Issue #792, opened 2018-08-01
Add NamedQubit.range(prefix, len)


Strilanc:
This makes more sense than LineQubit.range in contexts where connectivity is not linear.

`NamedQubit.range("a", 5)` would return `[NamedQubit("a0"), NamedQubit("a1"), NamedQubit("a2"), NamedQubit("a3"), NamedQubit("a4")]`

kevinsung:
What about making the prefix optional, so that `NamedQubit.range(5)` returns `[NamedQubit("0"), NamedQubit("1"), NamedQubit("2"), NamedQubit("3"), NamedQubit("4")]`?

Strilanc:
What use case did you have in mind where that would be useful?

A case I'm worried about is that NamedQubit.range(n) returns qubits that are impossible to tell apart from LineQubit.range(n) in circuit diagrams. Not necessarily bad, probably not common, but perhaps something to avoid making easy to do.

bryano:
Why not use `string.asciilowercase` instead?

Strilanc:
That's an interesting idea; just go a, b, c, ..., z, aa, ab, ... instead of numbering. Does feel more natural to the use case.

PR #818, opened 2018-08-07
Implement NamedQubit.range

Strilanc:
Fixes https://github.com/quantumlib/Cirq/issues/792

Current problem: the qubits don't sort into the correct order!


Strilanc:
Not happy with how this works yet.

Issue #824, opened 2018-08-07
Bug in assert_circuits_with_terminal_measurements_are_equivalent()


cduck:
Here are three different test cases that fail:

### Test case 1
```
from cirq import *
from cirq.google import *
import cirq


def test_test():
    a, b, d = LineQubit.range(3)
    c = Circuit.from_ops(
        #GateOperation(ExpWGate(half_turns=0.61351656, axis_half_turns=0.8034575038876517), (b,)),
        GateOperation(ExpZGate(), (a,)),
        GateOperation(ExpZGate(), (b,)),
        GateOperation(XmonMeasurementGate('all'), (a,)),)
    c2 = Circuit.from_ops(
        GateOperation(ExpZGate(), (b,)),
        GateOperation(XmonMeasurementGate('all'), (a,)),)
    c_orig = Circuit(c)

    #EjectZ().optimize_circuit(c)
    print(c)
    testing.assert_circuits_with_terminal_measurements_are_equivalent(
        c2, c, atol=1e-4)

    print(c)
```

### Test case 2
```
from cirq import *
from cirq.google import *
import cirq

def test_test():
    a, b, d = LineQubit.range(3)
    c = Circuit.from_ops(
        GateOperation(ExpWGate(half_turns=0.61351656, axis_half_turns=0.8034575038876517), (b,)),
        GateOperation(ExpZGate(half_turns=.5), (a,)),
        GateOperation(ExpZGate(half_turns=0.1), (b,)),
        GateOperation(XmonMeasurementGate('all'), (a, b)),
    )
    c2 = Circuit.from_ops(
        GateOperation(ExpWGate(half_turns=0.61351656, axis_half_turns=0.8034575038876517), (b,)),
        GateOperation(XmonMeasurementGate('all'), (a, b)),
    )
    print(
        dot(ExpZGate(half_turns=0.1).matrix(),
            ExpWGate(half_turns=0.61351656, axis_half_turns=0.803457503887651).matrix()))

    testing.assert_circuits_with_terminal_measurements_are_equivalent(
        c2, c, atol=1e-4)

    print(c)
```

### Test case 3
```
from cirq import *
from cirq.google import *
import cirq


def test_test():
    c = Circuit([
[...]

Issue #822, opened 2018-08-07
Rename ACTUAL/EXPECTED in assert_circuits_equivalent


Strilanc:
This is a bit misleading because it looks like the actual circuit should have bee the expected one, but really they just need to have similar functions. Maybe REFERENCE and ACTUAL?

Issue #828, opened 2018-08-08
Put failure information in AssertionError's message, instead of printing


Strilanc:
This more strongly associates the information with the failing test, instead of globbing it all together at the end. We have several testing methods that essentially say `if bad: print message`, but they should be `assert not bad, message`

Issue #813, opened 2018-08-06
Failing tests on windows


Strilanc:
I setup a dev environment on a windows machine, ran the tests, and got eight failures from three classes.

Class 1: Numpy doesn't have float128 on Windows? I tried a couple different versions of numpy and this attribute was consistently not present.

```
cirq\google\sim\xmon_simulator_test.py:363 (test_run_initial_state_ndarray_not_upconvertable[None])
cirq\google\sim\xmon_simulator_test.py:363 (test_run_initial_state_ndarray_not_upconvertable[moment_by_moment_schedule])
                    initial_state=np.array([0, 0, 1, 0],
>                                          dtype=np.float128))
E           AttributeError: module 'numpy' has no attribute 'float128'
```

Class 2: multiprocessing not working. We probably need to mark these tests as "don't run on windows":

```
cirq\google\sim\mem_manager_test.py:81 (test_with_multiprocessing_pool)
cirq\google\sim\mem_manager_test.py:92 (test_with_multiple_multiprocessing_pools)
cirq\google\sim\xmon_stepper_test.py:638 (test_large_circuit_unitary[2-True])
multiprocessing.pool.RemoteTraceback: 
    assert tup is not None
AssertionError
```

Class 3: differing circuits. Not sure why these would ever fail just because of an OS change.

```
cirq\contrib\paulistring\convert_to_pauli_string_phasors_test.py:22 (test_convert)
E       AssertionError: assert '0: ───[X]───...0.5────[Z]───' == '0: ───[X]────...0.5────[Z]───'
E         - 0: ───[X]──────────────────[Z]^0.125─────────
E         ?          ----------
E         + 0: ───[X]────────[Z]^0.125─────────
E           
E         - 1: ───[Y]^0.25───[Z]^0.0───[Y]^-0.5────[Z]───
E         ?               ----------
E         + 1: ───[Y]^0.25───[Y]^-0.5────[Z]───
```

```
cirq\contrib\paulistring\convert_to_pauli_string_phasors_test.py:44 (test_convert_keep_clifford)
E       AssertionError: assert '0: ───X─────...─H───────────' == '0: ───X──────...─H───────────'
E         - 0: ───X────────────────────[Z]^0.125───
E         ?        ----------
E         + 0: ───X──────────[Z]^0.125───
E           
E         - 1: ───[Y]^0.25───[Z]^0.0───H───────────
E         ?               ----------
E         + 1: ───[Y]^0.25───H───────────
```

```
[...]

PR #831, opened 2018-08-08
Fix assert_circuits_with_terminal_measurements_are_equivalent

Strilanc:
- The old strategy only worked when the chosen reference cells happened to correspond to linearly independent constraints
- New strategy just solves the linear system the hard way
- Fixed phase-correction qubit ordering being backwards
- Moved failure details into the error message instead of printing them to stdout

Fixes https://github.com/quantumlib/Cirq/issues/824
Fixes https://github.com/quantumlib/Cirq/issues/822
I think this fixes one of the failures in https://github.com/quantumlib/Cirq/issues/813
Part of https://github.com/quantumlib/Cirq/issues/828

Apologies to @cduck , who lost hours because of this method's subtle jankiness.


cirq/ops/gate_operation.py
```diff
@@ -85,7 +85,15 @@ def with_gate(self, new_gate: raw_types.Gate) -> 'GateOperation':
         return new_gate.on(*self.qubits)
 
     def __repr__(self):
-        return 'GateOperation({!r}, {!r})'.format(self.gate, self.qubits)
+        # Abbreviate when possible.
+        if self == self.gate.on(*self.qubits):
+            return '{!r}.on({})'.format(
+                self.gate,
+                ', '.join(repr(q) for q in self.qubits))
```
cduck: This can generate a repr that can't be evaulated.  Maybe use parenthesis if it's not a just letters.
```
>>> repr(cirq.X(q0) ** 0.5)
X**0.5.on(NamedQubit('q0'))
```

Strilanc: I think this should be a separate issue on `EigenGate` not using parens in its repr.

cduck: All the repr changes probably shouldn't be part of this PR anyways.

Strilanc: Done.

Issue #827, opened 2018-08-08
Include 'cirq.' in __repr__ outputs


Strilanc:
This makes the repr's a lot more useful, because you don't need to `from cirq import *` for them to work.

Strilanc:
Not that I am aware of.

vtomole:
Close for now? I haven't found any left after skimming.

PR #817, opened 2018-08-07
Implement OrderTester

Ashalynd:
Fixes #568

PR #833, opened 2018-08-08
Default qubit order now sorts by type then by whatever ordering is defined by each qubit type

### LINKED ISSUE #832: Update default qubit to first go by type name then ordering provided by qubit else str ordering
This allows qubits to specify custom orderings without violating the requirements of comparator methods when different qubit types interact within a circuit.

Strilanc:
- `QubitId` is now an abstract class. Child classes must implement `_comparison_key(self)`.
- `sorted(list_of_qubits)` now matches the behavior of `QubitOrder.DEFAULT.order_of(list_of_qubits)`.
- By default, `QubitId` sorts by type name, then by package, then by comparison key.
- Moved `NamedQubit` into its own file
- All instances of `cirq.QubitId()` or similar have been replaced with usage of a specific type of qubit.

~~Comes after #817~~

Fixes #832
Part of #827


cirq/ops/named_qubit.py
```diff
@@ -0,0 +1,93 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from cirq.ops import raw_types
+
+
+class NamedQubit(raw_types.QubitId):
+    """A qubit identified by name."""
+
```
dabacon: I'd document the ordering that this supports, since it is non-trivial.

Ashalynd:
#817 is now in the mergeable state, please review.

cirq/ops/named_qubit.py
Strilanc: Done.

Strilanc:
MergeBot is confused! It hurt itself in its confusion!

Strilanc:
I... killed one of the instances. But there's still two?

PR #734, opened 2018-07-20
[WIP] Replace KnownMatrix with a duck-typing based protocol

Strilanc:
- KnownMatrix.matrix() now returns None when the matrix is not known
- Classes that used to optionally implement KnownMatrix now unconditionally implement it and return None when no matrix
- Added KnownMatrix.has_matrix, static matrix_of, and static has_matrix_of.
- Replaced all usage of extensions with KnownMatrix with None checking
- Fixed cirq.dot failing when given a single argument

Part of https://github.com/quantumlib/Cirq/issues/672

PR #836, opened 2018-08-08
Define cirq.unitary and _unitary_, which will replace KnownMatrix

Strilanc:
- Any object can expose that it has a unitary matrix by defining a `_unitary_` method
- For objects that only sometimes have a valid matrix, `None` can be returned by `_unitary_` to indicate there is no matrix.
- `cirq.unitary(v)` returns the non-None result of `v._unitary_()` or else raises a `TypeError` if `v` doesn't have `_unitary_` or it returned `None`.
- `cirq.unitary(v, d)` will return `d` instead of raising a `TypeError` in failure cases.

Please comment on whether or not you think this is a good/bad replacement for KnownMatrix/extensions.

For an idea of what comes next, see https://github.com/quantumlib/Cirq/pull/734


Strilanc:
I have simplified the design. Take another look.

cirq/protocols/unitary.py
```diff
@@ -0,0 +1,88 @@
...
+# provides a different np.array([]) value.
+RaiseTypeErrorIfNotProvided = np.array([])
+
+TDefault = TypeVar('TDefault')
+
+
+class SupportsUnitary(Protocol):
+    """An object that may be describable by a unitary matrix."""
+
+    @abc.abstractmethod
```
maffoo: The [protocol examples](https://mypy.readthedocs.io/en/latest/protocols.html#defining-subprotocols-and-subclassing-protocols) in the mypy docs do not mark methods as `abstractmethod`; is there a reason you've done so here? This might mean that if you _explicitly_ inherit from `SupportsUnitary` instead of just relying on ducktyping, then when the method is not present you'll get an error at instance creation, rather than later when calling `cirq.unitary`. If that in fact works and we want to support explicit inheritance, we should document that and add some tests.

cirq/protocols/__init__.py
```diff
@@ -0,0 +1,4 @@
+from cirq.protocols.unitary import (
+    SupportsUnitary,
+    unitary,
+)
```
maffoo: It would be nice to have some module docs about the intent here. IIUC, the idea is that these "protocols" work like python builtin functions that are implemented by calling special `__methods__` that may or may not be defined on a particular object. We use mypy `Protocol` objects to describe our protocols to mypy and to allow explicit or implicit (ducktyped) implementation of those protocols. Also, we use single-underscoe `_methods_` to distinguish from python's built-in special methods.

cirq/protocols/unitary.py
```diff
@@ -0,0 +1,88 @@
...
+    def _unitary_(self) -> Optional[np.ndarray]:
```
maffoo: Would be helpful to give a brief example of where you'd want to define this method but sometimes return `None`. For example, might want this for parameterized gates that only have a matrix if all parameters are specified as numbers instead of symbols.

cirq/protocols/unitary.py
```diff
@@ -0,0 +1,88 @@
...
+    if result is not None:
+        return result
+    if default is not RaiseTypeErrorIfNotProvided:
+        return default
+
+    if get is None:
+        raise TypeError("object of type '{}' "
+                        "has no _unitary_ method.".format(type(val)))
+    raise TypeError("object of type '{}' does have a _unitary_ method, "
+                    "but it returned None.".format(type(val)))
```
maffoo: It might be helpful to include val itself in these error messages, not just its type.

cirq/protocols/unitary.py
```diff
@@ -0,0 +1,88 @@
...
+# whether or not the caller provided a 'default' argument. It must be of type
+# np.ndarray to ensure the method has the correct type signature in that case.
+# It is checked for using `is`, so it won't have a false positive if the user
+# provides a different np.array([]) value.
+RaiseTypeErrorIfNotProvided = np.array([])
+
+TDefault = TypeVar('TDefault')
+
+
+class SupportsUnitary(Protocol):
```
maffoo: Do we want to support runtime isinstance checks of protocols? https://mypy.readthedocs.io/en/latest/protocols.html#using-isinstance-with-protocols

cirq/protocols/unitary_test.py
```diff
@@ -0,0 +1,77 @@
...
+        _ = cirq.unitary(ReturnsNone())
+    assert cirq.unitary(ReturnsSome()) is m
+
+    assert cirq.unitary(NoMethod(), None) is None
+    assert cirq.unitary(ReturnsNone(), None) is None
+    assert cirq.unitary(ReturnsSome(), None) is m
+
+    assert cirq.unitary(NoMethod(), 1) == 1
+    assert cirq.unitary(ReturnsNone(), 1) == 1
+    assert cirq.unitary(ReturnsSome(), 1) is m
```
maffoo: Maybe add a test case with a false-y default value other than `None`, like `False` or `0`.

cirq/protocols/__init__.py
Strilanc: I don't think this is the right place to put those docs. How would the user end up here? It's far more likely they'd end up in `cirq/__init__.py` (by manual exploration) or in `cirq/protocols/unitary.py` (by go-to-definition).

I do think we will need a place in the documentation labelled "magic methods" or similar where these are all listed.

cirq/protocols/unitary.py
Strilanc: Done.

cirq/protocols/unitary.py
Strilanc: Acknowledged. I went with just the type because that's what python builtins do. E.g. when you call `len(5)` you get `TypeError: object of type 'int' has no len()`.

cirq/protocols/unitary.py
Strilanc: This made more sense when there were multiple methods and implementing one would make the others work. Now it makes so sense. Removed the abstractmethod.

cirq/protocols/unitary.py
Strilanc: Mostly I wanted mypy to understand the type. But yes I think it makes sense for the runtime check to work.

PR #837, opened 2018-08-08
Minor import cleanup in cirq/contrib and examples/

Strilanc:


examples/supremacy.py
```diff
@@ -17,10 +17,10 @@
 from typing import Callable, Iterable
 
 import cirq
-from cirq import devices, google
+import cirq.google as cg
```
dstrain115: Why import as cg?  Why not just as google?  Seems like it makes the code less readable.

Strilanc: It's for the same reason people import numpy as np.

Issue #829, opened 2018-08-08
Add __repr__ to XmonDevice


Strilanc:
Also, ensure that `repr(cirq.google.Foxtail)` is `'cirq.google.Foxtail'`.

PR #841, opened 2018-08-09
Add __repr__ to XmonDevice

vtomole:
`repr(cirq.google.Foxtail)` is now `cirq.google.Foxtail` based on https://github.com/quantumlib/Cirq/issues/829


vtomole:
@Strilanc Updated branch. Unrelated test fails.

cirq/google/xmon_device.py
```diff
@@ -195,6 +195,9 @@ def col(self, col: int) -> List[GridQubit]:
         """Returns the qubits in the given column, in ascending order."""
         return sorted(q for q in self.qubits if q.col == col)
 
+    def __repr__(self):
```
Strilanc: This repr should be printing out a string that you could paste into ipython in order to get another instance of the current xmondevice. Not all xmon devices are foxtail, so this current repr is incorrect.

https://github.com/quantumlib/Cirq/pull/840 added repr's specific to foxtail and bristlecone, so it's no longer necessary to worry about that.

Example of a correct `__repr__` from `TrialResult`:

```python
    def __repr__(self):
        return ('TrialResult(params={!r}, '
                'repetitions={!r}, '
                'measurements={!r})').format(self.params,
                                             self.repetitions,
                                             self.measurements)
```

vtomole: I knew it should be like that, but all XmonDevice has is 
```
self._measurement_duration
self._exp_w_duration 
self._exp_z_duration 
self.qubits
```
Where do I get the device's name?

vtomole: The GridQubits generated by this class are not the same every time it's instantiated. How should I write a test for it?

Strilanc:
Add a test to `xmon_device_test` with some simple parameters to cover the lines and fix the >80 char line lint warnings and we're good to go.

vtomole:
@Strilanc This is brutal, but I think i'm getting the hang of it :) . How's this?

cirq/google/xmon_device_test.py
```diff
@@ -55,6 +55,15 @@ def test_init():
         _ = d.duration_of(cirq.Gate().on(q00))
 
 
+@cirq.testing.only_test_in_python3
```
Strilanc: Because there are no strings in the `__repr__` (which is what causes the problems) you can omit this decorator.

cirq/google/xmon_device_test.py
```diff
@@ -55,6 +55,15 @@ def test_init():
...
+def test_repr():
+    d = square_device(2, 2, holes=[])
+
+    assert repr(d) == ("XmonDevice(measurement_duration=Duration(picos=1000), "
```
Strilanc: Almost there! Just need to also print out the qubits. Add `qubits={}` with value `sorted(self.qubits)` to the output. :+1:

Strilanc:
I'm going on vacation, so I've handed this review off to @dabacon .

vtomole:
@Strilanc Have a good break! :)

PR #845, opened 2018-08-10
Revert PauliString optimizer recombine short circuit

cduck:
Recombine would stop looking for possible placements too soon.


cduck:
Obsoleted by #847.

Issue #850, opened 2018-08-10
Docs not displaying .md links


vtomole:
For example, https://cirq.readthedocs.io/docs/development.md gives a `404` even though this page exists in the repository.

PR #851, opened 2018-08-10
Change link to development.md

vtomole:
Couldn't think of a better way to fix https://github.com/quantumlib/Cirq/issues/850 other than linking to the `.md` in the repo. Looking for feedback! :)


docs/install.md
```diff
@@ -6,7 +6,7 @@ Choose your operating system:
 - [Installing on Mac OS X](#installing-on-mac-os-x)
 - [Installing on Windows](#installing-on-windows)
 
-If you want to create a development environment, see [docs/development.md](/docs/development.md).
+If you want to create a development environment, see [docs/development.md](https://github.com/quantumlib/Cirq/blob/master/docs/development.md).
```
dabacon: The problem with the previous one was that it had the first "/".  Instead we can just use the fact that development.md is in this  directory.  Please change the part after the comma to
`see [these instructions](development.md).`
and this should fix this.

vtomole: Wouldn't that link to https://cirq.readthedocs.io/development.md? Which also doesn't exist.

PR #853, opened 2018-08-10
Support CircuitDag equality

### LINKED ISSUE #846: Implement equality for CircuitDag
Implement `__eq__` for CircuitDag using `networkx.is_isomorphic()`.  Use the node_match argument of is_isomorphic: `node_match=lambda n: n.val`.

This may be useful for #830.

cduck:
Resolves #846.

networkx.is_isomorphic() seems to require that nodes have a total ordering.  Right now I have only implemented `__lt__` for Unique.

Testing the ordering operators on Unique will be much easier once #568 is finished.

### DIFF
--- a/cirq/circuits/circuit_dag.py
+++ b/cirq/circuits/circuit_dag.py
@@ -14,6 +14,7 @@
 
 from typing import Any, Callable, Dict, Generic, Iterator, TypeVar
 
+import functools
 import networkx
 
 from cirq import ops, devices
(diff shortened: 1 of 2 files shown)


cirq/circuits/circuit_dag.py
```diff
@@ -38,6 +38,11 @@ def __init__(self, val: T) -> None:
     def __repr__(self):
         return 'Unique({}, {!r})'.format(id(self), self.val)
 
+    def __lt__(self, other):
```
dabacon: I think PEP8 likes you to implement all of these.  Maybe use functools.total_ordering()?

cduck: Done.

PR #854, opened 2018-08-12
Allow ParamResolver to resolve values into Symbols

kevinsung:
Here is the use case I have in mind. I want to define a parameterized operation. Then, I want to define something which has this parameterized operation repeated a bunch of times, but with each repetition introducing a new set of parameters. Concretely, the parameterized operation could be a Trotter step with angle names of the form `T_j` and I want to define a variational ansatz which has multiple repetitions, so the first repetition would have angles named `T_j-0` and the second angles named `T_j-1`, etc. I think a nice way to do this is to make the Trotter step a ParameterizableEffect and then use ParamResolvers to rename the Symbols as appropriate.

### DIFF
--- a/cirq/study/resolver.py
+++ b/cirq/study/resolver.py
@@ -32,7 +32,7 @@ class ParamResolver(object):
             assigned value.
     """
 
-    def __init__(self, param_dict: Dict[str, float]) -> None:
+    def __init__(self, param_dict: Dict[str, Union[float, Symbol]]) -> None:
         self.param_dict = param_dict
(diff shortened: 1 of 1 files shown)


Strilanc:
This breaks an invariant I've assumed so far, which is that resolving is idempotent. At the very least this change requires modifying all methods that do resolution to iteratively resolve variables until they hit bottom.

Strilanc:
Closed as stale, and also it has widespread impact on other code (e.g. need to do cycle detection) that is risky without thinking it through with the whole codebase in mind.

PR #863, opened 2018-08-15
Initial proposal for hardware specification

dabacon:


dabacon:
@mafoo sure you are swamped but if you have some spare cycles it would be great if you could look at this.

mafoo:
i think you have miss '@' me

dabacon:
Oops sorry off by a letter.

@maffoo sure you are swamped but if you have some spare cycles it would be great if you could look at this

cirq/api/google/v1/hardware.proto
```diff
@@ -0,0 +1,92 @@
...
+
+// Simple measurements of the performance of a gate.
+message Fidelities {
+  // If the gate is unitary, this is the average gate fidelity.
+  repeated float gate_fidelity = 1;
+
+  // The measurement results for a given preparation.
+  message Confusion {
+    // The key in this map is the measurement result and the value is the
+    // probability.  These may not sum to 1.0 due to leakage.
```
maffoo: I'm not sure that including a caveat here about leakage is particularly useful. Given that the measurements we have defined always return a 0 or 1, there is no _observable_ leakage from the user's perspective, just a probability of measuring state 0 as a 1 or vice-versa. Thus I would suggest replacing `confusion` with a `map <bool, float> measurement_error` that gives the observable measurement error for the 0 and 1 states, respectively.

(Note that for gate fidelities we are also hiding "leakage errors" inside a single gate fidelity number; we could of course expose a richer error model for both gates and measurements in the future.)

cirq/api/google/v1/hardware.proto
```diff
@@ -0,0 +1,92 @@
...
+// The performance of a calibrated.
+message CalibratedOperation {
+  // Which operation (gate plus placed qubits) was calibrated. If the gate is
+  // not parameterized then this will match exactly the prototype from the
+  // gate set. If the gate was parameterized, then this will be the
+  // actual parameter values (up to accuracy) that the calibration was done for.
+  Operation op = 1;
+
+  // Simple fidelities for the operation.
+  Fidelities fidelities = 2;
```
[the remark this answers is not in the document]
dabacon: OK, I reworked this to report randomized benchmarking, and to report errors (one minus fidelity).

cirq/api/google/v1/hardware.proto
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// Simple measurements of the performance of a gate.
+message Fidelities {
+  // If the gate is unitary, this is the average gate fidelity.
+  repeated float gate_fidelity = 1;
```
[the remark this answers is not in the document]
dabacon: removed

cirq/api/google/v1/hardware.proto
dabacon: OK, created a separate MeasruementError message

cirq/api/google/v1/hardware.proto
```diff
@@ -0,0 +1,114 @@
...
+  // ranges describing allowed parameter values and accuracies.  Note that
+  // multiple prototypes may have the same key and hence have the same
+  // parameter range.
+  map<string, ParameterRange> parameter_ranges = 2;
+}
+
+// The results of a randomized benchmarking protocol over a set of qubits.
+message RandomizedBenchmarking {
+  // The qubits over which the randomized benchmarking has been run.
+  repeated Qubit qubits = 1;
```
Strilanc: Aren't randomized benchmarking results specific to an operation, not a set of qubits? E.g. you can tell how much each CZ is hurting you.

cirq/api/google/v1/hardware.proto
```diff
@@ -0,0 +1,114 @@
...
+  float average_gate_error = 2;
+}
+
+// A simple representation of measurement errors on single qubits.
+message MeasurementError {
+  // Which qubit is being characterized.
+  Qubit qubit = 1;
+
+  // The probability that a given measurement values is incorrect.
+  map<bool, float> error = 2;
```
Strilanc: If we specified this as a confusion matrix, it would generalize to e.g. measuring the 2 state. Might be forward-looking to make it a map<int, ...>

cirq/api/google/v1/hardware.proto
```diff
@@ -0,0 +1,114 @@
...
+  // The resolution with which the gate can be offset (values of
+  // `incremental_delay_picoseconds` will be rounded to values in increments
+  // of this size).
+  uint64 timing_resolution_picoseconds = 2;
+}
+
+// The range that a parameter can take along with the accuracy with which
+// the parameter can be supplied.
+message ParameterRange {
+  // The lowest allowed value for the parameter.
```
Strilanc: Will parameters always be linear? Maybe we should switch from `step_size` to `distinct_value_count` and add an enum `interpolation` set to `linear`.

cirq/api/google/v1/hardware.proto
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// The supported operations on the hardware.
+message GateSet {
+  // Prototypes of the gate, including the qubits upon which the gate acts,
+  // and timing resolution. There is a prototype for every general gate
+  // that can be applied to specific qubits on the device (i.e. every single
+  // qubit gate on every qubit will have a prototype, and similar for every
+  // two qubit gate.)
+  repeated Prototype prototypes = 1;
```
Strilanc: `OperationPrototype`

cirq/api/google/v1/hardware.proto
```diff
@@ -0,0 +1,114 @@
...
+// The operations supported by the hardware and the current performance
+// characteristics of these operations.
+message HardwareSpec {
+  // The version name of the hardware.
+  //
+  // Currently supports: `bristlecone`.
+  string processor_name = 1;
+
+  // The supported operations on the hardware. This information is static
+  // and does not change for a given processor_name.
```
Strilanc: Even if there's a dead qubit?

cirq/api/google/v1/hardware.proto
dabacon: Yes, but @maffoo said: "We're probably going to want to quote things like average gate error over all single-qubit gates, as coming from randomized benchmarking, rather than breaking this down by specific operations."

dabacon:
This is stale and will be revisited soon.  So closing for now.

PR #864, opened 2018-08-16
Moving duration keyword for angles in from OpenFermion-Cirq

idk3:
Added `duration` (time-based) keyword to `chosen_angle_to_half_turns` and `chosen_angle_to_canonical_half_turns` from OpenFermion-Cirq. Should allow a fair amount of boilerplate removal from OpenFermion-Cirq.


idk3:
After discussion, we'll keep this only in OpenFermion-Cirq.

PR #866, opened 2018-08-18
Random circuits v2

sboixo:
Creates the random circuit instances in https://github.com/sboixo/GRCS
Some of these specific circuits where benchmarked in https://arxiv.org/abs/1807.10749


examples/examples_test.py
```diff
@@ -93,3 +94,30 @@ def test_example_runs_grover():
...
+    # We check that is exactly circuit inst_4x5_10_0
+    # in github.com/sboixo/GRCS cz_v2
+    assert len(circuit) == 11
+    assert sum( 1 for _ in circuit.findall_operations_with_gate_type(
+        cirq.ops.common_gates.Rot11Gate)) == 35
+    assert sum( 1 for _ in circuit.findall_operations_with_gate_type(
+        cirq.ops.common_gates.RotXGate)) == 15
+    assert sum( 1 for _ in circuit.findall_operations_with_gate_type(
+        cirq.ops.common_gates.RotYGate)) == 23
+    assert sum( 1 for _ in circuit.findall_operations_with_gate_type(
```
Strilanc: ``len(list(circuit.findall[...]))``

examples/examples_test.py
```diff
@@ -93,3 +94,30 @@ def test_example_runs_grover():
...
+def test_random_circuits_v2():
+    circuit = examples.random_circuits_v2.generate_random_v2_circuit_grid(
+        n_rows=4, n_cols=5, cz_depth=9, seed=0)
+    # We check that is exactly circuit inst_4x5_10_0
+    # in github.com/sboixo/GRCS cz_v2
+    assert len(circuit) == 11
+    assert sum( 1 for _ in circuit.findall_operations_with_gate_type(
+        cirq.ops.common_gates.Rot11Gate)) == 35
+    assert sum( 1 for _ in circuit.findall_operations_with_gate_type(
+        cirq.ops.common_gates.RotXGate)) == 15
```
Strilanc: Don't dig down when not needed; use e.g. `cirq.RotXGate` instead of `cirq.bla.bla.RotXGate`.

If moving to `experiments` you would need to `from cirq import ops` and use `ops.RotXGate` to avoid cyclic dependency issues.

examples/random_circuits_v2.py
```diff
@@ -0,0 +1,170 @@
+# Copyright 2018 The Cirq Developers
```
Strilanc: "random_circuits_v2" is too ambiguous of a name. There should be enough information in the name of the method to guess at which paper this method is implementing a circuit from. e.g. "generate_google_v2_random_circuit".

examples/random_circuits_v2.py
```diff
@@ -0,0 +1,170 @@
...
+    circuit.append(common_gates.H(qubit) for qubit in qubits)
+
+    return circuit
+
+
+def generate_random_v2_circuit_grid(n_rows: int, n_cols: int,
+                                    cz_depth: int, seed: int
+                                    ) -> cirq.Circuit:
+    """
+    Generates Google Random Circuits v2 as in github.com/sboixo/GRCS cz_v2.
```
Strilanc: Can you also link the relevant paper?

cirq/experiments/generate_google_v2_random_circuit.py
```diff
@@ -0,0 +1,171 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import random
+from typing import Callable, Iterable, TypeVar
+
+import cirq
+from cirq.circuits import InsertStrategy
+from cirq.devices.grid_qubit import GridQubit
+from cirq.ops import common_gates
```
Strilanc: `from cirq import ops`

When in a neighboring module, import at the module level instead of at the file level.

cirq/experiments/generate_google_v2_random_circuit.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def generate_random_v2_circuit(qubits: Iterable[GridQubit],
+                               cz_depth: int,
+                               seed: int) -> cirq.Circuit:
+
```
Strilanc: Add a docstring to this method.

cirq/experiments/generate_google_v2_random_circuit.py
```diff
@@ -0,0 +1,171 @@
...
+
+    ND_GATES = [common_gates.X**(1/2), common_gates.Y**(1/2)]
```
Strilanc: Is ND short for something?

Why is this variable all caps?

cirq/experiments/generate_google_v2_random_circuit.py
```diff
@@ -0,0 +1,171 @@
...
+                        circuit.append(common_gates.T(qubit),
+                                       strategy=InsertStrategy.EARLIEST)
+
+    # Add a final moment of Hadamards
+    circuit.append(common_gates.H(qubit) for qubit in qubits)
+
+    return circuit
+
+
+def generate_random_v2_circuit_grid(n_rows: int, n_cols: int,
```
Strilanc: `v2` isn't going to mean anything to people looking over the documentation. Rename this method to something like `generate_supremacy_circuit_1807_10749`.

cirq/experiments/generate_google_v2_random_circuit.py
```diff
@@ -0,0 +1,171 @@
...
+         ●─0─●─2─●─4─●─6─●─0─. . .
+        3│  7│  3│  7│  3│
+         .   .   .   .   .   .
+         .   .   .   .   .     .
+         .   .   .   .   .       .
+
+    Note that, for small devices, some layers will be empty because the layer
+    only contains edges not present on the device.
+
+    NOTE: This is the almost the function in supremacy.py,
```
Strilanc: This file should supplant the existing one. Delete that one, or replace its contents with a call to this one.

cirq/experiments/generate_google_v2_random_circuit.py
sboixo: v2 of course means v2, using the alternative you suggest will require a new paper if we want to update to v3

cirq/experiments/generate_google_v2_random_circuit.py
sboixo: Do you mean the one in examples/supremacy.py? 
That one does not have single qubits in neighboring qubits to two qubit gates, which is sometimes useful.

cirq/experiments/generate_google_v2_random_circuit.py
Strilanc: Sure, `v2` means `v2`, but in what context? What is is the second version of? Users aren't going to see this code already-knowing all of the surrounding context, they're going to see it when browsing through the various methods exported by cirq. And in that context `v2` makes you think that it refers to version 2 of *cirq* or of *this method within cirq*, instead of some external paper they've not heard of up to this point.

How about `generate_google_v2_supremacy_circuit` or `generate_supremacy_circuit_1807_10749_v2`?

`generate_random_v2_circuit_grid` just isn't a good enough name.

cirq/experiments/generate_google_v2_random_circuit.py
Strilanc: It's not too hard to put that back if needed. Unless we have an actual immediate use case for it, it should be replaced by this method.

cirq/experiments/generate_google_v2_random_circuit.py
sboixo: I am renaming this

cirq/experiments/generate_google_v2_random_circuit.py
sboixo: I think it would be better if you pull this without a deleting supremacy.py, and I also submit a pull deleting supremacy.py. Then this is easy to roll back. Is that OK?

Strilanc:
LGTM! Thanks for doing this.

PR #867, opened 2018-08-18
Implement wavefunction pretty printer

vtomole:
The wavefunction vector is sometimes hard to read. This procedure could help for debugging  and quantum programming in general.

Example:

```
import cirq

Q1 = cirq.GridQubit(0, 0)

circuit = cirq.Circuit.from_ops(cirq.H(Q1))
simulator = cirq.google.XmonSimulator()
result = simulator.simulate(circuit)

print(simulator.wavefunction(result))
```

### DIFF
--- a/cirq/google/sim/xmon_simulator.py
+++ b/cirq/google/sim/xmon_simulator.py
@@ -32,8 +32,9 @@ function, i.e. one can retrieve the final wave function from the simulation
 via.
     final_state = sim.simulate(circuit).final_state
 """
-
+import re
 import math
+import itertools
 import collections
 from typing import Dict, Iterator, List, Set, Union, cast
 from typing import Tuple  # pylint: disable=unused-import
@@ -166,6 +167,34 @@ class XmonSimulator:
         """
         self.options = options or XmonOptions()
 
+    def wavefunction(self, state):
+        """
+        Returns the wavefunction as a string in Dirac notation.
+
+        For example:
+        circuit = cirq.Circuit.from_ops(cirq.H(Q1))
+        simulator = cirq.google.XmonSimulator()
+        result = simulator.simulate(circuit)
+
+        print(wavefunction(result.final_state)) -> -0.71j|0> + -0.71j|1>
+        """
+        wvf = state.final_state
+        perm_list = ["".join(seq) for seq in itertools.product(
+            "01", repeat=int(np.log2(len(wvf))))]
+        wvf_string = ""
+
+        for x in range(0, len(perm_list)):
+            rounded_elem = np.around(wvf[x], decimals=2) + np.complex(0, 0)
+            rounded_elem = np.around(rounded_elem, decimals=2)
+
+            if rounded_elem != 0:
+                wvf_string += str(rounded_elem) + "|" + perm_list[x] + "> + "
+
+        wvf_string = re.split(r'(\s+)', wvf_string)[:-4]
+        wvf_string = "".join(wvf_string)
+
+        return wvf_string
+
     def run(
         self,
         circuit: Circuit,
@@ -236,9 +265,9 @@ class XmonSimulator:
         qubit_order = ops.QubitOrder.as_qubit_order(qubit_order)
         for param_resolver in param_resolvers:
             xmon_circuit, keys = self._to_xmon_circuit(
-                    circuit,
-                    param_resolver,
-                    extensions or xmon_gate_ext)
+                circuit,
+                param_resolver,
+                extensions or xmon_gate_ext)
             if xmon_circuit.are_all_measurements_terminal():
                 measurements = self._run_sweep_sample(xmon_circuit, repetitions,
                                                       qubit_order)
@@ -436,7 +465,7 @@ class XmonSimulator:
 
         # TODO: Use one optimization pass.
         xmon_circuit = circuit.with_parameters_resolved_by(
-                param_resolver, extensions)
+            param_resolver, extensions)
         converter.optimize_circuit(xmon_circuit)
         DropEmptyMoments().optimize_circuit(xmon_circuit)
         keys = find_measurement_keys(xmon_circuit)
@@ -516,7 +545,7 @@ def _simulator_iterator(
                 elif isinstance(gate, xmon_gates.XmonMeasurementGate):
                     if perform_measurements:
                         invert_mask = (
-                                gate.invert_mask or len(op.qubits) * (False,))
+                            gate.invert_mask or len(op.qubits) * (False,))
                         for qubit, invert in zip(op.qubits, invert_mask):
                             index = qubit_map[qubit]
                             result = stepper.simulate_measurement(index)
@@ -562,7 +591,7 @@ def _sample_measurements(circuit: Circuit,
         all_qubits.extend(op.qubits)
         current_index += len(op.qubits)
     sample = step_result.sample(all_qubits, repetitions)
-    return {k: [x[s:e] for x in sample] for k,(s, e) in bounds.items()}
+    return {k: [x[s:e] for x in sample] for k, (s, e) in bounds.items()}
 
 
 def find_measurement_keys(circuit: Circuit) -> Set[str]:
--- a/cirq/google/sim/xmon_simulator_test.py
+++ b/cirq/google/sim/xmon_simulator_test.py
@@ -94,6 +94,31 @@ def simulate(simulator, circuit, scheduler, **kw):
     return simulator.simulate(program, **kw)
 
 
+def test_wavefunction():
+    simulator = cirq.google.XmonSimulator()
+
+    hadamard = cirq.Circuit.from_ops(cirq.H(Q1))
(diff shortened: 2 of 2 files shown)


Strilanc:
For maximum pretty, use `⟩` instead of `>`.

vtomole:
Wow. It is pretty!

kevinsung:
You should actually use the character `⟩` in the source code. If you have trouble typing it out, copy and paste it.

vtomole:
@kevinsung Done.

cirq/google/sim/xmon_simulator.py
```diff
@@ -166,6 +167,35 @@ def __init__(self, options: XmonOptions = None) -> None:
...
+        For example:
+        circuit = cirq.Circuit.from_ops(cirq.H(Q1))
+        simulator = cirq.google.XmonSimulator()
+        result = simulator.simulate(circuit)
+
+        print(wavefunction(result.final_state)) -> -0.71j|0> + -0.71j|1>
+        """
+        wvf = state.final_state
+        perm_list = ["".join(seq) for seq in itertools.product(
+            "01", repeat=int(np.log2(len(wvf))))]
```
dabacon: I think you can use int's method "bit_length" here (i.e. wvf.bit_length())

cirq/google/sim/xmon_simulator.py
```diff
@@ -166,6 +167,35 @@ def __init__(self, options: XmonOptions = None) -> None:
         """
         self.options = options or XmonOptions()
 
+    def wavefunction(self, state, decimals=2):
```
dabacon: This is kind of a strange place to put the method since you use the simulator to produce the result and then use it to print it.  I think it would be better if this method was a method on the XmonSimulateTrialResult and also on the XmonStepResult, so that you don't have to pass in the result.  I'd make this a global function, as it is pretty useful even if you are just given a state, and then delegate to it in those two classes.  For the global function, it would also help if the name were more explicit, i.e. something like pretty_wavefunction or pretty_state.

cirq/google/sim/xmon_simulator.py
```diff
@@ -166,6 +167,35 @@ def __init__(self, options: XmonOptions = None) -> None:
...
+        result = simulator.simulate(circuit)
+
+        print(wavefunction(result.final_state)) -> -0.71j|0> + -0.71j|1>
+        """
+        wvf = state.final_state
+        perm_list = ["".join(seq) for seq in itertools.product(
+            "01", repeat=int(np.log2(len(wvf))))]
+        wvf_string = ""
+
+        for x in range(0, len(perm_list)):
```
dabacon: this could just be range(len(perm_list))

cirq/google/sim/xmon_simulator.py
```diff
@@ -166,6 +167,35 @@ def __init__(self, options: XmonOptions = None) -> None:
...
+            rounded_elem = np.around(wvf[x], decimals) + np.complex(0, 0)
+            rounded_elem = np.around(rounded_elem, decimals)
```
dabacon: I don't quite understand what the np.complex is doing here and why there are two arounds.

cirq/google/sim/xmon_simulator.py
```diff
@@ -166,6 +167,35 @@ def __init__(self, options: XmonOptions = None) -> None:
...
+    def wavefunction(self, state, decimals=2):
+        """
+        Returns the wavefunction as a string in Dirac notation.
+
+        For example:
+        circuit = cirq.Circuit.from_ops(cirq.H(Q1))
+        simulator = cirq.google.XmonSimulator()
+        result = simulator.simulate(circuit)
+
+        print(wavefunction(result.final_state)) -> -0.71j|0> + -0.71j|1>
```
dabacon: Does H really produce complex values here (i.e the j)?

cirq/google/sim/xmon_simulator.py
vtomole: Yeah. These lines are suspect. I should have left a comment or two justifying them. This was the only way I could think on how to handle `-0.`. `circuit = cirq.Circuit.from_ops(cirq.H(Q1))` prints `(-0-0.71j)|0⟩ + -0.71j|1⟩`. That negative 0 in front of `|0⟩ ` doesn't look good. To not print that `-0`, I use the fact that `-0 + 0 = 0` , so that `0` is not printed. The line doesn't need an np.complex, so I should change that to `rounded_elem = np.around(wvf[x], decimals) + 0.` The problem is that this line undos the rounding for some reason; making rounded_elem `-0.7099999785423279j|0⟩ + -0.7099999785423279j|1⟩`. That's why I call `round` again.

There has to be a better way! Just haven't thought of it.

cirq/google/sim/xmon_simulator.py
vtomole: Yes. It's kind of strange, but `abs(0.71j)^2` does get you `~0.5`.

Strilanc: It's because H is being decomposed into exp(i Y pi/2) * exp(i Z pi).

cirq/google/sim/xmon_simulator_test.py
```diff
@@ -94,6 +94,35 @@ def simulate(simulator, circuit, scheduler, **kw):
     return simulator.simulate(program, **kw)
 
 
+def test_wavefunction():
+    simulator = cirq.google.XmonSimulator()
+
+    hadamard = cirq.Circuit.from_ops(cirq.H(Q1))
+    result = simulator.simulate(hadamard)
```
Strilanc: Instead of using a gate, specify an initial state and simulate an empty circuit. This gives you complete control over what the output should be,

cirq/google/sim/xmon_simulator.py
vtomole: Maybe `rounded_elem = np.around(wvf[x], decimals) + 0.0` will eliminate the unnecessary round procedure call.

cirq/google/sim/xmon_simulator_test.py
dabacon: You need a circuit which acts on a qubit in order for the initial state to make sense (your example as no qubits, but an initial state of a single qubit)

vtomole: I want to simulate an empty circuit per Craig's suggestion. I'm assuming that's a circuit that doesn't call any gates (e.g "a circuit which acts on a qubit").

Strilanc:
You can specify a qubit order with unused qubits to pad it out.

Dave, do you think we should allow too-large superpositions as initial states? With the understanding that it was as if there was an extra anonymous qubit register.

vtomole:
Okay. I'm ready for the second round of feedback :)

cirq/google/sim/xmon_simulator.py
```diff
@@ -53,6 +54,37 @@
...
+def pretty_state(wvf, decimals=2):
+    """
+    Returns the wavefunction as a string in Dirac notation.
+
+    For example:
+    circuit = cirq.Circuit.from_ops(cirq.H(Q1))
+    simulator = cirq.google.XmonSimulator()
+    result = simulator.simulate(circuit)
+
+    print(wavefunction(result.final_state)) -> -0.71|0⟩ + -0.71|1⟩
```
dabacon: The doc now describes the result and not what this method does.  I'd just give an example where you pass in a simple wave function and pretty print it.

cirq/google/sim/xmon_simulator.py
```diff
@@ -53,6 +54,37 @@
 from cirq.study import ParamResolver, Sweep, Sweepable, TrialResult
 
 
+def pretty_state(wvf, decimals=2):
```
dabacon: since we call the object passed in here a state of final_state, I'd use state instead of wft

cirq/google/sim/xmon_simulator.py
vtomole: Oops! Forgot to update doc! Done now.

vtomole:
What request is that?

cirq/google/sim/xmon_simulator.py
```diff
@@ -53,6 +54,41 @@
...
+def pretty_state(state, decimals=2):
+    """
+    Returns the wavefunction as a string in Dirac notation.
+
+    For example:
+
+    Q1 = cirq.GridQubit(0, 0)
+
+    state = np.array([1/np.sqrt(2), 1/np.sqrt(2)], dtype=np.complex64)
+    simulator = cirq.google.XmonSimulator()
```
dabacon: it's still strange to reference the simulator here.  I would just shorten to document what happens on the input:

```
state = np.array([1/np.sqrt(2), 1/np.sqrt(2)], dtype=np.complex64)
print(pretty_state(state)) -> 0.71|0⟩ + 0.71|1⟩
```

vtomole:
@dabacon Anything else?

vtomole:
@dabacon My pleasure :)

PR #870, opened 2018-08-20
Use path.split instead of '..' hack.

dabacon:


docs/snippets_test.py
```diff
@@ -30,9 +30,8 @@
 def test_can_run_readme_code_snippets():
     # Get the contents of the README.md file at the project root.
     readme_path = os.path.join(
-        os.path.dirname(__file__),  # Start at this file's directory.
-        '..', 'docs',  # Hacky check that we're under docs/.
-        '..', 'README.rst')     # Get the readme one level up.
+        os.path.split(os.path.dirname(__file__))[0], 'README.rst')
```
Strilanc: Why are you accessing the first item? Does this work if someone runs pytest from a different directory than the git repo root?

dabacon: This works if you run it from root of the git repo and if you run it from docs or if you run it from any other directory (just tested this on my local machine).  Is there something odd about different platforms here?  The only way the first element does exist is if somehow you've moved this test to the root directory of your drive.

dabacon:
Ping @Strilanc

PR #875, opened 2018-08-22
Add eq, ne, and hash to Pauli

dabacon:
Equality test was passing because it was falling back to object equals and equality groups only had one element.

### DIFF
--- a/cirq/ops/pauli.py
+++ b/cirq/ops/pauli.py
@@ -39,6 +39,14 @@ class Pauli:
     def difference(self, second: 'Pauli') -> int:
         return (self._index - second._index + 1) % 3 - 1
 
+    def __eq__(self, other):
+        if not isinstance(other, type(self)):
+            return NotImplemented
+        return self._index == other._index
+
+    def __ne__(self, other):
+        return not self == other
+
     def __gt__(self, other):
         if not isinstance(other, type(self)):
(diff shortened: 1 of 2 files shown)


cirq/ops/pauli.py
```diff
@@ -39,6 +39,14 @@ def third(self, second: 'Pauli') -> 'Pauli':
     def difference(self, second: 'Pauli') -> int:
         return (self._index - second._index + 1) % 3 - 1
 
+    def __eq__(self, other):
+        if not isinstance(other, type(self)):
+            return NotImplemented
+        return self._index == other._index
+
+    def __ne__(self, other):
```
Ashalynd: Are you sure that this will also return NotImplemented for the objects of different types? Is there any test to verify that?

cduck: This is the normal way to do it.  If other is a different type, it will use `other.__eq__` and either return a bool or throw a TypeError.

cirq/ops/pauli.py
```diff
@@ -39,6 +39,14 @@ def third(self, second: 'Pauli') -> 'Pauli':
...
+        return not self == other
+
```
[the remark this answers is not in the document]
cduck: `__lt__` and `__gt__` are already defined although it is not a total ordering.

PR #876, opened 2018-08-23
cirq.H with EigenGate implementation Fixes #872

### LINKED ISSUE #872: Implement cirq.H with EigenGate so cirq.H**0.5 works
writing `cirq.H(ancilla_qubit) ** 0.5` gives the error:
"Expected a <class 'cirq.ops.gate_features.ExtrapolatableEffect'> but got cirq.H"

This is probably expected behavior but I don't understand it. Why isn't Hadamard extrapolatable?

vtomole:
Addressing: https://github.com/quantumlib/Cirq/issues/872

### DIFF
--- a/cirq/circuits/qasm_output_test.py
+++ b/cirq/circuits/qasm_output_test.py
@@ -46,7 +46,7 @@ def test_empty_circuit():
     q0, = _make_qubits(1)
     output = cirq.QasmOutput((), (q0,))
     assert (str(output) ==
-"""OPENQASM 2.0;
+            """OPENQASM 2.0;
 include "qelib1.inc";
 
 
@@ -57,11 +57,10 @@ qreg q[1];
 
 def test_header():
     q0, = _make_qubits(1)
-    output = cirq.QasmOutput((), (q0,), header=
-"""My test circuit
+    output = cirq.QasmOutput((), (q0,), header="""My test circuit
 Device: Bristlecone""")
     assert (str(output) ==
-"""// My test circuit
+            """// My test circuit
 // Device: Bristlecone
 
 OPENQASM 2.0;
@@ -72,13 +71,12 @@ include "qelib1.inc";
 qreg q[1];
 """)
 
-    output = cirq.QasmOutput((), (q0,), header=
-"""
+    output = cirq.QasmOutput((), (q0,), header="""
 My test circuit
 Device: Bristlecone
 """)
     assert (str(output) ==
-"""//
+            """//
 // My test circuit
 // Device: Bristlecone
 //
@@ -96,7 +94,7 @@ def test_single_gate_no_parameter():
     q0, = _make_qubits(1)
     output = cirq.QasmOutput((cirq.X(q0),), (q0,))
     assert (str(output) ==
-"""OPENQASM 2.0;
+            """OPENQASM 2.0;
 include "qelib1.inc";
 
 
@@ -112,7 +110,7 @@ def test_single_gate_with_parameter():
     q0, = _make_qubits(1)
     output = cirq.QasmOutput((cirq.X(q0) ** 0.25,), (q0,))
     assert (str(output) ==
-"""OPENQASM 2.0;
+            """OPENQASM 2.0;
 include "qelib1.inc";
 
 
@@ -124,11 +122,27 @@ rx(pi*0.25) q[0];
 """)
 
 
+def test_h_gate_with_parameter():
+    q0, = _make_qubits(1)
+    output = cirq.QasmOutput((cirq.H(q0) ** 0.25,), (q0,))
+    assert (str(output) ==
+            """OPENQASM 2.0;
+include "qelib1.inc";
+
+
+// Qubits: [q0]
+qreg q[1];
+
+
+rh(pi*0.25) q[0];
+""")
+
+
 def test_precision():
     q0, = _make_qubits(1)
     output = cirq.QasmOutput((cirq.X(q0) ** 0.1234567,), (q0,), precision=3)
     assert (str(output) ==
-"""OPENQASM 2.0;
+            """OPENQASM 2.0;
 include "qelib1.inc";
 
 
@@ -155,7 +169,7 @@ def test_save_to_file():
         with open(file_path, 'r') as f:
             file_content = f.read()
     assert (file_content ==
-"""OPENQASM 2.0;
+            """OPENQASM 2.0;
 include "qelib1.inc";
 
 
@@ -192,7 +206,6 @@ def _all_operations(q0, q1, q2, q3, q4, include_measurments=True):
     class DummyCompositeOperation(cirq.Operation, cirq.CompositeOperation):
         qubits = (q0,)
         with_qubits = NotImplemented
-
 
         def default_decompose(self):
             return cirq.X(self.qubits[0])
@@ -287,7 +300,7 @@ def test_output_unitary_same_as_qiskit():
     qiskit_unitary = result.get_unitary()
 
     cirq.testing.assert_allclose_up_to_global_phase(
-                    cirq_unitary, qiskit_unitary, rtol=1e-8, atol=1e-8)
+        cirq_unitary, qiskit_unitary, rtol=1e-8, atol=1e-8)
 
 
 def test_output_format():
@@ -301,7 +314,7 @@ def test_output_format():
                              precision=5)
     assert (filter_unpredictable_numbers(str(output)) ==
             filter_unpredictable_numbers(
-"""// Generated from Cirq
+        """// Generated from Cirq
 
 OPENQASM 2.0;
 include "qelib1.inc";
--- a/cirq/circuits/qasm_output_test.py
+++ b/cirq/circuits/qasm_output_test.py
@@ -134,7 +134,9 @@ include "qelib1.inc";
 qreg q[1];
 
 
-rh(pi*0.25) q[0];
+ry(pi*0.25) q[0];
+rx(pi*0.25) q[0];
+ry(pi*-0.25) q[0];
 """)
 
 
--- a/cirq/contrib/paulistring/convert_to_pauli_string_phasors.py
+++ b/cirq/contrib/paulistring/convert_to_pauli_string_phasors.py
@@ -66,13 +66,13 @@ class ConvertToPauliStringPhasors(PointOptimizer):
                                         mat: np.ndarray,
                                         qubit: ops.QubitId) -> ops.OP_TREE:
         rotations = decompositions.single_qubit_matrix_to_pauli_rotations(
-                                       mat, self.tolerance)
+            mat, self.tolerance)
         out_ops = []  # type: List[ops.Operation]
         for pauli, half_turns in rotations:
             if (self.keep_clifford
-                and self._tol.all_near_zero_mod(half_turns, 0.5)):
+                    and self._tol.all_near_zero_mod(half_turns, 0.5)):
                 cliff_gate = ops.CliffordGate.from_quarter_turns(
-                                    pauli, round(half_turns * 2))
+                    pauli, round(half_turns * 2))
                 if out_ops and not isinstance(out_ops[-1], PauliStringPhasor):
                     op = cast(ops.GateOperation, out_ops[-1])
                     gate = cast(ops.CliffordGate, op.gate)
@@ -83,7 +83,8 @@ class ConvertToPauliStringPhasors(PointOptimizer):
             else:
                 pauli_string = ops.PauliString.from_single(qubit, pauli)
                 out_ops.append(
-                    PauliStringPhasor(pauli_string, half_turns=half_turns))
+                    PauliStringPhasor(pauli_string,
+                                      half_turns=round(half_turns, 10)))
         return out_ops
 
     def _convert_one(self, op: ops.Operation) -> ops.OP_TREE:
@@ -93,14 +94,14 @@ class ConvertToPauliStringPhasors(PointOptimizer):
 
         if (self.keep_clifford
             and isinstance(op, ops.GateOperation)
-            and isinstance(op.gate, ops.CliffordGate)):
+                and isinstance(op.gate, ops.CliffordGate)):
             return op
 
         # Single qubit gate with known matrix?
         mat = self.extensions.try_cast(ops.KnownMatrix, op)
         if mat is not None and len(op.qubits) == 1:
             return self._matrix_to_pauli_string_phasors(
-                            mat.matrix(), op.qubits[0])
+                mat.matrix(), op.qubits[0])
 
         # Just let it be?
         if self.ignore_failures:
--- a/cirq/ops/common_gates.py
+++ b/cirq/ops/common_gates.py
@@ -524,11 +524,11 @@ class HGate(eigen_gate.EigenGate,
         return HGate(half_turns=exponent)
 
     def _eigen_components(self):
-        component0 = np.array([[(3 + 2 * np.sqrt(2)), (1 + np.sqrt(2))],
-                               [(1 + np.sqrt(2)), (1)]]) / 2 * (2 + np.sqrt(2))
+        component0 = np.divide(np.array([[(3 + 2 * np.sqrt(2)), (1 + np.sqrt(2))],
+                                         [(1 + np.sqrt(2)), (1)]]), 2 * (2 + np.sqrt(2)))
 
-        component1 = np.array([[(3 - 2 * np.sqrt(2)), (1 - np.sqrt(2))],
-                               [(1 - np.sqrt(2)), (1)]]) / 2 * (2 - np.sqrt(2))
+        component1 = np.divide(np.array([[(3 - 2 * np.sqrt(2)), (1 - np.sqrt(2))],
+                                         [(1 - np.sqrt(2)), (1)]]), 2 * (2 - np.sqrt(2)))
 
         return [(0, component0), (1, component1), ]
 
--- a/cirq/ops/common_gates.py
+++ b/cirq/ops/common_gates.py
@@ -524,11 +524,15 @@ class HGate(eigen_gate.EigenGate,
         return HGate(half_turns=exponent)
 
     def _eigen_components(self):
-        component0 = np.divide(np.array([[(3 + 2 * np.sqrt(2)), (1 + np.sqrt(2))],
-                                         [(1 + np.sqrt(2)), (1)]]), 2 * (2 + np.sqrt(2)))
+        component0 = np.divide(np.array([[(3 + 2 * np.sqrt(2)),
+                                          (1 + np.sqrt(2))],
+                                         [(1 + np.sqrt(2)), (1)]]),
+                               2 * (2 + np.sqrt(2)))
 
-        component1 = np.divide(np.array([[(3 - 2 * np.sqrt(2)), (1 - np.sqrt(2))],
-                                         [(1 - np.sqrt(2)), (1)]]), 2 * (2 - np.sqrt(2)))
+        component1 = np.divide(np.array([[(3 - 2 * np.sqrt(2)),
+                                          (1 - np.sqrt(2))],
+                                         [(1 - np.sqrt(2)), (1)]]),
+                               2 * (2 - np.sqrt(2)))
 
         return [(0, component0), (1, component1), ]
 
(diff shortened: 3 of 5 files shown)


vtomole:
Closing because too many tests fail. Need to rewrite.

cirq/ops/common_gates.py
```diff
@@ -490,17 +490,48 @@ def measure_each(*qubits: raw_types.QubitId,
...
+    def _canonical_exponent_period(self) -> Optional[float]:
+        return 2
+
+    def _with_exponent(self,
+                       exponent: Union[value.Symbol, float]) -> 'RotXGate':
+        return HGate(half_turns=exponent)
+
+    def _eigen_components(self):
+        return [
+            (0, np.array([[0.5, 0.5], [0.5, 0.5]])),
```
bryano: These are the eigencomponents for X. For H, the first one should be

[[3 + 2 sqrt(2), 1 + sqrt(2)], [1 + sqrt(2), 1]] / (2 * (2 + sqrt(2))

and something similar for the second.

vtomole: This is what I have 

```
def _eigen_components(self):
        two_plus_sqrt_two = 2 * (2 + np.sqrt(2))

        component = np.array([[(3 + 2 * np.sqrt(2)) / two_plus_sqrt_two, (1 + np.sqrt(2)) /
                               two_plus_sqrt_two], [(1 + np.sqrt(2)) / two_plus_sqrt_two, (1) / (2 * two_plus_sqrt_two)]])

        return [(0, component), (1, component), ]
```

cirq/ops/common_gates.py
```diff
@@ -515,6 +546,10 @@ def matrix(self):
         s = math.sqrt(0.5)
         return np.array([[s, s], [s, -s]])
 
+    def text_diagram_info(self, args: gate_features.TextDiagramInfoArgs
+                          ) -> gate_features.TextDiagramInfo:
+        return gate_features.TextDiagramInfo(('H',))
+
     def known_qasm_output(self,
```
Strilanc: The known_qasm_output needs to be updated to handle the non-unit-exponent case. I think returning None should be sufficient, but you should add a test in the qasm output code using H**0.5 to be sure.

bryano:
Testing the matrix for a few different values of half_turns should do it.

cirq/ops/common_gates.py
```diff
@@ -515,11 +551,19 @@ def matrix(self):
...
                           qubits: Tuple[raw_types.QubitId, ...],
                           args: gate_features.QasmOutputArgs) -> Optional[str]:
         args.validate_version('2.0')
-        return args.format('h {0};\n', qubits[0])
+        if self.half_turns == 1:
+            return args.format('h {0};\n', qubits[0])
+        else:
+            return args.format('rh({0:half_turns}) {1};\n',
```
Strilanc: I know the other cases don't do this yet, but could you add a comment linking to external documentation stating that `rh` exists and what it is?

cirq/ops/common_gates.py
```diff
@@ -490,17 +490,53 @@ def measure_each(*qubits: raw_types.QubitId,
...
+                              [(1 + np.sqrt(2)) / two_plus_sqrt_two,
+                               (1) / (2 * two_plus_sqrt_two)]])
+
+        return [(0, component), (1, component), ]
+
+    @property
+    def half_turns(self) -> Union[value.Symbol, float]:
+        return self._exponent
 
     def default_decompose(self, qubits):
```
Strilanc: default_decompose also needs to work in the general case

A possible decomposition is `Y**0.25, X**self.half_turns, Y**-0.25` (i.e. move X+Z onto X, rotate around X, move it back). I may have the negative sign in the wrong place.

cirq/ops/common_gates.py
vtomole: Should we change the name of this class from `HGate` to  `RotHGate`? It will break a couple of things, but that is what it does now.

Strilanc: I like @dabacon 's idea to rename to things like `HPowGate`

vtomole: `XPowGate`, `YPowGate` and `ZPowGate` too?

Strilanc: Yes, but don't do that in this PR.

vtomole: >Add a comment linking to external documentation stating that rh exists and what it is?

I haven't been able to find external documentation about `rh`. I;m not sure where to add it. Maybe in the `Single Qubit Gates` section? https://github.com/quantumlib/Cirq/blob/master/docs/api.rst

`rh` is qasm. Does a user need to know about it for something like `cirq.H(q0) ** 0.25` which compiles to `rh`? Maybe emulating how the Rotation gates are documented will suffice. 

For example:
`HPowGate(*, half_turns, float, …) | Fixed rotation around the X+Z axis of the Bloch sphere.`

vtomole:
`cirq/testing/random_circuit.py:67: error: Incompatible types in assignment (expression has type "Dict[EigenGate, int]", variable has type "Optional[Dict[Gate, int]]")` for `gate_domain = DEFAULT_GATE_DOMAIN`. I've tried tracking it down with no success so far.

cirq/ops/common_gates.py
```diff
@@ -515,11 +552,19 @@ def matrix(self):
         s = math.sqrt(0.5)
         return np.array([[s, s], [s, -s]])
```
kevinsung: The `matrix` method here should be deleted; it is implemented under EigenGate.

cirq/ops/common_gates.py
```diff
@@ -490,36 +490,76 @@ def measure_each(*qubits: raw_types.QubitId,
...
+        yield Y(q)**0.25,
+        yield X(q)**self.half_turns
+        yield Y(q)**-0.25
 
     def inverse(self):
         return self
 
-    def matrix(self):
-        """See base class."""
-        s = math.sqrt(0.5)
```
Strilanc: math is now an unused import

cirq/ops/common_gates.py
```diff
@@ -490,36 +490,76 @@ def measure_each(*qubits: raw_types.QubitId,
...
+
+    @property
+    def half_turns(self) -> Union[value.Symbol, float]:
+        return self._exponent
 
     def default_decompose(self, qubits):
         q = qubits[0]
-        yield Y(q)**0.5
-        yield X(q)
+        yield Y(q)**0.25,
```
Strilanc:
```python
if self._exponent == 1:
    yield Y(q)**0.5, X(q)
    return
```

cirq/ops/common_gates.py
Strilanc: How did you find out about `rh` in the first place if you never found any documentation on it?

cirq/ops/common_gates_test.py
```diff
@@ -152,6 +153,16 @@ def test_x_matrix():
                        np.array([[1 - 1j, 1 + 1j], [1 + 1j, 1 - 1j]]) / 2)
 
 
+def test_h_matrix():
+    ket_plus = np.array([[1/np.sqrt(2), 1/np.sqrt(2)],
```
Strilanc: Divide the whole array instead of each entry individually.

cirq/ops/common_gates_test.py
```diff
@@ -152,6 +153,16 @@ def test_x_matrix():
...
+                         [1/np.sqrt(2), -1/np.sqrt(2)]])
+
+    assert np.allclose(cirq.HGate(half_turns=1).matrix(), ket_plus)
+    assert np.allclose(cirq.HGate(half_turns=0.5).matrix(), ket_plus)
```
Strilanc: This assertion is incorrect.

cirq/ops/common_gates_test.py
```diff
@@ -152,6 +153,16 @@ def test_x_matrix():
                        np.array([[1 - 1j, 1 + 1j], [1 + 1j, 1 - 1j]]) / 2)
 
 
+def test_h_matrix():
+    ket_plus = np.array([[1/np.sqrt(2), 1/np.sqrt(2)],
+                         [1/np.sqrt(2), -1/np.sqrt(2)]])
+
```
Strilanc: A good one to test is that the square of (H**0.5).matrix() is nearly H.matrix().

cirq/ops/common_gates.py
```diff
@@ -490,36 +490,76 @@ def measure_each(*qubits: raw_types.QubitId,
...
+
+    def _eigen_components(self):
+        two_plus_sqrt_two = 2 * (2 + np.sqrt(2))
+
+        component = np.array([[(3 + 2 * np.sqrt(2)) / two_plus_sqrt_two,
+                               (1 + np.sqrt(2)) / two_plus_sqrt_two],
+                              [(1 + np.sqrt(2)) / two_plus_sqrt_two,
+                               (1) / (2 * two_plus_sqrt_two)]])
+
+        return [(0, component), (1, component), ]
```
Strilanc: This can't possibly be correct. The eigenspaces can't be the same.

cirq/ops/common_gates.py
vtomole: It didn't exist before.  This pull request is `rh`'s implementation hence the

```
if self.half_turns == 1:
            return args.format('h {0};\n', qubits[0])
        else:
            return args.format('rh({0:half_turns}) {1};\n',
                               self.half_turns, qubits[0])
```

Strilanc: What do you mean that "this is rh's implementation"? Is rh not an existing function in QASM? The QASM output code is supposed to produce valid QASM that can run anywhere; we can't just make up our own functions.

vtomole: Update: Hardcoded the generated code instead of calling `default_compose()` because that would require calling `QasmOutput()`  which can't called recursively as it generates the same header and qreg twice. This is what I was planning on doing

```
 def known_qasm_output(self,
                          qubits: Tuple[raw_types.QubitId, ...],
                          args: gate_features.QasmOutputArgs) -> Optional[str]:
        args.validate_version('2.0')

        lst = list(self.default_decompose(qubits))
        output = qasm_output.QasmOutput(lst, tuple(qubits))
        
        return str(output)
```

vtomole:
Ready for the second round.

cirq/ops/common_gates.py
```diff
@@ -524,11 +524,11 @@ def _with_exponent(self,
         return HGate(half_turns=exponent)
 
     def _eigen_components(self):
-        component0 = np.array([[(3 + 2 * np.sqrt(2)), (1 + np.sqrt(2))],
-                               [(1 + np.sqrt(2)), (1)]]) / 2 * (2 + np.sqrt(2))
+        component0 = np.divide(np.array([[(3 + 2 * np.sqrt(2)), (1 + np.sqrt(2))],
```
Strilanc: You can just divide with the binary operator.

cirq/ops/common_gates.py
```diff
@@ -349,8 +348,8 @@ class MeasurementGate(raw_types.Gate,
         key: The string key of the measurement.
         invert_mask: A list of values indicating whether the corresponding
             qubits should be flipped. The list's length must not be longer than
-            the number of qubits, but it is permitted to be shorted. Qubits with
-            indices past the end of the mask are not flipped.
+            the number of qubits, but it is permitted to be shorted.
+    Qubits with indices past the end of the mask are not flipped.
```
Strilanc: Why has this been moved out of the invert_mask description? It's describing invert_mask.

cirq/ops/common_gates.py
```diff
@@ -490,36 +489,85 @@ def measure_each(*qubits: raw_types.QubitId,
...
 
     def known_qasm_output(self,
                           qubits: Tuple[raw_types.QubitId, ...],
                           args: gate_features.QasmOutputArgs) -> Optional[str]:
         args.validate_version('2.0')
-        return args.format('h {0};\n', qubits[0])
+        if self.half_turns == 1:
+            return args.format('h {0};\n', qubits[0])
+        else:
+            return args.format('ry({0:half_turns}) {3};\nrx({1:half_turns})'
```
Strilanc: I would break the strings after the `\n` like so:

```python
return args.format('ry({0:half_turns}) {3};\n'
                   'rx({1:half_turns}) {3};\n'
                   ...
```

cirq/ops/common_gates_test.py
```diff
@@ -152,6 +153,22 @@ def test_x_matrix():
                        np.array([[1 - 1j, 1 + 1j], [1 + 1j, 1 - 1j]]) / 2)
 
 
+def test_h_matrix():
+    assert np.allclose(((cirq.H**0.5)**2.0).matrix(), (cirq.H).matrix())
```
Strilanc: This is not testing the matrix, it is testing that 0.5*2 = 1. You want something like this:

```python
sqrt = (cirq.H**0.5).matrix()
m = np.dot(sqrt, sqrt)
assert np.allclose(sqrt, cirq.H.matrix())
```

I would recommend also verifying that the handedness of the matrix is correct by checking the entries against the square root computed elsewhere.

This one is from quirk, by entering "X+Z" into the custom rotation gate and using an angle of 90 (with phase correction of 45) and then copying the raw values out of the address bar:

`{{0.8535534+0.1464466i,√⅛-√⅛i},{√⅛-√⅛i,0.1464466+0.8535534i}}`

Note: 0.853... = (1+sqrt(0.5))/2 and 0.146... = (1-sqrt(0.5))/2

Here's wolfram alpha:

image

cirq/ops/common_gates.py
vtomole: Dividing by binary operator is what I originally wrote, but it doesn't give the correct matrices.

```
component0 = (np.array([[(3 + 2 * np.sqrt(2)),
                                          (1 + np.sqrt(2))],
                                         [(1 + np.sqrt(2)),
                                          (1)]])) / 2 * (2 + np.sqrt(2))

        component1 = (np.array([[(3 - 2 * np.sqrt(2)),
                                          (1 - np.sqrt(2))],
                                         [(1 - np.sqrt(2)),
                                          (1)]])) / 2 * (2 - np.sqrt(2))
```

cirq/circuits/qasm_output_test.py
```diff
@@ -72,13 +71,12 @@ def test_header():
 qreg q[1];
 """)
 
-    output = cirq.QasmOutput((), (q0,), header=
-"""
+    output = cirq.QasmOutput((), (q0,), header="""
```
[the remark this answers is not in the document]
vtomole: Judging by the green check-mark on the tab, no.

cirq/contrib/paulistring/convert_to_pauli_string_phasors.py
```diff
@@ -93,14 +94,14 @@ def _convert_one(self, op: ops.Operation) -> ops.OP_TREE:
 
         if (self.keep_clifford
             and isinstance(op, ops.GateOperation)
-            and isinstance(op.gate, ops.CliffordGate)):
+                and isinstance(op.gate, ops.CliffordGate)):
```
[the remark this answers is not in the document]
vtomole: Actually, `autopep8` is correct because that `if` statement is broken down into one that is new lined.

cirq/ops/common_gates.py
vtomole: What has been moved? `Qubits with`? It's still describing `invert_mask`.

cirq/contrib/paulistring/convert_to_pauli_string_phasors.py
```diff
@@ -83,7 +83,8 @@ def _matrix_to_pauli_string_phasors(self,
             else:
                 pauli_string = ops.PauliString.from_single(qubit, pauli)
                 out_ops.append(
-                    PauliStringPhasor(pauli_string, half_turns=half_turns))
+                    PauliStringPhasor(pauli_string,
+                                      half_turns=round(half_turns, 10)))
```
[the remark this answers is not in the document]
Strilanc: Change the assertion to a `np.testing.assert_allclose`

cirq/ops/common_gates.py
Strilanc: The indentation now indicates it is about the method instead of about the parameter.

cirq/circuits/qasm_output_test.py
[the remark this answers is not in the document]
Strilanc: Hahaha, fair enough.

cirq/ops/common_gates_test.py
```diff
@@ -152,6 +153,24 @@ def test_x_matrix():
                        np.array([[1 - 1j, 1 + 1j], [1 + 1j, 1 - 1j]]) / 2)
 
 
+def test_h_matrix():
+    sqrt = (cirq.H**0.5).matrix()
+    m = np.dot(sqrt, sqrt)
+    assert np.allclose(m, cirq.H.matrix())
```
Strilanc: Specify `atol=1e-8` to avoid floating point brittleness.

cirq/contrib/paulistring/convert_to_pauli_string_phasors.py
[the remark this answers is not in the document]
vtomole: I was hesitant about changing a test. But now with the go-ahead, here we go!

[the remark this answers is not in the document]
Strilanc: Wait, really? And it started failing because of the changes we made? This is very surprising.

It must be that the matrix is slightly different, and the code is using the matrix decomposition. I'll need to think about this for a bit. We might need to add approximate equality to a lot of classes to properly fix this.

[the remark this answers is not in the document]
vtomole: >It must be that the matrix is slightly different, and the code is using the matrix decomposition

That's the reason for that almost insignificant rounding error? Here are the compared outputs side by side

```
cirq.Circuit(moments=[
    cirq.Moment(operations=[
        PauliStringPhasor({+, 0:Y}, half_turns=-0.5),
    ]),
    cirq.Moment(operations=[
        cirq.CZ.on(cirq.LineQubit(0), cirq.LineQubit(1)),
    ]),
    cirq.Moment(operations=[
        PauliStringPhasor({+, 0:Z}, half_turns=-0.12499999999999978),
    ]),
    cirq.Moment(operations=[
        PauliStringPhasor({+, 0:X}, half_turns=0.5),
    ]),
    cirq.Moment(operations=[
        PauliStringPhasor({+, 0:Z}, half_turns=0.5),
    ]),
])
cirq.Circuit(moments=[
    cirq.Moment(operations=[
        PauliStringPhasor({+, 0:Y}, half_turns=-0.5),
    ]),
    cirq.Moment(operations=[
        cirq.CZ.on(cirq.LineQubit(0), cirq.LineQubit(1)),
    ]),
    cirq.Moment(operations=[
        PauliStringPhasor({+, 0:Z}, half_turns=-0.125),
    ]),
    cirq.Moment(operations=[
        PauliStringPhasor({+, 0:X}, half_turns=0.5),
    ]),
    cirq.Moment(operations=[
        PauliStringPhasor({+, 0:Z}, half_turns=0.5),
    ]),
])
```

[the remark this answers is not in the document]
Strilanc: Well, previously we were making the matrix elements very directly as sqrt(0.5). Now we're doing stuff like adding 1 to sqrt(2) and then multiplying and adding stuff together. It could easily result in being off by an extra ulp, which could propagate through the gate decomposition code and result in an ever so slightly different phase.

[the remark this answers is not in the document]
Strilanc: If you want, you can put the "round" code back but add a message to the issue I opened mentioning that it needs to be removed when the approx circuit equality is added in.

Strilanc:
👍

vtomole:
@Strilanc Oh man. I had a feeling there had to be something. Modifying the test to do

```
assert np.allclose(c_opt.to_unitary_matrix(), c_expected.to_unitary_matrix())
```
Fixes everything without calling `round`. Should I send a new pull request with the reversion and this change to `test_optimize()`?

Strilanc:
But isn't the failure in the circuit comparison, not in the matrix comparison?

if you can cut the round call, go ahead and make a PR for that.

vtomole:
Yes. Thankfully this test can be implemented without circuit comparison.

Issue #880, opened 2018-08-27
Slow tests in acquaintance contrib package


dabacon:
16.54s call     cirq/contrib/acquaintance/executor_test.py::test_executor_random[8-5-gates33]
16.54s call     cirq/contrib/acquaintance/executor_test.py::test_executor_random[8-5-gates32]

PR #881, opened 2018-08-27
Removes expensive executor test

bryano:
Addresses #880.

The costly test was creating two 8-qubit unitaries to compare. Would using the simulator (and comparing only a sample of states) be significantly faster (i.e. fast enough)?


bryano:
The failing test seems unrelated to this PR.

PR #890, opened 2018-09-05
Describe EigenGate._eigen_components better

Strilanc:


cirq/ops/eigen_gate.py
```diff
@@ -64,23 +82,57 @@ def _with_exponent(self: TSelf,
...
+            The Pauli Z gate's eigencomponents are:
+
+                [
+                    (0, np.array([[1, 0],
+                                  [0, 0]])),
+                    (1, np.array([[0, 0],
+                                  [0, 1]])),
+                ]
+
+            Valid eigencomponents for Rz(π) = iZ are:
```
kevinsung: Rz(π) = exp(-i π Z / 2) is actually -iZ.

Strilanc: Oops. Fixed.

cirq/ops/eigen_gate.py
```diff
@@ -23,6 +23,24 @@
...
+        # The θ in λ = exp(i π θ) where λ is a unique eigenvalue. The exponent
+        # factor is used, instead of just a raw unit complex number, because it
+        # disambiguates several cases. For example, when λ=-1 you can set θ to
+        # -1 instead of +1 resulting in square root operations returning -i
+        # instead of +1.
+        ('eigenvalue_exponent_factor', float),
+
+        # The projection matrix onto the eigenspace of the eigenvalue. Must
+        # equal Σ_k |λ_k⟩⟨λ_k| where the |λ_k⟩ vectors form an orthonormal
+        # basis for the eigenspace.
```
dabacon: This one could probably do with an explicitly worked out example vector.

dabacon: Oh see you did this below.  OK.

PR #891, opened 2018-09-05
Use cirq.unitary instead of try_cast(KnownMatrix) in cirq/circuits

Strilanc:
~~Comes after https://github.com/quantumlib/Cirq/pull/836~~


cirq/circuits/qasm_output.py
```diff
@@ -277,20 +277,19 @@ def _write_operations(self,
...
-            if matrix_op is not None and len(op.qubits) == 1:
-                u_op = QasmUGate.from_matrix(matrix_op.matrix())(*op.qubits)
+            mat = protocols.unitary(op, None) if len(op.qubits) <= 2 else None
+            if mat is not None and len(op.qubits) == 1:
+                u_op = QasmUGate.from_matrix(mat).on(*op.qubits)
                 if top:
                     output_line_gap(1)
                     output('// {}\n'.format(comment))
-                output(u_op.known_qasm_output(self.args))
+                output(cast(str, u_op.known_qasm_output(self.args)))
```
maffoo: Why is this cast needed now?

Strilanc: It's because `known_qasm_output` is being called on the operation, not the gate, and the operation is a `GateOperation` which can't guarantee in general that it has qasm output. In this case it's guaranteed because of the underlying gate, but mypy has no way of seeing this.

PR #893, opened 2018-09-06
Use cirq.unitary instead of try_cast(KnownMatrix) in cirq/google

Strilanc:
~~Comes after https://github.com/quantumlib/Cirq/pull/836, https://github.com/quantumlib/Cirq/pull/895~~


cirq/google/decompositions_test.py
```diff
@@ -11,32 +11,27 @@
...
-    for gate in gates[1:]:
-        m = gate.matrix().dot(m)
-    return m
-
-
-def assert_gates_implement_unitary(gates, intended_effect):
-    actual_effect = _gates_to_matrix(gates)
+def assert_gates_implement_unitary(gates: List[cirq.SingleQubitGate],
+                                   intended_effect: np.ndarray):
+    actual_effect = cirq.dot(*reversed([cirq.unitary(g) for g in gates]))
```
maffoo: I would suggest doing:
```python
actual_effect = cirq.dot(*[cirq.unitary(g) for g in reversed(gates)])
```

cirq/google/merge_rotations.py
```diff
@@ -34,23 +33,21 @@ class MergeRotations(PointOptimizer):
     """Combines adjacent constant single-qubit rotations."""
 
     def __init__(self,
-                 tolerance: float = 1e-8,
-                 extensions = None) -> None:
+                 tolerance: float = 1e-8) -> None:
```
maffoo: nit: can fit on previous line.

cirq/google/merge_rotations.py
```diff
@@ -34,23 +33,21 @@ class MergeRotations(PointOptimizer):
...
         super().__init__()
         self.tolerance = tolerance
-        self.extensions = extensions or Extensions()
 
     def optimization_at(self, circuit, index, op):
         if len(op.qubits) != 1:
             return
 
-        indices, gates = self._scan_single_qubit_ops(circuit, index,
+        indices, opers = self._scan_single_qubit_ops(circuit, index,
```
maffoo: nit: `opers` sounds odd; I'd suggest `operations` instead.

cirq/google/merge_rotations.py
```diff
@@ -65,29 +62,28 @@ def _scan_single_qubit_ops(
...
             op = cast(ops.Operation, circuit.operation_at(qubit, index))
             if len(op.qubits) != 1:
                 break
-            operation = self.extensions.try_cast(ops.KnownMatrix, op)
-            if operation is None:
+            if protocols.unitary(op, None) is None:
                 break
             indices.append(index)
-            operations.append(operation)
+            operations.append(op)
```
maffoo: Might want to capture the unitary matrices here and return them as well. Then they don't have to be recomputed inside `_merge_rotations`. Could also compute the matrix product as we go to avoid storing all the matrices separately.

Strilanc: I would, but see the is-it-one-xmon-op check? Need the operation to do that check.

cirq/google/merge_rotations.py
Strilanc: Done.

cirq/google/merge_rotations.py
Strilanc: Done.

cirq/google/decompositions_test.py
Strilanc: Done. (You know this makes it not work on all iterables, right? But not relevant here.)

cirq/google/merge_rotations.py
maffoo: Yes, would have to return the list of ops _and_ the matrices (or the final matrix if you compute the matrix products as you go through the loop here).

PR #895, opened 2018-09-06
Fix cirq.dot(*args) failing when given a single matrix

Strilanc:


cirq/linalg/combinators.py
```diff
@@ -81,14 +81,16 @@ def dot(*values: Union[float, complex, np.ndarray, np.ndarray]
         ) -> Union[float, complex, np.ndarray, np.ndarray]:
```
maffoo: Why is `np.ndarray` specified twice in the input type and twice in the output type?

Alos, if you want to get fancy, could use `typing.overload` to more carefully specify the relationships between input and output types, but that's obviously not high priority.

Strilanc: Oh, that's just to make sure that it works.

...I have no idea why it's there twice. Fixed. Also fixed the fact that this would wrap singleton ints and floats into an array.

PR #898, opened 2018-09-06
Use cirq.unitary instead of try_cast(KnownMatrix) in contrib, decompositions, linalg, testing

Strilanc:


cirq/contrib/paulistring/convert_to_clifford_gates.py
```diff
@@ -89,9 +88,9 @@ def _convert_one(self, op: ops.Operation) -> ops.OP_TREE:
             return op
 
         # Single qubit gate with known matrix?
-        mat = self.extensions.try_cast(ops.KnownMatrix, op)
+        mat = protocols.unitary(op, None)
         if mat is not None and len(op.qubits) == 1:
```
maffoo: Might want to check `len(op.qubits)` first to avoid creating a potentially large matrix unneccessarily:
```python
if len(op.qubits) == 1:
    mat = protocols.unitary(op, None)
    if mat is not None:
        ...
```
Note this sort of thing will be much nicer once [pep 572](https://www.python.org/dev/peps/pep-0572/) is in:
```python
if len(op.qubits) == 1 and (mat := protocols.unitary(op, None)) is not None:
    # do something with mat
```
We just have to wait for python 3.8!

cirq/contrib/paulistring/convert_to_pauli_string_phasors.py
```diff
@@ -97,17 +97,17 @@ def _convert_one(self, op: ops.Operation) -> ops.OP_TREE:
             return op
 
         # Single qubit gate with known matrix?
-        mat = self.extensions.try_cast(ops.KnownMatrix, op)
+        mat = protocols.unitary(op, None)
         if mat is not None and len(op.qubits) == 1:
```
maffoo: See previous comment

cirq/decompositions_test.py
```diff
@@ -28,16 +28,10 @@ def _operations_to_matrix(operations, qubits):
...
-    m = gates[0].matrix()
-    for gate in gates[1:]:
-        m = gate.matrix().dot(m)
-    return m
-
 def assert_gates_implement_unitary(gates: Sequence[cirq.SingleQubitGate],
                                    intended_effect: np.ndarray,
                                    atol: float):
-    actual_effect = _gates_to_matrix(cast(Sequence[cirq.KnownMatrix], gates))
+    actual_effect = cirq.dot(*reversed([cirq.unitary(g) for g in gates]))
```
maffoo: I would suggest:
```python
actual_effect = cirq.dot(*[cirq.unitary(g) for g in reversed(gates)])
```

cirq/testing/sample_circuits_test.py
```diff
@@ -19,5 +19,5 @@ def test_nonoptimal_toffoli_circuit():
     q0, q1, q2 = cirq.LineQubit.range(3)
     cirq.testing.assert_allclose_up_to_global_phase(
         cirq.testing.nonoptimal_toffoli_circuit(q0, q1, q2).to_unitary_matrix(),
-        cirq.TOFFOLI(q0, q1, q2).matrix(),
+        cirq.unitary(cirq.TOFFOLI(q0, q1, q2)),
```
maffoo: Currently it seems that `cirq.unitary(cirq.TOFFOLI(q0, q2, q1))` (or any other permutation of the qubits) gives the same result here. This seems like a bad thing to me. It's not actually possible to understand these matrices, or to correctly combine them with matrices from other operations, unless we know the basis. I don't know the best way to deal with this, but I think it's something we need to reckon with because I could imagine this producing subtle bugs that are hard to track down. Could `cirq.unitary` take an extra argument which is the qubit basis in which to express the matrix? Or always return the matrix of an operation in some canonical basis with the involved qubits sorted? Or return an array that has extra information attached which gives the basis?

cirq/contrib/paulistring/convert_to_clifford_gates.py
Strilanc: Done.

cirq/contrib/paulistring/convert_to_pauli_string_phasors.py
Strilanc: Done.

cirq/decompositions_test.py
Strilanc: Done.

cirq/testing/sample_circuits_test.py
Strilanc: I don't think that's what we should do, because it would require every implementation of _unitary_ to take a basis. Presumably that's a list of qubits, but some of them don't have any qubits associated. We have to settle for the user using the object itself as a reference for what the basis is. In this case, they would have to combine the matrix with the qubits property.

PR #901, opened 2018-09-07
Remove round() call in convert_to_pauli_string_phasors.py and compare circuit.matrix() instead of circuit in test_optimize()

vtomole:


Strilanc:
Oh I see what you meant now. Actually, this is not the test that we want to perform. The goal was to check that we understood the output; it wasn't enough for it to just have the correct matrix (e.g. what if the method simply did nothing at all?)

You can see that the matrices are already being compared just above this check.

Oh well. I'm going to close this. Thanks for suggesting it, though!

PR #903, opened 2018-09-07
Convert remaining KnownMatrix children to using _unitary_

Strilanc:


cirq/google/xmon_gates.py
```diff
@@ -213,18 +212,11 @@ def to_proto_dict(self, *qubits):
...
-        return super().try_cast_to(desired_type, ext)
-
-    def has_matrix(self):
-        return not isinstance(self.half_turns, value.Symbol)
-
-    def matrix(self):
-        if not self.has_matrix():
-            raise ValueError("Don't have a known matrix.")
-        return protocols.unitary(ops.Rot11Gate(half_turns=self.half_turns))
+    def _unitary_(self) -> Optional[np.ndarray]:
```
maffoo: I thought the plan was to change the return type here to `Union[np.ndarray, type(NotImplemented)]`

cirq/google/xmon_gates.py
```diff
@@ -542,13 +527,11 @@ def inverse(self):
...
-    def has_matrix(self):
-        return not isinstance(self.half_turns, value.Symbol)
-
-    def matrix(self):
-        if not self.has_matrix():
-            raise ValueError("Don't have a known matrix.")
-        return np.diag([(-1j)**self.half_turns, 1j**self.half_turns])
+    def _unitary_(self) -> Optional[np.ndarray]:
+        if isinstance(self.half_turns, value.Symbol):
+            return None
```
maffoo: Why `None` here instead of `NotImplemented` like the methods above?

cirq/google/xmon_gates.py
Strilanc: Good catch. Done.

cirq/google/xmon_gates.py
Strilanc: Good catch. Done.

Strilanc:
@maffoo @dabacon take another look

cirq/circuits/convert_to_cz_and_single_gates_test.py
```diff
@@ -20,15 +20,15 @@
 
 
 def test_avoids_infinite_cycle_when_matrix_available():
-    class OtherX(cirq.Gate, cirq.KnownMatrix, cirq.CompositeGate):
-        def matrix(self):
+    class OtherX(cirq.Gate, cirq.CompositeGate):
+        def _unitary_(self):
```
dabacon: return type?  Here an elsewhere in this pull request.

PR #907, opened 2018-09-10
Add Deutsch's algorithm to examples.

dlyongemallo:
I needed an example for demonstration purposes which is less complex than existing examples, but also not completely trivial.


examples/deutsch.py
```diff
@@ -0,0 +1,79 @@
...
+
+import random
+
+import cirq
+from cirq.ops import H, X, CNOT, measure
+
+
+def main():
+    # Choose qubits to use.
+    q0, q1 = [cirq.LineQubit(i) for i in range(2)]
```
Strilanc: `cirq.LineQubit.range(2)`

examples/deutsch.py
```diff
@@ -0,0 +1,79 @@
...
+              │
+1: ───X───H───X─────────────────────
+Result f(0)⊕f(1):
+result=1
+"""
+
+import random
+
+import cirq
+from cirq.ops import H, X, CNOT, measure
```
Strilanc: Always import from the top level in tests and examples and external code. `from cirq import H, ...`

dlyongemallo:
How do I tell the incremental coverage test about code which is executed randomly? Every time I run the test, there's a 50% that each of the two yields inside `make_oracle` doesn't run (and so a 75% chance of the coverage test failing).

Strilanc:
Mark those lines as `# coverage: ignore`.

dlyongemallo:
Thanks, done.

examples/deutsch.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2018 The Cirq Developers
```
vtomole: You can remove this notice per: https://github.com/quantumlib/Cirq/blob/master/examples/README.rst

examples/deutsch.py
```diff
@@ -0,0 +1,94 @@
...
+
+Deutsch's algorithm is one of the simplest demonstrations of quantum parallelism
+and interference. It takes a black-box oracle implementing a Boolean function
+f(x), and determines whether f(0) and f(1) have the same parity using just one
+query.  This version of Deutsch's algorithm is a simplified and improved version
+from Nielsen and Chuang's textbook.
+
+=== REFERENCE ===
+
+Deutsch, David. "Quantum theory, the Church-Turing Principle and the universal
```
vtomole: Deutch's paper doesn't discuss the algorithm in terms of circuits which is what this implementation is based on. I suggest removing this even though this is where the algorithm originated.

examples/deutsch.py
```diff
@@ -0,0 +1,94 @@
...
+quantum computer." Proc. R. Soc. Lond. A, 400:97, 1985.
+
+Nielsen, Michael A. and Isaac L. Chuang. "Quantum Computation and Quantum
```
vtomole: A user might not have the book handy. We could reference a link. The Wikipedia page describing this algorithm: https://en.wikipedia.org/wiki/Deutsch%E2%80%93Jozsa_algorithm is correct. We could link to it. If you are not comfortable with that, there are lecture notes online we could link to.

vtomole (changes requested):

examples/deutsch.py
Strilanc: I think having the original reference is a good idea, but I wouldn't put it first.

examples/deutsch.py
Strilanc: Definitely link to wikipedia. Much more practical.

examples/deutsch.py
dlyongemallo: Done.

examples/deutsch.py
dlyongemallo: Okay, I've put the Wikipedia link first and the original paper second.

vtomole (approved):
Sorry for the wait. LGTM!

Strilanc:
@Strilanc resolve example conflict and merge

dlyongemallo:
*ping*

Strilanc (approved):

PR #909, opened 2018-09-11
Add InsertStrategy animations.

dlyongemallo:


dlyongemallo:
@dabacon Sorry for stealing your TODO, and I'm not really sure if this is what you were looking for, but I needed to make these animations anyway so I thought I might as well add them to the docs.

If you had something else in mind, feel free to ignore/close this PR.

Strilanc:
A major component missing from these animations is the "insert position". It's not clear why the operations stop in the middle of the diagram.

It's difficult to make an animation that increases understanding. At the moment I don't think these do that, except the one for `EARLIEST`. There are a lot of things you could try to improve them. For example, add labelled arrows showing how things are going to move / moving / have moved. Coloring new gates different from existing gates. Drawing a simplified boundary around existing gates. Showing the same situation in all four animations. Dropping the "- -" ellipsis. Making the circuits bigger, so as to make the details other than the motion and fitting less apparent. Etc etc etc.

bryano:
@dlyongemallo How did you make these animations?

dlyongemallo:
@bryano I used [gimp](https://www.gimp.org/).

dlyongemallo:
@Strilanc I just followed the examples used in the text because I needed to make some quick examples. I'll try to come up with a good example which can be used to illustrate all four cases.

Strilanc:
You used gimp? That's... that's a lot of manual work.

Strilanc:
Closed as inactive. Feel free to reopen later.

Issue #912, opened 2018-09-13
How does the Apache license work when it comes to attaching the boilerplate notice to a file?


vtomole:
I've read the licence, but I've noticed some files in `Cirq/examples/` like `quantum_fourier_transform.py` have the licence at the top of the file while files like `grover.py` don't.

Strilanc:
Not a lawyer, but I feel pretty strongly that the example code shouldn't be bogged down with license notifications. Why would we care if people copy-and-modify example code? That's what it's for.

dabacon:
We should maybe add a README about this in examples?

Strilanc:
That sounds reasonable. We could also mention things like "must run unconverted", "not included in the packaged library", etc.

PR #913, opened 2018-09-13
Add licences to example files that don't have one

vtomole:
Fixes https://github.com/quantumlib/Cirq/issues/912


vtomole:
Sure. The thing is that some files have notices and some don't. @Strilanc should it be all or none?

Strilanc:
None.

examples/README.rst
```diff
@@ -0,0 +1 @@
+The software in this directory (Cirq/examples/) is not included in the packaged library. Anyone is free to copy, modify, or publish this software.
```
Strilanc: "this example code" instead of "this software". And add "use" in the list of things they're allowed to do.

vtomole: Done.

PR #915, opened 2018-09-14
schnjaso

schnjaso2:


dabacon:
Hi @schnjaso2 this looks like you are trying to merge back master into the version that was used for a release?  I think this branch should be thought of as staying bristine to the pypi version.  See release.md for our procedure for this.

Strilanc:
We definitely don't want to merge later changes into an older version, so this is definitely not something we are going to merge. I am going to close this PR because of that. Feel free to continue discussion of course.

Issue #830, opened 2018-08-08
Add cirq.testing.assert_same_diagram


Strilanc:
This method would do two things:

1. Strip the diagrams for you, so you didn't have to worry about trailing newlines in the code.
2. Give better error messages than python's default string error message.

For point 2, the main thing that is required is to append the actual diagram and reference diagram without abbreviation. Pytest is far too abbreviation heavy for the diagrams.

Secondarily for point 2, a better diff should be included. The information is 2d instead of line-based, and the diff should be similarly 2d. Show a copy of the actual diagram with any character positions that differ from the reference diagram replaced by ■ or some other obtrusive char.

kevinsung:
For use cases I have encountered, I often want to ignore differences due to insertion strategies, i.e., I'd want to shove all gates to the left before comparing.

Strilanc:
That would be better described as a circuit equivalence rather than a diagram equivalence.

bryano:
`CircuitDag` would be good for that.

cduck:
Probably, but it would have to support customizing node equality.  The matching in the two graphs wouldn't compare equal by default.

cduck:
You can do it with is_isomorphic.

PR #917, opened 2018-09-14
Add cirq.testing.assert_same_diagram.

dlyongemallo:
Fixes https://github.com/quantumlib/Cirq/issues/830.

### DIFF
--- a/cirq/testing/circuit_compare.py
+++ b/cirq/testing/circuit_compare.py
@@ -165,24 +165,19 @@ def assert_circuits_with_terminal_measurements_are_equivalent(
     )
 
 
-def assert_same_diagram(
+def assert_has_diagram(
         actual: circuits.Circuit,
         desired: str,
-        use_unicode_characters: bool = True,
-        transpose: bool = False) -> None:
+        **kwargs) -> None:
     """Determines if a given circuit has the desired text diagram.
 
     Args:
         actual: The circuit that was actually computed by some process.
         desired: The desired text diagram as a string. Whitespace at the
             beginning and end are ignored.
-        use_unicode_characters: Determines if unicode characters are
-            allowed (as opposed to ascii-only diagrams).
-        transpose: Arranges qubit wires vertically instead of horizontally.
+        **kwargs: Keyword arguments to be passed to actual.to_text_diagram().
     """
-    actual_diagram = actual.to_text_diagram(
-        use_unicode_characters=use_unicode_characters, transpose=transpose
-    ).strip()
+    actual_diagram = actual.to_text_diagram(**kwargs).strip()
     desired_diagram = desired.strip()
     assert actual_diagram == desired_diagram, (
         "Circuit's text diagram differs from the desired diagram.\n"
@@ -193,7 +188,7 @@ def assert_same_diagram(
         'Desired text diagram:\n'
         '{}\n'
         '\n'
-        'Diff:\n'
+        'Highlighted differences:\n'
         '{}\n'.format(actual_diagram, desired_diagram,
                       _text_diagram_diff(actual_diagram, desired_diagram))
     )
--- a/cirq/testing/circuit_compare.py
+++ b/cirq/testing/circuit_compare.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from typing import Tuple, Sequence, TYPE_CHECKING
+from itertools import zip_longest
 
 import numpy as np
 
@@ -67,7 +68,8 @@ def _cancel_qubit_phase(m1: np.ndarray,
 
     # Dominant row phase differences.
     for row in range(n):
-        col = max(range(n), key=lambda c: min(abs(m1[row, c]), abs(m2[row, c])))
+        col = max(range(n), key=lambda c: min(
+            abs(m1[row, c]), abs(m2[row, c])))
         prob[row, -1] = np.angle(m1[row, col]) - np.angle(m2[row, col])
 
     # Gram-Schmidt.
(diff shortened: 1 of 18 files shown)


cirq/testing/circuit_compare_test.py
```diff
@@ -190,3 +190,41 @@ def test_known_old_failure():
...
+1: ───Z───
+
+Diff:
+0: ───@───
+      │
+1: ───█───
+
+"""
+
+    try:
```
Strilanc: Use `pytest.raises` instead of a manual try-except.

cirq/testing/circuit_compare.py
```diff
@@ -148,3 +163,37 @@ def assert_circuits_with_terminal_measurements_are_equivalent(
...
+    assert actual_diagram == desired_diagram, (
+        "Circuit's text diagram differs from the desired diagram.\n"
+        '\n'
+        'Diagram of actual circuit:\n'
+        '{}\n'
+        '\n'
+        'Desired text diagram:\n'
+        '{}\n'
+        '\n'
+        'Diff:\n'
```
Strilanc: How about `Highlighted differences:`?

cirq/testing/circuit_compare.py
```diff
@@ -148,3 +163,37 @@ def assert_circuits_with_terminal_measurements_are_equivalent(
         'Diagram of reference circuit with desired function:\n'
         '{}\n'.format(actual, reference)
     )
+
+
+def assert_same_diagram(
```
Strilanc: Rename to `assert_has_diagram`. I know that's the name I asked for, but it didn't occur to me that `same` is misleading since the two arguments have differing types.

cirq/testing/circuit_compare.py
```diff
@@ -148,3 +163,37 @@ def assert_circuits_with_terminal_measurements_are_equivalent(
...
+        actual: circuits.Circuit,
+        desired: str,
+        use_unicode_characters: bool = True,
```
Strilanc: Add a `*` before `use_unicode_characters` to force keyword arguments on the options.

You may want to take a `**kwargs` argument and forward it into `to_text_diagram`, instead of listing specific ones, to future-proof against features being added to the text diagram methods.

cirq/testing/circuit_compare.py
```diff
@@ -117,6 +119,19 @@ def _canonicalize_up_to_terminal_measurement_phase(
     return matrix1, matrix2
 
 
+def _text_diagram_diff(
+        actual_diagram: str,
+        desired_diagram: str) -> str:
+    diff = ""
```
Strilanc: Nit: use single-quotes consistently.

Strilanc:
It looks like `zip_longest` isn't present in python 2. You'll have to use something else.

dlyongemallo:
Python 2 has izip_longest. Is there a way to specify this should be used during the transpilation?

Strilanc:
We could search replace `itertools.zip_longest` with `itertools.izip_longest` in the conversion script.

cirq/testing/circuit_compare.py
```diff
@@ -13,6 +13,10 @@
 # limitations under the License.
 
 from typing import Tuple, Sequence, TYPE_CHECKING
+try:
```
Strilanc: This is far too hacky of a way to do this IMO.

I would recommend doing the following:

`import itertools`

use `itertools.zip_longest`

in `python2.7-generate.sh`, add a loop that search-replaces `itertools.zip_longest` with `itertools.izip_longest`. (Mimic the loop that does the `__str__` -> `__unicode__` translation)

dlyongemallo:
`E       UnicodeEncodeError: 'ascii' codec can't encode characters in position 88-90: ordinal not in range(128)`

Hmm, it seems I can't use `█` and am restricted to range(128) in Python 2 if I return `str`.

Strilanc:
Where is that being triggered? In general, we try to make sure unicode works.

dlyongemallo:
Strings don't automatically handle unicode in python2 so we'd have to change the test for python2 (and change `_text_diagram_diff` to return `unicode`) if we want it to run.

I've marked the test `@cirq.testing.only_test_in_python3`. Let me know if you want me to modify the test instead.

dlyongemallo:
So one way I can fix this is to wrap the `_text_diagram_diff` in a private class with a `__str__` method, and ditto for the assignment of `expected_error` in `circuit_compare_test.py`, so that the `python2.7-generate.sh` would take care of the unicode encoding. That seems kludgy though.

Strilanc:
It doesn't make sense that this specific test would have a failure when all the other ones work fine. We do quite a lot of tests involving unicode strings! There is something unique about this one that I don't understand that is triggering the issue.

Strilanc:
It's something to do with `pytest.match`

Strilanc:
Here's my current workaround:

```python
    with pytest.raises(AssertionError) as ex_info:
        cirq.testing.assert_has_diagram(circuit, u"""
0: ───@───
      │
1: ───Z───
""")
    assert expected_error in ex_info.value.args[0]
```

Strilanc:
This is a very strangely behaved thing. This passes:

```python
    with pytest.raises(AssertionError, match=u'█'.encode('utf-8')):
        raise AssertionError(u'█'.encode('utf-8'))
```

Whereas this fails:

```python
    with pytest.raises(AssertionError, match=u'█'.encode('utf-8')):
        assert False, u'█'.encode('utf-8')
```

dlyongemallo:
Could be a bug in pytest. It looks similar to this one: https://github.com/pytest-dev/pytest/issues/877

dlyongemallo:
I've added a commit with your work around.

(btw is there a way to do a diffbase in github, i.e., work on top of an uncommitted PR? This particular change is super useful for doing other stuff, e.g., #918.)

Strilanc:
The closest thing is to create a PR that merges into the other branch. Unfortunately, this means that none of the branch protections kick in (e.g. requiring a review, running status checks).

Issue #911, opened 2018-09-12
Add annotation that denotes Moments to ascii diagram


dabacon:
When the ascii diagram tool cannot show gates in a Moment without overlapping, it breaks them out so you can see them.  But then you cannot really see by looking whether they are in the same moment or not.  Suggestion: we can add something at the bottom to group moments?

```circuit = cirq.Circuit.from_ops(cirq.CZ(q0, q3), cirq.CZ(q1, q2))
print(circuit)
print(repr(circuit))

0: ───@───────
      │
1: ───┼───@───
      │   │
2: ───┼───@───
      │
3: ───@───────
cirq.Circuit(moments=[
    cirq.Moment(operations=[
        cirq.CZ.on(cirq.LineQubit(0), cirq.LineQubit(3)),
        cirq.CZ.on(cirq.LineQubit(1), cirq.LineQubit(2)),
    ]),
])
```

Maybe something like
```
0: ───@───────
      │
1: ───┼───@───
      │   │
2: ───┼───@───
      │
3: ───@───────

      +---+
```
Or another suggestion is to tighten the padding for a single moment.

Strilanc:
I think the current implementation has bugs. The following code produces a circuit diagram where the "padding forces" are not lining up with the indicators:

```python
import cirq.experiments.google_v2_supremacy_circuit as x
print(x.google_v2_supremacy_circuit_bristlecone(7, 4, 0))
```

This is the diagram that comes out:

```
                  ┌──────────┐
(0, 5): ───H───T────────────────────────@───────X^0.5─────────H───
                                        │
(0, 6): ───H───T────────────────────────@───────@─────────────H───
                                                │
(1, 4): ───H───T───@────────────────────X^0.5───┼────T────────H───
                   │                            │
(1, 5): ───H───T───┼────────────────────────────┼─────────────H───
                   │                            │
(1, 6): ───H───@───┼────@───────────────Y^0.5───@─────────────H───
               │   │    │
(1, 7): ───H───@───┼────┼───────Y^0.5───T─────────────────────H───
                   │    │
(2, 3): ───H───T───┼────┼───────────────────────@─────────────H───
                   │    │                       │
(2, 4): ───H───@───@────┼───────────────X^0.5───┼────T────────H───
               │        │                       │
(2, 5): ───H───@───Y^0.5┼───────────────@───────┼────@────────H───
                        │               │       │    │
(2, 6): ───H───T────────@───────────────@───────┼────┼Y^0.5───H───
[...]

PR #918, opened 2018-09-17
Add annotation that denotes columns belonging to the same Moment to text diagrams.

dlyongemallo:
Fixes https://github.com/quantumlib/Cirq/issues/911.


cirq/circuits/circuit_test.py
```diff
@@ -2113,3 +2113,45 @@ def test_save_qasm():
...
+        cirq.H.on(d),
+        cirq.CZ.on(a, d),
+        cirq.CZ.on(b, c),
+        cirq.CZ.on(a, c),
+        cirq.CZ.on(b, d),
+        cirq.H.on(b),
+        cirq.H.on(c),
+    )
+
+    assert circuit.to_text_diagram() == """
```
dlyongemallo: Reminder to myself: These should use cirq.testing.assert_same_diagram from https://github.com/quantumlib/Cirq/pull/917.

Strilanc:
This is a good start, but it's not visually clear enough yet.

1. The indicators themselves are not clear enough. They are too far below the circuit, and don't show clear boundaries.
2. There would also need to be an indicator above the circuit.
3. I think we should also drop any spacing (or maybe set spacing to 1) between within-moment columns.

Reducing the spacing may be sufficient. It's hard to give concrete direction because the test I'm applying is "know it when I see it" in terms of clarity.

dabacon:
We should try just tightening the spacing to see what that looks like.

If we go with the annotation, I what about "+----+"?

dlyongemallo:
You mean "+----+" in ascii? What about in non-ascii ("unicode") mode?

The problem with using the "+" sign is that if you remove the empty row, it looks like it connects to the previous row. It looks too much like the "crossing" character (┼), and in fact is used for the crossing in `use_unicode_characters=False` mode.

For the record, this is the version of text diagram produced the above comments are referring to (so we can discuss changes to it):
```
0: ───H───@───────@───────────
          │       │
1: ───────┼───@───┼───@───H───
          │   │   │   │
2: ───────┼───@───@───┼───H───
          │           │
3: ───H───@───────────@───────

          ╶───╴   ╶───╴
```

Strilanc:
Those look better.

Try having the indicators overshoot the column by one on each side, and using `|------|` instead of `+-----+`. I think the spacing of 0 works best, and that it won't be ambiguous because no symbol starts with or ends with | or +.

dlyongemallo:
Unicode version, overshooting by one on each side, with various intra-`Moment` column spacings:
```
         ⇤─────⇥ ⇤─────⇥
0: ───H───@───────@───────────
          │       │
1: ───────┼───@───┼───@───H───
          │   │   │   │
2: ───────┼───@───@───┼───H───
          │           │
3: ───H───@───────────@───────
         ⇤─────⇥ ⇤─────⇥

         ⇤────⇥ ⇤────⇥
0: ───H───@──────@──────────
          │      │
1: ───────┼──@───┼──@───H───
          │  │   │  │
2: ───────┼──@───@──┼───H───
          │         │
3: ───H───@─────────@───────
         ⇤────⇥ ⇤────⇥

         ⇤───⇥ ⇤───⇥
0: ───H───@─────@─────────
          │     │
1: ───────┼─@───┼─@───H───
          │ │   │ │
2: ───────┼─@───@─┼───H───
          │       │
3: ───H───@───────@───────
         ⇤───⇥ ⇤───⇥

         ⇤──⇥ ⇤──⇥
0: ───H───@────@────────
          │    │
1: ───────┼@───┼@───H───
          ││   ││
2: ───────┼@───@┼───H───
          │     │
3: ───H───@─────@───────
         ⇤──⇥ ⇤──⇥
```

dlyongemallo:
The ascii version using `+`:
```
         +-----+ +-----+
0: ---H---@-------@-----------
          |       |
1: -------+---@---+---@---H---
          |   |   |   |
2: -------+---@---@---+---H---
          |           |
3: ---H---@-----------@-------
         +-----+ +-----+

         +----+ +----+
0: ---H---@------@----------
          |      |
1: -------+--@---+--@---H---
          |  |   |  |
2: -------+--@---@--+---H---
          |         |
3: ---H---@---------@-------
         +----+ +----+

         +---+ +---+
0: ---H---@-----@---------
          |     |
1: -------+-@---+-@---H---
          | |   | |
2: -------+-@---@-+---H---
          |       |
3: ---H---@-------@-------
         +---+ +---+

         +--+ +--+
0: ---H---@----@--------
          |    |
1: -------+@---+@---H---
          ||   ||
2: -------+@---@+---H---
          |     |
3: ---H---@-----@-------
         +--+ +--+
```

Using `|` instead of `+`:
```
         |-----| |-----|
0: ---H---@-------@-----------
          |       |
1: -------+---@---+---@---H---
          |   |   |   |
2: -------+---@---@---+---H---
          |           |
3: ---H---@-----------@-------
         |-----| |-----|

         |----| |----|
0: ---H---@------@----------
          |      |
1: -------+--@---+--@---H---
          |  |   |  |
2: -------+--@---@--+---H---
          |         |
3: ---H---@---------@-------
         |----| |----|

         |---| |---|
0: ---H---@-----@---------
          |     |
1: -------+-@---+-@---H---
          | |   | |
2: -------+-@---@-+---H---
          |       |
3: ---H---@-------@-------
         |---| |---|

         |--| |--|
0: ---H---@----@--------
          |    |
1: -------+@---+@---H---
          ||   ||
2: -------+@---@+---H---
          |     |
3: ---H---@-----@-------
         |--| |--|
```

Strilanc:
I think the `|` with spacing 0 looks great.

For the unicode version you would use box drawing characters ┤and ├.

dlyongemallo:
Okay, I was avoiding re-using `┤`and `├` because they're already ascii-mapped to `<` and `>`, respectively, in `_BoxChars`, but I can work around that.

Strilanc:
Those aren't actually used anywhere yet. Feel free to override that mapping.

This is ambiguity of symbols is a problem we have to fix eventually.

dlyongemallo:
What about the transpose diagram?

dlyongemallo:
It gets quite complicated when the gates are more than one character:
```
         ├──┤ ├───────────────┤
0: ───H───@────@─────────────────────
          │    │
1: ───────┼@───┼────@────────────────
          ││   │    │
2: ───────┼@───┼────┼────@───────H───
          │    │    │    │
3: ───H───@────X^0.5┼────┼───────────
                    │    │
4: ─────────────────X^0.5┼───────H───
                         │
5: ──────────────────────X^0.5───────
         ├──┤ ├───────────────┤
```

And the transpose:
```
  0 1 2 3     4     5
  │ │ │ │     │     │
  H │ │ H     │     │
  │ │ │ │     │     │
┬ @─┼─┼─@     │     │
┴ │ @─@ │     │     │
  │ │ │ │     │     │
┬ @─┼─┼─X^0.5 │     │     ┬
│ │ │ │ │     │     │     │
│ │ @─┼─┼─────X^0.5 │     │
│ │ │ │ │     │     │     │
┴ │ │ @─┼─────┼─────X^0.5 ┴
  │ │ │ │     │     │
  │ │ H │     H     │
  │ │ │ │     │     │
```

dlyongemallo:
I've made the logic change for the overshoots.

cirq/circuits/text_diagram_drawer.py
```diff
@@ -253,7 +339,11 @@ def _pad_into_multiline(width: int,
         col_width = max(1, max(len(sub_row[col])
                                for row in multiline_grid
                                for sub_row in row))
-        col_width += horizontal_spacing
+        if groups_are_vertical:
+            col_width += horizontal_spacing
+        else:
+            col_width = _pad_border(border, col, col_width, horizontal_spacing,
```
Strilanc: This makes the previous `col_width` assignment a dead write in this case. Move that previous assignment into the `if`'s positive case.

cirq/circuits/text_diagram_drawer.py
```diff
@@ -221,20 +253,74 @@ def _cross_char(use_ascii: bool, horizontal_emph: bool, vertical_emph: bool
...
     multiline_grid = []  # type: List[List[List[str]]]
+    border = []
 
     # Vertical padding.
     for row in range(len(grid)):
         multiline_cells = [cell.split('\n') for cell in grid[row]]
         row_height = max(1, max(len(cell) for cell in multiline_cells))
-        row_height += vertical_spacing
+        if groups_are_vertical:
+            row_height = _pad_border(border, row, row_height, vertical_spacing,
```
Strilanc: This makes the previous `row_height` assignment a dead write in this case. Move that previous assignment into the `if`'s positive case.

cirq/circuits/text_diagram_drawer.py
```diff
@@ -171,13 +190,26 @@ def render(self,
         for (x, y), v in self.entries.items():
             grid[y][x] = v
 
+        # Prepare Moment groups.
+        if self.groups_are_vertical:
```
Strilanc: Maybe when defining the characters, we should define their transpose. That would allow these to be merged together and might be more generally useful.

cirq/circuits/text_diagram_drawer.py
Strilanc: Oops. Obsoleted by my huge design change comment.

cirq/circuits/text_diagram_drawer.py
Strilanc: Oops. Obsoleted by my huge design change comment.

cirq/circuits/text_diagram_drawer.py
Strilanc: Oops. Obsoleted by my huge design change comment.

dlyongemallo:
I agree that this approach is conceptually much better and cleaner. However, there are two issues with writing the "end caps" of the `Moment` intervals using `TextDiagramDrawer.write`. One is, how should the "overshoots" be handled? The second is, if they're written as text, then they can't be ascii-fied in the `TextDiagramDrawer` itself, the way that wires (vertical and horizontal lines) are.

Strilanc:
For the overshoots, you can add an extra column, on each side of the moment, into the diagram.

The diagram-creating code knows whether or not unicode is being used, so it can just "if unicode then this else that". Another idea is to add yet more parameters to `write`, including `ascii` and `ascii_transpose`.

cirq/circuits/circuit.py
```diff
@@ -1314,6 +1324,49 @@ def _draw_moment_in_diagram(moment: Moment,
...
+    h = out_diagram.height()
+
+    left_end = '├' if use_unicode_characters else '|'
+    top_end = '┬' if use_unicode_characters else '-'
+    right_end = '┤' if use_unicode_characters else '|'
+    bottom_end = '┴' if use_unicode_characters else '-'
+
+    # Insert columns starting from the back since the insertion
+    # affects subsequent indices.
+    for (x1, x2) in reversed(moment_groups):
```
dlyongemallo: The efficiency of this loop can be improved by doing all the padding at once. I'll leave it this way for now as it's easier to reason about. I can make the improvement later.

dlyongemallo:
Okay, I've moved the circuit-specific code out of `TextDiagramDrawer` and into `Circuit` as suggested.

cirq/circuits/circuit.py
Strilanc: Nit: unnecessary parens.

cirq/circuits/text_diagram_drawer.py
```diff
@@ -126,6 +138,42 @@ def height(self) -> int:
             max_y = max(max_y, v.y1, v.y2)
         return 1 + max_y
 
+    def force_horizontal_padding_after(self, index, padding) -> None:
```
Strilanc: Add doc strings.

cirq/circuits/text_diagram_drawer.py
```diff
@@ -126,6 +138,42 @@ def height(self) -> int:
...
+        self.horizontal_padding[index] = padding
+
+    def force_vertical_padding_after(self, index, padding) -> None:
+        self.vertical_padding[index] = padding
+
+    def shift_horizontally(self, index, offset) -> None:
```
Strilanc: Create a helper method `_transform_coordinates(func: Callable[[int, int], Tuple[int ,int]])` and implement both shift methods using the helper.

Strilanc: I think the name of this method could be better. Maybe `insert_empty_columns(x: int, amount: int = 1)`?

cirq/circuits/text_diagram_drawer.py
```diff
@@ -37,15 +37,22 @@ class TextDiagramDrawer:
 
     def __init__(self):
         self.entries = dict()  # type: Dict[Tuple[int, int], str]
+        self.transposed_entries = dict()  # type: Dict[Tuple[int, int], str]
```
Strilanc: Instead of adding a second dictionary, change the `entries` dictionary to have type `Dict[Tuple[int, int], Tuple[str, str]]`. Then, instead of using `Tuple[str, str]`, use a `NamedTuple`.

This will come in very handy if we add more entry value types.

cirq/circuits/text_diagram_drawer.py
```diff
@@ -37,15 +37,22 @@ class TextDiagramDrawer:
...
         self.vertical_lines = []  # type: List[_VerticalLine]
         self.horizontal_lines = []  # type: List[_HorizontalLine]
+        self.horizontal_padding = {}  # type: Dict[int, int]
+        self.vertical_padding = {}  # type: Dict[int, int]
 
-    def write(self, x: int, y: int, text: str):
+    def write(self, x: int, y: int, text: str, transposed_text: str = None):
```
Strilanc: Add `Args:` documentation now that the arguments are non-trivial.

cirq/circuits/text_diagram_drawer.py
```diff
@@ -37,15 +37,22 @@ class TextDiagramDrawer:
...
         """Adds text to the given location."""
+        if transposed_text == None:
```
Strilanc: Use `is None` instead of `== None`

cirq/circuits/text_diagram_drawer.py
```diff
@@ -37,15 +37,22 @@ class TextDiagramDrawer:
...
+            transposed_text = text
```
Strilanc:
```python
cur = self.entries.get((x, y), ('', ''))
self.entries[(x, y)] = Entry(cur.text + text, cur.transposed_text + transposed_text)
```

cirq/circuits/circuit.py
```diff
@@ -1314,6 +1324,49 @@ def _draw_moment_in_diagram(moment: Moment,
...
+        out_diagram.shift_horizontally(x2 + 1, 1)
+        out_diagram.force_horizontal_padding_after(x2, 0)
+
+        for y in [0, h]:
+            out_diagram.write(x2 + 1, y, right_end, bottom_end)
+        out_diagram.force_horizontal_padding_after(x2 + 1,
+            2 if not transpose else 0)
```
Strilanc: Why does this depend on being transposed or not?

cirq/circuits/circuit.py
```diff
@@ -1314,6 +1324,49 @@ def _draw_moment_in_diagram(moment: Moment,
...
+
+        for y in [0, h]:
+            out_diagram.horizontal_line(y, x1, x2 + 1)
+
+        out_diagram.shift_horizontally(x1, 1)
+        out_diagram.force_horizontal_padding_after(x1, 0)
+        for y in [0, h]:
```
Strilanc: Can the three y loops be fused into one?

cirq/circuits/circuit_test.py
```diff
@@ -2111,3 +2111,50 @@ def test_save_qasm():
...
+        cirq.CZ.on(b, c),
+        (cirq.CNOT**0.5).on(a, d),
+        (cirq.CNOT**0.5).on(b, e),
+        (cirq.CNOT**0.5).on(c, f),
+        cirq.H.on(c),
+        cirq.H.on(e),
+    )
+
+    cirq.testing.assert_has_diagram(circuit, """
+         ├──┤  ├───────────────┤
```
Strilanc: Use box corner characters instead of box side characters.

```
┌───┐

└───┘
```

Strilanc:
@dlyongemallo any issues holding you up? I can take over the PR from here if you'd like.

cirq/circuits/circuit.py
dlyongemallo: Not really, since the diagram changes in-between. I guess it's possible, but would involve a rewrite of the outer loop.

cirq/circuits/circuit.py
dlyongemallo: This is needed to match the spacing logic in `to_text_diagram`, which sets `horizontal_spacing=1 if transpose else 3`.

cirq/circuits/circuit_test.py
```diff
@@ -2425,3 +2425,50 @@ def test_reachable_frontier_from():
...
+│ │ @─@ │     │     │     │
+└ │ │ │ │     │     │     ┘
+┌ │ │ │ │     │     │     ┐
+│ @─┼─┼─X^0.5 │     │     │
+│ │ @─┼─┼─────X^0.5 │     │
+│ │ │ @─┼─────┼─────X^0.5 │
+└ │ │ │ │     │     │     ┘
+  │ │ H │     H     │
+  │ │ │ │     │     │
+""", transpose=True)
```
Strilanc: Add a test of the ascii behavior, tranposed and not transposed

cirq/circuits/text_diagram_drawer.py
```diff
@@ -126,6 +144,45 @@ def height(self) -> int:
...
 
+    def force_horizontal_padding_after(self, index: int, padding: int) -> None:
+        """Change the padding after the given column."""
+        self.horizontal_padding[index] = padding
+
+    def force_vertical_padding_after(self, index: int, padding: int) -> None:
+        """Change the padding after the given row."""
+        self.vertical_padding[index] = padding
+
+    def _transform_coordinates(
```
Strilanc: This is much nicer now.

cirq/circuits/text_diagram_drawer.py
```diff
@@ -126,6 +144,45 @@ def height(self) -> int:
...
+            for y, x1, x2, emph in self.horizontal_lines]
+        self.horizontal_padding = {func_x(x): padding
+            for x, padding in self.horizontal_padding.items()}
+        self.vertical_padding = {func_y(y): padding
+            for y, padding in self.vertical_padding.items()}
+
+    def insert_empty_columns(self, x: int, amount: int = 1) -> None:
+        """Insert a number of columns after the given column."""
+        def transform_columns(column: int, row: int) -> Tuple[int, int]:
+            return (column + (amount if column >= x else 0), row)
```
Strilanc: Nit: unnecessary parens

cirq/circuits/text_diagram_drawer.py
```diff
@@ -126,6 +144,45 @@ def height(self) -> int:
...
+        self._transform_coordinates(transform_columns)
+
+    def insert_empty_rows(self, y: int, amount: int = 1) -> None:
+        """Insert a number of rows after the given row."""
+        def transform_rows(column: int, row: int) -> Tuple[int, int]:
+            return (column, row + (amount if row >= y else 0))
```
Strilanc: Nit: unnecessary parens

cirq/circuits/circuit_test.py
dlyongemallo: Done. I noticed that the intersections weren't `+` as I would've expected, but `|` in the regular diagram and `-` in the transpose diagram. However, this is existing behaviour not introduced by this PR (see `test_to_text_diagram_teleportation_to_diagram` for example). 

Is this a bug?

Strilanc: No, that's on purpose. Subjectively, I found reading diagrams with +s harder than diagrams with |s. There was too much visual noise or something (which wasn't a problem in the unicode version because there's no whitespace breaks between the characters).

cirq/circuits/circuit_test.py
```diff
@@ -2425,3 +2425,84 @@ def test_reachable_frontier_from():
...
+
+    cirq.testing.assert_has_diagram(circuit, """
```
Strilanc: Use `r` strings to avoid double `\\`.

PR #919, opened 2018-09-17
Fix deprecation warning for np.sum with generator.

dlyongemallo:
Background:
https://docs.scipy.org/doc/numpy/release.html#numpy-1-15-0-release-notes
"Giving a generator to numpy.sum is now deprecated. This was undocumented behavior, but worked. Previously, it would calculate the sum of the generator expression. In the future, it might return a different result. Use np.sum(np.from_iter(generator)) or the built-in Python sum instead."

### DIFF
--- a/cirq/ops/eigen_gate.py
+++ b/cirq/ops/eigen_gate.py
@@ -134,8 +134,8 @@ class EigenGate(raw_types.Gate,
         if self.is_parameterized():
             raise ValueError("Parameterized. Don't have a known matrix.")
         e = cast(float, self._exponent)
-        return np.sum(1j**(half_turns * e * 2) * component
-                      for half_turns, component in self._eigen_components())
+        return sum(1j**(half_turns * e * 2) * component
+                   for half_turns, component in self._eigen_components())
 
     def extrapolate_effect(self: TSelf,
                            factor: Union[float, value.Symbol]) -> TSelf:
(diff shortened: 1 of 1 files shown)


cirq/ops/eigen_gate.py
```diff
@@ -134,8 +134,8 @@ def matrix(self) -> np.ndarray:
         if self.is_parameterized():
             raise ValueError("Parameterized. Don't have a known matrix.")
         e = cast(float, self._exponent)
-        return np.sum(1j**(half_turns * e * 2) * component
-                      for half_turns, component in self._eigen_components())
+        return sum(1j**(half_turns * e * 2) * component
```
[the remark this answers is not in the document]
dlyongemallo: Done.

Though I'm not sure what numpy was actually doing behind the scenes. It might have just been implicitly wrapping the generator in a list, or it might have been summing using the generator. It's a tradeoff between space and speed, possibly.

cirq/ops/eigen_gate.py
```diff
@@ -134,8 +134,8 @@ def matrix(self) -> np.ndarray:
         if self.is_parameterized():
             raise ValueError("Parameterized. Don't have a known matrix.")
         e = cast(float, self._exponent)
-        return np.sum(1j**(half_turns * e * 2) * component
-                      for half_turns, component in self._eigen_components())
+        return np.sum([1j**(half_turns * e * 2) * component for half_turns,
+                       component in self._eigen_components()], axis=0)
```
Strilanc: Is the axis argument required?

dlyongemallo: Yes, as otherwise np.sum will sum over all axes, which results in a scalar.

PR #926, opened 2018-09-18
Polish auto_merge script

Strilanc:
- Don't attempt to delete PR branch after merging from a fork
- Print an explanatory message when failing to merge master into a fork
- Include PR title in console output
- Mark script as executable


Strilanc:
Good bot

dev_tools/auto_merge.py
```diff
@@ -529,6 +536,14 @@ def auto_delete_pr_branch(pr: PullRequestDetails) -> bool:
         return False
 
     remote = pr.remote_repo
+    if (pr.repo.organization.lower() != remote.organization.lower() or
+            pr.repo.name.lower() != remote.name.lower()):
+        print('Not deleting branch {!r}. It belongs to a fork ({}/{}).'.format(
```
dabacon: so if the repo name is different but the organization is the same it won't delete it?  so branches created on quantumlib aren't deleted?

Strilanc: Correct. If you somehow forked into a different repo within quantumlib and then merged in, this would not delete that branch. I think that's the correct behavior.

PR #939, opened 2018-09-19
[WIP] Use __pow__ instead of ReversibleGate/ExtrapolatableGate

Strilanc:
- Expand `cirq.inverse` to attempt `val**-1` before reversing/inverting items in a list.
- Add `cirq.extrapolate` which calls `val**factor` but with controllable failure behavior.
- Rewrite all gates to have a `__pow__` method instead of `inverse`/`extrapolate_effect`/`has_inverse`/etc

This is a "big glob of changes" tracking PR for https://github.com/quantumlib/Cirq/issues/943 . I will pull off more sensible pieces for actual PRs.


Strilanc:
Obsoleted by https://github.com/quantumlib/Cirq/pull/1021

Issue #943, opened 2018-09-19
Refactor ReversibleEffect and ExtrapolatableEffect classes into a magic method protocol


Strilanc:
Part of https://github.com/quantumlib/Cirq/issues/672

Instead of defining a new magic method, the presence of `__pow__` will indicate invertability or extrapolability.

PR #944, opened 2018-09-19
Refactor cirq.inverse to work on values that define __pow__ or implement ReversibleEffect

Strilanc:
- Moved cirq.inverse into cirq/protocols
- Include temporary compatibility shim for working with ReversibleEffect
- Will eventually delete ReversibleEffect in favor of this method

Part of https://github.com/quantumlib/Cirq/issues/943

### DIFF
--- a/cirq/protocols/inverse.py
+++ b/cirq/protocols/inverse.py
@@ -1,0 +1,78 @@
+# Copyright 2018 The Cirq Developers
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


cirq/protocols/inverse.py
```diff
@@ -0,0 +1,78 @@
...
+        val: The value (or iterable of invertable values) to invert.
+        default: Determines the fallback behavior when `val` doesn't have
+            an inverse defined. If `default` is not set, a TypeError is raised.
+            If `default` is set to a value, that value is returned.
+
+    Returns:
+        If `val` has a __pow__ method that returns something besides
+        NotImplemented when given an exponent of -1, that result is returned.
+        Otherwise, if `val` is iterable, the result is a tuple with the same
+        items as `val` but in reverse order and with each item inverted.
```
dabacon: we have cases where the iterable case is used?

Strilanc: `cirq.inverse` used to be something like `inverse_of_op_tree`, so actually all existing uses were on the iterable case.

Issue #952, opened 2018-09-19
Merge cirq.google xmon gates into the basic gate set


Strilanc:
Currently, the xmon gates are adding negative value. They do essentially the same thing as some of the basic gates, and people using basic gates get bit by the totally minor differences.

The merge must preserve the matrices of the gates. That is to say, it is necessary for the basic gates to support global phase shifting before this can happen.

Comes after https://github.com/quantumlib/Cirq/issues/950

kevinsung:
Just curious, why is it essential for the Xmon gates to have a certain global phase?

Strilanc:
People have preferences about whether a 180 degree rotation around the X axis should equal the Pauli X gate or exp(i pi X) = iX != X.

Issue #950, opened 2018-09-19
Add global phase shifting support to EigenGate and X/Y/Z gates


Strilanc:
So that we can have `rx(pi/2) = iX` be distinct from `X**1 = X`, but equivalent up to global phase.

Issue #865, opened 2018-08-17
Rename RotPGates to PowPGate and make RotPGates follow standard conventions


babbush:
I find the output of the following code extremely surprising. In every textbook the rotation around X gate is always defined as R_X(theta) = e^{-i theta X / 2}. But that's not true in Cirq. Not only does RotXGate differ from this by a global phase, but the function is 2pi periodic instead of 4pi periodic. The last part is extremely confusing for those of us that try to use these gates for quantum simulation (we are thinking about evolving for certain amounts of time, rather than the geometry of the Bloch sphere). I was trying to make a controlled gate that corresponded to controlled evolution under X and because of these strange global phase conventions my circuit wasn't working. This is such unexpected behavior that I suggest we consider it a bug.

```
import numpy
import scipy
import cirq

class MyRotXGate(cirq.KnownMatrix, cirq.Gate):
    def __init__(self, rads):
        self.rads = rads   
    def matrix(self):
        X = numpy.array([[0, 1], [1, 0]])
        unitary = scipy.linalg.expm(-1.j * self.rads * X / 2.)
        return unitary

for angle in numpy.linspace(0, 3 * numpy.pi, 10):
    A = MyGate(rads=angle)
    B = cirq.RotXGate(rads=angle)
    print('My gate')
    print(A.matrix())
    print('Cirq gate')
    print(B.matrix())
    print()
```

bryano:
I think this is actually reasonable behavior. In cirq,  `RotXGate(rads=θ)` corresponds to _X<sup>θ/π<sup>_. In some sense, the difference is between thinking of X as observable vs. as a unitary. Maybe renaming `RotXGate` (and its siblings) to something like `PauliXGate` would make this clearer. 

It's important to distinguish between the two different issues here: the global phase and the factor of 2. 

The global phase issue is not just for the Pauli rotation gates; see #816. Personally, the decision to really embrace that global phase is physically irrelevant seems reasonable, especially given the availability of the helper function `cirq.linalg.predicates.allclose_up_to_global_phase`.

The factor of two is more inconvenient. One reasonable solution would be to have a second gate `RX` that includes the factor of 2; that would make it usable by people familiar with the physicists' convention without having to change any existing code.

maffoo:
[...]

Issue #816, opened 2018-08-07
Global phase error


dgawerc:
I'm frequently getting final states where there is a phase multiplying the correct final state.

e.g. `X|+>=|+>` except Cirq returns `-i|+>`.

This can be seen by executing:

```
q = cirq.LineQubit(1)
state = np.array([1/np.sqrt(2), 1/np.sqrt(2)], dtype=np.complex64)
c = cirq.Circuit()
c.append(cirq.X(q))
simulator = cirq.google.XmonSimulator()
result = simulator.simulate(c, initial_state=state)
print(result.final_state)
```

Strilanc:
This is because decompositions do not preserve global phase, and the gate is being automatically decomposed into the xmon gate set which (currently) has a different phase convention than the ones in common.

On the one hand, it's annoying that optimizers don't keep track of this minor detail and make it transparently work. On the other hand, global phase is completely unobservable and so why *would* you preserve it? It's extra simulation work for no difference in measurement expectations. This is why we have methods like `allclose_up_to_global_phase`.

I think I've gradually been moving from the latter camp (it doesn't matter and we should ignore it) to the former camp (users are bothered by it so we should bother).

Really fixing this probably requires a way to represent a global phase operation (should it be an operation with no target qubits? An attribute on moment? A circuit attribute? An operation attribute?). Then we'd need to modify all the existing decomposition/optimization code to output a correct global phase modifier.

dgawerc:
Thanks for the prompt response. Although global phases don't matter for measurements, I was just making sure this was intended behavior.

Strilanc:
You're not the first person to complain, and I think it's important to fix these "not really a bug but it confuses new users so maybe that makes it a bug anyways" issues.

Strilanc:
I received an email from someone else who was confused about this:

> As expected, the square-Pauli-X operator has matrix like this:
>>> import cirq
>>> x_2 = cirq.X(0)**(1/2)
>>> x_2.matrix()
array([[0.5+0.5j, 0.5-0.5j],
       [0.5-0.5j, 0.5+0.5j]])
>
>But, if I run in a simulator a simple circuit with only this operator:
>    (0, 0): ───X^0.5───
>    result.final_state: [0.7071+0.j     0.    -0.7071j]
>The result should be [<0|A|0>,<1|A|0>] = [A[0,0], A[1,0]]  = [0.5+0.5j,0.5-0.5j ], where A is matrix of the operator. 
[...]

PR #953, opened 2018-09-19
Refactor EigenGate to include a global phase shift

Strilanc:
- Use it in RotZGate
- Refactor ExpZGate to inherit from RotZGate

Part of https://github.com/quantumlib/Cirq/issues/865
Part of https://github.com/quantumlib/Cirq/issues/816
Part of https://github.com/quantumlib/Cirq/issues/950


cirq/ops/common_gates.py
```diff
@@ -275,22 +275,30 @@ class RotZGate(eigen_gate.EigenGate,
...
+        turn is used.
 
         Args:
             half_turns: The relative phasing of Z's eigenstates, in half_turns.
             rads: The relative phasing of Z's eigenstates, in radians.
             degs: The relative phasing of Z's eigenstates, in degrees.
+            global_shift_in_half_turns: Offsets the eigenvalues of the gate.
+                The default shift of 0 gives the Z gate's matrix eigenvalues of
+                +1 and -1, whereas a shift of -0.5 changes those eigenvalues to
+                i and -i. The shift is always specified assuming an exponent of
```
kevinsung: It would be clearer to say `-i and i` instead of `i and -i` so that the order matches with `+1 and -1`.

cirq/ops/eigen_gate.py
```diff
@@ -186,8 +219,11 @@ def _unitary_(self) -> Union[np.ndarray, type(NotImplemented)]:
         if self.is_parameterized():
             return NotImplemented
         e = cast(float, self._exponent)
-        return np.sum([1j**(half_turns * e * 2) * component for half_turns,
-                       component in self._eigen_components()], axis=0)
+        return np.sum([
+            component * 1j**(
+                    2 * e * (half_turns + self._global_shift_in_half_turns))
+            for half_turns, component in self._eigen_components()
+        ], axis=0)
```
kevinsung: What do you think of
```
            component * (-1)**(
                    e * (half_turns + self._global_shift_in_half_turns))
```
instead?

Strilanc: Acknowledged, but it doesn't give an exact result in the same cases.

`(-1+0j)**(1*0.5) == (6.123233995736766e-17+1j)`

`1j**(2*0.5) == 1j`

This has an observable impact, namely the X gate's matrix being perturbed away from [[0, 1], [1, 0]] which I find distasteful.

cirq/ops/common_gates.py
Strilanc: Done.

Issue #954, opened 2018-09-20
Create docker container


dabacon:
for cirq installation ease.

vtomole:
Please close.

vtomole:
Ping.

PR #957, opened 2018-09-21
Create Dockerfile

vtomole:
Looking for feedback before I edit https://github.com/quantumlib/Cirq/blob/master/docs/install.md.

For one, do we want users to `docker build`? Or should we regulary update an image on Docker Hub (https://hub.docker.com/) so users can `docker pull` instead?

Fixes https://github.com/quantumlib/Cirq/issues/954


vtomole:
@dabacon Made the request so he's suited for this :)

Ashalynd:
`docker build` can be useful for advanced users. I don't know of any caveats regarding it. I haven't used docker for quite a while though.

vtomole:
@Ashalynd  Docker is great for web-apps because you have full control of the execution environment. This will be useful for deploying remote simulators that run on more powerful servers than local machines e.t.c.

vtomole:
I propose that we create a Cirq account on Docker Hub so users don't have to build the docker image.

Currently:
```
git clone git@github.com:quantumlib/Cirq.git
cd Cirq
docker build -t cirq/cirq . # Takes several minutes
docker run -it --rm cirq/cirq python -c 'import cirq; print(cirq.google.Foxtail)'
```
With Docker Hub:
```
docker pull cirq/cirq
docker run -it --rm cirq/cirq python -c 'import cirq; print(cirq.google.Foxtail)'
```

vtomole:
We could use any appropriately named Docker Hub  account . Maybe `quantumlib`.

Strilanc:
@zoltanegyed has been looking into more of the details of what has to be done to ensure this works well. This PR is likely going to be obsoleted by that work.

PR #960, opened 2018-09-21
Add cirq.testing.assert_same_circuits method

Strilanc:


cirq/testing/circuit_compare.py
```diff
@@ -165,6 +165,48 @@ def assert_circuits_with_terminal_measurements_are_equivalent(
...
+                         expected)
+
+
+def _first_differing_moment_index(circuit1: circuits.Circuit,
+                                  circuit2: circuits.Circuit) -> Optional[int]:
+    differences = [
+        i
+        for i, (m1, m2) in enumerate(itertools.zip_longest(circuit1, circuit2))
+        if m1 != m2
+    ]
```
dabacon: Why not simply bail early for loop?
```
for i, (m1, m2) in enumerate(itertools.zip_longest(circuit1, circuit 2):
  if m1 != m2:
     return i
return None
```

Strilanc: Good point. Done.

Strilanc:
@dabacon PTAL

cirq/testing/circuit_compare.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Tuple, Sequence, TYPE_CHECKING
+from typing import Tuple, Sequence, TYPE_CHECKING, Optional
```
dabacon: nit alphabetical?

Strilanc: Acknowledged. I think this is a bit of a lost cause until we add yapf tooling to fix it automatically. It's not consistent anywhere.

PR #968, opened 2018-09-24
Rename cirq.CliffordGate to cirq.SingleQubitCliffordGate (#964)

vtomole:
Curious. Where is `TwoQubitCliffordGate`?


Strilanc:
The closest thing to `TwoQubitCliffordGate` is `PauliInteractionGate`, but that operation only covers two qubit operations that of the form of a CZ conjugated by a single-qubit Clifford basis change (e.g. it can do a CNOT but not a SWAP).

Issue #946, opened 2018-09-19
Refactor TextDiagrammable class into a magic method protocol


Strilanc:
Part of https://github.com/quantumlib/Cirq/issues/672

This task includes:

1. Creating a magic method (akin to `def _unitary_`) that indicates diagrammability.
2. Creating a utility method (akin to `cirq.unitary`) that gets the diagramming info, if present.
3. Refactoring existing code that *uses* text diagram info to instead use the utility method defined in (2). Drop any directly related extensions functionality while doing so.
4. Refactoring existing classes that *provide* text diagram info to instead use the magic method defined in (1). Drop any directly related extensions functionality while doing so.

PR #976, opened 2018-10-02
Refactor cirq.Circuit to use cirq.circuit_diagram_info

Strilanc:
- Change fallback representation from (gate:0, gate:1, ...) to (gate, #1, #2, ...)
- Fix qubit-symbol mismatch error saying the wrong number.
- Drop ext argument.
- Had to update contrib/qcircuit_diagram because it uses the ext argument.
- Fixed qcircuit_diagram not aligning its output to column boundaries. Looks much better now.
- Updaded GateOperation and ControlledGate to delegate to cirq.circuit_diagram_info
- Start forcing keywords for cirq.to_text_diagram

Part of https://github.com/quantumlib/Cirq/issues/946


Strilanc:
Interesting. For some reason it's trying to submit despite the checks not failing, noticing that the submit failed, trying again, then noticing the repetition and aborting (good bot!).

Strilanc:
Oh that is strange, the github API is actually saying this PR has status checks that have succeeded.

Strilanc:
It had to do with a status check being absent. All fixed now, and will submit the updates to cirqbot in another PR.

PR #977, opened 2018-10-02
Delete supremacy "example" now that better code is in experiments/

sboixo:


Strilanc:
I meant to do it as part of the other PR, but this will do fine. I'll wait until that one is merged.

PR #979, opened 2018-10-03
Implement more _apply_unitary_to_tensor_ methods

Strilanc:
- ControlledGate, CCZGate, CCXGate, and CSwapGate
- Add `cirq.testing.assert_apply_unitary_to_tensor_is_consistent_with_unitary` and use it


cirq/ops/three_qubit_gates.py
```diff
@@ -246,6 +272,18 @@ def _decompose_inside_control(self,
         yield common_gates.S(c)**-1
         yield common_gates.X(a)**-0.5
 
+    def _apply_unitary_to_tensor_(self,
+                                  target_tensor: np.ndarray,
+                                  available_buffer: np.ndarray,
+                                  axes: Sequence[int],
+                                  ) -> np.ndarray:
+        return protocols.apply_unitary_to_tensor(
+            controlled_gate.ControlledGate(common_gates.SWAP),
```
kevinsung: Did you really mean to implement this before implementing it for SWAP?

Strilanc:
I added the SWAP and other two qubit gates. I did some testing to get a sense of the speedup. Here's the summary:

```
CCZ: 30x faster
CSWAP: 8x faster
CCX: 7x faster when half_turns=1, 4x faster otherwise
SWAP: 3.5x faster when half_turns=1, would be slower for other half_turns so not changed
CNOT: 3.5x faster when half_turns=1, would be slower for other half_turns so not changed
iSWAP: 2x faster when half_turns=1, would be slower for other half_turns so not changed
```

Issue #164, opened 2018-02-28
Optimization barrier?


Strilanc:

andyliphys:
Is there any update with the optimization barrier?

dabacon:
@andyliphys do you have a good use case, we haven't encountered many and wanted to understand before implementing.

andyliphys:
The initial motivation is that when I compile a circuit with measurement gates, it raises an error
```
ValueError: Adjacent Exp11 operations: W(0.325)^0.5287110447050094((1, 1)) and W(0.552)^-0.5270693014384051((1, 2)) and CZ((0, 0), (0, 1)) and XmonMeasurementGate('2')((0, 2)).
```
and I think that an optimization barrier in front of the measurement gates will solve the problem.

Nonetheless, I solved the problem by optimizing the circuit without measurement gates first and then appending the measurement gates to it.
And I agree with you that the optimization barrier is not an urgent feature.

andyliphys:
@dabacon `cirq.google.optimized_for_xmon(circuit, new_device=cirq.google.Foxtail)`

Strilanc:
@andyliphys could you provide code that causes that issue? optimized_for_xmon is not supposed to fail like that in the first place. It may have been fixed by https://github.com/quantumlib/Cirq/issues/795

andyliphys:
```
import cirq
qubit = [cirq.GridQubit(0, 0), cirq.GridQubit(0, 1), cirq.GridQubit(0, 2)]

circuit = cirq.Circuit()
circuit.append([cirq.CZ(qubit[0], qubit[1]), cirq.CZ(qubit[1], qubit[2])])

measure_circuit = cirq.Circuit()
measure_circuit.append([cirq.MeasurementGate(key='0').on(qubit[0]),
                        cirq.MeasurementGate(key='1').on(qubit[1]),
                        cirq.MeasurementGate(key='2').on(qubit[2])])

cirq.google.optimized_for_xmon(circuit, new_device=cirq.google.Foxtail)
cirq.google.optimized_for_xmon(circuit + measure_circuit, new_device=cirq.google.Foxtail)
```

`cirq.google.optimized_for_xmon(circuit, new_device=cirq.google.Foxtail)` works perfectly.
[...]

PR #981, opened 2018-10-03
Fix XmonDevice disallowing measurements beside CZs

Strilanc:
Fixes an issues mentioned here: https://github.com/quantumlib/Cirq/issues/164#issuecomment-426663879


Strilanc:
Hmmmmmmm. Apparently this "status check" is only in the 'checks' API not the statuses API.

PR #983, opened 2018-10-04
Implement apply_unitary_to_tensor for SWAP and ISWAP

### LINKED ISSUE #978: Implement _apply_unitary_effect_to_tensor_ for SWAP and ISWAP

kevinsung:
Resolves #978

### DIFF
--- a/cirq/ops/common_gates.py
+++ b/cirq/ops/common_gates.py
@@ -773,6 +773,23 @@ class SwapGate(eigen_gate.EigenGate,
                        exponent: Union[value.Symbol, float]) -> 'SwapGate':
         return SwapGate(half_turns=exponent)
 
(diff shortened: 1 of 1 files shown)


Strilanc:
This ended up being redundant with https://github.com/quantumlib/Cirq/pull/979 , which has ever-so-slightly-faster implementations.

For the SWAP gate I managed to do it three assignments, instead of four, via the `tmp=a, a=b, b=tmp` approach. Same for iSwap, but additionally I avoided two copies because I used `*= 1j` instead of `x * 1j`.

Overall this means the SWAP in that PR is 5% faster. Interestingly, the iSWAPs are almost identical. It turns out that `*= 1j` is 25% slower than `* 1j` for some reason. The space usage should be different though...

PR #984, opened 2018-10-04
Define `apply_matrix_to_slices` method

Strilanc:
- This method could be useful when writing _apply_unitary_to_tensor_ methods


cirq/linalg/transformations.py
```diff
@@ -147,3 +147,83 @@ def targeted_left_multiply(left_matrix: np.ndarray,
...
+    if out is None:
+        out = np.copy(target)
+    else:
+        out[...] = target[...]
+    if workspace is None and slices:
+        workspace = np.copy(target[slices[0]])
+
+    # Apply operation.
+    for i, s_i in enumerate(slices):
+        s_i = slices[i]
```
kevinsung: Unnecessary

PR #985, opened 2018-10-04
Fix auto_merge failing to understand checks that aren't status checks

Strilanc:
- Detect if merge is blocked due to absent required status checks
- Add a fallback if the merge is blocked for some unknown reason
- Query the 'checks' API for check runs (which are not listed by the statuses API)


Strilanc:
Well, apparently it still needs a bit of work. But it's definitely better than before.

PR #986, opened 2018-10-05
Reduce number of files at top level

### LINKED ISSUE #748: Reduce number of files at top level
Super big nit: would be nice to have the Cirq logo show up without having to scroll down to see it :)

vtomole:
Fixes #748


vtomole:
I'm assuming this is not building because `.travis.yml` is not in it's configured location. I think changing this location to `continous-integration` is appropriate.

vtomole:
While we are at it, how about moving `check/` , `continuous-integration/` and `third_party/sphinx`   to `dev_tools/`?

vtomole:
I can't find a way to build `.travis.yml` without it being in root. Seems to be impossible based on https://docs.travis-ci.com/user/customizing-the-build/  `Travis CI uses .travis.yml file in the root of your repository to learn about your project and how you want your builds to be executed`. I will move it back because of this.

vtomole:
Python 2 is not happy with the procedure that has `->`at the end of `style.md`. Is there a way to avoid this without telling snippets to avoid running that file?

check/pytest2
```diff
@@ -35,7 +35,7 @@ trap "exit" INT
 # Convert code from python3 to python2.7.
 echo "Transpiling to python 2 with destination '${out_dir}'..."
 set -e
-./python2.7-generate.sh "${out_dir}"
```
Strilanc: I think moving `python2.7-generate.sh` into `dev_tools` is a good call.

check/pytest2
```diff
@@ -35,7 +35,7 @@ trap "exit" INT
...
+./dev_tools/python2.7-generate.sh "${out_dir}"
 set +e
```
Strilanc: `.readthedocs.yml` doesn't belong in `continuous-integration`. It's not really related to that concept.

check/pytest2
```diff
@@ -35,7 +35,7 @@ trap "exit" INT
...
 
 # Run tests against converted code.
```
Strilanc: `apt-system-requirements.txt` is a runtime requirements file and so should stay at the root.

Strilanc: I think `MANIFEST.in` has to be at the root for packaging to work correctly.

README.rst
```diff
@@ -66,7 +66,7 @@ Contributing
 ------------
 
 We welcome contributions. Please follow these
-`guidelines <https://github.com/quantumlib/cirq/blob/master/CONTRIBUTING.md>`__.
+`guidelines <https://github.com/quantumlib/cirq/blob/master/dev_tools/CONTRIBUTING.md>`__.
```
Strilanc: I'm not sure if there are legal reasons for CONTRIBUTING to have to be at the top level or not. I feel uncomfortable moving this file.

Strilanc:
I appreciate these changes, but I think each of these files needs to considered carefully before being moved. There are sometimes tools that check for the existence of these files, so we can't move them without thinking through those cases. I am going to tentatively close this, because it's just too big of a change in terms of the amount and diversity of non-automated verification work.

vtomole:
Fair. The number of files at top level is not that many. We should look out when adding more in the future. @dabacon did preface https://github.com/quantumlib/Cirq/issues/748 by saying it's not really that big of a deal, so I suggest we close that too. With his permission of course.

Issue #393, opened 2018-05-10
Replace cirq.Symbol with sympy.Symbol / sympy formulas


kevinsung:
The docstring says
>A constant plus the runtime value of a parameter with a given key.

Is using the constant offset implemented yet? Shouldn't that be an attribute?

Also, I will want to use multiplication instead of addition. Can we also support a constant _times_ the runtime value? I'm happy to help with this.

kevinsung:
I describe a use case here: https://github.com/quantumlib/OpenFermion-Cirq/issues/282

Strilanc:
We've decided to go ahead and do this. We're looking for a volunteer to get this done, e.g.a  20%-er, because it's a reasonably independent reasonably large project. It involves going over the whole code base and doing little refactorings, and merging sympy's featureset with our needs. So there's a bunch of coding work but also a little bit of design work.

vtomole:
I'm not familiar with Sympy. Could you please expand on what advantages and disadvantages "Allow Symbol to be a Sympy expression"  will have? Thanks!

Strilanc:
The main advantage is that sympy supports algebra on symbols (i.e. you can make formulas), and it does a much better job of it than we ever would. Instead of being limited to just `X^t` as a parameterized gate, you could make a `X^(t^2 + 3 arcos(theta) - 1)` gate.

Now, in practice we won't be able to communicate those gates with arbitrary formulas to the service. Don't worry about that. It just has to work with the local simulator.

dabacon:
Found a pretty bad place where this bites us pretty badly.  The way we define the Rx gate (which most people will use) we only allow the radians to be passed in and then we rescale it.  So you cannot use symbols with the Rx gate.

Strilanc:
@dabacon We were *already* being bitten by that. Having the method just makes it more blatant. (This is why @babbush has been bringing it up; they prefer radians and so have been running into it.)

The change @vtomole is making will address this as a side effect, though it might be a few iterations before it starts to work.

vtomole:
@Strilanc Don't close yet as that pull request didn't address not the whole repo. (paulistring).

Strilanc:
It auto-closed because somewhere in the PR it said "Fixes bla".

PR #990, opened 2018-10-09
Replace cirq.Symbol with sympy.Symbol

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/393

The first commit of this pull request is me getting familiar with sympy. Looking for some feedback @kevinsung. Thank you!


cirq/ops/common_gates.py
```diff
@@ -187,7 +188,10 @@ def known_qasm_output(self,
     def __str__(self) -> str:
         if self.half_turns == 1:
             return 'X'
-        return 'X**{!r}'.format(self.half_turns)
+        elif isinstance(self.half_turns, float):
+            return 'X**{!r}'.format(self.half_turns)
+        return 'X**symbol.Symbol({!r})'.format(
```
Strilanc: Does the `repr` of a sympy expression not return evaluable code? I would expect `X**{!r}` to be sufficient.

cirq/ops/eigen_gate.py
```diff
@@ -79,21 +79,20 @@ def __init__(self, *,  # Forces keyword args.
...
                 angles by (before multiplying by the exponent).
         """
-
         self._exponent = exponent
         self._global_shift_in_half_turns = global_shift_in_half_turns
 
         # Canonicalize the exponent.
         period = self._canonical_exponent_period()
-        if period is not None and not isinstance(exponent, value.Symbol):
+        if period is not None and not isinstance(exponent,(sympy.mul.Mul,
```
Strilanc: As part of this change we would actually completely delete `cirq.Symbol`.

cirq/ops/common_gates.py
vtomole: Could you please elaborate on your question?

cirq/ops/eigen_gate.py
vtomole: Yup. This is me being wary of not rocking the boat too much :D

cirq/ops/common_gates.py
Strilanc: If `repr(sympy_thing)` returns code that evaluates to something equal to `sympy_thing`'s value, then `"cirq.X**{!r}`.format(sympy_thing)` should be code that evaluates to something equal to the gate. There shouldn't be any `symbol.Symbol` because, well, that class is not going to exist anymore so that's not evaluable code.

Strilanc:
Hmmm, I can see that this change is getting a bit tricky. For example, avoiding `a * b` producing `1.0 * s` instead of `s` when a=1.0 b=sympy.Symbol('s').

It may be simpler to start by keeping `cirq.Symbol` and having its value be a `sympy` expression. Alternatively, `cirq.mul` could be adapted to notice sympy expressions and call simplify on them and then we could be careful in internal gates to use that instead of `*`.

vtomole:
@Strilanc Agree with the second option as we want to get rid of `cirq.Symbol`. Might as well rip off the band-aid. 31 tests to pass after this following commit.

vtomole:
@Strilanc How should we handle PauliString equality? This is what i'm getting.

```
def test_extrapolate_effect_with_symbol():
        eq = cirq.testing.EqualsTester()
        eq.add_equality_group(
            PauliStringPhasor(cirq.PauliString({}),
                              half_turns=Symbol('a')),
            PauliStringPhasor(cirq.PauliString({})) ** Symbol('a'))
```
`AssertionError: PauliStringPhasor({+, }, half_turns=a) and PauliStringPhasor({+, }, half_turns=-Mod(-1.0*a + 1.0, 2) + 1) can't be in the same equality group. They're not equal.`

cirq/study/resolver.py
```diff
@@ -54,8 +55,8 @@ def value_of(
         """
         if isinstance(value, str):
             return self.param_dict.get(value, Symbol(value))
-        if isinstance(value, Symbol):
-            return self.param_dict.get(value.name, value)
+        if isinstance(value, sympy.mul.Mul):
```
Strilanc: This is almost certainly incorrect. There could be arbitrarily complicated `sympy` expressions. You should use substitution: https://docs.sympy.org/latest/tutorial/basic_operations.html#substitution

vtomole: Thanks for letting me know about substitutions. I was looking for a way to generalize this and just using `Symbol`  wasn't working.

vtomole:
@Strilanc Non-Cirq obligations have been taking over my life lately. I also hate dealing with merge conflicts so i will make time for this as much as possible. As far as i can tell, most of the problems left have to do with `ParamResolver`. 

I also seriously need to clean up (Type annotations for example); so please ignore those when you are reviewing for now. Thank you.

Strilanc:
Hahaha, yes, sorry about all the merge conflicts as we make breaking changes leading up to a release.

cirq/value/angle.py
```diff
@@ -88,7 +88,7 @@ def canonicalize_half_turns(
         half_turns: Union[Symbol, float]
 ) -> Union[Symbol, float]:
     """Wraps the input into the range (-1, +1]."""
-    if isinstance(half_turns, Symbol):
+    if isinstance(half_turns, sympy.Basic):
```
Strilanc: There might be lots of different `sympy` things, right? We should invert this logic so that it cares about "not int and not float" instead of "is a specific brand of weird".

Strilanc:
I've done a lot of these "gigantic code base spanning change" things, and basically this is my workflow.

1. Try to do it as one gigantic PR, with work chaining into work chaining into work. Get it to work; barely.
2. Create a fresh new branch off of master and squash merge (without commiting) the half-working gigantic code changes into it.
3. Use a tool such as `git gui` or github's files changed view to review all the changes.
4. Pick a small separable subset (e.g. making methods that take angles work with `sympy` things), keep that, revert everything else. (Shim the methods with backwards-compatibility code if needed.)
5. Do (4) a few times in parallel, with different subsets of changes or maybe chains of small changes.
6. Get the small separable subsets merged, sync master into the gigantic branch, and repeat. Keep going until the gigantic branch's diff is small enough to get merged on its own.

Currently you are in the throws of step (1).

Strilanc:
`1 - (1 - x) % 2` maps any `x` into the range `-1 < x <= 1` while preserving the value of `x % 2`. It's an expression to canonicalize the angle.

I would recommend only bothering to do this if `x` is an int or a float, since for symbols it will be more hassle than it's worth. It's already almost more hassle than it's worth for floats, and is e.g. being removed from EigenGate.

vtomole:
I am currently looking at ` test_blocked_by_unknown_and_symbols()` and I can't figure out why the optimizer is not working correctly. I haven't grokked the optimizer yet; but could changing how we represent our symbols mess it up?

Strilanc:
Any logic that says "do this if symbol do that if number" could ultimately result in the optimizer doing the wrong thing when you subtly change how to check "if symbol".

Strilanc:
If I were you, I would step back at this point, look over what I'd learned as part of the refactoring, and start trying to come up with a small change that can be submitted on its own such as e.g. making sympy symbol work in a few places but not removing support for cirq.Symbol yet.

vtomole:
@Strilanc Will submit small changes in different PRs. One problem i ran into while trying to make a small change while keeping `cirq.Symbol` was that sometimes there was a conflict in the program between the Symbols. Will try to figure out how to get around that.

There are currently 9 tests left after the merge with master. Most look minor except the optimizers: https://gist.github.com/vtomole/3f3849c2e0c891338aad6a80363e0315

Strilanc:
Sounds good to me.

vtomole:
@Strilanc I think this a good step to getting `sympy` into the mainline because it passes the checks.

Files that currently use `cirq.Symbol`:
```
ops/common_gates_test.py    
ops/phased_x_gate_test.py
ops/matrix_gates_test.py   
ops/eigen_gate_test.py     
optimizers/merge_interactions_test.py
optimizers/eject_phased_paulis_test.py
optimizers/eject_z_test.py            
contrib/paulistring/pauli_string_phasor_test.py
```

cirq/circuits/circuit.py
```diff
@@ -141,6 +144,8 @@ def __bool__(self):
     def __eq__(self, other):
         if not isinstance(other, type(self)):
             return NotImplemented
+        #print("Self moments ", self._moments)
```
vtomole: Will remove debug prints after review.

cirq/contrib/jobs/job.py
```diff
@@ -39,6 +39,8 @@ def __init__(self,
     def __eq__(self, other):
         if not isinstance(other, type(self)):
             return NotImplemented
+        #print(self.circuit)
```
vtomole: Remove.

cirq/ops/eigen_gate.py
```diff
@@ -262,8 +266,9 @@ def _period(self) -> Optional[float]:
         real_periods = [abs(2/e) for e in exponents if e != 0]
         return _approximate_common_period(real_periods)
 
-    def __pow__(self: TSelf, exponent: Union[float, value.Symbol]) -> TSelf:
+    def __pow__(self: TSelf, exponent: Union[float, sympy.Symbol]) -> TSelf:
         new_exponent = protocols.mul(self._exponent, exponent, NotImplemented)
+        #print("Type of exponent in pow ", new_exponent)
```
vtomole: Remove.

cirq/ops/moment.py
```diff
@@ -46,6 +46,7 @@ def __init__(self, operations: Iterable[raw_types.Operation] = ()) -> None:
             ValueError: A qubit appears more than once.
         """
         self.operations = tuple(operations)
+        print("Operantions ", operations)
```
vtomole: Remove.

cirq/protocols/pow.py
```diff
@@ -89,7 +89,10 @@ def pow(val: Any,
         TypeError: `val` doesn't have a __pow__ method (or that method returned
             NotImplemented) and no `default` value was specified.
     """
+    #print("Val ", val)
```
vtomole: Remove prints.

Strilanc:
Hey Victory, what's the current status of this? At the moment the changed files look reasonable but travis is very unhappy (looks like you need to add sympy to the requirements).

vtomole:
@Strilanc  The.problem came from fixing merge conflicts on the browser instead of my local machine as I didn't run the checks locally. It's all good now. As we talked about earlier `Come up with a small change that can be submitted on its own`. I want to get this reviewed and into master so that we can establish a pattern for the rest of the migration to `sympy`.

cirq/circuits/circuit.py
```diff
@@ -1528,10 +1531,23 @@ def _get_operation_circuit_diagram_info_with_fallback(
...
+
+def _encode(text):
+    return json.JSONEncoder().encode(text)
 
 def _formatted_exponent(info: protocols.CircuitDiagramInfo,
                         args: protocols.CircuitDiagramInfoArgs
                         ) -> Optional[str]:
+
+    if isinstance(info.exponent, sympy.Basic):
+        name = str(info.exponent.free_symbols.pop())
```
Strilanc: Why not just `name = str(info.exponent)`? E.g. this would give the right value of "a + b" when the exponent is `sympy.Symbol('a') + sympy.Symbol('b')`, instead of just "a".

Add a test that setting the exponent to `sympy.Symbol('a') + sympy.Symbol('b')` gives a reasonable looking circuit diagram.

cirq/circuits/circuit.py
```diff
@@ -1528,10 +1531,23 @@ def _get_operation_circuit_diagram_info_with_fallback(
...
+        return (name
+                if _is_valid_identifier(name)
```
Strilanc: Given that we now expect there to be expressions, we should rename this to e.g. `_is_exposed_formula`.

cirq/google/programs.py
```diff
@@ -351,22 +351,22 @@ def raise_missing_fields(gate_name: str):
...
                      'message: {!r}'.format(message))
 
 
-def _parameterized_value_to_proto_dict(param: Union[value.Symbol, float]
+def _parameterized_value_to_proto_dict(param: Union[sympy.Symbol, float]
                                        ) -> Dict:
     out = {}  # type: Dict
-    if isinstance(param, value.Symbol):
-        out['parameter_key'] = param.name
+    if isinstance(param, sympy.Basic):
```
Strilanc: This is the wrong type check at this point. We only want `sympy.Symbol`. If it's a `sympy.Basic` but not a `sympy.Symbol`, raise a `NotImplementedError` explaining we don't support serializing it yet.

cirq/ops/eigen_gate.py
```diff
@@ -59,7 +60,7 @@ class EigenGate(raw_types.Gate):
     """
 
     def __init__(self, *,  # Forces keyword args.
-                 exponent: Union[value.Symbol, float] = 1.0,
+                 exponent: Union[sympy.Symbol, float] = 1.0,
```
Strilanc: I suspect a lot of the `sympy.Symbol` types should change to `sympy.Basic`, because they can be formulae. For now it's fine to leave at as symbol and ease in.

cirq/ops/eigen_gate.py
```diff
@@ -98,17 +99,20 @@ def __init__(self, *,  # Forces keyword args.
                 `cirq.Rx(t)` uses a `global_shift` of -0.5, which is why
                 `cirq.unitary(cirq.Rx(pi))` equals -iX instead of X.
         """
-        self._exponent = exponent
+        if isinstance(exponent, sympy.Basic):
```
Strilanc: Unconditionally assign exponent. We're happy with whatever we were given. Definitely don't try to pick a randomvariable out of an equation as the exponent.

vtomole: Done.

vtomole:
Failing due to `ImportError: No module named 'pandas'` from `cirq/contrib/tpu/circuit_to_tensorflow.py`.

Strilanc:
That's... very strange that that would start failing for this PR.

vtomole:
@Strilanc  Ready for second round of review :)

vtomole:
https://github.com/quantumlib/Cirq/issues/1262 Run on this:
```
vtomole@vtomole:~$ python test.py 
x: ───H───
```

cirq/circuits/circuit.py
```diff
@@ -1527,10 +1530,23 @@ def _get_operation_circuit_diagram_info_with_fallback(
...
 
+def _is_valid_identifier(text):
+    return re.match('[a-zA-Z_][a-zA-Z0-9_]*$', text)
 
-def _formatted_exponent(info: protocols.CircuitDiagramInfo,
+
+def _encode(text):
+    return json.JSONEncoder().encode(text)
+
+def _is_exposed_formula(info: protocols.CircuitDiagramInfo,
```
Strilanc: Rename back to `_formatted_exponent`. Methods with `_is` at the start of their name should return booleans.

cirq/circuits/circuit.py
```diff
@@ -1527,10 +1530,23 @@ def _get_operation_circuit_diagram_info_with_fallback(
 
     return protocols.CircuitDiagramInfo(wire_symbols=symbols)
 
+def _is_valid_identifier(text):
```
Strilanc: This is the method I wanted to be called `_is_exposed_formula`

cirq/circuits/circuit_test.py
```diff
@@ -297,6 +298,48 @@ def test_empty_moments():
...
+def test_symbol_addition_in_gate_exponent():
+    # 1-qubit test
+    qubit = cirq.NamedQubit('a')
+    circuit = cirq.Circuit.from_ops(
+        cirq.X(qubit)**0.5,
+        cirq.YPowGate(
+            exponent=sympy.Symbol('a') + sympy.Symbol('b')).on(qubit)
+    )
+    cirq.testing.assert_has_diagram(circuit,
+                                    "a: ───X^0.5───Y^Symbol(\"a + b\")───",
```
Strilanc: Use an r-string to avoid the backslashes. Or use single-quotes.

cirq/circuits/circuit_test.py
```diff
@@ -297,6 +298,48 @@ def test_empty_moments():
...
+                                    use_unicode_characters=True)
+
+
+    cirq.testing.assert_has_diagram(circuit,
+"""
```
Strilanc: backslashes shouldn't be needed

cirq/circuits/circuit_test.py
```diff
@@ -297,6 +298,48 @@ def test_empty_moments():
...
+X^0.5
+│
+Y^Symbol(\"a + b\")
+│
+""",
+                                    use_unicode_characters=True,
+     transpose=True)
+
+    cirq.testing.assert_has_diagram(circuit,
+                                    "a: ---X^0.5---Y^Symbol(\"a + b\")---",
```
Strilanc: single-quotes or r-string to avoid backslashes

cirq/google/programs_test.py
```diff
@@ -259,9 +260,9 @@ def test_z_proto_dict_convert():
             }
         }
     }
-    assert_proto_dict_convert(gate, proto_dict,
-                              cirq.GridQubit(2, 3))
-
+    with pytest.raises(NotImplementedError):
```
Strilanc: This should still be implemented. Do not convert this check into an expected failure, instead of an assertion of success.

What shouldn't be implemented is formulas such as `sympy.Symbol('k') + sympy.Symbol('k2')`.

cirq/google/programs_test.py
```diff
@@ -379,10 +381,12 @@ def test_w_to_proto_dict():
             }
         }
     }
-    assert_proto_dict_convert(gate, proto_dict,
-                              cirq.GridQubit(2, 3))
+    with pytest.raises(NotImplementedError):
```
Strilanc: This should still succeed. Do not expect a failure here.

cirq/google/programs_test.py
```diff
@@ -397,8 +401,9 @@ def test_w_to_proto_dict():
...
```
Strilanc: This should still succeed. Do not expect a failure here.

cirq/google/programs_test.py
```diff
@@ -449,8 +455,9 @@ def test_w_to_proto_dict():
...
```
Strilanc: This should still succeed. Do not expect a failure here.

cirq/google/programs_test.py
```diff
@@ -295,8 +296,9 @@ def test_cz_proto_dict_convert():
             }
         }
     }
-    assert_proto_dict_convert(gate, proto_dict,
-                              cirq.GridQubit(2, 3), cirq.GridQubit(3, 4))
+    with pytest.raises(NotImplementedError):
```
Strilanc: This should still succeed. Do not expect a failure here.

cirq/ops/common_gates_test.py
```diff
@@ -294,15 +295,17 @@ def test_text_diagrams():
         cirq.IdentityGate(2)(a,b))
 
     cirq.testing.assert_has_diagram(circuit, """
-a: ───×───X───Y───Z───Z^x───@───@───X───H───iSwap───iSwap──────I───I───
-      │                     │   │   │       │       │              │
-b: ───×─────────────────────@───X───@───────iSwap───iSwap^-1───────I───
+a: ───×───X───Y───Z───Z^Symbol("1.0*x")───@───@───X───H───iSwap───iSwap──────I───I───
```
Strilanc: We shouldn't have a `1.0*` in the exponent here. This may be way the other test cases were swiched to errors? We likely need to add a case to various gate pow methods to avoid multiplying symbols by 1.

cirq/circuits/circuit.py
vtomole: Done.

cirq/circuits/circuit.py
vtomole: Done,

cirq/circuits/circuit_test.py
vtomole: Done.

cirq/circuits/circuit_test.py
vtomole: Done.

cirq/google/programs_test.py
vtomole: Done. And all the other ones too.

vtomole:
@Strilanc Ready for third round :)

cirq/circuits/circuit.py
```diff
@@ -1527,10 +1530,23 @@ def _get_operation_circuit_diagram_info_with_fallback(
...
 
 def _formatted_exponent(info: protocols.CircuitDiagramInfo,
                         args: protocols.CircuitDiagramInfoArgs
                         ) -> Optional[str]:
+
+    if isinstance(info.exponent, sympy.Basic):
+        name = str(info.exponent)
+        return (name
+                if _is_exposed_formula(name)
+                else 'Symbol({})'.format(_encode(name)))
```
Strilanc: Good enough for now. May change to just parens `'({})'` later, since it may be a formula instead of a symbol.

vtomole: Done.

vtomole:
@Strilanc This is good. We should encourage everyone to use sympy from now on unless there is a good reason to fall back to `cirq.Symbol`. Thanks for being patient!!

Strilanc:
@vtomole No, thank you for putting all the effort over a span of months into making this work!

dabacon:
This is really awesome, thanks a ton @vtomole !!!

vtomole:
@dabacon  My pleasure!

PR #991, opened 2018-10-09
Define simulator interface and move xmon simulator over to using it.

dabacon:
This has one breaking change which is that the run method previous took qubit_order, but this was not needed, since run does not return wave function.


cirq/__init__.py
```diff
@@ -168,13 +168,24 @@
     moment_by_moment_schedule,
 )
 
+from cirq.sim import (
+    check_state,
+    decode_initial_state,
+    pretty_state,
+    RunSimulator,
```
Strilanc: I would call this `SampleSimulator`.

cirq/__init__.py
```diff
@@ -168,13 +168,24 @@
     moment_by_moment_schedule,
 )
 
+from cirq.sim import (
+    check_state,
+    decode_initial_state,
```
Strilanc: This method name is ambiguous out of context. Maybe `initial_state_to_wavefunction` or `state_to_wavefunction` or just `wavefunction(state)` or `state_vector(index=x, vector=y, dtype=z)`?

cirq/__init__.py
```diff
@@ -168,13 +168,24 @@
...
+    pretty_state,
```
Strilanc: `cirq.pretty_state` is ambiguous. Perhaps `cirq.describe_state` or `cirq.summarize_state_vector`?

cirq/google/sim/xmon_simulator.py
```diff
@@ -198,111 +117,40 @@ def __init__(self, options: XmonOptions = None) -> None:
...
-                repetitions=repetitions,
-                measurements={k: np.array(v) for k, v in measurements.items()}
-            ))
-        return trial_results
-
-    def _run_sweep_repeat(self, keys, circuit, repetitions, qubit_order):
-        measurements = {
-            k: [] for k in keys}  # type: Dict[str, List[np.ndarray]]
+    def _run_sweep_repeat(self, keys, circuit, repetitions):
+        measurements = {k: [] for k in
```
Strilanc: We should open an issue to extract most of this functionality into a method such as `sample_state_vector(vector=v, num_samples=k)`.

cirq/google/sim/xmon_simulator.py
```diff
@@ -312,257 +160,101 @@ def _run_sweep_sample(self, circuit, repetitions, qubit_order):
...
-        """
-        param_resolver = param_resolver or ParamResolver({})
-        xmon_circuit, _ = self._to_xmon_circuit(program,
+        circuit: circuits.Circuit,
+        param_resolver: study.ParamResolver,
+        qubit_order: ops.QubitOrderOrList,
+        initial_state: Union[int, np.ndarray],
+        extensions: extension.Extensions,
+        perform_measurements: bool = True,
+    ):
```
Strilanc: Add output annotation type.

cirq/google/sim/xmon_simulator.py
```diff
@@ -312,257 +160,101 @@ def _run_sweep_sample(self, circuit, repetitions, qubit_order):
...
+                                   perform_measurements)
+
+
+    def _base_iterator(
+        self,
+        circuit: circuits.Circuit,
+        qubit_order: ops.QubitOrderOrList,
+        initial_state: Union[int, np.ndarray],
+        perform_measurements: bool=True,
+    ) -> Iterator['XmonStepResult']:
```
Strilanc: Add a quick docstring, perhaps saying "See _simulator_iterator".

cirq/google/sim/xmon_simulator.py
```diff
@@ -312,257 +160,101 @@ def _run_sweep_sample(self, circuit, repetitions, qubit_order):
...
+        qubits = ops.QubitOrder.as_qubit_order(qubit_order).order_for(
+            circuit.all_qubits())
+        qubit_map = {q: i for i, q in enumerate(reversed(qubits))}
+        if isinstance(initial_state, np.ndarray):
+            initial_state = initial_state.astype(dtype=np.complex64,
```
Strilanc: We can't specify the dtype anymore?

cirq/__init__.py
```diff
@@ -168,13 +168,24 @@
     moment_by_moment_schedule,
 )
 
+from cirq.sim import (
+    check_state,
```
Strilanc: This method name is ambiguous out of context.

It almost feels better placed as `cirq.testing.assert_state_vector_is_normalized(vector=v, atol=1e-8)`.

cirq/sim/simulator.py
```diff
@@ -0,0 +1,436 @@
...
+            extensions: Extensions that will be applied during the simulate.
+                See documentation of class for details.
+
+        Yields:
+            StepResults from simulating a Moment of the Circuit.
+        """
+        raise NotImplementedError()
+
+
+class StepResult:
```
Strilanc: I think we should separate the concept of a simulator that gives you intermediate states from the concept of a simulator that gives you the final state. Some simulation strategies, e.g. path summing, cannot cheaply give the intermediate steps.

cirq/sim/simulator.py
```diff
@@ -0,0 +1,436 @@
...
+                numpy.kron.
+            decimals: How many decimals to include in the pretty print.
+
+        Returns:
+            A pretty string consisting of a sum of computational basis kets
+            and non-zero floats of the specified accuracy."""
+        return state.pretty_state(self.state, decimals)
+
+
+class SimulateTrialResult:
```
Strilanc: `SimulationTrialResult`

cirq/sim/simulator_test.py
```diff
@@ -0,0 +1,215 @@
...
+        cirq.SimulateTrialResult(
+            params=cirq.ParamResolver({'a': 2}),
+            measurements={'m': np.array([1, 2])},
+            final_state=np.array([0, 1, 0, 0])))
+    eq.add_equality_group(
+        cirq.SimulateTrialResult(
+            params=cirq.ParamResolver({'a': 2}),
+            measurements={'m': np.array([1, 2])},
+            final_state=np.array([0, 0, 1, 0])))
+    eq.add_equality_group(
```
Strilanc: Also vary the measurements.

cirq/sim/state.py
```diff
@@ -0,0 +1,102 @@
...
+            mapping to qubits follows the standard Kronecker convention of
+            numpy.kron.
+        decimals: How many decimals to include in the pretty print.
+
+    Returns:
+        A pretty string consisting of a sum of computational basis kets
+        and non-zero floats of the specified accuracy.
+    """
+    perm_list = ["".join(seq) for seq in itertools.product(
+        "01", repeat=int(len(state)).bit_length() - 1)]
```
Strilanc: This works on the empty ket? Pass in `state=[1]` (the output of `cirq.Circuit().apply_unitary_effect_to_state()`) and the output should be `|⟩`.

cirq/sim/state.py
```diff
@@ -0,0 +1,102 @@
...
+        if val != 0:
+            if round(state[x], decimals) == 1:
+                components.append(ket.format(perm_list[x]))
+            else:
+                components.append((format_str + ket).format(val, perm_list[x]))
+
+    return ' + '.join(components)
+
+
+def decode_initial_state(initial_state: Union[int, np.ndarray],
```
Strilanc: We should consider also supporting `Dict[int, complex]`.

cirq/sim/state.py
```diff
@@ -0,0 +1,102 @@
...
+    num_qubits: int, dtype: np.dtype = np.complex64) -> np.ndarray:
```
Strilanc: These arguments interplay in complicated ways. Should we infer `num_qubits` when `initial_state` is a vector?

cirq/__init__.py
dabacon: Done

cirq/sim/simulator.py
dabacon: OK, broke into two, one interface that gives full wave function at end only, and one that allows stepping.

cirq/google/sim/xmon_simulator.py
Strilanc: I just mean that it seems like "Oh I have a wavefunction, I want to sample bits from it" seems like a use case we'll run into multiple times and that our users are also likely to run into. So it would make sense to have an independent method that did so. We could just open an issue about it.

cirq/__init__.py
Strilanc: `dirac_notation` is better than my suggestions.

cirq/__init__.py
```diff
@@ -168,13 +168,25 @@
     moment_by_moment_schedule,
 )
 
+from cirq.sim import (
+    dirac_notation,
+    SampleSimulator,
```
Strilanc: A major UX problem with having this class at the top level is that people exploring the API will go "Oh, SampleSimulator, that must be the simulator I use to get samples. I'll remember that.". Then it doesn't give them samples, because it's an interface, and they feel tricked.

The same goes for the other `XSimulator` base classes.

I'm not sure what our naming convention should be to avoid this. Perhaps `SupportsSampleSimulator` or `AbstractSampleSimulator` or `SampleSimulatorInterface` or something along those lines.

dabacon: Maybe we can move the simulate word.

SimulatesSamples
SimulatesFinalWaveFunction
SimulatesIntermediateWaveFunction

I'm not as worried about the length of these names, since they actually are only used by those implementing the interface, so not really part of core use.

Thoughts?

PR #995, opened 2018-10-10
Implement Circuit.reachable_frontier_from and Circuit.findall_operations_between

Strilanc:
- Implemented a private priority queue (BuckerPriorityQueue) specialized to the needs of these methods

These methods are useful when attempting to find "spacelike-separated light cones" within a circuit. For example, if you want to grab as many operations as possible to fold into one big 8 qubit operation, you would do the following:

```python
start_frontier = {q: start_index for q in group_of_8_qubits}
end_frontier = circuit.reachable_frontier_from(start_frontier)
merged_operations = circuit.findall_operations_between(start_frontier, end_frontier)
```

### DIFF
--- a/cirq/circuits/circuit.py
+++ b/cirq/circuits/circuit.py
@@ -29,6 +29,7 @@ from typing import (
 import numpy as np
 
 from cirq import devices, ops, extension, study, protocols
+from cirq.circuits._bucket_priority_queue import BucketPriorityQueue
 from cirq.circuits.insert_strategy import InsertStrategy
 from cirq.circuits.moment import Moment
 from cirq.circuits.text_diagram_drawer import TextDiagramDrawer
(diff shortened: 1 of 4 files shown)


cirq/circuits/circuit.py
```diff
@@ -407,6 +408,208 @@ def _prev_moment_available(
...
+                is reachable and L is within the bounds of the circuit.
+
+            OR
+
+            c) There is an operation P covering L and, for every location
+                M = (q', moment_index) that P covers, the location
+                prev(M) = (q', moment_index-1) is reachable. Also, P must not be
+                classified as a blocker by the given `is_blocker` argument.
+
+        In other words, the reachable region extends horizontally along each
```
dabacon: horizontally is a bit awkward since sometimes we draw the diagrams vertically.  Think also you want to say forward in time.

cirq/circuits/circuit.py
```diff
@@ -407,6 +408,208 @@ def _prev_moment_available(
...
+                    cirq.LineQubit(2): 0,
+                    cirq.LineQubit(3): 0,
+                }
+
+            Then this is the reachable area:
+
+                0   1   2   3   4   5   6   7   8   9   10  11  12  13
+            0: ─██H███@██████████████████████████████████████─@───H───
+                      │                                       │
+            1: ─██████@███H██─@───────────────────────@───H───@───────
```
dabacon: according to the definition I don't understand why this goes through the first CZ

cirq/circuits/circuit.py
```diff
@@ -407,6 +408,208 @@ def _prev_moment_available(
...
+            A list of tuples. Each tuple describes an operation found between
+            the two frontiers. The first item of each tuple is the index of the
+            moment containing the operation, and the second item is the
+            operation itself. The list is sorted so that the moment index
+            increases monotonically.
+        """
+        result = BucketPriorityQueue[ops.Operation](drop_duplicates=True)
+
+        involved_qubits = set(start_frontier.keys()) | set(end_frontier.keys())
+        for q in ops.QubitOrder.DEFAULT.order_for(involved_qubits):
```
dabacon: should this method take in a qubit order that can be used here?

Strilanc: No, the sorting is just to ensure a deterministic output. The ordering doesn't matter, as long as it's consistent for testing. Added a comment saying such.

Also, I switched from `ops.QubitOrder.DEFAULT.order_for` to just `sorted`, now that that finally works. :smile:

PR #997, opened 2018-10-10
Add cirq.testing.assert_equivalent_repr

Strilanc:
- Checks that evaluating a value's repr produces an equivalent value
- Applied it to `cirq.Circuit`. Immediately caught the fact that Xmon gates are not including a `cirq.google` prefix.

We should be using this on essentially every class. I've opened followup issue #998 to actually use this method everywhere.


cirq/testing/equivalent_repr_eval.py
```diff
@@ -0,0 +1,61 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any
+
+
+def assert_equivalent_repr(
```
dabacon: assert_equivalent_reprs since this takes in multiple values?

PR #1000, opened 2018-10-11
Add alpha stage statements to README and install.md

vtomole:
Fixes https://github.com/quantumlib/Cirq/issues/750


README.rst
```diff
@@ -82,7 +82,8 @@ chemistry and materials science, we encourage exploring
 its sister library for compiling quantum simulation algorithms in Cirq,
 `OpenFermion-Cirq <https://github.com/quantumlib/openfermion-cirq>`__.
 
-Disclaimer
+Disclaimers
 ----------
-
-Copyright 2018 The Cirq Developers. This is not an official Google product.
+* Cirq is currently in Alpha
```
Strilanc: This is not enough information. We need to emphasize that we are going to be making breaking changes every single release, and e.g. that users should target specific versions if they don't want to experience unexpected breakage.

README.rst
```diff
@@ -82,7 +82,8 @@ chemistry and materials science, we encourage exploring
 its sister library for compiling quantum simulation algorithms in Cirq,
 `OpenFermion-Cirq <https://github.com/quantumlib/openfermion-cirq>`__.
 
-Disclaimer
+Disclaimers
 ----------
-
-Copyright 2018 The Cirq Developers. This is not an official Google product.
+* Cirq is currently in Alpha. This means that the developer team is not investing effort to provide backward compatibility between releases. Users should target specific versions to avoid backward incompatible changes.
```
Strilanc: It's not so much that we're not putting in effort as that we are purposefully making breaking changes to avoid laying a bad foundation.

```
**Cirq is currently in alpha.** We are still making breaking changes, and we *will* break your code when we make new releases. We recommend that you pin a specific version of cirq in `requirements.txt` files, and periodically bump to the latest release. That way you have control over when a breaking change affects you.
```

vtomole:
Done.

vtomole:
@Strilanc My pleasure; as always!

PR #1005, opened 2018-10-12
Refactor BoundedEffect into a magic method

dstrain115:
- Create _trace_distance_bound_ magic method
    - Create cirq.trace_distance_bound to utilize it
    - Refactor all instances to use these two methods.


cirq/protocols/trace_distance_bound.py
```diff
@@ -0,0 +1,60 @@
...
+# limitations under the License.
+
+from typing import Any, TypeVar
+from typing_extensions import Protocol
+
+
+TDefault = TypeVar('TDefault')
+
+
+class TraceDistanceBound(Protocol):
```
Strilanc: `SupportsTraceDistanceBound`

cirq/protocols/trace_distance_bound.py
```diff
@@ -0,0 +1,60 @@
...
+        """A maximum on the trace distance between this effect's input/output.
+
+        Generally this method is used when deciding whether to keep gates, so
+        only the behavior near 0 is important. Approximations that overestimate
+        the maximum trace distance are permitted. Even ones that exceed 1.
+        Underestimates are not permitted.
+        """
+
+def trace_distance_bound(val: Any) -> float:
+    """Returns a maximum on the trace distance between this effect's input
```
Strilanc:
```
between `val`'s input
```

cirq/protocols/trace_distance_bound.py
```diff
@@ -0,0 +1,60 @@
...
+    and output.  This method makes use of the effect's `_trace_distance_bound_`
```
Strilanc:
```
use of `val`'s
```

cirq/protocols/trace_distance_bound.py
```diff
@@ -0,0 +1,60 @@
...
+    method to determine the maximum bound on the trace difference between
+    before and after the effect.
+
+    Args:
+        val: The effect of which the bound should be calculated
+
+    Returns:
+        If `val` has a _trace_distance_bound_ method and its result is not
+        NotImplemented, that result is returned. Otherwise, positive infinity
```
Strilanc: We should default to `1` instead of positive infinity (1 means 100% distinguishable). Also, we might as well clamp the value returned to be at most 1.

cirq/protocols/trace_distance_bound.py
```diff
@@ -0,0 +1,60 @@
...
+    operation is very close to 0, so it would typically be considered
+    negligible.
+    """
+
+    def trace_distance_bound(self) -> float:
+        """A maximum on the trace distance between this effect's input/output.
+
+        Generally this method is used when deciding whether to keep gates, so
+        only the behavior near 0 is important. Approximations that overestimate
+        the maximum trace distance are permitted. Even ones that exceed 1.
```
Strilanc: If the value exceeds 1, we will clamp it down to 1 in `trace_distance_bound`.

cirq/protocols/trace_distance_bound_test.py
```diff
@@ -0,0 +1,38 @@
...
+            return NotImplemented
+
+    class ReturnsConstant:
+        def __init__(self, bound):
+            self.bound = bound
+
+        def _trace_distance_bound_(self) -> float:
+            return self.bound
+
+    infinity = float("inf")
```
Strilanc: Add a test of the clamping functionality.

cirq/protocols/trace_distance_bound.py
dstrain115: Changed the verbiage a bit, but added.

PR #1009, opened 2018-10-15
Merge cirq.google.XmonMeasurementGate into cirq.MeasurementGate

Strilanc:
~~Comes after #1007~~


Strilanc:
CLA false positive was due to me having a different base branch than master.

Issue #947, opened 2018-09-19
Refactor PhaseableEffect class into a magic method protocol


Strilanc:
Part of https://github.com/quantumlib/Cirq/issues/672

This task includes:

1. Creating a magic method (akin to `def _unitary_`) that indicates phaseability.
2. Creating a utility method (akin to `cirq.unitary`) to invoke the phaseability magic methods.
3. Refactoring existing code that *uses* operation phasing to instead use the utility method defined in (2). Drop any directly related extensions functionality while doing so.
4. Refactoring existing classes that *provide* phasing functionality to instead use the magic method defined in (1). Drop any directly related extensions functionality while doing so.

dstrain115:
I should be able to get BoundedEffect done this week (or at least our for review), so I think I can get this done in about 2 weeks.

PR #1010, opened 2018-10-15
Replace PhaseableEffect with a new SupportsPhase protocol

dstrain115:
- Add magic method `_phase_by_` to phase a gate or operation
- Add `cirq.protocols.phase_by` to call this magic method or return a
default value.

Note: eject_z logic changed slightly to check if each qubit phase
is allowed rather than checking it at the beginning.  However, this
should have no practical effect.

Fixes https://github.com/quantumlib/Cirq/issues/947


cirq/protocols/phase.py
```diff
@@ -0,0 +1,67 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any, TypeVar
+
+from typing_extensions import Protocol
+
+# This is a special value to indicate that the gate should be returned
+# unchanged if underlying implementation of _phase_by_ exists.
+ReturnSelfIfNotProvided = 1.234567
```
Strilanc: Use a value that has reference identity, so that you can guarantee it is impossible for a caller to produce the same value. `([],)` is a simple example of such a value. (You don't want to use `[]` because it sets off mutable-default-value alarms, and you don't want to use `()` because it's a singleton. `() is ()` but `([],) is not ([],)`)

cirq/protocols/phase.py
```diff
@@ -0,0 +1,67 @@
...
+
+TDefault = TypeVar('TDefault')
+
+
+class SupportsPhase(Protocol):
+    """An effect that can be phased around the Z axis of target qubits."""
+
+    def _phase_by_(self: Any, phase_turns: float, qubit_index: int):
+        """Returns a phased version of the effect.
+        For example, an X gate phased by 90 degrees would be a Y gate.
```
Strilanc: Nit: blank line after summary line

cirq/protocols/phase.py
```diff
@@ -0,0 +1,67 @@
...
+        Args:
+            phase_turns: The amount to phase the gate, in fractions of a whole
```
Strilanc: Add `Divide by 2pi to get radians.`

cirq/protocols/phase.py
```diff
@@ -0,0 +1,67 @@
...
+            qubit_index: The index of the target qubit the phasing applies to.
+        Returns:
+            The phased gate or operation.
+        """
+
+
+def phase_by(val: Any, phase_turns: float, qubit_index: int,
+             default: TDefault = ReturnSelfIfNotProvided):
+    """Returns a phased version of the effect.
+    For example, an X gate phased by 90 degrees would be a Y gate.
```
Strilanc: Nit: add blank line

cirq/protocols/phase.py
```diff
@@ -0,0 +1,67 @@
...
+    This works by calling `val`'s _phase_by_ method and returning
+    the result.
+
+    Args:
+        val: The value to describe with a unitary matrix.
+        phase_turns: The amount to phase the gate, in fractions of a whole
+            turn.
```
Strilanc: Add `Divide by 2pi to get radians.`

cirq/protocols/phase.py
```diff
@@ -0,0 +1,67 @@
...
+        qubit_index: The index of the target qubit the phasing applies to.
+
+    Returns:
+        If `val` has a _phase_by_ method and its result is not NotImplemented,
+        that result is returned. Otherwise, `val` is returned unchanged.
```
Strilanc: This description is not correct. That only happens if a default wasn't specified.

cirq/protocols/phase.py
Strilanc: The default behavior should be to raise a `TypeError`, not to return self. If a gate can't explain to you what conjugating it with a Z operation will do, it is very bad to assume it will simply do nothing.

dstrain115: Changed this behavior and will push a new commit once I get the tests working again.

dstrain115:
Yes please.  I am not quite sure why Exp11Gate disappeared.

Strilanc:
It's because I merged it into the normal CZ. I'll take it from here.

Strilanc:
Well that's a new race condition.

PR #1011, opened 2018-10-16
Remove cirq.ReversibleEffect in favor of cirq.inverse and __pow__

Strilanc:
Part of https://github.com/quantumlib/Cirq/issues/943

### DIFF
--- a/cirq/contrib/paulistring/pauli_string_phasor.py
+++ b/cirq/contrib/paulistring/pauli_string_phasor.py
@@ -30,8 +30,7 @@ class PauliStringPhasor(PauliStringGateOperation,
                         ops.CompositeOperation,
                         ops.TextDiagrammable,
                         extension.PotentialImplementation[Union[
-                            ops.ExtrapolatableEffect,
-                            ops.ReversibleEffect]]):
+                            ops.ExtrapolatableEffect]]):
     """An operation that phases a Pauli string."""
     def __init__(self,
                  pauli_string: PauliString,
@@ -93,9 +92,6 @@ class PauliStringPhasor(PauliStringGateOperation,
             return NotImplemented
         return self.extrapolate_effect(power)
 
-    def inverse(self) -> 'PauliStringPhasor':
-        return self.extrapolate_effect(-1)
-
     def can_merge_with(self, op: 'PauliStringPhasor') -> bool:
         return self.pauli_string.equal_up_to_sign(op.pauli_string)
 
@@ -143,8 +139,7 @@ class PauliStringPhasor(PauliStringGateOperation,
                     desired_type: Type[T_DESIRED],
                     ext: extension.Extensions
                     ) -> Optional[T_DESIRED]:
-        if (desired_type in [ops.ExtrapolatableEffect,
-                             ops.ReversibleEffect] and
+        if (desired_type in [ops.ExtrapolatableEffect] and
                 not self._is_parameterized_()):
             return cast(T_DESIRED, self)
         return super().try_cast_to(desired_type, ext)
--- a/cirq/google/xmon_gates.py
+++ b/cirq/google/xmon_gates.py
@@ -260,9 +260,7 @@ class Exp11Gate(XmonGate,
 class ExpWGate(XmonGate,
                ops.SingleQubitGate,
                ops.TextDiagrammable,
-               ops.PhaseableEffect,
-               PotentialImplementation[Union[
-                   ops.ReversibleEffect]]):
+               ops.PhaseableEffect):
     """A rotation around an axis in the XY plane of the Bloch sphere.
 
     This gate is a "phased X rotation". Specifically:
@@ -345,20 +343,6 @@ class ExpWGate(XmonGate,
         if protocols.is_parameterized(self) and power != 1:
             return NotImplemented
         return ExpWGate(half_turns=self.half_turns * power,
-                        axis_half_turns=self.axis_half_turns)
-
-    def try_cast_to(self, desired_type, ext):
-        if desired_type is ops.ReversibleEffect and self.has_inverse():
-            return self
-        return super().try_cast_to(desired_type, ext)
-
-    def has_inverse(self):
-        return not isinstance(self.half_turns, value.Symbol)
-
-    def inverse(self):
-        if not self.has_inverse():
-            raise ValueError("Don't have a known inverse.")
-        return ExpWGate(half_turns=-self.half_turns,
                         axis_half_turns=self.axis_half_turns)
 
     def _unitary_(self) -> Union[np.ndarray, type(NotImplemented)]:
--- a/cirq/google/xmon_gates.py
+++ b/cirq/google/xmon_gates.py
@@ -20,7 +20,6 @@ from typing import Dict, Optional, Tuple, Union
 import numpy as np
 
 from cirq import abc, ops, value, protocols
-from cirq.extension import PotentialImplementation
 from cirq.devices.grid_qubit import GridQubit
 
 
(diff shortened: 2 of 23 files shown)


cirq/contrib/paulistring/pauli_string_phasor.py
```diff
@@ -93,9 +92,6 @@ def __pow__(self, power: Union[float, value.Symbol]) -> 'PauliStringPhasor':
             return NotImplemented
         return self.extrapolate_effect(power)
 
-    def inverse(self) -> 'PauliStringPhasor':
-        return self.extrapolate_effect(-1)
-
```
dstrain115: Presumably, this deletion is fine since this class already supports _pow_.  Is that correct?

cirq/google/xmon_gates.py
```diff
@@ -347,20 +344,6 @@ def __pow__(self, power):
...
-
-    def has_inverse(self):
-        return not isinstance(self.half_turns, value.Symbol)
-
-    def inverse(self):
-        if not self.has_inverse():
-            raise ValueError("Don't have a known inverse.")
-        return ExpWGate(half_turns=-self.half_turns,
-                        axis_half_turns=self.axis_half_turns)
-
```
dstrain115: Actually, ExpWGate does not have _pow_ that I can see.  Doesn't this represent a slight drop in functionality then?

cirq/ops/reversible_composite_gate.py
```diff
@@ -23,34 +23,26 @@
 
 
 class ReversibleCompositeGate(gate_features.CompositeGate,
-                              gate_features.ReversibleEffect,
                               metaclass=abc.ABCMeta):
     """A composite gate that gets decomposed into reversible gates."""
 
-    def __pow__(self, power):
+    def __pow__(self: TOriginal,
+                power) -> '_ReversedReversibleCompositeGate[TOriginal]':
```
dstrain115: ReversedReversibleCompositeGate?  I guess it's okay, but it is a really goofy name.

cirq/protocols/inverse.py
```diff
@@ -12,16 +12,54 @@
...
 
 # This is a special indicator value used by the inverse method to determine
 # whether or not the caller provided a 'default' argument.
 RaiseTypeErrorIfNotProvided = ([],)
 
 
+TDefault = TypeVar('TDefault')
+
+
+# pylint: disable=function-redefined
```
dstrain115: I haven't done that much python with type checking.  Is this a common thing to do when a function can have multiple type inputs?  It looks weird.  Maybe it deserves a comment to explain what the purpose of all these stub functions are for?

docs/gates.md
```diff
@@ -16,18 +16,14 @@ print(CNOT(q0, q1))
...
-features a ``Gate`` has.  
-
-For example, one feature is ``ReversibleEffect``.
-A ``Gate`` that inherits this class is required to implement the method ``inverse`` which returns the inverse gate.
-Algorithms that operate on gates can use ``isinstance(gate, ReversibleEffect)`` to determine whether gates implements ``inverse`` method, and then use it.
-(Note that, even if the gate is not reversible, the algorithm may have been given an ``Extension`` with a cast from the gate to ``ReversibleEffect``.
-See the [extensions documentation](extensions.md) for more information.)
+### Magic Methods
+
+A raw ``Gate`` class can be applied to qubits to produce an ``Operation``, but that's about it.
```
dstrain115: I would remove ", but that's about it".  It seems very informal.

docs/gates.md
```diff
@@ -16,18 +16,14 @@ print(CNOT(q0, q1))
...
+In order to support other functionality, such as being understood by simulators or looking good in diagrams, it is necessary to implement several *magic methods*.
+A magic method is a special method that, if present on a value, enables new functionality (think `__add__`, `__eq__`, `__len__`, and etc).
```
dstrain115: This is an awkward sentence.  Here is my suggestion:

Standard magic methods in python are `__add__`, `__eq__`, and `__len__`.  Cirq defines several additional magic methods, specified below, that enable functionality such as parameterization, text diagramming, or unitary matrix representations.

cirq/contrib/paulistring/pauli_string_phasor.py
Strilanc: Correct.

cirq/ops/reversible_composite_gate.py
Strilanc: Ack. It's private, so it's fine if it's long-winded.

cirq/protocols/inverse.py
Strilanc: Ack. This is a standard feature of mypy. If you search `python @overload` or similar you pretty quickly come across what it means. This happens in many places in the code base, and it's not really practical to explain it in every place.

docs/gates.md
Strilanc: Fair point. That's my style of writing, even for technical things. Maybe especially for technical things. I would worry more about the documentation having "many voices" before I worried about it being informal.

This all has to be more extensively rewritten (e.g. the very next section is back to talking about extensions) so I won't worry about it too much here.

docs/gates.md
Strilanc: Incorporated. I rewrote the whole paragraph based on some of your suggestions.

cirq/google/xmon_gates.py
dstrain115: New doc looks fine.  Can you address this comment?  Seems like ExpWGate should have a pow implementation for -1?

Also, looks like build failed.

Strilanc: ExpW's `__pow__` method is defined on line 344.

Strilanc:
Triggered rebuild.

Travis has been really flaky lately.

cirq/google/xmon_gates.py
dstrain115: Ok, I must be really tired, I went through the file twice and didn't see it either time.  LGTM then.

Strilanc:
@dstrain115 You have to go to the 'Files Changes' tab, hit 'review changes' in the top right, and select approve then submit in order to LGTM.

cirq/contrib/paulistring/pauli_string_phasor.py
```diff
@@ -143,8 +139,7 @@ def try_cast_to(self,
                     desired_type: Type[T_DESIRED],
                     ext: extension.Extensions
                     ) -> Optional[T_DESIRED]:
-        if (desired_type in [ops.ExtrapolatableEffect,
-                             ops.ReversibleEffect] and
+        if (desired_type in [ops.ExtrapolatableEffect] and
```
dabacon: does this need to be "in" now?

docs/gates.md
```diff
@@ -16,20 +16,17 @@ print(CNOT(q0, q1))
...
-See the [extensions documentation](extensions.md) for more information.)
+Standard magic methods in python are `__add__`, `__eq__`, and `__len__`.
+Cirq defines several additional magic methods, for functionality such as parameterization, diagramming, and simulation.
+For example, if a gate specifies a `_unitary_` method that returns a matrix for the gate, then simulators will be able to simulate applying the gate.
+Or, if a gate specifies a `__pow__` method that works for an exponent of -1, then `cirq.inverse` will start to work on lists including the gate.
 
-We describe some gate features below.
+We describe some magic methods below.
 
 #### ReversibleEffect, SelfInverseGate
```
dabacon: Remove ReversibleEffect

cirq/contrib/paulistring/pauli_string_phasor.py
Strilanc: Acknowledged. It doesn't, but this code is going to be deleted within a week when the same thing happens to ExtrapolatableEffect so it's not worth polishing.

Issue #1006, opened 2018-10-15
Schedules.md has a bad circuit diagram but isn't failing the snippet tests


Strilanc:
Look for:

```
print(circuit)
# prints:
# (0, 0): ───Z───X───
#            │
# (1, 0): ───Z───────
```

Z-Z is not how CZs are printed anymore.

vtomole:
How do I self-assign to issues?

Strilanc:
I gave you "collaborator: read" access to the repo, but that may not include the ability to assign issues. If you can, you click on `Assignees` in the right hand column and set it to yourself.

If you just leave a comment saying "I am working on this issue, please assign me", that is sufficient for now. I'll think about some kind of workaround, e.g. CirqBot watching for particular comments or good 'ol complaining to github that read access should include the ability to self-assign unassigned issues.

vtomole:
> may not include the ability to assign issues

Yup. That only works for personal repos, not organizations'. Manual or CirqBot is fine.

PR #1012, opened 2018-10-16
Remove semicolon to run snippets test in schedules.md

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/1006


vtomole:
All the ones I looked at only had `# prints`. Completely missed the `# prints something like`. Fix coming up!

vtomole:
@Strilanc It's a spacing problem. Do we want to do an

`if '# prints' elif ' # prints:'`
Or check if '# prints' is in the first 8 characters? 
The second solution would make  `# printssomething like` legal.

Strilanc:
Let's add the following variations:

```
# print
# prints
# print:
# prints:
```

So, in regex form, `^#\s*prints?:?\s*$`. But perhaps it's clearer to do cases instead of that. Up to you.

vtomole:
@Strilanc That requires write access. Makes me wonder how different read access and no access is: https://help.github.com/articles/repository-permission-levels-for-an-organization/.

Strilanc:
If you aren't a collaborator at all, I can't assign issues to you.

docs/snippets_test.py
```diff
@@ -406,7 +406,8 @@ def find_expected_outputs(snippet: str) -> List[str]:
                 expected.append(rest)
             else:
                 printing = False
-        elif line.startswith(start_key):
+        # Matches '# print', '# prints', '# print:', and '# prints:'
+        elif re.match('^# prints?:?\s*$', line):
```
Strilanc: We should also match `#prints:` and similar. Throw in a `\s*` after the `#`.

Also, add tests exercising the cases.

vtomole: Yup. Not done yet as i realized this didn't fix the original error.

vtomole: Oh yeah the reason i didn't include the `\s*` was because it broke one of the tests. If we add it, this will be valid 
```
#prints something like (The other lines have spaces before comments begin except this one)
# transverse fields: [[-1, 1, -1], [1, -1, -1], [-1, 1, -1]]
# row j fields: [[1, 1, -1], [1, -1, 1]]
# column j fields: [[1, -1], [-1, 1], [-1, 1]]
```

vtomole: Not aesthetic, but if we want to give the user the option of doing it, i'll add it.

Strilanc: It's fine to require that, because it causes a loud failure instead of a silent failure. But I agree that it could be nice to allow the content to be indented (but the indent should be consistent across the whole block of text). Out of scope for this PR.

docs/snippets_test.py
```diff
@@ -419,10 +418,38 @@ def _indent(lines: List[str]) -> str:
...
+
+    assert find_expected_outputs("""
+# prints:
+# abc
+
+# def
+    """) == ['abc']
+
+    assert find_expected_outputs("""
+# prints:
```
Strilanc: Add one more test case for `#print:` and we're good to go.

Strilanc:
@vtomole Thanks, I'll keep that in mind. There was a huge slowdown these last couple weeks due to @dabacon being busy so more reviewers is useful.

PR #1017, opened 2018-10-18
Special case single-qubit unitaries in _apply_unitary_to_tensor_

Strilanc:
This is 20% faster than using einsum.

Setup:

```
import cirq
import numpy as np
c = cirq.Circuit.from_ops(cirq.X(q)**0.1 for q in cirq.LineQubit.range(24))
```

Before:

```
%time _ = c.apply_unitary_effect_to_state(dtype=np.complex64)
CPU times: user 4.05 s, sys: 611 ms, total: 4.66 s
Wall time: 4.66 s
```

After:

```
%time _ = c.apply_unitary_effect_to_state(dtype=np.complex64)
CPU times: user 3.38 s, sys: 279 ms, total: 3.66 s
Wall time: 3.66 s
```

### DIFF
--- a/cirq/linalg/transformations.py
+++ b/cirq/linalg/transformations.py
@@ -211,8 +211,9 @@ def apply_matrix_to_slices(
 
     # Apply operation.
     for i, s_i in enumerate(slices):
-        out[s_i] = 0
+        out[s_i] *= matrix[i, i]
         for j, s_j in enumerate(slices):
-            out[s_i] += target[s_j] * matrix[i, j]
+            if i != j:
+                out[s_i] += target[s_j] * matrix[i, j]
 
     return out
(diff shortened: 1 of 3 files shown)


Strilanc:
For whatever reason, this approach is slower with 4x4 matrices than einsum. Goes from 1.95s to 3.38s in a test I did with 22 operations on 22 qubits. If I had to guess, I'd say it's because we do 12 slightly-faster passes over the data instead of one slightly-slower one. Whereas for a 2x2 matrix it's 2 slightly-faster passes vs one slightly-slower one, and all the "slightlys" line up to make that faster whereas 4x4 is slower.

(Qubit adjacency isn't relevant to this particular method.)

PR #1023, opened 2018-10-18
Fix protocol definition for trace_distance_bound

dabacon:


Strilanc:
There was some kind of weird problem where travis put two statuses on the PR.

Strilanc:
Very confused about these double statuses

Issue #988, opened 2018-10-06
Lower numpy requirement


kevinsung:
What prevents us from getting down to 1.12, which is what is available on Debian stable?

vtomole:
I'm not sure; but i don't see why we should change it to make a specific OS and distro happy. If more than one OS (like Mac) also has 1.12 as the default then it would make sense to change it.

PR #1026, opened 2018-10-19
Loosen numpy version to >=1.12.0

Strilanc:
- Expose highlight_text_differences from cirq.testing
- Do a better job of canonicalizing numpy text output for snippe tests

I ran the checks locally while forcing the numpy version to 1.12 and it worked. Not sure how to guarantee we don't break the snippet tests in older versions of numpy. But... it's probably fine as long as they pass in *some* version.

Fixes https://github.com/quantumlib/Cirq/issues/988


python2.7-requirements.txt
```diff
@@ -3,7 +3,7 @@
 google-api-python-client~=1.6
 matplotlib~=2.1
 networkx~=2.1
-numpy~=1.14
+numpy>=1.12.0
```
dabacon: Why not ~=?

Strilanc: Done.

Issue #672, opened 2018-07-15
Remove Extensions in favor of protocols


Strilanc:
I've been stewing on this for the last few days.

Originally I added the extensions mechanism in order to solve something akin to the [expression problem](https://en.wikipedia.org/wiki/Expression_problem). Since then it kind of grew into a convenient way to ask things if they support a feature, to the point where we have `cirq.cast`.

But what value are we really getting out of this? It is easily the most abstract and complicated-to-use-correctly part of the library. Worryingly, it has wound its way into the way you perform basic tasks such as defining a gate. In principle it allows people to post-facto add features to a class and explain to library A how to use library B, but we're really not at the level of scale where that is a problem in the first place.

More seriously, I think extensions is doing something that is really a language-level task. It is not the job of a library to solve these kinds of problems. Or, at least, you would want the library to *only* be about that. Because otherwise people coming in from the rest of python-land have no idea why everything is done differently. It is an extremely unpythonic feature.

**I think we should cut the extensions mechanism out**. The key thing that would make this possible is to refactor all marker classes to have "can't do that" outputs. For example, `KnownMatrix.matrix()` would be allowed to return `None` to signify "I do not have a known matrix [e.g. because I am parameterized]". This would obsolete `PotentialImplementation`.

All of that being said, there may still be some use for extensions. But I think we should think of it more like a way to make existing gates smoothly go through your own code as part of implementing a feature, as opposed to defensively asking for an `ext` instance that you don't know what it would be used for.

Please chime in with your opinions.

Strilanc:
Very well, I think that makes a strong consensus. I will start cutting down the extensions tree.

Strilanc:
After playing this out a bit for `KnownMatrix`, and thinking about it overnight, I have a different idea of the refactor should play out.

Instead of requiring classes to implement `KnownMatrix`, it should be more like a "reference type" that we use to make mypy happy. At runtime all we should care about is the presence or absence of specific methods. This would put us inline with how the rest of the python ecosystem works. For example, there is no "Addable" interface there is only the `__add__` method. You make yourself addable by implementing `__add__`, end of story. Numpy has a similar strategy, but with methods like `exp`; if you call `np.exp(c)` and `c` is an instance of a custom class that defines an `exp` method, numpy will delegate to that method instead of failing.

[...]

PR #1029, opened 2018-10-19
Remove `extension` parameters and fields from most methods and classes

Strilanc:
- Also refactor `Circuit` to match `SupportsParameterization` protocol instead of having custom resolve/is methods

Part of https://github.com/quantumlib/Cirq/issues/672


cirq/circuits/circuit.py
```diff
@@ -1124,23 +1124,23 @@ def to_unitary_matrix(
...
             qubits_that_should_be_present: Qubits that may or may not appear
                 in operations within the circuit, but that should be included
                 regardless when generating the matrix.
             ignore_terminal_measurements: When set, measurements at the end of
                 the circuit are ignored instead of causing the method to
                 fail.
-            dtype: The numpy dtype for the returned unitary. Must be a complex
-                dtype.
+            dtype: The numpy dtype for the returned unitary. Defaults to
+                np.complex128. Specifying np.complex64 will run faster at the
```
dabacon: Aside: why default to complex128?  Isn't complex64 usually sufficient for NISQ.  Also this now does not match the quantum engine simulator.

Strilanc: The reason I defaulted to complex128 was because tests tend to fail at `atol=1e-8` when you use complex64.

I agree that we should probably change this. Makes it faster, doesn't give up much of relevance. Opened https://github.com/quantumlib/Cirq/issues/1030

PR #1032, opened 2018-10-19
First take at depolarizing channels

dabacon:
Some thoughts:
* Decided not to allow parameter for channel to be a symbol.  
* Used delegation.


cirq/ops/common_channels.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def __str__(self) -> str:
+        return ('AsymmetricDepolarizingChannel(p_x={!r},p_y={!r},p_z={!r})'
+                .format(self._p_x, self._p_y, self._p_z))
+
+    def _circuit_diagram_info_(self,
+        args: protocols.CircuitDiagramInfoArgs) -> protocols.CircuitDiagramInfo:
+        return protocols.CircuitDiagramInfo(
+            wire_symbols=('A',),
+            exponent='({!r},{!r},{!r})'.format(self._p_x, self._p_y, self._p_z))
```
Strilanc: If the exponent can't ever be 1, you might as well put it in the symbol. If you don't use any features beyond symbols the new protocol allows you to simply return the string.

cirq/ops/common_channels.py
```diff
@@ -0,0 +1,146 @@
...
+            if p < 0:
+                raise ValueError('{} was less than 0.'.format(p_str))
+            elif p > 1:
+                raise ValueError('{} was greater than 1.'.format(p_str))
+            return p
+        self._p_x = validate_probability(p_x, 'p_x')
+        self._p_y = validate_probability(p_y, 'p_y')
+        self._p_z = validate_probability(p_z, 'p_z')
+        self._p_i = 1 - validate_probability(p_x + p_y + p_z, 'p_x + p_y + p_z')
+
```
Strilanc: Add a `__pow__` method that scales the probabilities as if repeating the channel.

cirq/ops/common_channels.py
```diff
@@ -0,0 +1,146 @@
...
+                .format(self._p_x, self._p_y, self._p_z))
+
+    def _circuit_diagram_info_(self,
+        args: protocols.CircuitDiagramInfoArgs) -> protocols.CircuitDiagramInfo:
+        return protocols.CircuitDiagramInfo(
+            wire_symbols=('A',),
+            exponent='({!r},{!r},{!r})'.format(self._p_x, self._p_y, self._p_z))
+
+
+class DepolarizingChannel(raw_types.Gate):
```
Strilanc: Could we make this a constant instead of a class? Similar to how `cirq.X**t` gives a convenient way to get at X rotations we could have `HALF_DEPOLARIZE**t`. That does feel a bit more awkward..

cirq/ops/common_channels.py
```diff
@@ -0,0 +1,146 @@
...
+    def __repr__(self) -> str:
+        return '(cirq.DepolarizingChannel(p={!r}))'.format(self._p)
+
+    def __str__(self) -> str:
+        return 'DepolarizingChannel(p={!r})'.format(self._p)
+
+    def _circuit_diagram_info_(self,
+        args: protocols.CircuitDiagramInfoArgs) -> protocols.CircuitDiagramInfo:
+        return protocols.CircuitDiagramInfo(
+            wire_symbols=('D',), exponent='({!r})'.format(self._p))
```
Strilanc: Is the gate *actually* equal to some `D` to the power p?

cirq/ops/common_channels.py
```diff
@@ -142,5 +148,9 @@ def __str__(self) -> str:
...
     def _circuit_diagram_info_(self,
         args: protocols.CircuitDiagramInfoArgs) -> protocols.CircuitDiagramInfo:
-        return protocols.CircuitDiagramInfo(
-            wire_symbols=('D',), exponent='({!r})'.format(self._p))
+        return 'D({!r})'.format(self._p)
+
+
+def depolarize(p):
+    """Returns a DepolarizingChannel with given probability of error."""
+    return DepolarizingChannel(p)
```
Strilanc: Nit: newline

cirq/ops/common_channels.py
```diff
@@ -142,5 +148,9 @@ def __str__(self) -> str:
 
     def _circuit_diagram_info_(self,
         args: protocols.CircuitDiagramInfoArgs) -> protocols.CircuitDiagramInfo:
-        return protocols.CircuitDiagramInfo(
-            wire_symbols=('D',), exponent='({!r})'.format(self._p))
+        return 'D({!r})'.format(self._p)
+
+
+def depolarize(p):
```
Strilanc: Add `args` documentation and type annotations.

cirq/ops/common_channels.py
```diff
@@ -89,10 +89,13 @@ def __str__(self) -> str:
...
     def _circuit_diagram_info_(self,
-        args: protocols.CircuitDiagramInfoArgs) -> protocols.CircuitDiagramInfo:
-        return protocols.CircuitDiagramInfo(
-            wire_symbols=('A',),
-            exponent='({!r},{!r},{!r})'.format(self._p_x, self._p_y, self._p_z))
+        args: protocols.CircuitDiagramInfoArgs) -> str:
+        return 'A({!r},{!r},{!r})'.format(self._p_x, self._p_y, self._p_z)
+
+
+def asymmetric_depolarize(p_x, p_y, p_z):
```
Strilanc: Add `args` documentation and type annotations.

Strilanc: Could this be merged into the `depolarize` method?

```python3
def depolarize(p=0, *, p_x=0, p_y=0, p_z=0):
    p_x = 1 - (1 - p) * (1 - p_x)
    p_y = 1 - (1 - p) * (1 - p_y)
    p_z = 1 - (1 - p) * (1 - p_z)

    # alternatively, the individual ones override p as the default one

    return AsymmetricDepolarizingChannel(p_x, p_y, p_z)
```

Strilanc: Is this actually sufficient to specify any kind of depolarization? For example, measurement along the X+Z axis doesn't seem to be representable without introducing surrounding unitaries.

cirq/ops/common_channels.py
dabacon: fixed

cirq/ops/common_channels.py
```diff
@@ -0,0 +1,188 @@
...
+from cirq.ops import raw_types
+
+
+class AsymmetricDepolarizingChannel(raw_types.Gate):
+    """A channel that depolarizes asymmetrically along different directions."""
+
+    def __init__(self, p_x: float, p_y: float, p_z: float) -> None:
+        """The asymmmetric depolarizing channel.
+
+        This channel applies one of the three Pauli operators at random,
```
Strilanc: I didn't realize that these probabilities were describing disjoint events instead of independent events. Rewrite to make this clearer.

cirq/ops/common_channels.py
```diff
@@ -0,0 +1,188 @@
...
+        or does nothing.  The supplied probabilities must be valid probabilities
+        and the sum p_x + p_y + p_z must be a valid probability or else this
+        constructor will raise a ValueError.
+
+        This channel evolves a density matrix via
+            \rho -> (1 -p_x + p_y + p_z) \rho
+                    + p_x X \rho X + p_y Y \rho Y + p_z Z \rho Z
+
+        Args:
+            p_x: The probability that a Pauli X error occurs.
```
Strilanc: Add ` (and no other error)` to each of these

cirq/ops/common_channels.py
```diff
@@ -0,0 +1,188 @@
...
+        return (
+            np.sqrt(self._p_i) * np.eye(2),
+            np.sqrt(self._p_x) * np.array([[0, 1], [1, 0]]),
+            np.sqrt(self._p_y) * np.array([[0, -1j], [1j, 0]]),
+            np.sqrt(self._p_z) * np.array([[1, 0], [0, -1]])
+        )
+
+    def _eq_tuple(self):
+        return (AsymmetricDepolarizingChannel,
+                self._p_i, self._p_x, self._p_y, self._p_z)
```
Strilanc: Don't need to include `_p_i`.

cirq/ops/common_channels.py
```diff
@@ -0,0 +1,188 @@
...
+        def validate_probability(p, p_str):
+            if p < 0:
+                raise ValueError('{} was less than 0.'.format(p_str))
+            elif p > 1:
+                raise ValueError('{} was greater than 1.'.format(p_str))
+            return p
+        self._p_x = validate_probability(p_x, 'p_x')
+        self._p_y = validate_probability(p_y, 'p_y')
+        self._p_z = validate_probability(p_z, 'p_z')
+        self._p_i = 1 - validate_probability(p_x + p_y + p_z, 'p_x + p_y + p_z')
```
Strilanc: I would change `_p_i` into a property (but still validate it here).

cirq/ops/common_channels.py
```diff
@@ -0,0 +1,188 @@
...
+        if not isinstance(other, type(self)):
+            return NotImplemented
+        return self._eq_tuple() == other._eq_tuple()
+
+    def __ne__(self, other):
+        return not self == other
+
+    def __repr__(self) -> str:
+        return (
+            '(cirq.AsymmetricDepolarizingChannel(p_x={!r},p_y={!r},p_z={!r}))'
```
Strilanc: The outside parens in the repr are unnecessary.

cirq/ops/common_channels_test.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def test_asymmetric_depolarizing_channel_str():
+    assert (str(cirq.asymmetric_depolarize(0.1, 0.2, 0.3))
+            == 'AsymmetricDepolarizingChannel(p_x=0.1,p_y=0.2,p_z=0.3)')
+
+
+def test_asymmetric_depolarizing_channel_eq():
+    et = cirq.testing.EqualsTester()
+    et.add_equality_group(cirq.asymmetric_depolarize(0.0, 0.0, 0.0))
```
Strilanc: This doesn't actually check that anything is ever equal to anything else. It would be satisfied by `def __eq__(a, b): return a is b`.

cirq/ops/common_channels_test.py
```diff
@@ -0,0 +1,119 @@
...
+    cirq.testing.assert_equivalent_repr(cirq.DepolarizingChannel(0.3))
+
+
+def test_depolarizing_channel_str():
+    assert str(cirq.depolarize(0.3)) == 'DepolarizingChannel(p=0.3)'
+
+
+def test_depolarizing_channel_eq():
+    et = cirq.testing.EqualsTester()
+    et.add_equality_group(cirq.depolarize(0.0))
```
Strilanc: Need a group with more than one member (you can use the `et.make_equa...` method to produce two from a lambda)

cirq/ops/common_channels.py
dabacon: Ha yeah that's sort of the definition :)  Rewrote to make this clear.

cirq/ops/common_channels.py
dabacon: modified (not quite what you said we shouldn't use "error")

cirq/ops/common_channels.py
dabacon: done

cirq/ops/common_channels_test.py
dabacon: Thanks kind of annoying

cirq/ops/common_channels.py
dabacon: OK.  There are a bunch of places where there are ()s in reprs.

cirq/ops/common_channels.py
Strilanc: It's fine as is; just a passing comment.

cirq/ops/common_channels.py
Strilanc: The general rule of thumb is "if it's of the form X**Y then you need them, because then appending `.stuff()` won't work". Use `cirq.testing.assert_equivalent_repr` to validate that a parens-less repr is correct.

cirq/ops/common_channels_test.py
Strilanc: Well, *technically* it compares the objects against themselves. But if you completely forgot to implement `__eq__` the fallback `is` check will pass the tests.

This is just a general feature of tests, I think: no matter how many input cases you give, it might as well just be one if you didn't exercise multiple output cases.

cirq/ops/common_channels.py
Strilanc: You can specify this repr and str in terms of `asymmetric_depolarize` if you want.

cirq/ops/common_channels.py
```diff
@@ -0,0 +1,188 @@
...
+    def __eq__(self, other):
+        if not isinstance(other, type(self)):
+            return NotImplemented
+        return self._p == other._p
+
+    def __ne__(self, other):
+        return not self == other
+
+    def __repr__(self) -> str:
+        return '(cirq.DepolarizingChannel(p={!r}))'.format(self._p)
```
Strilanc: Unnecessary parens on repr. Consider using `cirq.depolarize(p)` instead, since that's the language users will use.

cirq/ops/common_channels.py
dabacon: Huh it looks like my change didn't get pushed.

Strilanc:
Still LGTM.

Issue #945, opened 2018-09-19
Refactor CompositeOperation class into a magic method protocol


Strilanc:
Part of https://github.com/quantumlib/Cirq/issues/672

This task includes:

1. Creating a magic method (akin to `def _unitary_`) that indicates decomposability.
2. Creating a utility method (akin to `cirq.unitary`) that gets the decomposition, if present.
3. Improving the documentation of what decomposition means / deciding what it should mean in the first place.
    - I think it should mean "give me something that eventually recursively decomposes into 1 and 2 qubit operations defined in `cirq.ops`".
4. Refactoring existing code that *uses* decomposition functionality to instead use the utility method defined in (2). Drop any directly related extensions functionality while doing so.
5. Refactoring existing classes that *provide* decomposition functionality to instead use the magic method defined in (1). Drop any directly related extensions functionality while doing so.
6. Clean up any remaining snags, such as optimization classes that ask for an extensions object but no longer use it.

Strilanc:
I would really like if `cirq.decompose` was a bit configurable, so that callers didn't have to handle recursive decomposition on their own and could generally control the direction decomposition was going. Something like this:

```python3
@overload
def decompose(circuit_or_operations: circuits.Circuit,
              decomposer_func: Callable[[ops.Operation], ops.OP_TREE]
              ) -> circuits.Circuit:
    pass


@overload
def decompose(circuit_or_operations: ops.OP_TREE,
              decomposer_func: Callable[[ops.Operation], ops.OP_TREE]
              ) -> List[ops.Operation]:
        pass


def _default_decompose_until_atomic(op: ops.Operation) -> ops.OP_TREE:
    c = extension.try_cast(ops.CompositeOperation, op)
    if c is None:
        return op
    return protocols.decompose(c)


def decompose(
        circuit_or_operations: Union[circuits.Circuit, ops.OP_TREE],
        decomposer_func: Callable[[ops.Operation],
                                  ops.OP_TREE] = _default_decompose_until_atomic
) -> Union[circuits.Circuit, List[ops.Operation]]:
    if isinstance(circuit_or_operations, circuits.Circuit):
        return circuits.Circuit.from_ops(
            _decompose_tree(circuit_or_operations.all_operations(),
                            decomposer_func))

    return list(ops.flatten_op_tree(
        _decompose_tree(circuit_or_operations, decomposer_func)))


def _decompose_tree(root: ops.OP_TREE,
[...]

PR #1035, opened 2018-10-19
Add cirq.decompose[_once][_with_qubits] protocol

Strilanc:
Part of https://github.com/quantumlib/Cirq/issues/945

Looking for comments on how this is designed. The reason `cirq.decompose` is recursive is because that seems to be how it's always used in the code.

For reference here is some code using `cirq.decompose`:

```python3
    def convert(self, root: ops.Operation) -> ops.OP_TREE:

        def single_qubit_converter(op):
            if len(op.qubits) == 1:
                mat = protocols.unitary(op, None)
                if mat is not None:
                    cliff_op = self._matrix_to_clifford_op(mat, op.qubits[0])
                    if cliff_op is not None:
                        return cliff_op
                    if self.ignore_failures:
                        return op

            return NotImplemented

        return protocols.decompose(
            root,
            intercepting_decomposer=single_qubit_converter,
            keep=lambda op: isinstance(op, ops.GateOperation) and
                            isinstance(op.gate, ops.SingleQubitCliffordGate),
            on_stuck_raise=(
                None
                if self.ignore_failures else
                lambda op: TypeError(
                    "Don't know how to work with {!r}. "
                    "It isn't a CompositeOperation or a 1-qubit operation "
                    "with a known clifford effect.".format(op))))
```


cirq/protocols/decompose.py
```diff
@@ -0,0 +1,234 @@
...
+            return r
+    return NotImplemented
+
+
+def decompose(
+    val: Any,
+    *,
+    intercepting_decomposer: Callable[['cirq.Operation'],
+                                      'cirq.OP_TREE'] = None,
+    decomposer: Callable[['cirq.Operation'],
```
dabacon: I don't quite see the point of having the decomposer arg in decompose.  When would you pass this in instead of an intercepting_decomposer?

cirq/protocols/decompose.py
```diff
@@ -0,0 +1,234 @@
...
+
+    return output
+
+
+def decompose_once(val: Any,
+                   default: TDefault=RaiseTypeErrorIfNotProvided,
+                   **kwargs
+                   ) -> Union[TDefault, List['cirq.Operation']]:
+    """Decomposes a value into operations, if possible.
+
```
dabacon: doc to that this is not recursive for completeness

cirq/protocols/decompose.py
```diff
@@ -0,0 +1,234 @@
...
+    if default is not RaiseTypeErrorIfNotProvided:
+        return default
+    if method is None:
+        raise TypeError("object of type '{}' "
+                        "has no _decompose_ method.".format(type(val)))
+    raise TypeError("object of type '{}' does have a _decompose_ method, "
+                    "but it returned NotImplemented.".format(type(val)))
+
+
+def decompose_once_with_qubits(
```
dabacon: could both decompose_once and decompose take in an optional Qubits for this case?

cirq/protocols/decompose.py
```diff
@@ -0,0 +1,234 @@
...
+            main decomposer when doing intermediate decompositions. If
+            `intercepting_decomposer` isn't specified, or returns
+            `NotImplemented`, the decomposition falls back to the main
+            `decomposer`. Otherwise the result of `intercepting_decomposer`
+            takes priority and `decomposer` isn't even called.
+        decomposer: Decomposes operations into more basic operations, or returns
+            `NotImplemented` to indicate that an operation cannot be decomposed
+            anymore. Defaults to a decomposer that calls operations'
+            `_decompose_` method.
+        keep: A predicate that determines if the initial operation or
```
dabacon: keep feels a bit ambiguous to me.  It makes sense in terms of creating the list of outputs, but I'm not sure stoping decomposing would be thought of as "keep"ing.  Maybe
is_decomposed?
is_fully_decomposed?
or invert it: 
needs_more_decomposition?

cirq/protocols/decompose.py
Strilanc: Hmmm... think you're right. I introduced `decomposer` first so it felt essential to me, but all the use cases I've written ended up using `intercepting_decomposer`. You can even emulate `decomposer` by throwing an exception out of `intercepting_decomposer`.

I wish I could come up with a better name than `intercepting_decomposer`. Stuff like `first_decomposer` or `lead_decomposer` seems like it might be confused with "at the top level you apply this one, but not when recursing".

cirq/protocols/decompose.py
Strilanc: Done.

cirq/protocols/decompose.py
Strilanc: I tried doing that initially. The code looked gross. This way is clearer and makes it harder for the caller to make a mistake such as thinking they need to pass in `operation.qubits` in order to decompose an operation.

cirq/protocols/decompose.py
Strilanc: Acknowledged. I think this is a good name in the same sense that `key` is a good argument for `sorted`. It's a vague word that only works because the context pushes you toward one particular interpretation. It makes up for that by being short and memorable.

cirq/protocols/decompose.py
Strilanc: Done.

cirq/protocols/decompose.py
dabacon: OK let's watch if this confuses people.  I think the problem is that keep is close to "keep going" in my head.

PR #1038, opened 2018-10-22
Fix circuit equivalence

### LINKED ISSUE #1019: assert_circuits_with_terminal_measurements_are_equivalent fails when CZs are present before measurements
The following assert should pass:

```python
import cirq
c1 = cirq.Circuit.from_ops(cirq.CZ(a, b), cirq.measure(a, b))
c2 = cirq.Circuit.from_ops(cirq.measure(a, b))
cirq.testing.assert_circuits_with_terminal_measurements_are_equivalent(c1, c2)
```

But actually it fails:

```
AssertionError: Circuit's effect differs from the reference circuit.

Diagram of actual circuit:
0: ───@───M───
      │   │
1: ───@───M───

Diagram of reference circuit with desired function:
0: ───M───
      │
1: ───M───
```

The issue is that `assert_circuits_with_terminal_measurements_are_equivalent` only canonicalizes the circuits' unitary matrices up to `Z**t` gates on the individual qubits. But actually it needs to consider arbitrary diagonal gates involving all measured qubits. This requires updating the linear constraints created by the method.

viathor:
Fixes #1019. Avoids LP and instead compares circuit matrices block-by-block up to per-block phase factor. Crucially, blocks are formed using measurement subspaces. Adds more tests.

This is a second attempt at #1036. First one was lost to [github incident on October 21st](https://blog.github.com/2018-10-21-october21-incident-report/). Includes changes to address comments in the original review.

### DIFF
--- a/cirq/testing/circuit_compare.py
+++ b/cirq/testing/circuit_compare.py
@@ -38,8 +38,7 @@ def highlight_text_differences(actual: str, expected: str) -> str:
 
 def _measurement_subspaces(
         measured_qubits: Iterable[ops.QubitId],
-        n_qubits: int,
-        qubit_order: ops.QubitOrderOrList = ops.QubitOrder.DEFAULT
+        n_qubits: int
 ) -> Sequence[Sequence[int]]:
     """Computes subspaces associated with projective measurement.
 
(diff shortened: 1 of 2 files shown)


Strilanc:
We'll have to wait for github to re-enable hooks then poke this branch to get travis to run it.

PR #1039, opened 2018-10-23
Add 'dottable' check to assert_equivalent_repr

Strilanc:
- For example, we want `{!r}.on(a)` to work, but `cirq.CCZ**0.5` doesn't have that property. You need to add parens.

### DIFF
--- a/cirq/testing/equivalent_repr_eval_test.py
+++ b/cirq/testing/equivalent_repr_eval_test.py
@@ -19,7 +19,7 @@ import cirq
 
 
 def test_external():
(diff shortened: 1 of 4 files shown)


cirq/testing/equivalent_repr_eval_test.py
```diff
@@ -19,7 +19,7 @@
 
 
 def test_external():
-    for t in [1, 2, 3, 'a', 0.5, 2**0.5, 1.1, 1j]:
+    for t in ['a', 1j]:
```
Strilanc: Dropped floats and ints because, well, they don't satisfy the property that we are holding ourselves to.

PR #1045, opened 2018-10-24
Refactor Rot[X/Y/Z/11]Gate into [CZ/X/Y/Z]PowGate

Strilanc:
- `__init__` now takes exponent instead of rads/degs/half_turns
- exposed property is exponent instead of half_turns

Part of https://github.com/quantumlib/Cirq/issues/865


cirq/ops/__init__.py
```diff
@@ -26,10 +26,10 @@
     measure,
     measure_each,
     MeasurementGate,
-    Rot11Gate,
-    RotXGate,
-    RotYGate,
-    RotZGate,
+    CZPowGate,
```
dabacon: we've been using alphabetical and this breaks that style though I agree it's nice to have these near each other

cirq/__init__.py
```diff
@@ -134,10 +134,10 @@
     QubitOrder,
     QubitOrderOrList,
     ReversibleCompositeGate,
-    Rot11Gate,
-    RotXGate,
-    RotYGate,
-    RotZGate,
+    CZPowGate,
```
dabacon: This breaks alphabetical

cirq/ops/__init__.py
Strilanc: Done.

cirq/__init__.py
Strilanc: Done.

PR #1047, opened 2018-10-24
Add XX parity gate

mlihc15:
Adding ion trap native 2-qubit gates.


mlihc15:
Signed CLA

ion/ionops_test.py
```diff
@@ -0,0 +1,43 @@
...
+# limitations under the License.
+
+import numpy as np
+import pytest
+
+import cirq
+import ion
+
+def test_CxNOTGate_init():
+    assert ion.CxNOTGate(exponent=8) == ion.CxNOTGate(exponent=0)
```
Strilanc: This is testing equality, not initialization. Just check that the `.exponent` property is the value you passed in. Then add a `test_CxNotGate_eq` method using `eq = cirq.testing.EqualsTester()` and add a few equivalence groups to `eq`.

ion/ionops_test.py
```diff
@@ -0,0 +1,43 @@
...
+    assert np.allclose(cirq.unitary(ion.CxNOTGate()),
+                       np.array([[s, 0, 0, -1j*s],
+                                 [0, s, -1j*s, 0],
+                                 [0, -1j*s, s, 0],
+                                 [-1j*s, 0, 0, s]]))
+    assert np.allclose(cirq.unitary(ion.CxNOTGate(exponent=4)),
+                       np.diag([-1, -1, -1, -1]))
+    assert np.allclose(cirq.unitary(ion.CxNOTGate()**8),
+                       np.diag([1, 1, 1, 1]))
+
```
Strilanc: Add a circuit diagram test (`c = cirq.Circuit.from_ops(cirq.CxNOT ...)` then `cirq.testing.assert_has_diagram(c, """...""")`.

ion/__init__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2018 The Cirq Developers
```
Strilanc: Move to `cirq/ion`

ion/ionops.py
```diff
@@ -0,0 +1,79 @@
...
+from typing import (
+    Union, Tuple, Optional, List, Callable, cast, Iterable, Sequence,
+)
+
+import numpy as np
+
+from cirq import value, protocols
+from cirq.ops import gate_features, eigen_gate
+
+class CxNOTGate(eigen_gate.EigenGate,
```
Strilanc: Rename to `XXGate`. I guess we should also add ZZ and YY while we're at it. Can be done after.

Strilanc:
After some conversation we realized that actually the goal is to implement an XX gate, not a CxNOT gate.

cirq/ops/common_gates.py
```diff
@@ -982,3 +982,82 @@ def Ry(rads: float) -> RotYGate:
 def Rz(rads: float) -> RotZGate:
     """Returns a gate with the matrix e^{-i Z rads / 2}."""
     return RotZGate(rads=rads, global_shift_in_half_turns=-0.5)
+
+
+class XXGate(eigen_gate.EigenGate,
```
Strilanc: Rename to `XXPowGate`

cirq/ops/common_gates.py
```diff
@@ -982,3 +982,82 @@ def Ry(rads: float) -> RotYGate:
...
+                gate_features.TwoQubitGate,
+                gate_features.InterchangeableQubitsGate):
+    """XX interaction.
+
+    The gate implements the following unitary:
+    exp(-i t X⊗X) = [cos(t) 0 0 -isin(t)]
```
Strilanc: The matrix of this gate should be `(X⊗X)^t = (-i)^t exp(-i pi t X⊗X)`. The current specified matrix is off by a global phase, and has a different scaling of `t`.

cirq/ops/common_gates.py
```diff
@@ -982,3 +982,82 @@ def Ry(rads: float) -> RotYGate:
...
+                    [0 cos(t) -isin(t) 0]
+                    [0 -isin(t) cos(t) 0]
+                    [-isin(t) 0 0 cos(t)]
+
+    """
+
+    def __init__(self, *,  # Forces keyword args.
+                 duration: Optional[float]=None,
```
Strilanc: This duration concept is sort of unique to openfermion-cirq and has no equivalent in cirq, so we should leave it out. If we want a gate specified with alternate units, we'll make a utility method analogous to `Rx`.

Is there a typical name for a function that takes radians and spits out exp(i XX t)?

cirq/ops/common_gates.py
```diff
@@ -982,3 +982,82 @@ def Ry(rads: float) -> RotYGate:
...
+                            [0, -0.5, 0.5, 0],
+                            [-0.5, 0, 0, 0.5]])),
+            (-0.25, np.array([[0.5, 0, 0, 0.5],
+                            [0, 0.5, 0.5, 0],
+                            [0, 0.5, 0.5, 0],
+                            [0.5, 0, 0, 0.5]]))
+        ]
+
+    def _canonical_exponent_period(self) -> Optional[float]:
+        return 8
```
Strilanc: The period of `f(t) = XX**t` is 2.

cirq/ops/common_gates.py
```diff
@@ -982,3 +982,82 @@ def Ry(rads: float) -> RotYGate:
...
+
+        super().__init__(exponent=power)
+
+    @property
+    def exponent(self) -> Union[value.Symbol, float]:
+        return self._exponent
+
+    def _eigen_components(self):
+        return [
+            (0.25, np.array([[0.5, 0, 0, -0.5],
```
Strilanc: Use 0 and 1 instead of 0.25 and -0.25

Strilanc:
Define the `MS(rads)` method inside of `cirq.ion`, but export it all the way to the top level.

PR #1049, opened 2018-10-24
Add _has_unitary_ magic method

dstrain115:
Add has_unitary to SupportsUnitary protocol to support cirq.has_unitary
Add _has_unitary_ to each gate / class that has _unitary_ and add tests for each of these classes.

### DIFF
--- a/cirq/protocols/unitary.py
+++ b/cirq/protocols/unitary.py
@@ -120,7 +120,4 @@ def has_unitary(val: Any) -> bool:
         return result
 
     # No _has_unitary_ function, use _unitary_ instead
-    getter = getattr(val, '_unitary_', None)
-    result = NotImplemented if getter is None else getter()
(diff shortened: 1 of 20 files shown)


cirq/protocols/unitary.py
```diff
@@ -90,3 +102,25 @@ def unitary(val: Any,
...
+        Returns False if neither function exists.
+    """
+    getter = getattr(val, '_has_unitary_', None)
+    result = NotImplemented if getter is None else getter()
+
+    if result is not NotImplemented:
+        return result
+
+    # No _has_unitary_ function, use _unitary_ instead
+    getter = getattr(val, '_unitary_', None)
```
Strilanc: Delegate to `return protocols.unitary(val, None) is not None` instead of re-implementing it.

Issue #731, opened 2018-07-19
Can we include the python 2 and python 3 artifacts in the same pypi release?


Strilanc:
As part of launching, I tested if I could put py2 and py3 artifacts into the same pypi version (e.g. just cirq 0.3.1 instead of cirq 0.3.1.27 and cirq 0.3.1.35), but my quick test indicated that pypi was pretending only the python 2 version was available when I did so. So I put them in separate versions just to be sure it would work.

But there's *no way* that's actually how it works, right? Surely there's a way to get both the py2 and py3 code into one version distribution? Did I just mess up my quick test, and including the artifacts in the same package actually just works? Does someone know offhand how this works, or a site that explains the details of doing this?

vtomole:
@Strilanc Just wondering; why is this project supporting Python 2?  I'm asking because most programs have been phasing out their Python 2 support lately.

Strilanc:
@vtomole There are two reasons:

1. A non-negligible number of people are still using it.
2. We came up with a low-cost way to support it.

The low-cost way to support it is by using a transpiler. This is very common in the javascript ecosystem, e.g. babel will take your ES6 code and transpile it down to ES5, and so it is very easy to use new features while maintaining compatibility.

When compared to babel and traceur and such, python has terrible transpilers. But 3to2 is good enough that, with a few additional tweaks, you can just write python 3 code and have it automatically work in python 2. There are still gotchas (e.g. using the `default` parameter of `max`, using non-indexed format string arguments to a custom string formatter, probably anything to do with async methods), but they are negligible compared to the costs of writing code that must run in both languages.

dabacon:
https://stackoverflow.com/questions/40114100/uploading-different-versions-python-2-7-vs-3-5-to-pypi

Think you have to do different versions?

Strilanc:
Using separate version numbers for the packages is what I did last time. I want to put them both in the same package with the same version.

... I think. My plan for this issue is to just experiment with the testing pypi when it comes near release time.

PR #1050, opened 2018-10-25
Stop using different version numbers for python 2 and python 3

Strilanc:
- Figured out why previous attempts to use the same package for both wheels didn't work (bad python_requires semantics)
- Added twine as a dev tool
- Tweaked produce-package.sh to force custom output directory
- Added publish-to-test-pypi.sh
- Updated pypi package instructions

Fixes https://github.com/quantumlib/Cirq/issues/731


Strilanc:
googlebot seems to be taking a break; manually approved CLAs

Issue #1048, opened 2018-10-24
Automatically compute the period of an EigenGate


Strilanc:
Current EigenGate requires the child class to implement a `_canonical_exponent_period` method saying what the period of the operation is. But this information can be derived from other information given to EigenGate; specifically the _global_shift_in_half_turns argument to `__init__` and the differences between the eigenvalue exponents coming out of the `_eigen_components` method.

This issue requires a) figuring out how exactly to compute the period (e.g. the least common multiple of the reciprocals of the relative phases and the global phase) and b) deleting `_canonical_exponent_period` and all usage of it with said computation.

PR #1055, opened 2018-10-25
Refactor EigenGate's period semantics

Strilanc:
- Renamed _canonical_exponent_period to _period
- _period is now computed automatically in most relevant cases, instead of always requiring child classes to specify it
- _exponent is no longer canonicalized, so that `(X**-1)**0.5` gives `X**-0.5`
- Added _canonical_exponent property, which is used for equality checks
- Renamed global_shift_in_half_turns to global_negate_rate
- Fix cirq.PauliInteractionGate's repr; update its exponent behavior

Fixes https://github.com/quantumlib/Cirq/issues/1048
Fixes https://github.com/quantumlib/Cirq/issues/1076


Strilanc:
I think the general consensus is that "half_turns" is a confusing unit. So I'm specifically avoiding it. I'm not really sure what to call this because it's not a familiar concept. I call it global negate rate because it is the t in (-1)^(t * exponent). If you can think of a better name I'm happy to change it.

Strilanc:
Perhaps global_phase_factor?

Strilanc:
I've renamed it to `global_shift` and documented it as shifting the eigenvalues at exponent=1

cirq/ops/eigen_gate.py
```diff
@@ -227,3 +234,45 @@ def _is_parameterized_(self) -> bool:
...
+
+def _common_rational_period(rational_periods: List[fractions.Fraction]
+                            ) -> fractions.Fraction:
+    """Finds the least common integer multiple of some fractions.
+
+    The solution is the smallest positive integer c such that there
+    exists integers n_k satisfying p_k * n_k = c for all k.
+    """
+    assert rational_periods, "no well-defined solution for an empty list"
+    common_denom = _lcm(p.denominator for p in rational_periods)
```
dabacon: can you use numpy's lcm?

cirq/ops/eigen_gate.py
```diff
@@ -227,3 +234,45 @@ def _is_parameterized_(self) -> bool:
...
+
+def _approximate_common_period(periods: List[float],
+                               initial_denom: int = 60,
+                               reject_atol: float = 1e-8) -> Optional[float]:
+    if not periods:
+        return None
+    if any(e == 0 for e in periods):
+        return None
+    approx_rational_periods = [
+        fractions.Fraction(int(np.round(p * initial_denom)), initial_denom)
```
dabacon: I think you can use numpy's rint instead of int(np.round(.))

cirq/ops/eigen_gate.py
```diff
@@ -227,3 +234,45 @@ def _is_parameterized_(self) -> bool:
...
+    t = 1
+    for r in vals:
+        t = t * r // fractions.gcd(t, r)
+    return t
+
+
+def _approximate_common_period(periods: List[float],
+                               initial_denom: int = 60,
+                               reject_atol: float = 1e-8) -> Optional[float]:
+    if not periods:
```
dabacon: doc these args.  initial_denom and reject_atol are currently not publicly accessible, I guess they are being used for testing?

cirq/ops/eigen_gate.py
```diff
@@ -165,28 +160,40 @@ def _eigen_components(self) -> List[Union[EigenComponent,
...
 
-    # virtual method
-    def _canonical_exponent_period(self) -> Optional[float]:
-        """Determines how the exponent parameter is canonicalized.
+    def _period(self) -> Optional[float]:
+        """Determines how the exponent parameter is canonicalized when equating.
 
         Returns:
             None if the exponent should not be canonicalized. Otherwise a float
             indicating the period of the exponent. If the period is p, then a
```
dabacon: a bit strange that this returns a float, when it only every returns an integer period.

Strilanc: It returns non integer in some cases, there are tests that trigger this behavior.

cirq/ops/eigen_gate.py
Strilanc: I wish I could but we support a version of numpy that doesn't have that function. Also in Python 2 there is no LCM function under the math module.

Strilanc:
Rint still returns a float. It looks equivalent to np.round to me.

cirq/ops/eigen_gate.py
Strilanc: Done.

PR #1059, opened 2018-10-25
Add method for taking channel and returning probabilities and unitaries represent channel if possible

dabacon:
There is no general way to tell if a channel can be interpreted as a stochastic set of probabilities (under the unitary degree of freedom of the Krauss operators), but this does this decomposition when it is possible.

This method means that we don't have to have noise models that declare they are stochastic unitary maps, which will be useful for the stochastic noise simulator.

### DIFF
--- a/cirq/protocols/channel.py
+++ b/cirq/protocols/channel.py
@@ -14,7 +14,7 @@
 
 """Protocol and methods for quantum channels."""
 
-from typing import Any, Iterable, Tuple, TypeVar, Union
+from typing import Any, Iterable, Optional, Tuple, TypeVar, Union
 
 import numpy as np
 from typing_extensions import Protocol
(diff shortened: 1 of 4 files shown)


cirq/protocols/channel.py
```diff
@@ -122,3 +123,46 @@ def channel(val: Any,
...
+
+def stochastic_unitary(
+    val: SupportsChannel,
+    rtol=1.e-5,
+    atol=1.e-8, ) -> Optional[Tuple[Tuple[float, np.ndarray], ...]]:
+    """Gives a stochastic representation of a channel by unitaries, if possible.
+
+    If a channel has Krauss operators A_0,A_1,...,A_{r-1} this checks whether
+    each A_i is proportional to a unitary matrix, U_i, and if so, returns
+    each each unitary U_i along with the probability of the unitary p_i,
```
kevinsung: "each" not "each each"

cirq/protocols/channel.py
```diff
@@ -122,3 +123,46 @@ def channel(val: Any,
                         "_unitary_ method.".format(type(val)))
     raise TypeError("object of type '{}' does have a _channel_  or _unitary_ "
                     "method, but it returned NotImplemented.".format(type(val)))
+
+
+def stochastic_unitary(
+    val: SupportsChannel,
+    rtol=1.e-5,
+    atol=1.e-8, ) -> Optional[Tuple[Tuple[float, np.ndarray], ...]]:
```
Strilanc: Nit: comma space without new line look sodd

cirq/protocols/channel.py
```diff
@@ -122,3 +123,46 @@ def channel(val: Any,
                         "_unitary_ method.".format(type(val)))
     raise TypeError("object of type '{}' does have a _channel_  or _unitary_ "
                     "method, but it returned NotImplemented.".format(type(val)))
+
+
+def stochastic_unitary(
```
Strilanc: I think we should delegate this task to the gates. Gates will now if they are equivalent to a stochastic unitary, and they can tell us instead of us trying to figure out it after forcing the gate explain it in a different language.

Basically, we should have a `_stochastic_unitary_` magic method.

dabacon: That's not DRY.  By having this method this means that you only need to implement channel.  In what you propose you would have to write essentially the same operators twice. We can avoid this by just a small amount of math.

dabacon: That's very unDRY.  If you have this magic method you would essentially repeat the code for the channel multiple times with subtle differences.  This means writing essentially the same code twice, and worse it can lead to inconsistencies.  By using this you never run into that problem at the cost of a small, albeit very standard, bit of linear algebra.

cirq/protocols/channel.py
dabacon: fixed

cirq/protocols/channel.py
Strilanc: Oh, you would avoid repetition by modifying `cirq.channel` to fallback onto `_stochastic_unitary_` and then onto `_unitary_`.

That being said, repetition can be justified by performance benefits and this could fall under that umbrella.

dabacon: My first thought is that i think that would be confusing for people.  Knowing that you could implement _stochastic_unitary_ would be something you would have to discover.  (And I bet most people would implement both _channel_ and _stochastic_unitary_).  I guess stochastic_unitary could fall back to attempting to decompose the channel.  Right now the only plans for this are that the simulator knows about it.  I do agree that performance might be worth thinking about.  The other problem is that there are other channel classes beyond stochastic unitary (unital for example) and I wonder where we draw the line.  

OK let me think about this a bit.

kkourin:
This is a nit but I think the more common term for these is _mixed_ unitary channels.

dabacon:
Closing this for now, though a similar method may be useful as a utility at some point.

dabacon:
Oh and thanks kkourin, yes "mixed" is the right word.

PR #1064, opened 2018-10-25
Add global_shift to PhasedXGate

Strilanc:


cirq/ops/phased_x_gate.py
```diff
@@ -135,13 +162,36 @@ def __str__(self):
...
+
+    def _period(self):
+        exponents = [self._global_shift, 1 + self._global_shift]
+        real_periods = [abs(2/e) for e in exponents if e != 0]
+        int_periods = [int(np.round(e)) for e in real_periods]
+        if any(i != r for i, r in zip(real_periods, int_periods)):
+            return None
+        if len(int_periods) == 1:
+            return int_periods[0]
+        return int_periods[0] * int_periods[1] / fractions.gcd(*int_periods)
```
dabacon: fractions.gcd is deprecated, I think you can use math.gcd?

Strilanc: `math.gcd` doesn't exist in python 2, so we're stuck with `fractions.gcd` for now. We could consider modifying the transpilation script to handle this case.

PR #1066, opened 2018-10-25
Mark X, Y, and PhasedX gates as Xmon natives

Strilanc:


cirq/google/programs.py
```diff
@@ -28,25 +28,78 @@
...
         return _cz_to_proto_dict(gate, *qubits)
 
     raise ValueError("Don't know how to serialize this gate: {!r}".format(gate))
 
 
+def _x_to_proto_dict(gate: ops.XPowGate, q: ops.QubitId) -> Dict:
+    exp_w = {
+        'target': cast(devices.GridQubit, q).to_proto_dict(),
+        'axis_half_turns':
+            xmon_gates.XmonGate.parameterized_value_to_proto_dict(0),
```
dabacon: Move parameterized_value_to_proto method to the module level

Strilanc: I don't think we should do that quite yet. In particular, I think we want one single "give me a proto" method instead of a family of them. Perhaps a magic method protocol...

PR #1067, opened 2018-10-27
Added More Noise

### LINKED ISSUE #928: Library of useful noise models
After #925  we will need actual implementations of some useful channels

- depolarizing
- amplitude damping
- dephasing
- over or under rotation gates

MichaelBroughton:
fixes #928 I added:
- Generalized Amplitude Damping
- Amplitude Damping
- Phase Damping
- Phase Flip
- Bit Flip
- Rotation Error

To go along with the existing Depolarizing


MichaelBroughton:
I signed it ?

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,528 @@ def depolarize(p: float) -> DepolarizingChannel:
         ValueError if p is not a valid probability.
     """
     return DepolarizingChannel(p)
+
+
+class GeneralizedAmplitudeDampingChannel(raw_types.Gate):
+    """ The generalized amplitude damping channel."""
```
Strilanc: Don't just repeat the name, define what the operation actually does (i.e. move most of the content from `__init__` to here).

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,528 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+            M_1 = \sqrt{p} \begin{bmatrix} 0 & \sqrt{\gamma} \\
+                            0 & 0 \end{bmatrix}
+            M_2 = \sqrt{1-p} \begin{bmatrix} \sqrt{1-\gamma} & 0 \\
+                            0 & 1 \end{bmatrix}
+            M_3 = \sqrt{1-p} \begin{bmatrix} 0 & 0 \\
+                            \sqrt{gamma} & 0 \end{bmatrix}
+
+        Args:
+            gamma: The damping rate
+            p: the probability of
```
Strilanc: of...?

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,528 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+
+
+class GeneralizedAmplitudeDampingChannel(raw_types.Gate):
+    """ The generalized amplitude damping channel."""
+
+    def __init__(self, p, gamma) -> None:
+
+        """
+        This channel models the effect of dissipation to the environment
+        at possibly non zero tempurature. Where gamma is the damping rate
```
Strilanc: You don't need to define gamma and p here, since they're listed in the args.

Strilanc: typo temperature

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,528 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+        and p is the probability of
+
+        This channel evovles a density matrix via
```
Strilanc: typo evolves

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,528 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+            \rho -> M_0 \rho M_0^\dagger + M_1 \rho M_1^\dagger
+                  + M_2 \rho M_2^\dagger + M_3 \rho M_3^\dagger
+
+        With
+            M_0 = \sqrt{p} \begin{bmatrix} 1 & 0  \\
```
Strilanc: Horizontally align the text like so:

```
\begin{bmatrix}
   1 & 0 \\
    0 & ...
```

so that it can be read by eye easily.

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,528 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+            elif p > 1:
+                raise ValueError('{} was greater than 1.'.format(p_str))
+            return p
+
+        self._gamma = validate_probability(gamma, 'gamma')
+        self._p = validate_probability(p, 'p')
+
+    def _channel_(self) -> Iterable[np.ndarray]:
+        return (
+            np.sqrt(self._p)
```
Strilanc: Precompute the `np.sqrt`s to avoid redoing work, and to make the matrices easier to read. E.g. `p0 = np.sqrt(self._p)`, `p1 = np.sqrt(1 - self._p)`.

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,528 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+        p: the probability of
+
+    Raises:
+        ValueError gamma or p is not a valid probability.
+    """
+    return GeneralizedAmplitudeDampingChannel(p, gamma)
+
+
+class AmplitudeDampingChannel(raw_types.Gate):
+    """ The amplitude damping channel."""
```
Strilanc: Nit: no space after the `"""`.

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,528 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+        gamma: the damping rate
+
+    Raises:
+        ValueError is gamma is not a valid probability.
+    """
+    return AmplitudeDampingChannel(gamma)
+
+
+class PhaseDampingChannel(raw_types.Gate):
+    """Phase Damping of a qubit"""
```
Strilanc: This is the dephasing channel?

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,528 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+    def __str__(self) -> str:
+        return 'phase_damping(gamma={!r})'.format(self._gamma)
+
+    def _circuit_diagram_info_(
+        self, args: protocols.CircuitDiagramInfoArgs
+    ) -> str:
+        return 'PD({!r})'.format(self._gamma)
+
+
+def phase_damping(gamma: float) -> PhaseDampingChannel:
```
Strilanc: Rename to `phase_damp`, similar to how we have `cirq.measure` instead of `cirq.measuring`. Method names should be verbs (usually).

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,528 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+    def __str__(self) -> str:
+        return 'phase_flip(p={!r})'.format(self._p)
+
+    def _circuit_diagram_info_(
+        self, args: protocols.CircuitDiagramInfoArgs
+    ) -> str:
+        return 'PF({!r})'.format(self._p)
+
+
+def phase_flip(p: float) -> PhaseFlipChannel:
```
Strilanc: I would have assumed that this was the unitary Z gate, and been very surprised about there being noisy-ness.

Perhaps what we could do is have `p` default to 1, actually return a Z gate if p equals 1, and rename `p` to `probability_of_apply` or something descriptive like that. The method's return type would change to `Union[ops.ZPowGate, PhaseFlipChannel]` and you would also declare an overload that returned `ops.ZPowGate` when there is no  `probability_of_apply` argument.

A probably-even-better alternative would be to encapsulate this entire paradigm into a "ProbableOperation" that you could wrap around any operation (whether unitary or channel). We could then add a `with_probability(p)` method to `cirq.Operation` and `cirq.Gate` to make these easy to make, so you could say `cirq.Z(q).with_probability(p)`.

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,528 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+    def __str__(self) -> str:
+        return 'bit_flip(p={!r})'.format(self._p)
+
+    def _circuit_diagram_info_(
+        self, args: protocols.CircuitDiagramInfoArgs
+    ) -> str:
+        return 'BF({!r})'.format(self._p)
+
+
+def bit_flip(p: float) -> BitFlipChannel:
```
Strilanc: Same comments as for the phase flip method.

cirq/ops/common_channels.py
MichaelBroughton: fixed

cirq/ops/common_channels.py
MichaelBroughton: fixed

cirq/ops/common_channels.py
MichaelBroughton: No, if you were going to dephase something, you would use `phase_flip`

cirq/ops/common_channels.py
MichaelBroughton: fixed

cirq/ops/common_channels.py
MichaelBroughton: Think i've got it fixed ?

cirq/ops/common_channels.py
MichaelBroughton: fixed

cirq/ops/common_channels.py
MichaelBroughton: think I've fixed everywhere ?

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,605 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+            return p
+
+        self._gamma = validate_probability(gamma, 'gamma')
+        self._p = validate_probability(p, 'p')
+
+    def _channel_(self) -> Iterable[np.ndarray]:
+        p0 = np.sqrt(self._p)
+        p1 = np.sqrt(1.0 - self._p)
+        return (
+            p0 * np.array([[1.0, 0.0], [0.0, np.sqrt(1.0 - self._gamma)]]),
```
Strilanc: Also precompute the gamme square roots..

I think you can drop the `.0`s, because a mix of ints and floats becomes dtype float.

MichaelBroughton: fixed. I dropped the .0's and precomputed square roots.

cirq/ops/common_channels.py
Strilanc: It doesn't look fixed...? Did you push?

MichaelBroughton: fixed

cirq/ops/common_channels.py
dabacon: I agree we need a better structure here but am not sure that this is the correct pull request to do that?  Righting these in the first place is an exercise to see what the structure is that we need.  The name is not horrible as a first take, but see where there could be confusion.

MichaelBroughton: I feel like this is something that should be tackled in separate PR. This is because some of the models in here have "probabilities", but others also have "rates", so the correct design of `with_probability` might not be so straightforward.

Strilanc:
I can't review this until you push your changes. There are still several "fixed" comments on code that is not actually fixed.

MichaelBroughton:
I'm Sorry, I can't seem to track down what you are referring to.
In terms of the unresolved conversations I see:

- p is the probability of... ? -> Which is fixed, I filled in what all the probabilities are
- typo in temperature and evolves -> which is fixed I can't find either of those kinds of typos anymore
- the latexing of the matrices is now indented according to your suggestions
- I've precomputed the sqrts which is in the code now
- I've removed the extra space in the comment
- I've renamed `phase_damping` to `phase_damp` same with `amplitude_damp`

I verified all of these by looking at the Files Changed tab (So I'm pretty sure the changes are pushed ?).

finally I thought the `with_probability` would be tackled in a later PR.

Could you help me understand what I'm still missing ?

dabacon:
I too only see the unresolved issue the question of bit_flip and phase_flip.  I think this also happened on my pull request recently, I assume Craig is looking at file changed tab and not the "outdated" inline.

Strilanc:
I'm really not sure what was going on. I did hit the 'files changed' tab to make sure I was looking at the up to date stuff... and there wasn't any "see latest" messages IIRC. Github is not so great at keeping this all straight for users.

cirq/ops/common_channels.py
Strilanc: Okay. Open an issue (you don't have to take it, just make it) to make a generalized unitary-to-probabilistic-unitary method and that is sufficient for now.

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,603 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+    def __str__(self) -> str:
+        return 'phase_flip(p={!r})'.format(self._p)
+
+    def _circuit_diagram_info_(
+        self, args: protocols.CircuitDiagramInfoArgs
+    ) -> str:
+        return 'PF({!r})'.format(self._p)
+
+
+def phase_flip(p: float) -> PhaseFlipChannel:
```
Strilanc: Have `p` default to `None`, change return type to `Union[ops.ZPowGate, PhaseFlipChannel]`, declare an overload with no argument returning `ops.ZPowGate` and an overload with the non-optional argument returning `PhaseFlipChannel`. If `p is None` then return `cirq.Z`.

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,603 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+    def __str__(self) -> str:
+        return 'bit_flip(p={!r})'.format(self._p)
+
+    def _circuit_diagram_info_(
+        self, args: protocols.CircuitDiagramInfoArgs
+    ) -> str:
+        return 'BF({!r})'.format(self._p)
+
+
+def bit_flip(p: float) -> BitFlipChannel:
```
Strilanc: Have `p` default to `None`, change return type to `Union[ops.XPowGate, BitFlipChannel]`, declare an overload with no argument returning `ops.XPowGate` and an overload with the non-optional argument returning `BitFlipChannel`. If `p is None` then return `cirq.X`.

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,603 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+        gamma: The damping rate
+
+    Raises:
+        ValueError is gamma is not a valid probability.
+    """
+    return PhaseDampingChannel(gamma)
+
+
+class PhaseFlipChannel(raw_types.Gate):
+    """
```
Strilanc: Style guide is to put a short one-line summary on the same line as the `"""`.

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,603 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+    Flips the sign of the phase of a qubit with a given probability.
+    """
+
+    def __init__(self, p) -> None:
+
+        """
```
Strilanc: Style guide is to put a short one-line summary on the same line as the `"""` before moving on to the more complicated body.

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,603 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+    return PhaseDampingChannel(gamma)
+
+
+class PhaseFlipChannel(raw_types.Gate):
+    """
+    Flips the sign of the phase of a qubit with a given probability.
+    """
+
+    def __init__(self, p) -> None:
+
```
Strilanc: Nit: extra new line

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,603 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+            M_1 = \sqrt{1-p} \begin{bmatrix}
+                                1 & 0 \\
+                                0 & -1
+                            \end{bmatrix}
+
+        Args:
+            p: the probability of a phase flip
+
+        Raises:
+            ValueError if p is not a valid probability.
```
Strilanc: `ValueError: ` so that it parses correctly in tools. The colon is part of the syntax.

cirq/ops/common_channels.py
```diff
@@ -187,3 +189,603 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+
+    def _eq_tuple(self):
+        return (BitFlipChannel, self._p)
+
+    def __eq__(self, other):
+        if not isinstance(other, type(self)):
+            return NotImplemented
+        return self._eq_tuple() == other._eq_tuple()
+
+    def __ne__(self, other):
```
Strilanc: Add a `__hash__` method to each of these types (just return `hash(self._eq_tuple())`).

cirq/ops/common_channels.py
MichaelBroughton: For this issue and the one above I wasn't exactly sure how you wanted things overloaded, so is how I've done it somewhat close to what you are after ?

cirq/ops/common_channels.py
MichaelBroughton: fixed this in most places I believe.

cirq/ops/common_channels.py
MichaelBroughton: fixed.

cirq/ops/common_channels.py
MichaelBroughton: fixed.

cirq/ops/common_channels.py
MichaelBroughton: fixed

cirq/ops/common_channels.py
MichaelBroughton: added

cirq/ops/common_channels.py
```diff
@@ -184,6 +195,700 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+    """
+
+    def __init__(self, p, gamma) -> None:
+        """The generalized amplitude damping channel.
+
+        Construct a channel to model the energy dissipation to the
+        environment at rate gamma, occuring with probability p
+
+        This channel evolves a density matrix via
+            \rho -> M_0 \rho M_0^\dagger + M_1 \rho M_1^\dagger
```
dabacon: maybe better to just write this as 
\rho -> \sum_{i=0}^3 M_i \rho M_i^\dagger
?  Not sure either way is fine.

cirq/ops/common_channels.py
```diff
@@ -184,6 +195,700 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+    This channel models the effect of energy dissipation to the
+    surrounding environment at some particular rate, with some
+    given probability at a non zero temperature.
+    """
+
+    def __init__(self, p, gamma) -> None:
+        """The generalized amplitude damping channel.
+
+        Construct a channel to model the energy dissipation to the
+        environment at rate gamma, occuring with probability p
```
dabacon: nit: period at end of comments (here and in args)

cirq/ops/common_channels.py
```diff
@@ -184,6 +195,700 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+        ValueError: if p is not a valid probability.
     """
     return DepolarizingChannel(p)
+
+
+class GeneralizedAmplitudeDampingChannel(raw_types.Gate):
+    """Dampen qubit amplitudes through non ideal dissipation.
+
+    This channel models the effect of energy dissipation to the
+    surrounding environment at some particular rate, with some
```
dabacon: I think the main point about the Generalized amplitude damping channel is that it models both dissipation into the environment, and also the environment causing the system to get excited.  The reference to "temperature" is sort of reference to this later, but it's a bit confusing.  Would suggest:

This channel models the effect of energy dissipation into an environment as well 
as the environment depositing energy into the system.  The probabilities with which
the energy exchange occur are give by one constant, gamma, and the
probability of the environment being not excited is given by another constant,
p.

The stationary state of this channel is the diagonal density matrix with probability
p of being |0> and probability 1-p of being |1>.

cirq/ops/common_channels.py
```diff
@@ -184,6 +195,700 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+                                \sqrt{1-\gamma} & 0 \\
+                                 0 & 1
+                              \end{bmatrix}
+            M_3 = \sqrt{1-p} \begin{bmatrix}
+                                 0 & 0 \\
+                                 \sqrt{gamma} & 0
+                             \end{bmatrix}
+
+        Args:
+            gamma: The damping rate
```
dabacon: I would call
p: the probability of the interaction being dissipative
gamma: the probability of the qubit and environment exchanging energy

cirq/ops/common_channels.py
```diff
@@ -184,6 +195,700 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+
+
+def generalized_amplitude_damp(
+    p: float, gamma: float
+) -> GeneralizedAmplitudeDampingChannel:
+    """
+    Returns a GeneralizedAmplitudeDampingChannel with the given damping rate
+    occuring with the given probability
+
+    This channel evolves a density matrix via
```
dabacon: this doc doesn't match the definition (this is amplitude_damp)

cirq/ops/common_channels.py
```diff
@@ -184,6 +195,700 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+
+    This channel models phase damping which is the loss of quantum
+    information without the loss of energy.
+    """
+
+    def __init__(self, gamma) -> None:
+        """The phase damping channel.
+
+        Construct a channel that enacts phase damping at the rate
+        given by gamma
```
dabacon: I wouldn't call this a rate.  Perhaps just say 

 Construct a channel that enacts phase damping constant gamma.

(If you don't care about the measurement info this is really a probabilistic unitary with a probability of flip given by (1+sqrt(1-gamma)) / 2

cirq/ops/common_channels.py
```diff
@@ -184,6 +195,700 @@ def depolarize(p: float) -> DepolarizingChannel:
...
+    Raises:
+        ValueError: if p is not a valid probability.
+    """
+    if p is None:
+        return _bit_flip_X()
+
+    return _bit_flip(p)
+
+
+class RotationErrorChannel(raw_types.Gate):
```
dabacon: I think we can design a more general version of this, but I'm fine with this being the starting point.

cirq/ops/__init__.py
```diff
@@ -23,7 +23,20 @@
     asymmetric_depolarize,
     AsymmetricDepolarizingChannel,
     depolarize,
-    DepolarizingChannel
+    DepolarizingChannel,
+    generalized_amplitude_damp,
+    GeneralizedAmplitudeDampingChannel,
+    amplitude_damp,
```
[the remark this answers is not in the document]
MichaelBroughton: fixed.

cirq/ops/__init__.py
```diff
@@ -23,7 +23,20 @@
...
+    AmplitudeDampingChannel,
+    phase_damp,
+    PhaseDampingChannel,
+    phase_flip,
+    PhaseFlipChannel,
+    bit_flip,
+    BitFlipChannel,
+    rotation_error,
+    RotationErrorChannel
+
```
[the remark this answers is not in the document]
MichaelBroughton: fixed

cirq/ops/common_channels.py
```diff
@@ -14,13 +14,13 @@
 
 """Quantum channels that are commonly used in the literature."""
 
-from typing import Iterable
+from typing import Iterable, Union, Optional
```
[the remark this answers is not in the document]
MichaelBroughton: fixed

cirq/ops/common_channels.py
```diff
@@ -14,13 +14,13 @@
...
 
 import numpy as np
 
 from cirq import protocols
 from cirq.ops import raw_types
-
+from cirq.ops.common_gates import ZPowGate, XPowGate
```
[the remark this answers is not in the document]
MichaelBroughton: fixed

cirq/ops/common_channels.py
MichaelBroughton: I kept it the way it it is , since this new form is pretty similar to what is already in protocols.

cirq/ops/common_channels.py
MichaelBroughton: fixed, added periods everywhere I could.

cirq/ops/common_channels.py
MichaelBroughton: fixed, update to the docs to almost exactly what you gave here

cirq/ops/common_channels.py
MichaelBroughton: fixed

cirq/ops/common_channels.py
MichaelBroughton: fixed.

cirq/ops/common_channels.py
MichaelBroughton: fixed, changed to be exactly "Construct a channel that enacts phase damping constant gamma."

cirq/ops/common_channels.py
MichaelBroughton: sounds good to me haha

Strilanc:
Interesting. That's definitely a bug in the merge script.

Issue #1061, opened 2018-10-25
Replace usages of `cirq.decompose_once` with `cirq.decompose`


Strilanc:
There is some leftover code throughout cirq that manually performs the recursion/iteration commonly associated with decomposition, instead of delegating this task to `cirq.decompose`. Such code should be updated.

This applies to almost every instance of `cirq.decompose_once` in the code base that's not in tests or the documentation.

For example, `cirq/circuits.expand_composite.py` uses `protocols.decompose_once` (the internal name for `cirq.decompose_once`) in a recursive way that should instead be written using `protocols.decompose`.

This issue is a bit advanced for a "good first issue", in that it requires carefully reasoning about recursion in the context of existing code. But it doesn't require much knowledge of how cirq works other than these specific methods and the code in the individual files so I've marked it such.

Strilanc:
Based on the errors, you probably forget to set `allow_partial_czs=False` somewhere. If you create a PR I can spot where the mistake is.

smitsanghavi:
I think this can be closed now. I don't see any non-test non-documentation uses of decompose_once anymore. 
Or is this issue supposed to cover decompose_once_with_qubits too?

Strilanc:
You're right, I think this is done. Good work!

PR #1069, opened 2018-10-28
Replace decompose_once with decompose in expand_composite

smitsanghavi:
Part of #1061


smitsanghavi:
Ignore the slew of commits. My origin master was corrupted by one of my old changes and I fumbled a bit to fix that.

cirq/circuits/expand_composite.py
```diff
@@ -54,12 +54,4 @@ def optimization_at(self, circuit, index, op):
...
-        """Recursively decompose composite gates into an OP_TREE of gates."""
-        skip = self.no_decomp(op)
-        if skip and (skip is not NotImplemented):
-            return op
-
-        decomposed = protocols.decompose_once(op, None)
-        if decomposed is None:
-            return op
-        return (self._decompose(op) for op in decomposed)
+        return protocols.decompose(op, keep=self.no_decomp, on_stuck_raise=None)
```
Strilanc: This function is so simple now that it should be inlined into the caller and then deleted.

Strilanc:
Looks good. One tweak (inlining the function) and we're good to go.

PR #1074, opened 2018-10-29
Replace decompose_once with decompose in qasm_output.py

smitsanghavi:
#1061

### DIFF
--- a/cirq/circuits/qasm_output.py
+++ b/cirq/circuits/qasm_output.py
@@ -249,52 +249,43 @@ class QasmOutput:
                           output: Callable[[str], None],
                           output_line_gap: Callable[[int], None],
                           top=True) -> None:
-        for op in ops.flatten_op_tree(op_tree):
-            out = protocols.qasm(op, args=self.args, default=None)
-            if out is not None:
-                output(out)
-                continue
+        for op_original in ops.flatten_op_tree(op_tree):
+            keep = lambda _op: (protocols.qasm(_op, args=self.args,
+                                               default=None)
+                                is not None)
+            for op in protocols.decompose(op_original, keep=keep, on_stuck_raise=None):
+                out = protocols.qasm(op, args=self.args, default=None)
+                if out is not None:
+                    output(out)
+                    continue
 
-            if isinstance(op, ops.GateOperation):
-                comment = 'Gate: {!s}'.format(op.gate)
-            else:
-                comment = 'Operation: {!s}'.format(op)
+                if isinstance(op, ops.GateOperation):
+                    comment = 'Gate: {!s}'.format(op.gate)
+                else:
+                    comment = 'Operation: {!s}'.format(op)
 
-            decomp = protocols.decompose_once(op, None)
-            if decomp is not None:
-                if top:
-                    output_line_gap(1)
-                    output('// {}\n'.format(comment))
-                self._write_operations(decomp,
-                                       output,
-                                       output_line_gap,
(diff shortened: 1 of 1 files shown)


cirq/circuits/qasm_output.py
```diff
@@ -249,52 +249,43 @@ def _write_operations(self,
...
+                else:
+                    comment = 'Operation: {!s}'.format(op)
+
+                mat = protocols.unitary(op, None) if len(op.qubits) <= 2 else None
+                if mat is not None and len(op.qubits) == 1:
+                    u_op = QasmUGate.from_matrix(mat).on(*op.qubits)
+                    if top:
+                        output_line_gap(1)
+                        output('// {}\n'.format(comment))
+                    output(cast(str, protocols.qasm(u_op, args=self.args)))
```
Strilanc: I believe the `cast` to `str` is no longer needed.

cirq/circuits/qasm_output.py
```diff
@@ -249,52 +249,43 @@ def _write_operations(self,
...
-                if top:
-                    output_line_gap(1)
-                continue
-
-            raise ValueError('Cannot output operation as QASM: {!r}'.format(op))
+        for op_original in ops.flatten_op_tree(op_tree):
+            keep = lambda _op: (protocols.qasm(_op, args=self.args,
+                                               default=None)
+                                is not None)
+            for op in protocols.decompose(op_original, keep=keep, on_stuck_raise=None):
```
Strilanc: Okay, basically, the issue you're running into is that the code is losing track of whether or not the initial operations are being decomposed or not and so the logic around commenting is going wrong.

To fix the issue you need to move this `out = qasm(...) if out: output(out); continue` block to the start of the outer loop, just before the current assignment of `keep`.

cirq/circuits/qasm_output.py
```diff
@@ -249,52 +249,43 @@ def _write_operations(self,
...
-                                       output,
-                                       output_line_gap,
-                                       top=False)
-                if top:
-                    output_line_gap(1)
-                continue
-
-            raise ValueError('Cannot output operation as QASM: {!r}'.format(op))
+        for op_original in ops.flatten_op_tree(op_tree):
+            keep = lambda _op: (protocols.qasm(_op, args=self.args,
```
Strilanc: Move this out of the loop, and `def` it as a local function instead of using lambda notation.

cirq/circuits/qasm_output.py
```diff
@@ -249,52 +249,43 @@ def _write_operations(self,
                           output: Callable[[str], None],
                           output_line_gap: Callable[[int], None],
                           top=True) -> None:
```
Strilanc: Rename `top` to `comment_matrix_fallbacks`

cirq/circuits/qasm_output.py
smitsanghavi: Moving that block up didn't make it any better. Getting slightly better results with setting comment_matrix_fallbacks to op==op_original (or just setting to true only if it's the first element in the inside loop). But I'm sure these are not the correct solutions.
https://pastebin.com/Fc6X74ii is the new diff

Strilanc:
I figured out the issue. It was quite tricky to get the cases exactly right, and to make it clean I even had to add a new parameter to `decompose`. I'm not surprised at all that you had trouble. Thanks for getting it started.

I can't push to your branch, so I opened #1078 obsoleting this PR.

PR #1078, opened 2018-10-29
Replace decompose_once with decompose in qasm_output.py

Strilanc:
- Add `fallback_decomposer` to `cirq.decompose`

Extends work by @smitsanghavi 
Obsoletes https://github.com/quantumlib/Cirq/pull/1074
Part of #1061


cirq/protocols/decompose.py
```diff
@@ -115,6 +102,10 @@ def decompose(
                                       Union[None,
                                             NotImplementedType,
                                             'cirq.OP_TREE']] = None,
+    fallback_decomposer: Callable[['cirq.Operation'],
```
dabacon: Think we should define a custom type that has this signature since this is used in multiple places

cirq/protocols/decompose.py
```diff
@@ -198,11 +200,18 @@ def decompose(
             "not possible to get stuck if you don't have a criteria on what's "
             "acceptable to keep.")
 
-    if intercepting_decomposer is None:
-        decomposer = _default_decomposer
-    else:
-        decomposers = [intercepting_decomposer, _default_decomposer]
-        decomposer = lambda op: _intercept_decompose(op, decomposers)
+    decomposers = [d
```
dabacon: boy it would be nice to use a filter here instead of this funky looking list comprehension, but I guess that won't work in 2.7

Strilanc: `functools` exists in both, I think. But eh, people like list comprehensions more than `map`/`filter` for whatever reason.

cirq/protocols/decompose.py
Strilanc: Acknowledged. For now I think it's okay to just say it so that people don't have to deref when reading the docs.

PR #1081, opened 2018-10-30
Replace decompose_once with decompose in convert_to_cz_and_single_gat…

smitsanghavi:
…es.py and convert_to_clifford_gates.py

#1061


cirq/circuits/convert_to_cz_and_single_gates.py
```diff
@@ -50,47 +50,39 @@ def __init__(self,
...
-        raise TypeError("Don't know how to work with {!r}. "
-                        "It isn't composite or an operation with a "
-                        "known unitary effect on 1 or 2 qubits.".format(op))
-
     def convert(self, op: ops.Operation) -> ops.OP_TREE:
-        converted = self._convert_one(op)
-        if converted is op:
-            return converted
-        return [self.convert(e) for e in ops.flatten_op_tree(converted)]
+        def keep(op: ops.Operation) -> bool:
```
Strilanc: Make this a private method on the class.

cirq/circuits/convert_to_cz_and_single_gates.py
```diff
@@ -50,47 +50,39 @@ def __init__(self,
...
+
+        def convert_one(op: ops.Operation) -> ops.OP_TREE:
+            # Known matrix?
+            mat = protocols.unitary(op, None)
+            if mat is not None and len(op.qubits) == 2:
+                return decompositions.two_qubit_matrix_to_operations(
+                    op.qubits[0],
+                    op.qubits[1],
+                    mat,
+                    allow_partial_czs=False)
```
Strilanc: Forward `self.allow_partial_czs` here, instead of unconditionally `False`. Any tests that disagree with this change are wrong and should be updated. If no tests catch it, add a test that a `cirq.TwoQubitMatrixGate` with the same matrix as `CZ**0.5` produces a circuit containing `CZ**0.5`.

cirq/circuits/convert_to_cz_and_single_gates.py
```diff
@@ -50,47 +50,39 @@ def __init__(self,
...
+            return NotImplemented
+
+        def on_stuck_raise(op: ops.Operation):
```
Strilanc: Rename `op` to `bad_op` to avoid shadowing the `op` in the outer method.

cirq/circuits/convert_to_cz_and_single_gates.py
```diff
@@ -50,47 +50,39 @@ def __init__(self,
...
+                    (isinstance(op, ops.GateOperation)
+                     and isinstance(op.gate, ops.CZPowGate)
+                     and (self.allow_partial_czs or op.gate.exponent == 1))
+                    # Measurement?
+                    or ops.MeasurementGate.is_measurement(op)
+                    # SingleQubit known matrix
+                    or (protocols.unitary(op, None) is not None
+                        and len(op.qubits) == 1))
+
+        def convert_one(op: ops.Operation) -> ops.OP_TREE:
```
Strilanc: Rename to `decompose_two_qubit_unitaries`.

cirq/contrib/paulistring/convert_to_clifford_gates.py
```diff
@@ -77,42 +77,36 @@ def _matrix_to_clifford_op(self, mat: np.ndarray, qubit: ops.QubitId
...
+                mat = protocols.unitary(op, None)
+                if mat is not None:
+                    cliff_op = self._matrix_to_clifford_op(mat, op.qubits[0])
+                    if cliff_op is not None:
+                        return cliff_op
+
+            return NotImplemented
+
+        def on_stuck_raise(op: ops.Operation):
+            if len(op.qubits) == 1 and protocols.unitary(op, None) is not None:
```
Strilanc: use the new `protocols.has_unitary`

cirq/circuits/convert_to_cz_and_single_gates.py
smitsanghavi: Why this one in particular? 
Made all of them private again.

cirq/circuits/convert_to_cz_and_single_gates.py
smitsanghavi: It's a different private method now so shouldn't be an issue

cirq/circuits/convert_to_cz_and_single_gates.py
smitsanghavi: Changed. But the test I am adding does not seem to be doing as expected. Please take a look, I'm not sure what to make of it.

cirq/circuits/convert_to_cz_and_single_gates_test.py
```diff
@@ -146,6 +146,16 @@ def test_allow_partial_czs():
...
+    circuit2 = cirq.Circuit.from_ops(
+        cirq.TwoQubitMatrixGate((np.array([[1, 0, 0, 0],
+                                           [0, 1, 0, 0],
+                                           [0, 0, 1, 0],
+                                           [0, 0, 0, 1j]]))).on(q0, q1)
+    )
+    cirq.ConvertToCzAndSingleGates(allow_partial_czs=True
+                                   ).optimize_circuit(circuit2)
+
+    assert circuit2 == c_orig
```
Strilanc: In this case there are two problems. First, the optimized circuit has a CZ**0.5 instead of the TwoQubitMatrixGate (as it should). Second, our 2-qubit decomposition is a bit, uh, lackadaisical when it comes to cleaning up single qubit operations that it generates and so there's a bunch of chaff operations.

Modify the check to assert that the list of 2-qubit operations in the circuit is of length 1 and that the operation in that list is a CZPowGate with an exponent near 0.5. `list(circuit.findall_operations(lambda e: isinstance(e, cirq.CZPowGate))) ...`.

Note: `cirq.testing.assert_same_circuit` gives much more informative error output.

cirq/circuits/convert_to_cz_and_single_gates.py
Strilanc: I don't really have a hard and fast rule, it just seemed like it was about the class instead of the method specifically.

cirq/circuits/convert_to_cz_and_single_gates.py
Strilanc: You need to assert something slightly different. See other comment.

cirq/circuits/convert_to_cz_and_single_gates_test.py
smitsanghavi: Thanks!

Issue #657, opened 2018-07-14
Use numpy einsum for base simulator 


Strilanc:
Based on my testing, `einsum` is twice as fast as the multi-processing based code in XmonSimulator. See https://github.com/quantumlib/Cirq/pull/656 (note: need to reduce precision from complex128 to complex64 to see comparable results).

dabacon:
I'm not seeing the improvement you were seeing on my box.  Here is benchmarking for circuits made up of 100 random gates selected from ExpW, ExpZ, Exp11 gates selected uniformly at random.  These are repeated 10 times (uses 2 prefix qubits = 4 shards).  This is with the unitary action downgraded to complex64

(cirq) [dabacon:cirq:~/git/cirq]$python dev_tools/profiling/benchmark_unitary_simulator.py 
num_qubits,seconds per gate
4,0.0001551498129265383
5,0.0001504874409874901
6,0.0001410450179828331
7,0.00014529275393579156
8,0.00015137688000686466
9,0.00016527387697715313
10,0.00018549127597361804
11,0.00021360345208086074
12,0.000289885765989311
13,0.0004153466949937865
14,0.0007009846249129623
15,0.001091724153026007
16,0.0018393511360045522
17,0.0035192482969723642
18,0.00644861755205784
19,0.012350965660065412
20,0.02310755885404069

(cirq) [dabacon:cirq:~/git/cirq]$python dev_tools/profiling/benchmark_xmon_simulator.py 
num_qubits,seconds per gate
4,0.0002477412990992889
5,0.00024116984999272972
6,0.00022343630704563111
7,0.0002296600310364738
8,0.0002298013960244134
9,0.00024218150600790979
10,0.00025756666297093036
11,0.00027730463002808393
12,0.0003192071299999952
13,0.0004105045989854261
14,0.0005838360510533676
15,0.0008769478329923003
16,0.0015365603070240468
17,0.0029610610279487445
18,0.004647299336036667
19,0.006038484486052767
20,0.01025167882500682

Strilanc:
Could you test this on a few more machines to get some idea of the distribution? Ideally some no-special-prep machines, no custom BLAS, etc.

Strilanc:
Oh! Also you need to make sure you're comparing equivalent precisions. I think the version in `Circuit` is using double precision at the moment, so you should change that to single if you want to compare against `XmonSimulator` properly.

kevinsung:
[...]

PR #1082, opened 2018-10-30
Methods for performing single or repeated measurements.

dabacon:


dabacon:
Part of #657

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
     norm = np.sum(np.abs(state) ** 2)
     if not np.isclose(norm, 1):
         raise ValueError('State is not normalized instead had norm %s' % norm)
+
+
+def sample_state(
```
Strilanc: Maybe `sample_state_vector`, since in the global context there may be other sorts of states?

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+    # We now have the probability vector, correctly ordered, so sample over
+    # it. Note that we us ints here, since numpy's choice does not allow for
+    # choosing from a list of tuples or list of lists.
+    result = np.random.choice(2 ** len(indices), size=repetitions, p=probs)
+    # Convert to bools and rearrange to match repetition being the outer list.
+    return np.transpose([(1 & (result >> i)).astype(np.bool) for i in
+                         range(len(indices))]).tolist()
+
+
+def measure_state(
```
Strilanc: `measure_state_vector`?

Note that users will likely see this in the autocomplete list when trying to find measurement operations for the first time. And they might try `cirq.measure(vector)` when trying to trigger this functionality. **We should either make that work, or throw a relevant error message when they pass in an np.ndarray instead of qubits to `cirq.measure`.**

If this doesn't actually affect the value of the input, it would better be to call it something like `measured_state_vector` or `state_vector_after_measurement`.

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+        indices: List[int]) -> Tuple[List[bool], np.ndarray]:
+    """Performs a measurement of the state in the computational basis.
+
+    Args:
+        state: The state to be measured. This state is assumed to be normalized.
+            The state must be of size 2 ** integer.  The state can be of shape
+            (2 ** integer) or (2, 2, ..., 2).
+        indices: Which qubits are measured. The state is assumed to be supplied
+            in big endian order. That is the xth index of v, when expressed as
+            a bitstring, has the largest values that the 0th index.
```
Strilanc: "has the largest values that the 0th index" doesn't part

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+
+    Returns:
+        A tuple of a list and an numpy array. The list is an array of booleans
+        corresponding to the measurement values (ordered by the indices). The
+        numpy array is the post measurement state. This state has the same
+        shape as the input state.
```
Strilanc: Same shape and dtype

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+    # choosing from a list of tuples or list of lists.
+    result = np.random.choice(2 ** len(indices), size=repetitions, p=probs)
+    # Convert to bools and rearrange to match repetition being the outer list.
+    return np.transpose([(1 & (result >> i)).astype(np.bool) for i in
+                         range(len(indices))]).tolist()
+
+
+def measure_state(
+        state: np.ndarray,
+        indices: List[int]) -> Tuple[List[bool], np.ndarray]:
```
Strilanc: We should provide an `out` parameter, so that callers trying to get performance don't incur allocations. If `out` isn't provided, it defaults to `np.zeros_like(state)`. And if `out` is `state`, we should modify it inline.

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+    mask[result_slice] = False
+
+    # Potentially reshape to tensor, and then set masked values to 0.
+    tensor = np.reshape(np.copy(state), [2] * num_qubits)
+    tensor[mask] = 0
+
+    # Restore original shape (if necessary) and renormalize.
+    return_state = np.reshape(tensor, state.shape) / np.sqrt(probs[b])
+
+    return (measurement_bits, return_state)
```
Strilanc: nit: unnecessary parens

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+    # Calculate the slice for the measurement result.
+    b = sum(2**i*b for i, b in enumerate(measurement_bits))
+    result_slice = linalg.slice_for_qubits_equal_to(indices, b)
+
+    # Create a mask which is False for only the slice.
+    mask = np.ones([2] * num_qubits, dtype=bool)
+    mask[result_slice] = False
+
+    # Potentially reshape to tensor, and then set masked values to 0.
+    tensor = np.reshape(np.copy(state), [2] * num_qubits)
```
Strilanc: Instead of calling `reshape`, directly assign `tensor.shape`. Avoids a copy.

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+    tensor[mask] = 0
```
Strilanc: Just use `result_slice` directly here, instead of using it to create a mask that you then use here. :wink:

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+    if len(indices) == 0:
+        return ([], np.copy(state))
+
+    # Calculate the measurement probabilities and then make the measurement.
+    probs = _probs(state, indices, num_qubits)
+    result = np.random.choice(2 ** len(indices), p=probs)
+    measurement_bits = [(1 & (result >> i)) for i in range(len(indices))]
+
+    # Calculate the slice for the measurement result.
+    b = sum(2**i*b for i, b in enumerate(measurement_bits))
```
Strilanc: ~~Use `b << i` instead of multiplying by `2**i`.~~

Wait, isn't this just `result` again? Try dropping `b` and just re-using `result`.

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+    # Create a mask which is False for only the slice.
+    mask = np.ones([2] * num_qubits, dtype=bool)
+    mask[result_slice] = False
+
+    # Potentially reshape to tensor, and then set masked values to 0.
+    tensor = np.reshape(np.copy(state), [2] * num_qubits)
+    tensor[mask] = 0
+
+    # Restore original shape (if necessary) and renormalize.
+    return_state = np.reshape(tensor, state.shape) / np.sqrt(probs[b])
```
Strilanc: Assign `shape` instead of `reshape`-ing.

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+    """
+    num_qubits = _validate_num_qubits(state)
+    _validate_indices(num_qubits, indices)
+
+    if len(indices) == 0:
+        return ([], np.copy(state))
+
+    # Calculate the measurement probabilities and then make the measurement.
+    probs = _probs(state, indices, num_qubits)
+    result = np.random.choice(2 ** len(indices), p=probs)
```
Strilanc: `len(probs)` instead of `2**len(indices)`.

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+    # Convert to bools and rearrange to match repetition being the outer list.
+    return np.transpose([(1 & (result >> i)).astype(np.bool) for i in
+                         range(len(indices))]).tolist()
+
+
+def measure_state(
+        state: np.ndarray,
+        indices: List[int]) -> Tuple[List[bool], np.ndarray]:
+    """Performs a measurement of the state in the computational basis.
+
```
Strilanc: Mention that `state` is not modified unless `out is state`.

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
     norm = np.sum(np.abs(state) ** 2)
     if not np.isclose(norm, 1):
         raise ValueError('State is not normalized instead had norm %s' % norm)
+
+
+def sample_state(
+    state: np.ndarray,
+    indices: List[int],
+    repetitions: int=1) -> List[List[bool]]:
```
Strilanc: This result might work better as a `TrialResult` or similar. We might need to streamline that class a bit to really make it work.

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+
+
+def _probs(state: np.ndarray, indices: List[int],
+        num_qubits: int):
+    """Returns the probabilities for a measurement on the given indices."""
+    # Tensor of squared amplitudes, shaped a rank [2, 2, .., 2] tensor.
+    prob_tensor = np.abs(np.reshape(state, [2] * num_qubits)) ** 2
+
+    # Calculate the probabilities for measuring the particular results.
+    probs = [np.sum(prob_tensor[linalg.slice_for_qubits_equal_to(indices, b)])
```
Strilanc:
```python3
probs = [np.norm(state[linalg.slice_for_qubits_equal_to(indices, b)], 2)
         for b ...]
```

Note that you will need to temporarily change `state`'s shape. Cache the original shape and restore it in a `finally` block, so that the caller is none the wiser.

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+    tensor[mask] = 0
+
+    # Restore original shape (if necessary) and renormalize.
+    return_state = np.reshape(tensor, state.shape) / np.sqrt(probs[b])
+
+    return (measurement_bits, return_state)
+
+
+def _probs(state: np.ndarray, indices: List[int],
+        num_qubits: int):
```
Strilanc: alignment, return type

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+    # To deal with rounding issues, ensure that the probabilities sum to 1.
+    probs /= sum(probs) # type: ignore
+    return probs
+
+
+def _validate_num_qubits(state: np.ndarray) -> int:
+    """Validates that state's size is a power of 2, returning number of qubits.
+    """
+    size = state.size
+    if (size & (size - 1)):
```
Strilanc: Nit: unnecessary outer parens

This check fails for `size==0`. Add a test ensuring that gets caught. You could use `1 << size.bit_length == size*2`.

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+        raise ValueError(
+            'State is not of size a power of two instead was {}'.format(size))
```
Strilanc: `"state.size ({}) isn't a power of two"`

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+    return size.bit_length() - 1
+
+
+def _validate_indices(num_qubits: int, indices: List[int]) -> None:
+    """Validates that the indices have values within range of num_qubits."""
+    if not all(index >= 0 for index in indices):
```
Strilanc: nit: `if any(index < 0 ...`

cirq/sim/state.py
```diff
@@ -121,3 +123,139 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+        raise IndexError('Negative index in indices: {}'.format(indices))
+    if not all(index < num_qubits for index in indices):
```
Strilanc: nit: `not all(p` -> `any(!p`

cirq/sim/state.py
dabacon: done

cirq/sim/state.py
dabacon: d

cirq/sim/state.py
dabacon: d

cirq/sim/state.py
dabacon: d

cirq/sim/state.py
dabacon: Done

cirq/sim/state.py
dabacon: done

cirq/google/sim/xmon_simulator_test.py
```diff
@@ -116,7 +116,7 @@ def test_run(scheduler):
     assert result.measurements['a'].dtype == bool
     assert result.measurements['b'].dtype == bool
     np.testing.assert_equal(result.measurements,
-                            {'a': [[True]], 'b': [[False]]})
+                            {'a': [[False]], 'b': [[True]]})
```
Strilanc: Why did this flip?

cirq/ops/common_gates.py
```diff
@@ -520,7 +520,15 @@ def measure(*qubits: raw_types.QubitId,
...
     Returns:
         An operation targeting the given qubits with a measurement.
+
+    Raises:
+        ValueError if the qubits are not instances of QubitId.
     """
+    if any(not isinstance(qubit, raw_types.QubitId) for qubit in qubits):
+        raise ValueError(
+                'measure() was called on objects not of type QubitId. '
+                'Perhaps you meant to call measure_state_vector on numpy array?'
```
Strilanc: Only give the ending hint if qubits is of type `np.ndarray`.

cirq/sim/state.py
```diff
@@ -121,3 +123,157 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+def sample_state_vector(
+    state: np.ndarray,
+    indices: List[int],
+    repetitions: int=1) -> List[List[bool]]:
+    """Samples repeatedly from measurements in the computational basis.
+
+    Note that this does not modify the passed in state.
+
+    Args:
+        state: The state to be measured. This state is assumed to be normalized.
```
Strilanc: Define what type of state. It's a multi-qubit state vector; an array of amplitudes with a power of 2 size.

cirq/sim/state.py
```diff
@@ -121,3 +123,157 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+
+    Raises:
+        ValueError if repetitions is less than one or size of state is not a
+            power of 2.
+        IndexError if the indices are out of range for the number of qubits
+            corresponding to the state.
+    """
+    if repetitions < 1:
+        raise ValueError(
+            'Number of repetitions cannot be negative. Was {}'.format(
```
Strilanc: Message doesn't match the check (0 isn't negative).

I think repetitions==0 should be allowed and explicitly tested for.

cirq/sim/state.py
```diff
@@ -121,3 +123,157 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+                repetitions))
+    num_qubits = _validate_num_qubits(state)
+    _validate_indices(num_qubits, indices)
+
+    if len(indices) == 0:
+        return [[]]
```
Strilanc: This result doesn't satisfy the invariant that the length of the other list is equal to the number of repetitions. Add a test that `sample_state_vector(..., indices=[], repetitions=2)` returns `[[], []]`.

If you return a list of tuples, instead of a list of lists, you can return `[()] * repetitions` instead of `[[] for _ in range(repetitions)]` (need to avoid aliasing between the items).

cirq/sim/state.py
```diff
@@ -121,3 +123,157 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+    This does not modify `state` unless the optional `out` is `state`.
+
+    Args:
+        state: The state to be measured. This state is assumed to be normalized.
+            The state must be of size 2 ** integer.  The state can be of shape
+            (2 ** integer) or (2, 2, ..., 2).
+        indices: Which qubits are measured. The state is assumed to be supplied
+            in big endian order. That is the xth index of v, when expressed as
+            a bitstring, has the largest values in the 0th index.
+        out: An optional place to store the result. If out is the same as
```
Strilanc: Put ticks around references to `out`, since they kinda parse as part of the sentences otherwise.

cirq/sim/state.py
```diff
@@ -121,3 +123,157 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+            is not None, then the result is put into out.  If out is None
+            a new value will be allocated. In all of these case out will be the
+            same as the returned ndarray of the method. The shape and dtype of
+            out will always match that of state.
+
+    Returns:
+        A tuple of a list and an numpy array. The list is an array of booleans
+        corresponding to the measurement values (ordered by the indices). The
+        numpy array is the post measurement state. This state has the same
+        shape and dtype as the input state.
```
Strilanc: Actually, it now matches the shape and dtype of `state if out is None else out`.

cirq/sim/state.py
```diff
@@ -121,3 +123,157 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+        IndexError if the indices are out of range for the number of qubits
+            corresponding to the state.
+    """
+    num_qubits = _validate_num_qubits(state)
+    _validate_indices(num_qubits, indices)
+
+    if len(indices) == 0:
+        return ([], np.copy(state))
+
+    # Cache initial state.
```
Strilanc: initial shape

cirq/sim/state.py
```diff
@@ -121,3 +123,157 @@ def validate_normalized_state(state: np.ndarray, num_qubits: int,
...
+    probs = _probs(state, indices, num_qubits)
+    result = np.random.choice(len(probs), p=probs)
+    measurement_bits = [(1 & (result >> i)) for i in range(len(indices))]
+
+    # Calculate the slice for the measurement result.
+    result_slice = linalg.slice_for_qubits_equal_to(indices, result)
+
+    # Create a mask which is False for only the slice.
+    mask = np.ones([2] * num_qubits, dtype=bool)
+    mask[result_slice] = False
```
Strilanc: > That reshapes the tensor and removes dimension. I think this is faster because it doesn't have to reshape the tensor and reshuffle the memory.

I don't understand what you mean by this. How can setting a subsection of the output to zero reshape it or remove dimensions? It's an assignment of the values, not the shape, and we force the shape back to the initial shape at the end.

cirq/sim/state.py
Strilanc: What do you mean by "not recurrent"? Not safe in the presence of concurrency? Does the reshaped view solve this problem?

I think you can still use `np.norm` instead of `np.sum`.

cirq/sim/state.py
Strilanc: What specifically is mypy barfing over when you set the return type to `np.ndarray`?

cirq/sim/state.py
Strilanc: It looks not done still.

cirq/sim/state.py
Strilanc: Looks not done.

cirq/sim/state.py
dabacon: I always disagree about this.  It is a common bug to end up with a do-nothing case that you don't realize because you are passing in something that is not correct and you spend lots of time before you notice this side effect.  The counter is that you should handle these edge cases so that code that hits this as an edge case is fine.  I find the later less compelling, I'd rather have my code yell at me and save me from the pain of the later than counter me not being able handle my edge cases correctly or explicitly.  

But I changed it because I'm generous that way.

cirq/sim/state.py
dabacon: Done but note that this is in conflict with your desire to support zero repetitions.

cirq/sim/state.py
dabacon: done

cirq/sim/state.py
dabacon: done

cirq/sim/state.py
dabacon:
```
>>> a = np.array([[1, 2], [3, 4]])
>>> a.shape
(2, 2)
>>> a[1,:]
array([3, 4])
>>> a[1,:].shape
(2,)
```
Setting a subsection to zero doesn't reshape it.  I said "that" in reference to your idea.  When you use the slice directly you end up with a new shape (see above) that excludes dimension.  If you put them back in I don't think numpy guarantees that it will fill back up the memory space correctly without shuffling things around (how could it?) Basically I'm forcing in place modification.  Remember you don't need to set the slice to zero you need to set everything but the slide to zero (hence the mask)

cirq/sim/state.py
dabacon: tensor is a view of shape, if you do state.shape = blah, you change the shape of state itself.  The code you wrote doesn't work because "state" may not be a tensor of the right shape.

Used np.linalg.norm

cirq/sim/state.py
dabacon: It doesn't understand division of a list by a float.

cirq/sim/state.py
Strilanc: > If you put them back in I don't think numpy guarantees that it will fill back up the memory space correctly without shuffling things around (how could it?)

I'm still not following what exactly you think will break.

My understanding (and the understanding that the entire `_apply_unitary_to_tensor_` code is based around) is that when you assign to a view of an array, numpy automatically takes care of scattering to the correct locations. So, for example, this code:

```python
v = np.zeros((2, 2, 2))
v[:, 1, :] = 1
```

is equivalent to this code:

```python
v = np.zeros((2, 2, 2))
v[0, 1, 0] = 1
v[0, 1, 1] = 1
v[1, 1, 0] = 1
v[1, 1, 1] = 1
```

> Remember you don't need to set the slice to zero you need to set everything but the slide to zero (hence the mask)

Ooooh, I see. Now it makes sense. Resolved.

cirq/sim/state.py
Strilanc: How is it in conflict? `[[] for _ in range(0)]` returns `[]`, which is the correct result for 0 repetitions.

Strilanc:
The way github changes how things behave based on if they are "outdated" or not is driving me *crazy*. I can't tell what's talking about what anymore.

cirq/sim/state.py
dabacon: done

cirq/ops/common_gates.py
```diff
@@ -481,7 +481,15 @@ def measure(*qubits: raw_types.QubitId,
 
     Returns:
         An operation targeting the given qubits with a measurement.
+
+    Raises:
+        ValueError if the qubits are not instances of QubitId.
     """
+    if any(isinstance(qubit, np.ndarray) for qubit in qubits):
```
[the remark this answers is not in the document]
dabacon: done

dabacon:
Yeah I wonder if they changed something about "outdated"

dabacon:
PTAL

Strilanc:
LGTM

cirq/sim/state.py
dabacon: Oh WTF now it's not yelling.  Why!?  Added annotation.

cirq/sim/state.py
dabacon: Do you see the change?

cirq/sim/state.py
dabacon: ah true!

dabacon:
Fixed type annotation.  Seems that it is not barfing now.  Why?  I don't know.

PR #1085, opened 2018-10-31
Add back generated proto files

dabacon:


check/pytest2-travis
```diff
@@ -0,0 +1,20 @@
...
+# Usage:
+#     check/pytest2-travis [--flags] [file-paths-relative-to-repo-root]
+#
+# You may specify additional flags or specific files to test. The file paths
+# must be relative to the repository root. If no files are specified, everything
+# is tested.
+################################################################################
+
+# Grab protoc compiler and add binary to path.
+curl -OL https://github.com/google/protobuf/releases/download/v3.5.1/protoc-3.5.1-linux-x86_64.zip
```
Strilanc: Put this directly into the yaml file as extra script lines

dev_tools/conf/apt-list-dev-tools.txt
```diff
@@ -1 +1,2 @@
+protobuf-compiler
```
Strilanc: Try this on a non-google-workstation machine and confirm that it works before adding it as a dependency. In my testing, it installs a too-old version of the protobuf compiler on external machines.

dabacon: Um, not sure what you'd do if you figured this out.  This is developer install instructions I think dealing with bad install because you have some strange distribution setup is out of scope of what we need to handle.

check/pytest2-travis
dabacon: Done

dev_tools/conf/apt-list-dev-tools.txt
dabacon: I removed it since apparently if you install this outside of ubuntu it can end up giving you the wrong version.  Instead I've added it to the developer isntructions and in the python 2 generating script it yells a warning at you if you don't have protoc installed and directs you to the github page describing how to install.

dabacon:
Thanks!

PR #1100, opened 2018-11-02
Implement dev_tools/profile_with_pyflame.sh

Strilanc:
Example profiled code:

```python
n = 10000
unitary = cirq.testing.random_unitary(4)
decomps = [cirq.kak_decomposition(unitary) for _ in range(n)]
```

Screenshot of chart view in Chrome:

image

Zoomed in:

image

I tested out what happens if I go around commenting out input and output checks in the kak decomposition (without keeping the tests passing in general) and profiled that. It gets about 4x faster even with those simple changes:

image

### DIFF
--- a/dev_tools/clean_up_pyflame_json_for_chrome.py
+++ b/dev_tools/clean_up_pyflame_json_for_chrome.py
@@ -14,7 +14,7 @@
 
 """Prettifies output from pyflame, so that it looks clearer in Chrome."""
 
-from typing import Dict, Any, Sequence, List
+from typing import Dict, Any, Sequence, List, Optional
 
 import json
 import sys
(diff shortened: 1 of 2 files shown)


Strilanc:
@dstrain115 please check that you can get this working without too much trouble before approving

dabacon:
Should this be reviewed, or did we decide not to pursue using pyflame?

dstrain115:
I'm of the opinion that this is of such limited use that it should not be committed to the repo.  If necessary, we can provide directions for devs on how to profile the code, but I don't think we should introduce another dependency just for the handful of devs who want to profile code.  Plus, there are a lot of tools to profile (I was using snakeviz, which was way easier to get running on a Goobuntu machine.)

Strilanc:
Alright, it will rest in the history; ready to be found by me six months by now when I get an urge to profile things.

PR #1102, opened 2018-11-02
Add XXPowGate with Molmer-Sorensen MS(rads) variant

mlihc15:
Adding XXPowGate for Pauli XX gate and MS gate for ion trap native Molmer-Sorensen entangling gate.


cirq/ion/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Types for representing and methods for manipulating iontrap operations.
+"""
+
+from cirq.ion.ion_gates import (
+    MSGate,
+    MS,
+)
```
vtomole: newline needed.

cirq/ion/ion_gates.py
```diff
@@ -0,0 +1,108 @@
...
+        rads = pi/4 is used.
+
+        Args:
+            rads: The angle of rotation, in radians.
+            exponent: The power that the gate operation will be raised to.
+        """
+
+        if rads is not None:
+            power = 4 * rads / np.pi
+        else:
```
vtomole: Is `else` needed here? I mean, if `rads` is `None`, `power` will be assigned to `chosen_angle_to_half_turns` without the `else`.

cirq/__init__.py
```diff
@@ -155,6 +155,8 @@
     X,
     Y,
     Z,
+    XXPowGate,
```
Strilanc: Use alphabetical order

cirq/__init__.py
```diff
@@ -217,6 +219,11 @@
     trace_distance_bound,
     phase_by,
 )
+
+from cirq.ion import (
+    MSGate,
```
Strilanc: Nit: shorter comes first in alphabetical order

cirq/ion/ion_gates.py
```diff
@@ -0,0 +1,108 @@
...
+
+import numpy as np
+
+from cirq import value, linalg, protocols
+from cirq.ops import gate_features, eigen_gate, raw_types, gate_operation
+from cirq.type_workarounds import NotImplementedType
+
+
+
+class MSGate(eigen_gate.EigenGate,
```
Strilanc: This shouldn't be a separate class, it should be the following method (plus docstring):

```python
def MS(rads: float) -> XXPowGate:
    return ops.XXPowGate(exponent=rads*np.pi, global_shift=-0.5)
```

cirq/ops/common_gates.py
```diff
@@ -982,3 +982,74 @@ def Ry(rads: float) -> RotYGate:
...
+    The gate implements the following unitary:
+    X⊗X = [0 0 0 1]
+          [0 0 1 0]
+          [0 1 0 0]
+          [1 0 0 0]
+
+    """
+
+    def __init__(self, *,  # Forces keyword args.
+                 exponent: Optional[Union[value.Symbol, float]] = None) -> None:
```
Strilanc: Add a `global_shift` parameter analogous to the one in `XPowGate`.

cirq/ops/common_gates.py
```diff
@@ -982,3 +982,74 @@ def Ry(rads: float) -> RotYGate:
...
+            exponent: The power that the gate operation will be raised to.
+        """
+
+        super().__init__(exponent=value.chosen_angle_to_half_turns(half_turns=exponent))
+
+    @property
+    def exponent(self) -> Union[value.Symbol, float]:
+        return self._exponent
+
+    def _eigen_components(self):
```
Strilanc: Add `def _eigen_shift(self): return (0, 1)`, which will accelerate the period computations.

cirq/ops/common_gates.py
```diff
@@ -982,3 +982,74 @@ def Ry(rads: float) -> RotYGate:
...
+            (0., np.array([[0.5, 0, 0, 0.5],
+                            [0, 0.5, 0.5, 0],
+                            [0, 0.5, 0.5, 0],
+                            [0.5, 0, 0, 0.5]]))
+        ]
+
+    def _canonical_exponent_period(self) -> Optional[float]:
+        return 2
+
+    def _with_exponent(self,
```
Strilanc: After `global_shift` parameter is added, this method can be removed.

cirq/ops/common_gates.py
```diff
@@ -982,3 +982,74 @@ def Ry(rads: float) -> RotYGate:
...
+                            [0, 0.5, -0.5, 0],
+                            [0, -0.5, 0.5, 0],
+                            [-0.5, 0, 0, 0.5]])),
+            (0., np.array([[0.5, 0, 0, 0.5],
+                            [0, 0.5, 0.5, 0],
+                            [0, 0.5, 0.5, 0],
+                            [0.5, 0, 0, 0.5]]))
+        ]
+
+    def _canonical_exponent_period(self) -> Optional[float]:
```
Strilanc: `EigenGate` now automatically computes the period, so you can delete this method. (You may need to sync to master for it to work; it was added in the last week.)

cirq/ops/common_gates.py
```diff
@@ -982,3 +982,74 @@ def Ry(rads: float) -> RotYGate:
...
+        return 2
+
+    def _with_exponent(self,
+                       exponent: Union[value.Symbol, float]) -> 'XXPowGate':
+        return XXPowGate(exponent=exponent)
+
+    def _circuit_diagram_info_(self, args: protocols.CircuitDiagramInfoArgs
+                               ) -> protocols.CircuitDiagramInfo:
+        return protocols.CircuitDiagramInfo(
```
Strilanc: Add

```python
    if self._global_shift == -0.5:
            return _rads_func_symbol(
                'MS',
                args,
                self._diagram_exponent(args, ignore_global_phase=False))
```

to the start of this method.

Check out https://github.com/quantumlib/Cirq/pull/1101/files#diff-08a88625437f21ce336680722c6942d5R104 where this method is defined. Note that that PR is not yet merged.

cirq/ops/common_gates.py
```diff
@@ -982,3 +982,74 @@ def Ry(rads: float) -> RotYGate:
...
+            wire_symbols=('XX', 'XX'),
+            exponent=self.exponent)
+
+    def __str__(self) -> str:
+        if self.exponent == 1:
+            return 'XX'
+        return 'XX**{!r}'.format(self.exponent)
+
+    def __repr__(self) -> str:
+        if self.exponent == 1:
```
Strilanc: Mimmic the logic in `XPowGate` to return `MS(...)` when `global_shift` is `-0.5`.

cirq/ops/common_gates.py
```diff
@@ -982,3 +982,74 @@ def Ry(rads: float) -> RotYGate:
...
+                       exponent: Union[value.Symbol, float]) -> 'XXPowGate':
+        return XXPowGate(exponent=exponent)
+
+    def _circuit_diagram_info_(self, args: protocols.CircuitDiagramInfoArgs
+                               ) -> protocols.CircuitDiagramInfo:
+        return protocols.CircuitDiagramInfo(
+            wire_symbols=('XX', 'XX'),
+            exponent=self.exponent)
+
+    def __str__(self) -> str:
```
Strilanc: Mimmic the logic in `XPowGate` to describe with `MS(...)` when `global_shift` is `-0.5`.

cirq/ion/ion_decomposition.py
```diff
@@ -0,0 +1,173 @@
+# Copyright 2018 The Cirq Developers
```
Strilanc: You are mixing later changes into this PR, which should specifically be about XXPowGate. Open a separate PR (or followup PR) for these changes.

cirq/ops/common_gates.py
```diff
@@ -911,3 +911,88 @@ def Ry(rads: float) -> YPowGate:
...
+                            [0, 0.5, 0.5, 0],
+                            [0, 0.5, 0.5, 0],
+                            [0.5, 0, 0, 0.5]]))
+        ]
+
+    def _circuit_diagram_info_(self, args: protocols.CircuitDiagramInfoArgs
+                               ) -> Union[str, protocols.CircuitDiagramInfo]:
+        if self._global_shift == -0.5:
+            return _rads_func_symbol(
+                'MS',
```
mlihc15: This is not correct. How do you do _rads_func_symbols for two-qubit gates? What would they look like?

cirq/ion/ion_gates.py
```diff
@@ -0,0 +1,36 @@
...
+# limitations under the License.
+
+"""Operations native to iontrap systems."""
+
+import numpy as np
+
+from cirq import ops
+
+
+def MSGate(rads: float) -> ops.XXPowGate:
```
Strilanc: just `def MS(`.

cirq/ion/ion_gates.py
```diff
@@ -0,0 +1,36 @@
...
+    native two-qubit operation in ion traps.
+    Effectively equivalents to rotation around the XX axis
+    in the two-qubit bloch sphere.
+
+    The gate implements the following unitary:
+    exp(-i t XX) = [cos(t) 0 0 -isin(t)]
+                   [0 cos(t) -isin(t) 0]
+                   [0 -isin(t) cos(t) 0]
+                   [-isin(t) 0 0 cos(t)]
+    rads = t, the rotation angle in radians
```
Strilanc: Put this in an `Args:` section so that tools parse it.

```
Args:
    rads: The rotation angle in radians.

Returns:
    Mølmer–Sørensen gate rotating by the desired amount.
```

cirq/ion/ion_gates.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    return ops.XXPowGate(exponent=rads*2/np.pi, global_shift=-0.5)
```
Strilanc: It gets multiplied by 2? So MS(pi) isn't equal to Rx(pi) * Rx(pi)? Is that consistent with the matrix in the doc?

cirq/ion/ion_gates_test.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def test_MSGate_str():
+    assert str(cirq.MSGate(np.pi/2)) == 'MS(np.pi/2)'
+    assert str(cirq.MSGate(np.pi)) == 'MS(np.pi/2*2.0)'
+
+
+def test_MSGate_matrix():
+    s = np.sqrt(0.5)
+    assert np.allclose(cirq.unitary(cirq.MSGate(np.pi/4)),
```
Strilanc: Use `np.testing.assert_allclose`, which has better error output and forces the `atol` argument.

cirq/ops/common_gates.py
```diff
@@ -911,3 +911,97 @@ def Ry(rads: float) -> YPowGate:
...
+            return 'MS(np.pi/2*{!r})'.format(self._exponent)
+        if self.exponent == 1:
+            return 'XX'
+        return 'XX**{!r}'.format(self._exponent)
+
+    def __repr__(self) -> str:
+        if self._global_shift == -0.5:
+            if self._exponent == 1:
+                return 'cirq.MSGate(np.pi/2)'
+            return '(cirq.MSGate(np.pi/2*{!r}))'.format(self._exponent)
```
Strilanc: You don't need the outer parens here, because the parameter is already enclosed in the argument list parens.

cirq/ops/common_gates.py
```diff
@@ -911,3 +911,97 @@ def Ry(rads: float) -> YPowGate:
...
+        if self._exponent == 1:
+            return 'cirq.XX'
+        return ('cirq.XXPowGate(exponent={!r}, '
```
Strilanc:
```python
if self._global_shift == 0:
    return '(cirq.XX**{!r})'.format(self._exponent)
```

cirq/ops/common_gates.py
```diff
@@ -911,3 +911,97 @@ def Ry(rads: float) -> YPowGate:
...
+                 global_shift: float = 0.0) -> None:
+        """
+        Initializes the gate.
+
+        The exponent may be specified.
+        If no argument is given, the default value of
+        exponent = 1 is used.
+
+        Args:
+            exponent: The t in XX**t. Phases the eigenstate of the Pauli XX
```
Strilanc: The -1 eigenstate

cirq/ops/common_gates_test.py
```diff
@@ -797,3 +797,66 @@ def test_rxyz_circuit_diagram():
...
+
+def test_XXPowGate_diagrams():
+    a = cirq.NamedQubit('a')
+    b = cirq.NamedQubit('b')
+    circuit = cirq.Circuit.from_ops(
+        cirq.SWAP(a, b),
+        cirq.X(a),
+        cirq.Y(a),
+        cirq.XX(a, b))
+    cirq.testing.assert_has_diagram(circuit, """
```
Strilanc: Do a test with a non-1 exponent

cirq/ion/ion_gates_test.py
```diff
@@ -0,0 +1,60 @@
...
+def test_MSGate_diagrams():
+    a = cirq.NamedQubit('a')
+    b = cirq.NamedQubit('b')
+    circuit = cirq.Circuit.from_ops(
+        cirq.SWAP(a, b),
+        cirq.X(a),
+        cirq.Y(a),
+        cirq.MSGate(np.pi).on(a, b))
+    cirq.testing.assert_has_diagram(circuit, """
+a: ───×───X───Y───MS(2.0π)───
```
Strilanc: You passed in pi, but the diagram says to pass in 2pi?

cirq/ops/common_gates.py
Strilanc: Sorry for the bad code before:

```
 if self._global_shift == -0.5:
    symbol = _rads_func_symbol(
        'MS',
        args,
        self._diagram_exponent(args, ignore_global_phase=False))
    return symbol, '#2'  
```

cirq/ion/ion_gates.py
mlihc15: MS(t) = exp(-i t XX) != exp(-i t/2 X) * exp(-i t/2 X).
I am using exponent=rads*2/np.pi because Pauli XX = -i exp(-i pi/2 XX)

Strilanc: Okay, I see. That's a bit unusual, that rotating by pi only gets a you a quarter of the way through the global-phase-invariant cycle.

The number being passed in is still considered to be in radians, right? People won't be confused by the argument being called `rads` instead of, like `half_rads`?

We'll have to make sure that the diagram exponent and so forth are all consistent with this convention.

mlihc15: Yes the number being passed in is still in radians.
RX(t) = exp(-i t/2 X) is the convention for single qubit gates because of the bloch sphere. I don't think such convention still stands for two-qubit gates. As long as people know they are doing MS(t) = exp(-i t XX) there should be no confusion.

mlihc15: And I am fixing the diagram exponent for MS right now. No need to change the _diagram_exponent function.

cirq/ion/ion_gates_test.py
```diff
@@ -30,12 +30,12 @@ def test_MS_str():
 
 def test_MS_matrix():
     s = np.sqrt(0.5)
-    assert np.testing.assert_allclose(cirq.unitary(cirq.MS(np.pi/4)),
+    np.testing.assert_allclose(cirq.unitary(cirq.MS(np.pi/4)),
```
Strilanc: You need to specify an absolute error tolerance. We typically use `atol=1e-8`.

PR #1103, opened 2018-11-02
Delete XmonGate class

Strilanc:


Strilanc:
Just a few more little cuts...

cirq/google/programs.py
```diff
@@ -68,8 +68,8 @@ def _x_to_proto_dict(gate: ops.XPowGate, q: ops.QubitId) -> Dict:
     exp_w = {
         'target': cast(devices.GridQubit, q).to_proto_dict(),
         'axis_half_turns':
-            xmon_gates.XmonGate.parameterized_value_to_proto_dict(0),
-        'half_turns': xmon_gates.XmonGate.parameterized_value_to_proto_dict(
+            xmon_gates._parameterized_value_to_proto_dict(0),
```
dabacon: always slightly strange when a module calls another modules _private method.  Maybe it doesn't belong in cirq.google but might still make it non _private.

cirq/google/xmon_gates.py
```diff
@@ -27,117 +26,96 @@
...
+        raise ValueError('invalid operation: {}'.format(proto_dict))
+
+
+def _parameterized_value_from_proto_dict(message: Dict
+                                         ) -> Union[value.Symbol, float]:
+    if 'raw' in message:
+        return message['raw']
+    if 'parameter_key' in message:
+        return value.Symbol(message['parameter_key'])
+    raise ValueError('No value specified for parameterized float.')
```
dabacon: show message in error string.

cirq/google/programs.py
Strilanc: Yup. There's an upcoming change where the xmon gates file is disappearing and will merge into programs.py.

cirq/google/xmon_gates.py
Strilanc: done.

Strilanc:
All covered now

cirq/google/xmon_gates.py
```diff
@@ -14,9 +14,8 @@
 
 """Gates that can be directly described to the API, without decomposition."""
 
-from typing import Dict, Optional, Union, Tuple, Any, cast
+from typing import Dict, Union, Tuple, Any, cast
```
dabacon: nit: alphabetical

Strilanc: Acknowledged. Leaving as is because this file is not long for this world.

PR #1104, opened 2018-11-02
Add PhasedXPowGate._qasm_ and assert_qasm_is_consistent_with_unitary

Strilanc:
- Added qiskit as a dev tool
- Added more compact qasm output for the phased x gate
- Fixed using now-deprecated-and-deleted methods to run the qasm simulation


vtomole:
@dabacon "ok I guess". Why the hesitation?

PR #1109, opened 2018-11-06
Add KakDecomposition type

Strilanc:
- Rewrite existing code using the big-complicated-tuple to instead use the named type
- Implement `_unitary_` on `KakDecomposition`

### DIFF
--- a/cirq/linalg/decompositions.py
+++ b/cirq/linalg/decompositions.py
@@ -15,7 +15,7 @@
 
 """Utility methods for breaking matrices into useful pieces."""
 
-from typing import Set  # pylint: disable=unused-import
+from typing import Set, NamedTuple  # pylint: disable=unused-import
 from typing import Callable, List, Tuple, TypeVar
 
 import math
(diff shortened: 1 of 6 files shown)


cirq/linalg/decompositions.py
```diff
@@ -264,12 +264,115 @@ def so4_to_magic_su2s(
...
-) -> Tuple[complex,
-           Tuple[np.ndarray, np.ndarray],
-           Tuple[float, float, float],
-           Tuple[np.ndarray, np.ndarray]]:
+class KakDecomposition:
+    """A convenient description of an arbitrary two-qubit operation.
+
+    Any two qubit operation U = can be decomposed into the form
+
+        U = g · (a1 ⊗ a0) · exp(i·(x·XX + y·YY + z·ZZ)) · (b1 ⊗ b0)
```
dstrain115: Unicode tensor product symbol.  Nice!

cirq/linalg/decompositions.py
```diff
@@ -264,12 +264,115 @@ def so4_to_magic_su2s(
...
+
+    This class stores g, (b0, b1), (x, y, z), and (a0, a1).
```
dstrain115: Would it be worth linking to a page (or the paper you linked to https://arxiv.org/abs/quant-ph/0507171) for more information?

cirq/linalg/decompositions.py
```diff
@@ -264,12 +264,115 @@ def so4_to_magic_su2s(
...
+        single_qubit_operations_after: a0, a1 from the above equation.
+    """
+
+    def __init__(self,
+                 *,
+                 global_phase: complex,
+                 single_qubit_operations_before: Tuple[np.ndarray, np.ndarray],
+                 interaction_coefficients: Tuple[float, float, float],
+                 single_qubit_operations_after: Tuple[np.ndarray, np.ndarray]):
+        """Initializes a decomposition for a two-qubit operation U.
```
dstrain115: This looks the same as the class pydoc.  Refer to that instead?

cirq/linalg/decompositions.py
```diff
@@ -264,12 +264,115 @@ def so4_to_magic_su2s(
...
+        after = np.kron(*self.single_qubit_operations_after)
+
+        def interaction_matrix(m: np.ndarray, c: float) -> np.ndarray:
+            return map_eigenvalues(np.kron(m, m),
+                                   lambda v: np.exp(1j * v * c))
+
+        x, y, z = self.interaction_coefficients
+        x_mat = np.array([[0, 1], [1, 0]])
+        y_mat = np.array([[0, -1j], [1j, 0]])
+        z_mat = np.array([[1, 0], [0, -1]])
```
dstrain115: Should all of this be computed at initialization rather than each time the unitary matrix is needed?
(Not quite sure which would be more efficient, if these will be re-used often)

cirq/linalg/decompositions_test.py
```diff
@@ -217,7 +207,72 @@ def test_kak_canonicalize_vector(x, y, z):
...
+        single_qubit_operations_after=(np.eye(2), cirq.unitary(cirq.Z)),
+    ))
+
+    eq.make_equality_group(lambda: cirq.KakDecomposition(
+        global_phase=1,
+        single_qubit_operations_before=(cirq.unitary(cirq.X),
+                                        cirq.unitary(cirq.Y)),
+        interaction_coefficients=(0.5, 0.2, 0.1),
+        single_qubit_operations_after=(np.eye(2), cirq.unitary(cirq.Z)),
+    ))
```
dstrain115: This just tests that all of these are not equal, right?  Should we have a group with more than one element?

cirq/linalg/decompositions.py
Strilanc: It is rare to ask for the unitary matrix, so it's not worth caching. You almost always go in the opposite direction. This is more of a debugging-and-testing method.

cirq/linalg/decompositions_test.py
Strilanc: Note that one of them is a `make_` call instead of `add_`. `make_` invokes the lambda twice and checks the results for equality. It exists exactly for this case, to avoid repetition.

cirq/linalg/decompositions.py
Strilanc: Acknowledged. They may appear in different places (e.g. based on what is being autocompleted in an IDE, or if the user is making-vs-consuming a KakDecomposition), so I think the repetition is worth it in this case. My opinion is likely to change as I look into the finer details of how our documentation is generated.

cirq/linalg/decompositions.py
Strilanc: Done.

PR #1111, opened 2018-11-06
Delete cirq.extensions

### LINKED ISSUE #672: Remove Extensions in favor of protocols

Strilanc:
~~Comes after #1110~~
Fixes #672


cirq/contrib/quirk/quirk_gate.py
```diff
@@ -12,12 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Optional, Any, Union
+from typing import Optional, Any, Union, cast, Dict, Callable
```
dabacon: nit: alphabetical

PR #1115, opened 2018-11-06
Add __pow__ to circuit and moments

dabacon:
We could restrict to integers if we think
Moment**0.5 shoudn't be defined as all op's raised to 0.5 power.

### DIFF
--- a/cirq/circuits/circuit.py
+++ b/cirq/circuits/circuit.py
@@ -238,6 +238,37 @@ class Circuit:
             return NotImplemented
         return self * repetitions
 
+    def __pow__(self, exponent: float):
+        """See `cirq.pow`.
+
+        For integer exponents, this corresponds to repetitions of the moments
+        of the circuit. When the exponent is negative, these circuits have
+        their moment order reversed and the inverse of each operation in the
+        moment is applied. This means circuit**-1 corresponds to the inverse
+        circuit.
+
+        For non-integer exponents, the integer closest to zero is calculated
+        and this is used as the integer exponent above. The remainder is
+        then applied as a power for each operation, again if the remainder is
+        negative the order is reversed.
+
+        Args:
+            exponent: The exponent to power by.
+
+        Returns:
+            A new circuit of the corresponding power.
+        """
+        d = 1 if exponent > 0 else -1
+        n, r = divmod(exponent, d)
+        result = Circuit()
+        if int(n) != 0:
+            result += int(n) * Circuit(
+                    [protocols.pow(moment, d) for moment in self[::d]])
+        if r != 0:
+            result += Circuit(
+                    [protocols.pow(moment, r) for moment in self[::d]])
+        return result
+
     def __repr__(self):
         if not self._moments and self._device == devices.UnconstrainedDevice:
             return 'cirq.Circuit()'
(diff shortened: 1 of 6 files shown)


cirq/circuits/circuit.py
```diff
@@ -238,6 +238,37 @@ def __rmul__(self, repetitions: int):
...
+    def __pow__(self, exponent: float):
+        """See `cirq.pow`.
+
+        For integer exponents, this corresponds to repetitions of the moments
+        of the circuit. When the exponent is negative, these circuits have
+        their moment order reversed and the inverse of each operation in the
+        moment is applied. This means circuit**-1 corresponds to the inverse
+        circuit.
+
+        For non-integer exponents, the integer closest to zero is calculated
```
Strilanc: Do we need the non-integer case?

I'm fine with defining a convention for raising moments to a fractional power. But it's odd that the convention is different from the convention used elsewhere in the library, which is that raising to a power maps directly onto the eigenvalues of a thing. For example, this definition is inconsistent with the definition we're using for `XXPowGate` meaning`circuit((X1 * X2)**t) != circuit((X1 * X2))**t`.

dabacon:
#1113

cirq/circuits/circuit.py
Strilanc: Hmmm, that is another interesting point; that we already define multiplication to be repetition.

I agree that getting the inverse to work is good. Integer powers are a bit ambiguous w.r.t. multiplication but probably worth it. I think we should hold off on fractional powers.

cirq/circuits/circuit_test.py
```diff
@@ -2532,3 +2532,52 @@ def test_decompose():
...
+    # Note that moment does not reverse order of operations in moment.
+    inverse_ops = [(cirq.CZ ** -1)(a, b), (cirq.X ** -0.5)(a),
+                   (cirq.Y ** -0.2)(b)]
+    inverse_sqrt_ops = [(cirq.CZ ** -0.5)(a, b), (cirq.X ** -0.25)(a),
+                        (cirq.Y ** -0.1)(b)]
+
+    expected_ops = inverse_ops + inverse_ops + inverse_sqrt_ops
+    cirq.testing.assert_same_circuits(base ** (-2.5),
+                                      cirq.Circuit.from_ops(expected_ops))
+
```
Strilanc: Some gates don't have inverses (e.g. measurement). Check that the behavior is correct when inverting measurement (i.e. that `cirq.inverse(..., default=None)` returns `None` instead of raising an error).

dabacon: done

cirq/circuits/circuit.py
```diff
@@ -247,36 +247,29 @@ def __rmul__(self, repetitions: int):
...
-        n, r = divmod(exponent, d)
-        result = Circuit()
-        if int(n) != 0:
-            result += int(n) * Circuit(
-                    [protocols.pow(moment, d) for moment in self[::d]])
-        if r != 0:
-            result += Circuit(
-                    [protocols.pow(moment, r) for moment in self[::d]])
-        return result
+        if exponent != -1:
```
Strilanc: I'm also okay with supporting `exponent==1`.

cirq/protocols/inverse.py
```diff
@@ -66,6 +66,16 @@ def inverse(val: 'cirq.OP_TREE',
     pass
 
 
+@overload
```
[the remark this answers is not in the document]
dabacon: done

PR #1118, opened 2018-11-07
Use numpy simulator for Simulator class

dabacon:
* Note choice of name, we could make it less generic but I assume this will be the main simulator folks will use.


README.rst
```diff
@@ -39,7 +39,7 @@ A simple example to get you up and running:
   print(circuit)
 
   # Simulate the circuit several times.
-  simulator = cirq.google.XmonSimulator()
+  simulator = cirq.Simulator()
```
Strilanc: Yesssss. One step closer to just `simulate` :wink:

cirq/google/sim/xmon_simulator.py
```diff
@@ -126,7 +127,7 @@ def _run(
         param_resolver: study.ParamResolver,
         repetitions: int,
     ) -> Dict[str, List[np.ndarray]]:
-        """Runs the entire supplied Circuit, mimicking the quantum hardware."""
+        """See definition in `sim.SimulatesSamples`."""
```
Strilanc: Prefer `cirq.SimulatesSamples`, because that is how the user will think about it.

cirq/google/sim/xmon_simulator.py
```diff
@@ -157,9 +158,8 @@ def _run_sweep_sample(self, circuit, repetitions):
         step_result = None
         for step_result in all_step_results:
             pass
-        return _sample_measurements(circuit,
-                                    step_result,
-                                    repetitions)
+        return wave_function.sample_terminal_measurements(circuit, step_result,
```
Strilanc: Nit: indentation

cirq/google/sim/xmon_simulator.py
```diff
@@ -169,7 +169,7 @@ def _simulator_iterator(
         initial_state: Union[int, np.ndarray],
         perform_measurements: bool = True,
     ) -> Iterator['XmonStepResult']:
-        """See definition in SimulatesFinalWaveFunction."""
+        """See definition in `sim.SimulatesIntermediateWaveFunction`."""
```
Strilanc: Prefer `cirq.[...]` in docstrings

cirq/sim/wave_function.py
```diff
@@ -282,3 +286,40 @@ def _validate_indices(num_qubits: int, indices: List[int]) -> None:
     if any(index >= num_qubits for index in indices):
         raise IndexError('Out of range indices, must be less than number of '
                          'qubits but was {}'.format(indices))
+
+
+def sample_terminal_measurements(
```
Strilanc: Is there a reason that we are exposing "terminal only" as a constraint on the method, instead of simply picking the appropriate method for the user based on the circuit they gave to us? Just as a way of communicating that this particular case will be much faster?

cirq/sim/wave_function.py
```diff
@@ -282,3 +286,40 @@ def _validate_indices(num_qubits: int, indices: List[int]) -> None:
...
+        circuit: circuits.Circuit,
+        last_step_result: 'simulator.StepResult',
+        repetitions: int) -> Dict[str, List]:
+    """Sample from measurements in the given circuit.
+
+    This should only be called if the circuit has only terminal measurements.
```
Strilanc: Explain why

cirq/sim/wave_function.py
```diff
@@ -282,3 +286,40 @@ def _validate_indices(num_qubits: int, indices: List[int]) -> None:
...
+            taken.
+        repetitions: The number of time to sample.
+
+    Returns:
+        A dictionary from the measurement keys to the measurement results.
+        These results are lists of lists, with the outer list corresponding to
+        the repetition, and the inner list corresponding to the qubits as
+        ordered in the measurement gate.
+    """
+    if last_step_result is None:
```
Strilanc: If the user calls this with a non-terminal measurement circuit, it silently gives bad results?

cirq/sim/wave_function_test.py
```diff
@@ -11,14 +11,15 @@
...
-"""Tests for wavefunction.py"""
+"""Tests for wave_function.py"""
 
 import itertools
 import pytest
 
 import numpy as np
 
 import cirq
+from cirq import sim
```
Strilanc: Just `import cirq` in tests

cirq/sim/wave_function.py
Strilanc: After looking more carefully I realize this method was not doing at all what I expected. We need a better name for it.

Maybe it should be an instance method of `StepResult`?

cirq/sim/wave_function.py
dabacon: Moved this into step result where it has a different meaning which should make it clear

cirq/sim/wave_function.py
dabacon: Moved into step result.  The semantics are that you need to pass it a set of measurement results that act on disjoint qubits.  Since it is on a step result it s clear that it means (at this time).  The part about being terminal is handled by the simulator class.

cirq/sim/wave_function_test.py
dabacon: The method I moved was only exposed at cirq.sim because I don't think it sould be part of the public interface.  But now it is moved into step result, and has a better signature.

docs/simulation.md
```diff
@@ -1,14 +1,15 @@
...
+Cirq comes with built in Python simulators for testing out
+small circuits.  One of these simulators works for generic
+gates that implement their unitary matrix, ``cirq.Simulator``.
+The other simulator is customized for the native gate set
+of Google's Xmon hardware ``cirq.google.XmonSimulator``.
+This later simulator can shard its simulation  across
+different processes/threads and so take advantage of
+multiple cores/CPUs.  Depending on your local computer
+architecture one or the other of these may be faster, but
+we recommend start with ``cirq.Simulator``.
```
Strilanc: starting

docs/simulation.md
```diff
@@ -175,24 +178,41 @@ for i, step in enumerate(simulator.simulate_moment_steps(circuit)):
 ```
 
 The object returned by the ``moment_steps`` iterator is a 
-``XmonStepResult``. This object has the state along with any 
+``tepResult``. This object has the state along with any
```
Strilanc: `StepResult`

docs/simulation.md
```diff
@@ -175,24 +178,41 @@ for i, step in enumerate(simulator.simulate_moment_steps(circuit)):
...
 measurements that occurred **during** that step (so does
 not include measurement results from previous ``Moments``).
-In addition, the``XmonStepResult`` contains ``set_state()``
+In addition, the``StepResult`` contains ``set_state()``
```
Strilanc: missing space after the

Why do you double-tick instead of single-tick?

docs/simulation.md
```diff
@@ -175,24 +178,41 @@ for i, step in enumerate(simulator.simulate_moment_steps(circuit)):
...
 will be set lie entirely in the computation basis state
 for the binary expansion of the passed integer.
 
+### XmonSimulator
+
+In addition to ``cirq.Simulator`` there is also a simulator
+which is specialized to the Google native gate set. In
+particular this simulator is specialized to use the
+``CZPowGate``, ``MeasurementGate``, ``PhasedXPowGate``,
+``XPowGate``, ``YPowGate``, and the ``ZPowGate``. This
```
Strilanc: Don't forget `PhasedXPowGate`

cirq/sim/sparse_simulator.py
```diff
@@ -0,0 +1,244 @@
...
+
+import numpy as np
+
+from cirq import circuits, study, ops, protocols
+from cirq.sim import simulator, wave_function
+
+
+class Simulator(simulator.SimulatesSamples,
+                simulator.SimulatesIntermediateWaveFunction):
+    """A sparse matrix wave function simulator that uses numpy.
```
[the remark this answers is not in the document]
Strilanc: Okay, sounds good.

docs/simulation.md
Strilanc: typo

docs/simulation.md
Strilanc: ... I only read the second line.

cirq/sim/sparse_simulator.py
```diff
@@ -0,0 +1,244 @@
...
+    def state(self) -> np.ndarray:
+        return self._state
+
+    def set_state(self, state: Union[int, np.ndarray]):
+        update_state = wave_function.to_valid_state_vector(state,
+                                                           len(self.qubit_map),
+                                                           self._dtype)
+        np.copyto(self._state, update_state)
+
+    def sample(self, qubits: List[ops.QubitId], repetitions: int = 1):
```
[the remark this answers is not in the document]
Strilanc: I do think that the "no repetitions set means not list" pattern is good, but it's fine to leave it as is for now (yay alpha!). Note that `itertools.product` works in exactly this fashion w.r.t. the `repeat` parameter, so it is an existing pattern within python.

I do agree that someone passing along a repetitions parameter that happens to be None can get quite the surprise. We could avoid this by making the default parameter unspeakable (e.g. via a private `([],)` value compared to with `is`).

dabacon:
Ugh merge broke, should fixed now

Issue #1128, opened 2018-11-07
Move no-longer-xmon-specific optimizers from `cirq.google` to `cirq.optimizers`


Strilanc:

dstrain115:
This has been completed.  All optimizers except ConvertToXmon() have been moved to cirq.optimizers.

https://github.com/quantumlib/Cirq/blob/master/cirq/google/optimizers/optimize_for_xmon.py#L37

PR #1130, opened 2018-11-08
Move cirq/google/eject_full_w to cirq/optimizers/eject_phased_paulis

Strilanc:
Part of https://github.com/quantumlib/Cirq/issues/1128

### DIFF
--- a/cirq/optimizers/eject_phased_paulis.py
+++ b/cirq/optimizers/eject_phased_paulis.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""An optimization pass that pushes some W gates later and later in the circuit.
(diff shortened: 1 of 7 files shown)


cirq/optimizers/eject_phased_paulis.py
```diff
@@ -115,6 +115,8 @@ def _absorb_z_into_w(moment_index: int,
                      state: _OptimizerState) -> None:
     """Absorbs a Z^t gate into a W(a) flip.
 
+    [Where W(a) is shorthand for PhasedX(phase_exponent=a).]
```
viathor: Optional: There seem to be a lot of places where we call this gate W. We're bound to miss some of the explanation of the symbol at some point. An alternative would be to adopt a short yet self-explanatory name such as "PhX" in place of W and remove the [Where...] comments. Your call.

Issue #1135, opened 2018-11-08
Controlled Rotation has the wrong relative phase


sleichen:
ControlledGate(Rx(pi)) produces a CNOT where it should be have an extra factor of -i in the lower part of the matrix. Here is some code:

Input:
```python
import cirq
import numpy as np
a = cirq.NamedQubit("a")
b = cirq.NamedQubit("b")
angle = np.pi
circuit = cirq.Circuit.from_ops(cirq.ControlledGate(cirq.Rx(angle)).on(a,b))
circuit.to_unitary_matrix()
```
Output:
```
array([[1.+0.j, 0.+0.j, 0.+0.j, 0.+0.j],
       [0.+0.j, 1.+0.j, 0.+0.j, 0.+0.j],
       [0.+0.j, 0.+0.j, 0.+0.j, 1.+0.j],
       [0.+0.j, 0.+0.j, 1.+0.j, 0.+0.j]])
```
An angle very close to pi appears to have the correct behavior:
Input:
```python
import cirq
import numpy as np
a = cirq.NamedQubit("a")
b = cirq.NamedQubit("b")
angle = 1.000001*np.pi
circuit = cirq.Circuit.from_ops(cirq.ControlledGate(cirq.Rx(angle)).on(a,b))
circuit.to_unitary_matrix()
```
Output:
```
array([[ 1.+0.j,  0.+0.j,  0.+0.j,  0.+0.j],
       [ 0.+0.j,  1.+0.j,  0.+0.j,  0.+0.j],
       [ 0.+0.j,  0.+0.j, -0.+0.j,  0.-1.j],
       [ 0.+0.j,  0.+0.j,  0.-1.j, -0.+0.j]])
```

Strilanc:
Good catch. This is because the `_apply_unitary_to_tensor_` method of `XPowGate` is failing to include the global phase.

kevinsung:
This will be fixed at the same time as #1053

Strilanc:
This is release blocking IMO, so I fixed it right away in https://github.com/quantumlib/Cirq/pull/1136

PR #1136, opened 2018-11-08
Refactor `assert_apply_unitary_to_tensor_is_consistent_with_unitary` into three methods

Strilanc:
- the variant `assert_has_consistent_apply_unitary` with no 'try multiple values' arguments
- the variant `assert_has_consistent_apply_unitary_for_various_exponents` that raises the value under test to various powers while testing
- the variant `assert_eigen_gate_has_consistent_apply_unitary` which raises to various exponents and adds various global shifts while testing
- fixed XPowGate and ZPowGate not accounting for global shift when applying their unitaries

Fixes https://github.com/quantumlib/Cirq/issues/1135

### DIFF
--- a/cirq/ops/common_gates_test.py
+++ b/cirq/ops/common_gates_test.py
@@ -74,9 +74,8 @@ def test_cz_matrix():
                                  [0, 0, 1, 0],
                                  [0, 0, 0, -1j]]))
 
-    cirq.testing.assert_apply_unitary_to_tensor_is_consistent_with_unitary(
-        val=cirq.CZ,
-        exponents=[1, -0.5, 0.5, 0.25, -0.25, 0.1, cirq.Symbol('s')])
+    cirq.testing.assert_has_consistent_apply_unitary_for_various_exponents(
+        cirq.CZ)
 
 
 def test_z_init():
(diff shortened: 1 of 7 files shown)


cirq/ops/common_gates_test.py
```diff
@@ -74,9 +74,8 @@ def test_cz_matrix():
                                  [0, 0, 1, 0],
                                  [0, 0, 0, -1j]]))
 
-    cirq.testing.assert_apply_unitary_to_tensor_is_consistent_with_unitary(
-        val=cirq.CZ,
-        exponents=[1, -0.5, 0.5, 0.25, -0.25, 0.1, cirq.Symbol('s')])
+    cirq.testing.assert_has_consistent_apply_unitary_for_various_exponents(
+        cirq.CZ)
```
kevinsung: I think it's confusing that these tests don't use `assert_eigen_gate_has_consistent_apply_unitary`, since they are EigenGates after all. I think we should get rid of all the overriding `__init__` methods on the Pow gates and just have the one test method.

cirq/testing/circuit_compare.py
```diff
@@ -245,6 +245,113 @@ def assert_has_diagram(
...
+
+    eye = np.eye(2 << n, dtype=np.complex128).reshape((2,) * (2 * n + 2))
+    actual = protocols.apply_unitary_to_tensor(
+        val=val,
+        target_tensor=eye,
+        available_buffer=np.ones_like(eye) * float('nan'),
+        axes=list(range(1, n + 1)),
+        default=None)
+
+    # If you don't have a unitary, you shouldn't be able to apply a unitary.
```
kevinsung: I'm not sure I agree with this; I think it's okay to define `apply_unitary_effect_to_state` without defining `unitary`.

Strilanc: I went back and forth on this as well, but ultimately I came down on "it's easier to make it looser later than stricter later". I suspect this may be resolved by a hypothetical possibly-maybe future change that implements `_unitary_` in terms of `_apply_unitary_to_tensor_` when only the latter is present.

PR #1138, opened 2018-11-09
fraction representation of exponents for circuit diagrams

### LINKED ISSUE #1098: Fractional exponents in diagrams
What do people think about replacing decimal fractions with ordinary fractions in diagrams (in cases where it is possible), e.g. displaying `X^(7/8)` rather than `X^0.875`?

An important property of quantum gates is that most of them are elements in finite [cyclic groups](https://en.wikipedia.org/wiki/Cyclic_group) and this is better exhibited by `X^(1/8)` than `X^0.125`. For example, the order of the minimal group containing the gate is clear from notation as it is equal to the denominator.

MichaelBroughton:
fixes #1098 . The only real logic change here is in `circuit.py`. The rest of changes are just modifying tests.


docs/tutorial.md
```diff
@@ -380,23 +380,23 @@ circuit.append(one_step(h, jr, jc, 0.1, 0.2, 0.3),
...
+#                             │       │       │
+# (1, 0): ───X^1/10───────────@^3/10──┼───────┼───────@───────X───────────────────────────@───────X───────────
+#                                     │       │       │                                   │
+# (1, 1): ───X^1/10───Z^1/5───────────@^3/10──┼───────┼───────X───────@───────X───X───────@^3/10──X───@───────
+#                                             │       │               │                               │
+# (1, 2): ───X^1/10───Z^1/5───────────────────@^3/10──┼───────@───────┼───────────────────────────────@^3/10──
+#                                                     │       │       │
+# (2, 0): ───X^1/10───────────────────────────────────@^3/10──┼───────┼───────────@───────────────────────────
+#                                                             │       │           │
+# (2, 1): ───X^1/10───Z^1/5───────────X───────────────────────┼───────@^3/10──X───@^3/10──@───────────────────
```
Strilanc: We should use `0.3` instead of `3/10`, since it's shorter and just as precise.

MichaelBroughton:
I can for sure add the parenthesis. When I was writing this I figured since the precision defaults to 3 in a lot of cases 2**(3 + 1) = 16 seemed fine. In one of the test cases 0.12343 -> 1/8. if precision was 4 it would stay 1/8, at precision = 5 it would turn into 5/57 as you go up in precision the fractions get progressively uglier, only at precision = 16 does it get as ugly as `6712/54379`. Are you sure you want this behavior that is always fixed at 16 ? (I'm fine either way)

Strilanc:
Yes, I would prefer to start with it fixed at 16. We can increase it later if it feels appropriate.

MichaelBroughton:
All set to 16

PR #1140, opened 2018-11-09
Bring back __init__ for compited protobufs

viathor:


cirq/ops/pauli_interaction_gate.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Hashable, List, Optional, Sequence, Tuple, Union
```
Strilanc: This wasn't causing a lint error before, *and* doesn't cause a lint error now?

viathor: It did cause a lint error when I ran pylint locally (hence this change):

```
$ ./check/pylint 
Using config file /usr/local/google/home/viathor/src/Cirq/dev_tools/conf/.pylintrc
************* Module cirq.ops.pauli_interaction_gate
W: 15, 0: Unused Optional imported from typing (unused-import)
$
```

Didn't investigate how this got into the repo in the first place.

PR #1141, opened 2018-11-10
Make bash_scripts_test more tolerant to different git tool versions

mrwojtek:
On my machine the `test_pytest_changed_files_file_selection` was failing because command added whitespaces before counts ("`Found        0 differing`").

The `test_pytest_and_incremental_coverage_branch_selection` has additional usage lines.


dev_tools/bash_scripts_test.py
```diff
@@ -283,7 +300,8 @@ def test_pytest_and_incremental_coverage_branch_selection():
     assert result.out == (
         'INTERCEPTED python '
         'dev_tools/run_pytest_and_incremental_coverage.py HEAD\n')
-    assert result.err == "Comparing against revision 'HEAD'.\n"
+    assert result.err.split('\n', 1)[0].split() == (
```
Strilanc: Can we `strip()` instead of splitting?

mrwojtek: I looked into that and I don't think so, I get multiline string like 

Comparing against revision 'HEAD'.\n
Usage ...

and I need the first line somehow, `strip()` takes care only about leading and trailing characters.

Strilanc: What is the exact output you are seeing?

What version of git are you using?

mrwojtek: I see exactly:
> Comparing against revision 'HEAD'.
> usage: find [-H | -L | -P] [-EXdsx] [-f path] path ... [expression]
>        find [-H | -L | -P] [-EXdsx] -f path [path ...] [expression]
> 
and
> $ git --version
> git version 2.19.0.605.g01d371f741-goog

mrwojtek: You're right, that was actually an error in the bash script. I didn't check what is actually going on here...

mrwojtek: Fixed

dev_tools/bash_scripts_test.py
```diff
@@ -78,28 +78,31 @@ def test_pytest_changed_files_file_selection():
...
     result = run(script_file='check/pytest-changed-files',
                  arg='HEAD~1',
                  setup='touch file_test.py\n'
                        'git add -A\n'
                        'git commit -m test --quiet --no-gpg-sign\n')
     assert result.exit_code == 0
-    assert result.out == 'INTERCEPTED pytest file_test.py\n'
-    assert result.err == ("Comparing against revision 'HEAD~1'.\n"
-                          "Found 1 differing files with associated tests.\n")
+    assert result.out.split() == 'INTERCEPTED pytest file_test.py\n'.split()
```
Strilanc: Is `split` still required here?

mrwojtek: No it is not - I did it for consistency. I reverted it back to keep the minimal changes required for this fix.

dev_tools/bash_scripts_test.py
```diff
@@ -88,8 +89,9 @@ def test_pytest_changed_files_file_selection():
                        'git commit -m test --quiet --no-gpg-sign\n')
     assert result.exit_code == 0
     assert result.out == 'INTERCEPTED pytest file_test.py\n'
-    assert result.err == ("Comparing against revision 'HEAD~1'.\n"
-                          "Found 1 differing files with associated tests.\n")
+    assert result.err.split() == (
```
Strilanc: There are still calls to `split` present. Could you explain why these ones were kept?

mrwojtek: Yes, those checks are problematic because of different formatting I experience (split() is a wait to ignore whitespaces):
```
>       assert result.err == (
            "Comparing against revision 'HEAD~1'.\n"
            "Found 0 differing files with associated tests.\n")
E       AssertionError: assert 'Comparing ag...ated tests.\n' == 'Comparing aga...ated tests.\n'
E         Skipping 33 identical leading characters in diff, use -v to show
E           1'.
E         - Found        0 differing files with associated tests.
E         ?       -------
E         + Found 0 differing files with associated tests.

dev_tools/bash_scripts_test.py:81: AssertionError
```

Strilanc: Oh, I didn't realize split had that behavior when you didn't give it arguments. Alright, that's fine then.

PR #1147, opened 2018-11-12
Fix README not formatting on pypi

Strilanc:
- Linking to image instead of github page describing image
- Using a PNG instead of an SVG because github says the ray SVG has text type
- Fixed missing newline in README.rst

Example package on test pypi (png header image won't load before this is submitted): https://test.pypi.org/project/cirq/0.4.0.dev4220181112134512/


README.rst
```diff
@@ -1,4 +1,5 @@
-.. image:: https://github.com/quantumlib/cirq/blob/master/docs/Cirq_logo_color.svg
+.. image:: https://raw.githubusercontent.com/quantumlib/Cirq/master/docs/Cirq_logo_color.png
```
viathor: This returns 404 for me atm.

PR #1151, opened 2018-11-13
Derive cirq.YY and cirq.ZZ from cirq.XX and openfermioncirq.ZZ

Strilanc:
- Move all three into cirq/ops/parity_gates.py
- Add consistency tests
- Fix SupportsDecompose[WithQubits] not being exported to top level
- Modify consistency tests to vacuously succeed when functionality is not present


cirq/ops/parity_gates.py
```diff
@@ -0,0 +1,204 @@
...
+
+import numpy as np
+
+from cirq import protocols, linalg
+from cirq.ops import gate_features, eigen_gate
+from cirq.ops.common_gates import _rads_func_symbol
+from cirq.type_workarounds import NotImplementedType
+
+
+class XXPowGate(eigen_gate.EigenGate,
```
viathor: Do you plan to add `_apply_unitary_to_tensor_()` for XX and YY in a follow-up PR? It's going to be a little surprising that ZZ has it, but the other two gates don't.

cirq/ops/parity_gates_test.py
```diff
@@ -0,0 +1,322 @@
...
+    np.testing.assert_allclose(cirq.unitary(cirq.XX),
+                               np.array([[0, 0, 0, 1],
+                                         [0, 0, 1, 0],
+                                         [0, 1, 0, 0],
+                                         [1, 0, 0, 0]]),
+                               atol=1e-8)
+    np.testing.assert_allclose(cirq.unitary(cirq.XX**2),
+                               np.eye(4),
+                               atol=1e-8)
+    c = np.cos(np.pi/4)
```
viathor: You'd cover more possible mistakes with other angles since sin(π/4) == cos(π/4). An easy choice is something like π/8 which corresponds to an exponent that may be represented exactly, but I recommend first checking whether errors with an angle like π/7 still fit within the tolerance (and if so, let's use π/7 or similar).

cirq/ops/parity_gates_test.py
```diff
@@ -0,0 +1,322 @@
...
+    np.testing.assert_allclose(cirq.unitary(cirq.YY),
+                               np.array([[0, 0, 0, -1],
+                                         [0, 0, 1, 0],
+                                         [0, 1, 0, 0],
+                                         [-1, 0, 0, 0]]),
+                               atol=1e-8)
+    np.testing.assert_allclose(cirq.unitary(cirq.YY**2),
+                               np.eye(4),
+                               atol=1e-8)
+    c = np.cos(np.pi/4)
```
viathor: See other comment above for better angles.

cirq/protocols/decompose.py
```diff
@@ -309,7 +309,10 @@ def decompose_once_with_qubits(val: Any,
 @overload
 def decompose_once_with_qubits(val: Any,
                                qubits: Iterable['cirq.QubitId'],
-                               default: TDefault,
+                               # NOTE: should be TDefault instead of Any, but
+                               # mypy has false positive errors when setting
+                               # default to None.
```
viathor: Well, mypy may be right: TDefault is a type parameter and it may be set to a type which doesn't admit `None` value. If we do need to allow `None` then default should have type `Optional[TDefault]`. Am I missing something?

cirq/ops/parity_gates.py
Strilanc: No plans for that, because I expect it would be slower than the default `einsum` based method. ZZ is done special because it is diagonal. XX and YY aren't particularly sparse, except at exponent=1 which I don't think happens often.

cirq/protocols/decompose.py
Strilanc: If a generic type parameter only occurs for a single input, that input should be used to infer the value of the type parameter. So if I pass `None` for default, `TDefault` should become `NoneType` and if I pass `"text"` it should become `str`.

cirq/ops/parity_gates_test.py
Strilanc: Done.

cirq/ops/parity_gates_test.py
Strilanc: Done.

Issue #1137, opened 2018-11-09
Inconsistent QASM for ControlledGate(X)


kevinsung:
Executing
```
cirq.testing.assert_qasm_is_consistent_with_unitary(cirq.ControlledGate(cirq.X))
```
raises an AssertionError.

Input:
```
import cirq
import qiskit
import numpy as np
np.set_printoptions(precision=3, linewidth=150, suppress=True)

qubits = cirq.LineQubit.range(2)
op = cirq.ControlledGate(cirq.X).on(*qubits)

unitary = cirq.unitary(op)

qasm = str(cirq.QasmOutput(op, qubits[::-1], precision=10))
result = qiskit.execute(
    qiskit.load_qasm_string(qasm),
    backend=qiskit.Aer.get_backend('unitary_simulator'))
qasm_unitary = result.result().get_unitary()

print(unitary)
print()
print(qasm_unitary)
```
Output:
```
[[1.+0.j 0.+0.j 0.+0.j 0.+0.j]
 [0.+0.j 1.+0.j 0.+0.j 0.+0.j]
 [0.+0.j 0.+0.j 0.+0.j 1.+0.j]
 [0.+0.j 0.+0.j 1.+0.j 0.+0.j]]

[[ 1.-0.j -0.+0.j  0.+0.j -0.+0.j]
 [-0.+0.j  0.+0.j  0.-0.j  1.-0.j]
 [ 0.+0.j -0.+0.j  1.-0.j -0.+0.j]
 [ 0.+0.j  1.-0.j  0.+0.j -0.+0.j]]
```

kevinsung:
In the above code, if you change `qubits[::-1]` to `qubits` then the unitaries match. However, you need `qubits[::-1]` to make it work for `cirq.CNOT`.

PR #1154, opened 2018-11-14
Fix qubit ordering of 2-qubit QASM decomposition and TwoQubitMatrixGate's phase_by

Strilanc:
- Use 'assert consistent' tests on matrix gates
- Make copies of matrix gate matrices before returning them, in case callers mutate
- Fix multi-line gate `__str__` causing bad qasm output
- Fix `ControlledGate` implementing `_phase_by_` but not accounting for relative phase by removing `_phase_by_`
- Refactor `QasmTwoQubitGate` to just store a `cirq.KakDecomposition`, and give it a `_unitary_` method + equality
- Add consistency tests to controlled gate

Fixes https://github.com/quantumlib/Cirq/issues/1137


cirq/circuits/qasm_output.py
```diff
@@ -110,8 +110,8 @@ def _decompose_(self, qubits: Sequence[ops.QubitId]) -> ops.OP_TREE:
         b = self.y * -2 / np.pi + 0.5
         c = self.z * -2 / np.pi + 0.5
 
-        yield self.before0(q0)
-        yield self.before1(q1)
+        yield self.before1(q0)
```
Strilanc: The 1 matched with 0 is a bit distressing. We must have an even number of 0-1 swaps on this path somehow.

kevinsung: Yes, quite distressing. What's the problem with switching the roles of `before0` and `before1` so that 0 matches with 0 and 1 with 1 here?

kevinsung:
Please add
```
    cirq.testing.assert_qasm_is_consistent_with_unitary(
        cirq.ControlledGate(cirq.X))
```
to the test in `testing/consistent_qasm_test.py`

cirq/circuits/qasm_output.py
Strilanc: Fixed by storing the kak decomposition directly instead of splitting it apart

Strilanc:
Added a test to controlled gate asserting that it has consistent qasm, as you asked. Also caught a few other things and fixed those...

Issue #1152, opened 2018-11-13
cirq.Simulator fails when measurements are involved


Strilanc:
The issue appears to be calling append instead of extend:

`                        measurements[cast(str, gate.key)].append(*corrected)`

should be:

`                        measurements[cast(str, gate.key)].extend(corrected)`

Though we really need to test the case to catch any follow-ons.

PR #1157, opened 2018-11-14
Fix bug when multiple qubits are measured in sparse simulator

dabacon:
Doh: append(*list) works when the list is length one but not really what was wanted.

Fixes https://github.com/quantumlib/Cirq/issues/1152


dabacon:
There was a py2test failure unrelated to this pull request.

PR #1160, opened 2018-11-15
Make HPowGate's _decompose_ preserve global phase

### LINKED ISSUE #1159: Broken Hadamard gate decomposition
Steps to reproduce:

```
In [1]: import cirq                                                                                                                                                            

In [2]: q = cirq.NamedQubit('q')                                                                                                                                               

In [3]: cirq.Circuit.from_ops(cirq.decompose([cirq.H(q)]))._unitary_()                                                                                                         
Out[3]: 
array([[ 0.5+0.5j,  0.5+0.5j],
       [ 0.5+0.5j, -0.5-0.5j]])

In [4]: cirq.Circuit.from_ops([cirq.H(q)])._unitary_()                                                                                                                         
Out[4]: 
array([[ 0.70710678+0.j,  0.70710678+0.j],
       [ 0.70710678+0.j, -0.70710678+0.j]])
```

Note that exponentiating the gate to a power different from 1.0 makes this work, suggesting a special casing of a decomposition is the culprit.

This affects other gates whose decomposition includes Hadamards (e.g. iSwaps). 

There is a unit test that compares the unitary given by the gate and by its decomposition, but the assert it uses makes the comparison merely up to global phase. I think this is incorrect. Consider two qubits q0 and q1 and a circuit that applies U0 to q0 and U1 to q1. Suppose that the decomposition of U0 yields a unitary that is consistent with U0 merely up to global phase. What happens we you replace U0 with its decomposition? Well, this alters the *relative* phase between q0 and q1 producing observable effect.

viathor:
Fixes #1159. Includes the fix for the gate and for the unit test.

### DIFF
--- a/cirq/ops/common_gates.py
+++ b/cirq/ops/common_gates.py
@@ -517,6 +517,11 @@ class HPowGate(eigen_gate.EigenGate, gate_features.SingleQubitGate):
     def _decompose_(self, qubits):
         q = qubits[0]
 
+        if self._exponent == 1:
+            yield cirq.Y(q)**0.5
+            yield cirq.XPowGate(global_shift=-0.25).on(q)
+            return
+
         yield Y(q)**0.25
         yield X(q)**self._exponent
         yield Y(q)**-0.25
--- a/cirq/ops/common_gates.py
+++ b/cirq/ops/common_gates.py
@@ -517,10 +517,6 @@ class HPowGate(eigen_gate.EigenGate, gate_features.SingleQubitGate):
     def _decompose_(self, qubits):
         q = qubits[0]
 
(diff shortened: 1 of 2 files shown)


cirq/ops/common_gates.py
```diff
@@ -517,10 +517,6 @@ def _apply_unitary_to_tensor_(self,
     def _decompose_(self, qubits):
         q = qubits[0]
 
-        if self._exponent == 1:
-            yield Y(q)**0.5, X(q)
```
Strilanc: Keep this case, but change it to

`cirq.Y(q)**0.5, cirq.XPowGate(global_shift=-0.25).on(q)`

Issue #1158, opened 2018-11-15
Simulation should fall back to decomposition when unitary isn't available


Strilanc:
A user should be able to implement a gate with a `_decompose_` into other known gates, and be able to simulate it.

...I actually thought it was working this way already (oops).

dabacon:
I'm not sure I agree with this.

Already we have had a regression because xmon simulator was using KAK and causing it to be very slow.  This means for a complicated gate it will fall back on whatever decompose is when it is faster and better to just implement the unitary.

Decompose itself is also very confusing.  Do we allow passing in a different compose?  

It also feels like we are sicking everything in the capsule: why are we putting transformations on the gates into the simulator?

kevinsung:
@dabacon even if it's faster and better to actually implement the unitary, I don't think we should _require_ it of the user. Making this change won't make anything slower; it'll just give the user more options for getting a gate to be simulated.

Also, XmonSimulator already does this! I.e., the following code works, but not if you replace XmonSimulator with Simulator
```
import cirq

class MyGate(cirq.Gate):
    
    def _decompose_(self, qubits):
        a = qubits[0]
        yield cirq.H(a)
        yield cirq.X(a)
        
circuit = cirq.Circuit.from_ops(MyGate()(cirq.NamedQubit('a')))

simulator = cirq.google.XmonSimulator()
simulator.simulate(circuit)
```

Strilanc:
> why are we putting transformations on the gates into the simulator?

The user said simulate, and it is our job to do our best to get that done. More generally, we want it to be easy to define the gates you want to use and having the simulator natively understand "oh I should try to decompose" gives users an additional implementation option.

That all being said, I think that it does make sense to have a strict "give me what I understand and nothing else" simulator. That would be the xmon simulator.

dabacon:
I agree the xmon simulator shouldn't do it.

dabacon:
I think the problem here is in decompose.  Decompose is  not well defined.  There are infinite number of decompositions, why did you chose one over the other?  It seems completely arbitrary.  Already we see people getting factors of 10 worse because they are using the default decompose on methods in a particular gate setting.  This is completely unexpected for most users.  There are places where decompose is NOT the most efficient decomposition.  People ask: "should I implement unitary or decompose or both?"

[...]

PR #1162, opened 2018-11-15
Fix sparse simulator not decomposing composite gates

Strilanc:
- Also fix SupportsChannel not being exposed at top level

Fixes https://github.com/quantumlib/Cirq/issues/1158


Strilanc:
I opened https://github.com/quantumlib/Cirq/issues/1163 to solve do that more generally. For now I think we should submit this as a workaround before releasing `0.4.0`.

Strilanc:
Done.

PR #1164, opened 2018-11-16
Improve decompose documentation

Strilanc:


cirq/protocols/decompose.py
```diff
@@ -44,20 +44,28 @@ def _value_error_describing_bad_operation(op: 'cirq.Operation') -> ValueError:
...
-    decomposition should move closer towards the basic 1-qubit and 2-qubit gates
-    included by default in cirq. It is not necessary to get all the way there,
-    just closer (callers will iteratively decompose if needed).
-
-    DECOMPOSITIONS MUST NEVER BE CYCLIC. If B has A in its decomposition, A
-    should not have B in its decomposition. Decomposition is often performed
-    iteratively, until a fixed point or some desired criteria is met. Cyclic
-    decompositions cause that kind of code to loop forever.
+    All decompositions methods should ultimately terminate on basic 1-qubit and
+    2-qubit gates included by default in cirq. For example, it is permitted for
```
dabacon: Which one and two qubit gates?  Clifford Gates?  When we add exponetiated Paulis, those?  What about gates that end up in contrib (which will happen)?

This is not a forward safe definition.

Strilanc: Any of those would be fine. Internally we will have some unique end state, but externally people just need to know "get to the internal types".

You might be thinking in terms of the users of `cirq.decompose`? I guess I could leave some phrasing about not waiting for specific gates but instead waiting for e.g. any unitary 2-qubit gate. We could vary the final gates from release to release to really nail this home for people, or just use TwoQubitMatrixGate as the end point.

docs/gates.md
```diff
@@ -28,6 +28,37 @@ Or, if a gate specifies a `__pow__` method that works for an exponent of -1, the
...
+The main requirements on the output of `_decompose_` methods are:
+
+1. DO NOT CREATE CYCLES. The `cirq.decompose` method will iterative decompose until it find values satisfying a `keep` predicate. Cycles cause it to enter an infinite loop.
+2. Head towards operations defined by cirq, because these operations have good decomposition methods that terminate in single-qubit and two-qubit gates that can be understood by essentially all code.
+3. All that matters is functional equivalence. Don't worry about staying within or reaching a particular gate set; it's too hard to predict what the caller will want. Gate-set-aware decomposition is useful, but *this is not the protocol that does that*. 
+
+For example, `cirq.CCZ` decomposes into a series of `cirq.CNOT` and `cirq.T` operations.
+This allows code that doesn't understand three-qubit operation to work with `cirq.CCZ`; by decomposing it into operations they do understand.
+As another example, `cirq.TOFFOLI` decomposes into a `cirq.H` followed by a `cirq.CCZ` followed by a `cirq.H`.
+Although the output contains a three qubit operation (the CCZ), that operation can be decomposed into two qubit and one qubit operations.
```
viathor: nit: consistency in hyphenation: s/one qubit/one-qubit/ same for two-qubit.

docs/gates.md
```diff
@@ -28,6 +28,37 @@ Or, if a gate specifies a `__pow__` method that works for an exponent of -1, the
...
+So code that doesn't understand three qubit operations can deal with Toffolis by decomposing it, and then decomposing the CCZ from the decomposition.
```
viathor: nit: consistency in (grammatical) number:

```
...with Toffolis by decomposing it,...
```

should be

```
...with Toffolis by decomposing them,...
```

PR #1165, opened 2018-11-16
Add and use functions to test consistency of protocol implementations

### LINKED ISSUE #1053: Add a "standard test battery" for new gates, use it on old gates
There is a lot of repetition in tests between gates. This leaves a lot of room for accidental omission. Some of this repetition could be reduced by having a standard set of tests that a gate must pass, such as "the decomposition is consistent with its unitary matrix".

Here's roughly what I was thinking:

```python
def assert_implements_consistent_protocols(
        val: Any,
        *,
        qubit_count: Optional[int] = None,
        exponents: Sequence[Any] = (1,
                                    -0.5,
                                    0.5,
                                    -1,
                                    0.25,
                                    -0.25,
                                    0.1,
                                    value.Symbol('s'))):
    """Checks that a value is internally consistent and has a good __repr__."""

    _assert_gate_meets_standards_helper(val, qubit_count)

    for exponent in exponents:
        p = protocols.pow(val, exponent, None)
        if p is not None:
            _assert_gate_meets_standards_helper(val**exponent, qubit_count)


def _assert_gate_meets_standards_helper(val: Any, qubit_count: Optional[int]):
    import cirq.testing  # HACK: actually should do file-level imports

    cirq.testing.assert_equivalent_repr(val)
    cirq.testing.assert_decompose_is_consistent_with_unitary(val)
    cirq.testing.assert_phase_by_is_consistent_with_unitary(val)
    cirq.testing.assert_apply_unitary_to_tensor_is_consistent_with_unitary(
        val, qubit_count=qubit_count)
```

This issue is to cover perhaps a few more cases, and to then incorporate this into the `cirq.testing` namespace

kevinsung:
Fixes #1053


Strilanc:
```
def _numpy_array_repr(arr: np.ndarray) -> str:
    return 'np.array({!r})'.format(arr.tolist())
```

use that for the numpy arrays

cirq/ops/common_gates_test.py
```diff
@@ -26,6 +26,27 @@
...
+@pytest.mark.parametrize('eigen_gate_type', [
+    cirq.CNotPowGate,
+    cirq.CZPowGate,
+    cirq.HPowGate,
+    cirq.ISwapPowGate,
+    cirq.SwapPowGate,
+    cirq.XPowGate,
+    cirq.YPowGate,
+    cirq.ZPowGate,
+    ]
```
Strilanc: Nit: deindent the square bracket, put the close parens on the same line

cirq/ops/common_gates_test.py
```diff
@@ -38,18 +59,7 @@ def test_cz_str():
     assert str(cirq.CZ**-0.25) == 'CZ**-0.25'
 
 
-def test_cz_repr():
```
Strilanc: I wouldn't get rid of thi  particular repr test, because it is asserting more than just consistency. It's asserting that you get a *simple* repr for these cases, instead of the fully detailed one.

cirq/ops/parity_gates.py
```diff
@@ -110,7 +110,7 @@ def _eigen_components(self):
         ]
 
     def _eigen_shifts(self):
-        return 0, 1
+        return [0, 1]
```
Strilanc: Why the square brackets?

cirq/testing/circuit_compare.py
```diff
@@ -364,6 +366,9 @@ def _infer_qubit_count(val: Any) -> int:
         return 3
     if isinstance(val, ops.ControlledGate):
         return 1 + _infer_qubit_count(val.sub_gate)
+    if protocols.has_unitary(val):
+        return protocols.unitary(val).shape[0].bit_length() - 1
```
Strilanc: Use `default` argument of unitary instead of calling `has` then `unitary`.

cirq/testing/consistent_decomposition.py
```diff
@@ -37,4 +39,6 @@ def assert_decompose_is_consistent_with_unitary(val: Any):
     actual = circuits.Circuit.from_ops(dec).to_unitary_matrix(
         qubit_order=qubits)
 
-    assert np.allclose(actual, expected, atol=1e-8)
+    lin_alg_utils.assert_allclose_up_to_global_phase(actual,
```
Strilanc: Add an `including_global_phase` boolean parameter that determines whether this or `np.allclose` is used.

cirq/ops/parity_gates.py
kevinsung: The return type is List[int]

kevinsung: So otherwise the consistent eigenshifts tests would fail

cirq/testing/circuit_compare.py
kevinsung: I realized this is actually unnecessary because because the function that uses this helper method already checks that case.

cirq/ops/parity_gates.py
Strilanc: Oh, I see.

The return type should be `Sequence[int]`, or maybe even `Iterable[int]`, not `List[int]`. The receiving code shouldn't be depending on the fact that it's a list or a tuple.

Issue #1073, opened 2018-10-29
Look into recognizing and rendering latex within docstrings for generated documentation


Strilanc:
Sometimes we use latex in our docstrings. It would be ideal if it was actually rendered as latex on the documentation website.

Strilanc:
Fixed by https://github.com/quantumlib/Cirq/pull/1184

PR #1167, opened 2018-11-16
First pass for implementing density matrices and bloch vectors.

### LINKED ISSUE #1094: Add 'trace' utility methods to TrialResult and similar classes
I think users of `TrialResult` will often have questions like "what's the bloch vector of qubit 5?" and "if I trace out qubit 4, what's the remaining density matrix?". We should have methods for these kinds of questions, and others.

MichaelBroughton:
fixes #1094 . Please double check my tests.


cirq/sim/simulator.py
```diff
@@ -525,3 +530,48 @@ def dirac_notation(self, decimals=2):
...
+        """Returns the sparse density matrix of the wavefunction.
+
+        Calculate the density matrix for the system on the given qubit
+        indices, with the qubits not in indices that are present in state
+        traced out. If indices is None the full density matrix for state
+        is returned.
+        For example:
+            state = np.array([1/np.sqrt(2), 1/np.sqrt(2)], dtype=np.complex64)
+            indices = None
+            gives us \rho = \begin{bmatrix}
```
vtomole: Is this Latex automatically rendered? Looking at: https://github.com/quantumlib/Cirq/issues/1073, it seems like there are a couple of doc strings in the repo that have Latex in them. An alternative is to type out the matrix similar to https://github.com/quantumlib/Cirq/blob/master/cirq/ops/parity_gates.py#L34; make it an easier read. I should add a comment on that issue, but @Strilanc, what is the plan on having Latex in the doc strings for people who won't generate the documentation to have it rendered?

Strilanc:
@vtomole I like it when people use latex, because it is very unambiguous and it opens the possibility of us later rendering it. Not enough to make people use it, but enough that I don't get rid of it when it's there.

cirq/sim/wave_function.py
```diff
@@ -270,6 +339,23 @@ def _probs(state: np.ndarray, indices: List[int],
     return probs
 
 
+def _single_partial_trace(density_matrix: np.ndarray,
+    n_qubits: int,  index: int) -> np.ndarray:
+    """Helper function to trace out one index from a density matrix."""
+    n_left = index
```
Strilanc: I think it's possible to specify this succinctly by using einsum.

I believe the following expression computes the two-qubit density matrix of the second qubit and fifth qubits of a five-qubit state vector v:

```python
v = v.reshape((2,) * 5)
#                 v lower case the included qubits
d = np.einsum('ABCDE,AbCDe->BEbe', v, np.conj(v))
#                         ^ list both cases of included qubits
return d.reshape(4, 4)
```

cirq/sim/simulator.py
```diff
@@ -525,3 +530,43 @@ def dirac_notation(self, decimals=2):
...
+        is returned.
+        For example:
+            self.state = np.array([1/np.sqrt(2), 1/np.sqrt(2)],
+                dtype=np.complex64)
+            indices = None
+            gives us \rho = \begin{bmatrix}
+                                0.5 & 0.5
+                                0.5 & 0.5
+                            \end{bmatrix}
+
```
dabacon: should document that it follows the np.kron order in relation to order of indices.

cirq/sim/wave_function.py
```diff
@@ -26,6 +26,75 @@
...
+        return np.outer(np.conj(state), state)
+
+    all_indices = [i for i in range(n_qubits)]
+    _validate_indices(n_qubits, indices)
+
+    to_trace_out = sorted(list(set(all_indices).difference(indices)))
+    rho = np.outer(np.conj(state), state)
+
+    num_traced = 0
+    for index in to_trace_out:
```
dabacon: I wonder if np.einsum can do this faster.

cirq/sim/wave_function.py
```diff
@@ -26,6 +26,76 @@
     from cirq.sim import simulator
 
 
+def bloch_vector(state: Sequence, index: int) -> np.ndarray:
```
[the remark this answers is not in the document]
MichaelBroughton: fixed

cirq/sim/wave_function.py
```diff
@@ -26,6 +26,76 @@
...
+    rho = density_matrix(state, [index])
+    v = np.zeros(3, dtype=np.complex64)
+    v[0] = rho[1][0] + rho[0][1]
+    v[1] = 1.0j * (rho[0][1] - rho[1][0])
+    v[2] = rho[0][0] - rho[1][1]
+
+    return v
+
+
+def density_matrix(state: Sequence, indices: List[int] = None) -> np.ndarray:
```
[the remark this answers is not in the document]
MichaelBroughton: fixed

cirq/sim/wave_function.py
```diff
@@ -270,6 +340,23 @@ def _probs(state: np.ndarray, indices: List[int],
     return probs
 
 
+def _single_partial_trace(density_matrix: np.ndarray,
+    n_qubits: int,  index: int) -> np.ndarray:
```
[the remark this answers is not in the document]
MichaelBroughton: removed this in favor of your mind blowing einsum formulation

cirq/sim/simulator.py
```diff
@@ -250,6 +250,12 @@ def bitstring(vals):
     def dirac_notation(self, decimals=2):
         return wave_function.dirac_notation(self.final_state, decimals)
 
+    def density_matrix(self, indices=None):
```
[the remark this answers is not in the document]
MichaelBroughton: fixed

cirq/sim/simulator.py
```diff
@@ -250,6 +250,12 @@ def bitstring(vals):
...
+        return wave_function.density_matrix(self.final_state, indices)
+
+    def bloch_vector(self, index):
```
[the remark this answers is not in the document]
MichaelBroughton: fixed

cirq/sim/simulator.py
```diff
@@ -525,3 +530,43 @@ def dirac_notation(self, decimals=2):
             A pretty string consisting of a sum of computational basis kets
             and non-zero floats of the specified accuracy."""
         return wave_function.dirac_notation(self.state, decimals)
+
+    def density_matrix(self, indices=None):
```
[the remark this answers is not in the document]
MichaelBroughton: fixed

cirq/sim/simulator.py
```diff
@@ -525,3 +530,43 @@ def dirac_notation(self, decimals=2):
...
+            indices: list containing indices for qubits that you would like
+                to include in the density matrix (i.e.) qubits that WON'T
+                be traced out.
+
+        Returns:
+            A numpy array representing the density matrix.
+        """
+        return wave_function.density_matrix(self.state, indices)
+
+    def bloch_vector(self, index):
```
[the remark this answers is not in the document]
MichaelBroughton: fixed

cirq/sim/wave_function.py
```diff
@@ -26,6 +26,75 @@
...
+        state: A sequence representing a wave function in which
+            the ordering mapping to qubits follows the standard Kronecker
+            convention of numpy.kron.
+        index: index of qubit who's bloch vector we want to find.
+
+    Returns:
+        A length 3 numpy array representing the qubit's bloch vector.
+    """
+    rho = density_matrix(state, [index])
+    v = np.zeros(3, dtype=np.complex64)
```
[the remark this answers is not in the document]
MichaelBroughton: fixed

cirq/sim/wave_function.py
```diff
@@ -26,6 +26,75 @@
...
+    v[0] = rho[1][0] + rho[0][1]
```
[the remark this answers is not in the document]
MichaelBroughton: fixed

cirq/sim/wave_function.py
```diff
@@ -26,6 +26,75 @@
...
+    v[1] = 1.0j * (rho[0][1] - rho[1][0])
```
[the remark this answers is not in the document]
MichaelBroughton: double checked and I THINK it's `rho[1][0]`

cirq/sim/wave_function.py
```diff
@@ -26,6 +26,75 @@
...
+    v[2] = rho[0][0] - rho[1][1]
```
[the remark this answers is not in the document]
MichaelBroughton: fixed

cirq/sim/wave_function.py
MichaelBroughton: I put this in, in favor of what I had before.

cirq/sim/simulator.py
MichaelBroughton: fixed

cirq/sim/wave_function.py
MichaelBroughton: The above np.einsum was a fair bit faster than my original logic.

MichaelBroughton:
The `np.einsum` logic limits the user to get the density matrix for a wavefunction on at most 26 qubits (after that we run out of letters in the alphabet) are people ok with this ?

cirq/sim/wave_function.py
```diff
@@ -26,6 +28,100 @@
...
+            with np.einsum.")
+
+    if indices is None:
+        return np.outer(state, np.conj(state))
+
+    _validate_indices(n_qubits, indices)
+
+    state = np.asarray(state).reshape((2,)*n_qubits)
+
+    upper_str = np.array([i for i in string.ascii_uppercase[:n_qubits]])
```
Strilanc: Let's fix the limitation to 26 by passing in tuples of ints instead of a string, and using only the integers we need.

Use the indices in 0..n to indicate the input axes, but instead of using n..2n for the kept axes, only use n..n+k where k is the number of kept axes. Something like:

```python
k = len(indices)
n = n_qubits
inp = list(range(n))
inp2 = list(inp)
next_index = n
for i in indices:
    inp2[i] = next_index
    next_index += 1
kept = list(range(n, n+k))
out = indices + kept
```

I probably got a bunch of ordering related things backwards in that code. Hopefully it gets across the general idea.

This will replace the qubits <= 26 constraint with qubits + kept <= 52, which is more reasonable. I think we can reasonably assume that kept is less than 17 and qubits is less than 34, since those both require 128 GiB of memory to store and it would take hours to actually extract the value. And even if computers got a lot faster in the far future, numpy would remove the constraint and we'd inherit that.

I wouldn't even bother checking if the user is over the limit. Just let numpy fail when they ask for something ridiculous, so they can complain to the right person.

MichaelBroughton: Thank you for the feedback, I've updated the logic now using integer indexing in einsum, what do you think ?

cirq/sim/wave_function.py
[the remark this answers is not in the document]
Strilanc: Okay.

You can use Quirk as a source of independent state-or-circuit-to-bloch-vector values. For example, according to Quirk, the bloch vector of a qubit that started in the 0 state and then had an X**(1/4) operation applied is x=0,y=-sqrt(0.5),z=sqrt(0.5).

image

cirq/sim/simulator.py
Strilanc: I think our audience is technical enough on average that we can expect them to read latex notation more precisely than whatever arbitrary text grid we use. So I'm fine with keeping it.

cirq/sim/wave_function_test.py
```diff
@@ -25,6 +25,102 @@ def assert_dirac_notation(vec, expected, decimals=2):
     assert cirq.dirac_notation(np.array(vec), decimals=decimals) == expected
 
 
+def test_bloch_vector_simple():
+    sqrt = np.sqrt(0.5)
+    test_state = np.array([sqrt, sqrt])
+
+    bloch = cirq.bloch_vector_from_state_vector(test_state, 0)
```
Strilanc: Test that the output is correct for the |0>+|1> state, the |0>+i|1> state, the |T> state, and the state that should produce the bloch vector x=y=z=sqrt(1/3).

cirq/sim/wave_function_test.py
```diff
@@ -25,6 +25,102 @@ def assert_dirac_notation(vec, expected, decimals=2):
...
+    np.testing.assert_array_almost_equal(bloch_0, desired_simple)
+
+
+def test_bloch_vector_multi_mixed():
+    sqrt = np.sqrt(0.5)
+    test_state = np.array([sqrt, 0., 0., sqrt])
+
+    bloch_0 = cirq.bloch_vector_from_state_vector(test_state, 0)
+    bloch_1 = cirq.bloch_vector_from_state_vector(test_state, 1)
+    zero = np.zeros(3)
```
Strilanc: Do a test involving five qubits, just to make sure the not-start-not-end case works.

cirq/sim/wave_function.py
```diff
@@ -26,6 +26,93 @@
...
+    v = np.zeros(3, dtype=np.float32)
+    v[0] = 2*np.real(rho[0][1])
+    v[1] = 2*np.imag(rho[1][0])
+    v[2] = np.real(rho[0][0] - rho[1][1])
+
+    return v
+
+
+def density_matrix_from_state_vector(state: Sequence,
+    indices: List[int] = None) -> np.ndarray:
```
Strilanc: nit: align indentation

cirq/sim/wave_function.py
```diff
@@ -26,6 +26,93 @@
...
+    n_qubits = _validate_num_qubits(state)
+
+    if indices is None:
+        return np.outer(state, np.conj(state))
+
+    _validate_indices(n_qubits, indices)
+
+    state = np.asarray(state).reshape((2,)*n_qubits)
+
+    sum_inds = np.array([i for i in range(n_qubits)], dtype=int)
```
Strilanc: This is equivalent to `sum_inds = np.array(range(n_qubits))`

cirq/sim/wave_function.py
```diff
@@ -26,6 +26,93 @@
...
+    sum_inds[indices] = (sum_inds + n_qubits)[indices]
```
Strilanc: This will fail if the user passes in a tuple or other value instead of a list. Convert `indices` to a list at the start of the method, and loosen the argument type from `List[int]` to `Iterable[int]`.

cirq/sim/wave_function.py
```diff
@@ -26,6 +26,93 @@
...
+
+    rho = np.einsum(state, [i for i in range(n_qubits)], np.conj(state),
```
Strilanc: `[i for i in range(n_qubits)]` is equivalent to `list(range(n_qubits))`

cirq/sim/wave_function.py
```diff
@@ -26,6 +26,93 @@
...
+        sum_inds.tolist(), indices + sum_inds[indices].tolist())
```
Strilanc: Is the `tolist` required? I'd have expected numpy to be happy with an ndarray.

cirq/sim/wave_function.py
Strilanc: `sum_inds[indices] += n_qubits`

cirq/sim/wave_function_test.py
```diff
@@ -25,32 +25,62 @@ def assert_dirac_notation(vec, expected, decimals=2):
...
-        _ = cirq.bloch_vector_from_state_vector(
-            np.array([0.5, 0.5, 0.5]), 0)
-    with pytest.raises(IndexError):
-        _ = cirq.bloch_vector_from_state_vector(
-            np.array([0.5, 0.5,0.5,0.5]), -1)
-    with pytest.raises(IndexError):
-        _ = cirq.bloch_vector_from_state_vector(
-            np.array([0.5, 0.5,0.5,0.5]), 2)
+def test_bloch_vector_simple_XH_zero():
+    XH_state = np.array([0.70710678, 0.70710678])
```
Strilanc: use `np.sqrt(0.5)` instead of the written out value

cirq/sim/wave_function_test.py
```diff
@@ -25,32 +25,62 @@ def assert_dirac_notation(vec, expected, decimals=2):
...
+    bloch = cirq.bloch_vector_from_state_vector(XH_state, 0)
+
+    desired_simple = np.array([1,0,0])
+    np.testing.assert_array_almost_equal(bloch, desired_simple)
+
+
+def test_bloch_vector_simple_YH_zero():
+    YH_state = np.array([0.-0.70710678j, 0.+0.70710678j])
```
Strilanc: `np.sqrt(0.5)`

cirq/sim/wave_function_test.py
```diff
@@ -25,32 +25,62 @@ def assert_dirac_notation(vec, expected, decimals=2):
...
+    bloch = cirq.bloch_vector_from_state_vector(YH_state, 0)
+
+    desired_simple = np.array([-1,0,0])
+    np.testing.assert_array_almost_equal(bloch, desired_simple)
+
+
+def test_bloch_vector_simple_ZH_zero():
+    ZH_state = np.array([0.70710678, -0.70710678])
```
Strilanc: `np.sqrt(0.5)`

cirq/sim/wave_function_test.py
```diff
@@ -25,32 +25,62 @@ def assert_dirac_notation(vec, expected, decimals=2):
...
+    bloch = cirq.bloch_vector_from_state_vector(ZH_state, 0)
+
+    desired_simple = np.array([-1,0,0])
+    np.testing.assert_array_almost_equal(bloch, desired_simple)
+
+
+def test_bloch_vector_simple_TH_zero():
+    sqrt = np.sqrt(0.5)
```
Strilanc: You can combine all of these tests into one parameterized test with state+blochvector arguments if you prefer.

cirq/sim/wave_function_test.py
```diff
@@ -25,32 +25,62 @@ def assert_dirac_notation(vec, expected, decimals=2):
...
+    TH_state = np.array([0.70710678+0.j , 0.5       +0.5j])
+    bloch = cirq.bloch_vector_from_state_vector(TH_state, 0)
+
+    desired_simple = np.array([sqrt,sqrt,0])
+    np.testing.assert_array_almost_equal(bloch, desired_simple)
+
+
+def test_bloch_vector_equal_sqrt3():
+    sqrt3 = 1/np.sqrt(3)
+    test_state = np.array([0.888074, 0.325058 + 0.325058j])
```
Strilanc: Good job finding the state

Strilanc:
Looks like you forgot to import `Type` from typing.

MichaelBroughton:
looks like it haha. It's not immediately clear to me how to fix the `Type` problem from travis, do you know how ?

Strilanc:
In `wavefunction.py` you deleted `Type` from the `typing` import list. Put it back and it should fix the error.

cirq/sim/wave_function_test.py
MichaelBroughton: fixed

cirq/sim/wave_function_test.py
MichaelBroughton: also fixed

cirq/sim/wave_function_test.py
MichaelBroughton: fixed

cirq/sim/wave_function_test.py
MichaelBroughton: I'm fine leaving them as is

MichaelBroughton:
Oops I must have missed that when I merged a little earlier on. Thanks!

PR #1174, opened 2018-11-19
Extract class equality protocol out of value equality.

mrwojtek:
Primarily exposes cirq.class_equals method which works on classes that support `_SupportsClassEquality` protocol that is implemented through `@cirq.class_equality decorator`. This functionality was already present but it was internal to `@cirq.value_equality` decorator.

This refactoring allows the `cirq.class_equals` to be used as an utility function in other comparison operators (like approximate equality) which is external to @cirq.value_equality and custom to the implementing class.

Note that in the process of refactoring an `isinstance` check was dropped. This makes equality comparator symmetric.

### DIFF
--- a/cirq/__init__.py
+++ b/cirq/__init__.py
@@ -219,6 +219,8 @@ from cirq.value import (
     canonicalize_half_turns,
     chosen_angle_to_canonical_half_turns,
     chosen_angle_to_half_turns,
+    class_equality,
+    class_equals,
     Duration,
     Symbol,
     Timestamp,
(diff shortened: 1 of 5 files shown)


mrwojtek:
That's also an option.

I would still like to define `cirq.protocols.SupportsIsClose` protocol with `cirq.protocols.isclose()` comparison method so that other classes (the ones which are not decorated by @value_equality) could also support this feature.
 
We can make value_equality automatically implement `_isclose_`. This will require a new dependency in `_SupportsValueEquality`, something like `_approximate_equals_(other) -> bool` which does the appropriate comparison inside. This method should be symmetric.

mrwojtek:
Moving to different approach (PR 1194 and 1223)

PR #1175, opened 2018-11-20
Install contrib dependencies during package verification

viathor:
Also, actually install dev deps, too. We need the dependencies in this environment since all tests (including those for code in contrib/) are run in it.


viathor:
Thanks for catching this.

I'm leery of distributing code without first running tests that come with it whether it's main package or contrib. I moved contrib deps installation and testing steps to after cirq tests finished. PTAL

PR #1176, opened 2018-11-20
Reorder unitary instead of reversing qubits in QASM tests

kevinsung:


cirq/circuits/qasm_output_test.py
```diff
@@ -470,3 +474,18 @@ def test_qasm_two_qubit_gate_unitary():
     u = cirq.testing.random_unitary(4)
     g = QasmTwoQubitGate.from_matrix(u)
     np.testing.assert_allclose(cirq.unitary(g), u)
+
+
+def _reorder_indices_of_matrix(matrix: np.ndarray, new_order: List[int]):
```
Strilanc: `einsum` should be ideal for this.

```python
reshaped_input = input.reshape((2,) *num_qubits)
output = np.einsum(list(range(num_qubits)), new_order, reshaped_input)
return output.reshape(input.shape)
```

PR #1177, opened 2018-11-20
Use QASM protocol in QASM tests

### LINKED ISSUE #1166: QASM test method doesn't use QASM protocol
The test method `assert_qasm_is_consistent_with_unitary` doesn't use the `protocols.qasm` method. Instead, it uses the line
```
qasm = str(circuits.QasmOutput(op, qubits[::-1], precision=10))
```
My understanding is that the reason `protocols.qasm` isn't used is that it doesn't let us pass in qubits in the reverse order, which is necessary due to qiskit using the reverse qubit ordering convention to convert a QASM to a unitary.
Blocking #1165

kevinsung:
Fixes #1166 .


kevinsung:
Fixed

PR #1178, opened 2018-11-20
Refactor `_apply_unitary_to_effect_` into `_apply_unitary_`

Strilanc:
- Combine arguments into an args object (for forwards compatibility and easier-to-remember signature)
- Add a `subspace_index` helper method to the args object
- `None` is now a valid value for indicating that the current value can't `_apply_unitary_` (equivalent to `NotImplemented`)
- Fixed a few type annotations along the way


cirq/protocols/apply_unitary.py
```diff
@@ -0,0 +1,238 @@
...
+        The receiving object may also write its output over a new buffer
+        that it created, in which case that new array is returned.
+
+    Raises:
+        TypeError: `unitary_value` doesn't have a unitary effect and `default`
+            wasn't specified.
+    """
+
+    # Check if the specialized method is present.
+    getter = getattr(unitary_value, '_apply_unitary_', None)
```
viathor: nit: `getter` is a poor name here (it suggests an existing value is accessed, in reality a new value is likely being computed), suggestion:

```python
apply = getattr(unitary_value, '_apply_unitary_', None)
if apply is not None:
    result = apply(args)
    ...
```

cirq/protocols/apply_unitary.py
```diff
@@ -0,0 +1,238 @@
...
+            that the result is in `args.available_buffer` (and so what was
+            `args.available_buffer` will become `args.target_tensor` in the next
+            call, and vice versa).
+
+            The receiving object is also permitted to allocate a new
+            numpy.ndarray and return that as its result.
+        """
+
+
+def apply_unitary(unitary_value: Any,
```
viathor: nit: possibly better name: `unitary`.

viathor: Clarification: s/unitary_value/unitary/

cirq/protocols/apply_unitary.py
```diff
@@ -0,0 +1,238 @@
...
+            unitary transformation or some other concept.
+        available_buffer: Pre-allocated workspace with the same shape and
+            dtype as the target tensor.
+        axes: Which axes the unitary effect is being applied to (e.g. the
+            qubits that the gate is operating on).
+    """
+
+    def __init__(self,
+                 target_tensor: np.ndarray,
+                 available_buffer: np.ndarray,
```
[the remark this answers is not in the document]
Strilanc: Yes. Actually, `cirq.apply_unitary` should probably allow a calling style where you say `state: Union[int, np.ndarray]`. But it's a bit messy, so I deferred it.

cirq/protocols/apply_unitary.py
Strilanc: Acknowledged. `unitary` would sound like it was asking for the matrix, I think. Left as is.

cirq/protocols/apply_unitary.py
Strilanc: Renamed to `func`.

Issue #95, opened 2018-02-13
Implement a `GraphQubit` type


dabacon:
Right now QubitLoc is  just a 2d point.  This is very reasonable for our hardware, but there are other qubit types that are very natural and which we should support (lines, staggered grids, general graph connectivities).  

Proposal:
Abstract out the interface that describes adjacency, i.e. make an interface type HasAdjacency which requires the method is_adjacent.  (Right  constant two qubit gate assumes that the qubits implement this method).

New Qubit types
GridQubit
LineQubit

Also a way to go from an Adjacency graph to a set of GraphQubits which implement HasAdjacency and  which obey the adjacency relation of the graph.

This would mostly be relevant for defining a single / two qubit gates that respect adjacency, and also for device which use this adjacency in some of their constraints.

bryano:
It seems more natural for adjacency and gate information to be part of the device rather than the qubit. Why not just give GraphDevice a graph attribute whose vertices can be generic `QubitId`s?

vtomole:
@bryano Is this related to all of your not-yet-merged pull requests?

bryano:
Yea, specifically #1181

bryano:
Can this be closed?

PR #1181, opened 2018-11-21
Added GraphDevice type

bryano:
Goal is two-fold: simplify device creation and expose more structure to compilers.


cirq/devices/graph_device.py
```diff
@@ -0,0 +1,230 @@
...
+# limitations under the License.
+
+import abc
+import itertools
+
+from typing import cast, FrozenSet, Hashable, Iterable, Optional
+
+from cirq import circuits, devices, ops, value, schedules
+
+from cirq.devices.hypergraph import UndirectedHypergraph
```
Strilanc: `from cirq.devices import hypergraph`

cirq/devices/graph_device.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+class HashQubit(ops.QubitId):
```
Strilanc: This definition is out of place here. Qubits are already hashable. I would just delete this class from the PR.

cirq/devices/graph_device.py
```diff
@@ -0,0 +1,230 @@
...
+        try:
+            device_edge = self.get_device_edge(operation)
+        except Exception as error:
+            vertices = self.get_vertices(operation)
+            if vertices not in self.device_graph.edges:
+                error =  ValueError('{} not in device graph edges'.format(
+                    vertices))
+            raise error
+        device_edge.validate_operation(operation)
+
```
Strilanc: Nit: spacing

cirq/devices/graph_device.py
```diff
@@ -0,0 +1,230 @@
...
+    def __init__(self,
+                 device_graph: UndirectedHypergraph,
+                 crosstalk_graph: Optional[UndirectedHypergraph]=None) -> None:
+        """
+
+        Args:
+            device_graph: An undirected hypergraph whose vertices correspond to
+                qubits and whose edges determine allowable operations and their
+                durations.
+            crosstalk_graph: An undirected hypergraph whose vertices are edges
```
Strilanc: What kind of constraints?

cirq/devices/hypergraph.py
```diff
@@ -0,0 +1,99 @@
...
+# limitations under the License.
+
+
+from typing import (
+        Any, Dict, FrozenSet, Hashable, Iterable, Optional, Set, Tuple)
+
+
+AdjacencyList = Set[FrozenSet[Hashable]]
+
+class UndirectedHypergraph:
```
Strilanc: Does `networkx` include a class for this already? Ideally we wouldn't be defining our own graph data structures.

cirq/devices/hypergraph.py
```diff
@@ -0,0 +1,99 @@
...
+            del self._labelled_edges[edge]
+            for neighbor in edge.difference((vertex,)):
+                self._adjacency_lists[neighbor].difference_update((edge,))
+        del self._adjacency_lists[vertex]
+
+
+    def remove_vertices(self, vertices):
+        for vertex in vertices:
+            self.remove_vertex(vertex)
+
```
Strilanc: one line between class methods

cirq/devices/hypergraph.py
```diff
@@ -0,0 +1,99 @@
...
+        """A labelled, undirected hypergraph.
+
+        Args:
+            vertices: The vertices.
+            labelled_edges: The labelled edges, as a mapping from (frozen) sets
+                of vertices to the corresponding labels. Vertices are
+                automatically added.
+        """
+
+        self._adjacency_lists = {} # type: Dict[Hashable, AdjacencyList]
```
Strilanc: It might be simpler to use just the edge dictionary (i.e. an edge set representation). Singleton vertices could be represented by the edge key `{node}`.

cirq/devices/graph_device.py
```diff
@@ -0,0 +1,230 @@
...
+
+def raise_crosstalk_error(*ops: ops.Operation):
+    raise ValueError('crosstalk on {}'.format(ops))
+
+
+class UndirectedGraphDevice(devices.Device):
+    """A device whose properties are represented by an edge-labelled graph.
+
+    Each (undirected) edge of the device graph is labelled by an
+    UndirectedGraphDeviceEdge or None. None indicates that any operation is
```
Strilanc: Shouldn't the edges represent capabilities, not constraints, since capabilities are often much sparser?

cirq/devices/graph_device.py
bryano: The name is terrible, but I think this is needed. The "hash" refers to the `value` method being hashable, not the qubit itself. The idea is something like `NamedQubit`, but where the analog to `name` is not restricted to being a string.

cirq/devices/graph_device.py
bryano: There are two types of edges (belonging to the two different graphs). Device edges corresponding to sets of qubits do indeed represent capabilities. Crosstalk edges correspond to sets of _device edges_ and represent constraints on the simultaneous capabilities of those device edges.

cirq/devices/hypergraph.py
bryano: Unfortunately not. I was tempted to use networks and restricted to (non-hyper-)graphs, but I think that could be overly limiting down the road. Two expected use cases for hyperedges are ion traps (which may have native more-than-two-qubit gates) and abstractions of superconducting devices that assume a "native" four-qubit gate (e.g. when compiling quantum chemistry circuits).

cirq/devices/hypergraph.py
bryano: The convention you suggest for singletons wouldn't work because `node` could be iterable. For example, an expected use case is lattice-based graph devices in which a reasonable vertex would be `(x, y)`.

Using just the edge dictionary is a possibility, but would require more time for common actions. The design choice I made reflects a particular space-time trade-off that seemed reasonable, but I'm open to others.

cirq/devices/hypergraph.py
bryano: Done

cirq/devices/graph_device.py
bryano: Done

cirq/devices/graph_device.py
bryano: What about `VertexQubit`?

Strilanc: `QubitId` already requires hashability. I think we should just use `QubitId`. This class just isn't pulling its weight in terms of added value; it's an extra detail that the user must remember when instead you could just say "any qubit works".

cirq/devices/graph_device.py
Strilanc: I would rename it to "disallowed_simultaneous_graph" or something like that, if that's what it is used for.

Strilanc: I'm unsure if this is the concept we want, as opposed to a set of pairs of operations that can't be performed simultaneously or a predicate that says yes or no when given a pair of operations. Do you have an example where using the graph is easier than using a predicate lambda or a set?

bryano: Note that the label of the constraint graph edge allows for specifying a set or predicate of forbidden pairs (or higher-order sets) of operations. By default, `None` indicates that no pair of operations is allowed; that's a good example of where it's easier to use the graph. Another point to make is that with respect to "pairs of operations that can't be performed simultaneously" is that really it's "pairs of operations that can't be performed simultaneously _on adjacent qubits_". The graph is what you the adjacency.

The other advantage of this class is what it exposes to a compiler. Suppose you're a compiler and in the course of doing your business you want to know "where can I do this logical gate?" With the existing predicate-only device class, you'd need to check every possible k-tuple of physical qubits; this class lets you iterate over a much smaller subset of possibilities.

cirq/devices/hypergraph.py
bryano: The removal of `HashQubit` obviates my objection to the singleton convention, but the point about the space-time trade-off remains relevant.

cirq/devices/graph_device.py
bryano: Done.

bryano:
See `/cirq/devices/uniform_graph_device.py` in #1212.

bryano:
Also, please take a look at the failing test in #1212. Seems strange.

Strilanc:
That PR seems to have more abstract specifications of what a device could be, but no actual device. What is a physical hardware device for which you would use this graph concept without having to add special cases?

bryano:
It's pretty general, so it's hard to imagine a physical device that wouldn't be easily expressed using this. Note also that it's not just for representing "physical" devices, but also abstract ones, e.g. line or 2D lattice, for enforcing spatial locality without bothering about native gate sets. This is very useful for testing compilers and getting reasonable resource counts.

Specifically, Rigetti's hardware could be implemented easily (as could Google's, if there weren't already a specialized class for it). For IBM, I would need to implement 'DirectedGraphDevice`, which would look pretty similar but have the obvious differences.

Strilanc:
But would we *want* to implement Rigetti's device in this way? I'd expect us to make a specialized class for it that encoded the structure more directly, instead of needing to use the abstract graph methods. For example, instead of storing all the connections it would just know the tiling pattern and tell you the neighbors according to that.

Have you been using this type in your own work?

bryano:
Yes, I've been using this and several derived types for defining devices to give to compilers that address connectivity constraints. The natural way of expressing the hardware half of an instance of the connectivity problem is as a hardware graph along these lines.

`GraphDevice` and a few helper functions will greatly simplify device specification. (I figure most devices could be specified in a few lines with the right abstractions.)

The bigger advantage is in what it exposes to a compiler. If we only care about validating circuits that were manually constructed to be valid, or implement compilation methods that are very device-specific, then this isn't so much of an issue. But it'll be very useful to have more general-purpose compilation procedures that take a circuit and a device specification and _try_ to do something reasonable. "Direct" implementations of devices aren't amenable to that.

Strilanc:
At the sync meeting we decided to put this in contrib for now.

bryano:
@Strilanc ping

vtomole:
@Strilanc ping +2

vtomole:
I'm good with getting this in if @Strilanc doesn't object. Would this close out https://github.com/quantumlib/Cirq/issues/95?

Strilanc:
Hey Bryano,

Sorry for leaving this hanging for ridiculously long. Thank you for moving the code into contrib. Once you get the CI to pass, I will merge it in.

dabacon:
This looks like it is just failing "format".  I think we should be able to just run the apply

bryano:
Done

PR #1191, opened 2018-11-22
Polish pass over the documentation

Strilanc:


cirq/ops/three_qubit_gates.py
```diff
@@ -121,7 +121,8 @@ class CCXPowGate(eigen_gate.EigenGate,
                  gate_features.InterchangeableQubitsGate):
     """A Toffoli (doubly-controlled-NOT) that can be raised to a power.
 
-    The matrix of CCX**t is an 8x8 identity except the bottom right 2x2 is X**t.
+    The matrix of `CCX**t` is an 8x8 identity except the bottom right 2x2 area
```
dabacon: This is awkward.  What is "bottom right"?  Actons as an identity except is a strange way to define things. 

Maybe:
The matrix of `CCX**t` acts as identity on all states except for the |110> and |111> state where it
acts like the matrix of `X**t` on these two states.

docs/gates.md
```diff
@@ -143,15 +130,18 @@ The matrix of `cirq.CZ**t` is ``exp(i pi |11><11| t)``.
...
-``XmonGates`` are hardware specific.  In addition Cirq has a
-number of more commonly named gates that are then implemented
-as ``XmonGates`` via decomposition or known unitaries. Some
-of these are our old friends:
+**S** the square root of Z gate
 
-**CNOT** the controlled-X gate, and **SWAP** the swap gate.
-As well as some other Clifford friends, **H** and **S**,
-and our error correcting friend **T**.
+and our error correcting friend **T**
```
dabacon: maybe say what this is?

cirq/ops/three_qubit_gates.py
Strilanc: Bottom right is.. bottom right? I'm not sure how to define it more clearly. You write out the identity matrix, then move your eyes right and down until you're looking at the corner, then the 2x2 block made by the intersection of the two columns and two rows you're closest to looking at will look like the 2x2 matrix of the X**t gate.

docs/gates.md
Strilanc: Slightly tweaked.

Issue #1131, opened 2018-11-08
Show empty moments in text diagrams (as extra spacing)


Strilanc:
For example, currently this code:

```python
op = cirq.X(cirq.NamedQubit('a'))
op_moment = cirq.Moment([op])
circuit = cirq.Circuit([op_moment, op_moment, cirq.Moment(), op_moment])
print(circuit)
```
prints:

```
a: ───X───X───X───
```

but it should print this instead:

```
a: ───X───X──────X───
```

It was originally intentional to remove the empty moments, but now I think they should be shown.

Strilanc:
@zoltanegyed

PR #1193, opened 2018-11-23
Render empty moments

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/1131

### DIFF
--- a/cirq/circuits/circuit.py
+++ b/cirq/circuits/circuit.py
@@ -847,8 +847,9 @@ class Circuit:
 
     @staticmethod
     def _pick_inserted_ops_moment_indices(operations: Sequence[ops.Operation],
-                                          start: int=0,
-                                          frontier: Dict[ops.QubitId, int]=None
+                                          start: int = 0,
+                                          frontier: Dict[ops.QubitId,
+                                                         int] = None
                                           ) -> Tuple[Sequence[int],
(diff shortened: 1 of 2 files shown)


cirq/circuits/circuit.py
```diff
@@ -1516,6 +1517,10 @@ def _draw_moment_in_diagram(moment: Moment,
                             precision: Optional[int],
                             moment_groups: List[Tuple[int, int]]):
     x0 = out_diagram.width()
+
+    if not moment.operations:
+        out_diagram.write(x0, 0, '─')
```
Strilanc: Does the empty string also work? This implementation will do the wrong thing for ascii-only diagrams.

cirq/circuits/circuit_test.py
```diff
@@ -144,6 +144,11 @@ def test_repr():
     ]),
 ], device=cirq.google.Foxtail)"""
 
+def test_empty_moment():
+    op = cirq.X(cirq.NamedQubit('a'))
+    op_moment = cirq.Moment([op])
+    circuit = cirq.Circuit([op_moment, op_moment, cirq.Moment(), op_moment])
+    assert str(circuit) == 'a: ───X───X───────X───'
```
Strilanc: Add an ascii-only test, and a 2-qubit test also.

vtomole:
@Strilanc Done.

PR #1194, opened 2018-11-23
Implement approximate equality protocol

mrwojtek:


cirq/__init__.py
```diff
@@ -242,6 +243,7 @@
     qasm,
     QasmArgs,
     SupportsApplyUnitary,
+    SupportsApproxEquality,
```
Strilanc: `Approximate`

cirq/protocols/approx_equality.py
```diff
@@ -0,0 +1,172 @@
...
+    """Object which can be approximately compared."""
+
+    def _approx_eq_(self, other: Any, rel_tol: float, abs_tol: float) -> bool:
+        """Approximate comparator.
+
+        Types implementing this protocol define their own logic for approximate
+        comparison with other types.
+
+        Args:
+        other: Target object for approximate comparison.
```
Strilanc: Indentation

cirq/protocols/approx_equality.py
```diff
@@ -0,0 +1,172 @@
...
+        rel_tol: The relative tolerance. See math.isclose() documentation for
```
Strilanc: Hmmmm, numpy uses `rtol` and `atol` but `math.isclose` uses `rel_tol` and `abs_tol`. I think for now we should stick to `rtol` and `atol`, and point at numpy's isclose. We can switch this later, to keep the whole library consistent.

cirq/protocols/approx_equality.py
```diff
@@ -0,0 +1,172 @@
...
+
+from typing import Any
+
+from typing_extensions import Protocol
+
+
+class SupportsApproxEquality(Protocol):
+    """Object which can be approximately compared."""
+
+    def _approx_eq_(self, other: Any, rel_tol: float, abs_tol: float) -> bool:
```
Strilanc: Force keyword args when setting the tolerances by adding a `, *,` argument after the `other` parameter.

cirq/protocols/approx_equality.py
```diff
@@ -0,0 +1,172 @@
...
+        abs_tol: float = 0.0) -> bool:
+    """Approximately compares two objects.
+
+    If `val` implements SupportsApproxEquality protocol then it is invoked and
+    takes precedence over all other checks:
+     - For primitive numeric types `int` and `float` approximate equality is
+       delegated to math.isclose().
+     - For complex primitive type the real and imaginary parts are treated
+       independently and compared using math.isclose().
+     - For `val` and `other` both iterable, consecutive elements are compared
```
Strilanc: Note that they need to have the same length.

cirq/protocols/approx_equality.py
```diff
@@ -0,0 +1,172 @@
...
+    approx_eq_getter = getattr(val, '_approx_eq_', None)
+    if approx_eq_getter is not None:
+        return approx_eq_getter(other, rel_tol, abs_tol)
+
+    val_t = type(val)
+    other_t = type(other)
+
+    # Fallback to math.isclose() for primitive types, which is symmetric
+    # contrary to numpy.isclose().
+    if val_t in [int, float]:
```
Strilanc: I think `if isinstance(val, (int, float))` is the more common way to do this.

cirq/protocols/approx_equality.py
```diff
@@ -0,0 +1,172 @@
...
+        if other_t not in [int, float]:
+            return NotImplemented
+        return math.isclose(val, other, rel_tol=rel_tol, abs_tol=abs_tol)
+
+    # For complex types, treat real and imaginary parts independently.
+    if val_t == complex:
```
Strilanc: If we delegated to `np.isclose`, this would be handled for us. That's also `cmath.isclose`.

cirq/protocols/approx_equality.py
```diff
@@ -0,0 +1,172 @@
...
+        NotImplemented when approximate equality is not implemented for given
+        types.
+    """
+
+    # Check if object defines approximate equality via _approx_eq_. This takes
+    # precedence over all other overloads.
+    approx_eq_getter = getattr(val, '_approx_eq_', None)
+    if approx_eq_getter is not None:
+        return approx_eq_getter(other, rel_tol, abs_tol)
+
```
Strilanc: Because approximate equality is symmetric, we should also check `getattr(other, '_approx_eq_', None)` before continuing to the default cases. For example, `cirq.approx_eq(5.0, fraction(5, 1))` should work.

cirq/protocols/approx_equality.py
```diff
@@ -0,0 +1,172 @@
...
+            abs_tol=abs_tol
+        ) and math.isclose(
+            val.imag,
+            other.imag,
+            rel_tol=rel_tol,
+            abs_tol=abs_tol
+        )
+
+    # Try to compare source and target recursively, assuming they're iterable.
+    return _approx_eq_iterables(val, other, rel_tol=rel_tol, abs_tol=abs_tol)
```
Strilanc: Another fallback we need, after everything else fails including the iterability check, is the fallback to `__eq__`.

Strilanc: You can explicitly check for iterability with `isinstance(val, collections.Iterable)`.

cirq/protocols/approx_equality_test.py
```diff
@@ -0,0 +1,73 @@
...
+        return math.isclose(
+            self.val,
+            other.val,
+            rel_tol=rel_tol,
+            abs_tol=abs_tol
+        )
+
+
+def test_approx_eq_supported():
+    assert cirq.approx_eq(A(0.0), A(0.1), abs_tol=0.1) is True
```
Strilanc: Don't `assert c is True`, just do the raw `assert c`

cirq/protocols/approx_equality_test.py
```diff
@@ -0,0 +1,73 @@
...
+    assert cirq.approx_eq(A(0.0), A(0.1)) is False
```
Strilanc: Don't `assert c is False`, just do the raw `assert not c`

cirq/protocols/approx_equality_test.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class B:
+    pass
+
+
+def test_approx_eq_not_supported():
+    assert cirq.approx_eq(B(), B()) is NotImplemented
```
Strilanc: The `NotImplemented` result should never be exposed to the user. If the approximate comparison is not implemented, we should assume the values are different.

We could consider adding a `default: Any=False` parameter, but I think this isn't useful for approximate equality because it has such a clear default result.

cirq/protocols/approx_equality.py
mrwojtek: Added `, *` for all the internal methods. I keep `approx_eq()` without forcing keyword args?

cirq/protocols/approx_equality.py
mrwojtek: This is satisfied, updated documentation.

cirq/protocols/approx_equality_test.py
mrwojtek: After dropping NotImplemented this makes much sense now, done.

cirq/protocols/approx_equality_test.py
mrwojtek: This makes a lot of sense (dropping NotImplemented). Applied changes.

mrwojtek: Agree with forcing default to `False`.

cirq/protocols/approx_equality.py
mrwojtek: Done.

cirq/protocols/approx_equality.py
mrwojtek: Done.

cirq/protocols/approx_equality_test.py
mrwojtek: Done.

cirq/protocols/approx_equality.py
mrwojtek: Done.

cirq/protocols/approx_equality.py
mrwojtek: Apparently iter() works for more cases (https://stackoverflow.com/a/1952655). Do you want me to change it still?

cirq/protocols/approx_equality.py
mrwojtek: Before I saw your comment I already implemented backward-compatible `_isclose()`. It fallbacks to `math.isclose()` for Python >= 3.5 and` numpy.isclose()` wht Python is older.

I think this is strictly better, since `math.isclose()` usage is much cleaner (to make `numpy.isclose()` symmetric, I needed to implement some additional logic, see my code). I don't think we'll be able to switch to `math.isclose()` very soon, since it requires pretty new Python dependency. Also, `math.isclose()` is more strict then `numpy.isclose()`: if some test passes on Python >= 3.5 it will pass on Python < 3.5 (up to NaN and other special values handling which I didn't check for differences).

cirq/protocols/approx_equality.py
mrwojtek: Keeping this one open for now until we resolve `_isclose()`.

mrwojtek:
@Strilanc, I didn't send out this PR for you to review earlier since I was still working on it. Thanks for early comments!

Do you know if I can check the generated documentation somehow and see if my docstrings are properly formatted?

Do you have any advice on failing incremental coverage test?

Strilanc:
You have to annotate the python2-only block with `# coverage: ignore`

Why are you doing different things in python 3 and python 2?

cirq/protocols/approx_equality.py
Strilanc: It's fine if you use `iter` and catch the exception to detect that it's not iterable. But make sure you forward a user-relevant exception about the method they called, instead of something specifically about iterability.

cirq/protocols/approx_equality.py
Strilanc: I think we should be consistent between the two versions of python. If a method doesn't exist in python 2, and doesn't translate in a way that can be handled by the transpilation scripts, then we should avoid using that method. The only place I'm happy to condition on python 3 or not is in the tests.

mrwojtek: Sounds reasonable. I'll remove `math.isclose` completely in that case. Do you think it's reasonable to keep my wrapper around `numpy.isclose` to make it symmetric?

mrwojtek: Done.

cirq/protocols/approx_equality.py
mrwojtek: Done.

cirq/protocols/approx_equality.py
mrwojtek: Done.

cirq/protocols/approx_equality.py
mrwojtek: I think this is handled as of now by delegating to __eq__; if one type is iterable and the other is not iterable the iterability approximate equality will fail with NotImplemented and delegate to __eq__ in the end. I don't throw any exception to the user.

Strilanc: Ah yes, that's much nicer.

cirq/protocols/approximate_equality.py
```diff
@@ -0,0 +1,170 @@
...
+            given types.
+        """
+        pass
+
+
+def approx_eq(
+        val: Any,
+        other: Any,
+        *,
+        atol: float = 0.0) -> bool:
```
Strilanc: Don't specify a default value for `atol`. If a user wants exact equality they'll use `==`.

cirq/protocols/approximate_equality_test.py
```diff
@@ -0,0 +1,94 @@
...
+        if not isinstance(self, type(other)):
+            return NotImplemented
+        return self.val == other.val
+
+
+def test_approx_eq_uses__eq__():
+    assert cirq.approx_eq(C(0), C(0))
+    assert not cirq.approx_eq(C(1), C(2))
+    assert cirq.approx_eq([C(0)], [C(0)])
+    assert not cirq.approx_eq([C(1)], [C(2)])
```
Strilanc: Test cases to add:

- empty list
- different sized lists
- different type of iterables being compared, tuple vs list vs generator
- sets and dicts...? Actually nevermind, we'll add that later only if we need it.
- `[]` vs `[[]]` (should be false)
- element vs singleton list containing element (should be false)
- NaN vs 0 (false)
- NaN vs NaN (false)
- Infinity vs -Infinity
- Infinity vs 5
- Infinity vs 0
- Infinity vs Infinity

cirq/protocols/approximate_equality.py
mrwojtek: Done.

cirq/protocols/approximate_equality_test.py
mrwojtek: The last case is interesting (inf vs inf), both math.isclose and np.isclose returns `True` in this case which seems somehow odd. Do you want me to special case this case so it gives `False`?

Strilanc: Infinity is equal to infinity, therefore they are also approximately equal. Should be true.

cirq/protocols/approximate_equality_test.py
```diff
@@ -0,0 +1,120 @@
...
+    assert cirq.approx_eq(1.0, 1.0 + 1e-10, atol=1e-09)
+    assert not cirq.approx_eq(1.0, 1.0 + 1e-10, atol=1e-11)
+    assert cirq.approx_eq(0.0, 1e-10, atol=1e-09)
+    assert not cirq.approx_eq(0.0, 1e-10, atol=1e-11)
+    assert cirq.approx_eq(complex(1, 1), complex(1.1, 1.2), atol=0.3)
+    assert not cirq.approx_eq(complex(1, 1), complex(1.1, 1.2), atol=0.1)
+
+
+def test_approx_eq_special_numerics():
+    assert not cirq.approx_eq(float('nan'), 0, atol=0.0)
```
Strilanc: I'd use a non-zero `atol` also, but I think this is sufficient for now.

mrwojtek:
np, I'm not very fast either (yet ;)

PR #1199, opened 2018-11-27
Add option to skip cleanup in two qubit decomposition

dstrain115:
_cleanup_operations seems to take up 30-40% of the time in the conversion to Xmon gates.  However, this does optimizations that are all done by later phases of the Xmon optimizer set. 

This changes the default to skip this step and adds an option to allow you to optimize the operation list if you want to anyway.  This should only affect two qubit decomposition and not the full circuit optimization, but this should be verified.

### DIFF
--- a/cirq/google/convert_to_xmon_gates.py
+++ b/cirq/google/convert_to_xmon_gates.py
@@ -57,7 +57,8 @@ class ConvertToXmonGates(PointOptimizer):
                 op.qubits[0],
                 op.qubits[1],
                 mat,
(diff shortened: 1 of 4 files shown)


Strilanc:
I think the default should be reversed. We actually added this change recently, because it's very surprising when this optimization spits out a glut of single qubit gates. Then you would modify `optimized_for_xmon` to specify no cleanup.

dstrain115:
Default state flipped.  Please take another look.

PR #1201, opened 2018-11-28
reorganized and augmented acquaintance module

bryano:
* added BipartiteSwapNetworkGate
* added device attribute to execution strategy
* added acquaintance_size=None option to SwapNetworkGate
* made strategies submodule
* added acquaintance DAG
* made CircularShiftGate decompose into SwapPermutationGate

### DIFF
--- a/cirq/contrib/acquaintance/bipartite_test.py
+++ b/cirq/contrib/acquaintance/bipartite_test.py
@@ -18,24 +18,17 @@ import pytest
 
 import cirq
 
-from cirq.contrib.acquaintance.bipartite import (
-        BipartiteSwapNetworkGate, BipartiteGraphType)
-from cirq.contrib.acquaintance.devices import (
-        get_acquaintance_size)
-from cirq.contrib.acquaintance.gates import (
-        AcquaintanceOpportunityGate)
-from cirq.contrib.acquaintance.permutation import (
-        SwapPermutationGate, update_mapping)
+import cirq.contrib.acquaintance as cca
 
 
 circuit_diagrams = {
-    ('undecomposed', BipartiteGraphType.COMPLETE, 1):
+    ('undecomposed', cca.BipartiteGraphType.COMPLETE, 1):
 """
 0: ───K_{1, 1}:(0, 0)↦(0, 0)───
       │
 1: ───K_{1, 1}:(1, 0)↦(1, 0)───
 """,
-    ('undecomposed', BipartiteGraphType.COMPLETE, 2):
+    ('undecomposed', cca.BipartiteGraphType.COMPLETE, 2):
 """
 0: ───K_{2, 2}:(0, 0)↦(0, 0)───
       │
@@ -45,7 +38,7 @@ circuit_diagrams = {
       │
 3: ───K_{2, 2}:(1, 1)↦(1, 1)───
 """,
-    ('undecomposed', BipartiteGraphType.COMPLETE, 3):
+    ('undecomposed', cca.BipartiteGraphType.COMPLETE, 3):
 """
 0: ───K_{3, 3}:(0, 0)↦(0, 0)───
       │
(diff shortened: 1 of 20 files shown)


cirq/contrib/acquaintance/bipartite_test.py
```diff
@@ -0,0 +1,305 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import itertools
+
+import pytest
+
+import cirq
```
Strilanc: Don't import cirq within cirq.

cirq/contrib/acquaintance/executor.py
```diff
@@ -17,21 +17,25 @@
...
+from cirq.contrib.acquaintance.mutation_utils import (
         expose_acquaintance_gates)
 
+
 if TYPE_CHECKING:
     # pylint: disable=unused-import
     from typing import Callable, List, DefaultDict
 
+
 class ExecutionStrategy(metaclass=abc.ABCMeta):
```
Strilanc: This class looks like it might be abstraction for abstraction's sake. Is there a reason not to downgrade it into a `NamedTuple` or just a small set of operations accepted by various methods?

cirq/contrib/acquaintance/executor.py
```diff
@@ -93,6 +111,19 @@ def optimization_at(self,
                          'are instances of AcquaintanceOpportunityGate or '
                          'PermutationGate.')
 
+
+class AcquaintanceOperation(ops.GateOperation):
+    """Represents an a realized acquaintance opportunity.
```
Strilanc: This description is not very clear. What is a "realized acquaintance opportunity"?

Is this class an operation that occurs as two qubits pass each other during a permutation, including details both of the physical qubits being acted upon and the logical qubits/modes those physical qubits are temporarily storing?

cirq/contrib/acquaintance/gates_test.py
```diff
@@ -124,10 +132,35 @@ def test_swap_network_gate():
...
+    swap_network_op = SwapNetworkGate(
+        part_lens, acquaintance_size=None)(*qubits)
+    swap_network = cirq.Circuit.from_ops(
+            swap_network_op, device=UnconstrainedAcquaintanceDevice)
+    initial_mapping = {q: i for i, q in enumerate(qubits)}
+
+    actual_opps = get_logical_acquaintance_opportunities(
+            swap_network, initial_mapping)
+    parts = []
+    q = 0
```
Strilanc: Move this loop to the top, and merge `q` into `n_qubits`

cirq/contrib/acquaintance/bipartite_test.py
bryano: I thought that was the convention within tests?

cirq/contrib/acquaintance/executor.py
bryano: It's nice to be able to have execution-strategy-specific methods; see `GreedyExecutionStrategy`.

cirq/contrib/acquaintance/executor.py
bryano: Yea, that's right. I updated the docstring.

cirq/contrib/acquaintance/gates_test.py
bryano: Done

cirq/contrib/acquaintance/bipartite_test.py
Strilanc: Ah, my bad, I was mislead by the fact that there were imports within sub modules.

You should export all your methods up to cirq.contrib.acquaintances and then import that as ca or something.

bryano: Done

Issue #982, opened 2018-10-03
Displays


kevinsung:
@Strilanc I'd like to know your thoughts on this "display" concept that you mentioned at #942 . Do you imagine this as a separate class whose instances would be bundled with the Circuits sent to Quantum Engine?

Strilanc:
A `Display` would be a special kind of gate that indicates "please tell me XX about the qubits I am targeting at this time". Displays are no-ops as far as the rest of the circuit is concerned (there would be dedicated runs of the circuit to compute each display's value, and for other runs the display is omitted).

Basically, it's exactly the concept of displays that I use in Quirk.

bryano:
It might be useful to integrate this with the concept of an "acquaintance gate". One would commonly be interested in "displaying" the logical qubits (or modes in the fermionic case) mapped to the corresponding physical qubits at some point in time.

kevinsung:
@Strilanc how does this affect the Quantum Engine API? Are you proposing to add a method separate from `run` that would be used to calculate displays?

PR #1202, opened 2018-11-29
Define unitary and apply_unitary for Pauli

kevinsung:


kevinsung:
Yes, exactly. I intend to use this for #982

PR #1203, opened 2018-11-29
Make PauliString an Operation

kevinsung:


cirq/ops/pauli_string.py
```diff
@@ -69,8 +69,12 @@ def __contains__(self, key: raw_types.QubitId) -> bool:
...
         return self._qubit_pauli_map.keys()
 
-    def qubits(self) -> KeysView[raw_types.QubitId]:
-        return self.keys()
+    @property
+    def qubits(self) -> Tuple[raw_types.QubitId, ...]:
+        return tuple(self.keys())
+
+    def with_qubits(self, *new_qubits: raw_types.QubitId) -> 'PauliString':
+        return PauliString(dict(zip(new_qubits, self.values())), self.negated)
```
viathor: IIUC, the user should expect the order of qubits passed to `with_qubits()` to match the order of qubits passed to `__init__()`.

Unfortunately, `PauliString` uses `dict` to receive and store qubit-to-Pauli mapping which doesn't preserve(*) order. For the API of `with_qubits` to have precise meaning, we need to change qubit mapping used internally and in `__init__()` from `dict` to something order-aware (e.g. [`OrderedDict`](https://docs.python.org/3/library/collections.html#collections.OrderedDict)). 

--
(*) Technically speaking, the new `dict` implementation in python 3.6 does maintain fixed order internally but [documentation](https://docs.python.org/3.6/whatsnew/3.6.html#new-dict-implementation) is explicit that this is merely an implementation detail that may change in future and should not be relied upon.

Strilanc: For now I would like to preserve the idea that there is an order to the qubits of an operation. It's so useful for telling the qubits of a CNOT or a QFT apart.

The solution in this case is to use the order as determined by `sorted`, e.g. to return `sorted(self.keys())` instead of `self.keys()`.

kevinsung: Done.

PR #1209, opened 2018-11-30
Use same type for measurement results everywhere

### LINKED ISSUE #1052: SimulatesSamples._run return type is different from type(TrialResult.measurements)
One is Dict[str, List[np.ndarray]] and the other is Dict[str, np.ndarray]. They should be the same, right?

kevinsung:
Fixes #1052 

I thought it was confusing that the same concept is being represented in different places as `Dict[str, np.ndarray]`, `Dict[str, List[np.ndarray]]`, and `Dict[str, List[List[bool]]]`. Is there a reason for this?


kevinsung:
Why won't CirqBot automerge for me?

PR #1212, opened 2018-11-30
More graph device

bryano:


bryano:
I can't reproduce the failing test on my machine. It looks like `sorted` isn't doing the right thing, which is weird.

Strilanc:
At the sync meeting we decided to put this in contrib.

vtomole:
@bryano  Please `QubitId` -> `Qid`.

Strilanc:
Still blocked on the Qid change, and getting very stale. I'll close in two weeks if there are no updates.

bryano:
@Strilanc PTAL

PR #1213, opened 2018-12-03
Define __iter__ and __len__ for Moment

kevinsung:


kevinsung:
@Strilanc Moments are already valid things to pass into `circuit.append`, with your desired behavior.

Strilanc:
Even when they're in a list?

kevinsung:
@Strilanc I've modified `append` to handle a Sequence of Moments.

cirq/circuits/circuit.py
```diff
@@ -1075,15 +1075,24 @@ def batch_insert(self,
...
+                to the circuit
             strategy: How to pick/create the moment to put operations into.
         """
-        self.insert(len(self._moments), moment_or_operation_tree, strategy)
+        if (isinstance(moments_or_operation_tree, Sequence)
+                and len(moments_or_operation_tree) > 0
+                and isinstance(moments_or_operation_tree[0], Moment)):
+            for moment in moments_or_operation_tree:
+                self.insert(len(self._moments), moment, strategy)
+        else:
```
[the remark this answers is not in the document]
kevinsung: Ah, I somehow didn't realize until now that this change makes Moments satisfy the OP_TREE contract.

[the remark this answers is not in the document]
kevinsung: ... which seems to necessitate modifying `flatten_op_tree` to yield Moments in addition to Operations. @Strilanc what do you think?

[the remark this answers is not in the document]
bryano: But then the output of `flatten_op_tree` wouldn't be flat.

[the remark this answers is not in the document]
kevinsung: True, that doesn't seem good. I guess I could rewrite the insertion logic to not use `flatten_op_tree`.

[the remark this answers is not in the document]
kevinsung: Or, we could add a `preserve_moments` boolean flag to `flatten_op_tree`.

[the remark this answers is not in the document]
Strilanc: I would add the `preserve_moments` flag.

cirq/circuits/circuit.py
```diff
@@ -763,16 +763,16 @@ def _can_commute_past(self,
     def insert(
             self,
             index: int,
-            moment_or_operation_tree: Union[Moment, ops.OP_TREE],
+            operations: ops.OP_TREE,
```
Strilanc: I think the moment or operation tree name is better, because it indicates to the user that moments are handled in a special way.

cirq/circuits/moment.py
```diff
@@ -14,9 +14,11 @@
 
 """A simplified time-slice of operations within a sequenced circuit."""
 
-from typing import Any, Iterable, TypeVar, Callable, Sequence
+from typing import Any, Iterable, TypeVar, Callable, Sequence, TYPE_CHECKING
 
-from cirq import ops
+if TYPE_CHECKING:
+    # pylint: disable=unused-import
+    from cirq import ops
```
Strilanc: Why did you need to move this into a type-checking only block? There shouldn't be any dependence of ops on circuits, and if there is the cycle should be broken there.

cirq/ops/op_tree.py
```diff
@@ -19,6 +19,7 @@
 
 from typing import Any, Iterable, Union, Callable
 
+from cirq.circuits.moment import Moment
```
Strilanc: Ah. Break cycle here instead of elsewhere.

kevinsung: I don't see a way to implement the `preserve_moments` boolean flag without using Moment in `isinstance` checks.

kevinsung:
mypy is very unhappy with this change ...

cirq/circuits/circuit.py
```diff
@@ -787,16 +787,17 @@ def _can_commute_past(self,
     def insert(
             self,
             index: int,
-            moment_or_operation_tree: Union[Moment, ops.OP_TREE],
+            moment_or_operation_tree: ops.OP_TREE,
```
Strilanc: This is not the correct type.

I recommend defining `OP_OR_MOMENT_TREE = Union[Operation, Moment, OP_TREE, Iterable[Any]]` in order to appease mypy.

cirq/ops/op_tree.py
Strilanc: That sounds reasonable.

cirq/circuits/circuit.py
kevinsung: Why isn't it the correct type? Operation and Moment are both examples of OP_TREE.

Strilanc: That's... Huh. I guess that's right. It just seems really weird to tell the user you're asking for "operations or moments" and then just say op tree.

kevinsung: That's why I initially renamed to variable to just `operations`. Also, mypy was mainly unhappy about the return type of `flatten_op_tree` being changed from `Iterable[Operation]` to `Iterable[Union[Operations, Moment]]`.

Strilanc: You would need to define overloads that said OP_TREE mapped to OP_TREE while the more complicated type mapped to the more complicated type.

kevinsung: @Strilanc do you still want me to define `OP_OR_MOMENT_TREE` or are you okay with just leaving `OP_TREE` and changing the variable name to simply `operations`?

Strilanc: In my opinion the variable name needs to say moments because that's important documentation that will affect user expectations. And the type name should probably also say moments. So yes, please change it.

kevinsung: I've renamed the variable to `moment_or_operation_tree` and the type to `Union[Moment, OP_TREE]`.

PR #1214, opened 2018-12-03
Change default insertion strategy to EARLIEST

smitsanghavi:


Strilanc:
Feel free to specify "NEW_THEN_INLINE" as the strategy for any tests that fail after the change, unless those tests are about the circuit layout being a specific way. E.g. for the supremacy circuits I would stick with NEW_THEN_INLINE.

docs/tutorial.md
```diff
@@ -151,8 +151,10 @@ gate to every qubit whose row index plus column index is odd.  To
 do this we write
 ```python
 circuit = cirq.Circuit()
-circuit.append(cirq.H(q) for q in qubits if (q.row + q.col) % 2 == 0)
-circuit.append(cirq.X(q) for q in qubits if (q.row + q.col) % 2 == 1)
+circuit.append([cirq.H.on(q) for q in qubits if (q.row + q.col) % 2 == 0],
+               strategy=cirq.InsertStrategy.NEW_THEN_INLINE)
```
Strilanc: I would avoid specifying the insertion strategy here; update the later text instead. It's confusing to throw this detail in in the middle of a tutorial.

smitsanghavi: Done

smitsanghavi:
> Feel free to specify "NEW_THEN_INLINE" as the strategy for any tests that fail after the change, unless those tests are about the circuit layout being a specific way. E.g. for the supremacy circuits I would stick with NEW_THEN_INLINE.

This did not work for xmon_simulator_test.py as I mentioned above. Also, I'm not able to determine the root cause for the precision mismatch in decompositions_test.py failure.

Strilanc:
@dabacon said he'd look at these failures and help out

smitsanghavi:
@dabacon Friendly ping

dabacon:
Sorry taking a look now.

dabacon:
OK the only xmon simulator tests that are failing are the ones that use the moment_by_moment conversion from scheduler to circuit.  When it gets simulated this then uses the schedule.to_circuit to convert it back to a circuit.  The last of these doesn't have the same behavior as previously, i.e. the circuit has a different moment structure.  This causes the test to fail because the measurements end up in different moments, and so end up coming from a different order.

To fix this, to_circuit in schedule.py needs to be fixed to work with the new insertion strategy

PR #1217, opened 2018-12-04
Add phase_exponent to XPowGate and YPowGate

### LINKED ISSUE #1216: Property missing for instances of PhasedXPowGate
Steps to reproduce:

```
In [2]: g1 = cirq.PhasedXPowGate(exponent=0.2, phase_exponent=0.4)                                                                                                                      

In [3]: g2 = cirq.PhasedXPowGate(exponent=0.2, phase_exponent=0.5)                                                                                                                      

In [4]: g1.phase_exponent                                                                                                                                                               
Out[4]: 0.4

In [5]: g2.phase_exponent                                                                                                                                                               
AttributeError                            Traceback (most recent call last)
<ipython-input-5-0b9226f7e654> in <module>
----> 1 g2.phase_exponent

AttributeError: 'YPowGate' object has no attribute 'phase_exponent'
```

The cause is the fact that `__new__()` intercepts and special-cases gate instantiations when phase_exponent is -1/2, 0, 1/2 or 1. It then returns an `XPowGate` or `YPowGate` which don't have `phase_exponent` property.

viathor:
Fixes #1216.

### DIFF
--- a/cirq/ops/common_gates.py
+++ b/cirq/ops/common_gates.py
@@ -111,6 +111,10 @@ class XPowGate(eigen_gate.EigenGate,
             return args.format('rx({0:half_turns}) {1};\n',
                                self._exponent, qubits[0])
 
+    @property
+    def phase_exponent(self):
+        return 0.0
+
     def _phase_by_(self, phase_turns, qubit_index):
         """See `cirq.SupportsPhase`."""
         return cirq.ops.phased_x_gate.PhasedXPowGate(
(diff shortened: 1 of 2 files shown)


viathor:
I agree. These implicit behind-the-scenes type memberships will continue to appear due to the inherently ambiguous nature of some of the objects we model (e.g. Paulis are both unitary and hermitian). Another way of saying this is that the objects we model are members of multiple intersecting algebraic structures we care about.

We need a consistent way of capturing these relationships throughout cirq.

PR #1218, opened 2018-12-04
Correct and update Circuit docstring with available methods

kevinsung:


kevinsung:
@CirqBot where you at??

Strilanc:
CirqBot is not running all the time. I try to keep an eye on it in case it decides to rebel.

Issue #1083, opened 2018-10-31
If decompose's on_stuck_raise is a lambda that returns None, no error should be raised


Strilanc:
Currently, passing in None instead of a lambda has this behavior. But this can result in duplicate logic when passing in the lambda or None and implementing the lambda.

andbe91:
Hi @iamvamsi, are you still working on this? If not, I'm happy to do it.

Strilanc:
@andbe91 Feel free to take over this one.

iamvamsi:
@andbe91 : I haven't yet started on this. Please feel free to take it.

PR #1219, opened 2018-12-04
Handle the case when on_stuck_return is a function returning None

andbe91:
#1083 
Please check if you agree with the argument type for on_stuck_raise. Is `Union[None, X]` or `Optional[X]` preferred by the way?

### DIFF
--- a/cirq/protocols/decompose.py
+++ b/cirq/protocols/decompose.py
@@ -144,7 +144,11 @@ def decompose(
                                         NotImplementedType,
                                         'cirq.OP_TREE']] = None,
     keep: Callable[['cirq.Operation'], bool] = None,
-    on_stuck_raise=_value_error_describing_bad_operation
+    on_stuck_raise: Union[None,
+                          Exception,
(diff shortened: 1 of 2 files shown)


Strilanc:
I prefer `Optional[X]` but not enough to consider it a problem.

cirq/protocols/decompose.py
```diff
@@ -234,8 +239,10 @@ def decomposer(op):
         if keep is not None and on_stuck_raise is not None:
             if isinstance(on_stuck_raise, Exception):
                 raise on_stuck_raise
-            else:
-                raise on_stuck_raise(item)
+            elif callable(on_stuck_raise):
+                error = on_stuck_raise(item)
+                if isinstance(error, Exception):
```
Strilanc: Change this to `if error is not None`. I would rather the code attempt to raise bad objects rather than silently keep going.

andbe91: Done

PR #1220, opened 2018-12-04
Adds a static range method to the NamedQubit class.

philiptmassey:


cirq/ops/named_qubit_test.py
```diff
@@ -60,3 +60,17 @@ def test_pad_digits():
     assert _pad_digits('a9') == 'a00000009:1'
     assert _pad_digits('a09') == 'a00000009:2'
     assert _pad_digits('a00000000:8') == 'a00000000:8:00000008:1'
+
+
+@cirq.testing.only_test_in_python3
+def test_named_qubit_range():
+    qubits = cirq.NamedQubit.range('a', 2)
```
Strilanc: Hmmmm. Something about the way this looks isn't quite right. It looks like you're saying "give me named qubits from A to 2.

I think we should change `prefix` from the first argument to a keyword-only argument.

```python
def range(*args, prefix: str) -> Iterable['NamedQubit']:
```

philiptmassey: I agree that the range does look a bit weird, and a keyword argument is a good solution. However, I also think that it looks like it isn't required, e.g. that this is reasonable behavior:
```
qubits = NamedQubit.range(5)
```
Maybe we should also provide a default prefix? "a", "cirq"?

Strilanc: I don't think there's an obvious winner for the default. We can always add one later.

PR #1215, opened 2018-12-03
[Parent PR] Implement Displays

kevinsung:
Fixes #982 .
Comes after #1203 , #1213 

The only simulator interface that I modified is SimulatesIntermediateWaveFunction. I leave modifying the other ones for a separate PR. Also, I might end up splitting this PR into multiple pieces.

PR #1221, opened 2018-12-04
Define SamplesDisplay, WaveFunctionDisplay, and Pauli string expectations

### LINKED ISSUE #982: Displays

kevinsung:
Fixes #982 

Peeled off of #1215


cirq/__init__.py
```diff
@@ -160,6 +151,19 @@
     ZPowGate,
     ZZ,
     ZZPowGate,
+    amplitude_damp,
```
Strilanc: Why'd these move to the end? Because lower case?

cirq/ops/display.py
```diff
@@ -0,0 +1,175 @@
...
+    from cirq.ops import pauli_string
+
+
+class SamplesDisplay(raw_types.Operation):
+    """A display whose value is computed from measurement results.
+
+    The value of a SamplesDisplay on some qubits is computed in the following
+    steps:
+        1. Perform some unitary operations on the qubits
+        2. Sample bitstrings from the resulting state a number of times
```
Strilanc: This description is not correct. It sounds like you're describing repeated measurement of one prepared state, instead of repeated prepare+measure.

cirq/ops/display.py
```diff
@@ -0,0 +1,175 @@
...
+        pass
+
+    @abc.abstractmethod
+    def measurement_basis_change(self) -> op_tree.OP_TREE:
+        """Operations to perform prior to measurement."""
+        pass
+
+    @property
+    @abc.abstractmethod
+    def repetitions(self) -> int:
```
Strilanc: I would rename this to `desired_sample_count` or similar.

cirq/ops/display.py
```diff
@@ -0,0 +1,175 @@
...
+        """The number of measurement samples to take."""
+        pass
+
+    @abc.abstractmethod
+    def value(self,
```
Strilanc: Is this supposed to be the function that derives the display's value from the samples? Rename to `samples_to_values` or similar, and make the docstring clearer.

cirq/ops/display.py
```diff
@@ -0,0 +1,175 @@
...
+class WaveFunctionDisplay(raw_types.Operation):
+    """A display whose value is computed from the full wavefunction."""
+
+    @property
+    @abc.abstractmethod
+    def key(self) -> Hashable:
+        pass
+
+    @abc.abstractmethod
+    def value(self,
```
Strilanc: `value_derived_from_wavefunction` or similar.

cirq/ops/display.py
```diff
@@ -0,0 +1,175 @@
...
+              state: np.ndarray,
+              qubit_map: Dict[raw_types.QubitId, int]
+              ) -> Any:
+        pass
+
+
+@value.value_equality
+class ApproxPauliStringExpectation(SamplesDisplay):
```
Strilanc: The name here could be better. I came up with `PauliStringObservableSamplesDisplay` but that's even worse.

cirq/ops/display.py
```diff
@@ -0,0 +1,175 @@
...
+    def value(self,
+              measurements: np.ndarray) -> float:
+        return np.mean([(-1)**sum(bitstring) for bitstring in measurements])
+
+    def _value_equality_values_(self):
+        return self._pauli_string, self._repetitions, self._key
+
+
+@value.value_equality
+class PauliStringExpectation(WaveFunctionDisplay):
```
Strilanc: It feels like these two pauli displays are actually the same display with slightly different options. In particular, it seems like it's a kind of "graceful degradation" display where in the simulator you might want idealized values but in the real hardware you want sampled values.

Maybe we could have a method like this:

```python
def pauli_vector_expectation(
        pauli_vector: Union[PauliVector, ..., ..., ...],
        *,
        samples: int,
        fake_statistical_noise_in_simulator: bool = True):
    ....
```

?

cirq/__init__.py
kevinsung: Yeah. `ord('a') > ord('Z')`.

cirq/ops/display.py
kevinsung: During a simulation you could want either the idealized value or an approximate value from samples. On hardware you have no choice; you can only get the latter.

Are you proposing that `pauli_vector_expectation` return one of these two displays based on the setting of `fake_statistical_noise_in_simulator`? But then the `samples` argument doesn't have any meaning if we're asking for the idealized wavefunction display, so it's weird to force the user to enter that. We could make it default to 0 or None to indicate the ideal one, and make a different value indicate the samples one.

cirq/ops/display.py
kevinsung: I agree that your suggestion is worse.

cirq/ops/display.py
Strilanc: Hmmmm. How about

```python
def pauli_vector_expectation(
        pauli_vector: Union[PauliVector, ..., ..., ...],
        *,
        samples: Optional[int] = None,
        idealized: bool = False):
```

If the user specifies only `samples`, then the simulator adds fake noise and it works on hardware. If the user specifies only `idealized=True` then the simulator doesn't add noise and it doesn't work on hardware. If the user specifies both then the simulator doesn't add noise and also it works on hardware.

I want to push users towards thinking "samples works everywhere so I want to use that, but sometimes I will additionally specify idealized when testing".

Strilanc: I think the main alternative is to default to idealized, and if the user specifies samples then we can run on hardware but also inject noise in the simulator. But from the user's perspective I really want them to think of this thing as one object with different options as opposed to two different objects.

kevinsung: What if the user specifies neither?

cirq/ops/display.py
kevinsung: I've made it hopefully clearer.

cirq/ops/display.py
kevinsung: I renamed it to `num_samples`.

cirq/ops/display.py
kevinsung: Done.

cirq/ops/display.py
kevinsung: Done

cirq/ops/display.py
kevinsung: I'm leaning toward defaulting to idealized. There would be one keyword argument, `samples`, and specifying a value for it would cause it to return the SamplesDisplay.

kevinsung: I've defined the helper function.

cirq/ops/display.py
```diff
@@ -0,0 +1,206 @@
...
+        return self._pauli_string, self._key
+
+
+def pauli_string_expectation(
+        pauli_string: 'pauli_string.PauliString',
+        num_samples: Optional[int] = None,
+        key: Hashable='') -> Union[ApproxPauliStringExpectation,
+                                   PauliStringExpectation]:
+    if num_samples is None:
+        return PauliStringExpectation(pauli_string, key=key)
```
Strilanc: I think we should go deeper and actually merge the two classes. This is going to be a really common thing that happens with displays.

kevinsung: Oh, you mean have PauliStringExpectation inherit from both SamplesDisplay and WaveFunctionDisplay? That sounds good.

Strilanc: Yes. It should be possible to do both at the same time.

kevinsung: What should the abstract method `num_samples` of SamplesDisplay return if the user doesn't specify the number of samples? Seems like we need to choose an arbitrary default value. 1? 1000?

kevinsung: Oh I just thought of one potential problem with having classes inherit from both SamplesDisplay and WaveFunctionDisplay. When running these on a simulator, how will the simulator know whether to compute the value of the display using the samples method or the wavefunction method?

kevinsung: Or do you think `num_samples` of SamplesDisplay should be allowed to return None? I don't think that's a good idea.

Strilanc: I think that this method should return something that isn't a samples display when given num_samples=None.

It would be ideal if we didn't need two distinct classes for the same concept, but if that's necessary behind the scenes then so be it. But I'd prefer that the user never realize that this is the case.

kevinsung: So you're still insisting that the user instantiate Pauli string expectation displays by calling a class constructor instead of a normal method, right? Okay then, I will have to think about how to make that possible. As I mentioned above, inheritance doesn't seem like it will work here.

kevinsung: Or, perhaps _all_ Displays should be able to have their values computed from the wavefunction. They could then opt into having a version that could be computed from samples.

Strilanc: It's fine if they call a method. Preferable, even. It makes it easier to isolate them from changes to the objects behind the scene.

kevinsung: That's what I have now ... what exactly is your complaint?

In any case, I'm thinking about renaming WaveFunctionDisplay to just Display and having SamplesDisplay be a subclass of it. The simulator would then always call the wavefunction method to compute the display value. This would let us merge the two PauliStringExpectation classes. What do you think of that?

kevinsung: Actually, there would still be the question of what `num_samples` should return if the user inputs None.

Strilanc: I don't think it makes sense for the sampling display to be a subclass of the wavefunction display. They're distinct input concepts.

I'm happy to have the two classes be separate and exposed via a common method. But it does seem like there's some kind of unifying concept we're missing here.

kevinsung: My reasoning is that the value of any sampling display can in principle be computed from the wavefunction, so we could require that the wavefunction method always be present.

dabacon: I'm confused as to whether Displays are supposed to make it to the hardware or not.  The way this is written seems to suggest that you are only thinking about Display's for simulation?

cirq/ops/display.py
```diff
@@ -0,0 +1,206 @@
...
+    from cirq.ops import pauli_string
+
+
+class SamplesDisplay(raw_types.Operation):
+    """A display whose value is computed from measurement results.
+
+    The value of a SamplesDisplay on some qubits is computed in the following
+    steps:
+        1. Repeat the following some number of times:
+            a. Start with the state which exists just prior to the Moment
```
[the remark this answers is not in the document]
kevinsung: No. See #1257 for how I would compute SamplesDisplay on SimulatesSamples. I intend to put basically the same method on Quantum Engine.

cirq/ops/display.py
```diff
@@ -0,0 +1,206 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Classes to represent displays.
+
+A Display is an operation that signifies extracting some information about the
+qubits it is applied to without actually performing any effect on those qubits.
+Each Display in a circuit has an associated key, and the result of computing
+the values of Displays in a circuit is a dictionary from Display key to
+Display value.
```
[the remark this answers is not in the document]
kevinsung: Only SamplesDisplays can be run on hardware. I'll make that clearer in the documentation.

PR #1223, opened 2018-12-04
Extend value_equality decorator with support for approximate equality

mrwojtek:


mrwojtek:
There are no tests, but I'd like to ask for early comments before creating them.

cirq/value/value_equality.py
```diff
@@ -73,33 +91,62 @@ def _value_equality_hash(self: _SupportsValueEquality) -> int:
...
+                   approximate: bool = False,
                    ) -> Union[Callable[[type], type], type]:
     """Implements __eq__/__ne__/__hash__ via a _value_equality_values_ method.
 
     _value_equality_values_ is a method that the decorated class must implement.
 
+    _value_equality_approximate_values_ is a method that the decorated class
+    might implement if special support for approximate equality is required.
+    This is only used when approximate argument is set. When approximate
+    argument is set and _value_equality_approximate_values_ is not defined,
```
Strilanc: Give an example of how it would be used, in terms of angles.

cirq/value/value_equality.py
```diff
@@ -36,6 +38,22 @@ def _value_equality_values_(self) -> Any:
...
+        supporting `SupportsApproximateEquality` protocol) then they can be
+        given as an iterable.
+
+        If this method is not defined by decorated class,
+        `_value_equality_values_` is going to be used instead.
+
+        Returns:
+            Any type supported by `cirq.approx_eq()`.
+        """
+        pass
```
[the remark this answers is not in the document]
mrwojtek: Done but how does that work? Is that only documentation change? This protocol is not directly used by any class, right?

I still need to keep
        if not hasattr(cls, '_value_equality_approximate_values_'):
            setattr(cls, '_value_equality_approximate_values_', values_getter)
below?

cirq/value/value_equality.py
mrwojtek: Done; Invented new name for angles comparator, WDYT?

cirq/value/value_equality.py
```diff
@@ -18,6 +18,8 @@
 
 from typing_extensions import Protocol
 
+import cirq
```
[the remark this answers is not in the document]
mrwojtek: Done.

mrwojtek:
Added tests and applied comments, please take a look.

The import change broke python 2.7 tests, do you have advice on this? I saw other parts of cirq just do "import cirq" and python 2.7 tests work?

mrwojtek:
Added tests and applied comments, please take a look.

The import change broke python 2.7 tests, do you have advice on this? I saw other parts of cirq just do "import cirq" and python 2.7 tests work?

mrwojtek:
Added tests and applied comments, please take a look.

The import change broke python 2.7 tests, do you have advice on this? I saw other parts of cirq just do "import cirq" and python 2.7 tests work?

cirq/value/value_equality.py
mrwojtek: I'm thinking about changing this name to `PeriodicEquivalence` actually.

Strilanc:
Python 2.7 is worse at handling cyclic dependencies than python 3. What's happening is that protocols imports value (e.g. in circuit_diagram_info) but value is also importing protocols now. And it is specifically importing value equality.

I think you can work around it by saying exactly `import cirq.protocols`.

mrwojtek:
I managed to get it working! If I make a new class declaration after the last test in test file, python2.7 conversion fails (some kind of indentation error).

cirq/value/value_equality.py
Strilanc: PeriodicValue is better.

cirq/value/value_equality.py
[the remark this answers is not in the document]
Strilanc: Yes, no other changes are needed.

PR #1228, opened 2018-12-07
Converts qcircuit module to protocol style

### LINKED ISSUE #1227: Fix QCircuit not wrapping unknown gates' text into a \gate box
```python
class G(cirq.Gate): 
    def __repr__(self): 
        return 'GGGGGGGGGGG' 
qs = cirq.LineQubit.range(10)
print(cirq.contrib.qcircuit.circuit_to_latex_using_qcircuit(cirq.Circuit.from_ops(G().on(a, b) for a, b in zip(qs, qs[1:]))))
```

```
\Qcircuit @R=1em @C=0.75em {
 \\
 &\lstick{\text{0}}& \qw&\text{GGGGGGGGGGG} \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &\qw\\
 &\lstick{\text{1}}& \qw&\text{\#2}         \qw\qwx&\text{GGGGGGGGGGG} \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &\qw\\
 &\lstick{\text{2}}& \qw&                   \qw    &\text{\#2}         \qw\qwx&\text{GGGGGGGGGGG} \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &\qw\\
 &\lstick{\text{3}}& \qw&                   \qw    &                   \qw    &\text{\#2}         \qw\qwx&\text{GGGGGGGGGGG} \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &\qw\\
 &\lstick{\text{4}}& \qw&                   \qw    &                   \qw    &                   \qw    &\text{\#2}         \qw\qwx&\text{GGGGGGGGGGG} \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &\qw\\
 &\lstick{\text{5}}& \qw&                   \qw    &                   \qw    &                   \qw    &                   \qw    &\text{\#2}         \qw\qwx&\text{GGGGGGGGGGG} \qw    &                   \qw    &                   \qw    &                   \qw    &\qw\\
 &\lstick{\text{6}}& \qw&                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &\text{\#2}         \qw\qwx&\text{GGGGGGGGGGG} \qw    &                   \qw    &                   \qw    &\qw\\
 &\lstick{\text{7}}& \qw&                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &\text{\#2}         \qw\qwx&\text{GGGGGGGGGGG} \qw    &                   \qw    &\qw\\
 &\lstick{\text{8}}& \qw&                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &\text{\#2}         \qw\qwx&\text{GGGGGGGGGGG} \qw    &\qw\\
 &\lstick{\text{9}}& \qw&                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &                   \qw    &\text{\#2}         \qw\qwx&\qw\\
 \\
}
```

image

bryano:
* Adds `get_circuit_diagram_info` and `qubit_namer` (and removes `qubit_name_suffix`) argument to `Circuit.to_text_diagram_drawer`
* Removes `_QCircuitOperation`, `QCircuitDiagrammable` and subclasses thereof
* Replaces above classes with `get_qcircuit_diagram_info` to try to wrangle an operation into qcircuit code. 

Fixes #1227


Strilanc:
This is doing way more than it says on the tin. Can you describe the changes?

bryano:
More description needed?

bryano:
Note that I kept all the same tests as before, with the only modification being the correction to enclose the gate names with "\gate{}".

cirq/circuits/circuit.py
```diff
@@ -1542,12 +1546,21 @@ def _formatted_exponent(info: protocols.CircuitDiagramInfo,
     return s
 
 
-def _draw_moment_in_diagram(moment: Moment,
-                            use_unicode_characters: bool,
-                            qubit_map: Dict[ops.QubitId, int],
-                            out_diagram: TextDiagramDrawer,
-                            precision: Optional[int],
-                            moment_groups: List[Tuple[int, int]]):
+
```
Strilanc: nit: two lines between module methods

cirq/protocols/circuit_diagram_info.py
```diff
@@ -233,3 +247,7 @@ def circuit_diagram_info(val: Any,
     raise TypeError("object of type '{}' does have a _circuit_diagram_info_ "
                     "method, but it returned NotImplemented.".format(type(val)))
 # pylint: enable=function-redefined
+
+
+OPERATION_DIAGRAM_INFO_GETTER = (Callable[
```
Strilanc: This isn't a protocol. It's not a methods or set of methods that, if a class implements, gives it new functionality in the eyes of other parts of the code base.

This is a short name for a specific type of method. I wouldn't even give this a name, I'd just use it directly it in signatures.

cirq/protocols/circuit_diagram_info.py
```diff
@@ -120,6 +121,19 @@ def __repr__(self):
                 self.precision,
                 self.qubit_map))
 
+    def as_dict(self):
```
Strilanc: Replace this with a `__copy__` method or `copy` method and mutation of the copy.

cirq/contrib/qcircuit/qcircuit_test.py
```diff
@@ -12,25 +12,24 @@
...
-    assert qx == qx2
-    assert qx != qy
-    assert sorted([x, qy, qx]) == [x, qx, qy]
+import cirq.contrib.qcircuit as ccq
+
+
+def assert_qcircuit_diagrams_equal(actual: str, expected: str):
+    zipped_lines = itertools.zip_longest(
+            actual.split(r'\\'), expected.split(r'\\'), fillvalue='')
+    for line_number, (actual_line, expected_line) in enumerate(zipped_lines):
```
Strilanc: `cirq.testing.highlight_text_differences`

cirq/protocols/circuit_diagram_info.py
bryano: Yea, this was just meant as a macro for concision. By "protocol style", I didn't mean that I introduced a new protocol, but more simply that I got rid of all the classes and used a functional style like that of, say, `decompose`.

Removed.

cirq/contrib/qcircuit/qcircuit_test.py
bryano: In this case, it's not necessary that the strings be exactly the same, because the "diagrams" are really the code to produce the diagrams, so most whitespace is irrelevant. This actually strips the entries of the matrix. It's not perfect, because it doesn't account for escaped ampersands, but it should work even when they're present. It also doesn't normalize whitespace within the entries. A few possibilities:

* Just test exact equality of the strings and use `highlight_text_differences`. This would be the prettiest by overly exacting.
* Test for equality in the current way, but use `highlight_text_differences` in reporting. Easiest way but a little confusing.
* Extend `highlight_text_differences` to split each line and strip surrounding whitespace of entries.

cirq/protocols/circuit_diagram_info.py
bryano: Done.

cirq/circuits/circuit.py
bryano: Fixed.

cirq/contrib/qcircuit/qcircuit_test.py
Strilanc: Whitespace is sometimes significant in latex and sometimes not. E.g. "\times a" is not the same thing as "\timesa". I also think of these tests as "I want it to be exactly this; the alignment is important" tests.

bryano: Done.

PR #1234, opened 2018-12-07
Added docker installation instructions and Dockerfile.

zoltanegyed:
Updated installation instructions on how to use docker, and added a Dockerfile which clones the repo for local editing.


vtomole:
Closing: https://github.com/quantumlib/Cirq/pull/957

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM ubuntu
+
+# Install dependencies.
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
+     python python3-pip python3-tk texlive-latex-base latexmk git emacs locales
```
vtomole: Don't forget Vim.

docs/install.md
```diff
@@ -106,3 +107,34 @@ That way you have control over when a breaking change affects you.
     # │        │        │        │        │        │        │        │        │        │        │
     # (1, 0)───(1, 1)───(1, 2)───(1, 3)───(1, 4)───(1, 5)───(1, 6)───(1, 7)───(1, 8)───(1, 9)───(1, 10)
     ```
+
+
+### Installing on Docker
+
+This will create a Docker image with the source code cloned from Github. This setup can be practical for development when you don't want to change any existing installation on your host machine.
+
+0. On your host sytem docker needs to be installed.
```
vtomole: Nit: Change to "Install Docker on your host system" with a link to the Docker installation page: https://docs.docker.com/install/#supported-platforms

Dockerfile
Strilanc: Are emacs or vim actually dependencies required for it to work?

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+# Configure UTF-8 encoding.
+RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && locale-gen
+ENV LANG en_US.UTF-8  
+ENV LANGUAGE en_US:en  
+ENV LC_ALL en_US.UTF-8 
+
+
+# Get a current copy of Cirq from Github.
+RUN git clone https://github.com/quantumlib/Cirq
```
Strilanc: Oh, this is a docker file for the dev environment? Is that what people normally expect the dockerfile to install?

Dockerfile
vtomole: No, but you might need an editor if you are in the container.
```
vtomole@vtomole:~/Cirq$ docker run -it cirq/cirq bash
root@f8928a0fb7e4:/usr/src/app# ls
#README.rst#	 Dockerfile~  WORKSPACE			   check		   dev_tools		  python2.7-requirements.txt  setup.cfg
AUTHORS		 LICENSE      apt-system-requirements.txt  cirq			   docs			  release.md		      setup.py
CONTRIBUTING.md  MANIFEST.in  ascii.py			   cirq.egg-info	   examples		  requirements.txt	      style.md
Dockerfile	 README.rst   ascii.py~			   continuous-integration  python2.7-generate.sh  res.txt		      third_party
root@f8928a0fb7e4:/usr/src/app# vim README.rst 
bash: vim: command not found

```

Dockerfile
zoltanegyed: The main goal is for dev environment, although it can be used for deployment too. The goal was to create an easy way to spin up a dev environment, without messing with host settings.

docs/install.md
```diff
@@ -106,3 +107,34 @@ That way you have control over when a breaking change affects you.
     # │        │        │        │        │        │        │        │        │        │        │
     # (1, 0)───(1, 1)───(1, 2)───(1, 3)───(1, 4)───(1, 5)───(1, 6)───(1, 7)───(1, 8)───(1, 9)───(1, 10)
     ```
+
+
+### Installing on Docker
+
+This will create a Docker image with the source code cloned from Github. This setup can be practical for development when you don't want to change any existing installation on your host machine.
```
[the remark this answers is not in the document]
zoltanegyed: Done

Dockerfile
zoltanegyed: Done

docs/install.md
```diff
@@ -106,3 +107,34 @@ That way you have control over when a breaking change affects you.
     # │        │        │        │        │        │        │        │        │        │        │
     # (1, 0)───(1, 1)───(1, 2)───(1, 3)───(1, 4)───(1, 5)───(1, 6)───(1, 7)───(1, 8)───(1, 9)───(1, 10)
     ```
+
+
+### Installing on Docker
```
Strilanc: Because this creates a development environment, this should be moved into the developer documentation. https://github.com/quantumlib/Cirq/blob/master/docs/development.md

We may want a different dockerfile for just the user install.

vtomole: I think this Dockerfile should be the user install; which could be done by getting rid of the `git clone` RUN command and adding `cirq` to the `pip install` RUN command. As far as i know, none of us use Docker for development. If needed, we could have the appropriate Dockerfile in `dev_tools`.

Strilanc: Yes, putting the dev dockerfile in dev_tools and having this one be a user install is a good idea.

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ubuntu
```
vtomole: Copy `Cirq/examples` into this container so users can execute commands like `docker run -it cirq/cirq python3 bell_inequality.py.` Big nit: making `python` the default will look better. Currently,  `docker run -it cirq/cirq python bell_inequality.py` doesn't work because python2 is used instead of python3 (by using `FROM python:3` instead).

To copy:
```
ADD examples/*.py /examples/
WORKDIR /examples
```

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+# Install dependencies.
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
+     python python3-pip python3-tk texlive-latex-base latexmk git vim locales
```
vtomole: Can get rid of `python` if using `FROM python:3`. Where do we use `python3-tk`?  Forgot `emacs` :)

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+# Configure UTF-8 encoding.
+RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && locale-gen
+ENV LANG en_US.UTF-8  
+ENV LANGUAGE en_US:en  
+ENV LC_ALL en_US.UTF-8 
+
+# Install Cirq with the needed Python libraries.
+RUN pip3 install numpy typing_extensions networkx sortedcontainers google-api-python-client cirq
```
vtomole: Installing Cirq installs everything in the requirements.txt with it. `RUN pip3 install cirq` is enough.

Dockerfile
zoltanegyed: python3-tk was listed as a prerequisite.

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM ubuntu
+
+# Install dependencies.
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
+     python3-pip python3-tk texlive-latex-base latexmk git emacs vim locales
```
vtomole: Where do we use `python3-tk`?

docs/install.md
```diff
@@ -106,3 +107,34 @@ That way you have control over when a breaking change affects you.
...
+
+This will create a Docker image containing Cirq's source code. It will isolate a Cirq installation from the rest of the system.
+
+0. [Install Docker](https://docs.docker.com/install/#supported-platforms) on your host sytem.
+
+1. Build the docker image for your system:
+    ```bash
+    git clone https://github.com/quantumlib/Cirq
```
Strilanc: Maybe I've fundamentally misunderstood how people use docker, but it seems very odd for the user docker setup to involve cloning the source code? Shouldn't it be more along the lines of "here is a docker file that includes cirq==0.4.* as a dependency, and here is the dependency line you can paste into a docker file to include cirq"?

dev_tools/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+ENV LC_ALL en_US.UTF-8 
+
+# Make python3 default
+RUN rm -f /usr/bin/python && ln -s /usr/bin/python3 /usr/bin/python
+
+# Get a current copy of Cirq from Github.
+RUN git clone https://github.com/quantumlib/Cirq
+
+# Install the needed Python libraries.
+RUN pip3 install numpy typing_extensions networkx sortedcontainers google-api-python-client
```
vtomole: pip install from from Cirq/requirements.txt instead.

Dockerfile
Strilanc: It's needed for matplotlib on some systems IIRC.

dev_tools/Dockerfile
Strilanc: Also `cirq/contrib/contrib-requirements.txt` and `dev_tools/conf/pip-list-dev-tools.txt`

docs/install.md
vtomole: That's true. All that's needed is the Dockerfile for manual building. We could host the Dockerfile independently so users can download it to avoid cloning the whole repo just to access the Dockerfile or; based on my comments from https://github.com/quantumlib/Cirq/pull/957, create a Docker Hub account so that the users don't have to clone the source or download the Dockerfile to build and use the image. They can just `docker pull quantumlib/cirq` and be ready to go. The latter is standard in the Docker community.

vtomole: We can also host the image independently.

dev_tools/Dockerfile
zoltanegyed: Done

docs/install.md
zoltanegyed: I have updated the development Dockerfile to use the requirements files, and the user Dockerfile now just pulls it it via cirq package. I suggest we do the prebuilt image in a different PR.

zoltanegyed:
Strilanc, the changes you have requested do not seem to show up in this view. Can you please elaborate what change you'd like to make?

zoltanegyed:
I have modified the top level Dockerfile (for deployment) to be able to have the examples subdirectory independently whether the Dockerfile is copied elsewhere. Now it stands on its own, without depending on the subdirectories created by a git clone which included this file. Now it gets the examples subdirectory inside the container, as part of the build process.

Issue #1226, opened 2018-12-06
Generalize TextDiagramDrawer to allow drawing outline boxes


Strilanc:
This issue is harder than it sounds. But I do think it's important for us to start being able to draw multi-qubit boxes in text diagrams and in the latex output.

1. Box interiors should cover lines but not text entries. The layer order is lines then boxes then text.

2. When a Box border intersects a line, it should produce the appropriate unicode character.

    ```
    ┌─┬┐
    │ ││
    ├─┼┤
    └─┴┘
    ┏━┳┓
    ┃ ┃┃
    ┣━╋┫
    ┗━┻┛
    ```

    - Including if the line is emphasized and the box is not
        ```
        ┍┯┑
        ┝┿┥
        ┕┷┙
        ┎┰┒
        ┠╂┨
        ┖┸┚
        ```

3. (Opinion) boxes may have their corners at half-coordinates, so that they can be used to surround a text cell without inserting an extra column or row.

4. A useful intermediate concept is a grid made up of a bunch of common cells. Each cell specifies a min width, min height, text, left/up/down/right line character, and a center line character. The more general render method can produce a bunch of cells, and then the cell rendered can handle picky layout details like sizing.

Untested code for 4:

```python
TextDiagramCell = NamedTuple('TextDiagramCell', [
    ('center', str),
    ('left_line', str),
    ('right_line', str),
    ('up_line', str),
    ('down_line', str),
    ('text', str),
    ('min_width', int),
    ('min_height', int),
])


def render_cell_lines(cell: TextDiagramCell,
                      width: int,
                      height: int) -> List[str]:
    out = collections.defaultdict(lambda: ' ')
    cy = height // 2
    cx = width // 2

    if cell.left_line:
        for x in range(cx):
            out[(x, cy)] = cell.left_line
    if cell.right_line:
        for x in range(cx + 1, width):
            out[(x, cy)] = cell.right_line
    if cell.up_line:
        for y in range(cy):
            out[(cx, y)] = cell.up_line
    if cell.down_line:
        for y in range(cy + 1, height):
            out[(cx, y)] = cell.down_line
    if cell.center:
        out[cx, cy] = cell.center

    lines = cell.text.split('\n')
    lines_height = len(lines)
    lines_width = max(len(line) for line in lines)
    off_x = (width - lines_width) // 2
    off_y = (height - lines_height) // 2
    for dy, line in enumerate(lines):
        for dx, c in enumerate(line):
[...]

PR #1236, opened 2018-12-08
Adding, shifting, and stacking for TextDiagramDrawer

bryano:


cirq/circuits/text_diagram_drawer.py
```diff
@@ -243,6 +260,73 @@ def render(self,
...
+        shifted_vertical_padding = {y + dy: padding
+                for y, padding in self.vertical_padding.items()}
+        return self.__class__(
+                entries=shifted_entries,
+                horizontal_lines=shifted_horizontal_lines,
+                vertical_lines=shifted_vertical_lines,
+                horizontal_padding=shifted_horizontal_padding,
+                vertical_padding=shifted_vertical_padding)
+
+    def __iadd__(self, other: 'TextDiagramDrawer') -> 'TextDiagramDrawer':
```
Strilanc: I don't think that this is there is a natural unambiguous meaning for "adding" two diagrams. E.g. I would have expected addition to do horizontal concatenation, as it does for strings and lists in python. I think we should give this a name instead of an operator e.g. "merge".

cirq/circuits/text_diagram_drawer.py
```diff
@@ -243,6 +260,73 @@ def render(self,
...
+        self.vertical_padding.update(other.vertical_padding)
+        return self
+
+    def __add__(self, other: 'TextDiagramDrawer') -> 'TextDiagramDrawer':
+        summed = self.copy()
+        summed += other
+        return summed
+
+    @staticmethod
+    def vstack(top: 'TextDiagramDrawer', bottom: 'TextDiagramDrawer',
```
Strilanc: Needs a docstring.

Why is there a 'padding resolver' argument?

Strilanc: In numpy vstack takes an iterable of tensors and merges them into one tensor. Shouldn't this then take a list of diagrams, instead of two separate diagram arguments?

bryano: To determine what the padding should be when the diagrams to be vertically stacked have different values for horizontal padding in the same column.

bryano: Done.

bryano: Docstring added.

cirq/circuits/text_diagram_drawer.py
bryano: Changed to "update".

bryano:
Do you see where the circular import is being introduced?

bryano:
I'm tempted to remove the use of `value_equality` to get this passing the tests, but it'd be nice to have it in and I'd like to understand why the tests are failing.

cirq/circuits/text_diagram_drawer.py
```diff
@@ -243,6 +262,90 @@ def render(self,
...
+
+    @classmethod
+    def vstack(cls,
+               diagrams: Iterable['TextDiagramDrawer'],
+               padding_resolver: Callable[[Sequence[int]], int]=max):
+        """Vertically stack text diagrams.
+
+        Args:
+            diagrams: The diagrams to stack, ordered from bottom to top.
+            padding_resolver: A function that takes a list of paddings
```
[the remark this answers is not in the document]
bryano: Consider the following:
```
d = TextDiagramDrawer()
d.write(0, 0, 'a')
d.write(1, 0, 'b')
d1, d2, d3, d4 = (d.copy() for _ in (1, 2, 3, 4))
d2.force_horizontal_padding_after(0, 3)
d3.force_horizontal_padding_after(0, 5)
d4.force_horizontal_padding_after(0, 2)
stacked = TextDiagramDrawer.vstack((d1, d2, d3, d4))
```
What should `stacked.horizontal_padding[0]` be? The smallest padding specified (2) or the largest (5)? The first (3) or the last (2)? The rounded mean? I imagine the answer will be context dependent, so I allowed the caller to specify how to resolve potential discrepancies.

Both `max` and `min` seem like reasonable defaults to me. An alternative default is to through an error  if different paddings are specified, in order to force the caller to say explicitly how to resolve them.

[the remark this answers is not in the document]
Strilanc: Oh I see, the issue is that columns may have incompatible padding requirements.

I would default the padding resolver to `None` and raise an error when conflicts occurs. A conflict occurs if the set of horizontal paddings for any column (with None meaning no forced padding) has a cardinality greater than 1.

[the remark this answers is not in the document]
bryano: To clarify, if one or more diagrams specify the same padding for a given column and one or more diagrams don't specify any padding for that column, should an error be thrown?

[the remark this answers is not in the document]
Strilanc: Correct.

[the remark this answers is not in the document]
bryano: Done.

Strilanc:
Conflicts with the text diagram changes I made last week. Needs to be updated.

bryano:
Done

cirq/circuits/text_diagram_drawer.py
```diff
@@ -305,3 +331,126 @@ def render(self,
...
+def _same_element_or_throw_error(elements: Sequence[Any]):
+    """Extract an element or throw an error.
+
+    Args:
+        elements: A sequence of something.
+
+    copies of it. Returns None on an empty sequence.
+
+    Raises:
+        ValueError: The sequence constai
```
Strilanc: constai?

cirq/circuits/text_diagram_drawer_test.py
```diff
@@ -15,8 +15,40 @@
...
+
+    Args:
+        actual: The text diagram.
+        desired: The desired rendering as a string.
+        **kwargs: Keyword arguments to be passed to actual.render.
+    """
+    actual_diagram = actual.render(**kwargs)
+    desired_diagram = desired
+    assert actual_diagram == desired_diagram, (
+    "Diagram's rendering differs from the desired rendering.\n"
```
Strilanc: Nit: indent

cirq/circuits/text_diagram_drawer_test.py
```diff
@@ -159,3 +191,165 @@ def test_multiline_entries():
...
+    hstacked_min = TextDiagramDrawer.hstack((d, dd), padding_resolver=min)
+    expected = """
+AB D
+
+ C E  F
+    """.strip()
+    assert_has_rendering(hstacked_min, expected)
+
+def test_drawer_eq():
+    assert TextDiagramDrawer().__eq__(23) == NotImplemented
```
Strilanc: Use `EqualsTester` to properly exercise the eq, neq, etc methods.

cirq/circuits/text_diagram_drawer.py
```diff
@@ -59,12 +62,35 @@ class TextDiagramDrawer:
...
+                else dict(horizontal_padding)
+                ) # type: Dict[int, Union[int, float]]
+        self.vertical_padding = (dict() if vertical_padding is None
+                else dict(vertical_padding)
+                ) # type: Dict[int, Union[int, float]]
+
+    def __eq__(self, other):
+        if not isinstance(other, type(self)):
+            return NotImplemented
+        attrs = ('entries', 'horizontal_lines', 'vertical_lines',
```
Strilanc: Looks like this is a prime candidate for `@cirq.value_equality` (well, `@value.value_equality`).

cirq/circuits/text_diagram_drawer_test.py
bryano: Done

cirq/circuits/text_diagram_drawer.py
bryano: fixed

cirq/circuits/text_diagram_drawer.py
bryano: I completely agree, but when I try to use that I get circular import errors that I don't know how to fix.

Strilanc: Might be a cycle via `cirq.protocols`... What's the stack trace?

cirq/circuits/text_diagram_drawer.py
```diff
@@ -83,14 +83,15 @@ def __init__(self,
                 else dict(vertical_padding)
                 ) # type: Dict[int, Union[int, float]]
 
+    def _value_equality_values_(self):
+        attrs = ('entries', 'horizontal_lines', 'vertical_lines',
+                 'horizontal_padding', 'vertical_padding')
+        return tuple(getattr(self, attr) for attr in attrs)
+
     def __eq__(self, other):
```
Strilanc: Delete the `__eq__` method when using value equality.

cirq/circuits/text_diagram_drawer.py
bryano: ../../.virtualenvs/quantumlib/lib/python3.6/site-packages/py/_path/local.py:668: in pyimport
    __import__(modname)
cirq/__init__.py:21: in <module>
    from cirq.circuits import (
cirq/circuits/__init__.py:17: in <module>
    from cirq.circuits.text_diagram_drawer import (
cirq/circuits/text_diagram_drawer.py:21: in <module>
    from cirq import value
cirq/value/__init__.py:26: in <module>
    from cirq.value.periodic_value import (
cirq/value/periodic_value.py:19: in <module>
    import cirq.protocols
cirq/protocols/__init__.py:16: in <module>
    from cirq.protocols.apply_unitary import (
cirq/protocols/apply_unitary.py:23: in <module>
    from cirq import linalg
cirq/linalg/__init__.py:28: in <module>
    from cirq.linalg.decompositions import (
cirq/linalg/decompositions.py:267: in <module>
    @value.value_equality
E   AttributeError: module 'cirq.value' has no attribute 'value_equality'

cirq/circuits/text_diagram_drawer.py
bryano: Yea, that's just temporary because the class isn't actually dressed with `value_equality` yet.

Strilanc: We need to resolve the dependency cycle so it can be annotated with it.

bryano: Done

bryano:
@Strilanc Tests are only failing in pytest2. Is that worth fixing given that Python 2 support is going to be removed?

vtomole:
@bryano Depends on if you want these features in the next release as we are removing Python 2 after it.

bryano:
I think I'll wait it out. pytest2 passes on my machine, so I can't even replicate the failures to address them.

bryano:
@Strilanc Tests passing now that pytest2 is gone.

vtomole:
@bryano @Strilanc Checks are passing with removal of python 2. Time to merge?

vtomole:
Related to: https://github.com/quantumlib/Cirq/issues/1226

bryano:
@Strilanc I noticed that fsim_gate.py imports cirq (rather than its submodules). Is there a reason for that?

bryano:
Also, what do you think about renaming either the file 'value_equality.py' or the decorator 'value_equality'. I think a lot of the import difficulties are caused by there being a module and method of the same name.

vtomole:
@Strilanc  Looks like https://travis-ci.com/quantumlib/Cirq/jobs/210593629 format is running into code that wasn't touched by this pull request.

cirq/circuits/text_diagram_drawer.py
```diff
@@ -305,3 +330,124 @@ def render(self,
...
+                              horizontal_padding=self.horizontal_padding)
+
+    def shift(self, dx: int = 0, dy: int = 0) -> 'TextDiagramDrawer':
+        self._transform_coordinates(lambda x, y: (x + dx, y + dy))
+        return self
+
+    def shifted(self, dx: int = 0, dy: int = 0) -> 'TextDiagramDrawer':
+        return self.copy().shift(dx, dy)
+
+    def update(self, other: 'TextDiagramDrawer') -> 'TextDiagramDrawer':
```
Strilanc: This method needs a better name. It overwrites entries but adds lines? Looks like `draw_on_top_of` or similar.

cirq/linalg/decompositions.py
```diff
@@ -329,8 +329,7 @@ def canonicalize(self, atol: float = 1e-8) -> 'AxisAngleDecomposition':
 
     def _value_equality_values_(self):
         v = self.canonicalize(atol=0)
-        return (value.PeriodicValue(v.angle,
-                                    period=math.pi * 2), v.axis, v.global_phase)
+        return (v.angle % (math.pi * 2), v.axis, v.global_phase)
```
Strilanc: This is an incorrect change! It breaks the approximate comparison logic.

cirq/ops/common_channels.py
```diff
@@ -548,7 +549,7 @@ def phase_damp(gamma: float) -> PhaseDampingChannel:
     return PhaseDampingChannel(gamma)
 
 
-@value.value_equality
+@value_equality
```
Strilanc: This was really the only way to break the cycle? It's quite odd... we really don't want to have packages outside of a tree digging into the guts of that tree.

cirq/contrib/acquaintance/permutation.py
```diff
@@ -112,14 +114,14 @@ class LinearPermutationGate(PermutationGate):
     def __init__(self,
                  num_qubits: int,
                  permutation: Dict[int, int],
-                 swap_gate: ops.Gate=ops.SWAP
-                 ) -> None:
+                 swap_gate: Optional[ops.Gate] = None) -> None:
```
Strilanc: I've fixed the bug in the formatter where it would compare against a too-recent version of master. You should be able to revert almosta ll the formatting changes.

cirq/contrib/acquaintance/permutation.py
```diff
@@ -89,7 +90,8 @@ def _circuit_diagram_info_(self, args: protocols.CircuitDiagramInfoArgs
 class SwapPermutationGate(PermutationGate):
     """Generic swap gate."""
 
-    def __init__(self, swap_gate: ops.Gate=ops.SWAP):
+    def __init__(self, swap_gate: Optional[ops.Gate] = None):
+        swap_gate = ops.SWAP if swap_gate is None else swap_gate
```
Strilanc: This functional change is not related to the PR.

cirq/contrib/acquaintance/permutation.py
```diff
@@ -35,9 +35,10 @@ class PermutationGate(ops.Gate, metaclass=abc.ABCMeta):
             qubits (e.g. SWAP or fermionic swap).
     """
 
-    def __init__(self, num_qubits: int, swap_gate: ops.Gate=ops.SWAP) -> None:
+    def __init__(self, num_qubits: int,
+                 swap_gate: Optional[ops.Gate] = None) -> None:
         self._num_qubits = num_qubits
-        self.swap_gate = swap_gate
+        self.swap_gate = ops.SWAP if swap_gate is None else swap_gate
```
Strilanc: This functional change is not related to the PR.

cirq/ops/common_channels.py
bryano: Better now?

cirq/linalg/decompositions.py
bryano: New cycle avoidance technique makes this irrelevant.

cirq/circuits/text_diagram_drawer.py
bryano: How about `superimpose`? In any case, it would be good for the name to indicate (as I think `update` does) that the changes are made in place. So maybe `superimpose_inplace` or `inplace_superimpose`?

bryano:
Okay, I think at this point the only two outstanding items are the cycle-avoidance and the `update` renaming.

cirq/contrib/acquaintance/permutation.py
bryano: This was a hack to avoid a mypy complaint that doesn't seem to be happening anymore.

cirq/circuits/text_diagram_drawer.py
Strilanc: Just `superimpose` is great. You can add `superimposed` as a variant that operates on a copy, similar to `shift` vs `shifted`.

Strilanc:
I agree.

It looks like the cycle is much better with the small number of import changes you're making now.

cirq/ops/common_channels.py
Strilanc: Yes

bryano:
Great. Just let me know what to rename `update` and then it's good to go.

Strilanc:
`superimpose` and a copy variant `superimposed`

bryano:
Done

bryano:
@Strilanc ping

dabacon:
This seems to have grown stale.  I'm closing.  If it can be salvaged, please open a new PR.

bryano:
I think this is ready to go and just awaiting final approval from @Strilanc

dabacon:
@Strilanc this has been sitting on you for 18 days now, could you take a look?

Strilanc:
Close as stale.

bryano:
@Strilanc I thought this was ready to merge. I'll get it back to up to sync if it'll get reviewed.

bryano:
@Strilanc PTAL

bryano:
@Strilanc ping

vtomole:
Does this close https://github.com/quantumlib/Cirq/issues/1226?

bryano:
No, but #1240 was a start towards #1226.

vtomole:
@bryano This should have been merged by now in my opinion. @Strilanc Is there a reason you are ambivalent toward this being in Cirq? I'll bring this up in the sync tommorrow.

Strilanc:
Darn history size page limits.

PR #1238, opened 2018-12-10
Moved __call__ up to OptimizationPass

bryano:


Strilanc:
This is a good change, but I think we should delete `OptimizationPass` instead of modifying it. It was a mistake for me to add it in the first place, because it's nothing but a definition of a particular kind of Callable (and not even a complicated one).

bryano:
I opened up an issue (#1241) to remove it; could we merge this in the meantime?

PR #1239, opened 2018-12-10
Renders swaps as crossing lines in qcircuit

bryano:


cirq/contrib/qcircuit/macros.py
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Tuple
+
+def line_macro(end: Tuple[int, int],
```
Strilanc: Add docstring

cirq/contrib/qcircuit/optimizers.py
```diff
@@ -0,0 +1,37 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from cirq import circuits, ops
+
+class _PadAfterSwapGates(circuits.OptimizationPass):
```
Strilanc: I'm not sure I like this optimization pass. Is there some other way to do this that's more coupled to the method that produces the qcircuit diagram, or to the swaps themselves?

cirq/protocols/circuit_diagram_info.py
```diff
@@ -43,23 +45,30 @@ def __init__(self,
                 (unless it's equal to 1). For example, the square root of X gate
                 has a text diagram exponent of 0.5 and symbol of 'X' so it is
                 drawn as 'X^0.5'.
-            connected: Whether or not to draw a line connecting the qubits.
+            vconnected: Whether or not to draw a vertical line connecting the
```
Strilanc: I don't think this is the right way to go about specifying more complicated drawings. I think we want to have options that are more along the lines of opting into particular functionality as opposed to opting out.

For example, in text diagrams we render disconnected swaps as little Xs joined by vertical lines. That's unlikely to change. Does that mean that the swap gate needs to understand the layout of the qubits in order to set the vconnected field? What if we show the connecting line in text diagrams but not in latex diagrams?

bryano: `vconnected` is just a renaming of `connected` to distinguish it from the new `hconnected`. All of the non-qcircuit text diagram code works exactly as before, setting `vconnected = hconnected = True` by default. For the qcircuit, `vconnected=False` only for gates that are to be drawn using `\multigate` (a feature that predates this PR). `hconnected` serves an analogous purpose for the swaps. In particular, this PR as-is "show[s] the connecting line in text diagrams but not in latex diagrams". 

Could you say more about your opt-in vs. opt-out distinction? It seems to me that the "functionality" here could be "include the line" just as well as "don't include the line".

bryano:
Could you say more about your "bucket of features" point? The essence of the PR is the following:

1) Introduction of `TextDiagramDrawer.horizontal_occlusions'.
2) Introduction of `CircuitDiagramInfo.hconnected`.
3) The optimization pass to pad after swaps.
4) `line_macro`.
5) `swap_qcircuit_diagram_info`.

Only (1) and (2) change the main code base in any way, and they seem like relatively minor additions.

cirq/contrib/qcircuit/macros.py
bryano: Done.

cirq/contrib/qcircuit/optimizers.py
Strilanc: Yes, I feel like if we kept implementing features this way we'd have a giant mess in no time. We need something simpler that includes this use case as a sub-case.

This feels like code that a user would add on to make the library work the way they like, and it would do well at that. But in the library itself it seems like a piece that doesn't play well with the other pieces.

cirq/protocols/circuit_diagram_info.py
Strilanc: This is certainly a straightforward way to communicate to the qcircuit drawer "please include \qwx or \qw or both or neither". But does it also apply to the text diagrams and to hypothetical 3d diagrams?

I think what I'm getting at is that we want a gate to have the ability to say "don't you dare *touch* my area in diagrams of style X. I will handle everything there.". I don't want to keep hacking on individual pieces to make individual gates work. More and more we're workaround around the diagramming system being limited in what it was designed to do. We need to redesign the system before our code is nothing but workarounds. This particular workaround isn't the problem in and of itself, it's just the straw breaking the camel's back.

bryano: That's fair enough. The fact that `TextDiagramDrawer` supports both the text diagrams and qcircuit diagrams is both nice in that it allows code re-use and not nice in that it seems forced when the logic of the two is slightly different.

I'm happy to help out with a more thorough overhaul, and can think more about what that should look like. I opened #1245 to host a discussion on the subject. It would help if you could enumerate the issues you see there. Could we do this one last workaround in the meantime?

Strilanc:
This PR isn't a bucket of features by itself, it's one feature in a growing bucket.

cirq/contrib/qcircuit/optimizers.py
bryano: This is exactly code that a user (me) added on to make the library work the way I wanted. I care less that it be in the library because it's easy enough to tack on afterwards (so long as `circuit_to_latex_using_qcircuit` takes an `optimizers` argument). Same goes for everything in this PR except `TextDiagramDrawer.horizontal_occlusions` and `CircuitDiagramInfo.hconnected`.

vtomole:
@Strilanc @bryano Where does this stand since we are planning on getting rid of qcircuit?  https://github.com/quantumlib/Cirq/issues/1492.

bryano:
Are we dropping support for qcircuit or just adding support for qpic?

Strilanc:
Someone has to actually do the qpic work before we can drop anything related to qcircuit, but the issue seems to be languishing. Get this PR synced with master and I can give it another look.

Strilanc:
Could you revert changes to files that were only formatted due to the old format script comparing to master instead of the merge base? It's creating a ton of noise making it hard to review.

bryano:
Sorry about that. The formatting changes got comingled with real changes, so I checked out from master any file that differed only in whitespace, which I think is either the same as undoing the formatting changes or at least close thereto.

Strilanc:
There are still a ton of format-only changes.

I'm not sure of a simple automated way to get rid of the format-only files. Look at the diff's original-file ranges, do incremental formatting on master vs those diffs, and see what matches up? Tricky to write.

bryano:
I went back in manually and got rid of most of the formatting-only changes. LMK if there's still too much.

bryano:
@Strilanc The only failing test is the formatting.

cirq/circuits/text_diagram_drawer.py
```diff
@@ -141,6 +144,9 @@ def horizontal_line(self,
         x1, x2 = sorted([x1, x2])
         self.horizontal_lines.append(_HorizontalLine(y, x1, x2, emphasize))
 
+    def horizontal_occlusion(self, x: int, y: int):
```
Strilanc: Add a docstring

cirq/contrib/qcircuit/optimizers.py
```diff
@@ -0,0 +1,63 @@
...
+            circuit.insert(i + 1, (ops.Moment(),) * padding)
+        padding = max(
+            itertools.chain((0,),
+                            (padding_needed(op, None) for op in circuit[-1])))
+        circuit.append((ops.Moment(),) * padding)
+
+    return optimize_circuit
+
+
+def swap_followed_by_non_swap(first_op: ops.Operation,
```
Strilanc: Docstring. Not clear at all what the results of this method mean or how they are used.

cirq/contrib/qcircuit/optimizers.py
```diff
@@ -0,0 +1,63 @@
...
+# limitations under the License.
+
+import itertools
+
+from typing import Callable, Optional
+
+from cirq import circuits, ops
+
+
+def EnsureFinalMomentEmpty(circuit: circuits.Circuit) -> None:
```
Strilanc: Use `snake_case` for method names.

Doc string.

cirq/contrib/qcircuit/optimizers.py
```diff
@@ -0,0 +1,63 @@
...
+    if circuit[-1]:
+        circuit.append(ops.Moment())
+
+
+def PadBetweenOps(
```
Strilanc: Use `snake_case` for method names.

Doc string.

cirq/contrib/qcircuit/optimizers.py
```diff
@@ -0,0 +1,63 @@
+# Copyright 2018 The Cirq Developers
```
Strilanc: These aren't really "optimizers", they're just utility methods for putting the circuit into a special form.

cirq/contrib/qcircuit/qcircuit_diagram.py
```diff
@@ -51,28 +55,61 @@ def _render(diagram: circuits.TextDiagramDrawer) -> str:
...
     return output
 
 
-def circuit_to_latex_using_qcircuit(
+def circuit_to_qcircuit_diagram(
         circuit: circuits.Circuit,
-        qubit_order: ops.QubitOrderOrList = ops.QubitOrder.DEFAULT) -> str:
-    """Returns a QCircuit-based latex diagram of the given circuit.
+        *,
+        optimizers: Iterable[
```
Strilanc: I would call this something more like "adjusters" or... I dunno. It's sort of a strange argument.

cirq/contrib/qcircuit/qcircuit_diagram.py
```diff
@@ -51,28 +55,61 @@ def _render(diagram: circuits.TextDiagramDrawer) -> str:
...
+            Callable[[circuits.Circuit], None]] = default_optimizers,
+        get_circuit_diagram_info: Optional[
```
Strilanc: I would call this `intercepting_circuit_diagram_info_getter` to be consistent with `intercepting_decomposer` from `cirq.decompose`.

cirq/contrib/qcircuit/qcircuit_diagram.py
```diff
@@ -51,28 +55,61 @@ def _render(diagram: circuits.TextDiagramDrawer) -> str:
...
+            Callable[[ops.Operation, protocols.CircuitDiagramInfoArgs],
+                     protocols.CircuitDiagramInfo]] = None,
+        qubit_namer: Optional[Callable[[ops.Qid], str]] = None,
```
Strilanc: The number of these arguments that are of the form "special case this specific object's behavior for this specific case" is indicative that we should perhaps be defining a protocol for latex output.

cirq/circuits/text_diagram_drawer.py
bryano: Done

cirq/contrib/qcircuit/optimizers.py
bryano: Done

cirq/contrib/qcircuit/optimizers.py
bryano: Done. (This was a vestige of the old optimizer class.)

cirq/contrib/qcircuit/optimizers.py
bryano: Done

cirq/contrib/qcircuit/qcircuit_diagram.py
bryano: Done

cirq/contrib/qcircuit/qcircuit_diagram.py
bryano: I'm fine with calling it something else, but there is a sense in which these optimize the circuit for qcircuit output.

dabacon:
This seems to have grown stale.  I'm closing.  If it can be salvaged, please open a new PR.

bryano:
@Strilanc I know you had some trepidation about this one. If I get it back in sync with master, will it be merged?

Strilanc:
Closed as stale.

Issue #996, opened 2018-10-10
Example algorithms from "Quantum Algorithm Implementation for Beginners" - Linear Systems


verult:
Problem V from https://arxiv.org/pdf/1804.03719.pdf

Implement the Harrow, Hassidim, Lloyd (HHL) algorithm as an example, first the simple instance of

<a href="https://www.codecogs.com/eqnedit.php?latex=A&space;=&space;\begin{bmatrix}&space;1.5&space;&&space;0.5&space;\\&space;0.5&space;&&space;1.5&space;\end{bmatrix}" target="_blank"><img src="https://latex.codecogs.com/gif.latex?A&space;=&space;\begin{bmatrix}&space;1.5&space;&&space;0.5&space;\\&space;0.5&space;&&space;1.5&space;\end{bmatrix}" title="A = \begin{bmatrix} 1.5 & 0.5 \\ 0.5 & 1.5 \end{bmatrix}" /></a>

<a href="https://www.codecogs.com/eqnedit.php?latex=b&space;=&space;\begin{bmatrix}&space;1\\&space;0&space;\end{bmatrix}" target="_blank"><img src="https://latex.codecogs.com/gif.latex?b&space;=&space;\begin{bmatrix}&space;1\\&space;0&space;\end{bmatrix}" title="b = \begin{bmatrix} 1\\ 0 \end{bmatrix}" /></a>

then a more generic version.

verult:
I'd like to work on this

vtomole:
Done. Please close.

Issue #755, opened 2018-07-24
Example code should run unmodified in python 2


Strilanc:
Our example code is supposed to be a demonstration of source code, as opposed to functionality. We can reasonably expect people to copy paste it into a project. Therefore it needs to work in both python 2 and python 3 _without conversion_.

For example, `1/2` evaluates to 0 in python 2 because `/` means integer division unless `from __future__ import division` is specified. And our current tests would miss this fact because the python 3 example code gets transpiled to python 2.

Fixing this requires:

1. Modify the python 2 transpiling script to copy the example code verbatim, instead of transpiling it. However, the test files should still be transpiled.
2. Modify example code to work in both python 2 and python 3.
3. Try to come up with tests that check basic functionality of the examples, since not doing transpiling drastically drops our expectations that the examples are outputting the correct thing.

PR #1244, opened 2018-12-12
HHL example

verult:
Implements https://github.com/quantumlib/Cirq/issues/996

I wrote a different circuit from the paper since I had trouble understanding how the example circuit mapped to the different components of the algorithm described (for example the Phase Estimation part doesn't look like a conventional phase estimation circuit to me; please correct me if I'm wrong!), and I was able to just use 1 register qubit instead of 2. The results check out with the paper, and I hope the new circuit matches the intuition of the algorithm better, but please double check my circuit.


examples/hhl.py
```diff
@@ -0,0 +1,145 @@
...
+
+    s = sum(post_select.values())
+    expected_value = 0
+    for output, occurrence in post_select.items():
+        v = 1 if output == '0' else -1
+        expected_value += v * occurrence / s
+    return expected_value
+
+
+def hhl_circuit(measure_key, observable):
```
Strilanc: Do you think it would be possible to take the matrix as a parameter, instead of requiring it to be a specific value? The example would be much stronger if it could apply to any 2x2 user-specified matrix instead of just a specific precomputed one that we may have over-compiled when producing the circuit.

verult: Challenge accepted!

verult:
In the mean time @Strilanc or anyone else interested, could you take a look at my approach at measuring the observables? I'm not sure if it's the right way to go about it.

kevinsung:
Really you should be using #1221 after it's done.

verult:
@kevinsung just want to clarify: the Displays introduced in your PR only operate on circuits directly? Then yeah I'd prefer to get observables from simulation results

kevinsung:
I'm not sure what you mean by "operate on circuits directly". Displays are operations which you can add to your circuit. Ultimately you'll be able to compute their value using the Simulator, but that is a separate PR (see https://github.com/quantumlib/Cirq/pull/1215)

examples/hhl.py
```diff
@@ -0,0 +1,145 @@
...
+               │                                        │       │
+(2, 0): ───────X────────────────────────────────────────X───────M───
+
+"""
+
+import cirq
+import math
+
+
+def bitstring(bits):
```
viathor: WDYT about adding type annotations in this file?

kevinsung: We decided that examples should run in Python 2 without modifications (https://github.com/quantumlib/Cirq/issues/755); I think that's why the examples have no type annotations.

viathor: Makes sense. Thanks.

verult:
Regarding the Travis failure of `pytest-and-incremental-coverage` - are examples expected to be covered?

vtomole:
@verult  Please call this code on `examples/examples_test.py` to cover it.

verult:
I'm close to figuring out a circuit for 2x2 matrices with only 1 qubit to store the eigenvalue, but the circuit may deviate from the actual HHL algorithm ever so slightly (although the technique is still very much HHL-inspired). Namely, the gate set used to compute the eigenvalue in Quantum Phase Estimation may not be e^(iAt).

@Strilanc Would you still consider this too compiled down? If HHL is followed precisely, my understanding is we need an arbitrary number of qubits to store eigenvalues to arbitrary precision, which may cause problems in simulators.

verult:
Implemented HHL for arbitrary 2x2 Hermitian matrices. Unfortunately setting parameters C and t in the algorithm still feels somewhat magical, although the original paper doesn't describe a precise technique for setting them, either.

This implementation is now general enough that extending to matrices with arbitrary dimensions should be fairly straightforward, but I'd love to have a review for this base algorithm first.

Implementation for `e^iAt` relies on `np.eig()` at the moment. Once https://github.com/quantumlib/Cirq/pull/1299 merges I'll migrate to the new linear operators framework.

examples/grover.py
```diff
@@ -10,7 +10,7 @@
 
 === REFERENCE ===
 Coles, Eidenbenz et al. Quantum Algorithm Implementations for Beginners
-https://arxiv.org/pdf/1804.03719.pdf
+https://arxiv.org/pdf/1804.03719
```
Strilanc: Link to the abstract instead of the pdf.

examples/hhl.py
```diff
@@ -0,0 +1,414 @@
...
+class PhaseEstimation(cirq.MultiQubitGate):
+    """
+    A gate for Quantum Phase Estimation.
+
+    U is the unitary matrix whose phases will be estimated.
+    The last qubit stores the eigenvector; all other qubits store the
+    estimated phase, in big-endian.
+    """
+
+    def __init__(self, num_qubits, U, invert=False):
```
Strilanc: make `U` a better name

examples/hhl.py
```diff
@@ -0,0 +1,414 @@
...
+class PhaseKickback(cirq.MultiQubitGate):
+    """
+    A gate for the phase kickback stage of Quantum Phase Estimation.
+
+    It consists of a series of controlled e^iAt gates with the memory qubit as
+    the target and each register qubit as the control, raised
+    to the power of 2 based on the qubit index.
+    """
+
+    def __init__(self, num_qubits, U, invert=False):
```
Strilanc: Better name for `U`

examples/hhl.py
```diff
@@ -0,0 +1,414 @@
...
+    estimated phase, in big-endian.
+    """
+
+    def __init__(self, num_qubits, U, invert=False):
+        super().__init__(num_qubits)
+        self.U = U
+        self.invert = invert
+
+    def _decompose_(self, qubits):
+        if self.invert:
```
Strilanc: Once https://github.com/quantumlib/Cirq/pull/1407 is merged, you can remove this inverse stuff.

examples/hhl.py
```diff
@@ -0,0 +1,414 @@
...
+    to the power of 2 based on the qubit index.
+    """
+
+    def __init__(self, num_qubits, U, invert=False):
+        super(PhaseKickback, self).__init__(num_qubits)
+        self.U = U
+        self.invert = invert
+
+    def _decompose_(self, qubits):
+        if self.invert:
```
Strilanc: Once https://github.com/quantumlib/Cirq/pull/1407 is merged, you can remove this inverse stuff.

examples/hhl.py
```diff
@@ -0,0 +1,414 @@
...
+    Swaps are omitted here because it's done implicitly in the PhaseKickback
+    gate by reversing the control qubit order.
+    """
+
+    def __init__(self, num_qubits, invert=False):
+        super().__init__(num_qubits)
+        self.invert = invert
+
+    def _decompose_(self, qubits):
+        if self.invert:
```
Strilanc: Once https://github.com/quantumlib/Cirq/pull/1407 is merged, you can remove this inverse stuff.

examples/hhl.py
```diff
@@ -0,0 +1,414 @@
...
+                if xor % 2 == 1:
+                    yield cirq.X(q)
+                xor >>= 1
+
+                # Build controlled ancilla rotation
+                kGate = cirq.ControlledGate(kGate)
+
+            yield kGate(*qubits)
+
+    def __ancilla_rotation(self, k):
```
Strilanc: Just one underscore.

examples/hhl.py
```diff
@@ -0,0 +1,414 @@
...
+    observables of the resulting qubit state |x>.
+    Expected observables are calculated from the expected solution |x>.
+    """
+
+    testcases = [
+        {
+            # Eigendecomposition: [(2, |+>), (1, |->)
+            # |b> = |0>
+            # |x> = (0.948683, -0.316228)
+            'A': np.array([[1.5, 0.5], [0.5, 1.5]]),
```
Strilanc: I would move these checks into the tests. For the example you should just take one matrix and show that it prints out the expected stuff.

verult:
Addressed most comments. I'll do some more debugging before having a single test case to a different file. Came up with some basic test cases that are failing.

verult:
The implementation passed all my test cases. Addressed remaining comments with the exception of the one about boilerplate inverse code. Ready for full review

verult:
Some things about Cirq I noticed while debugging HHL:

* I have a ton of gates with `_decompose_()`, and many layers of them. I wish there was a way to see a printed version of a particular HHL circuit, fully decomposed. That way I can quickly verify the generated circuit is correct. I understand "fully decomposed" could mean different things under different situations. To get around this problem, we could just have a way to specify the *decomposition depth* at which the printed circuit should represent for each gate. During debugging I could at least temporarily set the depth of decomposition. This suggestion applies to `simlate_moment_steps()` too.
* A section in the docs on `Debugging` would be nice. I envision it as a list of tools available in Cirq that could be used for debugging. A case study of the debugging process would also be extremely helpful.
* A way to *assert* certain statements about each moment (e.g. qubit A must be |0>, X observable of qubit B must be +1, etc) would've been helpful.  `simulate_moment_steps()` and Displays are already pretty good, but it takes the developer some extra time to scan through printouts and see exactly where things start to fail. Maybe an `Assertion` could be an `Operation` that can be appended in a circuit.

verult:
How should I get rid of the pylint error around branches of `NotImplemented` not being covered?

vtomole:
@verult  You need to implement cases in `examples_test.py` where `NotImplemented` is being called (calling `pow(exponent)` that is not -1 for example).

Strilanc:
https://github.com/quantumlib/Cirq/pull/1407 is now merged, so you should be able to remove the inverting logic from your gates

verult:
removed inversion logic

examples/hhl.py
```diff
@@ -0,0 +1,311 @@
...
+    """
+    A gate for Quantum Phase Estimation.
+
+    unitary is the unitary gate whose phases will be estimated.
+    The last qubit stores the eigenvector; all other qubits store the
+    estimated phase, in big-endian.
+    """
+
+    def __init__(self, num_qubits, unitary):
+        super().__init__(num_qubits)
```
vtomole: Backwards compatibility issues in Python 2 with respect to `super`.  Please change this line to `super(PhaseEstimation, self).__init__(num_qubits)` to fix.

Strilanc: It shouldn't be necessary to change this line. 3to2 inserts the arguments for you

Strilanc: Oh wait, nevermind, this is in the examples directory. Code in there must be python 2 and python 3 compatible without transformation.

examples/hhl.py
```diff
@@ -0,0 +1,311 @@
...
+        self.U = unitary
+
+    def _decompose_(self, qubits):
+        qubits = list(qubits)
+        yield cirq.H.on_each(*qubits[:-1])
+        yield PhaseKickback(self.num_qubits(), self.U)(*qubits)
+        yield (Qft(self._num_qubits-1)**-1)(*qubits[:-1])
```
Strilanc: You can apply the `**-1` at the end to avoid the parents. `X(q)**a` is equivalent to `(X**a)(q)`.

examples/hhl.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+class HamiltonianSimulation(cirq.EigenGate, cirq.SingleQubitGate):
+    """
+    A gate that represents e^iAt.
+
+    This EigenGate + np.linalg.eigh() implementation is temporary.
+    Actual implementation will use linear operators.
```
Strilanc: Why is the example talking about "temporary" and "actual implementation will"?

examples/hhl.py
```diff
@@ -0,0 +1,311 @@
...
+    """
+
+    def __init__(self, num_qubits):
+        super().__init__(num_qubits)
+
+    def _decompose_(self, qubits):
+        processed_qubits = []
+        for q_head in qubits:
+            for i, qubit in enumerate(processed_qubits):
+                yield (cirq.CZ**(1/2.0**(i+1)))(qubit, q_head)
```
Strilanc: Move `**` part to end to avoid parens.

examples/hhl.py
```diff
@@ -0,0 +1,311 @@
...
+    memory = cirq.GridQubit(register_size+1, 0)
+
+    c = cirq.Circuit()
+    hs = HamiltonianSimulation(A, t)
+    pe = PhaseEstimation(register_size+1, hs)
+    c.append([gate(memory) for gate in input_prep_gates])
+    c.append([
+        pe(*(register + [memory])),
+        EigenRotation(register_size+1, C, t)(*(register+[ancilla])),
+        (pe**-1)(*(register + [memory])),
```
Strilanc: `**-1` at end

examples/hhl.py
verult: Not sure what you mean - I can change the wording to something like: "This EigenGate + np.linalg.eigh() implementation is a hack to compute e^iAt; the linear operators framework will be used instead once that's available"

Strilanc: I guess I mean that you should do a better job of describing that a non-example implementation on a large matrix can't use this approach, and that it was chosen because it works well for demonstration purposes.

verult: I see, I was actually thinking of changing this once linear operators PR is merged, but maybe it's not high in our priorities - it's an example after all. Will update comments.

PR #1247, opened 2018-12-12
Use QubitId in stepresult DM.

### LINKED ISSUE #1232: StepResult's denisty matrix should not take in integer indices but qubits instead.
Figuring these out from the natural interfaces of the simulation requires using the qubit ordering or looking into the qubit_map.  The more natural way to do this is to use the qubit_map.

MichaelBroughton:
fixes #1232 .


cirq/sim/simulator.py
```diff
@@ -618,9 +618,11 @@ def density_matrix(self, indices: Iterable[int] = None) -> np.ndarray:
                 corresponding to the state.
         """
         return wave_function.density_matrix_from_state_vector(
-            self.state(), indices)
+            self.state(),
+            [self.qubit_map[q] for q in qubits] if qubits is not None else None
+        )
 
-    def bloch_vector(self, index: int) -> np.ndarray:
+    def bloch_vector(self, index: ops.QubitId) -> np.ndarray:
```
Strilanc: Rename `index` to `qubit`.

MichaelBroughton: changed

cirq/sim/simulator.py
```diff
@@ -605,7 +605,7 @@ def density_matrix(self, indices: Iterable[int] = None) -> np.ndarray:
                             \end{bmatrix}
```
kevinsung: You need to update the docstring to replace instances of `indices` with `qubits`.

kevinsung:
@Strilanc you mean that allowing integer indices could be confusing, right? I agree, and I think it's better to only allow specifying qubits.

cirq/sim/simulator.py
MichaelBroughton: fixed it

Issue #1041, opened 2018-10-23
Merge `cirq.Pauli.X` and `cirq.X` into one object (do same for Y and Z)


Strilanc:
Currently, there is an unnecessary distinction between `cirq.Pauli.X` (an instance of `cirq.Pauli`) and `cirq.X` (an instance of `cirq.RotXGate`). Some code wants the pauli x, some wants the x rotation. We need to unify these two things into one thing.

I'm not exactly sure how this should be done.

One way would be to introduce a `PauliXGate` that is the unification of both concepts, and falls back into one or the other as methods are applied (e.g. cirq.X would be a PauliGate but cirq.X**0.5 would again be a RotXGate).

Another way is to destroy the `Pauli` class and replace it with a type such as `Union[cirq.X, cirq.Y, cirq.Z]`, and all the class methods because just functions that can be applied to that type.

I'm labelling this "good part time project" because I think it's reasonably "meaty" design-wise and amount-of-touched-code-wise. I expect it to require iteration to hit something that works well.

dabacon:
I agree this is a problem, but worry that conflating gates with these other objects, which are really matrices / objects with algebraic relationships is dangerous. 

This ties in with our need to represent general matrices for expectation values (pauli string type idea).  I think it is OK if a gate is also one of these objects, but worry that we need to keep these interface designs separate.  

I'm not sure this a good part time project, I think this requires some careful thinking that will impact all of Cirq.

Strilanc:
"good part time project" is an approximation of "this is a strong bullet point you can add in perf-review as a 20% project". I think this meets that criteria.  Maybe a better name is needed for the tag?

--

> worry that conflating gates with these other objects, which are really matrices / objects with algebraic relationships is dangerous

I think, from the user's perspective, these things *are* fundamentally conflated. And that we should play into that, instead of trying to avoid it, if we can.

For example, have you ever tried to work out when "x^2+1" refers to a function, when it refers to a number, and when it refers to a constraint? Using only syntactic rules? It's hard! This ambiguity is not so great for computers. But in the hands of a human, the ability to play multiple roles and trigger multiple different intuitions is what makes the notation so strong.

[...]

PR #1249, opened 2018-12-13
Make cirq.Pauli a gate

viathor:
First step to #1041.


cirq/ops/common_gates.py
```diff
@@ -536,9 +536,9 @@ def _decompose_(self, qubits):
             yield cirq.XPowGate(global_shift=-0.25).on(q)
             return
 
-        yield Y(q)**0.25
-        yield X(q)**self._exponent
-        yield Y(q)**-0.25
+        yield YPowGate(exponent=0.25).on(q)
```
Strilanc: Why the switch in notation here? Avoiding a cycle?

cirq/ops/pauli.py
```diff
@@ -93,14 +75,59 @@ def __sub__(self, other_or_shift: Union['Pauli', int]
...
-    def __str__(self):
-        return self._name
 
-    def __repr__(self):
-        return 'cirq.Pauli.{!s}'.format(self)
+class _PauliX(Pauli, common_gates.XPowGate):
+    def __init__(self, *, exponent: Union[value.Symbol, float] = 1.0):
+        super(_PauliX, self).__init__(_index=0, _name='X', exponent=exponent)
+
+    def _value_equality_values_cls_(self):
```
Strilanc: This is very strongly coupled to an internal implementation detail of `value_equality`. I would instead re-annotate XPowGate with value equality, to override the options specified by eigen gate.

cirq/ops/pauli.py
```diff
@@ -11,29 +11,31 @@
...
 
 if TYPE_CHECKING:
     # pylint: disable=unused-import
     from typing import Tuple
 
 
-@value.value_equality
-class Pauli:
-    """Represents the X, Y, or Z axis of the Bloch sphere."""
+class Pauli(eigen_gate.EigenGate, metaclass=abc.ABCMeta):
```
Strilanc: I think we should significantly downgrade this class. Turns all of its methods about pairs of Paulis into private utility methods.

cirq/ops/pauli.py
```diff
@@ -11,29 +11,31 @@
...
+    """Represents the Pauli gates.
+
+    This is an abstract class with no public subclasses. The only instances
+    of private subclasses are the X, Y, or Z Pauli gates defined below.
+    """
     X = None  # type: Pauli
```
Strilanc: Delete the X/Y/Z attributes. Just `cirq.X`, no `cirq.Pauli.X`.

Not sure if the combined `XYZ` should stay or go.

cirq/ops/common_gates.py
viathor: Yes, exactly.

cirq/ops/pauli.py
viathor: Done for `difference()`. The other two already have clients, e.g. in clifford_gate.py and pauli_string.py.

(We could still do this, but it probably makes sense to do so in a separate PR.)

cirq/ops/pauli.py
Strilanc: Renaming it to pauli gates is fine.

cirq/ops/pauli.py
Strilanc: Okay, we can push it to a separate PR.

cirq/ops/pauli.py
```diff
@@ -11,73 +11,56 @@
...
+    """Represents the Pauli gates.
+
+    This is an abstract class with no public subclasses. The only instances
+    of private subclasses are the X, Y, or Z Pauli gates defined below.
+    """
+    _XYZ = None  # type: Tuple[Pauli, Pauli, Pauli]
 
-    def __init__(self, *, _index: int, _name: str) -> None:
+    @staticmethod
+    def by_index(index: int) -> 'Pauli':
```
Strilanc: Do these gates have specialized simulation methods? It's important that X and Z go fast. If not, open an issue to resolve.

viathor: Yes. Every object that we return here is either cirq.X, cirq.Y or cirq.Z and therefore `isinstance` of XPowGate, YPowGate or ZPowGate. All `isinstance`-based specializations for these types trigger for all Pauli gates appropriately.

I've added a unit test to verify `isinstance` behavior.

cirq/ops/pauli.py
viathor: Done.

PR #1251, opened 2018-12-14
[WIP] Adds a custom lint checker and a single case for unescaped docstrings

philiptmassey:
Sending this for some early comments. The overall structure isn't very difficult, but I want to make sure that my leaning on the dev_tools environment and check infrastructure is correct.


Strilanc:
Ah... I realize I asked you to do this, but it just occurred to me that pylint might already have this functionality ( e.g. https://github.com/PyCQA/pylint/issues/264 seems to be talking about it). Could you check if it's already present?

That being said, latex often contains stuff like `\\` so we might care about more cases than pylint does.

philiptmassey:
Would it just be enabling anomalous-backslash-in-string in the .pylintrc?

https://docs.pylint.org/en/1.6.0/features.html warning W1401

philiptmassey:
Just tested adding anomalous-backslash-in-string to the .pylintrc, and it doesn't seem to catch be adding "\text{something}" and "\\text{something}" in a random docstring.

Strilanc:
You should use r-strings to avoid nesting escape characters. Instead of `'\\\.+?\{.+?\}'` use `r'\.+?{.+?}'. (Why were you escaping the curly braces? This isn't a format string.)

Strilanc:
I think this was obsoleted by turning on a rule in pylint

PR #1252, opened 2018-12-14
Add method for computing Display values to SimulatesIntermediateWaveFunction

kevinsung:


kevinsung:
Any insight on this Python 2 conversion error?
```
/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/_pytest/python.py:403: in _importtestmodule
    mod = self.fspath.pyimport(ensuresyspath=importmode)
/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/py/_path/local.py:668: in pyimport
    __import__(modname)
/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/_pytest/assertion/rewrite.py:212: in load_module
    py.builtin.exec_(co, mod.__dict__)
examples/examples_test.py:2: in <module>
    import cirq
cirq/__init__.py:23: in <module>
    from cirq.circuits import (
cirq/circuits/__init__.py:22: in <module>
    from cirq.circuits.qasm_output import (
cirq/circuits/qasm_output.py:29: in <module>
cirq/ops/__init__.py:20: in <module>
    from cirq.ops.clifford_gate import (
cirq/ops/clifford_gate.py:22: in <module>
    from cirq import protocols, value
cirq/protocols/__init__.py:66: in <module>
    from cirq.protocols.resolve_parameters import (
cirq/protocols/resolve_parameters.py:19: in <module>
    from cirq.study import ParamResolver
cirq/study/__init__.py:19: in <module>
    from cirq.study.compute_displays_result import (
E     File "/tmp/cirq.pytest2.Qy9hlMtBlr0STkGA/cirq/study/compute_displays_result.py", line 23
E       from __future__ import absolute_import
E   SyntaxError: from __future__ imports must occur at the beginning of the file
```

cirq/sim/simulator.py
```diff
@@ -447,9 +448,134 @@ def _simulator_iterator(
...
+            qubit_map = {q: i for i, q in enumerate(qubits)}
+            _enter_moment_display_values_into_dictionary(
+                display_values, moment, state, qubit_order, qubit_map)
+
+            # Compute the displays in the rest of the Moments
+            all_step_results = self.simulate_moment_steps(circuit,
+                                                          param_resolver,
+                                                          qubit_order,
+                                                          initial_state)
+            for step_result, moment in zip(all_step_results, circuit[1:]):
```
Strilanc: Why are we the step results off-by-one with respect to the circuit moments? I'd have expected the first result to be the state after the first moment.

cirq/sim/simulator.py
```diff
@@ -447,9 +448,134 @@ def _simulator_iterator(
...
+
+
+def _enter_moment_display_values_into_dictionary(
+        display_values: Dict,
+        moment: ops.Moment,
+        state: np.ndarray,
+        qubit_order: ops.QubitOrder,
+        qubit_map: Dict[ops.QubitId, int]):
+    displays = (op for op in moment
+                if isinstance(op, (ops.SamplesDisplay,
```
Strilanc: This filtering is effectively already being done inside the for loop. Just iterate over the moment's operations while checking for wave function display else samples display.

cirq/sim/simulator.py
```diff
@@ -447,9 +448,134 @@ def _simulator_iterator(
...
+                                   ops.WaveFunctionDisplay)))
+    for display in displays:
+        if isinstance(display, ops.WaveFunctionDisplay):
+            display_values[display.key] = (
```
Strilanc: Does this correctly account for the fact that, if there are measurements before this display, it is necessary to sample?

We might want a 'density matrix display' concept for the more general case, at some point.

cirq/sim/simulator.py
kevinsung: Because the moments in a display are computed with respect to the state of the circuit _before_ that moment.

cirq/sim/simulator.py
kevinsung: Done.

cirq/sim/simulator.py
kevinsung: I don't understand. Measurements before this display can only be in a previous moment. Why would it be necessary to sample?

kevinsung: Ah, I see, you mean that the value of some displays (like idealized pauli string expectations) should not depend on the outcome of previous measurements, and rather take the value it would have from a density matrix simulations. Hmm ...

kevinsung: What do you think of adding a `repetitions` argument to `compute_displays`, indicating to average the value of displays over that many circuit repetitions?

Strilanc: I think that should be a per-display option, not a global option.

cirq/study/compute_displays_result_test.py
```diff
@@ -0,0 +1,26 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import cirq
+
+# Python 2 gives a different repr due to unicode strings being prefixed with u.
+@cirq.testing.only_test_in_python3
+def test_compute_displays_result_repr():
+    v = cirq.ComputeDisplaysResult(
```
Strilanc: Use the consistent repr test method in a not-gated-to-python-3 test.

Issue #1161, opened 2018-11-15
Make decompositions respect global phase


Strilanc:
I've seen enough people confused by the fact that decomposing a gate can change its global phase to be convinced that we should try to preserve it.

I think the way we should do this is by adding a "global phase gate" that moments or circuits automatically fold and diagrams treat specially in some way. Then we need to go through all the decompositions and add this global phase part. We would also need to update any tests that are now incorrectly only up to global phase.

viathor:
I changed the assert for unitary-vs-decomposition consistency to respect global phase in #1160. Hadamard was the only one broken by this change, so the last step above should be simple.

kevinsung:
Well, the tests in #1165 are failing because #1160 changed `assert_decompose_is_consistent_with_unitary` to require global phase consistency

Strilanc:
Oh. That's my mistake for letting that through before we were ready to really guarantee it. Revert it in your PR.

viathor:
Oops. Sorry for that. I thought that since all other gates passed the new stricter test we were ready to provide the stronger guarantee. Didn't know about #1165 .

An alternative to reverting would be a temporary switch for use in tests of gates that fail the stricter version of the test.

Strilanc:
Yes, I agree that a boolean switch defaulting to "phase insensitive" would be ideal.

kevinsung:
After #1165 , fixing this issue amounts to making the `include_global_phase` argument of `assert_decompose_is_consistent_with_unitary` default to True and then fixing the broken tests.

c-poole:
This looks it has been completed.

PR #1254, opened 2018-12-17
Make include_global_phase argument of assert_decompose_is_consistent_with_unitary default to True

smitsanghavi:
And fix broken tests. #1161


cirq/testing/consistent_protocols.py
```diff
@@ -52,6 +55,7 @@ def assert_eigengate_implements_consistent_protocols(
             0, 1, -1, 0.5, 0.25, -0.5, 0.1, value.Symbol('s')),
         global_shifts: Sequence[float] = (0, 0.5, -0.5, 0.1),
         qubit_count: Optional[int] = None,
+        is_global_phase_sensitive: bool=False,
```
kevinsung: I think we agreed in #1161 that global phase should always be preserved, so you should not be adding this flag.

Strilanc: Long term we want global phase to be preserved, but short term it makes sense to have this.

Rename to `ignoring_global_phase` defaulting to false. We want to be able to find tests opting into global phase insensitivity and update them as we fix individual issues.

cirq/testing/consistent_decomposition.py
```diff
@@ -22,7 +22,7 @@
 
 def assert_decompose_is_consistent_with_unitary(
         val: Any,
-        including_global_phase: bool=False):
+        including_global_phase: bool=True):
```
Strilanc: While we're switching the default value, let's rename this to `ignoring_global_phase` (inverting its meaning).

cirq/ops/three_qubit_gates_test.py
```diff
@@ -26,7 +26,7 @@
 )
 def test_eigen_gates_consistent_protocols(eigen_gate_type):
     cirq.testing.assert_eigengate_implements_consistent_protocols(
-            eigen_gate_type)
+            eigen_gate_type, is_global_phase_sensitive = True)
```
Strilanc: Isn't this the default now? Don't specify it.

cirq/ops/phased_x_gate_test.py
```diff
@@ -28,7 +28,14 @@ def test_phased_x_consistent_protocols(phase_exponent):
     cirq.testing.assert_implements_consistent_protocols(
             cirq.PhasedXPowGate(phase_exponent=phase_exponent,
                                 exponent=1.0,
-                                global_shift=0.1)
+                                global_shift=0),
+            is_global_phase_sensitive = False
+    )
+    cirq.testing.assert_implements_consistent_protocols(
```
Strilanc: Why the duplicated test? The global phase sensitive version passing implies the global phase insensitive version will pass. Undo changes to this file.

cirq/ops/common_gates_test.py
```diff
@@ -26,19 +26,20 @@
...
-@pytest.mark.parametrize('eigen_gate_type', [
-    cirq.CNotPowGate,
-    cirq.CZPowGate,
-    cirq.HPowGate,
-    cirq.ISwapPowGate,
-    cirq.SwapPowGate,
-    cirq.XPowGate,
-    cirq.YPowGate,
-    cirq.ZPowGate,
+@pytest.mark.parametrize('eigen_gate_and_phase_sensitivity', [
```
Strilanc: Split this into two different tests, one for the false case and one for true.

Strilanc: (You use commas to separate arguments in the spec string.)

cirq/testing/consistent_protocols.py
smitsanghavi: Renamed.

cirq/testing/consistent_decomposition.py
smitsanghavi: Renamed, hence inverting meaning. So didn't change the default value.

cirq/ops/three_qubit_gates_test.py
smitsanghavi: The default is False.

cirq/ops/phased_x_gate_test.py
smitsanghavi: It's not duplicated. The phase insensitive version passes only if global_shift==0. 
Added the insensitive test to verify that global phase doesn't change if the arg is 0.

cirq/ops/phased_x_gate_test.py
```diff
@@ -25,10 +25,18 @@
     ]
 )
 def test_phased_x_consistent_protocols(phase_exponent):
+    # If there is no global_shift, the gate is global phase insensitive.
     cirq.testing.assert_implements_consistent_protocols(
             cirq.PhasedXPowGate(phase_exponent=phase_exponent,
                                 exponent=1.0,
-                                global_shift=0.1)
+                                global_shift=0),
+            ignoring_global_phase = False
```
Strilanc: Don't bother to say you're using the default value.

cirq/ops/phased_x_gate_test.py
```diff
@@ -25,10 +25,18 @@
...
+    )
+    cirq.testing.assert_implements_consistent_protocols(
+            cirq.PhasedXPowGate(phase_exponent=phase_exponent,
+                                exponent=1.0,
+                                global_shift=0.1),
+            ignoring_global_phase = True
```
Strilanc: Nit: no spaces around the `=`.

PR #1255, opened 2018-12-17
Fix parameter retrieval in proto deserialization

viathor:


viathor:
@dabacon Suppose you have a circuit with a parametrized gate, e.g. `Z**t`. Then protobuf encoding it will look like

```
exp_z: {
    half_turns: {
        parameter_key: 't'
    }
}
```

but since in proto3 missing scalar fields are indistinguishable from fields set to their default values the corresponding proto dict will look like this:

```python
{'exp_z': {'half_turns': {'parameter_key': 't', 'raw': 0.0}}...}
```

and so deserialization (before this PR) will return `Z**0.0` instead of `Z**t`. This change ensures that if `parameter_key` is set then `raw` isn't used.

(Eventually, we probably want to use a single serialization format, not two. This is a patch until we do.)

Issue #1224, opened 2018-12-05
Mention contrib requirements install during dev setup documentation


Strilanc:

vtomole:
Please close.

PR #1256, opened 2018-12-17
Add contrib requirements to dev install

dabacon:
If you don't do this then `pytest .` fails due to tensorflow not being installed.


viathor:
Fixes #1224.

PR #1257, opened 2018-12-17
Add compute_samples_displays method to SimulatesSamples

kevinsung:
Comes after #1252 

Since SimulatesSamples and SimulatesIntermediateWaveFunction both have `compute_displays` methods with different signatures, mypy complains about classes that inherit from both:
```
cirq/sim/sparse_simulator.py:27: error: Definition of "compute_displays" in base class "SimulatesIntermediateWaveFunction" is incompatible with definition in base class "SimulatesSamples"
cirq/google/sim/xmon_simulator.py:96: error: Definition of "compute_displays" in base class "SimulatesIntermediateWaveFunction" is incompatible with definition in base class "SimulatesSamples"
```
Is this actually a problem?

Also, I don't understand why pylint is making the following complaint:
```
cirq/sim/sparse_simulator_test.py:439:19: E1003: Bad first argument 'SimulatesIntermediateWaveFunction' given to super() (bad-super-call)
cirq/sim/sparse_simulator_test.py:443:19: E1003: Bad first argument 'SimulatesIntermediateWaveFunction' given to super() (bad-super-call)
```


cirq/sim/simulator.py
```diff
@@ -111,6 +112,75 @@ def _run(
...
         raise NotImplementedError()
 
+    def compute_displays(
+            self,
+            program: Union[circuits.Circuit, schedules.Schedule],
+            param_resolver: Optional[study.ParamResolver] = None,
+    ) -> study.ComputeDisplaysResult:
+        """Computes displays in the supplied Circuit.
+
+        Only SamplesDisplays are computed; any WaveFunctionDisplays are ignored.
```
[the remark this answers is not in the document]
kevinsung: Or we could edit the docstring to say "Computes samples displays" and maybe change the name of the method to `compute_samples_displays`.

cirq/sim/simulator.py
```diff
@@ -114,6 +114,72 @@ def _run(
...
+                            if isinstance(op, ops.SamplesDisplay))
+                preceding_circuit = circuit[:i]
+                for display in displays:
+                    measurement_key = str(display.key)
+                    measurement_circuit = circuits.Circuit.from_ops(
+                        display.measurement_basis_change(),
+                        ops.measure(*display.qubits,
+                                    key=measurement_key)
+                    )
+                    measurements = self._run(
```
dabacon: This looks like it is scaling like O(moments^2)?

kevinsung: This scaling seems unavoidable to me with the current interface because of the need to generate new circuits to submit to `_run`. Do you see a good way around it?

kevinsung:
By the way, I intend to add basically the same method to Engine.

cirq/sim/simulator.py
Strilanc: I agree that it's unavoidable assuming you only have access to a sampling method.

cirq/sim/simulator.py
```diff
@@ -114,6 +114,73 @@ def _run(
...
+            List of ComputeDisplaysResults for this run, one for each
+            possible parameter resolver.
+        """
+        circuit = (program if isinstance(program, circuits.Circuit)
+                   else program.to_circuit())
+        param_resolvers = study.to_resolvers(params or study.ParamResolver({}))
+
+        compute_displays_results = []  # type: List[study.ComputeDisplaysResult]
+        for param_resolver in param_resolvers:
+            display_values = {}  # type: ignore
```
Strilanc: Why ignore instead of `type: Dict[Value, Key]`?

cirq/sim/sparse_simulator_test.py
```diff
@@ -432,6 +433,41 @@ def test_compute_displays(dtype):
                                atol=1e-7)
 
 
+@pytest.mark.parametrize('dtype', [np.complex64, np.complex128])
+def test_compute_samples_displays(dtype):
+    qubits = cirq.LineQubit.range(4)
```
Strilanc: You don't use qubit 0?

Suggestion: I like the style `a, b, c, d = cirq.LineQubit.range(4)`.

cirq/sim/simulator.py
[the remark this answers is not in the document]
Strilanc: I think that's fine for now. We can change it later if we need to.

cirq/sim/simulator.py
kevinsung: The correct annotation here would be `Dict[Hashable, Any]` which doesn't provide additional information. It is better than `ignore` though.

PR #1259, opened 2018-12-18
Install contrib deps in dev virtualenvs

viathor:
Without these, pytest fails in a fresh virtualenv complaining about missing tensorflow:

```
Traceback:
cirq/contrib/tpu/__init__.py:75: in <module>
    from cirq.contrib.tpu.circuit_to_tensorflow import (
cirq/contrib/tpu/circuit_to_tensorflow.py:20: in <module>
    import tensorflow as tf
E   ImportError: No module named 'tensorflow'
```


dabacon:
https://github.com/quantumlib/Cirq/pull/1256

Issue #1246, opened 2018-12-12
Change all Circuit insertion methods to preserve Moments


kevinsung:
#1213 did this for `insert` and `append`, but not for `insert_into_range`, `insert_at_frontier`, `batch_insert`, or `batch_insert_into` (and any other ones I might have missed). Note that Moments now satisfy the OP_TREE contract.

To clarify what this means: OP_TREEs can contain sub-trees which are Moments. These Moments should be inserted into the Circuit intact instead of being flattened and mixed with other operations.

iamvamsi:
@kevinsung : I am starting with the `insert_into_range` method. Let me state my understanding of the `insert_into_range`: It adds operations to the existing moments (between indices passed as args) until the method can do. We use `can_add_operation_into_moment` to check if it could be done for operations. I see that a similar thing is not available for the moments. As a general question, the way to deal with moments would also remain the same as of that operations expect that the moment (set of operations) should either fit together or not ?
LMKWYT

kevinsung:
@iamvamsi Moments should be treated differently from operations. While you can attempt to insert operations into an existing moment, you should insert a moment all by itself. You should let `flatten_op_tree` with the `preserve_moments` flag set to True handle the logic of inserting moments intact. See the `insert` method for an example of this.

kevinsung:
From a brief discussion with @Strilanc : the proliferation of Circuit insertion methods indicates that instead of resolving this issue the way I stated it, we should instead somehow combine the insertion methods into a more general one. In any case, we should probably discuss here what the exact behavior of the insertion methods should be on encountering Moments before attempting any changes.

bryano:
I'm not sure that `Circuit.insert` currently does the right thing when the first one or more elements of  `moments_and_operations` is a `Moment`. Specifically, I suggest that the [potential change from `NEW_THEN_INLINE` to `INLINE`](https://github.com/quantumlib/Cirq/blob/a2b981a3bbc50e4651975c4016f6998777a8f067/cirq/circuits/circuit.py#L839-L840) be de-indented so that it's triggered after the first `moment_or_op` regardless.

More generally, reasonable behavior would be that each `Moment` in `moments_and_operations` essentially resets the strategy to `NEW_THEN_INLINE`.
[...]

PR #1260, opened 2018-12-19
insert_into_range updated to preserve moments insertion #1246

iamvamsi:


iamvamsi:
@Strilanc @kevinsung : Updated the `insert_into_range` as per requirements. I want to get an eye before going further.

iamvamsi:
#1246

kevinsung:
I'm sorry @iamvamsi but I must admit I didn't quite think through what the logic should be for this method. It's possible that "preserving moments" doesn't quite make sense for this method. Can you explain what the behavior of the method is, with your changes?

iamvamsi:
@kevinsung : `insert_into_range` method, with my changes, when encountered a moment would insert it as new moment in the circuit next to the index it is currently in and skips the inserted moment for any future insertions. 
Eg: Lets say we have `Circuit([Moment([cirq.X(x)])] * 4)` and we plan to do `c.insert_into_range([cirq.Y(y), Moment([cirq.Z(y), cirq.Z(x)]), cirq.Y(y)], 2, 4)` the resulting circuit would be 
```
x: ───X───X───X───Z───X───

y: ───────────Y───Z───Y───

```
but lets say we plan to insert `c.insert_into_range([cirq.Y(y), Moment([ cirq.Z(x)]), cirq.Y(y)], 2, 4)` the resulting circuit would be 
```
x: ───X───X───X───Z───X───

y: ───────────Y──────Y───

```
I must say this may not be a very intuitive behaviour and as suggested by you we should discuss the same.

kevinsung:
@iamvamsi if you don't mind I'll close this for now; I think further progress needs to be made on #1246 before proceding with something like this.

iamvamsi:
Thanks.

Issue #1270, opened 2018-12-25
check.sh uses hardcoded python3 location for temporary test enironment


Ashalynd:
`Cirq/dev_tools/env_tools.py::prepare_temporary_test_environment` uses hardcoded location `/usr/bin/python3.5` which results in error if python3 is installed somewhere else. (Happened on Mac where I had python3.7 as the default python version, and no python3.5 present).

PR #1272, opened 2018-12-25
Nonhardcode python location in dev_tools/env_tools.py

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/1270

@Ashalynd  Please review.


Ashalynd:
Hm, we should also check that it doesn't refer to Python 2 interpreter executable.

vtomole:
@Ashalynd `sys.executable` uses the `PATH` variable. It refers to the interpreter that runs it; so it won't refer to a Python2 interpreter unless you are using a Python2 interpreter to run `env_tools.py`

PR #1273, opened 2018-12-25
Add dockerhub pull

vtomole:
I created an organization that hosts the Docker images: https://hub.docker.com/u/quantumlib. Please give me your Docker ID so I may add you.


vtomole:
@zoltanegyed Are you saying that we should use the same Dockerfile for development  and non-development? Maybe have an if statement in the Dockerfile that sepcifies whether to `pip3 install cirq` or `git clone cirq` depending on if we want the image to be for development or not? If yes, i agree.

zoltanegyed:
That is a good idea. We can also do that in a separate PR. I was just thinking that the instructions we give and the prebuilt images should be named the same (one name for dev, one for prod, but the same for prebuilt and self-built).

docs/development.md
```diff
@@ -13,12 +13,10 @@ git clone git@github.com:quantumlib/cirq.git
 cd cirq
 ```
 
-To do your development in a Docker virtual machine, you can use dev_tools/Dockerfile:
+To do your development in a Docker image, you can build one with Cirq/dev_tools/Dockerfile or pull an existing image:
```
dabacon: Cirq/dev_tools/Dockerfile doesn't exist?

vtomole: Here: https://github.com/quantumlib/Cirq/blob/master/dev_tools/Dockerfile

vtomole: @dabacon Ping :)

dabacon: Ah thanks.  Could you make the doc link to that file?

vtomole: Where should i put the link? On this file?  `docs/development.md`

Strilanc:
Jeez, DockerHub is very bad at displaying our README.

vtomole:
This is a sign to remove all the fancy stuff and make it a `README.txt`/s :laughing:

vtomole:
@Strilanc Docker Hub doesn't render `.rst` which is what our README is. It renders `.md`.

PR #1274, opened 2018-12-26
Fixes bug in Circuit.insert

bryano:


kevinsung:
Can you add a test that catches this bug?

cirq/circuits/circuit_test.py
```diff
@@ -27,6 +27,40 @@
 import cirq.google as cg
 
 
+def raise_value_error_if_wrong_type(obj, classinfo):
```
kevinsung: I think defining this function is needless generalization and actually makes the code less readable.

cirq/circuits/circuit_test.py
```diff
@@ -27,6 +27,40 @@
...
+    if not isinstance(obj, classinfo):
+        raise ValueError('not isinstance({!r}, {!r})'.format(obj, classinfo))
+
+
+class _MomentAndOpTypeValidatingDevice(
+        cirq.devices.unconstrained_device._UnconstrainedDeviceType):
```
kevinsung: Instead of inheriting from an unrelated and hidden class, just inherit directly from Device.

cirq/circuits/circuit_test.py
```diff
@@ -27,6 +27,40 @@
...
+    circuit = cirq.Circuit(device=moment_and_op_type_validating_device)
+
+    moment_or_operation_tree = [cirq.X(x), cirq.Moment([cirq.Y(x)])]
+    circuit.insert(0, moment_or_operation_tree)
+
+    moment_or_operation_tree = [[cirq.Moment([cirq.X(x)])]]
+    circuit.insert(0, moment_or_operation_tree)
+
+
+
```
kevinsung: Nit: extra newline (there should be 2).

cirq/circuits/circuit_test.py
```diff
@@ -27,6 +27,40 @@
 import cirq.google as cg
 
 
+def raise_value_error_if_wrong_type(obj, classinfo):
+    if not isinstance(obj, classinfo):
+        raise ValueError('not isinstance({!r}, {!r})'.format(obj, classinfo))
+
+
+class _MomentAndOpTypeValidatingDevice(
```
kevinsung: Rename the class _MomentAndOpTypeValidatingDeviceType

cirq/circuits/circuit_test.py
```diff
@@ -27,6 +27,47 @@
...
+
+    def duration_of(self, operation):
+        return cirq.Duration(picos=0) # coverage: ignore
+
+    def validate_schedule(self, schedule):
+        pass
+
+    def validate_scheduled_operation(self, schedule, scheduled_operation):
+        pass
+
```
kevinsung: Nit: one more newline.

PR #1277, opened 2018-12-28
Remove bloch_vector

### LINKED ISSUE #1231: Move bloch vector from StepResult to abstract class
StepResult is a very generic interface which is used for simulating any sized system, but bloch vector is specialized to a single qubit.  If one really needs it one can just do cirq.bloch_vector_from_state_vector(step_result.state()).

andbe91:
Fixes #1231


dabacon:
I've updated #1231 with the results of discussion about remove bloch vector.  We think a better solution is to have it in another abstract class that can be easily inherited from.  Sorry for the switch of direction.

andbe91:
No worries, I will work on the abstract class early next week!

Strilanc:
I will close this since we decided not to remove it. You can re-open this PR or a new PR when you have the updated changes. Sorry for asking for one thing then another.

PR #1278, opened 2018-12-28
Add examples to the documentation

### LINKED ISSUE #1205: Include files from examples/ in the documentation
We wrote all these examples, but if you go to readthedocs you have no idea they even exist.

As a starting point it is acceptable to do nothing but iterate over the source files (except test) and produce a page containing nothing but the source file as a code block.

andbe91:
Fixes #1205 

I couldn't find a good way to iterate over all files in the examples directory using only rst, so I manually added the examples. I can of course write a small bash or python script to automatically generate the required rst, but I didn't know if this is desired or not.

Let me know what you think.


vtomole:
The way quantum circuits in the doc strings are rendered does not match the ones in the source files for me.

vtomole:
I have `examples.html` open in Firefox if that makes a difference.

andbe91:
Could you please provide a screenshot of how it looks for you? For me, the circuits looks correct using Firefox on Debian. However, I will try other browsers and operating systems and see if I can reproduce it

vtomole:
This is what i get: https://imgur.com/a/D16T6IE

dabacon:
Let's get this one in and then we can fix those circuits if we need to.

PR #1248, opened 2018-12-13
Remove OptimizationPass

vtomole:
Fixes https://github.com/quantumlib/Cirq/issues/1241

I've modified two tests in cirq/optimizers/eject_z_test.py. Please look at them first and let me know if they break anything.

PR #1280, opened 2019-01-02
Remove documentation reference to the deleted class OptimizationPass (see #1248)

andbe91:
OptimizationPass was not removed from the documentation, causing warnings when building the docs.


dabacon:
Note that this will not update the documentation on readthedocs, as that requires a push of the documentation.  I believe we are keeping the read the docs in sync with the last release.

Issue #1190, opened 2018-11-22
Remove `cirq.Tolerance`


Strilanc:
This class ended up being nothing more than a variation on passing `atol` arguments, so it's not pulling enough weight to justify being in the API.

andbe91:
Hi @Strilanc 
This looks like a simple enough issue to start my contributions with. Just to make sure I understand, you only want it removed from the API, i.e. basically removing the import from the __init.py__ in the root directory?

vtomole:
@andbe91 I've started working on this but i didn't officially claim it so it's up to @Strilanc to decide; but yes, remove every reference to `Tolerance` (removing tolerance.py. e.t.c).

andbe91:
@vtomole since you have already started, it might be better if I work on another issue. Please let me know if you want me to take over for some reason.

Strilanc:
@vtomole is correct; this would involve completely deleting the class

Strilanc:
That being said, I think the main important arguments to propagate are `atol` and `rtol`, no need for the nan ones.

vtomole:
This is done. Please close.

dabacon:
We might want to take a pass through code to insure that we are using atol and rtol as parameters when it makes sense (probably most of the time)

PR #1281, opened 2019-01-04
Remove `cirq.Tolerance` in favor of passing `atol` and `rtol` in method parameters.

bbonnet:
Removes the `Tolerance` class in favor for passing `atol` (Absolute Tolerance) and `rtol` (Relative Tolerance) as float method parameters .

Addresses: https://github.com/quantumlib/Cirq/issues/1190

- [x] Remove `Tolerance` from the public API
- [x] Exposes Default `atol` and `rtol` constants internally to avoid setting defaults across the project
- [ ] WIP: Renames `tolerance` method paramters to `atol` and/or `rtol` to explicitly indicate what tolerance type is being set on the API

Hi! It's my first PR here and wanted to tackle this one to get familiar with the codebase.  Hope that this helps!  Looking for earlier feedback before wrapping up some loose ends.


bbonnet:
Rerun CLA

bbonnet:
Rerun CLA

vtomole:
@bbonnet Thanks for taking this. I recommend running the checks locally as using the travis logs for debugging could be inefficient.

bbonnet:
@vtomole I took a look at your PR however IMO I think it would be best to keep tolerance.py along with the utility methods for the following reasons:

- Prevent copying of logic for the all near zero and all near zero mod methods which are non-obvious.  This also makes it for future functionality to add tolerance by calling these utility methods.
- Keep the unit test coverage on these methods esp. because many of the methods that use the tolerance do not have test coverage across all tolerance settings.  Adding that test coverage at every implementation would probably be over-kill.

The changes in this PR should cover the main spirit of #1190 which is to remove the Tolerance class from the public API however I think the internals still do benefit from these helpers.  Happy to discuss further if you disagree.

vtomole:
@bbonnet  Sounds good.

bbonnet:
@Strilanc @dabacon Thanks for the review! I've updated based on your comments and this should be ready for another look.

Issue #1291, opened 2019-01-10
Conundrum: Protocols for variable qubit number gates


dabacon:
MeasurementGate is an example of a gate that can act on a variable number of qubits and turn into a GateOperation on these qubits.  We can imagine also a gate that acts on multiple qubits (a control^k-X gate for example) but that the number of qubits it acts on is not determined until "on" is applied to qubits.  This causes a problem.  Suppose you want to evaluate a protocol on an operation, for example cirq.unitary or cirq.channel.  Then if you call this on the operator, it can only delegate to the gate, and the gate doesn't know how many qubits it acts on.  

There are a couple of ways out of this.  
1. add an optional qubits parameter to these protocols.  Gates that need to know the number of qubits the gate is applied to.  This has the drawback of muddling this protocol (when do I care about the qubits and when do I not?)  
2. make the gates that need a variable number of qubits know the number of qubits they act on.  This would mean MeasurementGate(2).on(q1, q2).  Of course a convenience cirq.measure(q1, q2) would not change and similarly methods could be done for all of these variable number of qubit gates.  

I was leaning towards 1, but the more I think about it, I'm not quite convinced that 2 isn't slightly better.

Thoughts?

dabacon:
@Strilanc @kevinsung @viathor

dabacon:
Oh another option:
3. Have the protocols return the object that they known how to construct because they know how many qubits they act on, or a function which takes the qubits to act on (or number of qubits).

kevinsung:
How about:
4. The protocols are undefined for Gates. They must be applied to GateOperations.

dabacon:
5. Different protocols for cases with number of qubits needed and those without.

dabacon:
5 is what we do already for decompose

viathor:
Our abstractions are currently layered like this:
* Gates which don't know the qubits they act on,
* GateOperations which do.

The conundrum above stems from the fact that we have not explicitly specified whether Gates know the answer to a related question: how many qubits do I act on?

We should fix this.

Consider two abstractions: XPowGate which knows the number of qubits it acts on and MeasurementGate which doesn't. We currently pretend these are both Gates, i.e. we place them at the same level of our abstraction hierarchy. In reality, MeasurementGate is more abstract and XPowGate is more concrete. In fact, MeasurementGate is more like a "gate factory": only once you feed it a number of qubits to act on will it give you back something that's equivalent (in terms of concreteness) to other gates like XPowGate.

[...]

PR #1298, opened 2019-01-17
Proposal: make measurement gate have number of qubits

dabacon:
As part of #1291 one suggestion is to make MeasurementGate be a measurement gate for a number of qubits.  This is what such a change would look like.


dabacon:
See #1332.

I'm not sure where you see redundancy here, the only place the number of qubits shows up is in tests.

Issue #1037, opened 2018-10-22
Powering vector of pauli components


dabacon:
Add method to calculate (aI+bX+cY+dZ)^k for k integer as this can then be used to power a depolarizing channel.

Strilanc:
AKA implement nearly-quaternion arithmetic. We don't have a way to even make linear combinations of Paulis yet, right? I think there's a series of changes leading up to this change that need to be described.

viathor:
I'll review where we're at and will post a plan.

dabacon:
OK so this is actually connected to Fourier transforms over non-abelian groups!  Let me try to explain.  TLDR: take the QFT over the Pauli group of the a,b,c,d coeeficients, raise these to a power, and then inverse QFT to recover new Pauli vector!

Let G be a finite group.  Functions on this group to some field (lets just say complex numbers for now)  f: G -> C form a nice ring.  A convenient way to think about this is to represent \sum_x a_x x where a_x are elements of C and x is the group element.  Then you can see that you can multiply such objects together and represent it as another such object (\sum_x a_x x \sum_y b_y y  = \sum_{xy} a_x b_y xy and xy is just another member of the group), and you can also add and subtract linear combinations of them (\sum a_x x + \sum b_x x = \sum (a_x + b_x) x etc).  

Now recall that the Fourier transform is a map from functions over G to irreducible representations of G. Lets say the irreps of G are labeled by \rho and that the matrix representation of this irrep at x \in G is the matrix rho_{i,j}(x).  Then  the Fourier transform over G is a mapping that takes functions f: G -> C to functions \tilde{f} : \rho_{i,j} -> C.  Explicitly the FT of f over G at \rho_{i,j} is
\sum_x f(x) \rho_{i, j} (x)
The FT over G of a function f, just like normal FTs retains all of the information about f.  That is given all the \sum_x f(x) \rho_{i,j}(x) over all \rho,i,j values we can perform the inverse FT to recover F.  The inverse FT is given by
f(x) = {1 \over |G|} \sum_\rho d_\rho Tr( \rho(x^{-1}) \tilde{f}(\rho))
where d_\rho is the degree of the irrep \rho and the stuff inside the trace is the matrix product of the irreps.

Now suppose we want to calculate 
[\sum_x a_x x]^2
We can write this as 
[\sum_x a_x x ][\sum_y a_y y]
= \sum_{x,y} a_x a_y (xy)
[...]

Issue #962, opened 2018-09-23
Add __mul__ to cirq.PauliString and cirq.Pauli


Strilanc:
Note that the result has to be a `(complex, cirq.Pauli/cirq.PauliString])` tuple, because there may be an imaginary phase.

For example, `cirq.Pauli.X * cirq.Pauli.Y` should return `(1j, cirq.Pauli.Z)`.

bradyb:
What should the `cirq.Pauli` result be when you're squaring?

vtomole:
@bradyb The identity. It's not implemented in `cirq.Pauli` so i'm curious as to how we'll deal with that.

Strilanc:
I would make the result type `Tuple[complex, Optional[Pauli]]` and use `(1, None)` to represent the identity.

Strilanc:
The multiplication rules are quite simple; there's no need to go to the matrix arithmetic. Let r(I) be 00, r(X) be 01, r(Z) be 10 and r(Y) be 11. Then r(A*B) = r(A) xor R(b). Then you have to account for the sign of the product, which is based on the cycling direction: `s = 1j if A < B else -1j if A > B else 1`

Issue #932, opened 2018-09-18
Can and should we use python magic for numpy matrices


dabacon:
I think we can use rmul and lmul to create a class that wraps numpy matrices nicely such that you can write things like
H = 0.5 * X * X + 0.2 * Y * I
or if you want to use labeled qubits
H = 0.5 * X(q1) * X(q2) + 0.2 Y[1].

This would be useful as a base for at least a few cases I can think of
1) gate obtained by exponentiating these operators
2) returning expecatation values of these operators, in simulation, and also at hardware.

You could even use these as substitutes in other places where we are currently using numpy arrays.

Strilanc:
Is `*` the pointwise product, the matrix product, the tensor product, or the kronecker product?

dabacon:
* was tensor.  Pointwise (Hamadard) product is crazy ;) 

@ only appeared in python 3.5, so I'm not sure if we can use that for matrix multiplication?

dabacon:
Another thing is that there are implicit kroneckers in the first, and int he second the qubit order is necessary to define the kronecker order.

Strilanc:
I think it's fine for the tensor order to be implicit in the application order. On the other hand, X * X really *really* looks like it should be the identity to me.

Strilanc:
Obsoleted by the introduction of `PauliString` and `PauliSum`.

Issue #491, opened 2018-06-20
Representing Hamiltonians


kevinsung:
I was thinking about how one would contribute a QAOA example (it would be nice to have that at release time), and my first thought is that one would want to represent the problem Hamiltonian, which is diagonal, as a sum of tensor products of Pauli Z's. This is most naturally represented as a QubitOperator (from OpenFermion). Are there plans to have a QubitOperator or something like that? Or, does anyone think there is a nice way to do the QAOA without it?

viathor:
I'm working on ways to represent and manipulate hermitian operators in cirq. Since these operators form a real Hilbert space in their own right they come with a rich set of operations that we may potentially want to implement:
* tensor products #932,
* compositions #962,
* linear combinations #1037.

Also, their cousins anti-Hermitian operators have a useful property that exponentiation yields a unitary providing us with a neat new way of specifying evolution operators.

--

BTW, there is a question that I'm trying to answer that you may have input to. How would you interpret an expression such as `cirq.X * cirq.X` that is likely to feature in many Hamiltonians? Would you read the asterisk as composition or as tensor product?

Craig said he sees it as the former (see https://github.com/quantumlib/Cirq/issues/932#issuecomment-422594075 and https://github.com/quantumlib/Cirq/issues/962#issue-362958426). Dave indicated he may read it as the latter (see https://github.com/quantumlib/Cirq/issues/932#issue-361506057).

Note that the ambiguity disappears when qubits are specified explicitly and so I am leaning towards requiring the use of GateOperations instead of Gates in writing down a Hamiltonian in cirq. Is this acceptable for your use-cases?

@dabacon @kevinsung @Strilanc

kevinsung:
My immediate reaction is that it is composition. A thought that pops into mind is to make `@` mean composition instead.

Yes, I think it makes sense to require using Operations when making Hamiltonians. I think Hamiltonians are operations that have qubits associated with them.

kevinsung:
This is done now that we have cirq.PauliSum.

PR #1299, opened 2019-01-18
Initial version of general linear operator

viathor:
My eventual aim is to build an abstraction that enables representation and manipulations of:
1. Hamiltonians (see #491).
2. Operators encountered frequently in quantum operations formalism (see #1037).
3. Measurement operators to allow efficient computation of averages of observables.

This PR has the basic functionality for 1 and 2. In particular, it implements basic operations in the space of linear operators (i.e. operator addition and scalar multiplication) as well as operator powers (for depolarizing channel) and exponentiation (to enable creation of gates defined by the Hamiltonian). Operator composition (#962) and tensor products (#932) are deferred. So is all functionality specific to 3.

Here are some example manipulations you can perform using the new functionality:

```
In [1]: import cirq, numpy as np

In [2]: hamiltonian = -np.pi * (cirq.XX + cirq.YY) / 4

In [3]: gate1 = cirq.make_gate(np.e**(-1j * hamiltonian))

In [4]: cirq.protocols.unitary(gate1)
Out[4]: 
array([[1.+0.j, 0.+0.j, 0.+0.j, 0.+0.j],
       [0.+0.j, 0.+0.j, 0.+1.j, 0.+0.j],
       [0.+0.j, 0.+1.j, 0.+0.j, 0.+0.j],
       [0.+0.j, 0.+0.j, 0.+0.j, 1.+0.j]])

In [5]: I = cirq.X**0

In [6]: channel = .5*I + .5*cirq.X + .5*cirq.Y + .5*cirq.Z

In [7]: str(channel**2)
Out[7]: '1.000*I+0.500*X+0.500*Y+0.500*Z'

In [8]: gate2 = cirq.make_gate(np.sqrt(1/3)*(cirq.X + cirq.Y + cirq.Z))

In [9]: cirq.protocols.unitary(gate2)
Out[9]: 
array([[ 0.57735027+0.j        ,  0.57735027-0.57735027j],
       [ 0.57735027+0.57735027j, -0.57735027+0.j        ]])

In [10]: gate3 = cirq.forge_gate(cirq.X + cirq.Y + cirq.Z)

In [11]: cirq.protocols.unitary(gate3)
Out[11]: 
array([[ 0.57735027+0.j        ,  0.57735027-0.57735027j],
       [ 0.57735027+0.57735027j, -0.57735027+0.j        ]])

```


dabacon:
Yeah!

kevinsung:
And all that works with the Gates replaced with GateOperations?

viathor:
Not yet, but expect GateOperations to gain similar functionality in a follow-up PR. I started with Gates since GateOperation has a reference to a gate which will make the follow-up easy once this is in.

kevinsung:
It looks like here, Hamiltonians with Pauli expansions that include multi-qubit Pauli strings are represented internally by the entire matrix, rather than the Pauli expansion. We need a representation that just stores the Pauli expansion without storing the huge matrix (like QubitOperator from OpenFermion). Can you explain the plan for that?

kevinsung:
Sorry, I see now that you can always just store the Pauli expansion. So my question is basically, how do you plan to generalize Pauli expansion to include multi-qubit strings?

cirq/ops/linear_operator.py
```diff
@@ -0,0 +1,314 @@
...
+        """Returns or computes the matrix corresponding to self."""
+        matrix = self._matrix_()
+        if matrix is not None:
+            return matrix
+
+        pauli_expansion = self._pauli_expansion_()
+        if pauli_expansion is None:
+            return None
+        return operator_spaces.reconstruct_from_expansion(
+            pauli_expansion, operator_spaces.PAULI_BASIS)
```
kevinsung: To leave room for the possibility of using different bases in the future, I'd rewrite this as
```
        pauli_expansion = self._pauli_expansion_()
        if pauli_expansion is not None:
            return operator_spaces.reconstruct_from_expansion(
                pauli_expansion, operator_spaces.PAULI_BASIS)

        return None

File: cirq/ops/linear_operator.py [hunk 2]
```diff
@@ -0,0 +1,314 @@
...
+    def _pauli_expansion_(self) -> Optional[np.ndarray]:
+        """Returns the expansion of the unitary operator in the Pauli basis.
+
+        Gates which know the expansion, should override.
+        """
+        pass
+
+
+class LinearOperator(AbstractLinearOperator):
+    """General linear operator on a vector space.
```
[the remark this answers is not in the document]
viathor: Re requirement for a matrix:
I do assume we're only considering finite dimensional vector spaces. Updated comment.

Re requirement for a Pauli expansion:
I do permit complex coefficients in the expansions. Note that while real linear combinations of Paulis generate the space of Hermitian operators, complex ones generate the entire space of all linear operators on a Hilbert space. You can see this easily by counting free parameters. For example `2 * a\dagger` is just X - 1j * Y which incidentally is used as test input in `test_operator_power()`.

That said, I think it's ok to introduce new internal representations that aren't general enough to describe all operators since any given instance of an operator is allowed a choice as to which representations to support.

cirq/ops/linear_operator.py
viathor: Done.

cirq/ops/linear_operator.py
```diff
@@ -0,0 +1,316 @@
...
+    below. This class in turn equips then with basic linear algebraic
+    operations.
+    """
+    @abc.abstractmethod
+    def _n_dim_(self) -> Optional[int]:
+        """Returns the number of dimensions of the space self acts on."""
+        pass
+
+    @abc.abstractmethod
+    def _matrix_(self) -> Optional[np.ndarray]:
```
kevinsung: I'm thinking this should be a protocol instead of an abstract method. I can imagine us wanting to support representations of linear operators that are not committed to a particular basis.

cirq/ops/linear_operator.py
```diff
@@ -0,0 +1,316 @@
...
+        """Returns the matrix representation of self."""
+        pass
+
+    @abc.abstractmethod
+    def _pauli_expansion_(self) -> Optional[np.ndarray]:
```
kevinsung: I'm thinking this should be a protocol instead of an abstract method. I can imagine us wanting to support representations of linear operators that are not committed to the Pauli basis.

kevinsung:
I made some more comments but I'm now realizing that I'm basically taking issue with this PR's commitment to the Pauli basis. I'm a bit biased because I want to use whatever gets implemented here to do stuff beyond the Pauli basis, but perhaps this is a good start anyway. @dabacon might have an opinion on whether we should commit to the Pauli basis here.

cirq/ops/linear_operator.py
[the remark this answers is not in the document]
viathor: There are two different questions here:
1. What set of operators is being modeled (e.g. positive, hermitian, general...)?
2. What representation of the operators is being used (e.g. standard basis/matrix, Pauli expansion)?

In the comment, I meant "general" w.r.t. question 1, i.e. to signify lack of restriction to any particular subset of L(V) (space of linear operators on V). Not "general" as a synonym for "basis-independent".

Initially, I considered modeling these two notions separately. I had a group of classes to model operators in basis-independent way and a second group of classes to host all representation/basis-dependent logic. Eventually, I decided against this separation to avoid proliferation of classes. We can go back to that approach if we do in fact end up using many representations. I'd say that more than 3-4 bases would make the sheer size of the classes in the present monolithic approach ripe for break-up.

I'll clarify the comment tomorrow. Let me know if you have specific suggestions about what to include or how to rename the classes.

kevinsung:
@viathor On further thought, I agree with your decision not to make a separate class for basis-independent representations. It makes sense to always assume some basis.

I'm interested in using a basis of fermionic creation and annihilation operators, like the FermionOperator data structure from OpenFermion. FermionOperator has some limitations that are very difficult to overcome in OpenFermion, so I have some interest in seeing how one might represent the same class of operators using the tools you are developing here. I was wrong before when I said that FermionOperators have no matrix; there is indeed a reasonable choice of standard basis: the occupation number basis.

I need to take a closer look but I think I'm actually mostly fine with this weak commitment to the Pauli basis.

cirq/ops/linear_operator.py
```diff
@@ -0,0 +1,316 @@
...
+    def __mul__(
+            self,
+            other: Union[complex, float, int]
+    ) -> 'AbstractLinearOperator':
+        """Computes scalar multiplication in the space of linear operators."""
+        pauli_expansion = self._pauli_expansion_()
+        if pauli_expansion is not None:
+            pauli_expansion = pauli_expansion * other
+
+        matrix = self._matrix_()
```
kevinsung: Same comment as addition: I don't think we want to be computing the entire matrix for basic operations.

kevinsung:
The comments I just made apply more generally: I think we should consider the basic representation of operators to be the operator expansion in the chosen basis, with the explicit matrix representation a last resort that's basically only ever computed when the user asks for it. Almost all Hamiltonians we would consider have a relatively small operator expansion (roughly speaking, a polynomial number of terms).

cirq/ops/linear_operator.py
```diff
@@ -0,0 +1,316 @@
...
+        """Computes vector addition in the space of linear operators."""
+        pauli_expansion = None
+        pauli_expansion_a = self._pauli_expansion_()
+        pauli_expansion_b = other._pauli_expansion_()
+        if pauli_expansion_a is not None and pauli_expansion_b is not None:
+            pauli_expansion = pauli_expansion_a + pauli_expansion_b
+
+        matrix = None
+        matrix_a = self.matrix()
+        matrix_b = other.matrix()
```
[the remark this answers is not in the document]
viathor: Done and added test to this effect. Deferring sparse representation to follow-up.

cirq/ops/linear_operator.py
viathor: `_matrix_()` doesn't compute the matrix if one isn't stored in the operator.

cirq/ops/linear_operator.py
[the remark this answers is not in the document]
viathor: Done.

cirq/ops/linear_operator.py
```diff
@@ -0,0 +1,316 @@
...
+
+        return LinearOperator(matrix, pauli_expansion)
+
+    def exp(self) -> 'AbstractLinearOperator':
+        """Computes operator power of number e."""
+        return self.__rpow__(np.e)
+
+    def polar_decomposition(self) -> Tuple['AbstractLinearOperator',
+                                           'AbstractLinearOperator']:
+        """Computes (right) polar decomposition of self."""
```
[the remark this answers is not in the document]
viathor: Done.

Strilanc:
Summary of conversation we had in person:

1) Specialization. If we want to support other bases, such as creation/annihilation, that will be handled as its own thing. We are focused purely on Paulis.

2) Multiplying Paulis (`cirq.X(a) * cirq.Y(b)`) and scaling Paulis (`cirq.X(a) * (1 + 1j)`) should produce a `cirq.PauliString`. This will require generalizing `PauliString` to have a complex scalar, instead of just 'negated'.

3) Adding Paulis and/or PauliStrings should produce a (to be named better) `LinearCombinationOfPauliStrings`. Our analogue of `QubitOperator` from OpenFermion, except we can't use that name because it's too generic.

viathor:
Good summary, thanks!

kevinsung:
I might be a little late to the party but I actually think it's weird to associate a coefficient with PauliString. I think multiplying a PauliString by a scalar should just return a LinearCombinationOfPauliStrings. In fact, isn't that necessary if LinearCombinationOfPauliStrings is to be stored as a dictionary from term to coefficient?

Also, perhaps the method `pass_operations_over` should be refactored out of PauliString to get rid of the need to have the `negated` attribute. That's another issue though.

viathor:
I agree, @kevinsung. Given that we'll be able to represent linear combinations of gates and operations, it is redundant to have a coefficient embedded in a gate or operation.

Strilanc:
The reason I want a coefficient on PauliString itself is because `PauliString * PauliString` should return a `PauliString`. For example, this is what I would need if I was writing a stabilizer simulator.

Strilanc:
@viathor What's left to pull out of this PR?

viathor:
@Strilanc Not much: the algorithm for computing integer powers of Pauli combinations; operator exponentiation and some cool unit tests.

viathor:
I'll take care of closing it when I'm done.

dabacon:
I think this has been completely mined.  Closing.

PR #1313, opened 2019-01-24
Set approx_eq default tolerance to 1e-8

mrwojtek:
As suggested by @dabacon.


cirq/protocols/approximate_equality.py
```diff
@@ -46,7 +46,7 @@ def _approx_eq_(
         pass
 
 
-def approx_eq(val: Any, other: Any, *, atol) -> bool:
+def approx_eq(val: Any, other: Any, *, atol: Union[int, float] = 1e-8) -> bool:
```
dabacon: maybe add a comment that 1e-8 was chosen to match numpy?

Issue #1307, opened 2019-01-23
.DS_Store?


vtomole:
https://github.com/quantumlib/Cirq/blob/master/.DS_Store

What do we use it for? @smitsanghavi

Strilanc:
Fixed by https://github.com/quantumlib/Cirq/pull/1314

PR #1314, opened 2019-01-25
Removing .DS_Store 

lkuligin:


vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/1307

PR #1317, opened 2019-01-27
Add random_superposition()

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/675

I made it a class because `random_superposition()` is a procedure that calls to random_circuit with a `simulate()` appended. A new file for that seemed unnecessary to me.

@Strilanc I don't know how to test ` random_superposition()`. I thought that something along the lines of `assert state.size) == 2**n_qubits` would work, but this is not the case.


kevinsung:
@vtomole , the implementation of `random_superposition` should return a vector chosen according to the uniform distribution over the unit sphere. This can be achieved with the following steps:
- For each component of the vector, choose the real and imaginary parts according to the standard normal distribution.
- Normalize the resulting vector.

As for testing the method, I think it's fine to just test that the dimension is correct and the vector has length one.

cirq/testing/random_superposition.py
```diff
@@ -0,0 +1,20 @@
...
+
+def random_superposition(size: int) -> np.ndarray:
+    """
+    Args:
+        size: Specified size returns a 2^size length array.
+    Returns:
+        Normalized random array.
+    """
+    state_vector = np.array([])
+    for _ in range(0, 2**size):
```
Strilanc: Nit: unnecessary `0,`

cirq/testing/random_superposition.py
```diff
@@ -0,0 +1,20 @@
+import numpy as np
+
+def normalize(array: np.ndarray):
+    norm = np.linalg.norm(array)
+    return array / norm
+
+def random_superposition(size: int) -> np.ndarray:
+    """
+    Args:
+        size: Specified size returns a 2^size length array.
```
Strilanc: This convention is opposite to the one we used for `random_unitary`. Take a `dim: int` argument instead. We should keep them consistent.

cirq/testing/random_circuit.py
```diff
@@ -37,13 +37,20 @@
 }  # type: Dict[ops.Gate, int]
 
 
-def random_circuit(qubits: Union[Sequence[ops.QubitId], int],
-                   n_moments: int,
-                   op_density: float,
-                   gate_domain: Optional[Dict[ops.Gate, int]] = None
-                   ) -> Circuit:
-    """Generates a random circuit.
+class RandomCircuit:
```
Strilanc: I think that this should stay as a method. The `Type(args).type()` style is odd. Normally people would call it a factory or name the method `generate` or similar. I don't think we need to resort to factories yet.

cirq/testing/random_circuit.py
```diff
@@ -52,40 +59,48 @@ def random_circuit(qubits: Union[Sequence[ops.QubitId], int],
...
+                gate, arity = choice(tuple(self.gate_domain.items()))
+                op_qubits = sample(free_qubits, arity)
+                free_qubits.difference_update(op_qubits)
+                if random() <= self.op_density:
+                    operations.append(gate(*op_qubits))
+            moments.append(ops.Moment(operations))
+
+        return Circuit(moments)
+
+    def random_superposition(self):
```
Strilanc: This method does not have a clear name. It's generating a superposition using the circuit as a source of entropy? But the randomness of the superposition will strongly depend on the gate set of the circuit.

I don't think this method is needed. What purpose did you have in mind for it?

cirq/testing/random_superposition.py
```diff
@@ -0,0 +1,20 @@
...
+    Returns:
+        Normalized random array.
+    """
+    state_vector = np.array([])
+    for _ in range(0, 2**size):
+        state_vector = np.append(
+            np.array([complex(np.random.standard_normal(1),
+                              np.random.standard_normal(1))]),
+            state_vector)
+    return normalize(state_vector)
```
kevinsung: Instead of looping, just tell numpy how many entries you want:
```
state_vector = np.random.normal(size=n)
state_vector += 1j * np.random.normal(size=n)
state_vector /= np.linalg.norm(state_vector)
return state_vector
```

cirq/testing/random_circuit.py
vtomole: After talking to Kevin, i will be getting rid of this class and following his (simpler) suggestion for `random_superposition`. I was going to get rid of this on the most recent commit.

cirq/testing/random_circuit.py
vtomole: Deleting the class.

cirq/testing/random_superposition.py
vtomole: Done.

cirq/testing/random_superposition.py
vtomole: Done.

PR #1320, opened 2019-01-27
Controlled gate decompose & ControlledOperation

### LINKED ISSUE #1284: Make ControlledGate work with gates that only provide a decomposition
The following should work:

```
import cirq


class G(cirq.TwoQubitGate):
        a, b = qubits
        yield cirq.X(a)**0.5
        yield cirq.H(b)
        yield cirq.CZ(a, b)


cg = cirq.ControlledGate(G())
x, y, z = cirq.LineQubit.range(3)
c = cirq.Circuit.from_ops(cg(x, y, z))
print(c.to_unitary_matrix())
```

but currently it raises an exception:

```
Operation without a known matrix or decomposition: cirq.ControlledGate(sub_gate=[...G...].on(cirq.LineQubit(0), cirq.LineQubit(1), cirq.LineQubit(2))
```

because `ControlledGate` doesn't have a `_decompose_` method. This issue is to add that method.

### LINKED ISSUE #1302: Define ControlledOperation
Blocking https://github.com/quantumlib/Cirq/issues/1284

Here is a start, but also needs most of the magic methods to be implemented and tested:

```python
class ControlledOperation(raw_types.Operation):
    def __init__(self,
                 control: raw_types.QubitId,
                 sub_operation: raw_types.Operation):
        self.control = control
        self.sub_operation = sub_operation

    def qubits(self):
        return (self.control,) + self.sub_operation.qubits

    def with_qubits(self, *new_qubits):
        return ControlledOperation(
            new_qubits[0],
            self.sub_operation.with_qubits(*new_qubits[1:]))
```

balopat:
Fixes #1302.
Fixes #1284.

Tasks left: 

- [x] Clarify questions
- [x] implement potentially missing magic methods + remove not needed ones
- [x] Write missing tests


balopat:
@Strilanc, can you clarify if the current approach is the right direction? 
More specifically I have some questions regarding the magic methods. I summarized in a comparison table here: 

| magic | ControlledOperation| GateOperation | ControlledGate |  
| -------- | -------- | -------- | ----- | 
| `_decompose_`     |  it is useful to implement so that the recursive decompose protocol can make use of it, it will be similar to the controlledGate._decompose_     | N/A     | In progress |
| `_value_equality_values_` | must be implemented for correct equality behavior | Implemented | ❗`self.sub_gate` - not sure if this is correct - I'd include the "controlness" | 
|`_has_unitary_`, `_unitary_` |must be implemented  - it will be similar to controlledGate | self.gate | 𝐼𝑐𝑜𝑛𝑡𝑟𝑜𝑙⊕𝑀𝑠𝑢𝑏𝑔𝑎𝑡𝑒 |
|`_apply_unitary` | ❓I _think_ this might be required, similar to ControlledGate | implemented, delegates to gate | implemented, complex logic |
|`_circuit_diagram_info_` | must be implemented, similar to controlledGate | circuit_diagram_info(self.gate)| '@', circuit_diagram_info(self.sub_gate)| 
|`_phase_by` | ❓not sure we need this right now| phase_by(self.gate) |  Not implemented |
|`__pow__` |yes, similar to ControlledGate | implemented, simple delegation | ControlledGate(pow(subgate))| 
|`_is_parameterized_`, `_resolve_parameters_` | yes, similar to ControlledGate | implemented, simple delegation|  implemented, simple delegation|
|`_trace_distance_bound_` | yes, similar to ControlledGate | implemented, simple delegation |  implemented, simple delegation|
|`__str__`, `__repr__` | yes, similar to ControlledGate + show the control qubit ❓ | implemented, simple | implemented, simple |
|`_qasm_` | ❓not required | implemented, delegates to gate | not implemented | 

Also, `ControlledOperation` shares a lot of similar logic with `ControlledGate` in the `*unitary` methods for example, and at first it might seem like a good idea to dedupe some of the shared logic - however I don't like any of the options that came to my mind:

a.) abstract out from `ControlledGate` and `ControlledOperation` into a third `Controlled"Something"`. This might end up being more confusing than the duplication 
b.) make ControlledGate into a ControlledOperation - that is confusing - Operation meant to be an application of a Gate (or something else) on certain qubits

Thus I'd just keep the duplication around in logic for now. What do you think?

cirq/ops/controlled_operation.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+@value.value_equality
+class ControlledOperation(raw_types.Operation):
+    def __init__(self,
+                 control: raw_types.QubitId,
+                 sub_operation: raw_types.Operation):
+        self.control = control
+        self.sub_operation = sub_operation
+
```
Strilanc: Nit: just one line between class methods.

cirq/ops/controlled_operation.py
```diff
@@ -0,0 +1,97 @@
...
+    def qubits(self):
+        return (self.control,) + self.sub_operation.qubits
+
+    def with_qubits(self, *new_qubits):
+        return ControlledOperation(
+            new_qubits[0],
+            self.sub_operation.with_qubits(*new_qubits[1:]))
+
+    def _decompose_(self):
+        result = protocols.decompose_once_with_qubits(self.sub_operation,
```
Strilanc: `decompose_once` instead of `decompose_once_with_qubits`, because the sub operation will already know its qubits. This is the main difference between operations and gates.

cirq/ops/controlled_operation.py
```diff
@@ -0,0 +1,97 @@
...
+        return protocols.has_unitary(self.sub_operation)
+
+    def _unitary_(self) -> Union[np.ndarray, NotImplementedType]:
+        sub_matrix = protocols.unitary(self.sub_operation, None)
+        if sub_matrix is None:
+            return NotImplemented
+        return linalg.block_diag(np.eye(sub_matrix.shape[0]), sub_matrix)
+
+    def __str__(self):
+        return 'C{}'.format(self.control) + str(self.sub_operation)
```
Strilanc: This `__str__` will look odd when using named qubits e.g `CaX(b, c)`. I would do `"C{}({})".format(self.sub_operation, ", ".join(str(q) for q in self.qubits))`, matching what you see when you use a controlled gate.

cirq/ops/controlled_operation.py
```diff
@@ -0,0 +1,97 @@
...
+                                   NotImplemented)
+        if new_sub_op is NotImplemented:
+            return NotImplemented
+        return ControlledOperation(self.control, new_sub_op)
+
+    def _circuit_diagram_info_(self,
+                               args: protocols.CircuitDiagramInfoArgs
+                               ) -> protocols.CircuitDiagramInfo:
+        sub_info = protocols.circuit_diagram_info(self.sub_operation,
+                                                  args,
```
Strilanc: Need to rewrite args to have the correct known qubit count and known qubits. I know ControlledGate didn't do that; it's a bug. https://github.com/quantumlib/Cirq/issues/1321

cirq/ops/controlled_operation_test.py
```diff
@@ -0,0 +1,52 @@
...
+                           cirq.ControlledOperation(c2,
+                                                    cirq.GateOperation(g1, q1)))
+    eq.make_equality_group(lambda:
+                           cirq.ControlledOperation(c2,
+                                                    cirq.GateOperation(g2, q1)))
+    eq.make_equality_group(lambda:
+                           cirq.ControlledOperation(c1,
+                                                    cirq.GateOperation(g1, q2)))
+
+## TODO(balintp): more tests to cover magic methods
```
Strilanc: Definitely more tests.

cirq/ops/controlled_operation_test.py
```diff
@@ -0,0 +1,52 @@
...
+    c1 = cirq.NamedQubit('c1')
+    q1 = [cirq.NamedQubit('q1')]
+    c2 = cirq.NamedQubit('c2')
+    q2 = [cirq.NamedQubit('q2')]
+
+    eq = cirq.testing.EqualsTester()
+
+    eq.make_equality_group(lambda:
+                           cirq.ControlledOperation(c1,
+                                                    cirq.GateOperation(g1, q1)))
```
Strilanc: You can use stuff like `cirq.X(q1)` or `g1(q1)` to make these easier to write.

cirq/ops/controlled_operation.py
balopat: right, copy-paste error. Thank you.

cirq/ops/controlled_operation.py
balopat: ack

cirq/ops/controlled_operation.py
balopat: ack

cirq/ops/controlled_operation.py
balopat: ack

Strilanc:
The python 2 errors are because you accidentally created a circular dependency.

balopat:
@Strilanc this is good to merge now I think - PTAL. Are you squash & merging typically or merging only (in which case I'll tidy up my commit history)?

cirq/ops/controlled_operation.py
```diff
@@ -0,0 +1,106 @@
...
+        return ControlledOperation(self.control, new_sub_op)
+
+    def _circuit_diagram_info_(self,
+                               args: protocols.CircuitDiagramInfoArgs
+                               ) -> Optional[protocols.CircuitDiagramInfo]:
+
+        sub_args = protocols.CircuitDiagramInfoArgs(
+            known_qubit_count=(args.known_qubit_count - 1
+                               if args.known_qubit_count is not None else None),
+            known_qubits=(tuple(b for b in args.known_qubits
```
Strilanc: The known qubits have to be in order, so `args.known_qubits[1:]` will work when it's not none.

cirq/ops/controlled_operation_test.py
```diff
@@ -0,0 +1,166 @@
...
+    q = cirq.NamedQubit('q')
+    g = cirq.Gate()
+    v = cirq.GateOperation(g, (q,))
+    c = cirq.ControlledOperation(cb, v)
+    assert c.sub_operation == v
+    assert c.control == cb
+    assert c.qubits == (cb, q)
+
+    eq = cirq.testing.EqualsTester()
+    eq.add_equality_group(c, c.with_qubits(cb, q))
```
Strilanc: Just assert that `c == c.with_qubits(cb, q)` since you already have a test doing equality group stuff.

cirq/ops/controlled_operation_test.py
```diff
@@ -0,0 +1,166 @@
...
+    eq.make_equality_group(lambda: cirq.ControlledOperation(c2, cirq.Z(q1)))
+
+
+def test_str():
+    c1 = cirq.NamedQubit('c1')
+    c2 = cirq.NamedQubit('c2')
+    q2 = cirq.NamedQubit('q2')
+
+    assert (str(cirq.ControlledOperation(c1, cirq.CZ(c2, q2))) ==
+            "C(c1)CZ(c2, q2)")
```
Strilanc: Should be `"CCZ(c1, c2, q2)"`

cirq/ops/controlled_operation_test.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def test_repr():
+    qubits = cirq.LineQubit.range(3)
+
+    assert (repr(cirq.ControlledOperation(qubits[0], cirq.CZ(*qubits[1:]))) ==
```
Strilanc: Use `cirq.testing.assert_equivalent_repr` to ensure the repr evaluates.

cirq/ops/controlled_operation_test.py
```diff
@@ -0,0 +1,166 @@
...
+    c = cirq.Circuit()
+    c.append(cirq.ControlledOperation(qubits[0], MultiH()(*qubits[1:])))
+
+    cirq.testing.assert_has_diagram(c, """
+0: ───@──────
+      │
+1: ───H(1)───
+      │
+2: ───H(2)───
+""", use_unicode_characters=True)
```
Strilanc: use unicode is the default, it can be omitted.

cirq/ops/controlled_operation_test.py
Strilanc: Oh, I see now that this might be difficult. The current str is acceptable.

cirq/ops/controlled_operation.py
balopat: ack

cirq/ops/controlled_operation.py
```diff
@@ -0,0 +1,106 @@
...
+        return protocols.has_unitary(self.sub_operation)
+
+    def _unitary_(self) -> Union[np.ndarray, NotImplementedType]:
+        sub_matrix = protocols.unitary(self.sub_operation, None)
+        if sub_matrix is None:
+            return NotImplemented
+        return linalg.block_diag(np.eye(sub_matrix.shape[0]), sub_matrix)
+
+    def __str__(self):
+        return 'C({})'.format(self.control) + str(self.sub_operation)
```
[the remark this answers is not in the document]
balopat: ack

cirq/ops/controlled_operation_test.py
balopat: ack

cirq/ops/controlled_operation_test.py
Strilanc: We can't assume that.

Strilanc:
We always squash merge.

cirq/ops/controlled_operation_test.py
balopat: if we can't assume anything about the structure of the suboperation's `str`, then this is our best bet.

PR #1323, opened 2019-01-28
Add QAOA example

kevinsung:
At first I thought I would want to use a Display, but the measurements here are not complicated enough to warrant that.

Another note: at small system sizes (even up to 16 qubits) this example pretty much always finds the optimal answer due to the fact that we sample so many bitstrings. I set n=6 so that the example circuit in the docstring would fit within 80 lines.


examples/qaoa.py
```diff
@@ -0,0 +1,166 @@
...
+                        for bitstring in itertools.product(range(2), repeat=n))
+
+    # Print the results
+    print('The largest cut value found was {}.'.format(largest_cut_value_found))
+    print('The largest possible cut has size {}.'.format(max_cut_value))
+    print('The approximation ratio achieved is {}.'.format(
+        largest_cut_value_found / max_cut_value))
+
+
+def Rzz(rads):
```
dabacon: nit: style would say make this rzz

examples/qaoa.py
```diff
@@ -0,0 +1,166 @@
...
+Optimizing objective function ...
+The largest cut value found was 7.
+The largest possible cut has size 7.
+The approximation ratio achieved is 1.0.
+"""
+
+import itertools
+
+import networkx
+import numpy as np
```
dabacon: examples/qaoa.py:54:0: C0411: third party import "import numpy as np" should be placed before "import networkx" (wrong-import-order)

kevinsung: Done

examples/qaoa.py
kevinsung: We have cirq.Rz though

examples/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 The Cirq Developers
+# Copyright 2018 The Cirq Developersdef test_example_runs_basic_arithmetic():
```
balopat:
```suggestion
# Copyright 2018 The Cirq Developers```

File: examples/qaoa.py (hunk 1 above)
Comment (dabacon) [reply examples/qaoa.py]: cirq.Rz is a class.  Here you have a method.

**Comment by kevinsung** [comment]
Actually, cirq.Rz is a method:
```
def Rz(rads: Union[float, sympy.Basic]) -> ZPowGate:
    """Returns a gate with the matrix e^{-i Z rads / 2}."""
    pi = sympy.pi if protocols.is_parameterized(rads) else np.pi
    return ZPowGate(exponent=rads / pi, global_shift=-0.5)
```

dabacon:
Method name should not be `Rx` in Cirq.

PR #1324, opened 2019-01-28
Adds DAG and permutation utils

bryano:
* CircuitDag.all_qubits
* CircuitDag.is_topologically_sorted
* get_logical_operations
* random_circuit_dag
* random_topological_sort


cirq/circuits/circuit_dag.py
```diff
@@ -175,8 +175,28 @@ def get_next_node(succ: networkx.classes.coreviews.AtlasView
...
+
     def to_circuit(self) -> circuit.Circuit:
         return circuit.Circuit.from_ops(
                     self.all_operations(),
                     strategy=circuit.InsertStrategy.EARLIEST,
                     device=self.device)
+
+    def is_topologically_sorted(self, operations: ops.OP_TREE):
+        remaining_dag = self.copy()
+        frontier = list(node for node in remaining_dag.nodes()
```
Strilanc: Nit: use `[...]` instead of `list(...)`

cirq/circuits/circuit_dag.py
```diff
@@ -175,8 +175,28 @@ def get_next_node(succ: networkx.classes.coreviews.AtlasView
...
+    def all_qubits(self):
+        return frozenset(q for node in self.nodes for q in node.val.qubits)
+
     def to_circuit(self) -> circuit.Circuit:
         return circuit.Circuit.from_ops(
                     self.all_operations(),
                     strategy=circuit.InsertStrategy.EARLIEST,
                     device=self.device)
+
+    def is_topologically_sorted(self, operations: ops.OP_TREE):
```
Strilanc: Add a docstring explaining what this does (...what *does* this do? Why are the operations being passed in? Aren't DAGs topologically sorted by definition?). Give examples of passing and failing operations.

cirq/circuits/circuit_dag.py
bryano: Done, but out of curiosity, why? Visual appeal?

cirq/circuits/circuit_dag.py
bryano: Done.

This tests that the operations are topologically sorted according to the DAG.

cirq/circuits/circuit_dag.py
```diff
@@ -175,8 +175,51 @@ def get_next_node(succ: networkx.classes.coreviews.AtlasView
...
+    def all_qubits(self):
+        return frozenset(q for node in self.nodes for q in node.val.qubits)
+
     def to_circuit(self) -> circuit.Circuit:
         return circuit.Circuit.from_ops(
                     self.all_operations(),
                     strategy=circuit.InsertStrategy.EARLIEST,
                     device=self.device)
+
+    def is_topologically_sorted(self, operations: ops.OP_TREE) -> bool:
```
Strilanc: Oh, the operations are supposed to be the same as the ones in the circuit? Maybe we should an exception if there's some missing or extra ones?

cirq/circuits/circuit_dag.py
Strilanc: It's just shorter and more common.

cirq/circuits/circuit_dag.py
Strilanc: Yes, that is the intention.

If a user passes in an unexpected operation, do you expect that they did so as a mistake or do you expect that they intended to use this functionality?

bryano: I'm not sure. I'm using this to test code that takes a `CircuitDag` and a `Device` and outputs a `Schedule` including swaps, so I'm only interested in whether it's exactly right or not.

Having to distinguish between "bad order" and "different set of operations" would complicate the logic of the function, so I'm tempted to not do that. Assuming both are treated the same way, I'd rather peacefully return `False` and let the caller deal with what to do next, rather than always throw an error. As a side note, I wish the device validation methods did the same; in many case I want to know whether or not an operation is valid rather than assert that it is.

bryano: @Strilanc ping

bryano:
Failing test unrelated to this PR. See #1425.

cirq/circuits/circuit_dag.py
Strilanc: I'm still not convinced that we want to add this. I'd prefer to have it in contrib.

@dabacon what is your opinion?

Strilanc:
Still waiting for @dabacon to comment

@dabacon @dabacon @dabacon

Strilanc:
Okay, well, if Dave isn't going to comment on it then I'm going to go with my gut instinct that this should be in `contrib` until we run into more use cases that need it.

Strilanc:
Still needs to be moved into contrib. Will close in two weeks if no activity.

cirq/testing/random_dag.py
```diff
@@ -0,0 +1,35 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any, Iterable
+
+from cirq import circuits
+from cirq.testing.random_circuit import random_circuit
+
+
+def random_circuit_dag(*args, **kwargs) -> circuits.Circuit:
```
Strilanc: This method isn't doing enough to justify itself, and it acts as a documentation barrier by having indirectly-defined mystery arguments. Delete it and inline it where you're using it.

bryano: Done

PR #1327, opened 2019-01-29
Extract wavefunction convenience methods into abstract StateVector class

### LINKED ISSUE #1231: Move bloch vector from StepResult to abstract class

andbe91:
Fixes #1231 

I moved bloch_vector along with density_matrix and dirac_notation to a new class called State. 
Following #1247, these new methods also require that QubitIds are used instead of indices. 

On a similar note, SimulationTrialResult still uses indices for its state methods, should this be changed to QubitId as well?

Let me know what you think.


cirq/sim/wave_function.py
```diff
@@ -15,17 +15,131 @@
...
 
-from cirq import linalg
+from cirq import linalg, ops
 
 if TYPE_CHECKING:
     # pylint: disable=unused-import
     from cirq.sim import simulator
 
 
+class State:
```
Strilanc: This name is *waaay* too vague. Rename it to e.g. Wavefunction or StateVector.

cirq/sim/wave_function.py
```diff
@@ -15,17 +15,131 @@
...
+        qubit_map: A map from the Qubits in the Circuit to the the index
+            of this qubit for a canonical ordering. This canonical ordering is
+            used to define the state (see the state() method).
+    """
+
+    def __init__(self, qubit_map: Optional[Dict[ops.QubitId, int]]):
+        self.qubit_map = qubit_map or {}
+
+    @abc.abstractmethod
+    def state(self) -> np.ndarray:
```
Strilanc: Too vague. wavefunction or statevector.

cirq/sim/wave_function.py
```diff
@@ -15,17 +15,131 @@
...
+
+        Args:
+            decimals: How many decimals to include in the pretty print.
+
+        Returns:
+            A pretty string consisting of a sum of computational basis kets
+            and non-zero floats of the specified accuracy."""
+        return dirac_notation(self.state(), decimals)
+
+    def density_matrix(self, qubits: List[ops.QubitId] = None) -> np.ndarray:
```
Strilanc: I would rename this to `density_matrix_of`, because it's specific to the given qubits instead of the entire state.

cirq/sim/wave_function.py
```diff
@@ -15,17 +15,131 @@
...
+            ValueError: if the size of the state represents more than 25 qubits.
+            IndexError: if the indices are out of range for the number of qubits
+                corresponding to the state.
+        """
+        return density_matrix_from_state_vector(
+            self.state(),
+            [self.qubit_map[q] for q in qubits] if qubits is not None else None
+        )
+
+    def bloch_vector(self, qubit: ops.QubitId) -> np.ndarray:
```
Strilanc: Might want `bloch_vector_of` to be consistent with the density matrix method. It's clear that it doesn't apply to the state as a whole though, so not as important.

PR #1330, opened 2019-01-31
characterization of single- and two-qubit states and gates

XiaoMiQC:
Added the following files:
-qubit_characterizations.py: containing functions for Rabi oscillations, single- and two-qubit randomized benchmarking, single- and two-qubit tomography.
-qubit_characterizations_test.py: tests for the characterization functions.
-qubit_characterizations_example.py: examples for using the new functions above.


Strilanc:
(It helps if you respond to the individual comments with e.g. done.)

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+    """Class for storing and plotting data from a Rabi oscillation
+    experiment.
+    """
+
+    def __init__(self, num_shots: int, rabi_angles: Sequence[float],
+                 excited_state_pops: Sequence[float]):
+        """
+        Args:
+            num_shots: The number of repetitions of the circuit for each Rabi
+              angle.
```
Strilanc: Nit: indent four spaces for line continuation of arg docs.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+from mpl_toolkits.mplot3d import Axes3D
+
+
+class RabiExperimentResult(object):
+    """Class for storing and plotting data from a Rabi oscillation
+    experiment.
+    """
+
+    def __init__(self, num_shots: int, rabi_angles: Sequence[float],
+                 excited_state_pops: Sequence[float]):
```
Strilanc: rename `pops` to `probs`

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+        self._num_shots = num_shots
+        self._rabi_angles = rabi_angles
+        self._excited_state_pops = excited_state_pops
+
+    @property
+    def data(self) -> Dict[str, Sequence[float]]:
+        return {'rabi_angles': self._rabi_angles,
+                'excited_state_probability': self._excited_state_pops}
+
+    def plot(self) -> None:
```
Strilanc: Add a docstring describing the plot.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+              of the Bloch sphere.
+            excited_state_pops: The corresponding probabilities that the
+            qubit is in the excited state.
+        """
+        self._num_shots = num_shots
+        self._rabi_angles = rabi_angles
+        self._excited_state_pops = excited_state_pops
+
+    @property
+    def data(self) -> Dict[str, Sequence[float]]:
```
Strilanc: Docstring describing the data.

In the future we may want to go even more granular than probabilities and do the individual samples. People may want to analyze it for e.g. bursts of too-many-1s. For now this is good.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+        Args:
+            num_cfds_seq: The different numbers of Cliffords in the RB study.
+            gnd_state_pops: The corresponding average ground state
+              probabilities.
+        """
+        self._num_cfds_seq = num_cfds_seq
+        self._gnd_state_pops = gnd_state_pops
+
+    @property
+    def data(self) -> Dict[str, Sequence[float]]:
```
Strilanc: docstring

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+        return {'number_of_cliffords': self._num_cfds_seq,
+                'ground_state_probability': self._gnd_state_pops}
+
+    def plot(self) -> None:
```
Strilanc: docstring

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+
+    def __init__(self, density_matrix: numpy.ndarray):
+        """
+        Args:
+            density_matrix: The density matrix obtained from tomography.
+        """
+        self._density_matrix = density_matrix
+
+    @property
+    def data(self) -> numpy.ndarray:
```
Strilanc: docstring

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+        return self._density_matrix
+
+    def plot(self) -> None:
```
Strilanc: docstring

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+        _plot_density_matrix(self._density_matrix)
+
+
+def rabi_oscillations(sampler: sim.SimulatesSamples, qubit: devices.GridQubit,
+                      final_angle: float, num_shots: int, num_points: int) \
```
Strilanc: Nit: don't use `\` for line continuation. Put the close parens on a new line instead.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+        num_shots: The number of repetitions of the circuit for each Rabi angle.
+        num_points: The number of Rabi angles.
+
+    Returns:
+        A RabiExperimentalResult object that stores and plots the result.
+    """
+    circuit = circuits.Circuit()
+    theta = value.Symbol('theta')
+    circuit.append(ops.X(qubit) ** theta)
+    circuit.append(ops.measure(qubit, key='z'))
```
Strilanc: Nit: use `circuits.Circuit.from_ops(...)`.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+            excited_state_pops: The corresponding probabilities that the
+            qubit is in the excited state.
+        """
+        self._num_shots = num_shots
+        self._rabi_angles = rabi_angles
+        self._excited_state_pops = excited_state_pops
+
+    @property
+    def data(self) -> Dict[str, Sequence[float]]:
+        return {'rabi_angles': self._rabi_angles,
```
Strilanc: How about returning a dictionary mapping the angle to the excited state probability? Ideally the user should not have to memory strings like 'rabi_angles'.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+            num_cfds_seq: The different numbers of Cliffords in the RB study.
+            gnd_state_pops: The corresponding average ground state
+              probabilities.
+        """
+        self._num_cfds_seq = num_cfds_seq
+        self._gnd_state_pops = gnd_state_pops
+
+    @property
+    def data(self) -> Dict[str, Sequence[float]]:
+        return {'number_of_cliffords': self._num_cfds_seq,
```
Strilanc: Perhaps map number of cliffords directly to ground state probability? Again don't want users to have to memory strings. Could also define a `NamedTuple` type to return, because then IDEs will autocomplete.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+        excited_state_pops[i] = numpy.mean(results[i].measurements['z'])
+
+    return RabiExperimentResult(num_shots, half_turns, excited_state_pops)
+
+
+def single_qubit_randomized_benchmarking(sampler: sim.SimulatesSamples,
+                                         qubit: devices.GridQubit,
+                                         num_cfds_seq: Sequence[int],
+                                         num_circuits: int, num_shots: int,
+                                         basis='xz') \
```
Strilanc: Nit: avoid `\`

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,527 @@
+from typing import Sequence, Tuple
+from cirq import circuits, devices, ops, protocols, sim, study, value
+import numpy
+from matplotlib import pyplot
+from mpl_toolkits.mplot3d import Axes3D
+
+simulator = sim.Simulator()
```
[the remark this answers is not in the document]
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,527 @@
...
+        yield _single_qubit_gates(S1_HALF_X_IN_XY[idx_4], q_1)
+
+
+def _single_qubit_gates(gate_seq, qubit):
+    for gate in gate_seq:
+        yield gate(qubit)
+
+
+_identity = ops.XPowGate(exponent=0.0)
+_rot_x = ops.XPowGate(exponent=1.0)
```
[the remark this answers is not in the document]
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
Strilanc: No type annotation on basis. I would prefer if we had IDE-autocompletable arguments such as `use_xy_basis: bool = True`.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+        gnd_pops.append(1.0 - numpy.mean(excited_pops_l))
+
+    return RandomizedBenchmarkingResult(num_cfds_seq, gnd_pops)
+
+
+def two_qubit_randomized_benchmarking(sampler: sim.SimulatesSamples,
+                                      q_0: devices.GridQubit,
+                                      q_1: devices.GridQubit,
+                                      num_cfds_seq: Sequence[int],
+                                      num_circuits: int, num_shots: int) \
```
Strilanc: `\`

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+    Args:
+        sampler: The quantum engine or simulator to run the circuits.
+        qubit: The qubit under test.
+        circuit: The circuit to execute on the qubit before tomography.
+        num_shots: The number of measurements for each basis rotation.
+
+    Returns:
+        A StateTomographyResult object that stores and plots the density matrix.
+    """
+    circuit_z = circuit.copy()
```
Strilanc: You can use `circuit + circuits.Circuit.from_ops(ops.measure(qubit, key='z'))` to generate the modified circuits.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+        q_1: The second qubit under test.
+        circuit: The circuit to execute on the qubits before tomography.
+        num_shots: The number of measurements for each basis rotation.
+
+    Returns:
+        A StateTomographyResult object that stores and plots the density matrix.
+    """
+
+    def _measurement(circ):
+        circ.append(ops.measure(q_0, q_1, key='z'))
```
Strilanc: This is a dangerous variable name to use in a library called cirq ;)

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+        q_0: The first qubit under test.
+        q_1: The second qubit under test.
+        circuit: The circuit to execute on the qubits before tomography.
+        num_shots: The number of measurements for each basis rotation.
+
+    Returns:
+        A StateTomographyResult object that stores and plots the density matrix.
+    """
+
+    def _measurement(circ):
```
Strilanc: type annotations

Strilanc: You are shadowing the outside variable called circ.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+    sigma_1 = numpy.array([[0.0, 1.0], [1.0, 0.0]]) / 2.0
+    sigma_2 = numpy.array([[0.0, -1.0j], [1.0j, 0.0]]) / 2.0
+    sigma_3 = numpy.array([[1.0, 0.0], [0.0, -1.0]]) / 2.0
+    sigmas = [sigma_0, sigma_1, sigma_2, sigma_3]
+
+    pops = []
+    rots = [ops.X ** 0, ops.X ** 0.5, ops.Y ** 0.5]
+    for i in range(3):
+        for j in range(3):
+            circ = circuit.copy()
```
Strilanc: Not actually using the indices. Just iterate over rots directly.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+            circ.append(rots[i](q_1))
+            circ.append(rots[j](q_0))
```
Strilanc: `circuit + ...` will shorten this up.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+            pops.extend(_measurement(circ))
+
+    mat = numpy.zeros((27, 15))
+    s = numpy.array([[1.0, 1.0, 1.0], [-1.0, 1.0, -1.0], [1.0, -1.0, -1.0]])
+    mat[0:3, [2, 11, 14]] = s * numpy.tile([1.0, 1.0, 1.0], (3, 1))
```
Strilanc: There's way too many constants here. Is there a way to generate them automatically? At the very least they need to be documented.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+    c1_in_xy.append([rot_half_y, rot_half_x])
+    c1_in_xy.append([rot_half_y, rot_neg_half_x])
+    c1_in_xy.append([rot_neg_half_y, rot_half_x])
+    c1_in_xy.append([rot_neg_half_y, rot_neg_half_x])
+    c1_in_xy.append([rot_neg_half_x, rot_neg_half_y])
+    c1_in_xy.append([rot_half_x, rot_neg_half_y])
+    c1_in_xy.append([rot_neg_half_x, rot_half_y])
+    c1_in_xy.append([rot_half_x, rot_half_y])
+
+    c1_in_xz = []
```
Strilanc: These two giant lists can definitely be merged somehow.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,567 @@
...
+    c1_in_xz.append([rot_x])
+    c1_in_xz.append([rot_z, rot_x])
+    c1_in_xz.append([rot_z])
+    c1_in_xz.append([rot_half_x])
+    c1_in_xz.append([rot_neg_half_x])
+    c1_in_xz.append([rot_half_x, rot_half_z, rot_neg_half_x])
+    c1_in_xz.append([rot_half_x, rot_neg_half_z, rot_neg_half_x])
+    c1_in_xz.append([rot_half_z])
+    c1_in_xz.append([rot_neg_half_z])
+    c1_in_xz.append([rot_neg_half_x, rot_neg_half_z, rot_neg_half_x])
```
Strilanc: There must be a better way to generate this giant list. I think we have a way to enumerate cliffords via `cirq.ops.SingleQubitCliffordGate`. You could then do a decomposition of each into the XY gate set or XZ gate set with math.

cirq/experiments/__init__.py
```diff
@@ -3,3 +3,11 @@
...
     generate_supremacy_circuit_google_v2_grid,
 )
+
+from cirq.experiments.qubit_characterizations import (
+    rabi_oscillations,
+    single_qubit_randomized_benchmarking,
+    two_qubit_randomized_benchmarking,
+    single_qubit_state_tomography,
+    two_qubit_state_tomography,
+)
```
vtomole: newline needed.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: All "...pops" have been renamed to "...probs"

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done. \ is no longer used for line continuations anywhere.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done. Also made the same modification for the data property in RBResult.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: \ removed

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Renamed to two_qubit_circuit

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

XiaoMiQC: Shadowing removed after renaming variables

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Thanks for pointing this out. Yes, it was difficult to read. The idea is to build and solve an over-determined set of linear equations of the form M*C = P, where C is a vector of length 15 containing the coefficients for the operator-sum representation of the density matrix \rho = \sum_{i, j = 0, 1, 2, 3} c_{ij} \sigma_i *\ sigma_j (excluding c_00 which is set to 1). P is a vector of length 27 containing the probabilities P_00, P_01, P_10 from the nine measurements after different basis rotations. M is a 27 x 15 matrix where each row contains three non-zero elements \in {-1, 1} which one can determine from the basis rotations. I initially just did this on paper and hard-coded the matrix M.

In the new code, M is generated more compactly within the nested for loop. Hopefully this looks less cluttered. I have also added more explanation of the above method in the doc string.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: The API in ops.SingleQubitCliffordGate is a bit confusing to me. I have rewritten this function using some for loops and itertools.product. It is much shorter now.

cirq/experiments/__init__.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
+import itertools
+from typing import Sequence, Tuple
+from cirq import circuits, devices, ops, protocols, sim, study, value
+import numpy
```
Strilanc: `as np`

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+from matplotlib import pyplot
+from mpl_toolkits.mplot3d import Axes3D
+
+
+class RabiResult(object):
+    """Class for storing and plotting data from a Rabi oscillation
```
Strilanc: `Class for` is redundant. Just `Results from a Rabi oscillation experiment.` is fine.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+        """
+        fig = pyplot.figure()
+        pyplot.plot(self._rabi_angles, self._excited_state_probs, 'ro-',
+                    figure=fig)
+        pyplot.xlabel(r"Rabi Angle ($\pi$)")
+        pyplot.ylabel('Excited State Probability')
+
+
+class RBResult(object):
+    """Class for storing and plotting data from a randomized benchmarking
```
Strilanc: Redundant `Class for`

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+    experiment.
+    """
+
+    def __init__(self, num_cfds_seq: Sequence[int],
+                 gnd_state_probs: Sequence[float]):
+        """
+        Args:
+            num_cfds_seq: The different numbers of Cliffords in the RB study.
```
Strilanc: Don't use abbreviations for API elements exposed to the user. Spell out `ground` and `clifford`.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+        """
+        fig = pyplot.figure()
+        pyplot.plot(self._num_cfds_seq, self._gnd_state_probs, 'ro-',
+                    figure=fig)
+        pyplot.xlabel(r"Number of Cliffords")
+        pyplot.ylabel('Ground State Probability')
+
+
+class TomographyResult(object):
+    """Class for storing and plotting a density matrix obtained from a state
```
Strilanc: Redundant `Class for`

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+        Cliffords in the RB study.
+        """
+        fig = pyplot.figure()
+        pyplot.plot(self._num_cfds_seq, self._gnd_state_probs, 'ro-',
+                    figure=fig)
+        pyplot.xlabel(r"Number of Cliffords")
+        pyplot.ylabel('Ground State Probability')
+
+
+class TomographyResult(object):
```
Strilanc: Remove all `(object)` inheritance markers. They are redundant in python 3.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+        """Plots the real and imaginary parts of the density matrix as two
+        3D bar plots.
+        """
+        _plot_density_matrix(self._density_matrix)
+
+
+def rabi_oscillations(sampler: sim.SimulatesSamples, qubit: devices.GridQubit,
+                      final_angle: float, num_shots: int,
+                      num_points: int) -> RabiResult:
+    """
```
Strilanc: Set summary line of docstring (the one inline with the """) to `Runs a Rabi oscillation experiment.`. Empty line after that.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+    return RabiResult(half_turns, excited_state_probs)
+
+
+def single_qubit_randomized_benchmarking(sampler: sim.SimulatesSamples,
+                                         qubit: devices.GridQubit,
+                                         num_cfds_seq: Sequence[int],
+                                         num_circuits: int, num_shots: int,
+                                         use_xy_basis: bool = True) -> RBResult:
+    """
+    Clifford-based randomized benchmarking (RB) of a single qubit.
```
Strilanc: Summary line goes inline with `"""`.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+    excited_state_probs = numpy.zeros(num_points)
+    for i in range(num_points):
+        excited_state_probs[i] = numpy.mean(results[i].measurements['z'])
+
+    return RabiResult(half_turns, excited_state_probs)
+
+
+def single_qubit_randomized_benchmarking(sampler: sim.SimulatesSamples,
+                                         qubit: devices.GridQubit,
+                                         num_cfds_seq: Sequence[int],
```
Strilanc: Don't abbreviate.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+            excited_probs_l.append(numpy.mean(results.measurements['z']))
+        gnd_probs.append(1.0 - numpy.mean(excited_probs_l))
+
+    return RBResult(num_cfds_seq, gnd_probs)
+
+
+def two_qubit_randomized_benchmarking(sampler: sim.SimulatesSamples,
+                                      q_0: devices.GridQubit,
+                                      q_1: devices.GridQubit,
+                                      num_cfds_seq: Sequence[int],
```
Strilanc: Don't abbreviate.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+                                      num_circuits: int,
+                                      num_shots: int) -> RBResult:
+    """
+    Clifford-based randomized benchmarking (RB) of two qubits.
```
Strilanc: Summary inline with """.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+            circuit.append(ops.measure(qubit, key='z'))
+            results = sampler.run(circuit, repetitions=num_shots)
+            excited_probs_l.append(numpy.mean(results.measurements['z']))
+        gnd_probs.append(1.0 - numpy.mean(excited_probs_l))
+
+    return RBResult(num_cfds_seq, gnd_probs)
+
+
+def two_qubit_randomized_benchmarking(sampler: sim.SimulatesSamples,
+                                      q_0: devices.GridQubit,
```
Strilanc: Maybe a combined `qubits`? I don't feel strongly about it, just seems like `q_0` is an odd parameter name to expose to the user.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+
+    return RBResult(num_cfds_seq, gnd_probs)
+
+
+def single_qubit_state_tomography(sampler: sim.SimulatesSamples,
+                                  qubit: devices.GridQubit,
+                                  circuit: circuits.Circuit,
+                                  num_shots: int) -> TomographyResult:
+    """
+    Single-qubit state tomography.
```
Strilanc: Inline with """

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+    rho_11 = numpy.mean(results.measurements['z'])
+    rho_00 = 1.0 - rho_11
+
+    circuit_x = circuit.copy()
+    circuit_x.append(ops.X(qubit) ** 0.5)
+    circuit_x.append(ops.measure(qubit, key='z'))
+    results = sampler.run(circuit_x, repetitions=num_shots)
+    rho_01_im = numpy.mean(results.measurements['z']) - 0.5
+
+    circuit_y = circuit.copy()
```
Strilanc: I think this might work:

```python
circuit_y = circuits.Circuit.from_ops(
    circuit,
    ops.Y(qubit)**-0.5,
    ops.measure(qubit, key='z'))
```

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+
+    return TomographyResult(rho)
+
+
+def two_qubit_state_tomography(sampler: sim.SimulatesSamples,
+                               q_0: devices.GridQubit, q_1: devices.GridQubit,
+                               circuit: circuits.Circuit,
+                               num_shots: int) -> TomographyResult:
+    """
+    Two-qubit state tomography.
```
Strilanc: Inline with """

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+    rho_01 = rho_01_re + 1j * rho_01_im
+    rho_10 = numpy.conj(rho_01)
+
+    rho = numpy.array([[rho_00, rho_01], [rho_10, rho_11]])
+
+    return TomographyResult(rho)
+
+
+def two_qubit_state_tomography(sampler: sim.SimulatesSamples,
+                               q_0: devices.GridQubit, q_1: devices.GridQubit,
```
Strilanc: Consider merging into `qubits` arg.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+        num_shots: The number of measurements for each basis rotation.
+
+    Returns:
+        A StateTomographyResult object that stores and plots the density matrix.
+    """
+
+    def _measurement(two_qubit_circuit: circuits.Circuit) -> Sequence[float]:
+        two_qubit_circuit.append(ops.measure(q_0, q_1, key='z'))
+        results = sampler.run(two_qubit_circuit, repetitions=num_shots)
+        bit_strings = [r for r in results.measurements['z']]
```
Strilanc: Use `results.histogram` instead of counting for yourself.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+
+    sigma_0 = numpy.eye(2) / 2.0
+    sigma_1 = numpy.array([[0.0, 1.0], [1.0, 0.0]]) / 2.0
+    sigma_2 = numpy.array([[0.0, -1.0j], [1.0j, 0.0]]) / 2.0
+    sigma_3 = numpy.array([[1.0, 0.0], [0.0, -1.0]]) / 2.0
+    sigmas = [sigma_0, sigma_1, sigma_2, sigma_3]
+
+    probs = []
+    rots = [ops.X ** 0, ops.X ** 0.5, ops.Y ** 0.5]
+    mat = numpy.zeros((27, 15))
```
Strilanc: Magic numbers 27 and 15

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+    s = numpy.array([[1.0, 1.0, 1.0], [-1.0, 1.0, -1.0], [1.0, -1.0, -1.0]])
```
Strilanc: What determines the signs?

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+
+    c, _, _, _ = numpy.linalg.lstsq(mat, 4.0 * numpy.array(probs) - 1.0,
+                                    rcond=None)
+    c = numpy.concatenate(([1.0], c))
+    c = c.reshape(4, 4)
+
+    rho = numpy.zeros((4, 4))
+    for i in range(4):
+        for j in range(4):
+            rho = rho + c[i, j] * numpy.kron(sigmas[i], sigmas[j])
```
Strilanc: `+=`

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+    q_0_i = 3 - i
+    q_1_j = 3 - j
+    indices = [q_1_j - 1, 4 * q_0_i - 1, 4 * q_0_i + q_1_j - 1]
+    signs = [(-1) ** (j == 2), (-1) ** (i == 2), (-1) ** (i == 2 + j == 2)]
+    return mat_idx, indices, signs
+
+
+def _two_qubit_clifford_matrices(q_0, q_1, c1, s1, s1_x, s1_y):
+    mats = []
+    for i in range(11520):
```
Strilanc: 11520?

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+        circuit = circuits.Circuit()
+        circuit.append(_two_qubit_clifford(q_0, q_1, i, c1, s1, s1_x, s1_y))
```
Strilanc: `Circuit.from_ops`

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+
+
+def _random_single_q_clifford(qubit, num_cfds, cfds, cfd_matrices):
+    gate_ids = list(numpy.random.choice(24, num_cfds))
+    gate_sequence = []
+    for gate_id in gate_ids:
+        gate_sequence.extend(cfds[gate_id])
+    idx = _find_inv_matrix(_gate_seq_to_mats(gate_sequence), cfd_matrices)
+    gate_sequence.extend(cfds[idx])
+    circuit = circuits.Circuit()
```
Strilanc: `Circuit.from_ops(gate(qubit) for gate in gate_sequence)`

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+    for gate in gate_sequence:
+        circuit.append(gate(qubit))
+    return circuit
+
+
+def _random_two_q_clifford(q_0, q_1, num_cfds, cfd_matrices, c1, s1, s1_x,
```
Strilanc: Type annotations

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+
+
+def _find_inv_matrix(mat: numpy.ndarray, mat_sequence: numpy.ndarray) -> int:
+    mat_prod = numpy.einsum('ij,...jk->...ik', mat, mat_sequence)
+    diag_sums = list(numpy.absolute(numpy.einsum('...ii->...', mat_prod)))
+    idx = diag_sums.index(max(diag_sums))
+    return idx
+
+
+def _matrix_bar_plot(mat, z_label: str, kets=None, title=None) -> None:
```
Strilanc: Type annotations

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+
+
+def _gate_seq_to_mats(gate_seq: Sequence[ops.Gate]):
+    mat_rep = protocols.unitary(gate_seq[0])
+    for gate in gate_seq[1:]:
+        mat_rep = numpy.dot(protocols.unitary(gate), mat_rep)
+    return mat_rep
+
+
+def _two_qubit_clifford(q_0, q_1, idx, c1, s1, s1_x, s1_y):
```
Strilanc: Type annotations

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+    if idx_2 == 1:
+        yield ops.CZ(q_0, q_1)
+        yield ops.Y(q_0) ** -0.5
+        yield ops.Y(q_1) ** 0.5
+        yield ops.CZ(q_0, q_1)
+        yield ops.Y(q_0) ** 0.5
+        yield ops.Y(q_1) ** -0.5
+        yield ops.CZ(q_0, q_1)
+        yield ops.Y(q_1) ** 0.5
+    elif 1 < idx_2 < 11:
```
Strilanc: This needs comments. Where are the numbers from?

Also, make this 11 match the 10 in the next clause by using <= or >= in one of them.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+        yield ops.Y(q_0) ** 0.5
+        yield ops.X(q_1) ** -0.5
+        yield ops.CZ(q_0, q_1)
+        idx_3 = int((idx_2 - 11) / 3)
+        idx_4 = (idx_2 - 11) % 3
+        yield _single_qubit_gates(s1_y[idx_3], q_0)
+        yield _single_qubit_gates(s1_x[idx_4], q_1)
+
+
+def _single_qubit_gates(gate_seq, qubit):
```
Strilanc: Type annotations

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,544 @@
...
+    for gate in gate_seq:
+        yield gate(qubit)
+
+
+def _single_qubit_cliffords():
```
Strilanc: Type annotations

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: I rename q_0 and q_1 to first_qubit and second_qubit. Hopefully these look more intuitive. Same for the signatures in two qubit tomography.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Cool, it worked. Thanks.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: renamed to first_qubit and second_qubit.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done. Cool, this really shortens things.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: These are now cast into constants num_cols and num_rows which are set at the beginning of the function, and an annotation indicating that they denote the size of the linear equation system is added for clarification.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: They are the relative signs for the equations for P_00, P_01, P_10. I have expanded the doc string to include some example linear equations which hopefully further clarify how the density matrix is obtained from all the measured probabilities.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Got error: Cannot cast ufunc add output from dtype('complex128') to dtype('float64') with casting rule 'same_kind'.

I can in principle do `rho += (c[i, j] * numpy.kron(sigmas[i], sigmas[j])).astype('complex')` but it seems more complicated

cirq/experiments/qubit_characterizations.py
XiaoMiQC: This is the total number of two-qubit Cliffords. It's now made clear with a comment.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: This function now has a docstring explaining how a two-qubit Clifford is generated.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,639 @@
...
+
+    def plot(self) -> None:
+        """Plots the average ground state probability vs the number of
+        Cliffords in the RB study.
+        """
+        fig = pyplot.figure()
+        pyplot.plot(self._num_cfds_seq, self._gnd_state_probs, 'ro-',
+                    figure=fig)
+        pyplot.xlabel(r"Number of Cliffords")
+        pyplot.ylabel('Ground State Probability')
```
Strilanc: Finish all plot methods with `fig.show()` or equivalent. We can add a parameter later to guard that off, if needed.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,639 @@
...
+
+    def plot(self) -> None:
+        """Plots the real and imaginary parts of the density matrix as two
+        3D bar plots.
+        """
+        _plot_density_matrix(self._density_matrix)
+
+
+def rabi_oscillations(sampler: sim.SimulatesSamples, qubit: devices.GridQubit,
+                      final_angle: float, number_shots: int,
```
Strilanc: Final angle needs to be in radians. We can convert it internally, but people running rabi oscillation experiments will be thinking in radians instead of powers of Paulis.

Final angle should default to pi or 2pi; your choice. Maybe rename to `max_angle`.

Is there a reasonable default we could use for the number of repetitions or the number of points? E.g. one thousand repetitions and one hundred angles?

Force keyword parameters for shots and points by adding a `, *,` parameter before them, so that it isn't unclear what. Rename them to `repetitions` and `num_points` for better consistency with other methods in cirq.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,639 @@
...
+        around the x-axis).
+        """
+        fig = pyplot.figure()
+        pyplot.plot(self._rabi_angles, self._excited_state_probs, 'ro-',
+                    figure=fig)
+        pyplot.xlabel(r"Rabi Angle ($\pi$)")
+        pyplot.ylabel('Excited State Probability')
+
+
+class RBResult:
```
Strilanc: Don't abbreviate.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,639 @@
...
+                                                 self._gnd_state_probs)]
+
+    def plot(self) -> None:
+        """Plots the average ground state probability vs the number of
+        Cliffords in the RB study.
+        """
+        fig = pyplot.figure()
+        pyplot.plot(self._num_cfds_seq, self._gnd_state_probs, 'ro-',
+                    figure=fig)
+        pyplot.xlabel(r"Number of Cliffords")
```
Strilanc: Pass the `figure` argument to xlabel and ylabel to avoid touching global state. Also do this in other plot methods.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,639 @@
...
+        """
+        return [(angle, prob) for angle, prob in zip(self._rabi_angles,
+                                                     self._excited_state_probs)]
+
+    def plot(self) -> None:
+        """Plots excited state probability vs the Rabi angle (angle of rotation
+        around the x-axis).
+        """
+        fig = pyplot.figure()
+        pyplot.plot(self._rabi_angles, self._excited_state_probs, 'ro-',
```
Strilanc: When I used this method, my first reaction was "I want the points to be smaller". I think we can very hackily give control over that kind of thing by taking a **kwargs argument, and forwarding it into the `plot` call. Then users could say `results.plot(color='black')` or `results.plot(markersize=0.5, linewidth=0.2)` or etc.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,639 @@
...
+    results = sampler.run_sweep(circuit, params=sweep, repetitions=number_shots)
+    half_turns = np.linspace(0.0, final_angle, number_points)
+    excited_state_probs = np.zeros(number_points)
+    for i in range(number_points):
+        excited_state_probs[i] = np.mean(results[i].measurements['z'])
+
+    return RabiResult(half_turns, excited_state_probs)
+
+
+def single_qubit_randomized_benchmarking(sampler: sim.SimulatesSamples,
```
Strilanc: If you can pick reasonable defaults, do so.

Force keyword parameters after the qubit parameter.

Use `num_` instead of `number[s]_` to be more consistent with the rest of the library. `repetitions` instead of `number_shots`.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,639 @@
...
+            circuit = _random_single_q_clifford(qubit, num_cfds, c1, cfd_mats)
+            circuit.append(ops.measure(qubit, key='z'))
+            results = sampler.run(circuit, repetitions=number_shots)
+            excited_probs_l.append(np.mean(results.measurements['z']))
+        gnd_probs.append(1.0 - np.mean(excited_probs_l))
+
+    return RBResult(numbers_cliffords, gnd_probs)
+
+
+def two_qubit_randomized_benchmarking(sampler: sim.SimulatesSamples,
```
Strilanc: Reasonable defaults if possible, forced keyword args, `num_`, `repetitions`.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,639 @@
...
+            circuit.append(ops.measure(first_qubit, second_qubit, key='z'))
+            results = sampler.run(circuit, repetitions=number_shots)
+            gnds = [(not r[0] and not r[1]) for r in results.measurements['z']]
+            gnd_probs_l.append(np.mean(gnds))
+        gnd_probs.append(float(np.mean(gnd_probs_l)))
+
+    return RBResult(numbers_cliffords, gnd_probs)
+
+
+def single_qubit_state_tomography(sampler: sim.SimulatesSamples,
```
Strilanc: We can leave these tomography methods for now, but I think they make more sense as "Displays" that can be inserted into circuits. I'm unsure how to best expose the functionality to users. It's likely that someone new who wants to do state tomography will look for a method like this instead of in a display concept they don't know of yet. On the other hand, it's a bit weird for it to be in experiments..

Regardless, it's good to have the code ready to go. It can be refactored into the desired form at a later time.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,639 @@
...
+                                  qubit: devices.GridQubit,
+                                  circuit: circuits.Circuit,
+                                  number_shots: int) -> TomographyResult:
```
Strilanc: I want to see to rename this to `repetitions`, but it feels different from the other experiments. Like the handle the user wants is different here. Unsure.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,639 @@
...
+    of the iSWAP-like class (a total of 3*3*24*24 = 5184 possibilities).
+
+    Through the above process, all 11520 members of the two-qubit Clifford
+    group may be generated.
+
+    Args:
+        q_0: The first qubit under test.
+        q_1: The second qubit under test.
+        idx: An integer from 0 to 11519.
+        c1: Contains all 24 single-qubit Cliffords.
```
Strilanc: Isn't this a constant? Why is it being passed in?

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: All done.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Yeah this sounds like a good idea. Done.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Yeah the display method makes sense. I can refactor the code to do that later on if needs be.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: I renamed num_shots to repetitions. Although it does feel a bit different, it still is the parameter that gets fed to the repetitions parameter in the .run method.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Are you referring to the gate sequences in c1, s1 etc? Yes, they are constants. However, they are generated outside this function so that the code doesn't have to call `_single_qubit_cliffords()` and build all the lists of gates every time it needs to generate a random two-qubit Clifford.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,659 @@
+import itertools
+from typing import Sequence, Tuple, Iterator, Any
+
+import sympy
+
+from cirq import circuits, devices, ops, protocols, sim, study
+import numpy as np
```
Strilanc: Nit: third party imports before cirq imports

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,659 @@
...
+        matrix of the n-qubit system.
+        """
+        return self._density_matrix
+
+    def plot(self) -> None:
+        """Plots the real and imaginary parts of the density matrix as two
+        3D bar plots.
+        """
+        fig_re, fig_imag = _plot_density_matrix(self._density_matrix)
+        fig_re.show()
```
Strilanc: Show both figures simultaneously without one blocking the other, e.g. by merging the plots into one figure if necessary. When the user calls plot they should be seeing all the data right away.

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,659 @@
...
+
+    The above process is done for different numbers of Cliffords specified in
+    num_cliffords.
+
+    See Barends et al., Nature 508, 500 for details.
+
+    Args:
+        sampler: The quantum engine or simulator to run the circuits.
+        qubit: The qubit under test.
+        num_cliffords: The different numbers of Cliffords in the RB study.
```
Strilanc: Describe more generally why one would want to change this `num_cliffords` value, and what valid values look like. Can we pick a default that works well? Repeat for all methods with this argument.

It seems that this is setting the circuit sizes that will be sampled from, right? In that case I would try to use a different name, because `num_*` sounds like it should be an integer but this is a sequence of integers. Maybe `clifford_counts`? `lengths`?

Oh, maybe `num_clifford_range`! A `range(5, 20)` argument works right?

cirq/experiments/qubit_characterizations.py
```diff
@@ -0,0 +1,659 @@
...
+        excited_state_probs[i] = np.mean(results[i].measurements['z'])
+
+    return RabiResult(angles, excited_state_probs)
+
+
+def single_qubit_randomized_benchmarking(sampler: sim.SimulatesSamples,
+                                         qubit: devices.GridQubit,
+                                         num_cliffords: Sequence[int],
+                                         use_xy_basis: bool = True,
+                                         *, num_circuits: int = 20,
```
Strilanc: Put the * before `num_cliffords`, and on its own line.

cirq/experiments/qubit_characterizations.py
Strilanc: You could change it to a lazily initialized module-level constant. Alternatively, pack all these arguments together into a `NamedTuple` type.

cirq/experiments/qubit_characterizations_test.py
```diff
@@ -0,0 +1,105 @@
+import numpy as np
+
+from cirq import GridQubit
+from cirq import circuits, ops, sim
+from cirq.experiments import rabi_oscillations, \
```
Strilanc: Nit: use parens instead of `\`

cirq/experiments/qubit_characterizations_test.py
```diff
@@ -0,0 +1,105 @@
...
+    # small statistical error.
+    simulator = sim.Simulator()
+    qubit = GridQubit(0, 0)
+    results = rabi_oscillations(simulator, qubit, np.pi)
+    data = np.asarray(results.data)
+    angles = data[:, 0]
+    actual_pops = data[:, 1]
+    target_pops = 0.5 - 0.5 * np.cos(angles)
+    rms_err = np.sqrt(np.mean((target_pops - actual_pops) ** 2))
+    assert rms_err < 0.1
```
Strilanc: How flaky is this? Target something safe like failing once per million runs.

cirq/experiments/qubit_characterizations_test.py
```diff
@@ -0,0 +1,105 @@
...
+
+def test_single_qubit_randomized_benchmarking():
+    # Check that the ground state population at the end of the Clifford
+    # sequences is always unity.
+    simulator = sim.Simulator()
+    qubit = GridQubit(0, 0)
+    num_cfds = range(5, 20, 5)
+    results = single_qubit_randomized_benchmarking(simulator, qubit, num_cfds)
+    g_pops = np.asarray(results.data)[:, 1]
+    assert np.mean(g_pops) == 1.0
```
Strilanc: Floating point arithmetic is not exact, use an approximate comparison `np.isclose`.

cirq/experiments/qubit_characterizations_test.py
```diff
@@ -0,0 +1,105 @@
...
+def test_two_qubit_randomized_benchmarking():
+    # Check that the ground state population at the end of the Clifford
+    # sequences is always unity.
+    simulator = sim.Simulator()
+    q_0 = GridQubit(0, 0)
+    q_1 = GridQubit(0, 1)
+    num_cfds = range(5, 20, 5)
+    results = two_qubit_randomized_benchmarking(simulator, q_0, q_1, num_cfds)
+    g_pops = np.asarray(results.data)[:, 1]
+    assert np.mean(g_pops) == 1.0
```
Strilanc: Approximate comparison

examples/qubit_characterizations_example.py
```diff
@@ -0,0 +1,47 @@
...
+    tomography_1q.plot()
+
+    # State-tomography of a Bell state between q_0 and q_1.
+    cir_2q = cirq.Circuit.from_ops(cirq.H(q_0), cirq.CNOT(q_0, q_1))
+    tomography_2q = cirq.experiments.two_qubit_state_tomography(simulator,
+                                                                q_0, q_1,
+                                                                cir_2q)
+    tomography_2q.plot()
+
+    plt.show()
```
Strilanc: Is this redundant now? Please add a screenshot of the output as a reply to this comment.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done. The bar plots for the real&imaginary parts of the density matrix now fit into a single figure. The display looks fine on my computer. Let me know if it looks weird on your computer and I will figure out a better way to do it.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Thanks for the suggestion here. `num_cliffords` is now `num_clifford_range` with a default value. The doc string has been expanded to provide some guideline to the user on how to choose `num_clifford_range` judiciously.

cirq/experiments/qubit_characterizations.py
XiaoMiQC: Done. Same modification applied to the arguments of `two_qubit_randomized_benchmarking`

cirq/experiments/qubit_characterizations_test.py
XiaoMiQC: Done. Now the single-qubit clifford groups c_1, s_1 etc are all stored in a NamedTuple called Clifford.

cirq/experiments/qubit_characterizations_test.py
XiaoMiQC: Done.

cirq/experiments/qubit_characterizations_test.py
XiaoMiQC: Done.

cirq/experiments/qubit_characterizations_test.py
XiaoMiQC: I increased the repetitions to 10000 and am consistently getting rms_error of 0.0035 \pm 0.0005. I think satisfying <0.1 should be fairly safe.

examples/qubit_characterizations_example.py
XiaoMiQC: Without the plt.show() command, all the figures immediately close after opening unless I'm in the interactive mode of PyCharm. I left it for now. The output of the two-qubit tomography function looks like this (both bar plots in the same figure):

Screenshot from 2019-03-14 22-33-02

Strilanc: Wait, you're saying that calling `plt.show()` twice (once inside `tomography_2q.plot()` and once after it) is on purpose? That doesn't make sense to me. Pycharm closes the first one but not the second one? How does it behave when executing outside of pycharm?

XiaoMiQC: Sorry I didn't make this clear. There are actually two kinds of `.show` method, one that goes after a `Figure` object and one that goes after `plt`. The `.show` within the various `plot` methods in the experimental result classes come after the `Figures` object, whereas the `.show` within the example file comes after `plt` itself.

Calling `.show` after a `Figure` object will show the figure but close it soon afterwards so the program can keep running, whereas calling `.show` after 'plt' will show and hold the figure, but halts the execution of the program until the user closes the figure manually. That's why there are two calls to '.show' in the code. I hope this makes sense? A related post on this is here [https://stackoverflow.com/questions/9280171/matplotlib-python-show-returns-immediately](url).

In principle, we can get rid of all the `show` within the plot methods inside the experimental result classes to avoid redundancy. Alternatively we can save the plot to some user-specified directory. Please let me know what you prefer.

Strilanc: I'm not able to reproduce this flickering behavior.

It's acceptable to me if you use either `plt.show` or `fig.show`, as long as you only use one and it occurs inside the method.

Strilanc:
I think the only thing left to do in this PR is to get the continuous integration checks passing.

examples/qubit_characterizations_example.py
XiaoMiQC: Done, I removed the plt.show() from the example.

PR #1332, opened 2019-01-31
Add an abstract method for number of qubits to Gate

dabacon:
* New MultiQubitGate which helps eliminate boilerplate.
* Decided to make it a method, not a property, but could change.
* Lots of changes in contrib/acquaintance, since that is where most multi-qubit gates live.
* Decided not to touch circuit diagram args yet.  There is one place where we use the known_qubits, but none where the count is used.


cirq/contrib/acquaintance/executor_test.py
```diff
@@ -23,21 +23,22 @@
 import cirq.contrib.acquaintance as cca
 
 
-class ExampleGate(cirq.Gate):
+class ExampleGate(cirq.MultiQubitGate):
     def __init__(self, wire_symbols: Sequence[str]) -> None:
+        super(ExampleGate, self).__init__(len(wire_symbols))
```
Strilanc: just `super()`

cirq/ops/common_gates.py
```diff
@@ -334,26 +334,35 @@ def __repr__(self) -> str:
...
-class MeasurementGate(raw_types.Gate):
+class MeasurementGate(gate_features.MultiQubitGate):
     """A gate that measures qubits in the computational basis.
 
     The measurement gate contains a key that is used to identify results
     of measurements.
     """
 
     def __init__(self,
+                 num_qubits: int,
```
Strilanc: This is still requiring the user to redundantly specify information...

dabacon: Yes.  We need to fix this.

dabacon: If we don't we are going to end up making the other protocol choices which means we will specify with_qubits or optional qubits every where and this is way worse than having it in one place.  One interface change  >> multiple changes and added complexity in large number of protocols.

Strilanc: Okay. This is a relatively large breaking change. Tutorials need to be updated to use `measure` etc. I think the changes I want can be added after this PR, instead of continuing to block.

dabacon: You can blame me for this one #blamedave

cirq/contrib/acquaintance/executor_test.py
dabacon: Ah python 3.  Fixed here and throughout.

PR #1333, opened 2019-02-01
Bare bones identity gate 

dabacon:


dabacon:
Not sure why coverage is failing.  Seems to be failing at typing_extensions

dabacon:
Coverage passing now.

PR #1334, opened 2019-02-01
Add mixture protocol

dabacon:


cirq/protocols/mixture.py
```diff
@@ -0,0 +1,83 @@
...
+
+from typing_extensions import Protocol
+
+import numpy as np
+
+from cirq.type_workarounds import NotImplementedType
+
+# This is a special indicator value used by the inverse method to determine
+# whether or not the caller provided a 'default' argument.
+RaiseTypeErrorIfNotProvided = (None,)
```
Strilanc: Use `([],)` instead. In principle an implementation of python could make all expressions `(None,)` point to the same reference, so the caller could make this.

cirq/protocols/mixture.py
```diff
@@ -0,0 +1,83 @@
...
+
+    A mixture is described by an iterable of tuples of the form
+
+        (probability of object, object)
+
+    The probability components of the tuples must sum to 1.0 and be between
+    0 and 1 (inclusive).
+    """
+
+def _mixture_(self) -> Union[Iterable[Tuple[float, Any]], NotImplementedType]:
```
Strilanc: Indentation?

cirq/protocols/mixture.py
```diff
@@ -0,0 +1,83 @@
...
+
+    if getter is None:
+        raise TypeError(
+            "object of type '{}' has no _mixture_ method.".format(type(val)))
+
+    raise TypeError("object of type '{}' does have a _mixture_ method, "
+                    "but it returned NotImplemented.".format(type(val)))
+
+
+def validate_mixture(supports_mixture: SupportsMixture):
```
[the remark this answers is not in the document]
dabacon: I think it will be used by non test code.  I don't want testing to be a dumping ground for anything that smells of validation.

cirq/protocols/mixture.py
dabacon: Just copying your pattern ;)  Changed.

cirq/protocols/mixture.py
dabacon: Auto-indent :)  Changed.

cirq/protocols/mixture.py
dabacon: Ah saw we moved away from the (None,).

dabacon:
Also added doc string.

Strilanc:
The slow march continues :+1:

Strilanc:
You do you, CirqBot. You'll figure it out eventually.

PR #1345, opened 2019-02-04
fix controlled gate circuit diagram info passed to subgate

### LINKED ISSUE #1321: ControlledGate needs to pass modified circuit diagram info args into sub gate
For example, the qubit count reported to the sub gate will be one too large at the moment. This can result in bad diagram info.

balopat:
Fixes #1321.


cirq/ops/controlled_gate_test.py
```diff
@@ -18,7 +18,7 @@
 
 import cirq
 from cirq.type_workarounds import NotImplementedType
-
+from cirq import protocols
```
dabacon: for tests you should try to just use cirq.CircuitDiagramInfoArgs (so that it is like what users write).  So delete this import should make you fix the use of protocols.X below.

balopat: sounds good, thank you!

balopat:
Imports updated!

balopat:
@dabacon @Strilanc PTAL - also it seems the automerger is not working...

PR #1346, opened 2019-02-04
ControlledOperation._apply_unitary_

### LINKED ISSUE #1337: Implement ControlledOperation._apply_unitary_
- Copy over the implementation from ControlledGate and tweak it to work in the new context
- Then rewrite ControlledGate's apply unitary to delegate to ControlledOperation's

balopat:
Closes #1337.


cirq/ops/controlled_test_util.py
```diff
@@ -0,0 +1,54 @@
+from typing import Union, Tuple, cast
+
+import numpy as np
+import cirq
+from cirq.type_workarounds import NotImplementedType
+
+
+class GateUsingWorkspaceForApplyUnitary(cirq.SingleQubitGate):
```
Strilanc: Move these into the main test class adjacent to the test they're used in. I've had bad past experiences with "common test utilities" exploding into a nightmare of interdependency.

cirq/ops/controlled_test_util.py
```diff
@@ -0,0 +1,54 @@
...
+    def _apply_unitary_(self, args: cirq.ApplyUnitaryArgs
+                        ) -> Union[np.ndarray, NotImplementedType]:
+        args.available_buffer[...] = args.target_tensor
+        args.target_tensor[...] = 0
+        return args.available_buffer
+
+    def _unitary_(self):
+        return np.eye(2)
```
Strilanc: raise an exception instead of returning a bad result

cirq/ops/controlled_test_util.py
balopat: These are used in both tests - for controlled op and controlled gate - are we okay with the duplication?

cirq/ops/controlled_test_util.py
balopat: ack

balopat: actually - I'm not sure what you mean exactly here. 1) this was a pre-existing dummy gate class implementation that I reused  2) it does seem to need to return some kind of valid matrix otherwise it fails to be consistent

cirq/ops/controlled_test_util.py
balopat: I created the duplicated version - do you prefer this? or am I missing some pytest specific structure?

cirq/ops/controlled_operation_test.py
```diff
@@ -11,11 +11,63 @@
...
+        one = seed*a + (1, Ellipsis)
+        result = np.zeros(args.target_tensor.shape, args.target_tensor.dtype)
+        result[zero] = args.target_tensor[zero]*2 + args.target_tensor[one]*3
+        result[one] = args.target_tensor[zero]*5 + args.target_tensor[one]*7
+        return result
+
+    def _unitary_(self):
+        return np.array([[2, 3], [5, 7]])
+
+    def __pow__(self, factor):
```
Strilanc: This method is actually needed? It's not doing anything reasonable.

balopat: It is removed

Issue #1341, opened 2019-02-01
RotationErrorChannel is not normalized correctly nor does it make sense.


dabacon:
not sure what was the intent of this channel, but it isn't normalized correctly.  I think maybe the idea was to have a probability of a fixed rotation.  

consider removing it.

dabacon:
This has been removed in #1348

PR #1348, opened 2019-02-04
Add fallback to mixture to channel and moves common channels over to these.

dabacon:
* Also removed rotation channel, which is an invalid channel.  See #1341


dabacon:
Updated to actually run new probability tests.

cirq/ops/common_channels.py
```diff
@@ -676,20 +636,13 @@ def __init__(self, p) -> None:
...
+        self._p = value.validate_probability(p, 'p')
         self._delegate = AsymmetricDepolarizingChannel(1. - p, 0., 0.)
 
-    def _channel_(self) -> Iterable[np.ndarray]:
-        # Return just the I and X pieces.
-        return list(self._delegate._channel_())[:2]
+    def _mixture_(self) -> Sequence[Tuple[float, np.ndarray]]:
+        mixture = protocols.mixture(self._delegate)
+        # just return identity and x term
+        return (mixture[0], mixture[1])
```
Strilanc: This is a sneaky kind of way to couple this class to the implementation of `protocols.mixture`. Are we guaranteeing that it will never e.g. drop 0 terms?

cirq/protocols/channel.py
```diff
@@ -107,6 +107,13 @@ def channel(val: Any,
     if channel_result is not NotImplemented:
         return tuple(channel_result)
 
+    mixture_getter = getattr(val, '_mixture_', None)
+    print(mixture_getter)
```
Strilanc: Aha! You have to debug your code! I knew it!

cirq/ops/common_channels.py
dabacon: Moved over to just calling _mixture_ as it seems fine in this case.  I don't think we will drop 0 terms.

cirq/protocols/channel.py
dabacon: ha

Issue #1124, opened 2018-11-07
Add performance testing


Strilanc:
Given that we've now had our first proper performance bug (`kak_decomposition` being very slow), I think it's time that we added some kind of performance benchmarking process that can be run by the continuous integration scripts.

On another project, the way I did this was by decorating tests with a "performance goal" like this:

```
@perf_goal(avg_micros=50, args=[cirq.random_unitary_matrix()])
def test_kak_decomposition(matrix):
    return cirq.kak_decomposition(matrix)
```

And the test would fail if the code ran slower than the goal (and warn if it ran significantly faster than the goal). One particularly nice aspect of this approach is that the decorator can use the expected time to decide how many repetitions to perform, in order to balance statistical noise against total running time.

Overall, this issue involves:

1. Implementing the `@perf_goal` decorator (or some other mechanism for specifying performance tests)
2. Coming up with a convention for where we put performance tests (e.g. `file_name_perf.py`)
3. Hacking `pytest`'s test discovery to ignore/include performance tests as appropriate, etc.
4. Checking that the performance of code running on Travis-CI is actually stable enough for this to be meaningful (that being said, we don't really need high precision measurements; this is just to put pressure against gradual decay).

A particularly tricky issue here is that performance bugs introduced by PR A may start causing flaky failures in PRs B-Z. It's not clear that we should block merging due to performance bugs, but then it's also not clear how we incentivize actually fixing the issues.

vtomole:
Can we have CirqBot create a new issue for every performance test that fails? Maybe for each pull request that breaks performance tests as there could be too many issues created with the first approach.

Strilanc:
That's a decent idea. Test performance on PRs, but only *really* test it on master commits and when one of the tests starts failing open an issue (or update existing issue).

vtomole:
Assuming that the `assert` statement is in the decorator, If
```
@perf_goal(avg_micros=50, args=[cirq.random_unitary_matrix()])
def test_kak_decomposition(matrix):
    return cirq.kak_decomposition(matrix)
```
fails, pytest will show that it failed in the decorator instead of in `test_kak_decomposition()`. How do you get around that?

vtomole:
Assuming that the `assert` statement is in the decorator, If
```
@perf_goal(avg_micros=50, args=[cirq.random_unitary_matrix()])
def test_kak_decomposition(matrix):
    return cirq.kak_decomposition(matrix)
```
[...]

PR #1349, opened 2019-02-05
pytest-benchmark init

vtomole:
First step of https://github.com/quantumlib/Cirq/issues/1124


Strilanc:
I think a natural addition after this is to print an interpretation of the json to the console whenrunning the tests on travis.

vtomole:
Yup. Also, look my thoughts about comparisons on: https://github.com/quantumlib/Cirq/issues/1124 if you haven't already.

Issue #1285, opened 2019-01-08
Define `_control_` protocol and utility method


Strilanc:
A common thing that people want to do to operations is to control them. For example, this is intrinsic to how phase estimation works. A global utility method `cirq.control` that worked on any gate or OP_TREE would make that easier. Additionally, there would be a `_control_` protocol that gates with more efficient control mechanisms (e.g. decomposing a fredkin using two CNOTs and a Toffoli instead of three Toffolis) could define.

The signature of `_control_` would be `(self, control: cirq.QubitId)` and the result type would vary from object to object (e.g. gates would return gates but operations would return operations).

The signature of the global utility method would be `def control(controllee: Union[Gate, OP_TREE], control_qubit: Optional[Qubit] = None)`. If the control qubit is set to `None`, that means the user wants to specify it later in order to get an operation (i.e. it should work the way `ControlledGate(gate)` works, except it will also apply to operations).

If the control qubit is specified, then the method will attempt to use the controllee's `_control_` method or else fall back to strategies based on the controllee's known unitary or known decomposition in order to return a controlled result. Otherwise it fails.

Strilanc:
Also, once `cirq.control` exists, we can add `controlled_by(control)` to `Operation` if we want.

smitsanghavi:
For _control_ on a gate to return a gate when given the control qubit, do we have a concept of adding only a partial set of qubits to the gate and specifying the rest later?

Strilanc:
- for control(operation, qubit), use ControlledOperation
- for control(operation, None), raise an error explaining a qubit is needed for operations
- for control(gate, None), use ControlledGate
- for control(gate, qubit), make new type SpecifiedControlGate (or alternatively merge this concept into ControlledGate)

PR #1350, opened 2019-02-05
multi control_qubits for ControlledGate, ControlledOperation. Add controlled_by method and control protocol

smitsanghavi:
Support pre-specifying control_qubit for ControlledGate and add __control__ magic methods to Gate and Operation.
#1285


cirq/ops/controlled_gate.py
```diff
@@ -25,30 +25,40 @@
 class ControlledGate(raw_types.Gate):
     """Augments existing gates with a control qubit."""
 
-    def __init__(self, sub_gate: raw_types.Gate) -> None:
+    def __init__(self, sub_gate: raw_types.Gate,
+                 control_qubit: Optional[raw_types.QubitId] = None) -> None:
```
Strilanc: I think that, instead of the class radically changing function based on this argument, that we should have two parameters: `control_qubits` and `num_unspecified_control_qubits`. Then the class can do both things, and it will always be possible to automatically flatten nested controlled gates.

smitsanghavi: Done. Don't know why pytest2 fails.

smitsanghavi: Fixed

cirq/ops/controlled_gate.py
```diff
@@ -26,38 +26,76 @@
 class ControlledGate(raw_types.Gate):
     """Augments existing gates with a control qubit."""
 
-    def __init__(self, sub_gate: raw_types.Gate) -> None:
+    def __init__(self, sub_gate: raw_types.Gate,
+                 control_qubits: List[raw_types.QubitId] = None,
+                 num_unspecified_control_qubits: int = None) -> None:
```
viathor: The interface in #1388 is simpler (just the number of qubits). Can we adopt it here?

cirq/ops/controlled_gate.py
```diff
@@ -68,23 +106,28 @@ def _unitary_(self) -> Union[np.ndarray, NotImplementedType]:
         sub_matrix = protocols.unitary(self.sub_gate, None)
         if sub_matrix is None:
             return NotImplemented
-        return linalg.block_diag(np.eye(sub_matrix.shape[0]), sub_matrix)
+        for _ in range(self.total_control_qubits):
+            sub_matrix = linalg.block_diag(np.eye(sub_matrix.shape[0]),
+                                                  sub_matrix)
```
viathor: nit: confusing indent

cirq/ops/controlled_gate.py
```diff
@@ -68,23 +106,28 @@ def _unitary_(self) -> Union[np.ndarray, NotImplementedType]:
         sub_matrix = protocols.unitary(self.sub_gate, None)
         if sub_matrix is None:
             return NotImplemented
-        return linalg.block_diag(np.eye(sub_matrix.shape[0]), sub_matrix)
+        for _ in range(self.total_control_qubits):
```
viathor: We can very easily compute the size of both blocks, so we probably shouldn't loop.

cirq/ops/controlled_gate.py
```diff
@@ -26,38 +26,76 @@
...
+        return (self.sub_gate, self.total_control_qubits)
 
     def _apply_unitary_(self, args: protocols.ApplyUnitaryArgs) -> np.ndarray:
-        qubits = cirq.LineQubit.range(1 + self.sub_gate.num_qubits())
-        c_op = cop.ControlledOperation(qubits[0],
-                                       self.sub_gate.on(*qubits[1:]))
+        qubits = cirq.LineQubit.range(self.total_control_qubits +
+                                      self.sub_gate.num_qubits())
+        c_op = self.sub_gate.on(*qubits[self.total_control_qubits:])
+        for control in qubits[:self.total_control_qubits]:
```
viathor: This loop can also be avoided if GateOperation is made able to handle multiple controls.

cirq/ops/controlled_gate.py
```diff
@@ -107,11 +150,19 @@ def _circuit_diagram_info_(self,
...
     def __str__(self):
-        return 'C' + str(self.sub_gate)
+        return 'C'*self.total_control_qubits + str(self.sub_gate)
 
     def __repr__(self):
-        return 'cirq.ControlledGate(sub_gate={!r})'.format(self.sub_gate)
+        if (len(self.control_qubits) is 0 and
+            self.num_unspecified_control_qubits is 1):
+            return 'cirq.ControlledGate(sub_gate={!r})'.format(self.sub_gate)
+        else:
```
viathor: nit: You can remove this else and reduce indentation since preceding block ends in a return.

cirq/ops/controlled_gate.py
```diff
@@ -26,38 +26,76 @@
...
-        if len(qubits) < 1:
+        if len(qubits) < self.num_unspecified_control_qubits:
             raise ValueError('No control qubit specified.')
-        self.sub_gate.validate_args(qubits[1:])
+        self.sub_gate.validate_args(
+                          qubits[self.num_unspecified_control_qubits:])
 
     def _value_equality_values_(self):
-        return self.sub_gate
+        return (self.sub_gate, self.total_control_qubits)
```
viathor: nit: unnecessary parentheses

cirq/ops/controlled_gate.py
smitsanghavi: Agreed. Will change after that is implemented.

cirq/ops/controlled_gate.py
smitsanghavi: return linalg.block_diag(np.eye(sub_matrix.shape[0] << (self.total_control_qubits - 1)),] sub_matrix)
caused some tests to fail. Will look into this a bit more

cirq/ops/controlled_gate.py
smitsanghavi: This one specifically adds support for partial set of controls supplied initially. I followed the interface suggested by Craig so will wait for him to weigh in on this one.

Strilanc: After some more thought, I think I agree with @viathor that it's better to just have the number of control qubits. The basic reason comes down to the fact that, with this interface, we have no nice way of controlled the order, of saying "the first control is this one, then the second control is not specified, then the third control qubit is that one". But adding that to the class would be very arduous. Speaking more algebraically, I think the thing that's bothering me is that when the class is nested you can't flatten it.

Also, the way the default behavior of the class interacts with specifying control qubits feels off. Like, the default number of unspecified controls varies depending on whether or not you specified explicit controls.

So I think if we want to add explicit controls to gates, we should do it with a separate class. Cut it down to just the number of controls.

cirq/ops/controlled_gate.py
Strilanc: You should be able to copy over my code from https://github.com/quantumlib/Cirq/pull/1388

cirq/ops/controlled_gate.py
smitsanghavi: I have added logic to automatically flatten nested ControlledGates. The caveat there is that the qubit order for controlling is not preserved but I don't this that the qubit order matters once it's flattened. Even if I separate the partial qubit specification into a separate class, that class will have to deal with the same flattening issue and will probably inherit the same complexity as this.

About the default behavior of num_unspecified_control_qubits, I agree that the current logic is clumsy but I think that should be solvable if I give some more thought to it.

Strilanc: It is extremely important that the qubit order is preserved. If I call `cirq.control(obj, c)`, I expect `c` to end up in a specific consistent place regardless of `obj`.

smitsanghavi: I'm probably not aware of all the nuances here. When I am thinking of automatically flattened nested ControlledGates, I am thinking in case of circuits only. 

The flattened gate is basically n controls &ed together and controlling the sub_gate in the same Moment. In this final picture I see no concept of "qubit order" because all the controls would have been eventually specified to make an op/circuit. If the control qubits have values |q0, q1,..., qn>, regardless of what the qubit order was in the gate, the operation would have the same effect, right?

I don't know who will ever care about the internal qubit order for the gate before it's converted to either an Operation or a Circuit.

cirq/ops/controlled_gate.py
smitsanghavi: Yes, that's what I did for the above comment. It still doesn't work. There's a shape mismatch sometimes.

cirq/ops/controlled_gate.py
smitsanghavi: As per our offline discussion, I have simplified the interface to accept control_qubits and num_controls. Also made the internal list respect the qubit ordering.

cirq/ops/controlled_gate.py
```diff
@@ -68,23 +125,27 @@ def _unitary_(self) -> Union[np.ndarray, NotImplementedType]:
         sub_matrix = protocols.unitary(self.sub_gate, None)
         if sub_matrix is None:
             return NotImplemented
-        return linalg.block_diag(np.eye(sub_matrix.shape[0]), sub_matrix)
+        return linalg.block_diag(
+                    np.eye(sub_matrix.shape[0] << (self.num_controls - 1)),
+                    sub_matrix)
```
smitsanghavi: This is where it goes wrong. PTAL at the failing tests to see what cases break it.

cirq/ops/controlled_gate.py
```diff
@@ -12,52 +12,109 @@
...
 
-from typing import Any, Union
+from typing import Any, List, Union
 
 import numpy as np
 
 import cirq
 from cirq import linalg, protocols, value
-from cirq.ops import raw_types, controlled_operation as cop
+from cirq.ops import (raw_types, controlled_operation as cop, gate_operation)
```
Strilanc: Why the parens?

cirq/ops/controlled_gate.py
```diff
@@ -12,52 +12,109 @@
...
+            control_qubits = []
+        if num_controls < len(control_qubits):
+            raise ValueError('More specified control qubits than num_controls')
+
+        # Leave unspecified controls as Nones.
+        self.control_qubits = ([None]*(num_controls - # type: ignore
+                                       len(control_qubits)) +
+                               control_qubits)
+
+        self.num_controls = num_controls
```
Strilanc: This is redundant w.r.t. the length of the list. Don't store it.

cirq/ops/controlled_gate.py
```diff
@@ -12,52 +12,109 @@
...
+
+        # Flatten nested ControlledGates.
+        if isinstance(sub_gate, ControlledGate):
+            self.sub_gate = sub_gate.sub_gate # type: ignore
+            self.control_qubits += sub_gate.control_qubits
+            self.num_controls += sub_gate.num_controls
+        else:
+            self.sub_gate = sub_gate
+
+        self.num_unspecified_controls = self.control_qubits.count(None)
```
Strilanc: This isn't used often enough to justify caching it. Change it into a private property or just recompute it at each place it's used.

cirq/ops/controlled_gate.py
```diff
@@ -12,52 +12,109 @@
...
+
         if result is NotImplemented:
             return NotImplemented
 
-        return [cop.ControlledOperation(qubits[0], op) for op in result]
+        decomposed = []
+        for op in result:
+            controlled_op = op
+            for qubit in qubits[:self.num_controls]:
+                controlled_op = cop.ControlledOperation(qubit, controlled_op)
```
Strilanc: Is this the correct order? It looks like this will reverse the order of the qubits.

... I guess it doesn't matter in context.

We may want to generalized `ControlledOperation` at the same time, so we don't need to do this.

cirq/ops/controlled_gate.py
```diff
@@ -12,52 +12,109 @@
...
-            raise ValueError('No control qubit specified.')
-        self.sub_gate.validate_args(qubits[1:])
+        if len(qubits) < self.num_unspecified_controls:
+            raise ValueError('Not all control qubits specified.')
+        self.sub_gate.validate_args(qubits[self.num_unspecified_controls:])
+
+    def on(self, *qubits: raw_types.Qid) -> gate_operation.GateOperation:
+        if len(qubits) == 0:
+            raise ValueError(
+                "Applied a gate to an empty set of qubits. Gate: {}".format(
```
Strilanc: Use `{!r}` to get the repr. Pass in `self` instead of `repr(self)`.

cirq/ops/controlled_gate.py
```diff
@@ -12,52 +12,109 @@
...
+                    repr(self)))
+        self.validate_args(qubits)
+
+        # Merge specified controls and new controls.
+        j=0
+        merged_controls = []
+        for control in self.control_qubits:
+            if control is None:
+                merged_controls.append(qubits[j])
```
Strilanc:
```
remaining_qubits = list(quits)
...
...
    merged_controls(remaining_qubits.pop(0))
...
... + remaining_qubits
```

cirq/ops/controlled_gate.py
smitsanghavi: Removed

cirq/ops/controlled_gate.py
smitsanghavi: Removed. Converted to a method.

cirq/ops/controlled_gate.py
smitsanghavi: Removed. Recomputing.

cirq/ops/controlled_gate.py
smitsanghavi: Reversed all such operations for now. Working on adding the multiqubit ControlledOperation.

smitsanghavi: Done

cirq/ops/controlled_gate.py
```diff
@@ -12,52 +12,102 @@
...
 
 
 @value.value_equality
 class ControlledGate(raw_types.Gate):
     """Augments existing gates with a control qubit."""
 
-    def __init__(self, sub_gate: raw_types.Gate) -> None:
+    def __init__(self, sub_gate: raw_types.Gate,
+                 control_qubits: List[raw_types.Qid] = None,
+                 num_controls: int = 1) -> None:
```
Strilanc: Currently, if a user calls this like `ControlledGate(g, control_qubits=[a, b])` they will get a confusing exception about the `num_controls` argument.

Have `num_controls` default to `None`. If both `control_qubits` and `num_controls` are specified (i.e. not `None`), raise an exception about only specifying one of them. If neither is set, default to 1 unspecified control qubit. If just `num_controls` is set, default `control_qubits` to `[None]*num_controls`.

cirq/ops/controlled_gate.py
```diff
@@ -12,52 +12,102 @@
...
+        merged_controls = []
+        remaining_qubits = list(qubits)
+        for control in self.control_qubits:
+            if control is None:
+                merged_controls.append(remaining_qubits.pop(0))
+            else:
+                merged_controls.append(control)
+
+        # Convert to ControlledGate without control qubits that can be reused.
+        self.control_qubits = [None]*self.num_controls()
```
Strilanc: Do not mutate `self`! If the user re-uses this gate instance, their controls will be gone.

cirq/ops/controlled_gate.py
```diff
@@ -12,52 +12,102 @@
...
+
+        return gate_operation.GateOperation(self,
+                                            merged_controls+remaining_qubits)
 
     def _value_equality_values_(self):
-        return self.sub_gate
+        return self.sub_gate, self.num_controls()
```
Strilanc: `self.sub_gate, self.control_qubits`

Make `self.control_qubits` is a tuple, not a list, so that it is hashable.

cirq/ops/controlled_gate.py
```diff
@@ -68,23 +118,27 @@ def _unitary_(self) -> Union[np.ndarray, NotImplementedType]:
...
+                    sub_matrix)
 
     def __pow__(self, exponent: Any) -> 'ControlledGate':
         new_sub_gate = protocols.pow(self.sub_gate,
                                      exponent,
                                      NotImplemented)
         if new_sub_gate is NotImplemented:
             return NotImplemented
-        return ControlledGate(new_sub_gate)
+        return ControlledGate(new_sub_gate, self.control_qubits, # type: ignore
```
Strilanc: Why is the `type: ignore` needed here?

cirq/ops/controlled_gate.py
```diff
@@ -107,11 +161,17 @@ def _circuit_diagram_info_(self,
...
+            wire_symbols=('@',)*self.num_controls() + sub_info.wire_symbols,
             exponent=sub_info.exponent)
 
     def __str__(self):
-        return 'C' + str(self.sub_gate)
+        return 'C'*self.num_controls() + str(self.sub_gate)
 
     def __repr__(self):
-        return 'cirq.ControlledGate(sub_gate={!r})'.format(self.sub_gate)
+        if (self.num_controls() is 1 and self.control_qubits[0] is None):
```
Strilanc: Do not use `is` when comparing integers. I don't think it's guaranteed that every 1 is at the same location in memory. https://stackoverflow.com/questions/306313/is-operator-behaves-unexpectedly-with-integers

cirq/ops/controlled_operation.py
```diff
@@ -11,55 +11,69 @@
...
+                list(controls) + sub_operation.controls,
+                sub_operation.sub_operation)
+        return super().__new__(cls)
+
     def __init__(self,
-                 control: raw_types.Qid,
+                 controls: Sequence[raw_types.Qid],
                  sub_operation: raw_types.Operation):
-        self.control = control
+        self.controls = list(controls)
```
Strilanc: Use a tuple instead of a list.

cirq/ops/controlled_operation.py
```diff
@@ -11,55 +11,69 @@
...
         self.sub_operation = sub_operation
 
     @property
     def qubits(self):
-        return (self.control,) + self.sub_operation.qubits
+        return tuple(self.controls) + self.sub_operation.qubits
```
Strilanc: Just `self.controls + self.sub_operation.qubits` once `self.controls` is a tuple.

cirq/ops/controlled_operation.py
```diff
@@ -11,55 +11,69 @@
...
         result = protocols.decompose_once(self.sub_operation, NotImplemented)
         if result is NotImplemented:
             return NotImplemented
 
-        return [ControlledOperation(self.control, op) for op in result]
+        return [ControlledOperation(self.controls, op) for op in result]
 
     def _value_equality_values_(self):
-        return tuple([self.control, self.sub_operation])
+        return tuple(self.controls), self.sub_operation
```
Strilanc: Just `self.controls, self.suboperation` once `self.controls` is a tuple.

cirq/ops/controlled_operation.py
```diff
@@ -85,21 +94,29 @@ def _unitary_(self) -> Union[np.ndarray, NotImplementedType]:
...
     def __str__(self):
-        return 'C({}){}'.format(self.control, str(self.sub_operation))
-
+        if isinstance(self.sub_operation, gate_operation.GateOperation):
+            return '{}{}({})'.format(
+                'C' * len(self.controls),
+                self.sub_operation.gate,
+                ', '.join(map(str, self.qubits)))
+        return 'C({}, {})'.format(', '.join(str(q) for q in self.controls),
+                                  str(self.sub_operation))
```
Strilanc: Nit: blank line between methods.

cirq/ops/controlled_operation.py
```diff
@@ -134,3 +151,12 @@ def _circuit_diagram_info_(self,
...
             exponent=sub_info.exponent)
+
+def _positions_after_removals_at(initial_positions: Sequence[int],
+                                 removals: Sequence[int]) -> List[int]:
+    # TODO: O(n lg n) instead of O(n**2)
+    result = []
+    for p in initial_positions:
+        change = len([1 for r in removals if r < p])
+        result.append(p - change)
+    return result
```
Strilanc: Nit: new line at end of file.

cirq/ops/raw_types.py
```diff
@@ -160,12 +163,24 @@ def __pow__(self, power):
     def __call__(self, *args, **kwargs):
         return self.on(*args, **kwargs)
 
+    def __control__(self,
```
Strilanc: `_control_` instead of `__control__`. Double-underscore is reserve by python.

cirq/ops/raw_types.py
```diff
@@ -199,6 +214,21 @@ def transform_qubits(self: TSelf_Operation,
         """
         return self.with_qubits(*(func(q) for q in self.qubits))
 
+    def __control__(self,
```
Strilanc: `_control_` instead of `__control__`. Double-underscore is reserve by python.

What do you think of using the name `controlled_by` instead of `_control_`? E.g. `cirq.X(a).controlled_by(b)`

cirq/ops/raw_types.py
```diff
@@ -199,6 +214,21 @@ def transform_qubits(self: TSelf_Operation,
...
+                    control_qubits: List[Qid] = None) -> 'Operation':
```
Strilanc: Take a varargs `*control_qubits: Qid` instead of a list.

cirq/ops/raw_types.py
```diff
@@ -160,12 +163,24 @@ def __pow__(self, power):
     def __call__(self, *args, **kwargs):
         return self.on(*args, **kwargs)
 
+    def __control__(self,
+                    control_qubits: List[Qid] = None) -> 'Gate':
+        """Returns a controlled version of this gate.
+
+        Args:
+            control_qubits: Optional qubits to control the gate by.
```
Strilanc: Take a varargs `*control_qubits: Qid` instead of a list.

cirq/ops/controlled_gate.py
smitsanghavi: Agreed. But instead of always raising an exception when control_qubits and num_controls are specified (i.e. not None) we can handle cases where num_controls>len(control_qubits) by considering the others as yet-to-be-specified

cirq/ops/controlled_gate.py
smitsanghavi: This was my intent. When you specify rest of the qubits and create an operation from a gate that has partial qubits prespecified, you should consume all the qubits together and the gate should become a regular gate that can be reused instead of being tied up to some of the qubits before. I can change this, if needed.

cirq/ops/controlled_gate.py
smitsanghavi: incompatible type "Tuple[None, ...]"; expected "Optional[Sequence[Qid]]"

cirq/ops/controlled_gate.py
smitsanghavi: Changed.

cirq/ops/raw_types.py
smitsanghavi: Renamed to controlled_by

PR #1358, opened 2019-02-07
Validate gate-qubits combination when making a GateOperation

### LINKED ISSUE #1352: GateOperation qubit list is allowed to be inconsistent with gate.
A gate operation is formed by a list of qubits and a gate, but there is no check for consistency between the two. As an example, you can make a GateOperation with a CCX gate which must act on three qubits but only give the GateOperation one qubit.

    example_op = cirq.GateOperation(cirq.CCX, [cirq.GridQubit(1,1)])
    len(example_op.qubits) == example_op.gate.num_qubits()

This second line returns false in this example whereas it should always return true. The qubit list provided should need to pass the gate's validate_args method.

c-poole:
closes #1352


Strilanc:
Have you signed the CLA?

cirq/google/xmon_device_test.py
```diff
@@ -133,12 +133,17 @@ def test_validate_operation_supported_gate():
     d = square_device(3, 3)
 
     class MyGate(cirq.Gate):
-        pass
+        def __init__(self):
+            super().__init__()
```
Strilanc: These two lines are redundant and can be removed.

cirq/google/xmon_device_test.py
```diff
@@ -155,7 +160,7 @@ def test_validate_scheduled_operation_adjacent_exp_11_exp_w():
             cirq.CZ(q1, q2), cirq.Timestamp(), d),
     ])
     d.validate_schedule(s)
-
+
```
Strilanc: Nit: go into pycharm settings and find the option that automatically removes trailing space when you save the file.

cirq/ops/gate_operation.py
```diff
@@ -41,6 +41,7 @@ def __init__(self,
             gate: The gate to apply.
             qubits: The qubits to operate on.
         """
+        gate.validate_args(qubits)
```
Strilanc: Because this is now happening here, we should remove the call that is made during the `Gate.__call__` and `Gate.on` methods.

PR #1360, opened 2019-02-07
Add layer to Simulator interfaces to allow other step by step simulations

dabacon:
This will allow the density matrix and mixture simulators to work with the same interface for the top level interfaces.

Some notes:
* Change in the contract for the simulator iterator and empty circuits.  This eliminates so gnarly handling of this case.
* The type annotations had to be reduce to Iterator from Iterator[StepResult] because Iterator is covariant.  
* This now allows both StepResult and TrialResult for wave function simulations to implement the StateVectorMixin (which has now been made into a proper mixin).
* Mypy doesn't like init methods for mixins.


cirq/__init__.py
```diff
@@ -197,21 +197,26 @@
...
     sample_state_vector,
+    SimulatesSamples,
+    SimulatesFinalState,
+    SimulatesIntermediateState,
+    SimulatesIntermediateWaveFunction,
+    SimulationTrialResult,
+    Simulator,
+    SparseSimulatorStep,
+    StateVectorMixin,
+    StateVectorMixin,
```
Strilanc: Double import

cirq/sim/simulator.py
```diff
@@ -206,17 +212,15 @@ def simulate(
...
-            initial_state: If an int, the state is set to the computational
-                basis state corresponding to this state. Otherwise  if this
-                is a np.ndarray it is the full initial state. In this case it
-                must be the correct size, be normalized (an L2 norm of 1), and
-                be safely castable to an appropriate dtype for the simulator.
+            qubit_order: Determines the canonical ordering of the qubits. This
+                is often used in specifying the initial state, i.e. the
+                ordering of the computational basis states.
+            initial_state: The initial state for the simulation. The  form of
+                this state depends on the simulation implementation.  See
```
Strilanc: Could this state dependence be packaged into a generic parameter on the class? If it causes an explosion of generics everywhere, probably best not to for now.

cirq/sim/simulator_test.py
```diff
@@ -90,39 +96,21 @@ def steps(*args, **kwargs):
     np.testing.assert_equal(result.measurements['b'], [True, False])
     assert set(result.measurements.keys()) == {'a', 'b'}
     assert result.params == param_resolver
-    np.testing.assert_equal(result.final_state, final_state)
+    print(result)
```
Strilanc: print

cirq/sim/simulator_test.py
```diff
@@ -135,170 +123,26 @@ def steps(*args, **kwargs):
                                        qubit_order=qubit_order,
                                        initial_state=2)
     expected_results = [
-        cirq.SimulationTrialResult(
+        cirq.sim.simulator.SimulationTrialResult(
```
Strilanc: Should just be `cirq.SimulationTrialResult`.

cirq/__init__.py
dabacon: fixed

cirq/sim/simulator_test.py
dabacon: removed

cirq/sim/simulator_test.py
dabacon: done and cleaned up some imports

cirq/sim/simulator.py
dabacon: this is probably the worst part of this interface right now.  In particular note the qubit order, which is not actually always necessary for some simulators.   I agree we should try to find a better solution, maybe after we get new simulators in this will become more clear.

dabacon:
Had to add annotation for repr tests to avoid running them in python 2 due to unicode translation.

PR #1364, opened 2019-02-08
Add method to get valid density matrix from possible inputs

dabacon:
Similar to method in wave_function.y

### DIFF
--- a/cirq/sim/density_matrix.py
+++ b/cirq/sim/density_matrix.py
@@ -13,11 +13,10 @@
 # limitations under the License.
 """Code to handle density matrices."""
 
-from typing import Tuple, Type, Union
+from typing import Type, Union
 
 import numpy as np
 
-from cirq import protocols
 from cirq.sim import wave_function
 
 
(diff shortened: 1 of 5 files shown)


cirq/sim/density_matrix.py
```diff
@@ -0,0 +1,73 @@
...
+            raise ValueError('The density matrix is not hermitian.')
+        if not np.isclose(np.trace(density_matrix_rep), 1.0):
+            raise ValueError(
+                'Density matrix did not have trace 1 but instead {}'.format(
+                    np.trace(density_matrix_rep)))
+        if density_matrix_rep.dtype != dtype:
+            raise ValueError(
+                'Density matrix had dtype {} but expected {}'.format(
+                    density_matrix_rep.dtype, dtype))
+        if not np.all(np.linalg.eigvalsh(density_matrix_rep) > -1e-8):
```
Strilanc: Isn't this an extremely expensive check?

dabacon: It is.  But it is essentially same order as all the other density matrix simulation steps O((2^n)^3).  (In the same wave the normalization check for the sparse simulator is O(2^n) like most sparse steps.  I'll keep it for now, but lets discuss Monday.

PR #1366, opened 2019-02-11
Adds measure and sample for density matrices.

dabacon:


cirq/sim/density_matrix.py
```diff
@@ -71,3 +72,180 @@ def to_valid_density_matrix(
...
+    # Shape into a tensor
+    tensor = np.reshape(all_probs, [2] * num_qubits)
+
+    # Calculate the probabilities for measuring the particular results.
+    probs = [
+        np.sum(np.abs(tensor[linalg.slice_for_qubits_equal_to(indices, b)]))
+        for b in range(2 ** len(indices))]
+
+    # To deal with rounding issues, ensure that the probabilities sum to 1.
+    probs /= sum(probs) # type: ignore
```
Strilanc: `np.sum` will be faster

cirq/sim/density_matrix.py
```diff
@@ -71,3 +72,180 @@ def to_valid_density_matrix(
...
+    return probs
+
+
+def _validate_num_qubits(matrix: np.ndarray) -> int:
+    """Validates that matrix's shape is a valid shape for qubits.
+    """
+    shape = matrix.shape
+    half_index = len(shape) // 2
```
Strilanc: Did you check that the length is even anywhere?

cirq/sim/density_matrix.py
```diff
@@ -71,3 +72,180 @@ def to_valid_density_matrix(
...
+    row_size = np.prod(shape[:half_index]) if shape != (0,) else 0
```
Strilanc: Why would the shape ever be `(0,)`? It should be `()` instead in the no-qubit case.

cirq/sim/density_matrix.py
```diff
@@ -71,3 +72,180 @@ def to_valid_density_matrix(
...
+    col_size = np.prod(shape[half_index:]) if shape != (0,) else 0
+    if row_size != col_size:
+        raise ValueError(
+            'Matrix was not square. Shape was {}'.format(shape))
+    if row_size !=0 and row_size & (row_size - 1):
```
Strilanc: `row_size & (row_size - 1)` being true implies `row_size != 0`.

cirq/sim/density_matrix.py
```diff
@@ -71,3 +72,180 @@ def to_valid_density_matrix(
...
+        ValueError if the dimension of the matrix is not compatible with a
+            matrix of n qubits.
+        IndexError if the indices are out of range for the number of qubits
+            corresponding to the density matrix.
+    """
+    num_qubits = _validate_num_qubits(matrix)
+    _validate_indices(num_qubits, indices)
+
+    if len(indices) == 0:
+        return ([], np.copy(matrix))
```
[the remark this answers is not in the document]
dabacon: Added test.

cirq/sim/density_matrix.py
dabacon: done

dabacon:
Ping on this one.

PR #1367, opened 2019-02-11
Add has_channel and has_mixture methods.

dabacon:
* Implements them in common_channels.
* Mixture doesn't fallback like channel.  This is because mixture is a more general concept right now, and I think we want to use it for a fast density matrix object.  This just means the probabilistic simulator will have to do a little more work. Alternatively when we get there we could implement a has_mixture_channel that does the fallback for the unitary.


cirq/protocols/channel.py
```diff
@@ -66,10 +66,23 @@ def _channel_(self) -> Union[Iterable[np.ndarray], NotImplementedType]:
...
+
+        This method is used by the global `cirq.has_channel` method.  If this
+        method is not present, or returns NotImplemented, it will fallback
+        to similarly checking `cirq.has_mixture` or `cirq.has_unitary`. If none
+        of these are present or return NotImplemented, then `cirq.has_channel`
+        will fall back to checking whether `cirq.channel` has a non-default
+        value. Otherwise `cirq.has_channel` returns False.
+
+        Returns:
+          True if the value has a channel representation, False otherwise.
```
Strilanc: Indentation

cirq/protocols/channel.py
```diff
@@ -130,3 +143,38 @@ def channel(val: Any,
...
+        has a non-default value. Returns False if none of these functions
+        exists.
+    """
+    channel_getter = getattr(val, '_has_channel_', None)
+    result = NotImplemented if channel_getter is None else channel_getter()
+
+    if result is not NotImplemented:
+        return result
+
+    mixture_getter = getattr(val, '_has_mixture_', None)
```
Strilanc: Delegate to `cirq.has_mixture` instead of respecifying logic.

cirq/protocols/mixture.py
```diff
@@ -79,6 +95,24 @@ def mixture(
...
+
+    Returns:
+        If `val` has a `_has_mixture_` method and its result is not
+        NotImplemented, that result is returned. Otherwise, if the value
+        has a `_mixture_` method return if that has a non-default value.
+        Returns False if neither function exists.
+    """
+    getter = getattr(val, '_has_mixture_', None)
+    result = NotImplemented if getter is None else getter()
+    if result is not NotImplemented:
```
Strilanc: This should fallback to `has_unitary`, right? If a gate has a unitary that's a special case of a mixture.

cirq/protocols/channel.py
```diff
@@ -130,3 +143,38 @@ def channel(val: Any,
...
+    result = NotImplemented if mixture_getter is None else mixture_getter()
+
+    if result is not NotImplemented:
```
Strilanc: I think we should also check for `None`. This has been leaking into some of the other protocols; users are more familiar with it I think. It makes sense when the None result would clearly be invalid.

dabacon:
I think you missed my comment about using mixture more broadly?

I think maybe the solution is that has_mixture and mixture don't fall back, but has_mixture_channel and mixture_channel do fall back.  How does that sound?

dabacon:
That couples "mixture" to "unitary".  Suppose you want to think of a density matrix as a probabilistic combination of pure states.  Then why would it fall back to "unitary"?

PR #1369, opened 2019-02-11
Optimize single qubit optimization

dstrain115:
- Rather than use the complicated frontier
function, calculate the affected opeerations directly.
- Also, optimize the operate_on methods in moment.py

Results: Improves merge_single_qubit_gates_into_phased_x_z
by about 40-50%


dstrain115:
Are you sure about that sentiment?  I could move it into Circuit if you want, but my feeling was this was pretty specialized functionality, so I am not sure how reusable it is.  The main benefit is that we are finding the operations and indices at the same time as the frontier (simplifying the algorithm for one qubit also helps though).

If we were to add this to circuit, I think the better way would be to augment findall_operations rather than reachable_frontier.  Maybe this interface?

findall_operations_until_blocked(self,
                                   start_frontier: Dict[ops.QubitId, int],
                                   is_blocker: Callable[[ops.Operation], bool] = lambda op: False,
                                   omit_crossing_operations: bool = False
                                   ) -> List[Tuple[int, ops.Operation]]:

Which would start at the frontier and proceed until a blocking operation is hit, returning all intervening non-blocking operations with indices.  I'd have to think a bit about how to generalize this to multiple qubits and a varied frontier.

Does this sound good?

Strilanc:
Add specialized version of findall_operations, private, auto fallback

dstrain115:
I have finally implemented the circuit findall_operations_until_blocked as discussed above.  Let me know what you think.

cirq/circuits/circuit.py
```diff
@@ -659,6 +660,50 @@ def findall_operations_between(self,
 
         return list(result)
 
+    def findall_operations_until_blocked(self,
+                                   start_frontier: Dict[ops.QubitId, int],
```
Strilanc: Nit: indentation

cirq/circuits/circuit.py
```diff
@@ -659,6 +660,50 @@ def findall_operations_between(self,
...
+                                         = lambda op: False,
+                                   omit_crossing_operations: bool = False
+                                   ) -> List[Tuple[int, ops.Operation]]:
+        """
+        Finds all operations until a blocking operation is hit.  For each
+        qubit, this returns a list of all operations that involve that qubit
+        from the starting moment until a blocking operation that involves that
+        qubit.  Operations not involving that qubit are not considered blocking
+        for that qubit, even if they affect other qubits listed in the
+        start_frontier dictionary.  See reachable_frontier_from for a more
```
Strilanc: Use back ticks when referring to code elements such as `reachable_frontier_from`. They get used by the documentation system.

cirq/circuits/circuit.py
```diff
@@ -659,6 +660,50 @@ def findall_operations_between(self,
...
+        in depth example of reachable states.
+
+        Args:
+            start_frontier: A starting set of reachable locations.
+            is_blocker: A predicate that determines if operations block
+                reachability. Any location covered by an operation that causes
+                `is_blocker` to return True is considered to be an unreachable
+                location.
```
Strilanc: Doc `omit_crossing_operations`

cirq/circuits/circuit.py
```diff
@@ -1636,6 +1681,7 @@ def _draw_moment_in_diagram(
                 for index in indices:
                     out_diagram.write(x, index, '^' + exponent)
 
+
```
Strilanc: Nit: unnecessary line

cirq/circuits/circuit_test.py
```diff
@@ -1043,6 +1043,78 @@ def test_findall_operations_with_gate():
...
+        cirq.H(a))
+
+    # Always return true to test basic features
+    go_to_end = lambda op : False
+    stop_if_op = lambda op : True
+    stop_if_h = lambda op : op.gate == cirq.H
+
+    # Empty cases.
+    assert cirq.Circuit().findall_operations_until_blocked(
+        start_frontier={}, is_blocker=go_to_end) ==[]
```
Strilanc: Nit: space after `==`

cirq/circuits/circuit.py
```diff
@@ -659,6 +660,50 @@ def findall_operations_between(self,
...
 
+    def findall_operations_until_blocked(self,
+                                   start_frontier: Dict[ops.QubitId, int],
+                                   is_blocker: Callable[[ops.Operation], bool]
+                                         = lambda op: False,
+                                   omit_crossing_operations: bool = False
+                                   ) -> List[Tuple[int, ops.Operation]]:
+        """
+        Finds all operations until a blocking operation is hit.  For each
+        qubit, this returns a list of all operations that involve that qubit
```
Strilanc: This sentence makes it sound like the return type should be `Dict[Qid, List[...]]`. Rephrase to make it clearer how the lists are aggregated.

Strilanc:
Need to resolve conflicts

Strilanc:
Need to replace instances of QubitId with Qid.

dstrain115:
Merged in the Qid changes.  Want to take one last look that I merged correctly and then commit?

Issue #1357, opened 2019-02-07
Channels should support keys for labeling outcomes during simulations


dabacon:
there should be a way to get results out during simulation

dabacon:
Adding _measurement_key_ protocol.

dabacon:
There is a concrete proposal to move to a protocol `_acts_on_` and this would be fixed as part of this.

95-martin-orion:
Absorbing into #3566 to capture more recent discussion.

95-martin-orion:
Duplicate of #3566

PR #1374, opened 2019-02-14
Density matrix simulator

dabacon:
A working implementation of density matrix simulation.  Follow up work that is necessary after this pull request:
* Doesn't use apply_unitary or any similar methods (need apply_channel)
* Doesn't optimize for measurements at end of the gate.


cirq/sim/density_matrix_simulator.py
```diff
@@ -0,0 +1,425 @@
...
+class DensityMatrixSimulator(simulator.SimulatesSamples,
+                             simulator.SimulatesIntermediateState):
+    """A simulator for density matrices and noisy quantum circuits.
+
+    This simulator can be applied on circuits that are made up of operations
+    that have:
+        * a `_channel_` method
+        * a `_mixture_` method for a probabilistic combination of unitary gates.
+        * a `_unitary_` method
+        * a `_has_unitary_` and `_apply_unitary_` method.
```
Strilanc: Or decompose.

cirq/sim/density_matrix_simulator.py
```diff
@@ -0,0 +1,425 @@
...
+    results, but also the parameters used for the parameterized
+    circuit operations. The initial state of a run is always the all 0s state
+    in the computational basis.
+
+    By contrast the simulate methods of the simulator give access to the density
+    matrix of the simulation at the end of the simulation of the circuit.
+    These methods take in two parameters that the run methods do not: a
+    qubit order and an initial state. The qubit order is necessary because an
+    ordering must be chosen for the kronecker product (see
+    `DensityMatrixTrialResult` for details of this ordering). The initial
```
Strilanc: Does the density matrix include from simulate include measurement results, does it act like the measurement results are unknown, or does it require that all measurements be terminal?

cirq/sim/density_matrix_simulator.py
```diff
@@ -0,0 +1,425 @@
...
+        matrix = density_matrix_utils.to_valid_density_matrix(
+            initial_state, num_qubits, self._dtype)
+        if len(circuit) == 0:
+            yield DensityMatrixStepResult(matrix, {}, qubit_map, self._dtype)
+
+        def on_stuck(bad_op: ops.Operation):
+            return TypeError(
+                "Can't simulate operations that don't implement "
+                "SupportsUnitary, SupportsApplyUnitary, SupportsMixture, "
+                "SupportsChannel or is a measurement: {!r}".format(bad_op))
```
Strilanc: Future: we may want some way to include the concept of measurement into our protocols, so people can define e.g. measurements by different observables.

cirq/sim/density_matrix_simulator.py
```diff
@@ -0,0 +1,425 @@
...
+        final_simulator_state: 'DensityMatrixSimulatorState') \
+        -> 'DensityMatrixTrialResult':
+        return DensityMatrixTrialResult(
+            params=params,
+            measurements=measurements,
+            final_simulator_state=final_simulator_state)
+
+
+class DensityMatrixStepResult(simulator.StepResult):
+    """A singe step in the simulation of the DensityMatrixSimulator.
```
Strilanc: single

cirq/sim/density_matrix_simulator.py
```diff
@@ -0,0 +1,425 @@
...
+            `numpy.complex64` or `numpy.complex128`
+        """
+        if dtype not in {np.complex64, np.complex128}:
+            raise ValueError(
+                'dtype must be complex64 or complex128, was {}'.format(dtype))
+
+        self._dtype = dtype
+
+    def _run(self,
+        circuit: circuits.Circuit,
```
Strilanc: Your editor really seems to like indenting param lists 4 spaces instead of 8 spaces. Makes them visually harder to distinguish from the statements in the body.

cirq/sim/density_matrix_simulator.py
dabacon: Yep, this is #1357 I think.

PR #1376, opened 2019-02-14
Cleanup in linalg package

### LINKED ISSUE #1370: Use math.gcd in eigengate?
My pytest in verbose mode is dominated by a trillion of these warnings.  Was there a reason we were not using math.gcd()?  (3to2 problem?)
```
/tmp/test-cirq-yd_i4r4g/cirq/ops/eigen_gate.py:325: DeprecationWarning: fractions.gcd() is deprecated. Use math.gcd() instead.
  t = t * r // fractions.gcd(t, r)
```

Strilanc:
- Added check_preconditions parameters, to disable expensive checks
- kak_decompose now disables precondition checking
- Removed a postcondition check in _perp_eigendecompose
- Forced keyword arguments for atol/rtol
- Removed rtol parameters that had no effect
- Slightly optimized "all near zero" utility methods
- Worked around fractions.gcd deprecation warnings

Fixes #1370


cirq/linalg/diagonalize.py
```diff
@@ -165,14 +179,15 @@ def bidiagonalize_real_matrix_pair_with_symmetric_products(
...
-    if np.any(np.imag(mat1) != 0):
-        raise ValueError('mat1 must be real.')
-    if np.any(np.imag(mat2) != 0):
-        raise ValueError('mat2 must be real.')
-    if not predicates.is_hermitian(mat1.dot(mat2.T), rtol=rtol, atol=atol):
-        raise ValueError('mat1 @ mat2.T must be symmetric.')
-    if not predicates.is_hermitian(mat1.T.dot(mat2), rtol=rtol, atol=atol):
-        raise ValueError('mat1.T @ mat2 must be symmetric.')
+    if check_preconditions:
+        if np.any(np.imag(mat1) != 0):
```
dabacon: is this safe from floating errors?

Strilanc: The caller is supposed to send in a real matrix, not an approximately real matrix. In the actual context where this is used, that's done by sending in e.g. `np.real(u)` and `np.imag(u)` as the arguments.

PR #1377, opened 2019-02-14
Add support for qubit keys in TrialResult.histogram

Strilanc:


cirq/study/trial_result.py
```diff
@@ -75,6 +82,14 @@ def _keyed_repeated_bitstrings(vals: Dict[str, np.ndarray]
     return '\n'.join(keyed_bitstrings)
 
 
+def _key_to_str(key: TMeasurementKey) -> str:
+    if isinstance(key, str):
+        return key
+    if isinstance(key, ops.QubitId):
```
dabacon: maybe: you could make this isinstance(key, (ops.QubitId, str)) and get rid of first one.

Strilanc: I originally did that, but 3to2 fails to change `str` into `unicode` in that case and you get a test failure.

PR #1379, opened 2019-02-15
[Fixes #1370] uses math.gcd on python3

### LINKED ISSUE #1370: Use math.gcd in eigengate?

bryano:
while keeping python2 compatibility


Strilanc:
This is being fixed by happenstance in https://github.com/quantumlib/Cirq/pull/1376 by introducing a common `_compat.py` file. Also it deals with the fact that the two gcd methods disagree on the sign of the result. I should have assigned myself the issue.

Issue #1356, opened 2019-02-07
Measurement should implement channel


dabacon:
Because it is a channel.

dabacon:
Proposal:

If a gate implements channel and measurement_key, then when simulated it will record results in measurements and outputs will be conditional on this.  If it only implements channel, then the measurement results are not recorded, and outputs will need to be at least density matrices but recording that the information was not recorded.

Strilanc:
If a measurement is not recorded, and you are performing a density simulation, do we randomly pick one of the outcome branches (project to one of the poles) or do we keep going with a mixture of both (project to the Z axis)?

Strilanc:
@dabacon verify if this is done

Strilanc:
@dabacon says done

PR #1384, opened 2019-02-15
Add measurement_key protocol.

dabacon:
Also adds channel to measurement.  

See #1356 and #1357

### DIFF
--- a/cirq/protocols/measurement_key.py
+++ b/cirq/protocols/measurement_key.py
@@ -1,0 +1,84 @@
+# Copyright 2019 The Cirq Developers
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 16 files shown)


cirq/protocols/measurement_key.py
```diff
@@ -0,0 +1,84 @@
...
+    def _measurement_key_(self) -> str:
+        """Return the key that will be used to identify this measurement.
+
+        When a measurement occurs, either on hardware, or in a simulation,
+        this is the key value under which the results of the measurement
+        will be stored.
+        """
+
+
+def measurement_key(
```
Strilanc: What do you think about slightly generalizing this so that it works as a way of getting the key for measurement operations, but also as a way to get the canonical key for e.g. a list of qubits?

dabacon: I think lets start with "measurement key" but it could be that we want the general method at some point.

PR #1385, opened 2019-02-16
Increase speed of some tests

dabacon:


Strilanc:
It restricts the gate set, and it's not using the magical `np.einsum`.

vtomole:
@Strilanc 
> Magical `np.einsum` 

Heh.
Regarding slow tests/builds, they are going to get longer as the code base gets larger. I know we have a script that just tests changed files, but could we have a kind of "incremental" test procedure that runs in the background in the local dev environment? Then when the user pushes, Travis can use that info to be faster? Something along those lines... like an incremental compiler.

Strilanc:
That already exists, it's called `check/pytest-changed-files`

PR #1390, opened 2019-02-22
Pauli expansion protocol

viathor:
Logic for expanding an operator in the Pauli basis extracted from #1299 into a protocol.


cirq/linalg/operator_spaces.py
```diff
@@ -0,0 +1,69 @@
...
+
+PAULI_BASIS = {
+    'I': np.eye(2),
+    'X': np.array([[0., 1.], [1., 0.]]),
+    'Y': np.array([[0., -1j], [1j, 0.]]),
+    'Z': np.diag([1., -1]),
+}
+
+
+def kron_bases(basis1: Dict[str, np.ndarray],
```
Strilanc: I think this should have the same signature as `itertools.product`, e.g. you can specify an arbitrary number of bases and also there's a `repeat` parameter.

cirq/linalg/operator_spaces.py
```diff
@@ -0,0 +1,69 @@
...
+               basis2: Dict[str, np.ndarray]) -> Dict[str, np.ndarray]:
+    """Creates (n+k)-qubit basis from n-qubit basis1 and k-qubit basis2."""
+    return {
+        name1 + name2: np.kron(matrix1, matrix2)
+        for name1, matrix1 in basis1.items()
+        for name2, matrix2 in basis2.items()
+    }
+
+
+def hilbert_schmidt(m1: np.ndarray, m2: np.ndarray) -> complex:
```
Strilanc: Rename to `hilbert_schmidt_inner_product`.

cirq/linalg/operator_spaces.py
```diff
@@ -0,0 +1,69 @@
...
+    """Computes Hilbert-Schmidt inner product of two matrices.
+
+    Linear in second argument.
+    """
+    m1_dagger = np.conjugate(np.transpose(m1))
+    return np.trace(np.dot(m1_dagger, m2))
```
Strilanc: It should be possible to reduce this from cubic time to quadratic time by using `np.einsum` to refer specifically to the diagonal.

cirq/linalg/operator_spaces.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def expand_in_basis(m: np.ndarray,
```
Strilanc: Put the orthogonality assumption into the name, because it is so crucial. Also note that it's about matrices instead of e.g. vectors.

`expand_matrix_in_orthogonal_basis`

cirq/linalg/operator_spaces.py
```diff
@@ -0,0 +1,69 @@
...
+                    basis: Dict[str, np.ndarray]) -> Dict[str, complex]:
```
Strilanc: Does the key *have* to be a string? Could be `Any`, or perhaps even a list.

cirq/linalg/operator_spaces.py
```diff
@@ -0,0 +1,69 @@
...
+    product. We do not require that basis be orthonormal. Note that Pauli
+    basis (I, X, Y, Z) is orthogonal, but not orthonormal.
+    """
+    return {
+        name: hilbert_schmidt(b, m) / hilbert_schmidt(b, b)
+        for name, b in basis.items()
+    }
+
+
+def reconstruct_from_expansion(expansion: Dict[str, complex],
```
Strilanc: This name is not clear out of context. Perhaps `matrix_from_basis_coefficients`?

cirq/linalg/operator_spaces_test.py
```diff
@@ -0,0 +1,204 @@
...
+                        [0, 0, 0, 1j],
+                        [1j, 0, 0, 0],
+                        [0, -1j, 0, 0]]),
+        'ZI': np.diag([1, 1, -1, -1]),
+        'ZX': scipy.linalg.block_diag(X, -X),
+        'ZY': scipy.linalg.block_diag(Y, -Y),
+        'ZZ': np.diag([1, -1, -1, 1]),
+    }),
+    (STANDARD_BASIS, STANDARD_BASIS, {
+        'aa': _one_hot_matrix(4, 0, 0),
```
Strilanc: Could be defined with a generator.

```python
    'abcd'[i] + 'abcd'[j]: _one_hot_matrix(4, i, j)
    for i in range(4)
    for j in range(4)
```

cirq/ops/common_gates.py
```diff
@@ -263,6 +283,16 @@ def _eigen_components(self):
...
         ]
 
+    def _pauli_expansion_(self) -> Dict[str, complex]:
+        if protocols.is_parameterized(self):
+            return NotImplemented
+        phase = 1j**(2 * self._exponent * (self._global_shift + 0.5))
+        angle = np.pi * self._exponent / 2
+        return {
+            'I': phase * np.cos(angle),
+            'Z': -1j * phase * np.sin(angle),
```
Strilanc: How crazy would it be to use `cirq.Z` instead of the string `'Z'`...?

viathor: Not too crazy IMO. I considered it and deferred it until we have tensor product since this is needed to create keys for multi-qubit expansions (e.g. II, XX, YY and ZZ for swap gates).

cirq/linalg/operator_spaces.py
viathor: It doesn't have to be a string, but it should be a hashable, so `Any` is too broad and lists are excluded.

I tried using `Hashable` in the type annotation, but mypy isn't currently able to correctly decide which types are hashable. In particular, it doesn't recognize `str` as such (I think we want `str` even if we may also want other key types), see https://github.com/python/mypy/issues/1746.

Given the setbacks I'd keep this as-is until we find compelling reasons to change this. WDYT?

cirq/linalg/operator_spaces.py
viathor: Good idea! Done and added tests to cover repeats.

cirq/linalg/operator_spaces.py
viathor: Done.

cirq/linalg/operator_spaces.py
viathor: Oops. Fixed.

cirq/linalg/operator_spaces.py
viathor: Done.

cirq/linalg/operator_spaces.py
viathor: Done. BTW: Do you think we should expose these names directly under `cirq.` package? They're currently under `cirq.linalg.`.

cirq/linalg/operator_spaces_test.py
viathor: Done.

viathor:
This is ready for re-review. Coverage fails because protocol consistency checks have been disabled for parametrized code. They should pass once #1399 is merged here.

cirq/linalg/operator_spaces.py
Strilanc: I consider linalg to be one of the "flattened packages", so it should be exposed globally.

cirq/linalg/operator_spaces.py
Strilanc: SGTM

Strilanc:
(Blocked on https://github.com/quantumlib/Cirq/pull/1396)

PR #1391, opened 2019-02-22
ParallelGateOperation

### LINKED ISSUE #1352: GateOperation qubit list is allowed to be inconsistent with gate.

c-poole:
Implements ParallelGateOperation, which acts as a single operation where the same single qubit gate is applied to several qubits. Also closes #1352.


cirq/ops/parallel_gate_operation.py
```diff
@@ -0,0 +1,233 @@
...
+        diagram_info = protocols.circuit_diagram_info(self.gate,
+                                                      args,
+                                                      NotImplemented)
+        if diagram_info == NotImplemented:
+            return diagram_info
+
+        # Extract the formatted exponent and place it in wire symbols so that
+        # every gate on the diagram has an exponent instead of the last one
+        symbol = diagram_info.wire_symbols[0]
+        exponent = self._formatted_exponent(diagram_info, args)
```
Strilanc: Ah I see, you're trying to put the exponent on every one of the gates.

Instead of duplicating so much logic between here and `_circuit.py`, we should move this utility code into `CircuitDiagramInfo`. For example, it could have a `rendered_symbols` method.

cirq/ops/parallel_gate_operation.py
```diff
@@ -0,0 +1,233 @@
...
+    def _phase_by_(self, phase_turns: float,
+                   qubit_index: int) -> 'ParallelGateOperation':
+        phased_gate = protocols.phase_by(self._gate, phase_turns, qubit_index,
+                                         default=None)
+        if phased_gate is None:
+            return NotImplemented
+        return self.with_gate(phased_gate)
+
+    def _trace_distance_bound_(self) -> float:
+        return 1.
```
Strilanc: Delete this method if the bound is no better than the default bound.

cirq/ops/parallel_gate_operation.py
```diff
@@ -0,0 +1,233 @@
...
+            A new operation on the same qubits with the scaled gate.
+        """
+        new_gate = protocols.pow(self.gate,
+                                 exponent,
+                                 NotImplemented)
+        if new_gate is NotImplemented:
+            return NotImplemented
+        return self.with_gate(new_gate)
+
+    def _qasm_(self, args: protocols.QasmArgs) -> Optional[str]:
```
Strilanc: I believe this will work better if you don't implement `_qasm_` at all, because the qasm output code will auto decompose the gate and helpfully comment that it came from a parallel operation. Test it out.

cirq/ops/parallel_gate_operation.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def __init__(self,
+                 gate: raw_types.Gate,
+                 qubits: Sequence[raw_types.QubitId]) -> None:
+        """
+        Args:
+            gate: the gate to apply
+            qubits: lists of lists of qubits to apply the gate to.
+        """
+        if not isinstance(gate, gate_features.SingleQubitGate):
```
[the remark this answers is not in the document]
c-poole: Agreed, will do.

cirq/ops/parallel_gate_operation.py
```diff
@@ -0,0 +1,233 @@
...
+            raise ValueError("gate must be a single qubit gate")
+        if len(frozenset(qubits)) != len(qubits):
```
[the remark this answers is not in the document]
c-poole: Agreed, will do.

cirq/ops/parallel_gate_operation.py
```diff
@@ -0,0 +1,233 @@
...
+    def __repr__(self):
+        return 'cirq.ParallelGateOperation(gate={!r}, qubits={!r})'.format(
+            self.gate,
+            list(self.qubits))
+
+    def __str__(self):
+        return '{}({})'.format(self.gate,
+                               ', '.join(str(e) for e in self.qubits))
+
+    def _group_interchangeable_qubits(self) -> Tuple[
```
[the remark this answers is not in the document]
c-poole: Fair, will delete.

cirq/ops/parallel_gate_operation.py
```diff
@@ -0,0 +1,233 @@
...
+            Union[raw_types.QubitId,
+                  Tuple[int, FrozenSet[raw_types.QubitId]]],
+            ...]:
+        return tuple(((0, frozenset(self.qubits)),))
+
+    def _value_equality_values_(self):
+        return self.gate, self._group_interchangeable_qubits()
```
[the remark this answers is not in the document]
c-poole: Agreed.

cirq/ops/parallel_gate_operation.py
```diff
@@ -0,0 +1,233 @@
...
+        for axis in args.axes:
+            result = protocols.apply_unitary(self.gate,
+                                             protocols.ApplyUnitaryArgs(
+                                                 state,
+                                                 args.available_buffer,
+                                                 (axis,)),
+                                             default=NotImplemented)
+
+            if result is args.available_buffer:
+                args.available_buffer = state
```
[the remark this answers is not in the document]
c-poole: Agreed.

cirq/ops/parallel_gate_operation.py
c-poole: Agreed.

cirq/ops/parallel_gate_operation.py
c-poole: Removing the does provide qasm code with a neat comment specifying the the code block below it corresponds to the same operation. I like the aesthetics of that more, so I'll remove this implementation.

cirq/ops/parallel_gate_operation.py
c-poole: This exponent treatment was only necessary due to the fact that the `_draw_moment_in_diagram` method in `circuit.py` tacks on the exponent to the last symbol in the operation, so it seems like adjusting this behavior is the most straightforward approach. It feels natural to me that if the `connected` keyword argument of a `CircuitDiagramInfo` object is set to `False`, then the exponent should go on every symbol in the operation. An additional explicit keyword argument could be added to `CircuitDiagramInfo` instead. This circumvents the need for the exponent hack here. Does this sound acceptable?

Strilanc: That change sounds good to me.

cirq/ops/parallel_gate_operation_test.py
```diff
@@ -0,0 +1,176 @@
...
+def test_equivalent_circuit():
+    depth = 6
+    qreg = cirq.LineQubit.range(4)
+    oldc = cirq.Circuit()
+    newc = cirq.Circuit()
+    exponents = [0, -1, 1, 1/2, 1/3, 1/3+0.01]
+
+    for moment in range(depth):
+        if moment == 0:
+            gate = cirq.XPowGate(exponent=exponents[moment])
```
Strilanc: Make a list of gates and `zip` it with the exponents, instead of making a big if-else chain to manually index into an implicit list.

cirq/ops/parallel_gate_operation_test.py
```diff
@@ -0,0 +1,176 @@
...
+            gate = cirq.YPowGate(exponent=exponents[moment])
+        if moment == 5:
+            gate = cirq.ZPowGate(exponent=exponents[moment])
+
+        for qubit in qreg:
+            oldc.append(gate.on(qubit))
+
+        newc.append(cirq.ops.ParallelGateOperation(gate, qreg))
+
+    assert str(oldc) == str(newc)
```
Strilanc: Use `cirq.testing.assert_has_diagram(newc, oldc.to_text_diagram())`

cirq/ops/parallel_gate_operation_test.py
```diff
@@ -0,0 +1,176 @@
...
+    sim = cirq.Simulator()
+    old_result = sim.simulate(oldc)
+    new_result = sim.simulate(newc)
+    assert old_result.dirac_notation() == new_result.dirac_notation()
```
Strilanc: Use `cirq.testing.assert_circuits_with_terminal_measurements_are_equivalent`

(your circuit is vacuously one with only terminal measurements)

cirq/ops/parallel_gate_operation_test.py
```diff
@@ -0,0 +1,176 @@
...
+    eq.make_equality_group(lambda: cirq.ParallelGateOperation(g1, r1))
+    eq.make_equality_group(lambda: cirq.ParallelGateOperation(g2, r1))
+    eq.add_equality_group(cirq.ParallelGateOperation(g1, r12),
+                          cirq.ParallelGateOperation(g1, r21))
+
+
+def test_with_qubits_and_transform_qubits():
+    g = cirq.SingleQubitGate()
+    op = cirq.ParallelGateOperation(g, cirq.LineQubit.range(3))
+    assert op.with_qubits(*cirq.LineQubit.range(3, 0, -1)) \
```
Strilanc: Nit: google python style is to use parens for line continuations instead of `\`.

cirq/ops/parallel_gate_operation_test.py
```diff
@@ -0,0 +1,176 @@
...
+        == cirq.ParallelGateOperation(g, cirq.LineQubit.range(3, 0, -1))
+    assert op.transform_qubits(lambda e: cirq.LineQubit(-e.x)
+                               ) == cirq.ParallelGateOperation(g,
+                                                           [cirq.LineQubit(0),
+                                                           cirq.LineQubit(-1),
```
Strilanc: Nit: indentation mismatch

cirq/ops/parallel_gate_operation_test.py
```diff
@@ -0,0 +1,176 @@
...
+                                                           cirq.LineQubit(-2)])
+
+
+def test_decompose():
+    qreg = cirq.LineQubit.range(3)
+    op = cirq.ParallelGateOperation(cirq.X, qreg)
+    assert cirq.decompose(op) == cirq.X.on_each(qreg)
```
Strilanc: Test failure because we switched up the signature of on_each

cirq/ops/gate_operation_test.py
```diff
@@ -25,9 +25,17 @@ def test_gate_operation_init():
     assert v.qubits == (q,)
 
 
+def test_invalid_gate_operation():
+    three_qubit_gate = cirq.ThreeQubitGate()
+    single_qubit = [cirq.GridQubit(0, 0)]
+    with pytest.raises(ValueError):
```
Strilanc: Add `match="..."` clause to verify it's the expected type error.

(We've been really lax on this in general in the code base, and I'm trying to tighten up w.r.t. it)

cirq/ops/parallel_gate_operation_test.py
c-poole: Most of that test was designed to handle coverage with the copied `_formatted_exponent` method. I've cleaned it up with fewer gates but it still effectively performs the same tests.

cirq/ops/parallel_gate_operation_test.py
c-poole: Done

cirq/ops/parallel_gate_operation_test.py
c-poole: I've reworked the test to avoid repeated calls to the `range` method that makes using parens and matching indentation easier.

cirq/ops/parallel_gate_operation_test.py
c-poole: I've reworked the test to avoid repeated calls to the `range` method that makes using parens and matching indentation easier.

cirq/ops/parallel_gate_operation_test.py
c-poole: Fixed

cirq/ops/gate_operation_test.py
c-poole: I've added an assertion to check that the message is that of a three qubit gate complaining about not getting three qubits. I've added similar checks to the corresponding method in the `parallel_gate_operation` version.

cirq/ops/parallel_gate_operation_test.py
c-poole: Done

cirq/ops/parallel_gate_operation_test.py
```diff
@@ -14,7 +14,9 @@
 
 import pytest
 import numpy as np
+from sympy.core import Symbol
```
Strilanc: Just `import sympy`

cirq/ops/parallel_gate_operation_test.py
```diff
@@ -14,7 +14,9 @@
...
 import cirq
+from cirq import testing
```
Strilanc: Just use `cirq.testing.[...]`

Strilanc:
Cody, is this good to merge? The CLA issues were resolved with the university?

c-poole:
This is good to merge. There were a few different options in terms of how we needed to proceed with the CLA that we had to sort through but that's been resolved.

PR #1392, opened 2019-02-23
Make pylint catch strings with anomalous backslashes

### LINKED ISSUE #1187: Automatically detect docstrings that should be r-strings
When embedding latex into a docstring, it's common to use backslashes. Python interprets those as character escapes, so e.g. `\begin{bmatrix}` ends up turning into a bad character followed by `egin{bmatrix}`. We should automatically detect when this happens, and raise an error saying the user needs to prefix the docstring with `r`.

I think pylint might have a rule for this.

andbe91:
Closes #1187 

@Strilanc you are correct, pylint can catch strings containing unescaped backslahes. This of course makes pylint complain about _all_ strings in the codebase with anomalous backslashes, not only docstrings. I modified instances where warnings occurred according to what I found most appropriate. For regex patterns, I suppress the warnings from pylint, since making the pattern a raw string would change the meaning of the pattern.


cirq/circuits/qasm_output.py
```diff
@@ -124,7 +124,7 @@ def __repr__(self) -> str:
 
 
 class QasmOutput:
-    valid_id_re = re.compile('[a-z][a-zA-Z0-9_]*\Z')
+    valid_id_re = re.compile('[a-z][a-zA-Z0-9_]*\Z') # pylint: disable=anomalous-backslash-in-string
```
Strilanc: This warning is correct! Don't disable it, switch to an r-string `r'[a-z][a-zA-Z0-9_]*\Z'`.

cirq/contrib/qcircuit/qcircuit_diagram.py
```diff
@@ -47,7 +47,7 @@ def _render(diagram: circuits.TextDiagramDrawer) -> str:
         diagram2.write(2*w - 1, y, '&\\qw\\\\')
     grid = diagram2.render(horizontal_spacing=0, vertical_spacing=0)
 
-    output = '\Qcircuit @R=1em @C=0.75em {\n \\\\\n' + grid + '\n \\\\\n}'
+    output = '\\Qcircuit @R=1em @C=0.75em {\n \\\\\n' + grid + '\n \\\\\n}'
```
Strilanc: Use an r-string.

cirq/contrib/qcircuit/qcircuit_diagram_info.py
```diff
@@ -90,8 +90,8 @@ def multigate_qcircuit_diagram_info(
             str(op))
     if (info is not None) and (info.exponent != 1):
         name += '^{' + str(info.exponent) + '}'
-    box = '\multigate{' + str(n_qubits - 1) + '}{' + name + '}'
-    ghost = '\ghost{' + name + '}'
+    box = '\\multigate{' + str(n_qubits - 1) + '}{' + name + '}'
```
Strilanc: Use an r-string.

cirq/contrib/qcircuit/qcircuit_diagram_info_test.py
```diff
@@ -31,7 +31,7 @@ def test_get_qcircuit_diagram_info():
     actual_info = ccq.get_qcircuit_diagram_info(op, args)
     name = '{\\text{SWAP}^{0.5}}'
     expected_info = cirq.CircuitDiagramInfo(
-            ('\multigate{1}' + name, '\ghost' + name),
+            ('\\multigate{1}' + name, '\\ghost' + name),
```
Strilanc: Use an r-string.

cirq/contrib/qcircuit/qcircuit_diagram_info_test.py
```diff
@@ -47,7 +47,7 @@ def test_get_qcircuit_diagram_info():
             qubit_map=qubit_map)
     actual_info = ccq.get_qcircuit_diagram_info(op, args)
     expected_info = cirq.CircuitDiagramInfo(
-            ('\ghost{\\text{SWAP}}', '\multigate{1}{\\text{SWAP}}'),
+            ('\\ghost{\\text{SWAP}}', '\\multigate{1}{\\text{SWAP}}'),
```
Strilanc: Use an r-string.

dev_tools/check_pytest_v2.py
```diff
@@ -63,7 +63,7 @@ def _common_run_helper(env: env_tools.PreparedEnv,
             return True, 'Converted tests passed!'
 
         last_line = [e for e in output.split('\n') if e.strip()][-1]
-        fail_match = re.match('.+=== (\d+) failed', last_line)
+        fail_match = re.match('.+=== (\d+) failed', last_line) # pylint: disable=anomalous-backslash-in-string
```
Strilanc: This warning is correct! Use an r-string.

dev_tools/check_pytest_with_coverage.py
```diff
@@ -59,7 +59,7 @@ def perform_check(self, env: env_tools.PreparedEnv, verbose: bool):
             return True, 'Tests passed!'
 
         last_line = [e for e in output.split('\n') if e.strip()][-1]
-        fail_match = re.match('.+=== (\d+) failed', last_line)
+        fail_match = re.match('.+=== (\d+) failed', last_line) # pylint: disable=anomalous-backslash-in-string
```
Strilanc: This warning is correct! Use an r-string.

cirq/contrib/qcircuit/qcircuit_diagram.py
andbe91: To produce the correct output I need the line feeds, so I can't make the entire string an r-string. I can use a combination however, like this: `output = r'\Qcircuit @R=1em @C=0.75em {' '\n' r' \\' '\n' + grid + '\n' r' \\' '\n}'`. What do you prefer?

Strilanc: Ah, good point. If you can't use an r-string, use proper escaping with backashles intended for the output using `\\`. So it's good as is.

andbe91:
Done. Not really sure why pylint failed on Travis though. It passes locally on my computer..

Strilanc:
Fixing it in https://github.com/quantumlib/Cirq/pulls

Issue #343, opened 2018-04-25
It's too hard to run a noisy circuit correctly


olsonjonny:
I’m trying to apply the noise simulator to a circuit, but I run into this:

```
root@ce2cd89416a5:~/google_autoencoder/src/qae# python3 qae_noisy.py
Traceback (most recent call last):
  File "qae_noisy.py", line 170, in <module>
    run_sim(a,b,x,z,0.5,100,True,True,True)
  File "qae_noisy.py", line 136, in run_sim
    result = simulator.run(circuit_run, qubits = qub)
  File "/usr/local/lib/python3.5/dist-packages/cirq-0.1-py3.5.egg/cirq/google/sim/xmon_simulator.py", line 145, in run
    qubits, initial_state)[0]
  File "/usr/local/lib/python3.5/dist-packages/cirq-0.1-py3.5.egg/cirq/google/sim/xmon_simulator.py", line 192, in run_sweep
    all_step_results)
  File "/usr/local/lib/python3.5/dist-packages/cirq-0.1-py3.5.egg/cirq/google/sim/xmon_simulator.py", line 311, in simulator_iterator
    gate.half_turns)
  File "/usr/local/lib/python3.5/dist-packages/cirq-0.1-py3.5.egg/cirq/study/resolver.py", line 57, in value_of
    return self.param_dict[value.name]
KeyError: 'error_parameter0'
```
 
The implementation of the noisy simulator is done in these lines in the code I’m running:

```python
def noisy_job(a,b,x,z,alpha,exact=False):
    job = Job(test_circuit(a,b,x,z,alpha,True))
    noisy = DepolarizerChannel(probability=0.1)
    noisy_job = noisy.transform_job(job)
    return noisy_job.circuit
```
 
Notably, I do not seem to run into the error when I remove “probability=0.1”---however, I also don’t seem to be observing any error applied to the output state either.

dabacon:
We are proposing #929 as the path forward for our first version of noisy simulators.

Strilanc:
Obsoleted by all the other noisy simulator issues and work.

Issue #1355, opened 2019-02-07
Easy way to add noise to circuits


dabacon:
There should be easy helpers that take a noise free circuit and then add in noise after each gate (and after preparation).

peterse:
I can attempt this

dabacon:
We discussed a little off line: the most basic thing to start with is just adding in moments with noise operators.

Strilanc:
I think we might want the noise model to be an object that is passed into the simulator. This issue is very design problem heavy; not an easy one.

peterse:
@Strilanc could you elaborate a bit more on the role of the simulator here, compared to a (somewhat post hoc) set of moments implementing noisy channels on gates?

Strilanc:
I think that we need a concept of a `NoiseModel`, which associates error channels with gates, doing nothing, particular combinations of gates, etc. A `NoiseModel` *could* take a circuit and produce a circuit with injected noise, but it should also be possible to pass a noise model into simulate/run methods and have the transformation happen inline with the simulation.

Because this has such a large impact on the APIs we expose, and has to fit nicely with the rest of the library, there could be a lot of bike shedding over the particulars (strict spec) and we're not exactly sure what we want (vague spec). Not an ideal first issue.

You are of course free to try. But I think I would want to write up a design doc for you to follow, instead of leaving you to your own devices to implement it.

Strilanc:
(Also, this is one of the things we want to get done for the next release, so we're likely to do it ourselves rather than waiting on a third party. I don't want to send you down a garden path doing work we won't keep.)

peterse:
Okay, I understand completely

Can you think of dependencies on any specific Issues that this noise model would have? We're eager to see cirq noise support move forward and I'd like to contribute wherever it'd be more appropriate

Strilanc:
- Need to separate code not related to this issue from current PR
- @dabacon re-review PR

Strilanc:
Fixed by https://github.com/quantumlib/Cirq/pull/1520

PR #1395, opened 2019-02-26
Define cirq.run/run_sweep/NoiseModel/NO_NOISE/ConstantQubitNoiseMode

Strilanc:
- `cirq.NoiseModel` has semi-abstract noisy_moment/moments/operation methods (at least one must be implemented)
- `cirq.ConstantQubitNoiseModel` applies a given gate to every qubit at the end of every moment
- `cirq.NO_NOISE` is a noise model that returns its inputs unchanged
- `cirq.run`/`cirq.run_sweep` delegates to an appropriate simulator by analyzing their input
- `DensityMatrixSimulator` now has a `noise` attribute that takes a `NoiseModel`
- Simulators now force keyword args for `dtype`

Part of https://github.com/quantumlib/Cirq/issues/1355
Fixes https://github.com/quantumlib/Cirq/issues/343


Strilanc:
Please make comments on the design. Is this a reasonable way to define noise?

peterse:
I agree w Dave  - I think it would be useful for the NoiseModel to be applied before running the simulator so that the user could access the modified circuit (and we'd likely end up writing such a noise wrapper in our external library if its not available at the cirq level). 

I do like the flexibility in the NoiseModel though. One use we have for noise is to apply channels conditionally on gate type, e.g. greater damping for 2-qubit gates, as an ad hoc way of modelling gate fidelity. So the generality for the NoiseModel is nice in that respect

Strilanc:
> I worry a bit that we are again putting into the simulators transformations on the circuit. Of course the user has to supply the noise model, so that is good (better than decompose, which is hidden). But note that the user here cannot even see what the noisy circuit that they just ran looks like.

The fact that the user explicitly asks for the noise should make it not confusing. I think that was the main downside of automatic decomposition.

I agree that we should have a way to spit out the noisy circuit so the user can inspect it. Especially if the noise model is randomly inserting gates, instead of deterministically inserting channels. Perhaps we could include the simulated moment operations in the TrialResult?

Strilanc:
> I worry a bit about cirq.run because what about engine?

I was picturing an argument like `service` or similar that defaults to local simulation. I think local simulation is the right default.

dabacon:
> I was picturing an argument like `service` or similar that defaults to local simulation. I think local simulation is the right default.

I

Strilanc:
@dabacon Did you accidentally hit the close button while typing?

PR #1396, opened 2019-02-26
Delete cirq.Symbol, fix several snags with sympy

Strilanc:
- Fixed ParamResolver handling sympy formulas incorrectly
- `repr(sympy.bla)` doesn't return the right thing; introduced a `proper_repr` method in `_compat` and merged numpy-override code into the one place
- Improved PauliString's `__str__` method to return things like `'(+X(a)*Y(b))'`
- Tweaked several operation `__repr__` methods to handle sympy formulas in a special way
- Add `import sympy` to the default setup code of repr tests
- Reported and hackily worked around https://github.com/sympy/sympy/issues/16086, https://github.com/sympy/sympy/issues/16087, https://github.com/sympy/sympy/issues/16074

Fixes https://github.com/quantumlib/Cirq/issues/393

The following code now works!

```python
import cirq
import sympy
import numpy as np
q = cirq.NamedQubit('q')
θ = sympy.Symbol('θ')
c = cirq.Circuit.from_ops(cirq.Rx(θ).on(q), cirq.measure(q))
results = cirq.Simulator().run_sweep(c, cirq.Linspace('θ', 0, 2*np.pi, 100), repetitions=1000)
plt.plot([r.histogram(key=q)[1] for r in results])
plt.show()
```

image


bryano:
FYI, the new sympy changes break some tests in OpenFermion-Cirq related to apply_unitary.

cirq/_compat.py
```diff
@@ -22,10 +22,32 @@
...
         return abs(fractions.gcd(a, b))
 else:
     from math import gcd  # pylint: disable=unused-import
+
+
+def proper_repr(value: Any) -> str:
+    """Overrides a few cases that return strings that don't parse."""
+
+    if isinstance(value, sympy.Basic):
+        fixed_tokens = ['Symbol', 'pi', 'Mul', 'Add']
```
dabacon: do we know that these are a complete list of problematic tokens?

cirq/_compat.py
```diff
@@ -22,10 +22,32 @@
...
     import fractions   # pylint: disable=unused-import
 
     def gcd(a, b):
         return abs(fractions.gcd(a, b))
 else:
     from math import gcd  # pylint: disable=unused-import
+
+
+def proper_repr(value: Any) -> str:
+    """Overrides a few cases that return strings that don't parse."""
```
dabacon: I'd add a list of these cases to doc string.

cirq/_compat.py
Strilanc: We know that it isn't, but this is a hacky workaround until https://github.com/sympy/sympy/issues/16074 gets fixed.

cirq/_compat.py
Strilanc: Done.

cirq/_compat.py
Strilanc: Commented now.

Strilanc:
Fixed the pytest2 failures (a bunch of workarounds for sympy handling unicode a bit wonkily). Covered the uncovered code. A few more tweaks to make it smooth.

Strilanc:
@bryano which tests are broken by it?

Issue #1397, opened 2019-02-26
Sweeps such as `cirq.Linspace` should accept `sympy.Symbol`s (by turning them into strings)


Strilanc:
This code, which currently causes exceptions:

```python
var = sympy.Symbol('x')
return cirq.Linspace(var, 0, np.pi, 100)
```

should have the equivalent behavior as this code:

```python
return cirq.Linspace('x', 0, np.pi, 100)
```

via `Linspace` checking for `sympy.Symbol` and calling `str` on it to get the actual key if so.

vtomole:
I'll take this.

timothyman:
@vtomole Feel free to completely ignore this, if you have spent significant time on this already, but I have a pull request ready to go for this. I can wait for the next "good first issue", no problem.

I have a question about my implementation, though. Right now I do a type-check whether the `key` is of `sympy.Symbol` before calling `str`, but I'm wondering if that is at all necessary? Anything can be used as a `key`, as long as it's a string? Can we just call `str` on any input?

vtomole:
@timothyman It's yours!

timothyman:
Thanks, @vtomole!

Strilanc:
@timothyman I would avoid just calling `str` on the input, because that will mask mistakes where a nonsensical value is entered. Also, it would do things like cause the number 1 and the qubit LineQubit(1) to map down to the same thing, which is probably not desirable.

PR #1398, opened 2019-02-26
Linspace accepts sympy symbol

timothyman:
- Added K as type for key which can be `str` or `sympy.Symbol`
- Added call to `str` in `SingleSweep` class constructor

Fixes https://github.com/quantumlib/Cirq/issues/1397


timothyman:
I signed it!

timothyman:
I signed it!

cirq/study/sweeps.py
```diff
@@ -240,11 +240,14 @@ def __str__(self):
...
 
+K = TypeVar('K', str, Symbol)  # Key can be of type str or sympy.Symbol
+
+
 class SingleSweep(Sweep):
     """A simple sweep over one parameter with values from an iterator."""
 
-    def __init__(self, key: str) -> None:
-        self.key = key
+    def __init__(self, key: K) -> None:
```
Strilanc: Nit: I would just inline `Union[str, sympy.Symbol]` in each case.

cirq/study/sweeps.py
```diff
@@ -240,11 +240,14 @@ def __str__(self):
         return ' + '.join(str(s) for s in self.sweeps)
 
 
+K = TypeVar('K', str, Symbol)  # Key can be of type str or sympy.Symbol
```
Strilanc: You want `Union[str, sympy.Symbol]` instead of a type variable.

timothyman:
- Use `import sympy` instead of `from sympy import Symbol` to make `sympy.Symbol` explicit
- Replaced `TypeVar` for `Union[str, sympy.Symbol]`
- Added type-check for `sympy.Symbol` before calling `str` in `SingleSweep` class constructor

Issue #1139, opened 2018-11-09
Define _unitary_ for QasmUGate and QasmTwoQubitGate


kevinsung:

jaywha:
In [qasm_output.py](https://github.com/quantumlib/Cirq/blob/master/cirq/circuits/qasm_output.py), I see the 2 classes, but what should _\_unitary_\_ be?  
Is it a [unitary matrix](http://mathworld.wolfram.com/UnitaryMatrix.html)?

Strilanc:
@jaywha Yes, the output of `_unitary_` is the unitary matrix of the operation. E.g. for the Pauli Z gate the matrix is diag(1, -1).

PR #1145, opened 2018-11-12
_unitary_() init for qasm_output.py

vtomole:
Fixes https://github.com/quantumlib/Cirq/issues/1139

@Strilanc I don't think my implementation of `_unitary_` for QasmTwoQubitGate is correct as I tried to implement a test for CNOT, but it's not passing. Used: https://arxiv.org/pdf/quant-ph/0507171.pdf as a reference. Is it the test or the implementation?

PR #1402, opened 2019-02-27
_unitary_() init for qasm_output.py

Strilanc:
Fixes https://github.com/quantumlib/Cirq/issues/1139

Derived from https://github.com/quantumlib/Cirq/pull/1145 by @vtomole


Strilanc:
Good bot :)

Strilanc:
Bad bot :(

cirq/circuits/qasm_output.py
```diff
@@ -52,14 +52,20 @@ def _qasm_(self,
...
     def __repr__(self) -> str:
         return 'cirq.QasmUGate({}, {}, {})'.format(self.lmda,
                                                    self.theta,
                                                    self.phi)
 
+    def _unitary_(self) -> np.ndarray:
+        rz_phi_matrix = protocols.unitary(ops.Rz(self.phi * np.pi))
+        ry_theta_matrix = protocols.unitary(ops.Ry(self.theta * np.pi))
+        rz_lambda_matrix = protocols.unitary(ops.Rz(self.lmda * np.pi))
+        return rz_phi_matrix.dot(ry_theta_matrix.dot(rz_lambda_matrix))
```
viathor: Optional: You can also compute the four elements of this 2x2 matrix using formula (2) in [this paper](https://arxiv.org/abs/1707.03429). Something like this:

```python
pre_z_phase = np.exp(1j * np.pi * self.lmda)
post_z_phase = np.exp(1j * np.pi * self.phi)
y_phase = np.exp(1j * np.pi * np.theta)
c, s = y_phase.real, 1j * y_phase.imag
return np.array([[pre_z_phase.conj() * post_z_phase.conj() * c, ...]])
```

This should be a bit faster. The existing formulas would make a good unit test though.

Strilanc: I cleaned up the code a bit more, but didn't go all the way down to writing out the matrix. (This is not a method on a hot path.) I referenced that paper as the source.

Strilanc:
Google booooootttt...

PR #1403, opened 2019-02-27
Make circuit_compare._infer_qubit_count use Gate.num_qubits

Strilanc:
- Check for inconsistent qubit counts
- Fix repr eval failing with strange error messages in some cases


Strilanc:
Extracted this from https://github.com/quantumlib/Cirq/pull/1388

PR #1404, opened 2019-02-27
Pretty print TrialResult in ipython

Strilanc:


Strilanc:
Extracted from https://github.com/quantumlib/Cirq/pull/1395

Issue #1386, opened 2019-02-16
Don't automatically decompose circuits in Engine.program_as_schedule


Strilanc:
Instead, validate that the circuit is targeting a supported device and that the device validates the circuit.

dabacon:
Currently if a program has a device when passed into engine we validate it.  We don't do any decomposition and a user must specify a serializable gate set.

Think this issue is now stale for those reasons.  Closing.

PR #1405, opened 2019-02-27
XmonSimulator now requires given circuits to have an XmonDevice that validates the circuit

Strilanc:
- No longer auto decomposes incompatible circuits
- No longer allows circuits using the unconstrained device
- Fixed xmon_simulator_test breaking these rules constantly
- Benchmarking code is now careful to generate local operations

Part of https://github.com/quantumlib/Cirq/issues/1386


cirq/circuits/circuit.py
```diff
@@ -171,7 +171,7 @@ def __getitem__(self, key: int) -> ops.Moment:
 
     def __getitem__(self, key):
         if isinstance(key, slice):
-            return Circuit(self._moments[key])
+            return Circuit(self._moments[key], self.device)
```
dabacon: should maybe add test to check that device is propagated

cirq/google/sim/xmon_simulator.py
```diff
@@ -133,15 +143,20 @@ def _run(
         repetitions: int,
     ) -> Dict[str, List[np.ndarray]]:
         """See definition in `cirq.SimulatesSamples`."""
-        xmon_circuit, keys = self._to_xmon_circuit(
-            circuit,
-            param_resolver)
-        if xmon_circuit.are_all_measurements_terminal():
-            return self._run_sweep_sample(xmon_circuit, repetitions)
+
```
dabacon: update doc describing the verification and throwing of error

cirq/circuits/circuit.py
Strilanc: Done.

cirq/google/sim/xmon_simulator.py
Strilanc: Done.

PR #1406, opened 2019-02-27
Delete cirq/contrib/jobs

Strilanc:
- Obsoleted by new noisy simulation and coming NoiseModel

Part of https://github.com/quantumlib/Cirq/issues/343


Strilanc:
@dstrain115  Well, you did what I asked for so I considered it to have been completed successfully... but the idea fell apart in practice.

Strilanc:
Pylint looks like it failed to parse the code.

We should try to get that down to a few lines repro so it can get fixed.

PR #1408, opened 2019-02-27
Auto-wrap dictionaries into ParamResolver and skip empty resolutions

Strilanc:
- `cirq.resolve_parameters`, `cirq.Simulator`, etc can now be given a raw dictionary instead of one wrapped into `cirq.ParamResolver`
- `cirq.resolve_parameters` now bounces its input when given a vacuous parameter resolver
- `cirq.ParamResolver` now has a `__new__` method that auto-flattens when putting a resolver inside a resolver, and handles being given None


cirq/sim/simulator.py
```diff
@@ -44,7 +44,9 @@ class SimulatesSamples(metaclass=abc.ABCMeta):
     def run(
         self,
         program: Union[circuits.Circuit, schedules.Schedule],
-        param_resolver: Optional[study.ParamResolver] = None,
+        param_resolver: Union[None,
+                              Dict[str, float],
+                              study.ParamResolver] = None,
```
maffoo: I would suggest making a type alias for all the things that a param resolver can be created from, then you can use that in all these other places. For example:
```python
# in resolver.py
class ParamResolver:
    From = Union['ParamResolver', Dict[str, float]]

    # optional: a classmethod to create an instance from any of the union types
    @classmethod
    def from(cls, obj: From) -> 'ParamResolver':
        ...

# in simulator.py, etc.
def run(
    ...,
    param_resolver: Optional[study.ParamResolver.From] = None,
    ...):
```
The name in this example is inspired by Rust's [`From` trait](https://doc.rust-lang.org/std/convert/trait.From.html), but whatever the name I think this it's a good idea to capture these sorts of unions with a type alias. Also, note that the `from` method I've shown can be done in `__new__` as you've done below; my only hesitation with that is that I've seen cases where this logic grows quite complicated for converting from various possible types and having that in `__new__` means you get overhead on _every_ instantiation. That might not be so critical for `ParamResolver`, but it's worth considering.

Strilanc: Done. Introduced `cirq.ParamResolverOrSimilarType`.

cirq/__init__.py
```diff
@@ -231,6 +231,7 @@
     ComputeDisplaysResult,
     Linspace,
     ParamResolver,
+    ParamResolverOrSimilarType,
```
dabacon: Why not just ParamResolverType.  I mean "or similar" is kind of vague.  I guess ParamResolverType is also not specific enough.

cirq/protocols/resolve_parameters.py
```diff
@@ -56,7 +56,9 @@ def is_parameterized(val: Any) -> bool:
         return False
 
 
-def resolve_parameters(val: Any, param_resolver: 'cirq.ParamResolver') -> Any:
+def resolve_parameters(
+        val: Any,
+        param_resolver: 'cirq.ParamResolverOrSimilarType') -> Any:
```
dabacon: does this have to be quoted?  Can't you "from cirq.study import resolvers" and use resolvers.PROST?  This feels a bit better to me because my IDE treat it as a type I can jump to, whereas for a string it doesn't do this.

cirq/protocols/resolve_parameters.py
```diff
@@ -73,6 +75,15 @@ def resolve_parameters(val: Any, param_resolver: 'cirq.ParamResolver') -> Any:
         If `val` has no `_resolve_parameters_` method or if it returns
         NotImplemented, `val` itself is returned.
     """
+    # HACK: break cycle.
+    from cirq import ParamResolver
```
dabacon: nit: move down to right before it is used?  Not sure...imports in code are strange.

cirq/protocols/resolve_parameters.py
Strilanc: Yes, it has to be quoted. It breaks a dependency cycle.

When I ctrl+click on the string in pycharm, it takes me to the definition. At least some IDEs understand that string type annotations refer to code.

cirq/protocols/resolve_parameters.py
Strilanc: Done.

cirq/__init__.py
Strilanc: Acknowledged. Yeah, it's not specific enough.

PR #1413, opened 2019-02-28
Make pytest-benchmark optional for machines that don't have it installed.

vtomole:


vtomole:
Part of https://github.com/quantumlib/Cirq/issues/1124

vtomole:
Looks like the fixture is not automatically being set if it's set to `None`. Closing for now.

PR #1416, opened 2019-02-28
Adds ShiftSwapNetwork

bryano:


bryano:
Test failure unrelated to this PR.

cirq/contrib/acquaintance/bipartite.py
```diff
@@ -147,12 +147,16 @@ def _circuit_diagram_info_(self, args: protocols.CircuitDiagramInfoArgs
...
         return wire_symbols
 
-
     def __repr__(self):
-        return ('cirq.contrib.acquaintance.bipartite.BipartiteSwapNetworkGate'
-                '({}, {})'.format(repr(self.subgraph), self.part_size))
+        args = (repr(self.subgraph), repr(self.part_size))
+        if self.swap_gate != ops.SWAP:
+            args += (repr(self.swap_gate),)
+        return ('cirq.contrib.acquaintance.bipartite.BipartiteSwapNetworkGate' +
```
Strilanc: Nit: the + is not needed

cirq/contrib/acquaintance/shift_swap_network.py
```diff
@@ -0,0 +1,139 @@
...
+
+from cirq import ops, protocols
+from cirq.contrib.acquaintance.gates import acquaint
+from cirq.contrib.acquaintance.permutation import (
+        PermutationGate)
+from cirq.contrib.acquaintance.shift import (
+        CircularShiftGate)
+
+
+class ShiftSwapNetworkGate(PermutationGate):
```
Strilanc: This seems like a specific instance of a more general concept that seems like a "nested swap network" or perhaps "tensor product of two swap networks". Is there a way for you to define a general mechanism for combining permutations over groups of qubits, instead of introducing special cases of that idea?

cirq/contrib/acquaintance/bipartite.py
bryano: Removed

cirq/contrib/acquaintance/shift_swap_network.py
bryano: I see what you're saying, but I doubt the abstraction is worth it at this point.

Strilanc: Why did you add this specific instance? Did you run into it in many places?

bryano: It's a building block for a strategy that corresponds to UCCSD.

Strilanc:
Looks like this was impacted by the rename from QubitId to Qid

Issue #963, opened 2018-09-23
Extend cirq.CliffordGate.__pow__ to do square roots with Clifford results


Strilanc:
Only need to support exponents that return Clifford results. For example, `cirq.CliffordGate.Z**0.5` should be `cirq.CliffordGate.Z_sqrt`.

Strilanc:
Ah, good catch. Renamed issue to focus on the specific missing case.

vtomole:
Forgot to add "closes" on https://github.com/quantumlib/Cirq/pull/1431. Please close.

dstrain115:
Closed per vtomole.  Also added on assignee, since he did all the work!!!

PR #1417, opened 2019-03-02
CliffordGate**0.5/-0.5

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/963

### DIFF
--- a/cirq/circuits/circuit.py
+++ b/cirq/circuits/circuit.py
@@ -30,7 +30,6 @@ from typing import (
 import json
 import re
 import numpy as np
-import sympy
 
 from cirq import devices, ops, study, protocols
 from cirq.circuits._bucket_priority_queue import BucketPriorityQueue
@@ -1541,7 +1540,7 @@ def _formatted_exponent(info: protocols.CircuitDiagramInfo,
                         args: protocols.CircuitDiagramInfoArgs
                         ) -> Optional[str]:
 
-    if isinstance(info.exponent, sympy.Basic):
+    if protocols.is_parameterized(info.exponent):
         name = str(info.exponent)
         return (name
                 if _is_exposed_formula(name)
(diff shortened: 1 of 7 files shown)


cirq/ops/clifford_gate.py
```diff
@@ -16,6 +16,7 @@
 
 import numpy as np
 
+import cirq
```
Strilanc: Don't `import cirq` inside cirq; it tends to create import loops in python 2.

cirq/ops/clifford_gate.py
```diff
@@ -85,7 +87,7 @@ def from_single_map(pauli_map_to: Optional[Dict[Pauli, Tuple[Pauli, bool]]]
             z_to: The transform from cirq.Z
         """
         rotation_map = SingleQubitCliffordGate._validate_map_input(
-                                        1,
+            1,
```
Strilanc: Keep indentation consistent

cirq/ops/clifford_gate.py
```diff
@@ -122,7 +124,7 @@ def from_double_map(pauli_map_to: Optional[Dict[Pauli, Tuple[Pauli, bool]]]
             z_to: The transform from cirq.Z
         """
         rotation_map = SingleQubitCliffordGate._validate_map_input(
-                                        2,
+            2,
```
Strilanc: consistent indentation

cirq/ops/clifford_gate.py
```diff
@@ -207,8 +209,25 @@ def _value_equality_values_(self):
                 self.transform(pauli_gates.Z))
 
     def __pow__(self, exponent) -> 'SingleQubitCliffordGate':
-        if exponent != -1:
+        if exponent == 0.5:
```
Strilanc: Instead of a big if/else block, I would define a private module-level constant dictionary mapping (float, gate) to gate, with each of the special cases encoded.

PR #1421, opened 2019-03-05
added PeriodicValue._is_parameterized_

bryano:


bryano:
What do you think of `cirq.is_parameterized(sympy.Symbol('s'))` evaluating to `True`?

Strilanc:
@bryano I think that makes complete sense. Need to make sure resolve parameters also handles it.

PR #1422, opened 2019-03-05
Remove calls to benchmark in examples_test.py

vtomole:


vtomole:
Pulled request on a virtual machine that i guess my credentials were not good on. Closing.

PR #1429, opened 2019-03-13
Update Copyright Year

vtomole:


Strilanc:
Our internal guidelines are to leave these years alone (it should be creation date, not latest modified date).

PR #1430, opened 2019-03-13
[WIP] Drop empty moments

### LINKED ISSUE #1316: Circuit optimizations should remove empty Moments
Sometimes running a circuit optimizer leaves Moments empty. What reason is there to keep them around? They add some overhead when iterating through the circuit and also make text diagrams take up more space. I'm not too familiar with the code of the optimizers and it could be the case that removing empty Moments doesn't fit nicely into the logic there.

balopat:
Fixes #1316. 

However, I am not yet convinced that this is the best direction to go as the implicit empty moment removal gets a bit tricky to reason about - it could be that leaving empty moment removal an explicit post-operation is what the pattern should be that we support - that is, we should just close #1316 and leave things as they are.

Having said that, if we decide that we like this superclass direction, I can clean up this PR further with more unit tests and docs as needed. 

Feedback, thoughts more than welcome!

### DIFF
--- a/cirq/__init__.py
+++ b/cirq/__init__.py
@@ -24,6 +24,7 @@ from cirq.circuits import (
     InsertStrategy,
     PointOptimizationSummary,
     PointOptimizer,
+    Optimizer,
     QasmOutput,
     TextDiagramDrawer,
     Unique,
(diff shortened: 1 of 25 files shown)


Strilanc:
I agree that it's not the right way to go. Closing this PR and the associated issue. Thanks @balopat .

PR #1431, opened 2019-03-13
CliffordGate**0.5/-0.5

vtomole:
This one https://github.com/quantumlib/Cirq/pull/1417 automatically closed after not "pull rebasing".


vtomole:
@Strilanc Ready for the second round The formatting can be annoying. I always use [autopep8](https://github.com/hhatto/autopep8) before committing and "non relevant" code always seem to change. I'm wondering if we should define a standard formatter to avoid this.

Strilanc:
@vtomole Yes, we should definitely have a formatting check as part of the continuous integration.

Strilanc:
At some point I'll write an incremental format checker + incremental format applier, that only checks and only touches code you've edited.

vtomole:
@Strilanc  Fix the nits by using Emacs to format and it looks like it has different opinions from autopep8.

vtomole:
Ready for merge.

PR #1437, opened 2019-03-13
Should this example be corrected?

Ashalynd:
Assuming that matrix is represented like in numpy, the example looks confusing. (Unless I am overlooking something.)


Strilanc:
The issue is whether the array should be row major or col major?

Ashalynd:
> The issue is whether the array should be row major or col major?

Judging by the comments and by the example output, it is supposed to be row major (like in numpy). 
Unless I am overlooking something.

PR #1439, opened 2019-03-14
Add LinearDict

viathor:
This is the core of the logic behind linear combinations of gates and operations. Together with follow-ups will supersede #1299.

Follow-up PRs to use this logic:
 * linear combinations of gates,
 * linear combinations of operations,
 * change pauli expansions to use `LinearDict`.

Example of another potential future use of `LinearDict`: Schmidt decomposition for operators.


cirq/value/linear_dict.py
```diff
@@ -0,0 +1,173 @@
...
+        """Initializes linear combination from a collection of terms.
+
+        Args:
+            terms: Mapping of abstract vectors to coefficients in the linear
+                combination being initialized.
+        """
+        super(LinearDict, self).__init__()
+        self.update(terms)
+
+    def clean(self, *, tolerance: float=1e-9) -> None:
```
Strilanc: Rename `tolerance` to `atol`.

cirq/value/linear_dict.py
```diff
@@ -0,0 +1,173 @@
...
+        """Modifies self by deleting terms with small coefficients."""
+        negligible = [v for v, c in self.items() if abs(c) <= tolerance]
+        for v in negligible:
+            del self[v]
+
+    def copy(self) -> 'LinearDict':
+        return LinearDict(super(LinearDict, self).copy())
```
Strilanc: Just `super()`

cirq/value/linear_dict.py
```diff
@@ -0,0 +1,173 @@
...
+    a single-term linear combination with the sole coefficient equal to 1.0.
+    """
+    def __init__(self, terms: Mapping[Any, Scalar]) -> None:
+        """Initializes linear combination from a collection of terms.
+
+        Args:
+            terms: Mapping of abstract vectors to coefficients in the linear
+                combination being initialized.
+        """
+        super(LinearDict, self).__init__()
```
Strilanc: Just `super()`

cirq/value/linear_dict.py
```diff
@@ -0,0 +1,173 @@
...
+        """Modifies self by deleting terms with small coefficients."""
+        negligible = [v for v, c in self.items() if abs(c) <= tolerance]
+        for v in negligible:
+            del self[v]
+
+    def copy(self) -> 'LinearDict':
+        return LinearDict(super(LinearDict, self).copy())
+
+    def __getitem__(self, vector: Any) -> Scalar:
+        if vector in self:
```
Strilanc: Rename `vector` to `key` then just `return super().get(key, 0)`

cirq/value/linear_dict.py
```diff
@@ -0,0 +1,173 @@
...
+            return super(LinearDict, self).__getitem__(vector)
+        return 0.0
+
+    def __iadd__(self, other: Any) -> 'LinearDict':
+        if not isinstance(other, LinearDict):
+            other = LinearDict({other: 1})
```
Strilanc: I'm not convinced this autoconversion is a good idea. It doesn't provide much benefit, because `v[x] += 1` is just as short as `v += x`, and it seems like it turns a lot of type mistakes into silent wrong behavior.

cirq/value/linear_dict.py
```diff
@@ -0,0 +1,173 @@
...
+
+        for vector, coefficient in other.items():
+            self[vector] = self[vector] + coefficient
```
Strilanc: Can we delegate to `super()` instead of invoking our own methods?

cirq/value/linear_dict.py
```diff
@@ -0,0 +1,173 @@
...
+
+        Presence or absence of terms with coefficients exactly equal to
+        zero does not affect outcome.
+
+        Not appropriate for most practical purposes due to sensitivity to
+        numerical error in floating point coefficients. Use cirq.approx_eq()
+        instead.
+        """
+        if not isinstance(other, LinearDict):
+            other = LinearDict({other: 1})
```
Strilanc: *Definitely* don't do this. For example, this means LinearDict behaves unlike other values when nested. A singleton list is not equal to the one item in the list, and we should behave the same way.

If a value wants to compare as equal to a LinearDict that value should be the one to special case LinearDict instead of vice versa.

cirq/value/linear_dict.py
```diff
@@ -0,0 +1,173 @@
...
+
+        all_vs = set(self.keys()) | set(other.keys())
+        return all(self[v] == other[v] for v in all_vs)
+
+    def __ne__(self, other: Any) -> bool:
+        """Checks whether two linear combinations are not exactly equal.
+
+        See __eq__().
+        """
+        if not isinstance(other, LinearDict):
```
Strilanc: `return not self == other`

cirq/value/linear_dict.py
```diff
@@ -0,0 +1,173 @@
...
+            self[vector] *= a
+        self.clean(tolerance=0)
+        return self
+
+    def __mul__(self, a: Scalar) -> 'LinearDict':
+        result = self.copy()
+        result *= a
+        return result
+
+    def __rmul__(self, a: Scalar) -> 'LinearDict':
```
Strilanc: You need to override `__contains__` to return True iff a key is present with non-zero value.

cirq/value/linear_dict.py
```diff
@@ -0,0 +1,173 @@
...
+        return self.__mul__(a)
+
+    def __truediv__(self, a: Scalar) -> 'LinearDict':
+        return self.__mul__(1 / a)
+
```
Strilanc: We could override `__iter__` to not emit keys mapped to 0.

cirq/value/linear_dict.py
```diff
@@ -0,0 +1,173 @@
...
+            del self[v]
+
+    def copy(self) -> 'LinearDict':
+        return LinearDict(super(LinearDict, self).copy())
+
+    def __getitem__(self, vector: Any) -> Scalar:
+        if vector in self:
+            return super(LinearDict, self).__getitem__(vector)
+        return 0.0
+
```
Strilanc: Override `__setitem__` so that it deletes when setting to 0.

cirq/value/linear_dict.py
```diff
@@ -0,0 +1,173 @@
...
+    def __bool__(self) -> bool:
+        return not all(c == 0 for c in self.values())
+
+    def __eq__(self, other: Any) -> bool:
+        """Checks whether two linear combinations are exactly equal.
+
+        Presence or absence of terms with coefficients exactly equal to
+        zero does not affect outcome.
+
+        Not appropriate for most practical purposes due to sensitivity to
```
Strilanc: I would quickly scan through all of python's magic methods to see if we missed any more.

cirq/value/linear_dict.py
viathor: In future episodes, it turns out that this allows us to say things such as `2 * cirq.X + cirq.Z`. Here, `2 * cirq.X` is already an instance of `LinearDict` and a `Gate` that is added to it becomes a new key. You get `LinearDict{cirq.X: 2, cirq.Z: 1})`.

Strilanc: I agree that we want expressions like that, but I think that particular magic should live in `cirq.X` instead of in `LinearDict`.

cirq/value/linear_dict.py
viathor: I like `vector`. Argument name `key` just tells the reader what they already know: this is a type of dictionary so naturally `__getitem__()` takes a key as argument. OTOH, `vector` reminds them that `LinearDict` endows its keys with additional meaning: they represent vectors which are combined in the linear combination.

Agreed on the rest of your comment.

Strilanc: Perhaps `basis` would be even more accurate?

cirq/value/linear_dict.py
viathor: Agreed.

cirq/value/linear_dict.py
viathor: Considered it. Decided against to avoid the implication that elements of combination are linearly independent. We never check this and user should not expect it. In particular, this is legal: `LinearDict({cirq.X: 1, cirq.Z: 1, cirq.H: 1})` even though the operators are linearly dependent.

Will clarify in class docstring.

cirq/value/linear_dict.py
viathor: I moved it here since it was duplicated between Gate-keyed and Operation-keyed dicts. On second though, I agree with you it doesn't really belong here. Will remove.

cirq/value/linear_dict.py
viathor: Yeah, there is more, e.g. `__len__()` etc. Will check carefully.

cirq/value/linear_dict.py
viathor: Initially agreed. Then I discovered this leads surprising behavior when writing loops such as:

```python
for vector in linear_combination:
    linear_combination[vector] *= something
```

The surprising behavior is the `RuntimeError` exception thrown when dictionary changes size during iteration. I think we should use inherited `__setitem__()` to avoid this. Users can call `clean()` to remove the zero terms. Presence of these terms should not have any consequences for observable behavior of `LinearDict`.

Strilanc: We can bypass that issue by having `__iter__` return `list(super().__iter__())`

cirq/value/linear_dict.py
viathor: Done.

cirq/value/linear_dict.py
viathor: Done.

cirq/value/linear_dict.py
viathor: Done.

cirq/value/linear_dict.py
viathor: Done: `return super().get(vector, 0)` and docstring.

cirq/value/linear_dict.py
viathor: Done.

cirq/value/linear_dict.py
viathor: Done.

cirq/value/linear_dict.py
viathor: Done.

cirq/value/linear_dict.py
viathor: Done.

cirq/value/linear_dict.py
viathor:
```
In [1]: dir(dict())
Out[1]: 
['__class__',                       # override not needed
 '__contains__',                    # done
 '__delattr__',                     # override not needed
 '__delitem__'                      # override not needed
 '__dir__',                         # override not needed
 '__doc__',                         # set from docstring
 '__eq__',                          # done
 '__format__',                      # done
 '__ge__',                          # override not needed
 '__getattribute__',                # override not needed
 '__getitem__',                     # done
 '__gt__',                          # override not needed
 '__hash__',                        # override not needed
 '__init__',                        # done
 '__init_subclass__',               # override not needed
 '__iter__',                        # done
 '__le__',                          # override not needed
 '__len__',                         # done
 '__lt__',                          # override not needed
 '__ne__',                          # done
 '__new__',                         # override not needed
 '__reduce__',                      # override not needed
 '__reduce_ex__',                   # override not needed
 '__repr__',                        # done
 '__setattr__',                     # override not needed
 '__setitem__',                     # done
 '__sizeof__',                      # override not needed
 '__str__',                         # done
 '__subclasshook__',                # override not needed
 'clear',                           # override not needed
 'copy',                            # done
 'fromkeys',                        # done
 'get',                             # done
 'items',                           # done
 'keys',                            # done
 'pop',                             # override not needed
 'popitem',                         # override not needed
 'setdefault',                      # override not needed
 'update',                          # done
 'values']                          # done
```

cirq/value/linear_dict.py
viathor: Done. Employed this snapshot-grabbing approach in other similar places, too, e.g. `items()`.

cirq/value/linear_dict.py
viathor: Done.

cirq/value/linear_dict.py
viathor: Done.

cirq/value/linear_dict.py
```diff
@@ -0,0 +1,263 @@
...
+_TDefault = TypeVar('_TDefault')
+
+
+class LinearDict(Dict[TVector, Scalar]):
+    """Represents linear combination of things.
+
+    LinearDict implements the basic linear algebraic operations of vector
+    addition and scalar multiplication for linear combinations of abstract
+    vectors. Keys represent the vectors, values represent their coefficients.
+    The only requirement on the keys is that they be hashable.
```
dabacon: Do they also have to support equality (think you delegate to vector equality in __eq__?

viathor: Yes! This is implied by ['hashable'](https://docs.python.org/3/glossary.html#term-hashable). Added clarification.

cirq/value/linear_dict.py
```diff
@@ -0,0 +1,265 @@
...
+                combination being initialized.
+        """
+        super().__init__()
+        self.update(terms)
+
+    @classmethod
+    def fromkeys(cls, vectors, coefficient=0):
+        return LinearDict(dict.fromkeys(vectors, complex(coefficient)))
+
+    def clean(self, *, atol: float=1e-9) -> 'LinearDict':
```
Strilanc: Needs a docstring.

cirq/value/linear_dict.py
```diff
@@ -0,0 +1,265 @@
...
+        for v in negligible:
+            del self[v]
+        return self
+
+    def copy(self) -> 'LinearDict':
+        return LinearDict(super().copy())
+
+    def keys(self) -> KeysView[TVector]:
+        snapshot = self.copy().clean(atol=0)
+        return super(LinearDict, snapshot).keys()
```
Strilanc: `super()`

cirq/value/linear_dict.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def values(self) -> ValuesView[Scalar]:
+        snapshot = self.copy().clean(atol=0)
+        return super(LinearDict, snapshot).values()
```
Strilanc: `super()`

cirq/value/linear_dict.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def items(self) -> ItemsView[TVector, Scalar]:
+        snapshot = self.copy().clean(atol=0)
+        return super(LinearDict, snapshot).items()
```
Strilanc: `super()`

cirq/value/linear_dict_test.py
```diff
@@ -0,0 +1,439 @@
...
+
+@pytest.mark.parametrize('terms, bool_value', (
+    ({}, False),
+    ({'X': 0}, False),
+    ({'Z': 1e-12}, True),
+    ({'Y': 1}, True),
+))
+def test_bool(terms, bool_value):
+    linear_dict = cirq.LinearDict(terms)
+    if linear_dict:
```
Strilanc: `assert bool(linear_dict) == bool_value`

viathor: Done.

cirq/value/linear_dict.py
viathor: This would pick up `self` instead of `snapshot`.

(This would ultimately lead to `RuntimeError` being raised when one iterates over a LinearDict and happens to set a coefficient to zero (since this removes the element from the underlying dictionary). Confirmed by making the change and seeing the corresponding unit test fail.)

cirq/value/linear_dict.py
viathor: See below.

cirq/value/linear_dict.py
viathor: See below.

cirq/value/linear_dict.py
viathor: Done.

PR #1441, opened 2019-03-14
Pauli expansion for Fredkin and Toffoli gates

viathor:
Missed these gates in #1390.


viathor:
Yes, they do. For additional corroboration, I derived these on paper instead of using the machinery in cirq. For Fredkin, derived expansion failed the tests, but I was able to track it down to a silly error on paper. Also, learned about [Walsh functions](https://en.wikipedia.org/wiki/Walsh_function) in the process :-)

PR #1446, opened 2019-03-15
Call escape_text_latex before creating qcircuit qubit name

andbe91:


Strilanc:
I like the idea behind this change, but I'm a bit cautious about how this would generalize.

For example, what if the name is `a_2` instead of `a2`? Or `a_2_3`? Or `a_2^3`? Do we also want those cases to work? Another route entirely would be to define a `LatexQubit` with a latex string name that was not escaped during conversion.

andbe91:
I see what you mean. I actually just tested what happens in the master branch if you name a qubit `a_2`. It of course just generates `\lstick{\text{a_2}}`, however this is not valid latex since `a_2` wont work in text mode. To solve this we could remove the `\text{}` around the qubit name, which also means that things like `a_2^3` will work automatically. However, maybe there is a good reason why `\text` is there? If we remove `\text`, we probably don't want the substitution form this PR.

In general, there will be a problem if a qubit has a name which is not valid latex. I don't really see how to avoid this without using some kind of latex linter.

Strilanc:
Oh, I didn't realize we weren't calling `escape_text_for_latex` on the `str(qubit)` before handing it off. The code should definitely be modified to do that.

andbe91:
I will just implement `return r'\lstick{' + escape_text_for_latex(str(qubit)) + '}&'` for now. And then we could add the LatexQubit if we really want latex code.

Issue #887, opened 2018-09-03
Feature request: classical registers


vtomole:
Intermediate representations like QASM have `creg c0[1];` to initialize classical registers and `measure q[0] -> c0[0];` for assigning measurement results to classical memory. Are there equivalent procedures in Cirq?

bryano:
My understanding is that there was a conscious design decision to not support this because current hardware can't do intermediate measurements.

Strilanc:
@bryano is correct. We don't know what the limitations of feedforward will actually be, so for now we've avoided codifying it. The closest thing we have to any kind of classical processing at the moment is the invert-result mask on measurement gates.

vtomole:
Reopening based on comments by @balopat.

balopat:
Related #3231.

daxfohl:
@vtomole I think all the work in classical controls means this can be closed.

PR #1449, opened 2019-03-19
Add Quantum Teleportation to examples.

davemc84:
Very simplistic Quantum Teleporation example that might serve as an instructive file.

### DIFF
--- a/examples/quantum_teleportation.py
+++ b/examples/quantum_teleportation.py
@@ -1,0 +1,90 @@
+"""Quantum Teleportation.
+Quantum Teleportation is a process by which a quantum state can be transmitted
+by sending only two classical bits of information. This is accomplished by
+pre-sharing an entangled state between the sender and the receiver. This
+entangled state allows the receiver of the two classical bits of information
+to possess a qubit with the same state as the one held by the sender.
+
+In the following example output, qubit 0 is set to a random state by applying
+X and Y gates. By sending two classical bits of information after qubit 0 and
+qubit 1 are measured, the final state of qubit 2 will be identical to the
+original random state of qubit 0. This is only possible given that an
+entangled state is pre-shared between the sender and receiver.
+
+=== REFERENCE ===
+https://en.wikipedia.org/wiki/Quantum_teleportation
+https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.70.1895
+
+=== EXAMPLE OUTPUT ===
+Circuit:
+0: ---X^0.25---Y^0.125-----------@---H---M-------@---
+                                 |               |
+1: ----------------------H---@---X-------M---@---|---
+                             |               |   |
+2: --------------------------X---------------X---@---
+
+Bloch Sphere of Qubit 0 After Random X and Y Gates:
+x:  0.2706 y:  -0.7071 z:  0.6533
+
+Bloch Sphere of Qubit 2 at Final State:
+x:  0.2706 y:  -0.7071 z:  0.6533
+
+"""
+
+import random
+import numpy as np
+import cirq
+
+
+def make_quantum_teleportation_circuit():
+    circuit = cirq.Circuit()
+    q0, q1, q2 = cirq.LineQubit.range(3)
+
+    # Creates a random state for q0
+    circuit.append([cirq.X(q0)**random.random(), cirq.Y(q0)**random.random()])
+    # Creates Bell State to be shared on q1 and q2
+    circuit.append([cirq.H(q1), cirq.CNOT(q1, q2)])
+    # Bell measurement of q0 (qubit to be teleported) and q1 (entangled qubit)
+    circuit.append([cirq.CNOT(q0, q1), cirq.H(q0)])
+    circuit.append([cirq.measure(q0), cirq.measure(q1)])
+    # Uses q0 and q1 to perform operation on q2 to recover the state of q0
+    circuit.append([cirq.CNOT(q1, q2), cirq.CZ(q0, q2)])
+
+    return circuit
+
+
+def main():
+    state = []
+    circuit = make_quantum_teleportation_circuit()
+
+    print("Circuit:")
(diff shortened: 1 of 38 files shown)


davemc84:
As a heads-up, although the following line in [quantum_teleportation.py](https://github.com/quantumlib/Cirq/blob/f47a81f5a3140b0b1f9087343eed620edba9e620/examples/quantum_teleportation.py) passes the Travis CI tests:

`(cirq.bloch_vector_from_state_vector(step.state_vector(), 0))`

That line needs to be revised to the following for it to run correctly:

`(cirq.bloch_vector_from_state_vector(step.state(), 0))`

Strilanc:
What do you mean it needs to be revised? Why isn't it revised already?

examples/quantum_teleportation.py
```diff
@@ -0,0 +1,91 @@
...
+    print("x: ", np.around(b0X, 4),
+          "y: ", np.around(b0Y, 4),
+          "z: ", np.around(b0Z, 4))
+
+    # Records the final state of the simulation
+    final_results = sim.simulate(circuit)
+
+    print("\nBloch Sphere of Qubit 2 at Final State:")
+    # Prints the Bloch Sphere of q2 at the final state
+    b2X, b2Y, b2Z = \
```
Strilanc: Put the arguments on a new line, instead of the whole call. Avoid `\`.

examples/quantum_teleportation.py
```diff
@@ -0,0 +1,91 @@
...
+
+    print("Circuit:")
+    print(circuit)
+
+    sim = cirq.Simulator()
+
+    # Records in a list each state of q0 for the simulation
+    step_results = sim.simulate_moment_steps(circuit)
+    for step in step_results:
+        state.append \
```
Strilanc: Avoid `\`.

examples/quantum_teleportation.py
```diff
@@ -0,0 +1,91 @@
...
+    circuit = cirq.Circuit()
+    q0, q1, q2 = cirq.LineQubit.range(3)
+
+    # Creates a random state for q0
+    circuit.append([cirq.X(q0)**random.random(), cirq.Y(q0)**random.random()])
+    # Creates Bell State to be shared on q1 and q2
+    circuit.append([cirq.H(q1), cirq.CNOT(q1, q2)])
+    # Bell measurement of q0 (qubit to be teleported) and q1 (entangled qubit)
+    circuit.append([cirq.CNOT(q0, q1), cirq.H(q0)])
+    circuit.append([cirq.measure(q0), cirq.measure(q1)])
```
Strilanc: Use `measure(q0, q1)` to ensure they don't get separated.

examples/quantum_teleportation.py
```diff
@@ -0,0 +1,91 @@
...
+import numpy as np
+import cirq
+
+
+def make_quantum_teleportation_circuit():
+    circuit = cirq.Circuit()
+    q0, q1, q2 = cirq.LineQubit.range(3)
+
+    # Creates a random state for q0
+    circuit.append([cirq.X(q0)**random.random(), cirq.Y(q0)**random.random()])
```
Strilanc: The message should be appended after the entanglement.

examples/quantum_teleportation.py
```diff
@@ -0,0 +1,91 @@
...
+
+"""
+
+import random
+import numpy as np
+import cirq
+
+
+def make_quantum_teleportation_circuit():
+    circuit = cirq.Circuit()
```
Strilanc: `cirq.Circuit.from_ops` would avoid a lot of `append` boilerplate.

examples/quantum_teleportation.py
```diff
@@ -0,0 +1,91 @@
...
+    q0, q1, q2 = cirq.LineQubit.range(3)
```
Strilanc: `msg, alice, bob = `

examples/quantum_teleportation.py
davemc84: Absolutely - will do!

examples/quantum_teleportation.py
davemc84: Got it - understood. I have made this change.

examples/quantum_teleportation.py
davemc84: Thanks very much for this feedback. Change has been incorporated.

examples/quantum_teleportation.py
davemc84: Great thought. I have made this change and have amended the associated text accordingly.

examples/quantum_teleportation.py
davemc84: Got it - made this change as well. Thanks so much!

examples/quantum_teleportation.py
davemc84: Thanks very much for this feedback. Please let me know if I misunderstood your comment, but I have amended the code to be as follows: `message = sim.simulate(cirq.Circuit.from_ops([cirq.X(0)**ranX, cirq.Y(0)**ranY]))` which avoids the `append` boilerplate to get the bloch sphere of the message.

Again, please do let me know if I misunderstood what you were looking for here and thanks so much for this suggestion!

davemc84:
> What do you mean it needs to be revised? Why isn't it revised already?

My apologies that my earlier comment was unclear. This code has been removed in the latest version, so this is no longer an issue.

In case it is helpful, the point I was trying to make is that Travis CI liked `step.state_vector()`, but that did not seem to work when I tried to run the code, with this working for me instead: `step.state()`. As such, I was unsure which I should use. That said, as I mentioned above, this has been entirely rewritten, so is no longer an issue in the actual code.

Thanks again for all of your help - I really do appreciate it!

vtomole:
Tagging: https://github.com/quantumlib/Cirq/issues/887

PR #1450, opened 2019-03-20
Put failure information in AssertionError's message, instead of printing

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/828

### DIFF
--- a/cirq/linalg/diagonalize_test.py
+++ b/cirq/linalg/diagonalize_test.py
@@ -134,6 +134,12 @@ def test_diagonalize_real_symmetric_matrix_fails(matrix):
     with pytest.raises(ValueError):
         _ = cirq.diagonalize_real_symmetric_matrix(matrix)
 
+def test_diagonalize_real_symmetric_matrix_assertion_error():
+    with pytest.raises(AssertionError):
+        matrix =  np.array([[0.5, 0], [0, 1]])
+        m = np.array([[0, 1], [0, 0]])
+        p = cirq.diagonalize_real_symmetric_matrix(matrix)
+        assert_diagonalized_by(m, p)
 
 @pytest.mark.parametrize('s,m', [
     ([1, 1], np.eye(2)),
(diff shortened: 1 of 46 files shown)


cirq/linalg/diagonalize_test.py
```diff
@@ -68,53 +68,38 @@ def random_bi_diagonalizable_pair(
     b = cirq.dot(u, z, v)
     return a, b
 
+def _get_assert_diagonalized_by_str(m, p, d):
+    # coverage: ignore
+    return 'm.round(3) : %f, p.round(3) : %f, ' \
```
Strilanc: Prefer `.format(...)` to `%`, just to be consistent with the rest of the code base.

cirq/linalg/diagonalize_test.py
```diff
@@ -68,53 +68,38 @@ def random_bi_diagonalizable_pair(
...
-        print(np.round(m, 3))
-
-        print("p.round(3)")
-        print(np.round(p, 3))
-
-        print("np.log10(np.abs(p.T @ m @ p)).round(2)")
-        print(np.log10(np.abs(d)).round(2))
-
-        raise
+    # coverage: ignore
```
Strilanc: Why is this ignoring coverage?

vtomole: ` _get_assert_diagonalized_by_str(m, p, d)` is not called if the `assert` passes, so `coverage` never picks this line in that case.

Strilanc: There should be a test that triggers the failure case. If there isn't, we should add one.

```python
with pytest.raise(AssertError, match="bla bla"):
    ...
```

vtomole: No entirely sure about this. Lemme remove and see.

Strilanc: It's possible that the old version didn't have one because the `print` lines would end up in the output. This is part of the benefit of switching to the `assert bool, str` style: we can properly cover.

cirq/linalg/diagonalize_test.py
```diff
@@ -68,53 +68,36 @@ def random_bi_diagonalizable_pair(
...
-
-        print("np.log10(np.abs(p.T @ m @ p)).round(2)")
-        print(np.log10(np.abs(d)).round(2))
-
-        raise
+    assert cirq.is_orthogonal(p) and cirq.is_diagonal(d, atol=atol), \
+        _get_assert_diagonalized_by_str(m, p, d)
 
+def _get_assert_bidiagonalized_by_str(m, p, q, d):
+    # coverage: ignore
```
Strilanc: These ignore tags should also be removed.

cirq/linalg/diagonalize_test.py
vtomole: Seems like my local `pytest-and-incremental-coverage` doesn't reflect remote. Will add tests that trigger failure cases tomorrow.

Strilanc:
Just add a few more `with pytest.raises` tests to cover those lines.

Issue #1451, opened 2019-03-20
Make methods like `random_superposition` sample from the Haar measure


Strilanc:
Currently, the outputs of some of the random_x testing methods are non-uniform over the space of relevant objects. Replaced the biased sampling with an unbiased sampling method.

hsim13372:
Could I possibly work on this? 

If I didn't misunderstand, do we want to sample a Haar random pure state?
If so, I think you can do this by applying a Haar-random unitary to a product state (e.g. all-zero state), but I wasn't sure if this is the most efficient/direct way to do this.

c-poole:
I think the method you propose would work but the issue I see with it is that in order to produce a d-dimensional random quantum state you need to produce a d x d dimensional unitary. Since d is already 2^qubits, this becomes a memory hog very quickly.

hsim13372:
That's a good point. I just didn't know of a more memory-efficient method.

c-poole:
Appendix A of https://arxiv.org/pdf/quant-ph/9804024.pdf gives a method for uniformly sampling the PDF of a N-dimensional quantum state. A PDF doesn't uniquely determine a quantum state, but I would expect that putting i.i.d. uniformly distributed phases on each term should produce a uniform quantum state.

kevinsung:
@Strilanc I think for `random_superposition` you mean that it should sample from the uniform measure, which it already does.

hsim13372:
Hmm I think you're right, @kevinsung. I just checked and convinced myself that the way it's currently implemented is sampled uniformly. 

Maybe the only thing would be a minor documentation fix:

```def random_superposition(dim: int) -> np.ndarray:
    """
    Args:
        dim: Specified size returns a 2^dim length array.
    Returns:
        Normalized random array.
    """
    state_vector = np.random.standard_normal(dim).astype(complex)
    state_vector += 1j * np.random.normal(dim)
    state_vector /= np.linalg.norm(state_vector)
    return state_vector```

changing `dim` to `2**dim` in the code or changing the "2^dim" in the documentation to "dim."

Strilanc:
Oh, it's using a gaussian already? My mistake.

There *is* at least one instead of a TODO saying that code needs to be updated to sample from the Haar measure.

To close this issue, just verify and document that all these methods sample from the Haar measure.

hsim13372:
I verified and documented these methods and made a pull request. 

[...]

PR #1455, opened 2019-03-21
Verified and documented methods sampling from Haar measure.

hsim13372:
Fixes https://github.com/quantumlib/Cirq/issues/1451

- Checked to make sure methods in `cirq.testing.linalg` is sampling from the Haar measure.
- Added a few more simple tests in `cirq.testing.linalg.lin_alg_utils_test.py`

### DIFF
--- a/cirq/testing/lin_alg_utils.py
+++ b/cirq/testing/lin_alg_utils.py
@@ -42,19 +42,46 @@ def random_unitary(dim: int) -> np.ndarray:
 
 
 def random_orthogonal(dim: int) -> np.ndarray:
-    # TODO(craiggidney): Distribute with Haar measure.
-    m = np.random.randn(dim, dim) * 2 - 1
+    """Returns a random orthogonal matrix distributed with Haar measure.
+
+    Args:
+      dim: The width and height of the matrix.
+
+    Returns:
+      The sampled orthogonal matrix.
+
+    References:
+        'How to generate random matrices from the classical compact groups'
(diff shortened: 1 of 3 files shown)


kevinsung:
It looks to me to still be incorrect ...

It seems that the last line of `random_unitary` should actually be
```
    return q * (d / abs(d)) * q
```
and `random_orthogonal` should be the same as `random_unitary` except starting with a real matrix.

hsim13372:
Hmm for the `random_unitary`, I followed the reference by Mezzadri where his version was:

`from scipy import *
def haar_measure(n):
’’’’’’A Random matrix distributed with Haar measure’’’’’’
z = (randn(n,n) + 1j*randn(n,n))/sqrt(2.0)
q,r = linalg.qr(z)
d = diagonal(r)
ph = d/absolute(d)
q = multiply(q,ph,q)
return q`

I believe this scipy `multiply` routine's third argument is for the location into which the result is stored so q = q*ph.

For `random_orthogonal`, I removed the multiplication by `ph = d/absolute(d)` (as is done for the unitary case) because I thought scipy's QR decomposition for a real matrix already ensures that the diagonals of R are positive. I could be wrong though!

hsim13372:
Sorry, I take back the part about assuming positive diagonals for R. It seems like I should multiply by `ph = d/absolute(d)`. @Strilanc, is it possible to add to this pull request, or should I start a new one?

Strilanc:
@hsim13372 You have to create a new pull request.

PR #1464, opened 2019-03-23
Add 'update branch' button workaround to auto merge script

Strilanc:
Basically this has CirqBot actually loading the web page and clicking on the button, because the github api doesn't support this functionality.

How it's used:

1. In a browser, you login as the robot account. (I didn't want to implement code to run a two factor auth flow.)
2. Get the 'user_session' cookie out of your browser, and export it into your bash environment.
3. Clear github cookies out of your browser (don't hit the log out button!)
4. Run auto merge script as normal. Robot account will use the user session cookie to hit the button.


Strilanc:
@vtomole Cheng is looking into using Prow; the system kubernetes uses.

dev_tools/auto_merge.py
```diff
@@ -485,6 +486,59 @@ def delete_comment(repo: GithubRepository, comment_id: int) -> None:
...
+                                 ) -> Union[bool, CannotAutomergeError]:
+    session_cookie = os.getenv('CIRQ_BOT_UPDATE_BRANCH_COOKIE')
+    if session_cookie is None:
+        return attempt_sync_with_master(pr)
+
+    # Get the pull request page.
+    pr_url = 'https://github.com/{}/{}/pull/{}'.format(
+        pr.repo.organization,
+        pr.repo.name,
+        pr.pull_id)
```
viathor: This makes me uneasy. Can we validate each field? For example:
 * for organization we could check it equals "quantumlib" and raise otherwise,
 * for repo name we could check it equals "Cirq" and raise otherwise,
 * for pull ID we can check it is a sequence of digits.

If you don't want to hard-code organization and repo name, maybe pass them in as flags to the script (and maybe use quantumlib and Cirq as defaults).

dev_tools/auto_merge.py
```diff
@@ -485,6 +486,59 @@ def delete_comment(repo: GithubRepository, comment_id: int) -> None:
...
+    # Press the update branch button.
+    data = {
+        'utf8': '✓',
+        'authenticity_token': form_guts.group(2),
+        'expected_head_oid': form_guts.group(3),
+    }
+    update_url = 'https://github.com/{}/{}/pull/{}/update_branch'.format(
+        pr.repo.organization,
+        pr.repo.name,
+        pr.pull_id)
```
viathor: Ditto

dev_tools/auto_merge.py
Strilanc: This is a utility method that is supposed to work for any repository; it is not specific to quantumlib/cirq. For example, the pr argument could refer to fork of cirq.

In actual execution, all instances of `GithubRepository` source from either the main method (which specifies cirq explicitly) or from the `remote_repo` method (which refers to forks). it so happens that `remote_repo`'s result never gets passed into a method, therefore the repo name in this method is guaranteed to refer to cirq. But I do not consider that a precondition of the method, in the same way that I would not expect `list.append` to check that I was passing in the string "cirq". Lists aren't supposed to care about that sort of thing.

PR #1465, opened 2019-03-23
Implement approx_eq in Circuit

mrwojtek:


cirq/circuits/circuit.py
```diff
@@ -146,6 +146,14 @@ def __eq__(self, other):
             return NotImplemented
         return self._moments == other._moments and self._device == other._device
 
+    def _approx_eq_(self, other: Any, atol: Union[int, float]) -> bool:
+        """See `cirq.protocols.SupportsApproximateEquality`."""
+        if not isinstance(other, type(self)):
+            return NotImplemented
+        return cirq.protocols.approx_eq(self._moments, other._moments,
+                                        atol=atol) and \
```
Strilanc: Nit: use parens instead of `\`.

cirq/circuits/circuit_test.py
```diff
@@ -109,6 +110,49 @@ def test_equality():
...
+
+    assert cirq.approx_eq(Circuit([Moment([cirq.X(a)])]),
+                          Circuit([Moment([cirq.X(a)])]))
+    assert not cirq.approx_eq(Circuit([Moment([cirq.X(a)])]),
+                              Circuit([Moment([cirq.X(b)])]))
+
+    assert cirq.approx_eq(Circuit([Moment([cirq.XPowGate(exponent=0)(a)])]),
+                          Circuit([Moment([cirq.XPowGate(exponent=1e-9)(a)])]))
+
+    assert not cirq.approx_eq(Circuit([Moment([cirq.XPowGate(exponent=0)(a)])]),
```
Strilanc: None of these tests specify the atol property. Check that increasing the tolerance makes the equality more approximate.

cirq/ops/gate_operation_test.py
```diff
@@ -73,6 +73,22 @@ def p(*q):
...
+def test_gate_operation_approx_eq():
+    a = [cirq.NamedQubit('r1')]
+    b = [cirq.NamedQubit('r2')]
+
+    assert cirq.approx_eq(cirq.GateOperation(cirq.XPowGate(), a),
+                          cirq.GateOperation(cirq.XPowGate(), a))
+    assert not cirq.approx_eq(cirq.GateOperation(cirq.XPowGate(), a),
+                              cirq.GateOperation(cirq.XPowGate(), b))
+
+    assert cirq.approx_eq(cirq.GateOperation(cirq.XPowGate(exponent=0), a),
```
Strilanc: Use `cirq.X(a)**t` form, since it is much shorter.

cirq/circuits/circuit.py
mrwojtek: Done, did you mean that?

cirq/circuits/circuit_test.py
mrwojtek: Added tests for here and in the remaining two tested files as well.

cirq/ops/gate_operation_test.py
mrwojtek: Done. I wanted to be consistent with the rest of a file.

PR #1472, opened 2019-03-27
Remove MeasurementGate.is_measurement and add replacements

dabacon:


cirq/circuits/circuit.py
```diff
@@ -1767,12 +1766,10 @@ def on_stuck(bad_op):
 
 
 def _decompose_measurement_inversions(op: ops.Operation) -> ops.OP_TREE:
-    if ops.MeasurementGate.is_measurement(op):
-        gate = cast(ops.MeasurementGate, cast(ops.GateOperation, op).gate)
-        return [ops.X(q)
-                for q, b in zip(op.qubits, gate.invert_mask)
-                if b]
-
+    if ops.op_has_gate_type(op, ops.MeasurementGate):
```
Strilanc: I would make this a method on `Operation` itself, e.g. `op.has_gate_is_instance(...)`.

I agree that it's a bit hacky, but given how often we do it it may be justified to work this way:

`isinstance(op.gate, ops.MeasurementGate)`

where `op.gate` is a property that returns None and `GateOperation` overrides this attribute to return its gate.

dabacon: I'm not following you.  Making a method in Operation that refers to something on GateOperation only seems....like a bad coupling?

Then you are advocating making gate a property on Operation?  Most of the time when you are asking if a gate is of a particular type you will need to cast because you want to use properties of that specific gate type (we avoid this when we can use a protocol, which is why they are nice).

Strilanc: I guess it does couple.

Maybe instead of `op_has_gate_type` the method should be `op_gate_of_type(T) -> Optional[T]`.

dabacon:
If we add gate to Operation, every Operation will autocomplete with a gate which often isn't anything.  Making it "optional" just annoys people who are looking for the gate for an Operation.  Maybe I'm just too java-y, but having an optional interface feels very much like a violation of the interface segregation principle.  op_has_gate_type isn't too discoverable, but the method on GateOperation is, and we could use it to call out that method.  Anyway must people will just convert to GateOperation and then proceed, this is mostly an unimportant helper, not something should be central to an interface change.

Strilanc:
The utility method `cirq.op_gate_of_type` is sufficient for me. Takes a type, returns None if the operation isn't a gate operation or the gate is not an instance of that type.

dabacon:
I can't tell what it should return from that name if it is an GateOperation with a GateType.

dabacon:
PTAL

cirq/ops/gate_operation.py
```diff
@@ -61,6 +62,9 @@ def with_qubits(self, *new_qubits: raw_types.Qid) -> 'GateOperation':
     def with_gate(self, new_gate: raw_types.Gate) -> 'GateOperation':
         return new_gate.on(*self.qubits)
 
+    def has_gate_of_type(self, gate_type: Type[raw_types.Gate]) -> bool:
```
Strilanc: I would remove this method, since it is redundant with `isinstance(op.gate, gate_type)`.

cirq/protocols/measurement_key.py
```diff
@@ -82,3 +84,12 @@ def measurement_key(
...
                     "but it returned NotImplemented.".format(type(val)))
+
+
+def is_measurement(val: Any) -> bool:
+    """Returns whether or not the given value is a measurement.
+
+    A measurement must implement the `measurement_key` protocol and have a
+    channel, as represented by the `has_channel` protocol returning true.
+    """
+    return measurement_key(val, None) is not None and has_channel(val)
```
Strilanc: Nit: newline

cirq/ops/gate_operation.py
dabacon: done

cirq/protocols/measurement_key.py
dabacon: done

PR #1475, opened 2019-03-27
Capture type of default in Pauli expansion protocol

viathor:
This enables type checker to realize that `None` cannot be returned from `cirq.pauli_expansion` unless default is set to `None`. We do it in some of the other protocols.

### DIFF
--- a/cirq/protocols/pauli_expansion.py
+++ b/cirq/protocols/pauli_expansion.py
@@ -14,11 +14,12 @@
 
 """Protocol for obtaining expansion of linear operators in Pauli basis."""
 
-from typing import Any, Dict, Optional
+from typing import Any, Dict, TypeVar, Union
 
 from cirq.linalg import operator_spaces
 from cirq.protocols.unitary import unitary
 
+TDefault = TypeVar('TDefault')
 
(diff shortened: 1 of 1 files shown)


cirq/protocols/pauli_expansion.py
```diff
@@ -32,9 +33,9 @@ def _filter_coefficients(expansion: Dict[str, complex],
 def pauli_expansion(
     val: Any,
     *,
-    default: Optional[Dict[str, complex]] = RaiseTypeErrorIfNotProvided,
+    default: Union[Dict[str, complex], TDefault] = RaiseTypeErrorIfNotProvided,
```
Strilanc: This doesn't trigger the "mutable default value" warning?

In other cases I had to specify overloads with and without the default value in order to get mypy to understand.

viathor: Re-checked and mypy is happy with this.

viathor: Re-run pip install with dev reqs to make sure I'm using correct mypy version. Mypy is still happy. As is the one run by travis, so I'm going to keep as-is.

PR #1480, opened 2019-03-28
Fixed generation of random orthogonal matrices.

hsim13372:
Fixed the routine that returns a random orthogonal matrix, making sure the diagonal elements of r are positive (as shown in 'How to generate random matrices from the classical compact groups' http://arxiv.org/abs/math-ph/0609050).

### DIFF
--- a/cirq/testing/lin_alg_utils.py
+++ b/cirq/testing/lin_alg_utils.py
@@ -70,8 +70,9 @@ def random_orthogonal(dim: int) -> np.ndarray:
         http://arxiv.org/abs/math-ph/0609050
     """
     m = np.random.randn(dim, dim)
-    q, _ = np.linalg.qr(m)
-    return q
+    q, r = np.linalg.qr(m)
+    d = np.diag(r)
+    return q * (d / abs(d))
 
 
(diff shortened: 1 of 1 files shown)


cirq/testing/lin_alg_utils.py
```diff
@@ -70,8 +70,9 @@ def random_orthogonal(dim: int) -> np.ndarray:
         http://arxiv.org/abs/math-ph/0609050
     """
     m = np.random.randn(dim, dim)
-    q, _ = np.linalg.qr(m)
-    return q
+    q, r = np.linalg.qr(m)
+    d = np.diag(r)
+    return q * (d / abs(d))
```
Strilanc: Is d guaranteed not to be zero? We could use `np.sign` instead.

viathor: I guess the assumption is that the probability of `m` being singular is negligible. IIUC, it's not strictly speaking impossible, but is very, very unlikely. Perhaps we should just assert that `abs(d) > 0`.

BTW: This concern applies in `random_unitary()` above, too.

Strilanc: Good enough

PR #1481, opened 2019-03-28
Linear combination of gates

viathor:
Follow-up to #1439. Some context on #1299. This is a step towards #491, #932 and #1037. Follow-ups:
 * linear combination of operations,
 * products (for linear combination of operations only),
 * raising Pauli combinations to integer powers,
 * exponentiation.

I'm a little unhappy about the name. It's long and makes type annotations unwieldy. Suggestions?


cirq/ops/linear_combinations.py
```diff
@@ -0,0 +1,67 @@
...
+import numpy as np
+
+from cirq import protocols, value
+from cirq.ops import raw_types
+
+
+class LinearCombinationOfGates(value.LinearDict[raw_types.Gate]):
+    """Represents linear operator defined by a linear combination of gates.
+
+    Note that the operator may not be unitary or even normal. Examples of
```
Strilanc: This doesn't explain what the object is; it's just repeating the same words.

Say explicitly that this thing is supposed to be the sum of the matrices defining the gates.

cirq/ops/linear_combinations.py
```diff
@@ -0,0 +1,67 @@
...
+    operators that may be represented as LinearCombinationOfGates:
```
Strilanc: Users may go looking for how to do this in cirq; we don't have it yet.

cirq/ops/linear_combinations.py
```diff
@@ -0,0 +1,67 @@
...
+     * creation and annihilation operators.
+    """
+    def __init__(self, terms: Mapping[raw_types.Gate, value.Scalar]) -> None:
+        """Initializes linear combination from a collection of terms.
+
+        Args:
+            terms: Mapping of gates to coefficients in the linear combination
+                being initialized.
+        """
+        super().__init__(terms, validator=self._is_compatible)
```
Strilanc: I'm always surprised that this works in python. In C#, referring to a method on `self` before the parent class has been initialized is a compile time error.

cirq/ops/linear_combinations.py
```diff
@@ -0,0 +1,67 @@
...
+                self.num_qubits() == gate.num_qubits())
+
+    def matrix(self) -> np.ndarray:
+        num_qubits = self.num_qubits()
+        if num_qubits is None:
+            raise ValueError('Unknown number of qubits')
+        num_dim = 2 ** num_qubits
+        result = np.zeros((num_dim, num_dim), dtype=np.complex128)
+        for gate, coefficient in self.items():
+            result += protocols.unitary(gate) * coefficient
```
Strilanc: Some gates aren't unitary. Is this class only for unitary gates? Maybe that should be in the name of the class, or in the name of this method e.g. `sum_of_unitaries` or `total_matrix` or `combined_matrix` or `combined_matrix_from_unitaries`...

@dabacon Is there something reasonable we can do when the user passes in a general channel?

viathor: I think for linear combinations with non-unitary terms, we should just raise an exception. Present code does this implicitly.

cirq/ops/linear_combinations.py
viathor: Removed.

cirq/ops/linear_combinations.py
viathor: Clarified.

cirq/ops/linear_combinations.py
viathor: Well, in python even typos in method names aren't errors until you try to execute them. Compared to that full initialization of the target object seems like a minor concern.

cirq/ops/linear_combinations.py
Strilanc: We could include `has_unitary` in our validator.

cirq/ops/linear_combinations.py
```diff
@@ -0,0 +1,80 @@
+# Copyright 2018 The Cirq Developers
```
dabacon: nit: 2019

cirq/ops/linear_combinations.py
```diff
@@ -0,0 +1,80 @@
...
+                self.num_qubits() == gate.num_qubits())
+
+    def matrix(self) -> np.ndarray:
+        num_qubits = self.num_qubits()
+        if num_qubits is None:
+            raise ValueError('Unknown number of qubits')
+        num_dim = 2 ** num_qubits
+        result = np.zeros((num_dim, num_dim), dtype=np.complex128)
+        for gate, coefficient in self.items():
+            result += protocols.unitary(gate) * coefficient
```
dabacon: We've been loosy goosy with dtypes.  I think you would want to make sure that unitary has type that is compatible with summing against a complex128.

cirq/ops/linear_combinations_test.py
```diff
@@ -0,0 +1,145 @@
+# Copyright 2018 The Cirq Developers
```
dabacon: nit: 2019

cirq/ops/raw_types.py
```diff
@@ -134,6 +135,37 @@ def on(self, *qubits: Qid) -> 'gate_operation.GateOperation':
...
         return gate_operation.GateOperation(self, list(qubits))
 
+    def _wrap_in_linear_combination(
+            self,
+            coefficient: Union[complex, float, int]=1
+            ) -> 'linear_combinations.LinearCombinationOfGates':
+        from cirq.ops import linear_combinations
+        return linear_combinations.LinearCombinationOfGates({self: coefficient})
+
+    def __add__(self,
```
dabacon: probably gate doc deserves a note that gates can be added, multiplied by numeric, etc.

cirq/ops/linear_combinations.py
viathor: We have indeed.

I've chosen `np.complex128` assuming that it's safe against all types we use in practice, i.e. that all our unitaries can be promoted to complex128 (and hence can be added into a complex128 array). Tests I did suggest this is the case. Am I missing a type which isn't safe?

An alternative would be to use [`np.result_type()`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.result_type.html#numpy.result_type) to explicitly determine the output type, but AFAICT this will never be a type that cannot be promoted to `np.complex128` anyway.

cirq/ops/raw_types.py
viathor: Done.

cirq/ops/linear_combinations_test.py
viathor: Done.

cirq/ops/linear_combinations.py
viathor: I think it should be legal to create linear combinations of gates without unitaries. However, a user of such a linear combination accepts that `matrix()` raises an exception. Added comment to make this contract explicit.

cirq/ops/linear_combinations.py
viathor: Done.

cirq/ops/linear_combinations.py
Strilanc: We can't properly ask for a dtype without also adding it as a parameter to `cirq.unitary`. But that would make it harder for people to implement it..

dabacon: I think the strange case is if someone returns complex256 for unitary?

Maybe protocols.unitary should always upconvert to complex128?

Issue #1163, opened 2018-11-15
Make `unitary`, `has_unitary`, and `apply_unitary` fallback to `decompose`


Strilanc:
This will make them consistent with how the simulators currently work, and probably better represents the user's intent (i.e. "I have a thing I don't know the function of, please do your best to tell me what it does").

Note that this will naturally result in lists of unitary operations having a unitary matrix according to cirq.

We might need to add an optional `qubit_order` value to the `unitary` method to make this work cleanly. If not specified, the order should be inferred (e.g. from operations' qubits property).

As part of this issue, a lot of special case "decompose to get unitary" code strewn around the code base (e.g. in simulators) should be rewritten.

dabacon:
I'm down with this but think we need to make sure we have pinned down what decompose actually means first #930

andbe91:
I can work on this. However, I will maybe need some guidance. 

If I understand correctly, `unitary`  (`has_unitary`) should first try to decompose the object instead of raising `NotImplemented` (returning `False`)? However, if decompose fail, it should still raise  NotImplemented (return False)? 

I can't find a `apply_unitary_to_tensor` method anywhere in the code base.

I will probably get back to you about suitable test cases and the final task of rewriting "decompose to get unitary" in e.g. simulators.

kevinsung:
`apply_unitary_to_tensor` was renamed to `apply_unitary`.

Yes, these methods should try to obtain a unitary effect through decomposition before giving up.

Strilanc:
You are going to run into the snag "what order are the qubits in?" when attempting to reassemble multiple little unitaries into one larger unitary. This might be hard enough to deal with that tagging the issue "good first issue" was a bad idea.

An acceptable compromise for this issue is to condition the decomposition fallback on the object being a `cirq.Operation`, because operations have a `qubits` property that returns a list that corresponds to the qubit order.

For unapplied gates things are significantly trickier, because you need to know how many qubits but there's... not really a way to do that. We will open up a follow-up issue to resolve that after this one is done.

> unitary (has_unitary) should first try to decompose the object instead of raising NotImplemented (returning False)? However, if decompose fail, it should still raise NotImplemented (return False)?

Correct. Note that you should use `cirq.decompose_once` instead of `cirq.decompose`, because you will be calling `unitary` / `apply_unitary` on the items. These calls will handle further decompositions if needed.

> I can't find a apply_unitary_to_tensor method anywhere in the code base.

It was renamed to `apply_unitary`.

[...]

PR #1483, opened 2019-03-29
 Decompose gates in unitary and has_unitary

andbe91:
Implements the gate part of #1163.

The broken test is due to infinite recursion when calling `has_unitary` on either for these two gates:

    class OtherX(cirq.SingleQubitGate):
        # coverage: ignore
        def _unitary_(self) -> np.ndarray:
            return np.array([[0, 1], [1, 0]])

            return OtherOtherX().on(*qubits)

    class OtherOtherX(cirq.SingleQubitGate):
        # coverage: ignore
        def _unitary_(self) -> np.ndarray:
            return np.array([[0, 1], [1, 0]])

            return OtherX().on(*qubits)

Since none of the gates have `_has_unitary_`, we try to decompose the gates, but they are each others decomposition. In this particular case, I can solve it by falling back to `return unitary(val, None) is not None` before trying to decompose, but I don't know if that is desired? And it only solves the problem when the gates have `_unitary_` methods. Do you have any other ideas for how to solve this @Strilanc?


andbe91:
Done.

dabacon:
I believe this fixes #1163 will close that but re-open if that is not correct.

PR #1484, opened 2019-03-30
Remove MultiQubitGate

### LINKED ISSUE #1461: Get rid of MultiQubitGate
Now that `Gate` implements a `validate_args` that checks the qubit count, `MultiQubitGate` does nothing but replace `def num_qubits(self): return X` with `super().__init__(X)`, which is not enough to justify its existence.

Find all cases where `MultiQubitGate` is used and replace them with inheritance from just `Gate` with a defined qubit count.

balopat:
Fixes #1461.

Note: I'd think this is a bit of marginal benefit - so I opened this as a draft PR first to give another last chance for discussion - but Draft PR doesn't run Travis tests so please consider it Draft :) 

- this increases the boilerplate for users of Gate - even though small, but still annoying/error-prone, and introduces friction for someone who wants to design a new gate as there is no explicit direction/pattern given around it (we can fix that with docs to some extent)
- because every Gate _is_ a subclass of `raw_types.Gate`, we still have the issue of hard to distinguish between "scalable" multi-qubit gates vs fixed gates (1/2/3/.../n qubit gates). The lack of dependent types in python creates this issue and that we conflate 3+ qubit gates with gates designed to be scalable (dependent on qubit count). One idea/direction could be to introduce `ScalableGate` (similar/same as current multiqubitGate), and generate the types for `GateOn2Qubits`,`GateOn3Qubits`, ... `GateOn100Qubits` so that gate authors can signal the design intention of "this gate was design to work with only 4 qubits" easily for their users in the type hierarchy rather than in an overriden method somewhere in the body of the class. See an alternative implementation in https://github.com/quantumlib/Cirq/pull/1485

I wish python had dependent types.


c-poole:
If someone wants to make a fixed N-qubit gate for N>3, then I think it is fine to make them write a `num_qubits` method rather than having a `MultiQubitGate` class that generates the method for them.  The `num_qubits` method in `Gate` is an `abstractmethod`, so a user making a gate like this should know what went wrong if they omit this and try make an instance of their gate based on the errors they receive. 

The more interesting case is how we want to handle gates like Identity and Measurement, which can be applied to a variable number of qubits. In practice, a N-qubit Identity gate is equivalent to N single-qubit identities applied in parallel. We have a `ParallelGateOperation` class that handles that sort of scenario at the `Operation` level for single-qubit gates and also have the `on_each` method that produces the equivalent list of `GateOperation`s. 

How we want to approach describing these sorts of gates depends on what information we want `Gate`s to know and what information we want `Operation`s to know. We have in the past decided that `Gate`s need to know how many qubits they act on and `Operation`s need to know the set of qubits the gate is applied to. These variable qubit gates are a sort of "Compound Gate" which are just parallel applications of several associated "Base Gates". "Compound Gates" seem to live in a bit of a grey area in that they know how many qubits they act on in total and can also know how many qubits the associated "Base Gate" acts on, but they don't know the actual qubit set that they act on. I can't think of a compelling reason why we would need the `Gate` level classes to have this sort of flexibility. It makes more sense to promote having every `Gate` class be a description of a "Base Gate" and to let the `Operation` level classes deal with their compound forms. For the case of the Identity and Measurement gates, this would mean making them `SingleQubit` gates and letting users call the `on_each` method or the `ParallelGateOperation` constructor to get an `Operation` that applies them to numerous qubits.

balopat:
Thank you for this context, @c-poole! 
Fixed count gates: I can be convinced that just using an abstract num_qubits method is sufficient - the `FixedQubitCountGate` idea would just add one more layer of convenience where you can do the below dynamic superclass, and query for it in higher level algorithms (is this gate an instance of a fixedqubitcountgate?): 
```
class GoodGateNoPauliExpansion(cirq.FixedQubitCountGate(4)):
   ...
```
Variable count qubit gates: The CompoundGate case is definitely interesting! I have two thoughts there:
- I wonder if the CompoundGate use case is just a special case for variable count qubit gates and whether there are other multi-qubit cases: For example Qft and Adder in the examples folder are "scalable" but not necessarily compounded from independent operations that can be parallelized. (Maybe these shouldn't be a subclass of `Gate` but something else, e.g. `Algorithm`?)
- Identity and Measurement gate being forced to be a SingleQubitGate and applied to more qubits only through Operations level would be inconsistent with the rest of the model and hence probably introduce a bit of surprise for the users - but I can see the validity of it - especially if _all_ variable qubit gates fall into this category (i.e. my first point is not valid)

c-poole:
I'm in favor of encouraging code to check what `num_qubits` outputs instead of using `isinstance` or `type` information to make decisions based on how many qubits a gate acts on. I find it really difficult to come up with a compelling scenario where knowing that the gate you are working with is a `ScalableGate` or a `NQubitGate` for some fixed N gives you any more information than if you just asked the gate how many qubits it acts on. `ScalableGate` sounds like a pseudonym for "number of qubits unknown until run-time gate", which is essentially equivalent to our current implementation of `MultiQubitGate`.  It's worth noting that the `Operation` level classes sanity check whether the qubit list and gate that form the operation are consistent, which should catch errors at the Operation construction stage rather than the circuit simulation stage.
There are other cases of Gates that only know the number of qubits they act on at run time in addition to `CompoundGate`s. `ControlledGate`s with run-time determined controls is one, and the Gates in Examples are another (although these seem to be much more like Gate Factories since they are designed to be decomposed). It seems more appropriate to introduce `gate_features` that address special cases of concern such as `CompoundGate`, `ControlledGate` , or `Gate_Factory` that bring in additional utility beyond elminating the `num_qubits` boilerplate rather than trying to make a catch-all `ScalableGate` that we then need to distinguish from "Fixed N not equal to 1, 2, or 3 gate".

dabacon:
Re: the `IdentityGate`, there is at least some use for multi-qubit gates.  For example this will come up when you use linear combinations of gates.  

I think making people implement the abstract methods as is done here is good.

balopat:
Thank you for the input from both of you - it looks like the abstract method direction then is the simplest compromise for now. I'll close the other PR, and remove the WIP notice from this one, and we can merge it.

PR #1485, opened 2019-03-30
[DO NOT MERGE - discuss] scalable gate + fixed number qubit gates

balopat:
This is an alternative for #1484,  #1461.
Unfortunately, mypy doesn't fully support dynamically created classes so I had to use type annotations. 
Also type creation needs string over unicode for the first argument - this is what Travis is failing on now - as the python2.7 transpiler [forces](https://github.com/quantumlib/Cirq/blob/master/dev_tools/python2.7-generate.sh#L108-L111) unicode  - I'd need an advice here if we go down this route.


cirq/testing/consistent_pauli_expansion_test.py
```diff
@@ -39,9 +39,7 @@ def _unitary_(self) -> np.ndarray:
         return np.eye(2)
 
 
-class GoodGateNoPauliExpansion(cirq.MultiQubitGate):
-    def __init__(self) -> None:
-        super().__init__(num_qubits=4)
+class GoodGateNoPauliExpansion(cirq.FixedQubitCountGate(4)):
```
balopat: this is the "cool" part about this construct - it's kind of a dependent type implementation. However we might want to generate and "cache" the first 10 types upfront - I'm not sure about the cost of creating the same type over and over again.

balopat:
Closing in favor for #1484

Issue #1460, opened 2019-03-22
When a method expects qubits as arguments, verify that it received qubits


c-poole:
Currently `on_each` in `SingleQubitGate` does this and then calls `on` in `Gate` once it is content that is has been given valid arguments. This verification should be moved to `on` or `validate_args` in `Gate.` This is particularly relevant since the signature for `on` was recently changed from expecting a list of qubits to an unpacked list of qubits. I would push for `validate_args` since it covers the most cases of user error. If we are going to perform this explicit type-checking anywhere, it makes the most sense in `validate_args`.
 
Consider the following example:
`cirq.X.on(*cirq.LineQubit.range(3))`
`cirq.X.on_each(*cirq.LineQubit.range(3))`
The second line is a valid way to get an `op_tree` of 3 operations with different qubits all acted on by an X gate. The first will raise an error because cirq.X is a `SingleQubitGate` and will complain about being applied to 3 qubits.

Now consider this code:
`cirq.X.on(cirq.LineQubit.range(3))`
`cirq.X.on_each(cirq.LineQubit.range(3))`
The second line raises an error because the list of qubits we passed in isn't a qubit. The first line doesn't do this because it doesn't check that it received qubits and instead only checks that the number of arguments it received is consistent with the number of qubits the gate acts on. Instead the user gets a nonsensical `GateOperation` object that will give them debugging issues down the line. In fact, `cirq.GateOperation(cirq.X, "Q")` will also not complain when it should.

Strilanc:
So basically this amounts to modifying `GateOperation` to verify that the qubits it was given are in fact instances of `Qid`.

c-poole:
Pretty much. I think `Gate.validate_args` is the right place since almost everything that doesn't already perform explicit type-checking on qubit arguments already defer to it. Overridden `validate_args` methods will need to have super calls added to them as well. Most of the work will probably be in fixing tests that pass non-qubits into methods that defer to `validate_args`.

Strilanc:
Yeah, we should be more consistent about calling the parent method when overriding it.

timothyman:
Can I give this a shot?

vtomole:
Go for it!

PR #1486, opened 2019-03-30
Verifying if input are qubits in validate args

timothyman:
- Moved the verification if the input to the `on_each` method of the `SingleQubitGate` is a `Sequence[Qid]` from the `on_each` method to the `validate_args` method in the `Gate` superclass.
- Also moved an argument validation in the `on` method to `validate_args`.
- Added two test cases in gate_features_test.

Fixes: #1460.


cirq/ops/gate_features.py
```diff
@@ -108,7 +104,4 @@ def num_qubits(self) -> int:
         return self._num_qubits
 
     def validate_args(self, qubits):
-        if len(qubits) != self.num_qubits():
-            raise ValueError(
-                '{}-qubit gate was applied to {} qubits'.
-                    format(self.num_qubits(), len(qubits)))
+        super().validate_args(qubits)
```
Strilanc: Just delete the method if it forwards directly to `super()`.

timothyman: This entire class was deleted by an incoming merge from master.

PR #1494, opened 2019-04-04
Add monte carlo simulation of mixtures of sparse simulator

dabacon:
When the spares simulator hits a mixture, it now randomly selects one of these unitaries to apply.

Note that this can cause some confusion, in that similar to measurements, this adds non-determinism to the simulate* methods.


cirq/circuits/circuit.py
```diff
@@ -780,9 +780,23 @@ def findall_operations_with_gate_type(
             yield index, gate_op, cast(T_DESIRED_GATE_TYPE, gate_op.gate)
 
     def are_all_measurements_terminal(self):
+        """Whether all measurement gates are at the end of the circuit."""
+        return self.are_all_terminal(protocols.is_measurement)
+
+    def are_all_terminal(self, predicate: Callable[[ops.Operation], bool]):
```
Strilanc: `are_all_matches_terminal`

cirq/sim/sparse_simulator.py
```diff
@@ -234,36 +248,53 @@ def keep(potential_op: ops.Operation) -> bool:
...
+                                    buffer,
+                                    indices))
+                    if result is buffer:
+                        buffer = state
+                    state = result
+                elif protocols.is_measurement(op):
+                    # Do measurements second, since there may be mixtures that
+                    # operate as measurements.
+                    # TODO: support measurement outside the computational basis.
+                    meas = ops.op_gate_of_type(op, ops.MeasurementGate)
```
Strilanc: Extract each block from this big if/else-if into a method.

cirq/sim/sparse_simulator.py
```diff
@@ -93,6 +102,11 @@ class Simulator(simulator.SimulatesSamples,
         for step_result in simulate_moments(circuit):
            # do something with the wave function via step_result.state
 
+    Note also that simulations can be stochastic, i.e. return differenet results
```
balopat:
```suggestion
    Note also that simulations can be stochastic, i.e. return different results
```

dabacon: fixed

PR #1496, opened 2019-04-04
Remove fixtures from ion device test

dabacon:
Fixtures shouldn't be called directly.  Here the fixture doesn't really provide any benefit over a global method.  Renamed method to avoid it being a test.

### DIFF
--- a/cirq/ion/ion_device_test.py
+++ b/cirq/ion/ion_device_test.py
@@ -20,8 +20,7 @@ import cirq
 import cirq.ion as ci
 
 
-@pytest.fixture
-def test_ion_device(chain_length) -> ci.IonDevice:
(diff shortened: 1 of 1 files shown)


dabacon:
This is generating warnings in pytest.

vtomole:
@dabacon What warnings is it generating? I can't see anything "significant" in the build logs.

dabacon:
`~/.virtualenvs/cirq/lib/python3.6/site-packages/_pytest/python.py:198: RemovedInPytest4Warning: Fixture "test_ion_device" called directly. Fixtures are not meant to be called directly, are created automatically when test functions request them as parameters. See https://docs.pytest.org/en/latest/fixture.html for more information.`

PR #1500, opened 2019-04-05
use `cirq.Symbol` instead of `sympy.Symbol` in docs tutorial

gnperdue:
I am pretty sure you need `cirq.Symbol` in the tutorial - that works for me with cirq 0.4.0 while the `sympy.Symbol` throws errors about type coercion.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -503,11 +503,10 @@ Luckily for us, we have written our code so that using parameterized
 values is as simple as passing `Symbol` objects where we previously
 passed float values.
 ```python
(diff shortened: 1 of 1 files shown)


vtomole:
The docs on mainline are based on the features that we have on master. We replaced `cirq.Symbol` with `sympy.Symbol` after the 0.4.0 release . Please reference https://cirq.readthedocs.io/en/stable/ if you are on 0.4.0.

gnperdue:
Ah, okay, sorry for the noise!

PR #1508, opened 2019-04-09
Make PauliString powers produce a PauliStringPhasor

Strilanc:
- move PauliStringPhasor out of contrib
- refactor PauliStringPhasor to have separate phase exponents for both eigenspaces
- Implement `__pow__` and `__rpow__` and numpy's exp method on PauliString
- Single-qubit Pauli strings turn into X/Y/ZPowGate when raising/exponentiating

~~Comes after https://github.com/quantumlib/Cirq/pull/1504~~
Part of https://github.com/quantumlib/Cirq/issues/1037


cirq/ops/pauli_string_test.py
```diff
@@ -346,6 +436,31 @@ def test_pos():
...
+def test_pow():
+    a, b = cirq.LineQubit.range(2)
+
+    assert cirq.PauliString({a: cirq.X})**0.25 == cirq.X(a)**0.25
+    assert cirq.PauliString({a: cirq.Y})**0.25 == cirq.Y(a)**0.25
+    assert cirq.PauliString({a: cirq.Z})**0.25 == cirq.Z(a)**0.25
+
+    p = cirq.PauliString({a: cirq.X, b: cirq.Y})
+    assert p**1 == p
+    assert p**-1 == p
```
vtomole: Unrelated: Could you please tell me why this is the case?  Why is the inverse of a PauliString acting like an identity operation?

Strilanc: It's not acting like 1, it's acting like -1. (-1)**(-1) == -1

Strilanc:
@dabacon it phases the two eigenspaces independently. It can represent more things than exponentiation.

cirq/ops/pauli_string_test.py
vtomole: If `==-1`, why is `p**-1 == p` ? I think i'm missing a `-1` somewhere, but i can't see it from the `p` that's been defined here.

Strilanc: What's the inverse of the X gate? The X gate. X**-1 == X. This is true of Pauli products more generally.

vtomole: Ah! Thanks.

cirq/ops/pauli_string_phasor.py
```diff
@@ -0,0 +1,180 @@
...
+import sympy
+
+from cirq import value, protocols
+from cirq.ops import (raw_types, common_gates, pauli_string as ps, pauli_gates,
+                      op_tree, pauli_string_raw_types)
+
+
+@value.value_equality(approximate=True)
+class PauliStringPhasor(pauli_string_raw_types.PauliStringGateOperation):
+    """An operation that phases a Pauli string."""
```
dabacon: explain what "phases" means.  It is definitely not something most people in quantum computing will understand.

cirq/ops/pauli_string_phasor.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def __init__(self,
+                 pauli_string: ps.PauliString,
+                 *,
+                 exponent_neg: Union[int, float, sympy.Basic] = 1,
+                 exponent_pos: Union[int, float, sympy.Basic] = 0) -> None:
+        """Initializes the operation.
+
+        At most one angle argument may be specified. If more are specified,
```
dabacon: I don't see this in the code

vtomole: Cancelled merged cause this wasn't addressed.

cirq/ops/pauli_string_test.py
viathor: IOW, (aX + bY + Z)^2 is identity (assuming [a, b, c] is a normalized vector in R^3). You can see this by writing out the 3-by-3 table of all products in this expression. Diagonal products are squares of Paulis, i.e. identity and off-diagonal elements can be paired up to cancel out since Paulis anti-commute.

cirq/ops/pauli_string_phasor.py
vtomole: This wasn't addressed either. Was your approval premature or is this not a big deal now?

cirq/ops/pauli_string_test.py
vtomole: Thank you for the clarification.

Strilanc:
@dabacon remember to mention in the approval that the fixes are still expected so people don't accidentally merge them

cirq/ops/pauli_string_phasor.py
Strilanc: Fixed.

cirq/ops/pauli_string_phasor.py
Strilanc: Fixed.

PR #1510, opened 2019-04-09
Update pylint version from 2.1.* to 2.3.* and remove "unnecessary pass" warnings.

vtomole:


Strilanc:
What is up with travis-ci lately? Random double-notifications, this one hanging forever...

Strilanc:
Restarting one of the jobs seems to have triggered it to actually send the status completion.

verult:
Seeing the error below in Travis install step:

pylint 2.3.1 has requirement astroid<3,>=2.2.0, but you'll have astroid 2.1.0 which is incompatible.

Should astroid version be bumped too?

Strilanc:
I suspect that will trigger the parsing bug. Kind of surprised pylint managed to work with an older version.

vtomole:
@verult Bumping it will break pylint. I'm not seeing that "incompatible" warning on Travis builds. Where is it?

vtomole:
This appears when `check.sh` is run. I vote for leaving it until that astroid bug is fixed but i'm willing to revert if people think differently.

verult:
I'm seeing it here:
https://travis-ci.com/quantumlib/Cirq/jobs/193471026#L752
I'm surprised the install is shown as succeeding in the end, but hey as long as tests pass :)

OK with leaving as is too

Issue #1511, opened 2019-04-09
float128 not supported on my Windows Anaconda


vtomole:
```(base) C:\>python
Python 3.7.0 (default, Jun 28 2018, 08:04:48) [MSC v.1912 64 bit (AMD64)] :: Anaconda, Inc. on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
>>> np.array([1,1,1],dtype=np.float128)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'numpy' has no attribute 'float128'
>>>
```

@c-poole I ran into this while running pytest on my Windows. Does pytest pass for you on Windows?

vtomole:
This seems to be a win32 bug: https://github.com/winpython/winpython/issues/613

vtomole:
I suggest we replace `np.float128` with `np.longdouble` and `np.complex256` with `np.clongdouble`. These won't change anything on Linux but it will give us `np.float64` and `np.complex128` on Windows and won't break tests on Windows builds (this change also depends on how many devs we have on Windows ).

Kind of related: @Strilanc  Should we add a Windows build on our Travis?

LuisM78:
Hi, Is there a solution for this problem? I am using Windows 10. can you please point me to the right install for numpy? 
Thanks,
Luis

vtomole:
We already solved this; by not using float128. What commands are you getting to run into this error?

LuisM78:
Hi, I am using in a function.
def softmax(x):
    x = np.array(x, dtype=np.float128)
    e_x = np.exp(x)
    return e_x / e_x.sum(axis=0) 

I have np version 1.15.4

vtomole:
The snippet you posted doesn't use Cirq.

LuisM78:
yes, sorry... 
I couldn't figure out how to do it in Windows. So I switched to an Ubuntu virtual box and created a python environment with the specific pip requirements and versions ... and problem solved...
Thank you!
Luis

hafez-ahmad:
any support for windows 10 and python 3.8, I am getting the same error.
AttributeError: module 'numpy' has no attribute 'float128'

balopat:
@hafez-ahmad can you share the code you're using where you're running into this using Cirq?

hafez-ahmad:
@balopat here it is https://github.com/josuemtzmo/trackeddy/issues/9

vtomole:
[...]

Issue #1143, opened 2018-11-12
Test failures on windows machine related to numpy types being present


Strilanc:
```
cirq\circuits\circuit_test.py:1532 (test_apply_unitary_effect_to_state)
>       for dt in (np.complex64, np.complex128, np.complex256):
E       AttributeError: module 'numpy' has no attribute 'complex256'
```

```
cirq\google\sim\xmon_simulator_test.py:360 (test_run_initial_state_ndarray_not_upconvertible[None])
        with pytest.raises(TypeError):
            _ = run(simulator, basic_circuit(), scheduler,
                    initial_state=np.array([0, 0, 1, 0],
>                                          dtype=np.float128))
```

Also there's this failure related to floating point rounding error (ConvertToPauliStringPhasors and friends should default to a non-zero tolerance):

```
convert_to_pauli_string_phasors_test.py:64: 

E       AssertionError: Circuit's text diagram differs from the desired diagram.
E       
E       Diagram of actual circuit:
E       0: ───X────────────────────[Z]^0.125───
E       
E       1: ───[Y]^0.25───[Z]^0.0───H───────────
E       
E       Desired text diagram:
E       0: ───X──────────[Z]^0.125───
E       
E       1: ───[Y]^0.25───H───────────
```

dabacon:
We should check numpy  to see if the type is supported and if not, not do the test.

PR #1512, opened 2019-04-09
Only test float128/complex256 precision on systems that support it

Strilanc:
Fixes https://github.com/quantumlib/Cirq/issues/1511
Fixes https://github.com/quantumlib/Cirq/issues/1143


Strilanc:
@vtomole Could you check that the type-related failures no longer happen on your windows machine?

PR #1513, opened 2019-04-10
Adds partial trace to linalg

dabacon:


dabacon:
This will be used in making measurements for general channels that have keys (Best to take partial trace, then compute probabilities, then apply relevant single krauss operator instead of applying all possible single krauss operator calculating probabilities and then selecting one).  Also useful in general.  Will be also useful for density_matrix class in case of dense representations.

cirq/linalg/transformations.py
```diff
@@ -217,3 +217,38 @@ def fractional_swap(target):
                 out[s_i] += target[s_j] * matrix[i, j]
 
     return out
+
+
+def partial_trace(tensor: np.ndarray,
+                  keep_indices: List[int]):
```
Strilanc: Annotate output type

dabacon: done

Issue #1466, opened 2019-03-23
cirq/optimizers/decomposition_test.py fails undeterministicaly


mrwojtek:
While trying to merge #1465, one of the decomposition tests (`cirq/optimizers/decompositions_test.py::test_single_qubit_matrix_to_native_gates_fuzz_half_turns_always_one_gate`) failed on input `(0.8209689113482108, 0.32097649795628813)` with message:
```
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=1e-08
E       
E       (mismatch 25.0%)
E        x: array([ 0.+0.000000e+00j, -1.-9.533613e-05j,  1.+0.000000e+00j,
E               0.+0.000000e+00j])
E        y: array([ 0.+0.000000e+00j, -1.-4.766806e-05j,  1.+0.000000e+00j,
E               0.+0.000000e+00j])
```
This looks like a rounding problem but most likely require some investigation.

Regardless of the issue at hand, I noticed that this test generates random set of test parameters. This seems like a dangerous practice: with this project growing if we have more tests like that which fail randomly even with very low probability p, then the test run success probability degrades as (1-p)^N, which doesn't require large N to significantly impact development sanity. Is this random test here a one-time thing, or it is a desirable practice to have in this project?

Strilanc:
We do fuzz testing in several places. I don't think this is a bad practice, because the failures we see are actual failures.

We do have some probabilistic tests with false positive failures, but when we add those I try to keep the probability of statistical failure very low. Perhaps 1 in a million, informally. If we had a 1000 such tests, we would see a failure every 1000 runs or so. That is an acceptably low flake number.

The main issue with the randomized tests is that numpy gets very angry at them if you attempt to run the tests in parallel. The different threads compute different test parameters, and the inconsistency is considered to be an error.

mrwojtek:
You're probably right; it just might get more frustrating to work with the flaky project if it gets out of control. Fortunately there is re-run button in Travis.

On the other hand, if we suspect that there is something wrong maybe it's good idea to just test it extensively once, when creating the functionality? Out of pure curiosity I run the concerning test 100000 times and it failed 30 times. (Each case with the similar inconsistency at the order of 1e-4.)

Strilanc:
Fixed by #1516

PR #1516, opened 2019-04-10
Fix tolerance arguments in decomposition_test

mrwojtek:


mrwojtek:
This fixes #1466 (verified)

cirq/optimizers/decompositions_test.py
```diff
@@ -215,16 +215,18 @@ def test_single_qubit_matrix_to_native_gates_cases(intended_effect):
...
+    atol = 1e-6
+
     intended_effect = cirq.dot(
         cirq.unitary(cirq.Z**(2 * pre_turns)),
         cirq.unitary(cirq.X),
         cirq.unitary(cirq.Z**(2 * post_turns)))
 
     gates = cirq.single_qubit_matrix_to_phased_x_z(
-        intended_effect, atol=0.0001)
+        intended_effect, atol=atol)
```
Strilanc: This tolerance refers to a per-gate effect, whereas the other tolerance refers to an aggregated effect. Add a safety factor of 10 to the aggregated one.

vtomole:
@mrwojtek I ran the original test 50,000 times with no failures. How were you able to verify this?

cirq/optimizers/decompositions_test.py
mrwojtek: Sounds good, done.

mrwojtek:
Victory: I ran this test 100000 times before and after my change. It's not a proof but it failed about 30 times before fixing, and 0 times after fixing.

vtomole:
I managed to run the original 1000000 times with no errors (took 1.5 hours!). I wonder what could be the difference. Could it  different verions of numpy? (I'm on `1.16.2`). Python versions? (i'm on `3.6.7`).

mrwojtek:
Ahh! In that case this is most likely my macOS. I had similar problems with other equality comparisons for some of the Cirq tests (that's why I implemented approx_eq). It works on Linux but not on my macOS.

BTW: It took me about 10min to do the same calculation on my MacBook laptop ;-)

PR #1517, opened 2019-04-10
Make automerge script keep going when an error is encountered

Strilanc:
- Have been watching it long enough that I expect to have seen situations where this would cause continuous havoc, but haven't seen them


Strilanc:
Thanks Victory! Much faster than Dave.

Issue #1409, opened 2019-02-27
How do we make circuit translation and circuit simulation fit together in a nice way?


Strilanc:
Example technical challenges:

- I simulated a circuit, but the simulation was noisy. What noise was added?
- I optimized a circuit, and there was a problem with the output. How do I explain the problem in terms of my original circuit?
- I submitted a circuit using LineQubits to be run on hardware, and it was automatically placed onto a 2d grid. Which qubits did the placement use?
- I parameterized my circuit, and the simulator ran some concrete version. Which version did it run?

dstrain115:
@Strilanc This issue has a bunch of somewhat unrelated (and big) challenges in a very vague and unspecified way.  Some of it has been addressed (optimizers to transformers API with a proposed visualization and logging).   

I generally think this issue is too broad and too vague to make progress on.  I propose closing it as stale and creating more concrete issues, unless you have strong objections.

tanujkhattar:
Following up on the previous comment, I think we should close this issue and open more concrete issues with specific examples and use cases. I'll go ahead and close it.

PR #1520, opened 2019-04-11
Define cirq.NoiseModel and cirq.NO_NOISE/cirq.ConstantQubitNoiseMode

Strilanc:
- `cirq.NoiseModel` has semi-abstract noisy_moment/moments/operation methods (at least one must be implemented)
- `cirq.ConstantQubitNoiseModel` applies a given gate to every qubit at the end of every moment
- `cirq.NO_NOISE` is a noise model that returns its inputs unchanged
- `DensityMatrixSimulator` now has a `noise` attribute that takes a `NoiseModel`
- Simulators now force keyword args for `dtype`

Part of https://github.com/quantumlib/Cirq/issues/1355
Fixes https://github.com/quantumlib/Cirq/issues/343
Defers https://github.com/quantumlib/Cirq/issues/1409


cirq/devices/noise_model.py
```diff
@@ -0,0 +1,153 @@
+# Copyright 2018 The Cirq Developers
```
dabacon: nit: 2019

cirq/devices/noise_model.py
```diff
@@ -0,0 +1,153 @@
...
+
+from cirq import ops, value
+
+if TYPE_CHECKING:
+    # pylint: disable=unused-import
+    import cirq
+
+
+class NoiseModel:
+    """Replaces operations and moments with noisy counterparts."""
```
dabacon: Add documentation on how to use this.

dabacon: in particular the restriction on overriding.

cirq/devices/noise_model.py
```diff
@@ -0,0 +1,153 @@
...
+
+@value.value_equality
+class _NoNoiseModel(NoiseModel):
+    """A default noise model that adds no noise."""
+
+    def noisy_moments(self,
+                      moments: 'Iterable[cirq.Moment]',
+                      system_qubits: Sequence['cirq.Qid']
+                      ):
+        return list(moments)
```
dabacon: can't just be moments?

cirq/devices/noise_model.py
dabacon: also other files

vtomole:
> Simulators now force keyword args for `dtype`

Why?

Strilanc:
> Why?

Rule of thumb is that forcing keyword args is safer than not, because you can stop forcing it without breaking compatibility. I try to do it whenever there is not a clear order of arguments, or when there are very similar methods that have additional arguments where the order is not clear.

cirq/devices/noise_model.py
Strilanc: Done.

cirq/devices/noise_model.py
Strilanc: Done.

cirq/devices/noise_model.py
Strilanc: It can be, but in general I've been shying away from returning iterables because they are harder to work with (e.g. you can only iterate them once).

cirq/devices/noise_model.py
```diff
@@ -0,0 +1,156 @@
...
+    """Replaces operations and moments with noisy counterparts.
+
+    A child class must override *at least one* of the following three methods:
+
+        noisy_moments
+        noisy_moment
+        noisy_operation
+
+    The methods that are not overriden will be implemented in terms of the ones
+    that are.
```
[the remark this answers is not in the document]
Strilanc: The emphasized `*at least one*` seems pretty clear to me..

PR #1525, opened 2019-04-12
Support cirq.inverse in LinearPermutationGate

mrwojtek:


cirq/contrib/acquaintance/permutation.py
```diff
@@ -149,6 +149,16 @@ def _value_equality_values_(self):
     def __bool__(self):
         return bool(_canonicalize_permutation(self._permutation))
 
+    def __pow__(self, exponent):
+        if exponent == 1:
+            return self
+        if exponent == -1:
+            return LinearPermutationGate(
+                self._num_qubits,
+                dict((v, k) for k, v in self._permutation.items()),
```
Strilanc: `{v: k for k, v in ...}`

Issue #1526, opened 2019-04-12
Check script doesn't run format check


dabacon:

vtomole:
What do you mean?

Strilanc:
He means `continuous-integration/check.sh` doesn't do this particular check.

vtomole:
Ah! I forgot `check.sh` exists.

PR #1529, opened 2019-04-12
Fix format script half ignoring revision argument and attempting to format deleted files

Strilanc:


vtomole:
Should we knock https://github.com/quantumlib/Cirq/issues/1526 out on here while we are at it or would that be too big a PR?

Strilanc:
It's separate

PR #1532, opened 2019-04-12
Add `cirq.final_wavefunction`

Strilanc:
- Basically this is an easy-to-use version of `apply_unitary`


cirq/sim/mux.py
```diff
@@ -18,7 +18,7 @@
```
vtomole: Unrelated: why is this file call mux? Something related to mulitplexer?

cirq/sim/mux.py
```diff
@@ -52,6 +52,53 @@ def sample(program: Union[circuits.Circuit, schedules.Schedule],
             repetitions=repetitions)
 
 
+def out_vector(program: Union[ops.Operation, ops.Gate, circuits.Circuit, schedules.Schedule],
```
vtomole: What's the motivation for calling it an `out_vector`?

cirq/sim/mux.py
dabacon: +1 for maybe a better name?

cirq/sim/mux.py
dabacon: wonder if we could make this more of a verb?

cirq/sim/mux.py
```diff
@@ -52,6 +52,53 @@ def sample(program: Union[circuits.Circuit, schedules.Schedule],
...
+            ordering of the computational basis states.
+        initial_state: If an int, the state is set to the computational
+            basis state corresponding to this state. Otherwise  if this
+            is a np.ndarray it is the full initial state. In this case it
+            must be the correct size, be normalized (an L2 norm of 1), and
+            be safely castable to an appropriate dtype for the simulator.
+        dtype: The `numpy.dtype` used by the simulation. Typically one of
+            `numpy.complex64` or `numpy.complex128`.
+            Favors speed over precision by default, i.e. uses `numpy.complex64`.
+    """
```
dabacon: doc return value. note the order for the gate case.

cirq/sim/mux.py
```diff
@@ -52,6 +52,53 @@ def sample(program: Union[circuits.Circuit, schedules.Schedule],
...
+
+    if not protocols.has_unitary(program):
```
dabacon: I don't think Schedule has_unitary

cirq/sim/mux.py
```diff
@@ -52,6 +52,53 @@ def sample(program: Union[circuits.Circuit, schedules.Schedule],
...
+        raise ValueError(
+            "Given object is not unitary. "
+            "Non-unitary operations don't have "
+            "a single well defined output state vector. "
+            "Maybe you wanted `cirq.sample_wavefunction`?")
+
+    if isinstance(program, ops.Gate):
```
Ashalynd: It might look at bit more intuitive like that:

```
if isinstance(program, ops.Gate):
# convert to ops.Operation
  program = program.on(*line.LineQubit.range(program.num_qubits()))
program = circuits.Circuit.from_ops(program)
```

dabacon:
This seems to be getting stale.  Is there something blocking the requested changes?

cirq/sim/mux.py
Strilanc: It multiplexes over the simulators. It's a mux.

... Okay fine, it's not a very good name. Terrible, even.

cirq/sim/mux.py
Strilanc: This shouldn't be called "get out vector" because I want to be consistent with `cirq.unitary`, which also avoids using a verb. We are asking for a property.

Renamed to `final_wavefunction`.

cirq/sim/mux.py
Strilanc: Done.

cirq/sim/mux.py
Strilanc: Fixed

cirq/sim/mux.py
Strilanc: This does the wrong thing for schedules.

Strilanc:
@dabacon Brought it back up to date; ready for review again.

I think this is one of the key methods that we will want to have around, because it disambiguates what the user is asking for when they do a simulation.

cirq/sim/mux.py
vtomole: At the top of this file it says `Sampling/simulation methods that delegate to appropriate simulators` which makes sense cause that's what a mux means. I think this is fine if we explain the file's name in that description.

PR #1533, opened 2019-04-12
Define 'Sampler' object

Strilanc:
Common interface for hardware and simulators.


cirq/sim/sampler.py
```diff
@@ -0,0 +1,64 @@
...
+            repetitions: The number of repetitions to simulate.
+
+        Returns:
+            TrialResult for a run.
+        """
+        return self.run_sweep(program, study.ParamResolver(param_resolver),
+                              repetitions)[0]
+
+    @abc.abstractmethod
+    def run_sweep(
```
maffoo: Is there a way to handle extra, sampler-specific info here? For example, the engine sample implementation may want to allow specifying a job_config, priority, route, etc.

Strilanc: That information has to be provided when creating the sampler. The interface to call it should be unaware of details like that.

cirq/sim/sampler.py
```diff
@@ -0,0 +1,64 @@
+# Copyright 2018 The Cirq Developers
```
dabacon: nit: 2019

cirq/sim/sampler.py
```diff
@@ -0,0 +1,64 @@
...
+
+from cirq import circuits, schedules, study
+
+
+class Sampler(metaclass=abc.ABCMeta):
+    """Something capable of sampling quantum circuits. Simulator or hardware."""
+
+    def run(
+            self,
+            program: Union[circuits.Circuit, schedules.Schedule],
```
dabacon: I'm a bit worried about this not being compatible with some important patterns for engine:
1) create program
In loop
2) create job with run context and get results

If we don't separate out 1) from 2) then people will loop on 1) and 2) and this will take a hit for latency of uploading the program.

Strilanc: This interface is supposed to be identical w.r.t. the engine and the simulator, but we don't really have a concept of a saved program for the simulator. So for now I will leave it out, but we can add it later if needed.

Strilanc:
@vtomole `Sampler` covers `SimulatesSamples` but also (eventually) quantum engine and other physical hardware services. Methods taking a `Sampler` will work on both, and can be tested vs simulators.

PR #1537, opened 2019-04-16
Change engine from using route to processor_selector.

lindmarkm:


vtomole:
@lindmarkm The formatting is checked incrementally, so you only need to change the lines you've added.

cirq/google/engine/engine.py
```diff
@@ -185,15 +185,15 @@ def __init__(self,
...
+            self,
             *,  # Force keyword args.
             program: Union[circuits.Circuit, Schedule],
             job_config: Optional[JobConfig] = None,
             param_resolver: ParamResolver = ParamResolver({}),
             repetitions: int = 1,
             priority: int = 50,
-            target_route: str = '/xmonsim'
-            ) -> TrialResult:
+            processor_ids: List[str] = ['xmonsim']) -> TrialResult:
```
Strilanc: Use an immutable tuple instead of a mutable list for the default value. https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

The type `Sequence[str]` matches both.

cirq/google/engine/engine.py
```diff
@@ -314,15 +315,15 @@ def program_as_schedule(self,
...
-                  ) -> 'EngineJob':
+    def run_sweep(
+            self,
+            *,  # Force keyword args.
+            program: Union[circuits.Circuit, Schedule],
+            job_config: Optional[JobConfig] = None,
+            params: Sweepable = None,
+            repetitions: int = 1,
+            priority: int = 500,
+            processor_ids: List[str] = ['xmonsim']) -> 'EngineJob':
```
Strilanc: tuple instead of list

cirq/google/engine/engine.py
lindmarkm: Done.

cirq/google/engine/engine.py
lindmarkm: Done.

PR #1538, opened 2019-04-16
Fix bug in measurement gate channel

dabacon:
Completely wrong from >1 qubits.


cirq/ops/common_gates.py
```diff
@@ -446,12 +446,11 @@ def _measurement_key_(self):
...
-        size = 2 ** self.num_qubits()
-        zero = np.zeros((size, size))
-        zero[0][0] = 1.0
-        one = np.zeros((size, size))
-        one[-1][-1] = 1.0
-        return (zero, one)
+        def delta(i):
+            result = np.zeros((2 ** self.num_qubits(), ) * 2)
+            result[i][i] = 1
+            return result
```
Ashalynd: 1. This returns a generator instead of tuple, is that the intent?
2. Keeping the `size` allows to not repeat `2 ** self.num_qubits()` twice.

dabacon: oh good calls.  Yes made tuple.  pulled size outside.

cirq/ops/common_gates_test.py
```diff
@@ -540,16 +540,26 @@ def test_measurement_channel():
     np.testing.assert_allclose(
             cirq.channel(cirq.MeasurementGate(1)),
             (np.array([[1, 0], [0, 0]]), np.array([[0, 0], [0, 1]])))
+    # yapf: disable
```
vtomole: Why is yapf wrong?

dabacon: because formatting a big 4 by 4 array in a readable format is way more readable than cludging it together.

PR #1540, opened 2019-04-17
Add Travis build step for Windows.

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/813


Strilanc:
I am kicking myself for not thinking of this.

vtomole:
@Strilanc Should we add a Windows build? I'm personally against it because i have build time in mind. Devs will have to wait longer for things to pass. Also, Windows builds can't be run locally if on Linux. Maybe we can make this a weekly build (A feature maybe Prow has?).

CC: @verult

Strilanc:
I think it's important to test on other platforms, and that we've been guilty of neglecting it.

Strilanc:
I'm not too worried about having another check, because the builds can be run in parallel. Also, the auto merge script fixes the most serious problem of longer build times (someone having to actively poke github every 5-10 minutes when there's a glut of merges).

Note that you can see the individual build results before the whole thing has finished if you click through to travis. Unfortunately I don't think there's an option to have it split out at github, though that would be much better.

Strilanc:
Oh, I see what you mean by the duration... it takes 13 minute for it to run. Five of those minutes are just setting up the python 3 environment..

.travis.yml
```diff
@@ -59,3 +59,15 @@ matrix:
       - unzip protoc-3.5.1-linux-x86_64.zip -d protoc3.5
       - PATH=$(pwd)/protoc3.5/bin:${PATH}
     script: check/pytest2
+
+  - os: windows
+    language: sh
+    python: "3.5.2"
```
Strilanc: This says python 35, but below you're using python 37.

.travis.yml
```diff
@@ -59,3 +59,15 @@ matrix:
       - unzip protoc-3.5.1-linux-x86_64.zip -d protoc3.5
       - PATH=$(pwd)/protoc3.5/bin:${PATH}
     script: check/pytest2
+
+  - os: windows
+    language: sh
```
Strilanc: Give an `env: NAME=windows`.

.travis.yml
```diff
@@ -59,3 +59,15 @@ matrix:
...
+    python: "3.5.2"
+    before_install:
+    - choco install python3
```
Strilanc: You've confirmed this step is actually needed? I would have expected that asking for python 3.5.2 above to ensure it was installed.

vtomole: Just confirmed that choco is necessary.  The `python: "3.5.2"` displays what version is being run on the dashboard.

vtomole:
Ugh. Need to figure out how to run Travis locally...

vtomole:
Looks like everything is passing on this except for `use_unicode_characters=True` in `cirq\ion\convert_to_ion_gates_test.py` and `cirq\ion\ion_device_test.py`.

vtomole:
@Strilanc We just got the Windows build to pass on this. Please take a look. I would like to get this in before some non Windows stuff makes it into the mainline again.

cirq/contrib/contrib-requirements.txt
```diff
@@ -1,4 +1,4 @@
 # Runtime requirements for contrib.
 
 pylatex==1.3.*
-tensorflow
+tf-nightly
```
vtomole: Tensorflow for Python3.7 is currently only supported in the nightly build. I will add a comment above this line stating this case in the next commit.

cirq/protocols/approximate_equality_test.py
```diff
@@ -50,13 +50,13 @@ def test_numpy_dtype_compatibility():
         assert not cirq.approx_eq(u_type(i_a), u_type(i_c), atol=1)
 
     f_a, f_b, f_c = 0, 1e-8, 1
-    f_types = [np.float16, np.float32, np.float64, np.float128]
+    f_types = [np.float16, np.float32, np.float64]
```
vtomole: Cody has a better solution. I will revert to mainline on next commit.

requirements.txt
```diff
@@ -3,7 +3,7 @@
 google-api-python-client~=1.6
 matplotlib~=2.2
 networkx~=2.1
-numpy~=1.12
+numpy~=1.16
```
vtomole: Took this opportunity because tf-nightly is not running with older versions of numpy (got a `ModuleNotFoundError: No module named 'numpy.core._multiarray_umath'` ) when trying to import.

vtomole:
@Strilanc Last ping was premature. Cody's recent fix passes the build so i think this is good to go in.

c-poole:
This should also close #813 when merged.

Strilanc:
Thanks for doing all this, Victory. This is a huge help to Windows users over the long term.

Issue #1541, opened 2019-04-17
multiprocessing_pool tests failing on Windows


vtomole:
3 tests are failing on Windows due to multiprocessing errors. 

One of them:
```
Traceback (most recent call last):
  File "c:\python37\lib\multiprocessing\pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "c:\python37\lib\multiprocessing\pool.py", line 44, in mapstar
    return list(map(*args))
  File "C:\Users\travis\build\quantumlib\Cirq\cirq\google\sim\mem_manager_test.py", line 28, in get_shared_mem
    return mem_manager.SharedMemManager.get_array(handle)
  File "C:\Users\travis\build\quantumlib\Cirq\cirq\google\sim\mem_manager.py", line 182, in get_array
    return SharedMemManager._instance._get_array(handle)
  File "C:\Users\travis\build\quantumlib\Cirq\cirq\google\sim\mem_manager.py", line 124, in _get_array
    tup = self._arrays[handle]
IndexError: list index out of range
```
See: https://travis-ci.com/quantumlib/Cirq/jobs/193736308

PR #1549, opened 2019-04-18
Turn off multiprocessing for windows

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/1541

Will remove "Travis windows" once approved.


cirq/google/sim/mem_manager_test.py
```diff
@@ -79,6 +80,8 @@ def test_fills_gaps():
     mem_manager.SharedMemManager.free_array(new_handle)
 
 
+@pytest.mark.skipif(sys.platform.startswith('win'),
+                    reason="skipping linux-only tests")
```
dabacon: "skipping linux-only tests" makes it sound like it is skipping linux tests. 
I'd make the reason something more about why this test
"skipping test on windows due to lack of multiprocessor support"

vtomole: Done.

cirq/google/sim/mem_manager_test.py
```diff
@@ -79,6 +80,9 @@ def test_fills_gaps():
     mem_manager.SharedMemManager.free_array(new_handle)
 
 
+@pytest.mark.skipif(sys.platform.startswith('win'),
+                    reason='Skipping test on Windows due'
```
dabacon: nit space after 'due' so string doesn't say 'dueto'

vtomole: Long day! Haha

Issue #1539, opened 2019-04-17
Observables instead of Displays?


Strilanc:
I was talking with @ncrubin and @mpharrigan about displays, and they made some good points against them. Mainly it comes down to: this feature has a ton of engineering complexity, but cover a lot of use cases outside of what researchers actually need to do. We would likely be better served by, instead of a "compute_displays" method where there can be a series of displays strewn throughout the circuit, having an "estimate_observable" method that specifies a single observable we would like to apply to the output of the circuit.

So the proposal is:

- Rename Display into Observable
- Don't allow Observables to be added into circuits
- Refactor methods computing display contents to take an observable instead of searching in the circuit for displays

@dabacon @kevinsung

mpharrigan:
> to take an observable instead of searching in the circuit for displays

I think this is a really nice consequence of constraining displays into observables

Strilanc:
@babbush @john6060 @jarrodmcc

babbush:
I think this is a nice idea. But when you say "don't allow observables to be added into circuits", what do you mean exactly? Do you mean that all observables will be understood to be taken at the end of a circuit? While this is appropriate for most of the early NISQ experiments, eventually we'll want to be able to schedule measurements during a circuit. I realize that fast feedforward is still years away - that's not what I'm talking about. I'm talking about some scheme like (https://arxiv.org/abs/1711.07500) where measurements during the circuit are used to create dissipative maps that extract entropy. Its something I expect we'll want in the next couple years.

babbush:
Also, I do think that while doing classical simulations it will often be a useful debugging tool to be able to request expectation values at intermediate parts of the circuit, even though that is clearly not practical to do on the chip.

Strilanc:
> eventually we'll want to be able to schedule measurements during a circuit. I realize that fast feedforward is still years away

Displays cannot be used in that way. Later parts of the circuit do not have access to the statistical result that was computed by the display during its associated sampling runs.

> it will often be a useful debugging tool to be able to request expectation values at intermediate parts of the circuit

Yes. But that functionality can be emulated relatively easily by a user submitting the first half of a circuit.

Strilanc:
Here is some rough prototyping code:

```python
async def sample_observable(sampler: Sampler,
                            circuit: cirq.Circuit,
                            observable: Observable[TKey, TVal],
                            desired_concurrent_jobs: int,
                            desired_total_samples: int) -> TVal:
[...]

Issue #228, opened 2018-03-26
Define async result type for engine calls


Strilanc:

dabacon:
#1982 #1875

dabacon:
Does this make sense any more?  Was the idea for jobs to have an async method that returns the results?

MichaelBroughton:
FYI Going to hijack this issue label section briefly for testing in Triage party

tanujkhattar:
This is probably stale by now. @verult Please reopen the issue if you think this is still relevant. I'll go ahead and close it.

verult:
+1 Engine now has async methods which can be awaited on, so I believe this is obsolete. cc @wcourtney

PR #1551, opened 2019-04-19
Add cirq.SampleCollector and associated utilities

Strilanc:
This is an idea for how we can specify the equivalent of an 'observable' to measure at the end of a circuit, in a way that allows us to collect the information without being sensitive to latency.

There are two main parts: the `cirq.SampleCollector` interface, and the actual driving code `cirq.async_collect_samples`. All of the complicated async shenanigans are packed inside the async collect samples method, while all the logic about what to collect and what to do with it are inside the collector. I've included a quick and dirty example of a thing that collects pauli string expectations.

Here is what using the functionality looks like:

```python
async def main():
    pauli_collector = cirq.PauliStringSampleCollector(
        circuit=cirq.Circuit.from_ops(cirq.H(a), cirq.CNOT(a, b)),
        samples_per_term=1000,
        terms=cirq.LinearDict({
            cirq.X(a) * cirq.X(b): 10,
            cirq.Y(a) * cirq.Y(b): 100,
            cirq.Z(a) * cirq.Z(b): 1
        }))

    await cirq.async_collect_samples(
        collector=pauli_collector,
        sampler=cirq.Simulator(),
        concurrency=10)

    print("final result", pauli_collector.estimated_energy())
```

Part of https://github.com/quantumlib/Cirq/issues/1539


dabacon:
#228

cirq/work/collector.py
```diff
@@ -0,0 +1,140 @@
...
+
+from cirq import circuits, study
+from cirq.work import sampler, work_pool
+
+TResult = TypeVar('TResult')
+TKey = TypeVar('TKey')
+
+
+class CircuitSampleJob(NamedTuple('CircuitSampleJobBase', [
+    ('circuit', circuits.Circuit),
```
dabacon: what about parameters?

cirq/work/collector.py
```diff
@@ -0,0 +1,140 @@
...
+    pass
+
+
+class SampleCollector(Generic[TKey, TResult], metaclass=abc.ABCMeta):
+    """An interface for concurrently collecting sample data."""
+
+    @abc.abstractmethod
+    def next_job(self) -> Union[None,
+                                CircuitSampleJob[TKey],
+                                Iterable[CircuitSampleJob[TKey]]]:
```
dabacon: why also iterable?

cirq/work/collector.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # If no jobs were started or running, we're in a steady state. Halt.
+        if not pool.num_uncollected:
+            break
+
+        # Forward next job result from pool.
+        done_job, done_val = await pool.__anext__()
+        collector.on_job_result(done_job, done_val)
+
+    return collector.result()
```
dabacon: suggest returning collector and user can extract result or other info that has been gathered.

cirq/work/pauli_string_collector.py
```diff
@@ -0,0 +1,99 @@
...
+        self._circuit = circuit
+        self._samples_per_job = 1000
+        self._terms = value.LinearDict({
+            abs(k): v * k.coefficient for k, v in terms.items()
+        })
+        self._zeros = collections.defaultdict(lambda: 0)
+        self._ones = collections.defaultdict(lambda: 0)
+
+    def next_job(self) -> Optional[collector.CircuitSampleJob[ops.PauliString]]:
+        pauli = random.choice(list(self._terms.keys()))
```
dabacon: does randomness really help here?  it might lead to frustration in debugging and also means that for work sent engine the order will vary, which I don't think is desirable

cirq/work/pauli_string_collector.py
```diff
@@ -0,0 +1,99 @@
...
+    measure_layer = ops.Moment([
+        ops.measure(*sorted(pauli_string.keys()), key='out')
+    ])
+
+    circuit = circuit.copy()
+    fix_map = {
+        ops.X: ops.Y**0.5,
+        ops.Y: ops.X**-0.5,
+    }
+    transform_layer = ops.Moment([
```
dabacon: nit: seems like the fix_map code should live over in pauli land.  This is a common thing to want.

cirq/work/collector.py
mpharrigan: big +1

cirq/work/pauli_string_collector.py
```diff
@@ -0,0 +1,99 @@
...
+        Args:
+            circuit: Produces the state to be tested.
+            terms: The pauli product observables to measure. Their sampled
+                expectations will be scaled by their coefficients and their
+                dictionary weights, and then added up to produce the final
+                result.
+            samples_per_job: How many samples to request at a time.
+        """
+        self._circuit = circuit
+        self._samples_per_job = 1000
```
kevinsung: `samples_per_job` instead of `1000`

cirq/work/pauli_string_collector.py
kevinsung: It actually already exists as `PauliString.to_z_basis_ops`.

dabacon:
It would be great to get this in for a coming demo :)

dabacon:
What is the status of this?

cirq/work/collector.py
Strilanc: But the user passed the collector in... they already have it? I'll instead modify the method to have an awaitable None result.

cirq/work/pauli_string_collector.py
Strilanc: That's much better.

cirq/work/collector.py
```diff
@@ -0,0 +1,140 @@
...
+        pass
+
+
+async def async_collect_samples(collector: SampleCollector[Any, TResult],
+                                sampler: sampler.Sampler,
+                                *,
+                                min_concurrent_jobs: int = 1,
+                                max_total_samples: Optional[int] = None) -> TResult:
+    """Concurrently collects samples from a simulator or hardware.
+
```
dabacon: Think this should include an example of how to use this.  Many won't be aware of "await" and async in python so especially important here.

Strilanc:
Status is that I'm polishing it up so that it can be submitted

cirq/work/collector.py
Strilanc: Just makes it easier for users who want to yield a bunch of jobs at once.

cirq/work/pauli_string_collector.py
Strilanc: Replaced with just iterating over the terms.

cirq/work/pauli_string_collector.py
Strilanc: Done.

cirq/work/pauli_string_collector.py
```diff
@@ -0,0 +1,103 @@
...
+            for pauli_string, coef in terms.items()
+        ]
+        self._zeros = collections.defaultdict(
+            lambda: 0)  # type: MutableMapping[ops.PauliString, int]
+        self._ones = collections.defaultdict(
+            lambda: 0)  # type: MutableMapping[ops.PauliString, int]
+        self._samples_per_term = samples_per_term
+        self._total_samples_requested = 0
+
+    def next_job(self) -> Optional[collector.CircuitSampleJob]:
```
dabacon: this is sort of explicitly the pattern we don't want to follow for this type of estimation.  It causes the jobs to be broken up into pieces that could be combined using parameterizations.

cirq/work/pauli_string_collector_test.py
```diff
@@ -0,0 +1,34 @@
...
+
+import cirq
+
+
+def test_pauli_string_sample_collector():
+    a, b = cirq.LineQubit.range(2)
+    p = cirq.PauliStringSampleCollector(cirq.Circuit.from_ops(
+        cirq.H(a), cirq.CNOT(a, b), cirq.X(a), cirq.Z(b)),
+                                        samples_per_term=100,
+                                        terms=cirq.LinearDict({
```
dabacon: please test empty terms

cirq/work/work_pool.py
```diff
@@ -0,0 +1,89 @@
...
+        assert not self._no_more_work_coming
+        output = asyncio.Future()  # type: asyncio.Future
+        self._out_queue.append(output)
+        self._work_queue.append(output)
+        asyncio.ensure_future(self._async_handle_work_completion(work),
+                              loop=self._loop)
+
+    @property
+    def num_active(self) -> int:
+        """The amount of work in the pool that has not completed."""
```
dabacon: Mention that this and property below are thread safe (deque is thread safe for pop append and len)

cirq/work/sampler.py
```diff
@@ -14,14 +14,27 @@
...
+from typing import (List, Union, Awaitable)
 
 from cirq import circuits, schedules, study
 
 
 class Sampler(metaclass=abc.ABCMeta):
     """Something capable of sampling quantum circuits. Simulator or hardware."""
 
+    @abc.abstractmethod
+    def async_sample(self, program: Union[circuits.Circuit, schedules.Schedule],
```
maffoo: This is an async version of the `run` method below so I'd suggest calling it `async_run` or `run_async` (I personally prefer the latter with "async" as a suffix so that these two will appear together in autocompletion lists, etc.)

cirq/work/work_pool.py
```diff
@@ -0,0 +1,89 @@
...
+        self._allow_anext = asyncio.Semaphore(value=0)
+        self._no_more_work_coming = False
+        self._done_event = asyncio.Future()  # type: asyncio.Future
+        self._work_queue = collections.deque()  # type: collections.deque
+        self._out_queue = collections.deque()  # type: collections.deque
+
+    def include_work(self, work: Union[Awaitable, asyncio.Future]) -> None:
+        """Adds asynchronous work into the pool and begins executing it."""
+        assert not self._no_more_work_coming
+        output = asyncio.Future()  # type: asyncio.Future
```
maffoo: use `self._loop.create_future()`.

cirq/work/work_pool.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class CompletionOrderedAsyncWorkPool:
+    """Ensures given work is executing, and exposes it in completion order."""
+
+    def __init__(self, loop: Optional[asyncio.AbstractEventLoop] = None):
+        self._loop = loop if loop is not None else asyncio.get_event_loop()
+        self._allow_anext = asyncio.Semaphore(value=0)
+        self._no_more_work_coming = False
+        self._done_event = asyncio.Future()  # type: asyncio.Future
```
maffoo: use `self._loop.create_future()`.

cirq/work/work_pool.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def _react_to_progress(self) -> None:
+        """Common cleanup after something interesting happens."""
+        if self._no_more_work_coming and not self._out_queue:
+            self._done_event.set_result(True)
+        self._allow_anext.release()
+
+    def set_all_work_received_flag(self) -> None:
+        """Indicates to the work pool that no more work will be added."""
+        assert not self._no_more_work_coming
```
maffoo: Need to also set `self._no_more_work_coming = True`.

cirq/work/collector.py
```diff
@@ -0,0 +1,164 @@
...
+    @abc.abstractmethod
+    def on_job_result(self, job: CircuitSampleJob,
+                      result: study.TrialResult) -> None:
+        """Called by driving code when sample results have become available.
+
+        The results should be incorporated into the collector's state.
+        """
+
+
+async def async_collect_samples(collector: SampleCollector,
```
maffoo: Consider making this an instance method on `SampleCollector`?

cirq/work/collector.py
```diff
@@ -0,0 +1,164 @@
...
+@value.value_equality(unhashable=True)
+class CircuitSampleJob:
+    """Describes a sampling task."""
+
+    # pylint: disable=redefined-builtin
+    def __init__(self,
+                 circuit: circuits.Circuit,
+                 *,
+                 repetitions: int,
+                 id: Any = None):
```
maffoo: Maybe call this `tag` or `label` to avoid redefining the `id` builtin?

cirq/work/collector.py
```diff
@@ -0,0 +1,164 @@
...
+import abc
+from typing import Optional, Any, Iterable, Union, List
+
+import numpy as np
+
+from cirq import circuits, study, value
+from cirq.work import sampler, work_pool
+
+
+@value.value_equality(unhashable=True)
```
maffoo: Not for this PR, obviously, but have you considered using [`attrs`](https://www.attrs.org/en/stable/) for this kind of thing?

cirq/work/collector.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def _flatten_jobs_helper(
+        out: List[CircuitSampleJob],
+        given: Union[None, CircuitSampleJob, Iterable[CircuitSampleJob]]):
+    if isinstance(given, CircuitSampleJob):
+        out.append(given)
+    elif given is not None:
+        for item in given:
+            _flatten_jobs_helper(out, item)
```
maffoo: This implementation allows nested iterables of `CircuitSampleJob`, which the type annotation does not allow. Did you intend to allow nested interables but can't express it in the types?

cirq/work/collector.py
```diff
@@ -0,0 +1,164 @@
...
+            new_jobs = _flatten_jobs(collector.next_job())
+
+            # If no jobs were given, stop asking until something completes.
+            if not new_jobs:
+                break
+
+            # Start new sampling jobs.
+            for new_job in new_jobs:
+                remaining_samples -= new_job.repetitions
+                pool.include_work(_start_async_job(new_job))
```
maffoo: Note that this could exceed the specified `concurrency` limit if a collector returns a large chunk of jobs from `next_job`. I think it would be better to include backpressure in the work pool; you would pass the concurrency to the work pool on creation and then here you'd do something like
```python
await pool.ready()
pool.include_work(_start_async_job(new_job))
```
where `pool.ready()` would complete only when the number of outstanding work items drops below the specified limit. Of course, in that case you'd need to run two separate coroutines to submit work to the pool and send results back to the collector as they complete, but I think that would be conceptually better than having a single loop that does both jobs anyway.

cirq/ops/pauli_string.py
```diff
@@ -115,6 +117,11 @@ def __rmul__(self, other):
             return PauliString(self._qubit_pauli_map, self._coefficient * other)
         return NotImplemented
 
+    def __truediv__(self, other):
+        if isinstance(other, (int, float, complex)):
```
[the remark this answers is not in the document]
Strilanc: This doesn't include concepts like quaternions?

cirq/ops/pauli_string.py
```diff
@@ -72,6 +71,9 @@ def _value_equality_values_cls_(self):
     def equal_up_to_coefficient(self, other: 'PauliString') -> bool:
         return self._qubit_pauli_map == other._qubit_pauli_map
 
+    def __abs__(self) -> 'PauliString':
+        return PauliString(self._qubit_pauli_map)
```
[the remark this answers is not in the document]
Strilanc: Renamed to `unit`.

cirq/work/collector.py
Strilanc: I fixed this by just keeping a queue of work.

I agree that it would be ideal for the pool to have a method like `await pool.active_jobs_drops_below(x)`.

cirq/work/collector.py
Strilanc: Updated the type annotation.

cirq/work/collector.py
Strilanc: I hadn't. Could be useful.

cirq/work/collector.py
Strilanc: Done.

cirq/work/collector.py
Strilanc: Done.

cirq/work/work_pool.py
Strilanc: Yup.

cirq/work/work_pool.py
Strilanc: Done

cirq/work/work_pool.py
Strilanc: Done

cirq/work/sampler.py
Strilanc: Done.

cirq/sim/simulator.py
```diff
@@ -23,25 +23,40 @@
...
+        done = asyncio.Future()  # type: asyncio.Future
+        loop = asyncio.get_event_loop()
+
+        def run():
+            result = self.run(program, repetitions=repetitions)
+            loop.call_soon_threadsafe(lambda: done.set_result(result))
+
+        t = threading.Thread(target=run)
+        t.start()
+        return await done
```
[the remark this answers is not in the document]
Strilanc: I've moved this entire default implementation into the base Sampler class.

cirq/sim/simulator.py
```diff
@@ -23,25 +23,40 @@
...
 
+    async def async_sample(self,
+                           program: Union[circuits.Circuit, schedules.Schedule],
+                           *, repetitions: int) -> study.TrialResult:
+        done = asyncio.Future()  # type: asyncio.Future
+        loop = asyncio.get_event_loop()
+
+        def run():
+            result = self.run(program, repetitions=repetitions)
+            loop.call_soon_threadsafe(lambda: done.set_result(result))
```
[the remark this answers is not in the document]
Strilanc: Done.

cirq/sim/simulator.py
```diff
@@ -23,25 +23,40 @@
...
+class SimulatesSamples(work.Sampler, metaclass=abc.ABCMeta):
     """Simulator that mimics running on quantum hardware.
 
     Implementors of this interface should implement the _run method.
     """
 
+    async def async_sample(self,
+                           program: Union[circuits.Circuit, schedules.Schedule],
+                           *, repetitions: int) -> study.TrialResult:
+        done = asyncio.Future()  # type: asyncio.Future
```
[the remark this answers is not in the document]
Strilanc: Done.

Strilanc:
> wrap up the queue into a generator

This causes the consuming code to be unable to distinguish whether or not the last thing it did was ask for work or provide more information (because the generator hides whether or not a job was provided as part of a bundle or as a singleton). This can result in the consumer asking for more work even if the producer just returned [] or None (which is supposed to mean "stop calling me until you give me more results").

cirq/work/work_pool.py
```diff
@@ -0,0 +1,90 @@
...
+        self._no_more_work_coming = True
+        self._react_to_progress()
+
+    async def _anext_helper(self):
+        await self._allow_anext.acquire()
+        if self._out_queue:
+            return await self._out_queue.popleft()
+        else:
+            self._allow_anext.release()
+            raise StopAsyncIteration('no_more_work')
```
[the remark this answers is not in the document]
Strilanc: That flag is guaranteed to be set at this point. There is a test that exercises this case.

The trick is that `await self._allow_anext.acquire()` is waiting on a semaphore that only increments when work arrives (pushing into out queue) or all-received is set (not pushing). So either `_out_queue` will have a new entry to consume (indicating we were triggered by work arriving) or it will be empty (indicating we were triggered by no more work arriving).

I'll add an assert.

cirq/ops/pauli_string.py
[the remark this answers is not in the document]
Strilanc: Okay. Used numbers.

cirq/work/collector.py
Strilanc: Tried. It has slightly wrong semantics (as discussed).

cirq/work/pauli_string_collector.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def __init__(self,
+                 circuit: circuits.Circuit,
+                 samples_per_term: int,
+                 terms: value.LinearDict[ops.PauliString],
+                 max_samples_per_job: int = 1000000):
+        """
+        Args:
+            circuit: Produces the state to be tested.
+            terms: The pauli product observables to measure. Their sampled
```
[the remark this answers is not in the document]
Strilanc: Done.

cirq/work/sampler.py
```diff
@@ -54,11 +56,42 @@ def run_sweep(
...
         """
+
+    async def run_async(self,
+                        program: Union[circuits.Circuit, schedules.Schedule], *,
+                        repetitions: int) -> Awaitable[study.TrialResult]:
+        """Asynchronously samples from the given Circuit or Schedule.
+
+        By default, this method calls `run` on another thread and yields the
+        result via the asyncio event loop. However, child classes are free to
+        override it to use other strategies.
```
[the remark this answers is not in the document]
Strilanc: Will extract that method if we end up using this in more places.

cirq/work/work_pool.py
```diff
@@ -0,0 +1,90 @@
...
+    def num_uncollected(self) -> int:
+        """The amount of work still in the pool that has completed."""
+        return len(self._out_queue)
+
+    async def _async_handle_work_completion(self, work: Awaitable):
+        try:
+            result = await work
+            self._work_queue.popleft().set_result(result)
+        except BaseException as ex:
+            self._work_queue.popleft().set_exception(ex)
```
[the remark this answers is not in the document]
Strilanc: Used the else.

cirq/work/work_pool.py
```diff
@@ -0,0 +1,90 @@
...
+    async def _anext_helper(self):
+        await self._allow_anext.acquire()
+        if self._out_queue:
+            return await self._out_queue.popleft()
+        else:
+            self._allow_anext.release()
+            raise StopAsyncIteration('no_more_work')
+
+    def __anext__(self) -> Awaitable:
+        return asyncio.ensure_future(self._anext_helper())
```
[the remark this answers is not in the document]
Strilanc: Specified the loop. I like the idea of inlining the anext but it causes test failures due to the lack of ensure_future on the caller side.

cirq/sim/simulator.py
```diff
@@ -25,18 +25,25 @@
...
 from typing import (
-    Any, Dict, Hashable, Iterator, List, Tuple, Union, Optional)
+    Any,
+    Dict,
+    Hashable,
+    Iterator,
+    List,
+    Tuple,
+    Union,
+    Optional,
```
dabacon: nit: might remove final "," since this is a lot of vert space for an import

cirq/work/collector.py
```diff
@@ -0,0 +1,221 @@
...
+    # pylint: disable=unused-import
+    import cirq
+
+
+@value.value_equality(unhashable=True)
+class CircuitSampleJob:
+    """Describes a sampling task."""
+
+    def __init__(self,
+                 circuit: circuits.Circuit,
```
dabacon: In cases where the circuit will need to be uploaded, and should only be uploaded once, how will this work?  Will this somehow show up in tag?

cirq/work/pauli_sum_collector.py
```diff
@@ -0,0 +1,103 @@
...
+from cirq import circuits, study, ops
+from cirq.work import collector
+
+
+class PauliSumCollector(collector.Collector):
+    """Estimates the energy of a linear combination of Pauli observables."""
+
+    def __init__(self,
+                 circuit: circuits.Circuit,
+                 hamiltonian: ops.PauliSum,
```
dabacon: I don't like the name hamiltonian as that is a the energy observable.  Could we just use pauli_sum or something less specific?

cirq/work/pauli_sum_collector_test.py
```diff
@@ -0,0 +1,28 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import cirq
+
+
```
dabacon: you've got an ```assert pauli_string``` but no test that triggers it.

cirq/work/pauli_sum_collector_test.py
```diff
@@ -0,0 +1,28 @@
...
+    a, b = cirq.LineQubit.range(2)
+    p = cirq.PauliSumCollector(
+        circuit=cirq.Circuit.from_ops(cirq.H(a), cirq.CNOT(a, b), cirq.X(a),
+                                      cirq.Z(b)),
+        hamiltonian=cirq.X(a) * cirq.X(b) - 16 * cirq.Y(a) * cirq.Y(b) +
+        4 * cirq.Z(a) * cirq.Z(b),
+        samples_per_term=100)
+    completion = p.collect_async(sampler=cirq.Simulator())
+    cirq.testing.assert_asyncio_will_have_result(completion, None)
+    assert p.estimated_energy() == 11
```
dabacon: test case where expectation is imaginary?

Strilanc: Done.

cirq/work/pauli_sum_collector_test.py
Strilanc: That assert is actually never supposed to fail; it's documentation. Added a test that uses an identity term, and ensured it ends up in the output.

cirq/work/pauli_sum_collector.py
Strilanc: Renamed to observable.

cirq/work/collector.py
Strilanc: We'll have to refactor in the future, e.g. by adding a parameters field and expecting the driving method to recognize re-used circuits.

cirq/sim/simulator.py
Strilanc: Acknowledged, but I prefer this style (and the formatter accepts it).

Strilanc:
Very strange. The response from https://api.github.com/repos/quantumlib/cirq/pulls/1551/reviews really does indicate no approved reviews are present? And yet the UI clearly shows Dave approved.

PR #1553, opened 2019-04-22
Metric proto

dabacon:


dabacon:
This will fail format.  Going to add an yapf ignore in a separate pull request

cirq/api/google/v2/metrics.proto
```diff
@@ -0,0 +1,79 @@
...
+package cirq.api.google.v1;
+
+option java_package = "com.google.cirq.api.google.v2";
+option java_outer_classname = "CalibrationProto";
+option java_multiple_files = true;
+
+// Performance metrics for a quantum processor.
+message Metrics {
+
+  // The time the metrics were collected, in milliseconds since Jan 1, 1970.
```
Strilanc: This quantity is ambiguous. Refer to a specific time standard, such as unix time or UTC. Include it in the field name.

cirq/api/google/v2/metrics.proto
```diff
@@ -0,0 +1,79 @@
...
+  // Metric collection take time, so this is the time at which all of the
+  // metrics have been collected.
+  uint64 collection_time_ms = 1;
+
+  // Map from a metric type to a collection of the metrics all of this type.
+  // Keys should be snake_case and fairly descriptive of the metric.
```
Strilanc: Why is the format encouraging a style guide on its keys?

cirq/api/google/v2/metrics.proto
```diff
@@ -0,0 +1,79 @@
+syntax = "proto3";
+
+package cirq.api.google.v1;
+
+option java_package = "com.google.cirq.api.google.v2";
+option java_outer_classname = "CalibrationProto";
+option java_multiple_files = true;
+
+// Performance metrics for a quantum processor.
+message Metrics {
```
Strilanc: It's confusing that a `Metrics` contains many `MetricCollection`s instead of many `Metric`s.

Maybe `NamedMetricCollections`?

cirq/api/google/v2/metrics.proto
```diff
@@ -0,0 +1,79 @@
...
+  uint64 collection_time_ms = 1;
+
+  // Map from a metric type to a collection of the metrics all of this type.
+  // Keys should be snake_case and fairly descriptive of the metric.
+  map<string, MetricCollection> metrics = 2;
+}
+
+// A collection of metrics of the same type.
+message MetricCollection {
+  // The metrics in the collection. These should all be of the same type.
```
Strilanc: Why not have the oneof at this level if we're going to require that they all be the same type?

cirq/api/google/v2/metrics.proto
dabacon: because standards for casing make you not have to remember whether it was T1_time or T1Time etc.

cirq/api/google/v2/metrics.proto
dabacon: you can't have repeated oneofs, so if you do that you will end up making container classes for each metric type.  Do you think that's better?

cirq/api/google/v2/metrics.proto
dabacon: Unix time is in UTC.  

I've added words to say it is unix time.  I think calling it "unix time" doesn't really help, but labeling the time unit is important.

cirq/api/google/v2/metrics.proto
dabacon: Ugh.  That's like java ;)  
 How about PerformanceSnapshot?

dabacon:
Hey look I responded faster than the label could be removed.

cirq/api/google/v2/metrics.proto
Strilanc: Oh right this is coming from the server, so we can enforce things like that.

cirq/api/google/v2/metrics.proto
Strilanc: Yeah that's better.

Strilanc:
You didn't respond to the table suggestion ;)

cirq/api/google/v2/metrics.proto
Strilanc: I think we could encapsulate this into the "target" of a metric, with a special all target. It will also give us flexibility to talk about subsets of qubits.

dabacon:
Ah missed your main comment.

I think that solution is fine as well (though we need to have collection time).  Let me refactor real quick here.

dabacon:
Note that you'll want repeated values for things like tomography, so I don't think you want a single "value"

dabacon:
PTAL.  Simplified per suggestion.  Did keep repeated values (since this makes sense for things like tomography).  Dumped float value and added string.

cirq/api/google/v2/metrics.proto
```diff
@@ -0,0 +1,51 @@
...
+  // The values of this Metrics. Repeated to support multivalued metrics like
+  // tomography.
+  repeated Value values = 3;
+}
+
+// A generic metric value. Metrics of the same type should all support only
+// one of the types.
+message Value {
+  // The raw value.
+  oneof val {
```
Strilanc: Maybe we could have the repetition in here, instead of in `Metric`? So you can have a `Union[..., List[Float]]` instead of a `List[Union[..., Float]]`.

cirq/api/google/v2/metrics.proto
```diff
@@ -0,0 +1,51 @@
...
+    double double_val = 1;
+
+    int32 int32_val = 2;
+
+    int64 int64_val = 3;
+
+    string str_value = 4;
```
Strilanc: val val val value

cirq/api/google/v2/metrics.proto
```diff
@@ -0,0 +1,51 @@
...
+
+package cirq.api.google.v1;
+
+option java_package = "com.google.cirq.api.google.v2";
+option java_outer_classname = "CalibrationProto";
+option java_multiple_files = true;
+
+// A snapshot of the performance metrics for a quantum processor at a
+// particular time.
+message MetricSnapshot {
```
Strilanc: `MetricsSnapshot`

cirq/api/google/v2/metrics.proto
```diff
@@ -0,0 +1,51 @@
...
+
+  // The time the metrics were collected, in unix time (milliseconds since
+  // Epoch minus leap seconds). Metric collection take time, so this is the
+  // time at which all of the metrics have been collected.
+  uint64 collection_time_ms = 1;
```
Strilanc: How about `timestamp_ms`? There's not really a single well defined time when all the metrics were collected, but there can be a well defined time when we stamped it.

cirq/api/google/v2/metrics.proto
dabacon: Repeated oneofs add a lot of indirection.  Plus it's not clear they need to be of the same type, actually.

cirq/api/google/v2/metrics.proto
dabacon: done

cirq/api/google/v2/metrics.proto
dabacon: fixed.

cirq/api/google/v2/metrics.proto
dabacon: I've deleted the comment about same type.

dabacon:
PTAL

Strilanc:
Seems to be missing some of the changes.

dabacon:
I see the changes.

PR #1554, opened 2019-04-22
Exclude pb2.py files from format

dabacon:
Monkey banging at shell script.


check/format-incremental
```diff
@@ -66,6 +66,17 @@ if [ -z "${rev}" ]; then
 fi
 echo "Comparing against revision '${rev}'." >&2
 
+# Copy over .yapfignore or use local version. Cleanup if copying.
+ignore_exists='[ -f .yapfignore ]'
+if ! $ignore_exists; then
+  function cleanup {
+    rm .yapfignore
```
Strilanc: Can we pass this file in as a command line parameter, instead of by copying and deleting?

.yapfignore
```diff
@@ -0,0 +1 @@
+**/*_pb2.py
```
vtomole: newline needed

dev_tools/conf/.yapfignore
```diff
@@ -0,0 +1 @@
+**/*_pb2.py
```
vtomole: newline

check/format-incremental
```diff
@@ -66,6 +66,17 @@ if [ -z "${rev}" ]; then
...
+  }
+  trap cleanup EXIT;
+fi
+cp -n dev_tools/conf/.yapfignore .yapfignore
```
vtomole: Oh this is where the 2 files are comming from.

check/format-incremental
dabacon: yapf doesn't allow a config path to be passed in :(

dabacon:
PTAL.  Just filtered it out of changed files directly.

check/format-incremental
```diff
@@ -68,7 +68,7 @@ echo "Comparing against revision '${rev}'." >&2
 
 # Get the _test version of changed python files.
 needed_changes=0
-changed_files=$(git diff --name-only ${rev} -- | grep "\.py$")
+changed_files=$(git diff --name-only ${rev} -- | grep "\.py$" | grep -v "pb2\.py$")
```
Strilanc: You should also include the underscore. I can imagine a file accidentally ending in `pb2`, but `_pb2` not so much.

dabacon: done

Issue #1535, opened 2019-04-16
warnings while building docs


dabacon:
From running build-docs.sh
```
WARNING: [autosummary] failed to import 'cirq.ReversibleCompositeGate': no module named cirq.ReversibleCompositeGate
WARNING: [autosummary] failed to import 'cirq.RotationErrorChannel': no module named cirq.RotationErrorChannel
WARNING: [autosummary] failed to import 'cirq.Symbol': no module named cirq.Symbol
WARNING: [autosummary] failed to import 'cirq.Tolerance': no module named cirq.Tolerance
/usr/local/google/home/dabacon/.virtualenvs/cirq/lib/python3.6/importlib/_bootstrap.py:219: RuntimeWarning: numpy.ufunc size changed, may indicate binary incompatibility. Expected 216, got 192
  return f(*args, **kwds)
/usr/local/google/home/dabacon/.virtualenvs/cirq/lib/python3.6/importlib/_bootstrap.py:219: RuntimeWarning: numpy.ufunc size changed, may indicate binary incompatibility. Expected 192 from C header, got 216 from PyObject
  return f(*args, **kwds)
WARNING: [autosummary] failed to import 'cirq.rotation_error': no module named cirq.rotation_error
```

Issue #857, opened 2018-08-14
broken link in tutorial


sboixo:
In the tutorial, the link on "circuit" in this sentence is broken:
To build the above variational quantum algorithm using Cirq, one begins by building the appropriate circuit.

From Craig:
It's because we are using relative links, but once it's in read the docs it's relative to that instead of GitHub.

Tell Dave we should not update the links, but instead update our tooling to translate the links. That goes for both read the docs and the package generation.

vtomole:
I think this is already fixed. The source says `one begins by building the appropriate [circuit](circuits.md)`. This will link to the appropriate `.md` file. @dabacon

Strilanc:
There are probably many many relative links, throughout the docs, with this issue.

dabacon:
#856 just "fixes" if you look at gitub.  We still need to update the tool for readthedocs links to work.

PR #1555, opened 2019-04-23
Get rid of warnings when building readthedocs

vtomole:
This gets rid of all the warnings and
Fixes: https://github.com/quantumlib/Cirq/issues/1535

We've been through the relative vs absolute link wringer before (https://github.com/quantumlib/Cirq/issues/857), but this is a temporary solution until I figure out how to display sourcefiles in readthedocs; if it's possible.

### DIFF
--- a/cirq/circuits/circuit.py
+++ b/cirq/circuits/circuit.py
@@ -792,7 +792,7 @@ class Circuit:
             predicate: A predicate on ops.Operations which is being checked.
 
         Returns:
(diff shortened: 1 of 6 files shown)


vtomole:
@dabacon Yes. The Python docs do it by attaching the release source link onto the page based on it's branch: https://docs.python.org/3/library/argparse.html#module-argparse (top left, `Source code`:  [Lib/argparse.py](https://github.com/python/cpython/blob/3.7/Lib/argparse.py)). Something like that.

PR #1559, opened 2019-04-24
Sync v0.5.0

Strilanc:


docs/noise.md
```diff
@@ -101,7 +103,7 @@ such that each Pauli gate occurs with probability `p/3`.
 To construct channels, useful helpers are provided `cirq.asymmetric_depolarize`
 and `cirq.depolarize`.
 
-Another common case is when only a Pauli X (bit flip) can occur, or
+Another common case is when only a Pauli X (bit flip) can occur, or()
```
vtomole: Missed this on previous merge.
```suggestion
Another common case is when only a Pauli X (bit flip) can occur, or
```

Strilanc: Open an issue for it. Not worth blocking release for this typo.

Strilanc:
Can't do fast forward here. Trying something else...

PR #1560, opened 2019-04-24
Remove unnecessary parens in noise.md

vtomole:


Strilanc:
Thanks for catching this.

vtomole:
@Strilanc If i caught it a min earlier!

PR #1565, opened 2019-04-24
Methods that take cirq.Duration should also take datetime.timedelta

### LINKED ISSUE #1491: Methods that take cirq.Duration should also take datetime.timedelta
We still need duration, because timedelta only goes to microseconds.

timothyman:
- Added factory method `from_timedelta` to `Duration`
- Arithmetic operation overloading in `Duration` to support `timedelta`
- Arithmetic operation overloading in `Timestamp` to support `timedelta`
- Added `timedelta` in constructor of `Schedule` and `ScheduledOperation`
- Added `timedelta` in constructor of `XMON`
- Added `timedelta` in constructor of `IonDevice`
- Added `timedelta` in constructor of `NeutralAtomDevice`

Fixes #1491


cirq/google/xmon_device.py
```diff
@@ -42,6 +42,13 @@ def __init__(self,
             exp_11_duration: The maximum duration of an ExpZ operation.
             qubits: Qubits on the device, identified by their x, y location.
         """
+        if isinstance(measurement_duration, timedelta):
```
Strilanc: Move this conditional logic into `Duration.from`. Have `Duration.from` return its argument if that argument is already a duration, wrap it if it's a time delta, and otherwise raise an error.

timothyman: And that would be a TypeError?

Strilanc: or a value error. Either is fine.

PR #1566, opened 2019-04-25
adds repr for PeriodicValue

bryano:


cirq/value/periodic_value.py
```diff
@@ -76,6 +76,9 @@ def _approx_eq_(self, other: Any, atol: float) -> bool:
 
         return cirq.protocols.approx_eq(low, high, atol=atol)
 
+    def __repr__(self):
+        return 'cirq.PeriodicValue({!r}, {!r})'.format(self.value, self.period)
```
Strilanc: I think value may be a symbol, and sympy.symbol unfortunately has a bad repr. There's a workaround method to deal with it.

Add a test checking this case.

bryano: Done

Issue #1372, opened 2019-02-13
Drop support for Python 2


vtomole:
Some third party libraries that Cirq uses are phasing out Python 2 support (https://github.com/quantumlib/Cirq/issues/1370). I think we should do the same.

See https://github.com/quantumlib/Cirq/issues/731 for opposing arguments.

Strilanc:
@babbush what is your opinion on this? Do you think openfermion users are still very attached to python 2?

hugovk:
Python 2.7 is EOL on 2020-01-01.

See https://python3statement.org/#sections30-projects for a list of projects which have pledged to drop support in or before 2020.

babbush:
OpenFermion users would probably survive if we dropped python 2 support. If we do this can we also drop 3.4 and 3.5 support and just focus on 3.6? I'm also curious what @jarrodmcc and @ncrubin think.

babbush:
I mean if numpy is dropping python 2 support in 2020 then we'd have no business maintaining it!

hugovk:
> If we do this can we also drop 3.4 and 3.5 support and just focus on 3.6?

At least Python 3.4 is EOL next month (2019-03-16), and I've seen a few other projects also ditching 3.5 to make full use of the new features in 3.6+.

ncrubin:
I would hope all new users would opt for python3 development but I don't think it's too large a problem to support python2 until EOL.  Given the tooling we have for testing in a python2 environment it's not too much of a burden to develop 3.6 code without features that can be translated back to 2.7 (for example, this means don't use @ for matrix multiplication).  

As for the numpy argument...sure...but...tensorflow still supports 2.7, 3.3-6 . :)

hugovk:
TensorFlow is also dropping 2.7 on 2020-01-01.

https://github.com/python3statement/python3statement.github.io/pull/179

Strilanc:
It sounds like 2020-01-01 is a Schelling point for dropping support. I vote we drop it then.

vtomole:
We've decided to drop earlier than 2020 (next release) if no one complains.

vtomole:
@dabacon @Strilanc @viathor  

Anything we've missed?

Strilanc:
I think this is done enough to close. Whatever's left can be eroded gradually.

Good job victory!

PR #1574, opened 2019-04-29
Remove python 2 packaging functionality

Strilanc:
Part of https://github.com/quantumlib/Cirq/issues/1372


dev_tools/packaging/verify-published-package.sh
```diff
@@ -61,13 +61,9 @@ cd "${tmp_dir}"
 trap "{ rm -rf ${tmp_dir}; }" EXIT
 
 # Test both the python 2 and python 3 versions.
```
dabacon: remove for loop?

dabacon: update comment

vtomole: I assumed there are different python3 versions?

vtomole: Opps merged too soon. Will correct!

Strilanc: That was the idea.

Strilanc: Prefer to defer to the author of a PR as the one to merge it, unless it's blocking another PR or of course if they aren't able to merge it in the first place. The author is in the best position to know if it's ready or not.

I got into the habit of doing the merges for people because most other contributors can't merge, and because I usually have a good idea of if it's ready or not. I'll try to tone that down.

PR #1575, opened 2019-04-29
Delete continuous-integration/check.sh and simple_check.sh

Strilanc:
- These have fallen out of sync with how the build is run on travis, and are slower than the stuff under check/
- Also remove files that only existed to make these two scripts work

@dabacon I know you still sometimes use these scripts. What do you think?


dabacon:
I still use them yeah.  What should I do instead?

vtomole:
How are these programs different from a script that runs `check/mypy, format, pylint e.t.c at once?"

Strilanc:
These scripts set up entire virtual envs and run the checks in those virtual envs. That's the main difference.

Dave, I personally just use the individual scripts in `check/`.

vtomole:
> set up entire virtual envs and run the checks in those virtual envs

So travis-like locally? Have there been instances where `check/` missed something then travis failed but it would have been caught by `check.sh`?

Strilanc:
Yes, when e.g. we didn't pin the version of pylint.

There's also been the opposite, due to e.g. check.sh not including the format check. Nothing is actually verifying that this script continues to work.

Strilanc:
Got some pushback so it'll stay... *for now*

PR #1577, opened 2019-04-29
Make _compat.py non-python2

vtomole:
Part of https://github.com/quantumlib/Cirq/issues/1372


cirq/ops/phased_x_gate.py
```diff
@@ -15,11 +15,12 @@
 """An `XPowGate` conjugated by `ZPowGate`s."""
 from typing import Union, Sequence, Tuple, Optional, cast
 
+from math import gcd
```
Strilanc: Just import math use `math.gcd`.

PR #1578, opened 2019-04-30
Delete cirq/testing/mock.py, which was only used for python2 compatibility

vtomole:


Strilanc:
I see you picked up on the hint to do bite-sized chunks :+1:

mpharrigan:
Isn't this backwards? On py3 it should be 

```
from unittest import mock
```

and you can drop `mock` backport as a dependency

vtomole:
@mpharrigan Fixed here: https://github.com/quantumlib/Cirq/pull/1599

Issue #1519, opened 2019-04-10
Raise an error when attempting to sample a circuit with no measurements


Strilanc:
Should apply in the SparseSimulator, DensitySimulator, WaveFunctionSimulator, and XmonSimulator.

Introduce a utility method `has_measurements` on `Circuit` to avoid too much code duplication.

dabacon:
This code is in run?

Strilanc:
Could you point at it? When I run a circuit with no measurements I just get an empty trial result. For example, `cirq.Simulator().run(cirq.Circuit())` produces no error.

dabacon:
Whah?  I was asking whether you want to put this in the "run*" methods?  

Yeah current behavior is that no measurements produce an empty trial result.

Strilanc:
Oh, I thought you were asking "is this code in run".

Yes, I think it should be limited to the `run` methods, because the `simulate` methods have useful outputs when given no measurements.

smitsanghavi:
I can take this.

smitsanghavi:
@Strilanc I'm curious, you listed these 4 Simulators specifically. What kind of simulators should not throw error here? Is there an underlying property here that we can use in a parent class itself?

Strilanc:
I was just trying to list them all. I think this applies globally to any simulator, as long as we're talking about sampling the measurement results.

PR #1581, opened 2019-04-30
Raise an error when attempting to sample a circuit with no measurements

smitsanghavi:
Only applies to SimulatesSamples class and its children.
#1519


cirq/google/sim/xmon_simulator_test.py
```diff
@@ -61,6 +61,12 @@ def large_circuit():
     return circuit
 
 
+def empty_measurement_circuit(qubit):
```
Strilanc: You use this method once, and it can be compressed into a single expression `cirq.Circuit.from_ops(cirq.measure(qubit, key='meas'))`. Inline it.

PR #1585, opened 2019-05-01
Remove python2-generate.sh, requirements and update development and i…

vtomole:
…nstall docs


Strilanc:
You searched for all instances of `python2-generate` and confirmed there are none left?

vtomole:
Yes. It was only mentioned in that `development.md`

PR #1587, opened 2019-05-01
Remove `only_in_python3_on_posix`, circular dependency imports and `derive_temporary_python2_environment()`

vtomole:


vtomole:
mypy error when it comes to `pauli_gates`

Issue #1570, opened 2019-04-26
Set y axis range of rabi oscillation experiment plot to [0, 1]


Strilanc:

vtomole:
Closing manually as Bot doesn't work for "Fix" apparently.

PR #1588, opened 2019-05-01
set y axis range of rabi oscillation experiment plot to [0, 1] 

Ashalynd:
Fix for #1570 ?
Screen Shot 2019-05-01 at 19 03 17


Strilanc:
That's not a rabi oscillation plot.

```python
cirq.experiments.rabi_oscillations(cirq.DensityMatrixSimulator(noise=cirq.ConstantQubitNoiseModel(cirq.amplitude_damp(0.3))), cirq.LineQubit(0)).plot()
```

Currently produces

image

Where the y range is determined by the data instead of the range of possible probabilities. This makes the plot visually misleading, since the curve basically always looks the same.

Ashalynd:
Oh I see. This fix takes care of it too.
Screen Shot 2019-05-02 at 11 17 28
Screen Shot 2019-05-02 at 11 21 52

Ashalynd:
Should the same reasoning be applied to Clifford-based randomized benchmarking graphs, and to 3d graphs, as long as they show probabilities / probability densities?
I have added the corresponding changes, let me know if you want to keep them.
Here are all the graphs from qubit_characterization_example.py after the changes:
Screen Shot 2019-05-02 at 12 22 03
Screen Shot 2019-05-02 at 12 21 58
Screen Shot 2019-05-02 at 12 21 40
Screen Shot 2019-05-02 at 12 21 32
Screen Shot 2019-05-02 at 12 21 20

Strilanc:
I agree we should clamp the y axis to 0:1 whenever we're talking about probabilities that naturally span that range.

PR #1589, opened 2019-05-01
Add Pauli expansion to GateOperation

viathor:


cirq/ops/gate_operation.py
```diff
@@ -101,6 +101,9 @@ def _decompose_(self) -> op_tree.OP_TREE:
                                                     self.qubits,
                                                     NotImplemented)
 
+    def _pauli_expansion_(self) -> value.LinearDict[str]:
+        return protocols.pauli_expansion(self.gate)
+
```
dabacon: if you are bored, a lot of these delegates use the private _gate versus the property gate.  Would be nice to make consistent.

viathor: Made them use the property instead of the private field. Thanks for spotting this.

viathor:
Made a trivial formatting change to trigger cla/google.

Strilanc:
Why did you need to trigger the cla check?

In case of annoyance, I have an override script.

viathor:
It appeared to have gotten stuck... or maybe was just taking very long. I'm used to cla/google reporting its status before Travis and in this case it was still waiting T minutes after Travis completed. Don't remember the value of T, though.

PR #1591, opened 2019-05-03
Remove `only_in_python3_on_posix`, and `derive_temporary_python2_environment()`

vtomole:


cirq/sim/wave_function.py
```diff
@@ -258,9 +258,8 @@ def dirac_notation(state: Sequence, decimals: int=2) -> str:
     ket = "|{}⟩"
     for x in range(len(perm_list)):
         format_str = "({:." + str(decimals) + "g})"
-        # Python 2 rounds imaginary numbers to 0, so need to round separately.
-        val = (round(state[x].real, decimals)
-               + 1j * round(state[x].imag, decimals))
+        val = round(state[x], decimals)
```
viathor: In python complex numbers cannot be rounded:

```
In [1]: round(1j)
TypeError                                 Traceback (most recent call last)
<ipython-input-1-81fdcefa71b7> in <module>
----> 1 round(1j)

TypeError: type complex doesn't define __round__ method
```

and the separate round of the real and imaginary parts is a common workaround.

The reason the tests pass is that they only deal with numpy types which for the most part **can** be rounded:

```
In [1]: import numpy as np

In [2]: round(np.complex128(1j))
Out[2]: 1j

In [3]: round(np.complex64(1j))
Out[3]: 1j

In [4]: round(np.complex(1j))
TypeError                                 Traceback (most recent call last)
<ipython-input-4-48bae1814ad4> in <module>
----> 1 round(np.complex(1j))

TypeError: type complex doesn't define __round__ method

In [5]: round(1j)
TypeError                                 Traceback (most recent call last)
<ipython-input-5-81fdcefa71b7> in <module>
----> 1 round(1j)

TypeError: type complex doesn't define __round__ method
```

I think we should:
 * remove the comment,
 * keep separate rounding on the real and imaginary parts as it was,
 * possibly add a unit test for `dirac_notation()` that uses python's native `complex` type for amplitudes in `state`.

vtomole: Interesting! Didn't know Python/Numpy behaved this way. I haven't been able to reproduce this.
```
 cirq.dirac_notation(np.array([1j, 2j]), decimals=2)
```
Runs fine on this implementation. What am i missing?

viathor: Elements of a numpy array have numpy types such as `np.complex128`. Use a python array instead of a numpy array:

```python
cirq.dirac_notation([1j, 2j], decimals=2)
```

PR #1592, opened 2019-05-03
 Add cirq.axis_angle utility method

Strilanc:
Ran into the need for this while trying to decompose gates into fsims.

Also fixed 1 not approximately equalling 1+0j.


cirq/linalg/decompositions.py
```diff
@@ -264,6 +264,110 @@ def so4_to_magic_su2s(
...
+class AxisAngleDecomposition:
+    """Represents a unitary operation as an axis, angle, and global phase."""
+
+    def __init__(self, *, angle: float, axis: Tuple[float, float, float],
+                 global_phase: Union[int, float, complex]):
+        self.global_phase = complex(global_phase)
+        self.axis = tuple(axis)
+        self.angle = float(angle)
+
+    def canonicalize(self) -> 'AxisAngleDecomposition':
```
dabacon: doc what the canonicalization does

cirq/linalg/decompositions.py
```diff
@@ -264,6 +264,110 @@ def so4_to_magic_su2s(
     return a, b
 
 
+@value.value_equality(approximate=True)
+class AxisAngleDecomposition:
+    """Represents a unitary operation as an axis, angle, and global phase."""
```
dabacon: Give an explicit unitary that this corresponds to, i.e. what unitary produces.  Because we want to make sure we get the +\- consistent.

cirq/linalg/decompositions.py
Strilanc: Done.

cirq/linalg/decompositions.py
Strilanc: I added one. Is it the one you expected?

(Please check factors of 2 and such..)

dabacon: Where?

Strilanc: Oh, I added it to the unitary method. I guess it does make more sense on the class.

Strilanc: Moved.

PR #1594, opened 2019-05-06
Return self when controlling an operation with no controls

Strilanc:


cirq/ops/raw_types.py
```diff
@@ -268,12 +268,9 @@ def controlled_by(self, *control_qubits: Qid) -> 'Operation':
...
-            raise ValueError(
-                "Can't get controlled operation without control qubit. Op: {}"
-                .format(repr(self)))
-        else:
-            return ControlledOperation(control_qubits, self)
+        if len(control_qubits) == 0:
+            return self
+        return ControlledOperation(control_qubits, self)
 
 @value.value_equality
```
viathor: nit: formatter wants more empty lines here

viathor:
@dabacon I think tests exists and are in fact failing due to this change. Probably sufficient to update them to cover the changed code.

PR #1596, opened 2019-05-06
Allow arg values to be floats or bytes

dabacon:
Bytes will be used, for example, as invert_mask for a measurement.  Could also be used for control operations (i.e. do or don't do a measurement dependent on this value).

### DIFF
--- a/cirq/api/google/v2/program.proto
+++ b/cirq/api/google/v2/program.proto
@@ -127,8 +127,8 @@ message Arg {
   // Arguments are either a number, a symbol, or an argument function
   // (which recursively depends on Arg).
   //
-  // Number is used to specify an argument that does not vary depending on
-  // RunContext.
+  // ArgValue is used to specify an argument that does not vary
+  // depending on RunContext.
   //
   // Symbol is used when an argument will be resolved (supplied a value)
   // by a Run Context.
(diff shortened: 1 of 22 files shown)


cirq/api/google/v2/program.proto
```diff
@@ -127,21 +127,29 @@ message Arg {
...
+  // depending on RunContext.
   //
   // Symbol is used when an argument will be resolved (supplied a value)
   // by a Run Context.
   //
   // Functions are used to define a simple s-expression tree describing
   // how to combine numbers and symbols mathematically.
   oneof arg {
-    float number = 1;
+    ArgValue arg_value = 1;
```
Strilanc: Why are you putting a oneof inside another oneof, instead of flattening?

Just add a "bytes" option here, instead of introducing ArgValue.

dabacon: Because I want them to be in sync with the values that can be supplied in run_context.

dabacon: this will keep "Points" in run_context to not explode in a strange way.

Strilanc: Why do we want to keep those two values in sync, given that we have an example of them diverging?

dabacon: We want the types of things you feed as an arg to gate to match what a parameter can be passed in to fulfill that arg.  We could just repeat ourselves, but this way if we decide to add a new arg type we automatically get it in both places.

Issue #1597, opened 2019-05-06
Separate perf tests into a different travis check


Strilanc:
Currently the perf output is just a wall of text that test failures hide around. Perf testing is different from correctness testing in relevant ways, so we should keep them separate.

dabacon:
Thanks for fixing this!  Less scrolling!

PR #1600, opened 2019-05-07
Separate perf tests into a different travis check

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/1597


.travis.yml
```diff
@@ -46,4 +46,13 @@ matrix:
     install:
       - pip install -r requirements.txt
       - pip install -r dev_tools/conf/pip-list-dev-tools.txt
-    script: check/pytest --ignore=cirq/contrib
+    script: check/pytest --ignore=cirq/contrib --benchmark-skip
+
+
```
viathor: nit: these other blocks are separated by a single newline, so let's remove one here for consistency.

.travis.yml
```diff
@@ -46,4 +46,13 @@ matrix:
...
+    - os: linux
+    env: NAME=Performance tests
+    python: "3.7"
+    install:
+      - pip install -r requirements.txt
+      - pip install -r dev_tools/conf/pip-list-dev-tools.txt
+    script: check/pytest --benchmark-only
```
viathor: We should probably keep `--ignore=cirq/contrib` here.

.travis.yml
```diff
@@ -46,4 +46,13 @@ matrix:
     install:
       - pip install -r requirements.txt
       - pip install -r dev_tools/conf/pip-list-dev-tools.txt
-    script: check/pytest --ignore=cirq/contrib
+    script: check/pytest --ignore=cirq/contrib --benchmark-skip
+
+
+    - os: linux
+    env: NAME=Performance tests
+    python: "3.7"
```
viathor: Why run performance tests with a different python version (3.7 vs 3.6)?

.travis.yml
vtomole: Left it like this if we want to benchmark conrib programs (Do we want to?)

.travis.yml
vtomole: Wanted to have a variety of interpreters in travis cause users do; so it increases the chances of us catching version dependent problems. Now that you mention it, different interpreters can have different performances! I will make it consistent as you suggest.

.travis.yml
viathor: Our CI setup skips testing contrib for correctness (skipped above), so we should probably also skip testing contrib for performance.

.travis.yml
viathor: I see. Your argument about covering more interpreter versions makes sense to me. Either is OK.

vtomole:
I tagged you for `pytest-benchmark` feedback based on the compatibility issues you've been experiencing.

.travis.yml
Strilanc: I disagree that we should skip contrib perf tests. The without contrib check is to ensure that we don't accidentally depend on contrib dependencies in the main codebase. We only need one check of this style.

Strilanc:
Travis doesn't seem to be running the checks... weird.

I wonder if we tripped over a security feature, where they want to avoid people using it to mine bitcoins?

Strilanc:
Ah, nope, the yaml doesn't parse:

image

.travis.yml
```diff
@@ -46,4 +46,13 @@ matrix:
     install:
       - pip install -r requirements.txt
       - pip install -r dev_tools/conf/pip-list-dev-tools.txt
-    script: check/pytest --ignore=cirq/contrib
+    script: check/pytest --ignore=cirq/contrib --benchmark-skip
+
+
+    - os: linux
```
Strilanc: This line is indented two spaces too far, and preventing the file from parsing.

vtomole: Done.

PR #1602, opened 2019-05-07
QASM importer - first version

### LINKED ISSUE #1548: Add QASM importer
As the other leg of #44 and maybe a partial solution to #862 depending on the gate sets - as we discussed on today's sync meeting, a QASM importer would be useful. 

I'm happy to design and implement it.

balopat:
Closes #1548. 

The first iteration of a QASM importer. It's a very simple subset of a possible fully featured QASM importer 
- some features are not supported, for example: classical control, arbitrary gate definitions and even some of the gates that don't have a one-to-one representation in Cirq are not yet supported.
 - The functionality should be sufficient to import interesting quantum circuits. 
 - Error handling is very simple - on any lexical or syntactical error a `QasmException` is raised

As I placed it in contrib, it requires the contrib requirements to be installed (ply==3.4).


cirq/contrib/__init__.py
```diff
@@ -20,3 +20,4 @@
 
 from cirq.contrib import acquaintance
 from cirq.contrib.qcircuit import circuit_to_latex_using_qcircuit
+from cirq.contrib.qasm_import import QasmCircuitParser
```
vtomole: newline

cirq/contrib/contrib-requirements.txt
```diff
@@ -2,3 +2,4 @@
 
 pylatex==1.3.*
 tensorflow
+ply==3.4
```
vtomole: Make it alphabetical (before pylatex)

balopat:
Not sure how Travis got into this state of duplicated build suites with one not progressing but at least one of them are passing:)

cirq/circuits/circuit.py
```diff
@@ -1559,6 +1559,11 @@ def save_qasm(self,
         """
         self._to_qasm_output(header, precision, qubit_order).save(file_path)
 
+    @classmethod
+    def from_qasm(cls, qasm: str):
+        from cirq.contrib import QasmCircuitParser
```
Strilanc: Don't reference contrib from inside non-contrib.

cirq/circuits/qasm_output.py
```diff
@@ -69,6 +69,12 @@ def _unitary_(self) -> np.ndarray:
         ]
         return linalg.dot(*map(protocols.unitary, operations))
 
+    def __eq__(self, other):
+        return isinstance(other, QasmUGate) and \
```
Strilanc: Use parens instead of `\` to combine lines.

cirq/circuits/circuit_test.py
```diff
@@ -3056,3 +3056,16 @@ def test_moment_groups():
 (0, 7): ────H──────H─────────────────────
            └──┘   └───┘   └───┘   └──┘
 """, use_unicode_characters=True)
+
+
+def test_from_qasm():
```
Strilanc: You should add a consistency test asserting that the QASM we produce can be parsed.

Ideally it would be the case that parse(x) == parse(produce(parse(x))).

cirq/circuits/circuit.py
balopat: ah, alright, I can totally remove this. 
I figured it would be a useful utility method.

cirq/circuits/circuit_test.py
balopat: That should be possible on the unitary level, is that what you're thinking ? It will be more challenging on the actual gate level - we'll have to implement the inverse of the `__decompose__` functions in an unambiguous way for that.

cirq/circuits/qasm_output.py
balopat: ack. I did that too many places, I'll review and replace.

balopat:
Yeah, I recognize that this looks like a large change with it's 1700 lines added, even though structurally simple:
- a lexer that simply creates tokens based on regexes ~ 100 lines 
- a parser that builds a circuit during parsing expressions  ~ 400 lines 

The rest of it is tests...

I can slice it up to smaller chunks if that's easier to review but then
 - it's harder to see the whole picture
 - functionally it's not going to be useful until we get to this point 

One way to go would be by going instruction-by-instruction and then gate-by-gate and increase the end-to-end (lexer + parser + test) recognized language PR by PR: 

1) OPENQASM header  -> empty circuit + version format error handling 
2) comments 
3) include qelib1.inc statement 
4) quantum register definition 
5) classical register definition 
6) CX statement - this is bigger (requires argument list parsing), maybe split into two: 
   6.a) qubits only 
   6.b) registers and qubits 
7.) one qubit, unparameterized gates (x,y,z,h,s,t,sdg,tdg) 
8.) two qubit, unparameterized gates (cx, swap) 
9.) three qubit, unparametrized gates (ccx, cswap) 
10.) one qubit, one parameter gates - rotation gates (rx,ry,rz)  -- introduces parameters and expressions - so...
   10. a.) floats and pi as parameters 
   10. b.) binary expressions 
   10. c.) unary expressions 
   10. d.) external functions 
11.) one qubit three parameter gates - u3 gate 
12.) U statement (same as u3 gate) 

So at the finest granularity this could take up to 16 PRs. That would be probably very slow to review and get to a working state though (~10-16 weeks?)...

We can look at this also as "just a contrib", and advertise it as being very experimental, making it lower-risk but making progress with it and take in higher chunks. 

I'm open for any suggestions - I think this is useful, but I'm in no rush to push this in. There shouldn't be too much merge conflicts with this code being pretty independent either.

balopat:
Closing this in favor for smaller PRs. If people are interested, please feel free to use this branch as a prototype - it should be fully functional, any feedback is welcome!

Issue #1506, opened 2019-04-09
Define an `equal_up_to_global_phase` protocol


Strilanc:
A class implementing `_equal_up_to_global_phase_(self, atol: float)` should allow comparisons that ignore global phase to be made. The utility method `cirq.equal_up_to_global_phase` would first try for this method, then fallback to `cirq.approx_eq`.

c-poole:
This has been completed

PR #1603, opened 2019-05-07
defines an `equal_up_to_global_phase` protocol

peterse:
Addresses #1506 

my main concern is that this implementation doesn't support typecasting, so `equal_up_to_global_phase(1.0, 1.0j)` will return "False". I feel that this should return "True", but that would lead to inconsistent behavior when this protocol falls back on `approximate_equality` (which does not support typecasting for comparing complex vs. float vs. int)


cirq/protocols/equal_up_to_global_phase.py
```diff
@@ -0,0 +1,171 @@
...
+        val: Source object for approximate comparison.
+        other: Target object for approximate comparison.
+        atol: The minimum absolute tolerance. This places an upper bound on
+        the differences in _magnitudes_ of two compared complex numbers.
+
+    Returns:
+        True if objects are approximately equal up to phase, False otherwise.
+    """
+
+    # fall back to _equal_up_to_global_phase_ for val.
```
Strilanc: Not really a fallback if it comes first.

cirq/protocols/equal_up_to_global_phase.py
```diff
@@ -0,0 +1,171 @@
...
+    other_eq_up_to_phase_getter = getattr(
+        other, '_equal_up_to_global_phase_', None
+    )
+    if other_eq_up_to_phase_getter is not None:
+        result = other_eq_up_to_phase_getter(val, atol)
+        if result is not NotImplemented:
+            return result
+
+    # Try to compare the magnitude of two complex numbers.
+    # FIXME: why restrict casting int, float -> complex for comparison?
```
Strilanc: I agree. Merge them all together into a common thing where their `abs` is compared. (Use the builtin `abs` instead of `np.abs`).

cirq/protocols/equal_up_to_global_phase.py
```diff
@@ -0,0 +1,171 @@
...
+                return True
+
+        try:
+            other_next = next(other_it)
+        except StopIteration:
+            return False
+        # check that a single value describes all phase differences between
+        # corresponding values in the containers
+        if global_phase is None:
+            global_phase = _phase_difference(val_next, other_next)
```
Strilanc: This check is not correct, because it will fail when the first two elements are zero (and so appear to have the same phase even if the two sequences don't).

I think we should avoid handling arbitrary iterables, because it's not clear to me whether users will think that the "up to phase" means per-pair or across-whole-array. We could potentially support numpy arrays, because I think in that case it is clear what is meant and also we have existing code this case in linalg.

peterse: image

^ not sure what this "mypy" test is complaining about or how to address it.

Strilanc:
I've made a branch `equal_up_to_global_phase` that fixes the type errors, and also cleans up some issues with the linalg methods. Pull it into yours (`git pull origin equal_up_to_global_phase`).

The relevant commit: https://github.com/quantumlib/Cirq/commit/728507aa54bd949ff82f7c10e83389997337ce58

cirq/protocols/equal_up_to_global_phase.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+
+
+def equal_up_to_global_phase(val: Any,
+                             other: Any,
+                             *,
+                             atol: Union[int, float] = 1e-8) -> bool:
+    """Determine whether two objects are equal up to global phase.
+
+    If `val` implements SupportsEqualUpToGlobalPhase protocol then it is
```
Strilanc: Document in terms of the method to implement, not the name of the protocol, as this is more direct information for the user.

cirq/protocols/equal_up_to_global_phase.py
```diff
@@ -0,0 +1,107 @@
...
+    invoked and takes precedence over all other checks:
+     - For complex primitive type the magnitudes of the values are compared.
+     - For `val` and `other` both iterable of the same length, consecutive
+       elements are compared recursively. Types of `val` and `other` does not
+       necessarily needs to match each other. They just need to be iterable and
+       have the same structure.
+     - For all other types, fall back to _approx_eq_
```
Strilanc: Put backticks "`" around method names, especially when they are underscored (because markdown will interpret this as italics).

cirq/protocols/equal_up_to_global_phase.py
```diff
@@ -0,0 +1,107 @@
...
+
+    Args:
+        val: Source object for approximate comparison.
+        other: Target object for approximate comparison.
+        atol: The minimum absolute tolerance. This places an upper bound on
+        the differences in _magnitudes_ of two compared complex numbers.
```
Strilanc: Use `*` for emphasis instead of _, to avoid ambiguity with method names.

cirq/protocols/equal_up_to_global_phase.py
```diff
@@ -0,0 +1,107 @@
...
+
+    Returns:
+        True if objects are approximately equal up to phase, False otherwise.
+    """
+
+    # attempt _equal_up_to_global_phase_ for val.
```
Strilanc: Nit: we write comments like sentences, with capitalization.

(Also we use periods. You do that here but not elsewhere.)

cirq/protocols/equal_up_to_global_phase_test.py
```diff
@@ -0,0 +1,120 @@
...
+    assert cirq.equal_up_to_global_phase(1.0 + 1j, 1.0 - 1j, atol=1e-09)
+    assert cirq.equal_up_to_global_phase(np.exp(1j * 3.3),
+                                         1.0 + 0.0j,
+                                         atol=1e-09)
+    assert cirq.equal_up_to_global_phase(np.exp(1j * 3.3), 1.0, atol=1e-09)
+    assert cirq.equal_up_to_global_phase(np.exp(1j * 3.3), 1, atol=1e-09)
+    assert cirq.equal_up_to_global_phase(1j, 1, atol=1e-10)
+    # atol is determined by magnitude of complex vector, not components
+    assert cirq.equal_up_to_global_phase(0.0, 1e-10, atol=1e-09)
+    assert cirq.equal_up_to_global_phase(0, 0, atol=1e-09)
```
Strilanc: You don't check any not-equal cases here. If the code said "if int then return True", this test would pass.

Cause the equality to fail both due to the tolerance getting higher, and due to the values being very different.

cirq/protocols/equal_up_to_global_phase_test.py
```diff
@@ -0,0 +1,120 @@
...
+def test_equal_up_to_global_numeric_iterables():
+    # TODO: tuples, lists, numpy arrays
+    assert cirq.equal_up_to_global_phase([], [], atol=0.0)
+    assert cirq.equal_up_to_global_phase([[]], [[]], atol=0.0)
+    assert cirq.equal_up_to_global_phase([1j, 1], [1j, 1])
+    assert cirq.equal_up_to_global_phase([1j, 1j], [1, 1])
+    assert cirq.equal_up_to_global_phase([1j, 1], [1, -1j])
+    assert not cirq.equal_up_to_global_phase([1j, -1j], [1, 1])
+    assert not cirq.equal_up_to_global_phase([1j, 1], [1, 1j])
+    assert not cirq.equal_up_to_global_phase([1j, 1], [1j, 1, 0])
```
Strilanc: Verify that the tolerance argument matters.

cirq/protocols/equal_up_to_global_phase.py
peterse: done

PR #1605, opened 2019-05-08
Add += shortcut for circuit.append()

vtomole:
For less typing

```python
import cirq

# Pick a qubit.
qubit = cirq.GridQubit(0, 0)

# Create a circuit

circuit = cirq.Circuit()

circuit += cirq.X(qubit)**0.5
circuit += cirq.measure(qubit, key='m')

print("Circuit:")
print(circuit)

# Simulate the circuit several times.
simulator = cirq.Simulator()
result = simulator.run(circuit, repetitions=20)
print("Results:")
print(result)

```


Strilanc:
I'm not sure about this. This meaning of `+=` clashes with numpy (where += means to do per-item operations) and with python lists (where += means to create a new concatenated list and assign it).

mpharrigan:
I find it aesthetically pleasing

vtomole:
@mpharrigan I forgot to add that this is adopted from Pyquil ;)

dabacon:
I don't find it odd that it differs from numpy...broadcast is not what I'd naturally think of.

So I guess what @Strilanc is asking for is that the thing being added needs to be list of ops?
circuit += [cirq.X(a), cirq.H(b)]
It's slightly more verbose, but also has the benefit of allowing you to append more than one op at a time.

(I thought += doesn't create a new list in python? https://stackoverflow.com/questions/2347265/why-does-behave-unexpectedly-on-lists)

Strilanc:
> I thought += doesn't create a new list in python?

Oh!

In that case I would be satisfied if `circuit += [op1, op2, ...]` and `circuit += circuit2` caused append operations. I am still wary of `circuit += op`... but I guess if we generalize to optrees then it makes sense.

cirq/circuits/circuit.py
```diff
@@ -216,6 +216,9 @@ def __delitem__(self, key: Union[int, slice]):
         del self._moments[key]
 
     def __iadd__(self, other):
+        if isinstance(other, ops.gate_operation.GateOperation):
```
Strilanc: Can this just call `self.append(other); return self` and get the same behavior? Maybe after a bit of minimal type checking (if it's not an operation and not iterable then return not implemented).

Strilanc: @vtomole ping

vtomole: Thanks for the ping! Missed the notification. Are you referring to something of this nature?

```python
def __iadd__(self, other):
        if not isinstance(other, ops.gate_operation.GateOperation) and not isinstance(other, Iterable):
            return NotImplemented
        if hasattr(other, 'device'):
            if (other.device != self._device and other.device != devices.UnconstrainedDevice):
                raise ValueError("Other circuit's device is not compatible.")
        if isinstance(other, Iterable):
            for moment in other:                      
                self._device.validate_moment(moment)
            self._moments += other._moments
            return self        
        self.append(other)
        return self
```

Strilanc: The append call already does validation. I mean literally just

```
def __iadd__(self, other):
    self.append(other)
    return self
```

Strilanc: I don't think that we need to check that other.device is similar, because the asymmetric nature of iadd makes it clear which should win.

vtomole: PTAL

PR #1606, opened 2019-05-08
Remove Python 2 references in setup.py, snippets_tests.py check/py, a…

vtomole:
…nd simple_check.sh


setup.py
```diff
@@ -33,37 +33,16 @@
...
-    description=description,
-    long_description=long_description,
-    packages=cirq_packages,
-    package_data={'cirq.api.google.v1': ['*.proto']})
+setup(name='cirq',
+      version=__version__,
+      url='http://github.com/quantumlib/cirq',
+      author='The Cirq Developers',
+      author_email='cirq@googlegroups.com',
+      python_requires=('!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*, !=3.5.0, '
```
Strilanc: This technically allows python 1...

Change it to just be `>=3.5.2`.

vtomole: PTAL

PR #1612, opened 2019-05-08
Linear combination of gate operations

viathor:
Follow-up to #1439 and #1481 and a step towards #491 and #932. Follow-ups to this PR:
 * arithmetic overrides for `GateOperation`, `PauliString` etc,
 * products,
 * raising combinations of Paulis to integer powers,
 * exponentiation.


cirq/ops/linear_combinations.py
```diff
@@ -111,3 +111,127 @@ def _pauli_expansion_(self) -> value.LinearDict[str]:
...
+    Rather than creating LinearCombinationOfGateOperations instance explicitly,
+    one may use overloaded arithmetic operators. For example,
+
+        cirq.LinearCombinationOfGateOperations({cirq.X(q1): 2, cirq.Z(q2): -2})
+
+    is equivalent to
+
+        2 * cirq.X(q1) - 2 * cirq.Z(q2)
+
+    TODO: Overloaded arithmetic operators are not implemented yet.
```
Strilanc: Open an issue instead of adding a TODO.

cirq/ops/linear_combinations.py
```diff
@@ -111,3 +111,127 @@ def _pauli_expansion_(self) -> value.LinearDict[str]:
         for gate, coefficient in self.items():
             result += protocols.pauli_expansion(gate) * coefficient
         return result
+
+
+class LinearCombinationOfGateOperations(
```
Strilanc: Why is this specific to `GateOperation` instead of `Operation` more generally?

cirq/ops/linear_combinations_test.py
```diff
@@ -172,5 +181,356 @@ def test_in_place_manipulations(gates):
...
+         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1.1, 0, 0],
+         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0.1, 0],
+         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1.1],
+     ])),
+    ({
+        cirq.I(q3): 0.1,
+        cirq.FREDKIN(q0, q1, q2): 1
+    },
+     np.array([
+         [1.1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
```
Strilanc: `np.diag([1.1] * 16)`

cirq/ops/linear_combinations.py
viathor: Due to fairly limited functionality of bare `Operation`, linear combinations of bare `Operations` wouldn't have much interesting to offer beyond what is already in `LinearDict`, but supporting them would require adding runtime type switches for the interesting things that we can provide for `GateOperations`.

Whenever a linear combination of `Operation`s is needed, one should just use `LinearDict[Operation]`.

(That said, it's very easy to generalize this class to support `Operation` if/when we discover uses for this. I decided to start more concrete and defer generalization.)

cirq/ops/linear_combinations.py
viathor: Filed #1623.

cirq/ops/linear_combinations_test.py
viathor: There are non-zero off-diagonal elements in this and all other matrices here. I notice that imaginary numbers are often highlighted in a different color and used this fact to make those stand out.

cirq/ops/linear_combinations.py
Strilanc: I don't agree with this.

We have too much code that is specific to GateOperation. There have been a few recent bugs due to this fact.

I don't see any functionality in the class, as it is implemented, that is specific to GateOperation. In particular, you never access the `gate` property!

viathor: Good point. I wrongly assumed other operation types wouldn't have much to offer in terms of unitaries or Pauli expansions, but in fact they do.

Done.

viathor:
There is only one test failure:

```
TypeError: object of type '<class 'cirq.ops.pauli_string.PauliString'>' has no _unitary_ method.
```

which will go away after #1656 or #1661.

cirq/ops/linear_combinations.py
```diff
@@ -111,3 +111,114 @@ def _pauli_expansion_(self) -> value.LinearDict[str]:
...
+                 terms: Mapping[raw_types.Operation, value.Scalar]) -> None:
+        """Initializes linear combination from a collection of terms.
+
+        Args:
+            terms: Mapping of gate operations to coefficients in the linear
+                combination being initialized.
+        """
+        super().__init__(terms, validator=self._is_compatible)
+
+    def _is_compatible(self, gate: raw_types.Operation) -> bool:
```
Strilanc: gate -> operation

cirq/ops/linear_combinations.py
```diff
@@ -111,3 +111,114 @@ def _pauli_expansion_(self) -> value.LinearDict[str]:
...
+            a = extend(a, qubits[1:], all_qubits[1:])
+            b = extend(b, qubits[1:], all_qubits[1:])
+            c = extend(c, qubits[1:], all_qubits[1:])
+            d = extend(d, qubits[1:], all_qubits[1:])
+            return np.block([[a, b], [c, d]])
+
+        num_dim = 2**len(self.qubits)
+        result = np.zeros((num_dim, num_dim), dtype=np.complex128)
+        for op, coefficient in self.items():
+            u = protocols.unitary(op)
```
Strilanc: This could be made much more efficient by using `apply_unitary`. It would also handle a lot of the expansion logic that you have had to duplicate.

```python
raw = np.eye(num_dim, dtype=np.complex128)
workspace = np.empty_like(raw)
axes = [q_to_i_map[q] for q in op.qubits]
result += protocols.apply_unitary(...Args(raw, workspace, axes))
```

cirq/ops/linear_combinations_test.py
```diff
@@ -172,5 +181,383 @@ def test_in_place_manipulations(gates):
...
+         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0.1, 1, 0],
+         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0.1, 0],
+         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1.1],
+     ])),
+    ({
+        cirq.I(q2): 0.1,
+        cirq.FREDKIN(q0, q1, q3): 1
+    },
+     np.array([
+         [1.1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
```
Strilanc: It's quite hard to see the non-zero elements in these grids...

Maybe something crazy like `o = 0` and using `o`? Or `_`? Not necessary.

viathor: Done.

Used `_` and I think the result is really good. Perhaps we should use it when we hardcode large sparse matrices in tests more generally.

cirq/ops/linear_combinations.py
viathor: Done.

cirq/ops/linear_combinations.py
viathor: Done.

viathor:
All comments including the LGTM condition ("use apply_unitary") are addressed, so I'll just merge it later today if no new feedback comes in.

PR #1613, opened 2019-05-08
The foundations of the QASM import functionality

balopat:
- ply dependency in contrib
 - a lexer and a parser wired up to produce a cirq.Circuit
 - support for some basic statements:
   - OPENQASM format
   - include qelib1.inc
   - qreg and creg registration with duplication checking - we are allowing zero length registers to be created (size is not specified in the spec).
 - syntactical and lexical error handling in place
 - added first example for consistency check, parsing the output of cirq.qasm(circuit) 

This is a piecemeal re-implementation of #1602.


cirq/contrib/contrib-requirements.txt
```diff
@@ -1,4 +1,5 @@
 # Runtime requirements for contrib.
 
+ply==3.4
```
balopat: @Strilanc - what is the recommendation here? Based on the google3 availability, I pinned it to 3.4. However qiskit uses the latest and breaks with 3.4 which is old, so users who have qiskit installed will have some issues if they run `pip install -e cirq/contrib/contrib-requirements.txt`

vtomole: CCing @dstrain115  on this.

Strilanc: You may have to contact the person responsible for the google3 import and get them to sync.

balopat: We talked this through with @dstrain115, ply 3.4 should be fine as a dependency. 
I came to the conclusion that `ply>=3.4` might be the right way to go, if we don't want to break people who want to install cirq into an env that already has qiskit installed. 
However, the QASM importer should be tested against 3.4 in Travis to make sure it works internally too. So I would leave this as `ply ==3.4`, unless users ask for loosening the range.

cirq/contrib/qasm_import/_lexer.py
```diff
@@ -0,0 +1,86 @@
...
+from cirq.contrib.qasm_import.exception import QasmException
+
+
+class QasmLexer(object):
+    NATURAL_NUMBER = "NATURAL_NUMBER"
+
+    def __init__(self, qasm: str):
+        self.qasm = qasm
+        self.lex = lex.lex(object=self, debug=False)
+        self.lex.input(qasm)
```
Strilanc: This is doing things that make more sense in a function than in a class initializer. Why is lexing tied to the construction of the lexer? Is this class a thing-that-lexes or is it a thing-that-has-been-lexed? Is it both? I don't think it should be both.

cirq/contrib/qasm_import/_parser_test.py
```diff
@@ -0,0 +1,153 @@
...
+import cirq.testing as ct
+from cirq import Circuit
+from cirq.contrib.qasm_import import QasmException
+from cirq.contrib.qasm_import._parser import QasmParser
+
+
+def test_format_header_circuit():
+    parser = QasmParser("OPENQASM 2.0;")
+
+    parsed_qasm = parser.parse()
```
Strilanc: Shouldn't the source go into the parse method, not the constructor?

cirq/contrib/qasm_import/qasm.py
```diff
@@ -0,0 +1,29 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from cirq import circuits
+from cirq.contrib.qasm_import._parser import QasmParser
+
+
+class QasmCircuitParser:
+    """QasmCircuitParser is currently partially developed, not functional,
+    DO NOT USE.
```
Strilanc: Put a TODO(issue link)

cirq/contrib/qasm_import/_lexer.py
```diff
@@ -0,0 +1,86 @@
...
+import re
+from typing import Optional
+
+import ply.lex as lex
+
+from cirq.contrib.qasm_import.exception import QasmException
+
+
+class QasmLexer(object):
+    NATURAL_NUMBER = "NATURAL_NUMBER"
```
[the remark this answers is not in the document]
balopat: oops, no, I wanted to use it in the tests - but then did not use it in the end. 
I'll remove it.

cirq/contrib/qasm_import/_lexer.py
balopat: The reason I did it because I was thinking about this as a one-time use object alongside with the parse. The user shouldn't have to call input multiple times and this way I don't have to think about "what if the user passes in another qasm string, how do I validate that I'm still in a valid state?". 

Another inspiration is that a lexer is nothing else than a fancy tokenizer (in python it's a [generator function](https://docs.python.org/3/library/tokenize.html), in Java it's a one-time use [class](https://docs.oracle.com/javase/7/docs/api/java/util/StringTokenizer.html)) : 

`lex.input` doesn't do the lexing - it only sets up the initial state for lexing, after which the `token()` function can stream the :

```
    def input(self, s):
        # Pull off the first character to see if s looks like a string
        c = s[:1]
        if not isinstance(c, StringTypes):
            raise ValueError('Expected a string')
        self.lexdata = s
        self.lexpos = 0
        self.lexlen = len(s)
```

If you feel strongly about it - I'm happy to rewire it, it will work that way too of course - it's just a tiny bit more code.

cirq/contrib/qasm_import/_parser_test.py
balopat: Same argument as in the lexer (above) - the initialization is setup in the constructor for a one-time use object. I'm happy to move it around if you feel strongly about it.

cirq/contrib/qasm_import/qasm.py
balopat: Ack!

cirq/contrib/qasm_import/_lexer.py
Strilanc: I think it should be split out. Rule of thumb: for testability reasons, you generally want to avoid doing work in constructors ( see this: https://youtu.be/RlfLCWKxHJ0?t=332 ). Don't ask for stuff that allows you to make the thing you need, just directly ask for the thing you need.

Admittedly this is not always great. And we do break this rule. But it's a good rule of thumb.

balopat: Sounds good - I refactored it.

PR #1614, opened 2019-05-08
Pauli expansion for FSim

viathor:


cirq/ops/fsim_gate.py
```diff
@@ -79,6 +79,21 @@ def _unitary_(self) -> Optional[np.ndarray]:
             [0, 0, 0, c],
         ])
 
+    def _pauli_expansion_(self) -> value.LinearDict[str]:
+        if protocols.is_parameterized(self):
```
dabacon: aside: this makes me think that we could allow linear dict coefficients to be Symbols?

viathor: Good idea, thanks! Filed #1617 to track.

PR #1615, opened 2019-05-08
Added swap networks example

bryano:


examples/swap_networks.py
```diff
@@ -0,0 +1,83 @@
+"""Demonstrates swap networks."""
```
vtomole: So this demonstrates swap networks by generating a QAOA circuit as an example?  If so, is this another way of implementing https://github.com/quantumlib/Cirq/pull/1323/files?

@Strilanc  This uses `contrib` and i'm wondering if we should have rules about using `contrib` procedures in our examples.

examples/swap_networks.py
```diff
@@ -0,0 +1,83 @@
...
+        edges: The list of edges.
+        beta: The mixing angle.
+        gamma: The phase separation angle.
+
+    Returns:
+        A circuit implementing a single round of QAOA with the specified
+        angles.
+    """
+
+    ## G_{i,j} ∝ exp(i gamma (|01><01| + |10><10|))
```
vtomole: I'm not familiar with this double comment `##` convention.

examples/swap_networks.py
vtomole: It also does more than QAOA. Apologies.

bryano: It does both more and less than QAOA (i.e. no optimization, evaluation, etc.). The goal was to just show how to use the swap network code, and QAOA is a common, simple use case.

I'm fine with something like a "contrib examples" directory, though it doesn't seem necessary to me. (The examples are standalone files, so there's no risk of introducing indirect dependencies.)

examples/swap_networks.py
bryano: That's a very diplomatic way of pointing out a typo :)

Strilanc:
I think it's fine if examples use contrib.

examples/swap_networks.py
Strilanc: Describe in a lot more detail the purpose and mechanism of the example. Users are going to be looking at this file with zero context.

examples/swap_networks.py
```diff
@@ -0,0 +1,83 @@
...
+        beta: The mixing angle.
+        gamma: The phase separation angle.
+
+    Returns:
+        A circuit implementing a single round of QAOA with the specified
+        angles.
+    """
+
+    # G_{i,j} ∝ exp(i gamma (|01><01| + |10><10|))
+    phase_sep_gates = {edge: cirq.ZZ**gamma for edge in edges}
```
babbush: So I was just playing around with this code and I want to point out what I view as a somewhat insidious trap I ran into. Basically, I am using a different data structure for representing the graphs I want to simulate and so for various reasons, it was natural for me to keep track of edges in a slightly different way. I built this dictionary using keys that represented edges between qubits and looked like (cirq.LineQubit(i), cirq.LineQubit(j)) as opposed to edges that looked like (i, j). This did not raise an error when I called the GreedyExecutionStrategy. Instead, it just ignored the gates I was trying to insert. Instead, I think it should raise an exception that says something like "error, your edges look wrong". Let me know if what I'm saying is clear and if you agree. To be clear, I think the problem here is more with the code that silently ignores dictionary entries that must live in GreedyExecutionStrategy, not in the demo per se.

examples/swap_networks.py
```diff
@@ -0,0 +1,83 @@
+"""Demonstrates swap networks."""
+
+import itertools
+import random
+
+import cirq
+import cirq.contrib.acquaintance as cca
+
+
+def get_random_graph(n_vertices, edge_prob=0.5):
```
viathor: The code would be easier to read if the reader knew how vertices, edges etc are represented. Maybe add type annotations throughout.

This may have helped avoid the trap described by @babbush below.

examples/swap_networks.py
viathor: One tool that could help here is type annotations. They inform whoever reads or modifies the code how vertices, edges etc are represented here. See comment above.

bryano: @babbush It's actually perfectly fine to use keys that are tuples of qubits rather than integers; the initial mapping just needs to do that as well. The example code now shows both ways.

@viathor Type annotations added. These are actually just copied and pasted from cirq/contrib/acquaintance/permutation.py, but that seemed more concise than importing them and then having their definitions in a comment.

examples/swap_networks.py
bryano: I added some explanation. Let me know if you'd like more.

Is there a style convention for example docstrings? What I wrote is a weird mix of python, unicode, and (unrendered) tex. Rendered latex would be nicest, but right now the example files are shown in RTD verbatim.

examples/swap_networks.py
```diff
@@ -0,0 +1,158 @@
...
+    # opportunity gates.
+    circuit = cca.complete_acquaintance_strategy(qubits, 2)
+    if verbose:
+        print('Circuit containing a single complete linear swap network:')
+        print(circuit)
+        print('\n\n')
+
+    acquaintance_opportunities = cca.get_logical_acquaintance_opportunities(
+        circuit, initial_mapping)
+    assert set(frozenset(edge) for edge in gates) <= acquaintance_opportunities
```
bryano: @babbush I'm hesitant to make GreedyExecutionStrategy assert that it has inserted every logical gate, because there might be use cases in which that's not what you want, but this does essentially the same thing, by asserting that it has the opportunity to insert every logical gate.

bryano:
Is this ready to be merged?

vtomole:
Looks like you've addressed @babbush comments. Let's merge once he approves this.

vtomole:
A 3 day wait is enough IMO.

babbush:
Yes, LGTM!

PR #1616, opened 2019-05-09
Remove remaining references to 'python 2'

vtomole:
I don't remove the `_cmp_` tests because i think they are nice to have documentation (they don't affect coverage). 

With regards to the circular dependency issue, after tinkering with it, it seems to me like it's not specific to Python 2. Please let me know otherwise.

### DIFF
--- a/cirq/schedules/schedule_test.py
+++ b/cirq/schedules/schedule_test.py
@@ -178,8 +178,6 @@ def test_query_overlapping_operations_exclusive():
     assert schedule.query(time=zero + 3.5 * ps, duration=ps) == []
 
 
-# In python 2, timedelta doesn't support multiplication with floats
-
 def test_query_timedelta():
     q = cirq.NamedQubit('q')
     zero = cirq.Timestamp(picos=0)
--- a/cirq/type_workarounds.py
+++ b/cirq/type_workarounds.py
@@ -15,7 +15,6 @@
 from typing import Any
 
 # At the moment there's no reliable way to say 'NotImplementedType'.
-# - There is a NotImplementedType in python 2, but not python 3.
 # - type(NotImplemented) causes mypy to error.
 # - Just NotImplemented causes runtime errors (it's not a type).
 # - The string "NotImplemented" causes runtime errors (in some python versions).
--- a/cirq/value/duration_test.py
+++ b/cirq/value/duration_test.py
@@ -84,10 +84,6 @@ def test_cmp():
     assert (Duration() != 0)
 
 
-# In python 2, comparisons fallback to __cmp__ and don't fail.
-# But a custom __cmp__ that does fail would result in == failing.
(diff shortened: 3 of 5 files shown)


cirq/value/duration_test.py
```diff
@@ -84,10 +84,6 @@ def test_cmp():
     assert (Duration() != 0)
 
 
-# In python 2, comparisons fallback to __cmp__ and don't fail.
-# But a custom __cmp__ that does fail would result in == failing.
-# So we throw up our hands and let it be.
```
viathor: This raises the question: do we want to implement comparability of `Duration` (and `Timestamp`) with other types now that we're off python 2? If so, we should file an issue to track before removing this comment.

cirq/schedules/schedule_test.py
```diff
@@ -178,8 +178,6 @@ def test_query_overlapping_operations_exclusive():
     assert schedule.query(time=zero + 3.5 * ps, duration=ps) == []
 
 
-# In python 2, timedelta doesn't support multiplication with floats
```
viathor: IIUC, this comment was added to justify an absence of a unit test that would not pass in python 2. Maybe implement it now to replace the comment you're removing?

vtomole: https://github.com/quantumlib/Cirq/issues/1618

cirq/type_workarounds.py
```diff
@@ -15,7 +15,6 @@
 from typing import Any
 
 # At the moment there's no reliable way to say 'NotImplementedType'.
-# - There is a NotImplementedType in python 2, but not python 3.
```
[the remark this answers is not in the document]
vtomole: @Strilanc I'm assuming this is for documentation purposes? I mean, instead of using `Any` where we meant `NotImpemented`?

cirq/value/duration_test.py
vtomole: CC @Strilanc

Strilanc: What other values could timestamp be compared with?

I don't see any reason not to define comparison between duration and the builtin timedelta type.

cirq/type_workarounds.py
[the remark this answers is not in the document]
viathor: nvm, I misread this as saying "There is a NotImplementedType in python 3, but not python 2." (which would suggest we want to switch to `NotImplementedType` now that we're on py3).

cirq/schedules/schedule_test.py
timothyman: Actually, that comment was there to clarify the now missing annotation that that test should not run (and will not pass) in Python 2 because timedelta doesn't support multiplication with floats. The test itself uses that functionality and passed in Python 3. AFAIU, #1618 can be closed?

PR #1620, opened 2019-05-10
Adding PhaseXPowGate into the ion trap native gate set

mlihc15:


cirq/ion/ion_device.py
```diff
@@ -195,4 +194,4 @@ def _verify_unique_measurement_keys(operations: Iterable[ops.Operation]):
             key = protocols.measurement_key(meas)
             if key in seen:
                 raise ValueError('Measurement key {} repeated'.format(key))
-            seen.add(key)
+            seen.add(key)
```
dabacon: nit: missing a newline at end of file

PR #1627, opened 2019-05-13
Swap deprecation flag with comment in Program proto.

kunalq:
Until we get `RunContext`s plumbed through Cirq, we'll rely on implementers to handle deprecation warnings, as the current compile-time deprecations are causing hard errors in our build system.


viathor:
Berore you merge, can you check that the `deprecated` proto option doesn't affect generated code?

kunalq:
That's a great point, @viathor -- checking now

kunalq:
Chatted with @lindmarkm -- we no longer need this :)

PR #1638, opened 2019-05-14
Remove `Note: avoiding 'from/as' because it creates a circular dependency in python 2.`

vtomole:


Strilanc:
Not sure why mypy suddenly decided to complain about those particular things.

Adding

```
    def num_qubits(self):
        return 1
```

Into the Pauli class seems to do the trick. Likely results in a coverage failure; feel free to suppress it.

cirq/optimizers/decompositions.py
```diff
@@ -68,7 +69,8 @@ def is_no_turn(half_turns):
         linalg.deconstruct_single_qubit_matrix_into_angles(mat))
     z_ht_before = z_rad_before / np.pi - 0.5
     m_ht = y_rad / np.pi
-    m_pauli = ops.pauli_gates.X  # type: ops.pauli_gates.Pauli
+    m_pauli = ops.pauli_gates.X  # type: ignore \
```
Strilanc: Just shorten this to `ops.X` to fix it. Don't reference a pull request in comments.

Issue #1522, opened 2019-04-11
Introduce SweepTrialResult and return it instead of List[TrialResult] from sweeping sample methods


Strilanc:
For example, this would allow us to support functionality like "give me the results where the sweep variable was 0.25".

smitsanghavi:
There can be multiple sweep variables, right? IF we want to support arbitrary querying like this, the result of such a query on a SweepTrialResult will return another SweepTrialResult without the params specified in the query.

I am thinking of implementing SweepTrialResult as Dict[TrialID, TrialResult] and an index of the form Dict[str(param+val), Set[TrialID]]]. Both of these can be created from List[TrialResult].

GetResults query as above will take in a ParamResolver that describes the query. It then performs lookups on the index for each param in the input resolver, unions the sets and gets a list of TrialResults from the TrialResult dict. It then returns SweepTrialResult(results=List[TrailResults], ignore_params=self.ignore_params+input_resolver.param_dict.keys)

Strilanc:
I do like the idea of having a method like

`sweep_trial_result.slice_where_params_match({'a': 5})`

Strilanc:
Another common issue that people run into is how to count the proportion of times a particular measurement occurred. We should have utility methods for that.

smitsanghavi:
This would be same as a multi_measurement_histogram on the combined set of measurements for all underlying TrialResults, correct?

Strilanc:
I wrote some experimental code that might be useful for this. It has a sort of SQL feel to it.

Usage example:

```python
    x = Results(
        input_keys=('theta', 'phi'),
        output_keys=('x', 'z'),
        data=(
            ResultBin((1, 2), np.array([
                [0.1, 0.4],
                [0.2, 0.3],
            ])),
            ResultBin((3, 2), np.array([
                [0.2, 0.5],
                [0.1, 0.1],
            ])),
        )
    )
    print(str(x))
    print()
    print(str(x.where(theta=3)))
    print()
    print(str(x.where(sympy.Symbol('z') >= 0.2)))
    print()
    y = x.group_by(sympy.Symbol('z') >= 0.2)
    for k, v in y.items():
        print(k)
        print(v)
```

Outputs:

```
  theta phi │ x   z
╶───────────┼─────────
  1     2   │ 0.1 0.4
  1     2   │ 0.2 0.3
╶───────────┼─────────
  3     2   │ 0.2 0.5
  3     2   │ 0.1 0.1
            │

            ╷
[...]

PR #1643, opened 2019-05-15
SweepTrialResult

smitsanghavi:
#1522


cirq/study/sweep_trial_result.py
```diff
@@ -0,0 +1,97 @@
...
+
+from cirq import value
+from cirq.study import resolver, trial_result
+
+if TYPE_CHECKING:
+    # pylint: disable=unused-import
+    import cirq
+
+
+@value.value_equality(unhashable=True)
```
Strilanc: Why unhashable? This isn't a mutable type.

cirq/study/sweep_trial_result.py
```diff
@@ -0,0 +1,97 @@
...
+class SweepTrialResult:
+    """The combination of multiple TrialResults with potentially different
+       set of parameters.
+
+    Attributes:
+        trial_results: A list of TrialResults.
+    """
+
+    def __init__(self,
+                 trial_results: List[trial_result.TrialResult] = None) -> None:
```
Strilanc: I would drop the default value here, because I can't picture cases where someone would construct this class without having run a sweep (except in tests).

cirq/study/sweep_trial_result.py
```diff
@@ -0,0 +1,97 @@
...
+        self.trial_results = trial_results if trial_results is not None else []
+        self._index = {}  # type: Dict[str, Set[int]]
+        for i, result in enumerate(self.trial_results):
+            for key, value in result.params.param_dict.items():
+                self._index.setdefault(key + str(value), set()).add(i)
```
Strilanc: Why not use the tuple `(key, value)` as the index key? The string concatenation is ambiguous.

Strilanc: Use a `defaultdict` type instead of `setdefault`.

cirq/study/sweep_trial_result.py
```diff
@@ -0,0 +1,97 @@
...
+        Returns:
+            A list of TrialResults.
+        """
+        trial_results_match = set(range(len(self.trial_results)))
+        for key, value in resolver.ParamResolver(param_dict).param_dict.items():
+            trial_results_match.intersection_update(
+                self._index.get(key + str(value), set()))
+        return [self.trial_results[i] for i in trial_results_match]
+
+    def slice_where_params_match(self,
```
Strilanc: Refactor this into `where(self, *filters, **keyword_filters)` so that users can call it like this:

```python
results.where(a=2, b=3)
```

Note that `"a"` can *either* be a measurement key or a parameter key. If it's both, we should raise an exception and say to call a more specific method `where_measurements` / `where_params`.

We should still take param resolvers, so that keywords that aren't valid python identifiers can be used:

```python
results.where({'un speakable #5': 4})
```

We may also want to get ambitious and support use cases like this:

```python
results.where(sympy.Symbol('a')**2 >= 3)
```

Where you would would basically notice it was a sympy thing and evaluate it on each case to determine the filter result.

cirq/study/sweep_trial_result.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __init__(self,
+                 trial_results: List[trial_result.TrialResult] = None) -> None:
+        self.trial_results = trial_results if trial_results is not None else []
+        self._index = {}  # type: Dict[str, Set[int]]
+        for i, result in enumerate(self.trial_results):
+            for key, value in result.params.param_dict.items():
+                self._index.setdefault(key + str(value), set()).add(i)
+
+    def trials_where_params_match(
```
Strilanc: This method is strictly worse than `slice_where_params_match`. Drop it.

cirq/study/sweep_trial_result.py
```diff
@@ -0,0 +1,97 @@
...
+
+        Args:
+            param_dict: A dictionary from the ParameterValue key (str) to its
+                assigned value.
+        Returns:
+            A SweepTrialResult limited to TrialResults that match the specified
+            input.
+        """
+        return SweepTrialResult(self.trials_where_params_match(param_dict))
+
```
Strilanc: Implement `__iter__`, `__getitem__` and `__len__` so that this acts a bit more like a list of trial results.

cirq/study/sweep_trial_result.py
```diff
@@ -0,0 +1,97 @@
+# Copyright 2018 The Cirq Developers
```
dabacon: nit: 2019

dabacon: (in other files as well)

Strilanc:
It was mentioned at the weekly meeting that the sorts of things I'm asking for (with e.g `where`) are common operations in the `pandas` library. Maybe we should expose our results that way instead?

smitsanghavi:
Shouldn't have missed this meeting!
I'll look into pandas too

Strilanc:
I think we're going to move in the direction of using pandas for both TrialResult and this multi-trial result thing (in fact we may merge the classes). It makes a lot of things nicer.

Strilanc:
I am going to close this PR for now. Open another when the pandas stuff is ready.

Issue #1630, opened 2019-05-14
DensityMatrixSimulator fails on PauliString phasing operations


Strilanc:
There's a nasty not done TODO.

```python
import cirq
a, b = cirq.LineQubit.range(2)
c = cirq.Circuit.from_ops((cirq.Z(a)*cirq.Z(b))**0.5)
r = cirq.DensityMatrixSimulator().simulate(c)
```

Produces

```
  File "cirq\sim\density_matrix_simulator.py", line 261, in _base_iterator
    gate = cast(ops.GateOperation, op).gate
AttributeError: 'PauliStringPhasor' object has no attribute 'gate'
```

PR #1644, opened 2019-05-15
Use apply_channel in DensityMatrixSimulator

Strilanc:
And make it work on operations, not just gate operations.

Fixes https://github.com/quantumlib/Cirq/issues/1630

### DIFF
--- a/cirq/sim/density_matrix_simulator.py
+++ b/cirq/sim/density_matrix_simulator.py
@@ -30,6 +30,14 @@ if TYPE_CHECKING:
     from typing import Any, Hashable
 
 
+class _StateAndBuffers:
+
+    def __init__(self, n: int, state: np.ndarray):
+        self.n = n
+        self.state = state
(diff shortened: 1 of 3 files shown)


cirq/sim/density_matrix_simulator.py
```diff
@@ -30,6 +30,14 @@
     from typing import Any, Hashable
 
 
+class _StateAndBuffers:
+
+    def __init__(self, n: int, state: np.ndarray):
```
dabacon: num_qubits?  or add doc about what n is.

dabacon: i'd call the inner state "matrix" to remind people it's the density matrix, but that's a nit.

Strilanc: Done.

Strilanc: Done.

PR #1645, opened 2019-05-15
Add cirq.one_hot method

Strilanc:
- Creates a state vector with a single non-zero entry

### DIFF
--- a/cirq/linalg/states.py
+++ b/cirq/linalg/states.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 """Utility methods for creating vectors and matrices."""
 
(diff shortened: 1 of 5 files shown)


cirq/linalg/states.py
```diff
@@ -0,0 +1,41 @@
...
+from typing import Sequence, Union, Type
+
+import numpy as np
+
+
+def one_hot(*,
+            index: Union[None, int, Sequence[int]] = None,
+            shape: Union[int, Sequence[int]],
+            dtype: Type[np.number]) -> np.ndarray:
+    """Returns a numpy array that with a single 1 and otherwise 0.
```
vtomole: Grammar.

Strilanc:
I have other use cases that are coming; this is a pre-emptive PR.

cirq/linalg/states.py
Strilanc: Done.

PR #1646, opened 2019-05-16
approx_eq patch

### LINKED ISSUE #1608: Complex numbers can be approximately equal to integers
but approx_eq incorrectly disagrees

peterse:
Resolves #1608.


cirq/ops/raw_types.py
```diff
@@ -122,6 +122,8 @@ def validate_args(self, qubits: Sequence[Qid]) -> None:
                     repr(self)))
 
         if len(qubits) != self.num_qubits():
+            print(len(qubits))
```
Strilanc: Remove debug code

cirq/protocols/approximate_equality.py
```diff
@@ -86,13 +87,8 @@ def approx_eq(val: Any, other: Any, *, atol: Union[int, float] = 1e-8) -> bool:
...
 
     # Compare primitive types directly.
-    if isinstance(val, (int, float)):
-        if not isinstance(other, (int, float)):
-            return False
-        return _isclose(val, other, atol=atol)
-
-    if isinstance(val, complex):
-        if not isinstance(other, complex):
+    if isinstance(val, numbers.Number):
```
Strilanc: Is numpy compatible for all these values? Does `fractions.Rational` count as a number?

peterse: I've added support for `fractions.Fraction` and `decimal.Decimal` types and added more thorough testing for numpy types.

If there's number types I missed (which obviously I can't think of), `approx_eq` should now fallback to __eq__ instead of throwing an error

PR #1649, opened 2019-05-17
 Switch to just `import cirq` in circuit_test.py

Strilanc:


Strilanc:
We've *been* doing it. There's not many left.

PR #1650, opened 2019-05-17
Add cirq.apply_unitaries method

Strilanc:
- There are a few places in the codebase that repeat this basic pattern, and I expect more
- Add ApplyUnitaryArgs.default helper method
- Use apply_unitaries in ParallelGateOperation and Circuit
- Fix circuit.to_unitary_matrix ignoring desired dtype


cirq/protocols/apply_unitary.py
```diff
@@ -15,7 +15,16 @@
 """A protocol for implementing high performance unitary left-multiplies."""
 
 
-from typing import Any, Union, TypeVar, Tuple, Iterable
+from typing import (
+    Any,
+    Union,
+    TypeVar,
+    Tuple,
+    Iterable,
```
dabacon: nit: abc

cirq/protocols/apply_unitary.py
```diff
@@ -79,6 +91,12 @@ def __init__(self,
         self.available_buffer = available_buffer
         self.axes = tuple(axes)
 
+    @staticmethod
+    def default(num_qubits: int) -> 'ApplyUnitaryArgs':
+        """A default instance starting in state |0⟩."""
+        state = linalg.one_hot(shape=(2,) * num_qubits, dtype=np.complex128)
```
dabacon: be explicit here and set index. otherwise I have to dig to understand why this is 0.

cirq/circuits/circuit.py
```diff
@@ -1307,7 +1307,7 @@ def to_unitary_matrix(
             self.all_qubits().union(qubits_that_should_be_present))
         n = len(qs)
 
-        state = np.eye(1 << n, dtype=np.complex128)
```
dabacon: there should probably have been a test for this.

cirq/protocols/apply_unitary.py
Strilanc: Done

cirq/protocols/apply_unitary.py
Strilanc: Done. Not nearly as clean since you need a multi-dimensional index, though.

PR #1653, opened 2019-05-17
[AFTER #1650] Refactor apply_unitary to go directly to magic methods

Strilanc:
We were beginning to form cycles between the various protocol methods.
It was becoming to difficult to edit the protocol code without causing infinite loops.
This refactor (and others) will remove the seeds that are starting these cycles, at the cost of some extra redundancy in where magic method names are stated.

Comes after #1650


Strilanc:
This spiraled out of control.

PR #1654, opened 2019-05-17
Add travis build for OSX 

vtomole:


Strilanc:
This is a great idea.

cirq/linalg/decompositions_test.py
```diff
@@ -240,6 +240,18 @@ def test_kak_decomposition(target):
...
 
+def test_kak_decomposition_interaction_coefficients():
+    a, b = cirq.LineQubit(0), cirq.LineQubit(1)
+    theta = 0.3
+    circuit = cirq.Circuit()
+    circuit.append(cirq.ISWAP.on(a, b))
+    circuit.append(
+        cirq.ZZPowGate(exponent=2 * theta / np.pi, global_shift=-0.5).on(a, b))
+
+    assert cirq.kak_decomposition(cirq.unitary(circuit)).interaction_coefficients == \
```
Strilanc: Don't use exact equality with floating point numbers.

vtomole: I have removed this because it was an attempt to reproduce the kak bug now that we know the reason we were getting it.

dev_tools/install_macos.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+if [ "$TRAVIS_OS_NAME" == "osx" ]; then
+
+    # Install some custom requirements on macOS
+    # e.g. brew install pyenv-virtualenv
+    export PYTHON=3.7.0
```
Strilanc: Why is this being exported instead of only set?

dev_tools/install_macos.sh
```diff
@@ -0,0 +1,18 @@
...
+    brew update
+    brew install openssl readline pyenv-virtualenv
+    brew outdated pyenv || brew upgrade pyenv
+    pyenv install $PYTHON
+    export PYENV_VERSION=$PYTHON
```
Strilanc: Is it possible to pass this as a flag instead of via an exported variable?

dev_tools/install_macos.sh
```diff
@@ -0,0 +1,18 @@
...
+    export PATH="/Users/travis/.pyenv/shims:${PATH}"
+    pyenv virtualenv venv
+    source /Users/travis/.pyenv/versions/3.7.0/envs/venv/bin/activate
+    mkdir ~/.matplotlib && echo "backend: TkAgg" >> ~/.matplotlib/matplotlibrc
```
Strilanc: Add a comment explaining this line in detail.

.travis.yml
```diff
@@ -56,6 +56,18 @@ matrix:
       - pip install -r cirq/contrib/contrib-requirements.txt
       - pip install -r dev_tools/conf/pip-list-dev-tools.txt
     script: check/pytest --benchmark-only
+
+  - os: osx
+    env: NAME=pytest (macOS)
+    python: "3.7.0"
+    language: generic
+    install:
+      - ./dev_tools/install_macos.sh
```
Strilanc: There is an odd mix of lines here and lines in the bash script. There is some redundancy. What made you put some things here and some there?

dev_tools/install_macos.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+if [ "$TRAVIS_OS_NAME" == "osx" ]; then
```
Strilanc: If this script is specific to the continuous integration, give it a more descriptive filename so that users do not execute it by accident.

.travis.yml
vtomole: I was trying to keep it consistent with the other build steps (keeping the `pip installs` out here for example). Didn't notice that i activate the virtual env outside and inside the script. Will fix that.

dev_tools/install_macos.sh
vtomole: Why would we want to pass it in as a flag if it's always set to `$PYTHON`?

Strilanc: I wanted to avoid setting `$PYTHON`.

dev_tools/continous_integration_macos.sh
```diff
@@ -0,0 +1,19 @@
...
+    brew install openssl readline pyenv-virtualenv
+    brew outdated pyenv || brew upgrade pyenv
+    pyenv install $PYTHON
+    export PYENV_VERSION=$PYTHON
+    export PATH="/Users/travis/.pyenv/shims:${PATH}"
+    pyenv virtualenv venv
+    
+    #`import matplotlib` throws a  RuntimeError without this
+    mkdir ~/.matplotlib && echo "backend: TkAgg" >> ~/.matplotlib/matplotlibrc
+    fi
```
Strilanc: Indentation

.travis.yml
```diff
@@ -59,6 +59,12 @@ matrix:
       - pip install -r cirq/contrib/contrib-requirements.txt
       - pip install -r dev_tools/conf/pip-list-dev-tools.txt
     script: check/performance
+
+  - os: osx
+    env: NAME=pytest (macOS)
+    python: "3.7.0"
+    language: generic      
+    script: ./dev_tools/continous_integration_macos.sh
```
Strilanc: This should be the `install` step, and then the `script` step is `pytest`. You may need to use `source dev_tools/continous_integration_macos.sh` in order for it to work properly.

(Right now you're not actually running the tests!)

dev_tools/continous_integration_macos.sh
```diff
@@ -0,0 +1,19 @@
...
+
+    # Install some custom requirements on macOS
+    # e.g. brew install pyenv-virtualenv
+    PYTHON=3.7.0
+    brew update
+    brew install openssl readline pyenv-virtualenv
+    brew outdated pyenv || brew upgrade pyenv
+    pyenv install $PYTHON
+    export PYENV_VERSION=$PYTHON
+    export PATH="/Users/travis/.pyenv/shims:${PATH}"
```
Strilanc: Can the start of path we are adding be obtained e.g. by using `pwd`? Right now it is *very* specific to the setup travis-ci uses.

dev_tools/continous_integration_macos.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+# This script should only be run by Travis
+if [ "$TRAVIS_OS_NAME" == "osx" ]; then
```
Strilanc: Print an error message and `exit 1` if the os name is not as expected.

dabacon:
What is the status of this?

vtomole:
@dabacon Totally forgot about this. Thanks for the ping. I'll respond to Craig's comments tonight.

vtomole:
@Strilanc I think this is ready to go in. The warnings can be fixed in another PR since I don't have a mac and I've already pushed 90+ commits to this PR. If you have a mac please confirm that you are getting similar warnings.

Strilanc:
Test failure was a flake fixed in https://github.com/quantumlib/Cirq/compare/flakes0?expand=1

vtomole:
Bad bot.

Strilanc:
I think this might be because of a pagination issue caused by the massive number of commits.

Strilanc:
Or it was just the google cla bot being slow.

PR #1661, opened 2019-05-18
Add _unitary_ for PauliString

viathor:
Found it was missing when trying to use PauliString as an Operation in unit tests for #1612.

### DIFF
--- a/cirq/ops/pauli_string.py
+++ b/cirq/ops/pauli_string.py
@@ -20,7 +20,7 @@ import math
 
 import numpy as np
 
-from cirq import value
+from cirq import protocols, value
 from cirq.ops import (
     raw_types,
     gate_operation,
@@ -178,6 +178,12 @@ class PauliString(raw_types.Operation):
             factors.append(str(cast(raw_types.Gate, self[q]).on(q)))
 
         return prefix + '*'.join(factors)
+
+    def _unitary_(self) -> np.ndarray:
+        u = np.array([self.coefficient], dtype=np.complex128)
+        for pauli in self.values():
+            u = np.kron(u, protocols.unitary(pauli))
+        return u
 
     def zip_items(self, other: 'PauliString') -> Iterator[
             Tuple[raw_types.Qid, Tuple[pauli_gates.Pauli, pauli_gates.Pauli]]]:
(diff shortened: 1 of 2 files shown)


cirq/ops/pauli_string.py
```diff
@@ -179,6 +179,12 @@ def __str__(self):
 
         return prefix + '*'.join(factors)
 
+    def _unitary_(self) -> np.ndarray:
+        u = np.array([self.coefficient], dtype=np.complex128)
```
Strilanc: Need to return `None` when the coefficient is not on the complex unit circle. Add a test checking this.

cirq/ops/pauli_string.py
```diff
@@ -179,6 +179,12 @@ def __str__(self):
 
         return prefix + '*'.join(factors)
 
+    def _unitary_(self) -> np.ndarray:
```
Strilanc: Add a `_has_unitary_` method also.

cirq/ops/pauli_string.py
```diff
@@ -179,6 +179,12 @@ def __str__(self):
...
+        u = np.array([self.coefficient], dtype=np.complex128)
+        for pauli in self.values():
+            u = np.kron(u, protocols.unitary(pauli))
```
Strilanc: If we're really being safe I would also add `_apply_unitary_`, to ensure a 20-entry pauli string does not cause a memory error.

cirq/ops/pauli_string_test.py
```diff
@@ -616,3 +616,27 @@ def test_with_qubits():
     for q in new_qubits:
         assert new_pauli_string[q] == cirq.Pauli.by_index(q.x)
     assert new_pauli_string.coefficient == -1
+
+
+@pytest.mark.parametrize('paulis, coefficient, expected_unitary', (
```
Strilanc: Do we call the "is operation consistent" test on a Pauli string instance? If not, add it.

dabacon:
@viathor may be out for a bit.  @Strilanc maybe you could take over and get someone else to do a final review?

Strilanc:
Obsoleted by https://github.com/quantumlib/Cirq/pull/1743

PR #1665, opened 2019-05-23
Adjust failing Windows tests to make them pass

c-poole:
We had another instance of a dtype that doesn't exist in windows numpy showing up in some tests. Modified to skip in Windows.

### DIFF
--- a/cirq/protocols/approximate_equality_test.py
+++ b/cirq/protocols/approximate_equality_test.py
@@ -12,16 +12,79 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from fractions import Fraction
+from decimal import Decimal
+from numbers import Number
+import numpy as np
 import cirq
 
 
 def test_approx_eq_primitives():
+    assert not cirq.approx_eq(1, 2, atol=1e-01)
     assert cirq.approx_eq(1.0, 1.0 + 1e-10, atol=1e-09)
     assert not cirq.approx_eq(1.0, 1.0 + 1e-10, atol=1e-11)
     assert cirq.approx_eq(0.0, 1e-10, atol=1e-09)
     assert not cirq.approx_eq(0.0, 1e-10, atol=1e-11)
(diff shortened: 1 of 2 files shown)


vtomole:
What i did was delete the non-supporting types instead. See: https://github.com/quantumlib/Cirq/pull/1540/files#diff-15443b8109d9fb57ef919d410ae5f5d0R53

cirq/protocols/approximate_equality_test.py
```diff
@@ -38,6 +40,9 @@ def test_approx_eq_mixed_primitives():
     assert not cirq.approx_eq(1, 1.0 + 1e-10, atol=1e-11)
 
 
+@pytest.mark.skipif(sys.platform.startswith('win'),
```
vtomole: Let's skip if `float128` or `complex256` is not supported like https://github.com/quantumlib/Cirq/blob/0452fd817fd6a52e23682712484a8a5c047d30d0/cirq/google/sim/xmon_simulator_test.py#L375

cirq/protocols/approximate_equality_test.py
```diff
@@ -73,6 +78,9 @@ def test_decimal_compatibility():
     assert not cirq.approx_eq(Decimal('NaN'), Decimal('-Infinity'), atol=1e-9)
 
 
+@pytest.mark.skipif(sys.platform.startswith('win'),
```
vtomole: skipif for types like my previous comment.

cirq/protocols/approximate_equality_test.py
c-poole: Adjusted skip conditions in both adjusted tests.

c-poole:
Whenever a solution for this gets merged, all tests outside of contrib pass on my Windows machine. There are two circuit diagram comparison tests in contrib that are failing for reasons I suspect are similar to the ion_device tests.

c-poole:
I modified this change so that instead of skipping the tests if there is a bad data type inside them, the test only utilizes the bad data types if the machine recognizes them. This seems like the best route to take for this. Ready for re-review @vtomole.

PR #1666, opened 2019-05-23
Add serialization and deserialization of xmon gate set for v2 protos

dabacon:
Includes some changes to the new protos:
* Args are now named.
* Instead of bytes for an arg type I moved this to repeated bool.  This is for invert mask, and the compression for bytes is likely not worth it.

### DIFF
--- a/cirq/api/google/v2/program.proto
+++ b/cirq/api/google/v2/program.proto
@@ -93,14 +93,12 @@ message Operation {
   // Which gate this operation corresponds to.
   Gate gate = 1;
 
-  // Arguments needed to fully specify the gate.
-  repeated Arg args = 2;
+  // Map from the argument name to the Argument needed to fully specify
+  // the gate.
+  map<string, Arg> args = 2;
 
   // Which qubits the operation acts on.
   repeated Qubit qubits = 3;
-
-  // If the operation produces results, these are referenced by these keys.
-  repeated string keys = 4;
 }
 
 // The instruction identifying the action taken on the quantum computer.
@@ -118,7 +116,7 @@ message Qubit {
   //
   // Typically ids for qubits on a line are simple string versions of integers,
   // while for qubits on a square grid these are integers separated by a
-  // underscore, i.e. '0_1', '1_2', etc.
+  // comma, i.e. '0, 1', '1, 2', etc.
   string id = 2;
 }
 
@@ -146,8 +144,14 @@ message Arg {
 message ArgValue {
   oneof arg_value {
     float float_value = 1;
-    bytes bytes_value = 2;
+    RepeatedBoolean bool_values = 2;
+    string string_value = 3;
   }
+}
+
+// A repeated boolean value.
+message RepeatedBoolean {
+  repeated bool values = 1;
 }
 
 // A function of arguments. This is an s-expression tree representing
--- a/cirq/api/google/v2/program.proto
+++ b/cirq/api/google/v2/program.proto
@@ -116,7 +116,7 @@ message Qubit {
   //
   // Typically ids for qubits on a line are simple string versions of integers,
   // while for qubits on a square grid these are integers separated by a
-  // comma, i.e. '0, 1', '1, 2', etc.
+  // underscore, i.e. '0_1', '1_2', etc.
   string id = 2;
 }
 
--- a/cirq/devices/grid_qubit.py
+++ b/cirq/devices/grid_qubit.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 
-from typing import Dict
+from typing import Dict, List
 
 from cirq import ops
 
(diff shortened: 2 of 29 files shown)


dabacon:
Fixing some python 3.5 issues and coverage.

cirq/google/op_serializer.py
```diff
@@ -0,0 +1,174 @@
...
+            serialized_gate_id: The string id of the gate when serialized.
+            args: A list of specification of the arguments to the gate when
+                serializing, including how to get this information from the
+                gate of the given gate type.
+        """
+        self.gate_type = gate_type
+        self.serialized_gate_id = serialized_gate_id
+        self.args = args
+
+    def to_proto(self, op: ops.GateOperation) -> Dict:
```
maffoo: Could this instead produce actual proto message instances? There could be a parallel `to_proto_dict` method (or `to_json`) that produces a json-compatible dict (I presume the conversion could be automatic since there is a canonical json encoding for proto v3).

cirq/devices/grid_qubit.py
```diff
@@ -44,16 +44,24 @@ def __repr__(self):
...
+    def to_proto_dict(self, v2_proto=False) -> Dict:
         """Return the proto in dictionary form."""
+        # TODO: Deprecate v1 proto method.
         return {
             'row': self.row,
             'col': self.col,
         }
 
+    def proto_id(self):
+        return '{}_{}'.format(self.row, self.col)
```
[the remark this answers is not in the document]
dabacon: Ah yeah I changed it then changed it back.  

I was tempted to start relying on str on the qubit instead.

cirq/google/op_serializer.py
dabacon: We could but I'm not sure of the utility.  The python proto library is kind of a mess and we will just end up passing this as a protodict to engine code (I don't think it is ever in proto message form in Cirq right now).

maffoo: One advantage of proto classes is that we could use something like ['mypy-protobuf'](https://github.com/dropbox/mypy-protobuf) to generate `.pyi` stubs and then mypy can typecheck the code.

dabacon: That would be nice.  I'm going to leave as proto dicts for now.  I've relabeled the methods to say proto_dict in the serializer and deserializer.

dabacon:
OK, fixed coverage, now ready for review.

cirq/google/op_serializer.py
```diff
@@ -0,0 +1,164 @@
...
+                'Argument {} had type {} but gate returned type {}'.format(
+                    arg.serialized_name, arg.serialized_type, type(value)))
+
+    def _arg_value_to_proto(self, value: arg_func_langs.ArgValue) -> Dict:
+        arg_value = lambda x: {'arg_value': x}
+        if isinstance(value, (float, int)):
+            return arg_value({'float_value': float(value)})
+        if isinstance(value, str):
+            return arg_value({'string_value': str(value)})
+        if (isinstance(value, (list, tuple)) and
```
maffoo: Allow numpy ndarray here as well?

cirq/google/serializable_gate_set.py
```diff
@@ -0,0 +1,179 @@
...
+"""Support for serializing and deserializing cirq.api.google.v2 protos."""
+
+from typing import cast, Dict, Iterable, Optional, Tuple, Union
+
+from cirq.google import op_deserializer, op_serializer
+
+from cirq import circuits, devices, ops, schedules, value
+
+
+class SerializableGateSet():
```
maffoo: nit: don't need parens

cirq/google/gate_sets.py
```diff
@@ -13,5 +13,115 @@
...
+                    gate_getter=protocols.measurement_key),
+                op_serializer.SerializingArg(serialized_name='invert_mask',
+                                             serialized_type=List[bool],
+                                             gate_getter='invert_mask')
+            ])
+    ],
+    deserializers=[
+        op_deserializer.GateOpDeserializer(
+            serialized_gate_id='exp_w',
+            gate_constructor=ops.PhasedXPowGate,
```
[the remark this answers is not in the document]
dabacon: I does because the constructor for PhasedXPowGate returns gates of those types as appropriate.  But XPowGate is not a subclass of PhasedXPowGate, so to get the serialization to work you need the serializer cases.

cirq/api/google/v2/program.proto
```diff
@@ -93,14 +93,12 @@ message Operation {
...
-  repeated Arg args = 2;
+  // Map from the argument name to the Argument needed to fully specify
+  // the gate.
+  map<string, Arg> args = 2;
 
   // Which qubits the operation acts on.
   repeated Qubit qubits = 3;
-
-  // If the operation produces results, these are referenced by these keys.
-  repeated string keys = 4;
```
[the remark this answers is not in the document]
dabacon: So I do think putting it as an arg makes things much cleaner, and it fits into how this is used while deserializing quite nicely.  

But what you are describing would be nice., to be able to determine the amount of information this gate will provide.  My first thought would be to record this separately.  One worry I have is that will for computational basis state measurements this is a simple "this gate produces this much data", if we had an "expectation value" type gate this will then need to know repetitions / parameter counts (which is fine, I guess).  But more worrisome is that we might have an argument to an expectation value gate that changes the amount of info we return back.

cirq/api/google/v2/program.proto
```diff
@@ -146,10 +144,16 @@ message Arg {
 message ArgValue {
   oneof arg_value {
     float float_value = 1;
-    bytes bytes_value = 2;
+    RepeatedBoolean bool_values = 2;
```
dabacon: The more I stare at this, the more I think this should be 
RepeatedArgValue, which includes repeated versions of the arg values (not ArgValues, but things like repeated float_values).    I'd be tempted to use this in the Arg oneof.

cirq/google/op_deserializer.py
```diff
@@ -0,0 +1,129 @@
...
+from cirq import devices, ops
+from cirq.google import arg_func_langs
+
+
+class DeserializingArg(
+        NamedTuple(
+            'DeserializingArg',
+            [('serialized_name', str), ('constructor_arg_name', str),
+             ('value_func', Optional[Callable[[arg_func_langs.ArgValue], Any]]),
+             ('required', bool)])):
```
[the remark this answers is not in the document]
dabacon: done

cirq/google/op_deserializer.py
```diff
@@ -0,0 +1,129 @@
...
+                serialized_name,
+                constructor_arg_name,
+                value_func=None,
+                required=True):
+        return super(DeserializingArg,
+                     cls).__new__(cls, serialized_name, constructor_arg_name,
+                                  value_func, required)
+
+
+class GateOpDeserializer():
```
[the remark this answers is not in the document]
dabacon: done

cirq/google/op_deserializer.py
```diff
@@ -0,0 +1,129 @@
...
+    """Describes how to deserialize a proto to a given Gate type.
+
+    Attributes:
+        serialized_gate_id: The id used when serializing the gate.
+    """
+
+    def __init__(self,
+                 serialized_gate_id: str,
+                 gate_constructor: type,
+                 args: Iterable[DeserializingArg],
```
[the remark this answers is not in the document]
dabacon: moved to Sequence.

cirq/google/op_deserializer.py
```diff
@@ -0,0 +1,129 @@
...
+        gate = self.gate_constructor(**args)
+        return gate.on(*qubits)
+
+    def _args_from_proto(self, args_proto_dict: Dict
+                        ) -> Dict[str, arg_func_langs.ArgValue]:
+        return_args = {}
+        for arg in self.args:
+            if arg.serialized_name not in args_proto_dict and arg.required:
+                raise ValueError(
+                    'Argument {} not in serialized input for but is required.'.
```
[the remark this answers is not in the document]
dabacon: fixed

cirq/google/serializable_gate_set.py
dabacon: done

cirq/api/google/v2/program.proto
dabacon: Ah that doesn't really help, since you still want to have the base recorded data be a repeat of a raw type.

cirq/google/serializable_gate_set.py
```diff
@@ -0,0 +1,179 @@
...
+"""Support for serializing and deserializing cirq.api.google.v2 protos."""
+
+from typing import cast, Dict, Iterable, Optional, Tuple, Union
+
+from cirq.google import op_deserializer, op_serializer
+
+from cirq import circuits, devices, ops, schedules, value
+
+
+class SerializableGateSet:
```
Strilanc: Do we really need all this machinery around the concept of serializing a gate set, instead of just serializing the gate set? I'm worried we're pre-emptively creating abstractions that won't actually be useful in more than one case, and that this abstraction won't be more concise than the direct just-do-it imperative code.

cirq/devices/grid_qubit.py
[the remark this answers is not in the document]
dabacon: I don't think we should do it this way.  Api stuff should be consumers of cirq objects, not modifiers.

cirq/google/op_serializer.py
```diff
@@ -0,0 +1,164 @@
...
+# Type for variables that are subclasses of ops.Gate.
+Gate = TypeVar('Gate', bound=ops.Gate)
+
+
+class SerializingArg(
+        NamedTuple('SerializingArg',
+                   [('serialized_name', str),
+                    ('serialized_type', Type[arg_func_langs.ArgValue]),
+                    ('gate_getter',
+                     Union[str, Callable[[ops.Gate], arg_func_langs.ArgValue]]),
```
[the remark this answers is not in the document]
dabacon: We can easily update this to deal with ops when we need to.   However I suspect that the hardware side will not be fundamentally tied to "operations that depend on qubits" any time soon, because that's not how you build a modular system.

cirq/google/op_serializer.py
dabacon: done

cirq/google/serializable_gate_set.py
dabacon: Yes we do need it.

PR #1668, opened 2019-05-23
Add GridQubit utility methods square, rect, from_pic

### LINKED ISSUE #1634: Class methods on GridQubit to create common lattices
Some obvious shapes
```GridQubit.square(3, 3)```
```GirdQubit.rect(2, 4)```
returning list of list with row and column.   Default to 0,0 as one corner, but option to set corner.

Ability to construction from an ASCII diagram like xmon devices
```GridQubit.from_pic```
which operations on maps like
```
...A...
..AAA..
.AAAAA.
AAAAAAAA
````

### PR DESCRIPTION
**Pull request by c-poole**
Closes #1634  with @dabacon  approval.

### DIFF
--- a/cirq/devices/grid_qubit.py
+++ b/cirq/devices/grid_qubit.py
@@ -53,8 +53,8 @@ class GridQubit(ops.Qid):
         return GridQubit.rect(diameter, diameter, top=top, left=left)
 
     @staticmethod
-    def rect(rows: int, cols: int,
-             top: int = 0, left: int = 0) -> List['GridQubit']:
+    def rect(rows: int, cols: int, top: int = 0,
+             left: int = 0) -> List['GridQubit']:
         """Returns a rectangle of GridQubits.
 
         Args:
@@ -68,8 +68,8 @@ class GridQubit(ops.Qid):
         """
         return [
             GridQubit(row, col)
-            for row in range(top, top+rows)
-            for col in range(left, left+cols)
+            for row in range(top, top + rows)
+            for col in range(left, left + cols)
         ]
 
     @staticmethod
--- a/cirq/devices/grid_qubit.py
+++ b/cirq/devices/grid_qubit.py
@@ -39,7 +39,8 @@ class GridQubit(ops.Qid):
                 abs(self.row - other.row) + abs(self.col - other.col) == 1)
 
     @staticmethod
-    def square(diameter: int, top: int = 0, left: int = 0) -> List['GridQubit']:
+    def square(diameter: int, top: int = 0,
+               left: int = 0) -> List[List['GridQubit']]:
         """Returns a square of GridQubits.
 
         Args:
@@ -48,13 +49,13 @@ class GridQubit(ops.Qid):
             left: Column number of the leftmost row
 
         Returns:
-            A list of GridQubits filling in a square grid
+            A matrix of GridQubits filling in a square grid
         """
         return GridQubit.rect(diameter, diameter, top=top, left=left)
 
     @staticmethod
     def rect(rows: int, cols: int, top: int = 0,
-             left: int = 0) -> List['GridQubit']:
+             left: int = 0) -> List[List['GridQubit']]:
         """Returns a rectangle of GridQubits.
 
         Args:
@@ -64,13 +65,10 @@ class GridQubit(ops.Qid):
             left: Column number of the leftmost row
 
         Returns:
-            A list of GridQubits filling in a rectangular grid
+            A matrix of GridQubits filling in a rectangular grid
         """
-        return [
-            GridQubit(row, col)
-            for row in range(top, top + rows)
-            for col in range(left, left + cols)
-        ]
+        return [[GridQubit(row, col) for row in range(top, top + rows)]
+                for col in range(left, left + cols)]
 
     @staticmethod
     def from_diagram(diagram: str) -> List['GridQubit']:
--- a/cirq/devices/grid_qubit.py
+++ b/cirq/devices/grid_qubit.py
@@ -67,7 +67,8 @@ class GridQubit(ops.Qid):
         Returns:
             A matrix of GridQubits filling in a rectangular grid
         """
-        return [[GridQubit(row, col) for row in range(top, top + rows)]
+        return [[GridQubit(row, col)
+                 for row in range(top, top + rows)]
                 for col in range(left, left + cols)]
 
     @staticmethod
--- a/cirq/devices/grid_qubit.py
+++ b/cirq/devices/grid_qubit.py
@@ -39,8 +39,7 @@ class GridQubit(ops.Qid):
                 abs(self.row - other.row) + abs(self.col - other.col) == 1)
 
     @staticmethod
-    def square(diameter: int, top: int = 0,
-               left: int = 0) -> List[List['GridQubit']]:
+    def square(diameter: int, top: int = 0, left: int = 0) -> List['GridQubit']:
         """Returns a square of GridQubits.
 
         Args:
@@ -49,13 +48,13 @@ class GridQubit(ops.Qid):
             left: Column number of the leftmost row
 
         Returns:
-            A matrix of GridQubits filling in a square grid
+            A list of GridQubits filling in a square grid
         """
         return GridQubit.rect(diameter, diameter, top=top, left=left)
 
     @staticmethod
     def rect(rows: int, cols: int, top: int = 0,
-             left: int = 0) -> List[List['GridQubit']]:
+             left: int = 0) -> List['GridQubit']:
         """Returns a rectangle of GridQubits.
 
         Args:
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: cirq/devices/grid_qubit.py [hunk 1]
```diff
@@ -38,6 +38,61 @@ def is_adjacent(self, other: ops.Qid) -> bool:
         return (isinstance(other, GridQubit) and
                 abs(self.row - other.row) + abs(self.col - other.col) == 1)
 
+    @staticmethod
+    def square(size: int, top: int = 0, left: int = 0) -> List['GridQubit']:
```
Strilanc: Rename size to `diameter`.

cirq/devices/grid_qubit.py
```diff
@@ -38,6 +38,61 @@ def is_adjacent(self, other: ops.Qid) -> bool:
...
+
+        Args:
+            size: Length of a side of the square
+            top: Row number of the topmost row
+            left: Column number of the leftmost row
+
+        Returns:
+            A list of GridQubits filling in a square grid
+        """
+        return [GridQubit(row, col)
```
Strilanc: `return GridQubit.rect(diameter, diameter, top, left)`.

cirq/devices/grid_qubit.py
```diff
@@ -38,6 +38,61 @@ def is_adjacent(self, other: ops.Qid) -> bool:
...
+                for row in range(top, top+size)
+                for col in range(left, left+size)]
+
+    @staticmethod
+    def rect(rows: int, cols: int,
+             top: int = 0, left: int = 0) -> List['GridQubit']:
+        """Returns a rectangle of GridQubits
```
Strilanc: Sentences need periods.

cirq/devices/grid_qubit.py
```diff
@@ -38,6 +38,61 @@ def is_adjacent(self, other: ops.Qid) -> bool:
...
+
+        Returns:
+            A list of GridQubits filling in a rectangular grid
+        """
+        return [GridQubit(row, col)
+                for row in range(top, top+rows)
+                for col in range(left, left+cols)]
+
+    @staticmethod
+    def from_pic(s: str) -> List['GridQubit']:
```
Strilanc: Give a better argument name, like `diagram`.

Strilanc: Rename to `from_diagram` or similar. Picture implies bitmap implies give me a filepath.

cirq/devices/grid_qubit.py
```diff
@@ -38,6 +38,61 @@ def is_adjacent(self, other: ops.Qid) -> bool:
...
+        """Parse ASCIIart device layout into info about qubits and connectivity.
```
Strilanc: space after ascii

cirq/devices/grid_qubit.py
```diff
@@ -38,6 +38,61 @@ def is_adjacent(self, other: ops.Qid) -> bool:
...
+
+        Args:
+            s: String representing the qubit layout. Each line represents a row,
```
Strilanc: Give two examples of diagrams. Specify that the top row is row 0, left column is column 0.

cirq/devices/grid_qubit.py
```diff
@@ -38,6 +38,61 @@ def is_adjacent(self, other: ops.Qid) -> bool:
...
+                character is a hyphen '-'.
+
+        Returns:
+            A list of GridQubits corresponding to the provided picture
+        """
+        lines = s.strip().split('\n')
+        qubits = []
+        for row, line in enumerate(lines):
+            for col, c in enumerate(line.strip()):
+                if c != '-':
```
Strilanc: Don't just test for not hyphen, also define an acceptable character set for "yes qubit here".

Shouldn't the not-a-qubit character be space?

cirq/devices/grid_qubit_test.py
```diff
@@ -32,6 +32,36 @@ def test_xmon_qubit_eq():
...
 
+def test_square():
+    assert cirq.GridQubit.square(2, top=1) == [cirq.GridQubit(1, 0),
+                                               cirq.GridQubit(1, 1),
+                                               cirq.GridQubit(2, 0),
+                                               cirq.GridQubit(2, 1)]
+
+
+def test_rec():
+    assert cirq.GridQubit.rect(1, 2, top=1, left=2) == [cirq.GridQubit(1, 2),
```
Strilanc: Avoid repeating argument values in tests, or use multiple calls. Also check default behavior.

cirq/devices/grid_qubit_test.py
```diff
@@ -32,6 +32,36 @@ def test_xmon_qubit_eq():
     eq.make_equality_group(lambda: cirq.GridQubit(50, 25))
 
 
+def test_square():
+    assert cirq.GridQubit.square(2, top=1) == [cirq.GridQubit(1, 0),
```
Strilanc: Doesn't confirm left works. Doesn't confirm top has default.

cirq/devices/grid_qubit.py
c-poole: I was following the same convention used in for how to translate the diagrams. It seemed intuitive enough to just translate the parts of this private method that do the desired task over to GridQubit. Thoughts?
https://github.com/quantumlib/Cirq/blob/0452fd817fd6a52e23682712484a8a5c047d30d0/cirq/google/known_devices.py#L22-L31

cirq/devices/grid_qubit_test.py
c-poole: Now tests default case for both and specified case for top, left.

cirq/devices/grid_qubit_test.py
c-poole: Adjusted argument values, added test for default behavior.

cirq/devices/grid_qubit.py
c-poole: Adjusted name

c-poole: Adjusted argument name

cirq/devices/grid_qubit.py
c-poole: adjusted argument name

cirq/devices/grid_qubit.py
c-poole: I added two example diagrams, one where all qubits have the same character and one where not at qubits do. I only described the shape of grid in the picture rather than prescribing the exact output list of GridQubits.

cirq/devices/grid_qubit.py
Strilanc: I have sort of a "whatever you do in the privacy of your own module is your business" attitude to these things. When the method was private I didn't worry about which characters were used. But if we're exposing it then that question is suddenly very relevant.

I would say we should allow the use of dot, dash, and space as no-qubit characters and limit the qubit characters to be alphanumeric.

c-poole: Ok, I'll implement that and update the documentation.

cirq/devices/grid_qubit.py
```diff
@@ -38,6 +38,93 @@ def is_adjacent(self, other: ops.Qid) -> bool:
...
+        """Returns a rectangle of GridQubits.
+
+        Args:
+            rows: Number of rows in the rectangle
+            cols: Number of columns in the rectangle
+            top: Row number of the topmost row
+            left: Column number of the leftmost row
+
+        Returns:
+            A list of GridQubits filling in a rectangular grid
```
dabacon: I wonder if it would be more useful if this were a matrix and not a list.  That way I could do q[0][1] to get row 0 and column 1?

c-poole: That sounds reasonable. I'll change it to a list of lists.

Strilanc: Most methods that we have take a list instead of a list of lists of qubits. If you want random access into the grid, it seems like you would just use GridQubit directly anyway?

I vote leave it as a list, because then the signature matches from_diagram.

c-poole: We now have test passing commits either way. Whenever it is decided, we can just pick the corresponding commit.

dabacon: qubits = cirq.GridQubit.square(5)
then accessing qubits[i][j] is slightly nicer than cirq.GriqQubit(i,j).
but I think actually there is a bunch of stuff that could naturally go in an object that you produce via GridQubit... i.e. a grid.  Now that I've written a could versions of variational algorithms I find myself doing a lot of 
```
for x in range(d):
  for y in range(d-1):
      nearest neighbor
```
This gets even worse with the odder sized lattice.

So for now probably list is good, but I think we may need an actual Grid object helper.  I'll open an issue for that.

dabacon: Let's go with @Strilanc 's suggestion.  Sorry for the detour.

c-poole: I've reverted the commits to the List[GridQubit] return type version.

PR #1669, opened 2019-05-23
Implement transformations between Pauli**exponent and Rp(theta)

### LINKED ISSUE #1635: Transformation that turns Pauli**exponent gates to Rp(theta) gates and vice versa
These are the same up to global phase, but different people like different version of this gate.

c-poole:
Add `as_rotation` to turn a Pauli^exponent into a Rp(theta)
Add `as_pow_gate` to turn a Rp(theta) into a Pauli^exponent.
Closes #1635


cirq/ops/common_gates.py
```diff
@@ -81,6 +81,12 @@ def _apply_unitary_(self, args: protocols.ApplyUnitaryArgs
             args.available_buffer *= p
         return args.available_buffer
 
+    def as_rotation(self):
```
Strilanc: Need better names. The expression `(X**0.5).as_rotation()` is a bit of a mystery, I think. It's already a rotation..

I think what you want is `in_su2` or `with_non_imaginary_global_phase` or something like that. Describe at a lower level what it means for it to be "a rotation" and put that in the name.

cirq/ops/common_gates.py
```diff
@@ -81,6 +81,12 @@ def _apply_unitary_(self, args: protocols.ApplyUnitaryArgs
...
+        return XPowGate(exponent=self._exponent, global_shift=-0.5)
+
+    def as_pow_gate(self):
```
Strilanc: Need better name. `XPowGate(...).as_pow_gate()` just looks wrong.

Maybe `with_global_phase_matching_smallest_period_cycle` or `as_pure_power_without_global_shift` or `with_no_phase_on_plus_one_eigenspace` or ... okay none of these are very good. But you get the idea.

cirq/ops/common_gates.py
c-poole: I like in_su2 since that describes the exact purpose why someone would want to use this transformation. I'll go with that.

cirq/ops/common_gates.py
c-poole: I didn't really envision the use as someone calling it in the way you described since it would only changed the gate if they specified a non-default `global_shift` in the constructor, in which case why would they specify that only to undo it. I agree that the name could use work nonetheless. I'm comfortable with relying on the fact that the user understands what `global_shift` parameter means since the only reason they would want to use these utility methods is to adjust it to their preferred gate specification. I can add docstrings to the methods to make their use more explicit as well.
How do the following sound?
`with_default_global_shift`
`with_default_global_phase`
`with_zero_global_shift`

Strilanc: Because global shift is really not a concept I expect users to be familiar with, I want to downplay it. I really see the reason I would call this method as a way of canonicalizing the gate, where I decided to do so by moving all phase on to the - eigenbasis with no phase on the + eigenbasis (of the Pauli).

Perhaps what we need is a succinct way of referring to the space of single-qubit operations where one of the eigenvalues is equal to 1. Then we could just say `in_thatspace`. Maybe `with_determinant_equal_to_relative_phase`?

We could also go with something vague but user-relevant like `with_canonical_global_phase()`.

c-poole: I like `with_canonical_global_phase` because it implies a library default which is what this method provides.  I imagined these methods as translating between the "Cirq default" that you get if you didn't utilize `global_shift` at all and the "SU(2)" rotations. I expect the `in_su2` method to get much greater use by comparison. My expected use case for `with_canonical_global_phase` would be as a handy sanity check method for converting gates that you made with some `global_shift`, whether by directly specifying it in the constructor or by calling `in_su2` back into the "Cirq default". I'll make an update using those names.

cirq/ops/common_gates_test.py
```diff
@@ -67,6 +67,28 @@ def test_cz_init():
...
+    expectedPowx = cirq.X**(0.4 / pi)
+    recievedPowx = initialRx.as_pow_gate()
+    backToRx = recievedPowx.as_rotation()
+    assert recievedPowx == expectedPowx
+    assert backToRx == initialRx
+    initialRy = cirq.Ry(0.123)
+    expectedPowy = cirq.Y**(0.123 / pi)
+    recievedPowy = initialRy.as_pow_gate()
+    backToRy = recievedPowy.as_rotation()
+    assert recievedPowy == expectedPowy
```
Strilanc: `received`

cirq/ops/common_gates.py
```diff
@@ -81,6 +81,12 @@ def _apply_unitary_(self, args: protocols.ApplyUnitaryArgs
             args.available_buffer *= p
         return args.available_buffer
 
+    def in_su2(self):
```
Strilanc: Need a return type annotation and a one-liner docstring.

cirq/ops/common_gates.py
```diff
@@ -81,6 +81,12 @@ def _apply_unitary_(self, args: protocols.ApplyUnitaryArgs
...
+        return XPowGate(exponent=self._exponent, global_shift=-0.5)
+
+    def with_canonical_global_phase(self):
```
Strilanc: Need a return type annotation and a one-liner docstring.

cirq/ops/common_gates.py
```diff
@@ -184,6 +190,12 @@ class YPowGate(eigen_gate.EigenGate,
     `cirq.Y`, the Pauli Y gate, is an instance of this gate at exponent=1.
     """
 
+    def in_su2(self):
```
Strilanc: Need a return type annotation and a one-liner docstring.

cirq/ops/common_gates.py
```diff
@@ -184,6 +190,12 @@ class YPowGate(eigen_gate.EigenGate,
...
+        return YPowGate(exponent=self._exponent, global_shift=-0.5)
+
+    def with_canonical_global_phase(self):
```
Strilanc: Need a return type annotation and a one-liner docstring.

cirq/ops/common_gates.py
```diff
@@ -298,6 +310,12 @@ def _apply_unitary_(self, args: protocols.ApplyUnitaryArgs
             args.target_tensor *= p
         return args.target_tensor
 
+    def in_su2(self):
```
Strilanc: Need a return type annotation and a one-liner docstring.

cirq/ops/common_gates.py
```diff
@@ -298,6 +310,12 @@ def _apply_unitary_(self, args: protocols.ApplyUnitaryArgs
...
+        return ZPowGate(exponent=self._exponent, global_shift=-0.5)
+
+    def with_canonical_global_phase(self):
```
Strilanc: Need a return type annotation and a one-liner docstring.

PR #1674, opened 2019-05-25
on_each of SingleQubitGate takes list of Qid of arbitrary depth as input

### LINKED ISSUE #1624: Improve error message if on_each gets a list
When you do `cirq.H.on_each([q0, q1])` instead of the unpacked version  `cirq.H.on_each(q0, q1)` for some qubits `q0` and `q1`, the error message you get is **Gate was called with type different than Qid**. 

Maybe either flatten (because most of the time you'll have your qubits in a list or a list of lists), or give a more descriptive error message.

timothyman:
Fixes #1624

### DIFF
--- a/cirq/ops/gate_features.py
+++ b/cirq/ops/gate_features.py
@@ -50,7 +50,7 @@ class SingleQubitGate(raw_types.Gate, metaclass=abc.ABCMeta):
         Raises:
             ValueError if targets are not instances of Qid or List[Qid].
         """
-        operations = [] # type: List[raw_types.Operation]
+        operations = []  # type: List[raw_types.Operation]
         for target in targets:
             if isinstance(target,
                           collections.Iterable) and not isinstance(target, str):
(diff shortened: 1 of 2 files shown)


timothyman:
I changed the title of this PR, but unfortunately don't have the credentials for changing the original issue.

cirq/ops/gate_features.py
```diff
@@ -45,9 +45,15 @@ def on_each(self, *targets: raw_types.Qid) -> op_tree.OP_TREE:
             Operations applying this gate to the target qubits.
 
         Raises:
-            ValueError if targets are not instances of Qid.
+            ValueError if targets are not instances of Qid or List[Qid].
         """
-        return [self.on(target) for target in targets]
+        operations = []
+        for target in targets:
+            if isinstance(target, list):
```
Strilanc: We shouldn't be specializing to list. If we're going to flatten, we should flatten everything.

If it's a qubit id, apply the operation.
It it's a collections.Iterable, recurse.
If it's neither, or if it's a known problematic type like string (which is an infinitely descending iterable), raise an error.

Add a test making sure the string case terminates. Add a test that provides qubits using a generator.

timothyman: I followed your suggestions in the newest push.

PR #1677, opened 2019-05-28
Implement ThreeQubitDiagonalGate

### LINKED ISSUE #980: Implement ThreeQubitDiagonalGate
A three-qubit diagonal gate is a gate whose unitary matrix is of the form `np.diag([a, b, c, d, e, f, g, h])`. Such gates can be decomposed and simulated efficiently, and this decomposition/simulation code would be useful in several places (e.g. when defining the `CXXYY` gate in openfermion-cirq).

The `ThreeQubitDiagonalGate` class should have the following features:

- An `__init__` method taking the rotation angle for each element along diagonal. (The *angles*, not the complex values the angles produces. Also store the angles, not the complex values. This makes the behavior of `__pow__` better.)
- A `def _unitary_(self)` method that returns its matrix.
- A `def _circuit_diagram_info_(self, args)` method that uses symbols like `diag(a, b, ...)----#2----#3`.
- A `def default_decompose(self)` method that looks almost identical to the CCZ decomposition, except that the various T/phasing gates are replaced with gates that depend on the diagonal components of the matrix.
    - For reference, this is the CCZ code:

        ```python
        def default_decompose(self, qubits):
            """An adjacency-respecting decomposition.
    
            0: ───p───@──────────────@───────@──────────@──────────
                      │              │       │          │
            1: ───p───X───@───p^-1───X───@───X──────@───X──────@───
                          │              │          │          │
            2: ───p───────X───p──────────X───p^-1───X───p^-1───X───
    
            where p = T**self._exponent
            """
            a, b, c = qubits
    
            # Hacky magic: avoid the non-adjacent edge.
            if hasattr(b, 'is_adjacent'):
                if not b.is_adjacent(a):
                    b, c = c, b
                elif not b.is_adjacent(c):
                    a, b = b, a
    
            p = common_gates.T**self._exponent
            sweep_abc = [common_gates.CNOT(a, b),
                         common_gates.CNOT(b, c)]
    
            yield p(a), p(b), p(c)
            yield sweep_abc
            yield p(b)**-1, p(c)
            yield sweep_abc
            yield p(c)**-1
            yield sweep_abc
            yield p(c)**-1
            yield sweep_abc
            ```

    Each of the `p`s in the code controls the phasing of four of the eight diagonal entries in the matrix. By varying all 7 you can implement any diagonal matrix (up to global phase). You will have to write code to solve the actual_diagonal-to-phasing_angles problem.

- A `def _apply_unitary_to_tensor_` method that multiplies the appropriate entries by the appropriate phase values from the matrix.
     - For reference, here is how the CCZ would be implemented:
        ```python
        def _apply_unitary_to_tensor_(self,
                                      target_tensor: np.ndarray,
                                      available_buffer: np.ndarray,
                                      axes: Sequence[int],
                                      ) -> np.ndarray:
            ooo = linalg.slice_for_qubits_equal_to(axes, 0b111)
            target_tensor[ooo] *= np.exp(1j * self.exponent * np.pi)
            return target_tensor
        ```
    The code should avoid doing multiplications where the phase values equal 1.

- A `__pow__` method to extrapolate the phasing effect.
- `__eq__`/`__ne__`/`__hash__`

bradyb:
Fixes #980


cirq/ops/three_qubit_gates.py
```diff
@@ -145,6 +144,125 @@ def __str__(self) -> str:
         return 'CCZ**{}'.format(self._exponent)
 
 
+class ThreeQubitDiagonalGate(gate_features.ThreeQubitGate):
+    """A gate given by a diagonal 8x8 matrix."""
+
+    def __init__(self,
+                 a: float = 0.0,
```
Strilanc: It's a bit silly to take this many arguments instead of just a list of floats. Also, make sure to include `radians` in the name, and document that it's in radians.

cirq/ops/three_qubit_gates.py
```diff
@@ -145,6 +144,125 @@ def __str__(self) -> str:
...
+        self._diag_angles = []  # type: List[float]
+        self._diag_angles.append(a)
+        self._diag_angles.append(b)
+        self._diag_angles.append(c)
+        self._diag_angles.append(d)
+        self._diag_angles.append(e)
+        self._diag_angles.append(f)
+        self._diag_angles.append(g)
+        self._diag_angles.append(h)
+        self._matrix = np.diag(
```
Strilanc: Don't store the matrix, store the diagonal.

cirq/ops/three_qubit_gates.py
```diff
@@ -145,6 +144,125 @@ def __str__(self) -> str:
...
+            [np.exp(1j * angle) for angle in self._diag_angles])
+
+    def _unitary_(self) -> np.ndarray:
```
Strilanc: Implement `_has_unitary_`.

cirq/ops/three_qubit_gates.py
```diff
@@ -145,6 +144,125 @@ def __str__(self) -> str:
...
+            p_gates[5](c),
+            sweep_abc,
+            p_gates[6](c),
+            sweep_abc,
+        ]
+
+    def _apply_unitary_(self, args: protocols.ApplyUnitaryArgs) -> np.ndarray:
+        if np.allclose(self._matrix, np.identity(8)):
+            return args.target_tensor
+        return linalg.targeted_left_multiply(self._matrix.astype(
```
Strilanc: This can be implemented much more efficiently by taking advantage of the fact that the matrix is diagonal. Iterate over the diagonal elements while phasing the corresponding subsection of the tensor. There is a utility method `subspace_index` on the `args` value to make it easier to get at the relevant subsection of the tensor.

cirq/ops/three_qubit_gates.py
```diff
@@ -145,6 +144,125 @@ def __str__(self) -> str:
...
+            args.target_tensor.dtype).reshape((2,) * (2 * len(args.axes))),
+                                             args.target_tensor,
+                                             args.axes,
+                                             out=args.available_buffer)
+
+    def __eq__(self, other):
```
Strilanc: Should also support hashing. Easiest way to do this is to use `@value.value_equality` and implement `_value_equality_values_(self)` that returns a tuple of the angles.

cirq/ops/three_qubit_gates.py
```diff
@@ -145,6 +144,125 @@ def __str__(self) -> str:
...
 
 
+class ThreeQubitDiagonalGate(gate_features.ThreeQubitGate):
+    """A gate given by a diagonal 8x8 matrix."""
+
+    def __init__(self,
+                 a: float = 0.0,
+                 b: float = 0.0,
+                 c: float = 0.0,
+                 d: float = 0.0,
```
Strilanc: We should include support for sympy expressions in addition to floats. This involves making `_has_unitary_` return False, `_unitary_` return NotImplemented etc when one of the angles is symbolic. Also you'd add `_resolve_parameters_` / `_is_parameterized_` methods.

cirq/ops/three_qubit_gates.py
```diff
@@ -145,6 +144,125 @@ def __str__(self) -> str:
...
+                a, b = b, a
+        sweep_abc = [common_gates.CNOT(a, b), common_gates.CNOT(b, c)]
+        phase_matrix = np.array([[0, 0, 1, 0, 1, 1, 1], [0, 1, 0, 1, 1, 0, 1],
+                                 [0, 1, 1, 1, 0, 1, 0], [1, 0, 0, 1, 1, 1, 0],
+                                 [1, 0, 1, 1, 0, 0, 1], [1, 1, 0, 0, 0, 1, 1],
+                                 [1, 1, 1, 0, 1, 0, 0]])
+        shifted_angles_tail = [
+            angle - self._diag_angles[0] for angle in self._diag_angles[1:]
+        ]
+        phase_solutions = np.linalg.solve(phase_matrix, shifted_angles_tail)
```
Strilanc: Won't the solution be essentially the same every time? This should be a matrix multiplication instead of a call to a linear solver.

cirq/ops/three_qubit_gates.py
```diff
@@ -145,6 +144,125 @@ def __str__(self) -> str:
...
+        p_gates = []
+        for solution in phase_solutions:
+            p_gates.append(common_gates.T**(4 * solution / np.pi))
```
Strilanc: Use Z instead of T to drop the factor of four.

cirq/ops/three_qubit_gates.py
```diff
@@ -145,6 +144,125 @@ def __str__(self) -> str:
...
+        sweep_abc = [common_gates.CNOT(a, b), common_gates.CNOT(b, c)]
+        phase_matrix = np.array([[0, 0, 1, 0, 1, 1, 1], [0, 1, 0, 1, 1, 0, 1],
+                                 [0, 1, 1, 1, 0, 1, 0], [1, 0, 0, 1, 1, 1, 0],
+                                 [1, 0, 1, 1, 0, 0, 1], [1, 1, 0, 0, 0, 1, 1],
+                                 [1, 1, 1, 0, 1, 0, 0]])
+        shifted_angles_tail = [
+            angle - self._diag_angles[0] for angle in self._diag_angles[1:]
+        ]
+        phase_solutions = np.linalg.solve(phase_matrix, shifted_angles_tail)
+        p_gates = []
```
Strilanc: Use a list comprehension.

cirq/ops/three_qubit_gates_test.py
```diff
@@ -167,30 +184,42 @@ def test_decomposition_respects_locality(gate):
...
+      |   |       |   |   |       |      |      |
+2: ---X---X^0.5---@---|---|-------swap---swap---#3-----
                       |   |       |
-3: -------------------@---@^0.5---swap----------
-""", use_unicode_characters=False)
+3: -------------------@---@^0.5---swap-----------------
+""",
+                                    use_unicode_characters=False)
+
+
```
Strilanc: Call `cirq.testing.assert_implements_consistent_protocols` on an instance of the gate.

cirq/ops/three_qubit_gates.py
```diff
@@ -145,6 +144,125 @@ def __str__(self) -> str:
...
+
+    def _unitary_(self) -> np.ndarray:
+        return self._matrix
+
+    def _circuit_diagram_info_(self, args: protocols.CircuitDiagramInfoArgs
+                              ) -> protocols.CircuitDiagramInfo:
+        return protocols.CircuitDiagramInfo(('diag', '#2', '#3'))
+
+    def __pow__(self, exponent: Any) -> 'ThreeQubitDiagonalGate':
+        if not isinstance(exponent, (int, float)):
```
Strilanc: This will refuse to work on symbolic exponents. (There's probably a lot of places in the code that do this... it used to be a limitation. But it's time to change it now that the limit is gone.)

Allow `sympy.Basic` as one of the allowed types. Use `protocols.mul` instead of `*`, because `*` doesn't do the right thing for expressions like `1.0 * sympy.Symbol('a')`, unfortunately.

cirq/ops/three_qubit_gates.py
bradyb: OK, so each entry may be a sympy expression or a float?

cirq/ops/three_qubit_gates_test.py
bradyb: Done in test_consistent_protocols.

cirq/ops/three_qubit_gates.py
bradyb: Done

cirq/ops/three_qubit_gates.py
bradyb: Done.

cirq/ops/three_qubit_gates.py
bradyb: Done.

cirq/ops/three_qubit_gates.py
bradyb: Done.

cirq/ops/three_qubit_gates.py
bradyb: Let me know if the subspace indexing is too hacky.

cirq/ops/three_qubit_gates.py
bradyb: Done

cirq/ops/three_qubit_gates.py
bradyb: Inverted the matrix and made it matrix multiplication.

cirq/ops/three_qubit_gates.py
bradyb: Done

cirq/ops/three_qubit_gates.py
bradyb: I removed the check all together (like FSimGate).

cirq/ops/three_qubit_gates.py
```diff
@@ -14,10 +14,12 @@
 
 """Common quantum gates that target three qubits."""
 
-from typing import Optional, Tuple
+from typing import Any, List, Optional, Tuple, Union
 
 import numpy as np
+import sympy
 
+import cirq
```
bradyb: I keep getting a lint warning because of the ParamResolver type annotation. Is it ok to use # pylint: disable=unused-import?

cirq/ops/three_qubit_gates.py
Strilanc: Yes

cirq/ops/three_qubit_gates.py
Strilanc: Use the `if TYPE_CHECKING` thing used elsewhere in the codebase

cirq/ops/three_qubit_gates.py
```diff
@@ -145,6 +147,126 @@ def __str__(self) -> str:
...
+
+    The list of angles passed to the constructor should be in radians.
+    """
+
+    def __init__(self,
+                 diag_angles_radians: List[Union[float, sympy.Basic]]) -> None:
+        self._diag_angles_radians = diag_angles_radians  # type: List[float]
+
+    def _is_parameterized_(self):
+        return any([
```
Strilanc: The square brackets are not needed

cirq/ops/three_qubit_gates.py
```diff
@@ -145,6 +147,126 @@ def __str__(self) -> str:
...
+            for angle in self._diag_angles_radians
+        ])
+
+    def _circuit_diagram_info_(self, args: protocols.CircuitDiagramInfoArgs
+                              ) -> protocols.CircuitDiagramInfo:
+        return protocols.CircuitDiagramInfo(('diag', '#2', '#3'))
+
+    def __pow__(self, exponent: Any) -> 'ThreeQubitDiagonalGate':
+        return ThreeQubitDiagonalGate([
+            protocols.mul(angle, exponent, NotImplemented)
```
Strilanc: This will do the wrong thing when `NotImplemented` is returned. Add a test that confirms `cirq.pow(ThreeQubitDiagonalGate(...), "test", None) is None`.

cirq/ops/three_qubit_gates.py
```diff
@@ -145,6 +147,126 @@ def __str__(self) -> str:
...
+
+        where p_i = T**(4*x_i) and x_i solve the system of equations
+                    [0, 0, 1, 0, 1, 1, 1][x_0]   [l_1]
+                    [0, 1, 0, 1, 1, 0, 1][x_1]   [l_2]
+                    [0, 1, 1, 1, 0, 1, 0][x_2]   [l_3]
+                    [1, 0, 0, 1, 1, 1, 0][x_3] = [l_4]
+                    [1, 0, 1, 1, 0, 0, 1][x_4]   [l_5]
+                    [1, 1, 0, 0, 0, 1, 1][x_5]   [l_6]
+                    [1, 1, 1, 0, 1, 0, 0][x_6]   [l_7]
+        where l_i is self._diag_angles_radians[i].
```
Strilanc: Use `r` instead of `l`, so it looks like like the 1s.

cirq/ops/three_qubit_gates_test.py
```diff
@@ -32,6 +33,10 @@ def test_eigen_gates_consistent_protocols(eigen_gate_type):
 def test_consistent_protocols():
     cirq.testing.assert_implements_consistent_protocols(cirq.CSWAP)
 
+    diag_angles = [np.pi * i / 4 for i in range(0, 8)]
```
Strilanc: Pick angles more arbitrary than this, such as the first 8 primes.

cirq/ops/three_qubit_gates_test.py
```diff
@@ -87,6 +92,13 @@ def test_unitary():
         [0, 0, 0, 0, 0, 0, 0, 1],
     ]), atol=1e-8)
 
+    diagonal_angles = [np.pi * i / 4 for i in range(0, 8)]
```
Strilanc: This is not testing one of the angles, because the first angle equals 0.

cirq/ops/three_qubit_gates_test.py
```diff
@@ -167,30 +180,62 @@ def test_decomposition_respects_locality(gate):
...
-      │   │       │   │   │       │   │
-1: ───@───@───────X───@───@───────┼───×───
-      │   │       │   │   │       │   │
-2: ───X───X^0.5───@───┼───┼───────×───×───
+        cirq.TOFFOLI(a, b, c)**0.5, cirq.CCX(a, c, b), cirq.CCZ(a, d, b),
+        cirq.CCZ(a, d, b)**0.5, cirq.CSWAP(a, c, d), cirq.FREDKIN(a, b, c),
+        cirq.ThreeQubitDiagonalGate(diagonal_angles)(a, b, c))
+    cirq.testing.assert_has_diagram(
+        circuit, """
+0: ───@───@───────@───@───@───────@───@───diag───
```
Strilanc: Test the diagonal gate's diagram separately.

cirq/ops/three_qubit_gates_test.py
```diff
@@ -167,30 +180,62 @@ def test_decomposition_respects_locality(gate):
...
 """)
-    cirq.testing.assert_has_diagram(circuit, """
-0: ---@---@-------@---@---@-------@------@------
-      |   |       |   |   |       |      |
-1: ---@---@-------X---@---@-------|------swap---
-      |   |       |   |   |       |      |
-2: ---X---X^0.5---@---|---|-------swap---swap---
+    cirq.testing.assert_has_diagram(circuit,
+                                    """
+0: ---@---@-------@---@---@-------@------@------diag---
```
Strilanc: The diagonal gate needs to describe the phases it is using. This will make the diagram quite wide or perhaps quite tall, but it is needed. Look at what `TwoQubitMatrixGate` does for reference.

cirq/ops/three_qubit_gates.py
bradyb: Ok, added.

cirq/ops/three_qubit_gates.py
bradyb: Removed.

cirq/ops/three_qubit_gates.py
bradyb: Ah I see, added the test.

cirq/ops/three_qubit_gates.py
bradyb: Done.

cirq/ops/three_qubit_gates_test.py
bradyb: I kept it flat so that the circuit is wide.

cirq/ops/three_qubit_gates_test.py
bradyb: Done.

cirq/ops/three_qubit_gates_test.py
bradyb: OK, used the first 8 primes.

cirq/ops/three_qubit_gates_test.py
```diff
@@ -19,6 +19,7 @@
 
 import cirq
 
+FIRST_EIGHT_PRIMES = list(sympy.primerange(0, 20))
```
Strilanc: Haha, that's a bit much. Just hardcode this list in each method, to keep them as concrete and decoupled as possible.

bradyb: got it, done.

cirq/ops/three_qubit_gates_test.py
bradyb: Changed the angles used.

cirq/ops/three_qubit_gates.py
```diff
@@ -145,6 +149,132 @@ def __str__(self) -> str:
...
+@value.value_equality()
+class ThreeQubitDiagonalGate(gate_features.ThreeQubitGate):
+    """A gate given by a diagonal 8x8 matrix.
+
+    The list of angles passed to the constructor should be in radians.
+    """
+
+    def __init__(self,
+                 diag_angles_radians: List[Union[float, sympy.Basic]]) -> None:
+        self._diag_angles_radians = diag_angles_radians  # type: List[float]
```
Strilanc: This type annotation is incorrect. The actual type is List[Union[float, sympy.Basis]].

cirq/ops/three_qubit_gates.py
```diff
@@ -145,6 +149,132 @@ def __str__(self) -> str:
...
+            sweep_abc,
+            p_gates[6](c),
+            sweep_abc,
+        ]
+
+    def _apply_unitary_(self, args: protocols.ApplyUnitaryArgs) -> np.ndarray:
+        if self._is_parameterized_():
+            return NotImplemented
+        for index, angle in enumerate(self._diag_angles_radians):
+            slice_index = '0b' + str(index % 2) + str(index % 4 // 2) + str(
```
Strilanc: It looks like you're reversing the bit order? Instead of doing this, just directly pass in `index` and re-order the matrix rows or cols to make that work.

cirq/ops/three_qubit_gates.py
bradyb: Fixed.

cirq/ops/three_qubit_gates.py
bradyb: Right, I was ensuring the endianness of `index` because I noticed, for example, `args.subspace_index(1) = (slice(None, None, None), 1, 0, 0, Elllipsis)` and 
`args.subspace_index(2) = (slice(None, None, None), 0, 1, 0, Elllipsis)` and so on.

I made it less messy in my latest commit, but I think directly passing `index` to `args.subspace_index` will still require some of this twiddling but let me know if you think I'm missing something.

cirq/ops/three_qubit_gates.py
```diff
@@ -145,6 +149,132 @@ def __str__(self) -> str:
         return 'CCZ**{}'.format(self._exponent)
 
 
+@value.value_equality()
+class ThreeQubitDiagonalGate(gate_features.ThreeQubitGate):
+    """A gate given by a diagonal 8x8 matrix.
+
+    The list of angles passed to the constructor should be in radians.
```
Strilanc: Move this into the `Args:` section of the docstring of `__init__`

bradyb: Done!

dabacon:
I don't see any outstanding changes that requested that haven't been fixed.

Issue #1658, opened 2019-05-17
Simulator state shouldn't expose external details


Strilanc:
```
import cirq
d = cirq.asymmetric_depolarize(0.1, 0, 0)
s = cirq.DensityMatrixSimulator()
a, b = cirq.LineQubit.range(2)
out = s.simulate(cirq.Circuit.from_ops(d(a), d(b)))
assert out.final_simulator_state.density_matrix.shape == (4, 4)
```

Fails saying the shape is `(2, 2, 2, 2)`. If you use an empty circuit (and specify the qubits) you instead get shape `(4, 4)`.

dabacon:
Good catch. 

I think numpy is smart enough when you reshape tensors, but I assume we should always do (n, n) for density matrix or (n,) for state.  Should we consider adding versions of these that return the tensor form of these?

Strilanc:
I think we should always do `(n, n)` for now.

dabacon:
Ah note that you are doing `out.final_simulator_state`.  That's the internal "simulator state".  The right way to get this is `out.final_density_matrix`.

dabacon:
So yeah the issue here is that the public facing stuff should be in trial results and step results, but not in the "simulator state".  That way you can't do it the "wrong way" without considerable effort and using "_" classes.

dabacon:
Fixed in #1678

PR #1678, opened 2019-05-28
Hide simulator_state

dabacon:
The simulator state for simulators is an internal state used by the implementations for simulating state when stepping through a simulation.  It shouldn't be exposed "publically".  See #1658 for where callers where using this state instead of going to the public methods on step result or trial result.

### DIFF
--- a/cirq/sim/density_matrix_simulator.py
+++ b/cirq/sim/density_matrix_simulator.py
@@ -475,7 +475,7 @@ class DensityMatrixStepResult(simulator.StepResult):
         self._qubit_map = qubit_map
         self._dtype = dtype
 
-    def simulator_state(self) -> 'DensityMatrixSimulatorState':
+    def _simulator_state(self) -> 'DensityMatrixSimulatorState':
         return DensityMatrixSimulatorState(self._density_matrix,
                                            self._qubit_map)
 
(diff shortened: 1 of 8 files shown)


dabacon:
Fixes #1658

cirq/sim/simulator.py
```diff
@@ -397,8 +397,12 @@ def __init__(self,
         self.measurements = measurements or collections.defaultdict(list)
 
     @abc.abstractmethod
-    def simulator_state(self) -> Any:
-        """Returns the simulator_state of the simulator after this step.
+    def _simulator_state(self) -> Any:
+        """Returns the simulator state of the simulator after this step.
+
+        This method starts with an underscore to indicate that it is private.
+        State that should be exposed should exposed on the child StepResult
```
viathor: nit: This is a little convoluted. Maybe "Instead, use public state exposed by StepResult.".

cirq/sim/density_matrix_simulator.py
```diff
@@ -618,10 +619,10 @@ def __init__(self,
...
                         sorted(self.measurements.items())}
-        return (self.params, measurements, self.final_simulator_state)
+        return (self.params, measurements, self._final_simulator_state)
 
     def __repr__(self):
         return ("cirq.DensityMatrixTrialResult(params={!r}, measurements={!r}, "
-                "final_simulator_state={!r})"
-                .format(self.params, self.measurements,
-                        self.final_simulator_state))
+                "final_simulator_state={!r})".format(
```
Strilanc: If we don't want to expose the final simulator state, why are we including it in the trial results objects and printing it out?

dabacon: This is a repr. We want it to generate the code needed to create the object, no?  If it were __str__ I'd agree.

cirq/sim/simulator.py
dabacon: done

mpharrigan:
I was using `qubit_map` to pass to `cirq.pauli_string_expectation`. Seems sortof important to have this tagging along (in a public way) when implicitly labeling qubits by e.g. row index. Is there a preferred workaround?

dabacon:
we should expose it in the result methods.

PR #1680, opened 2019-05-29
Adjust kak coefficient canonicalize

### LINKED ISSUE #1647: instability in kak decomposition
In some cases the kak decomposition appears to have an instability in the sign of the ZZ coefficient. Consider the following example:
```
a, b = cirq.LineQubit(0), cirq.LineQubit(1)

theta = 0.3
circuit_1 = cirq.Circuit()
circuit_1.append(cirq.ISWAP.on(a, b))
circuit_1.append(
    cirq.ZZPowGate(exponent=2 * theta / numpy.pi, global_shift=-0.5).on(a, b))
print(cirq.kak_decomposition(cirq.unitary(circuit_1)).interaction_coefficients)

theta += 0.001
circuit_2 = cirq.Circuit()
circuit_2.append(cirq.ISWAP.on(a, b))
circuit_2.append(
    cirq.ZZPowGate(exponent=2 * theta / numpy.pi, global_shift=-0.5).on(a, b))
print(cirq.kak_decomposition(cirq.unitary(circuit_2)).interaction_coefficients)

print(numpy.amax(numpy.absolute(cirq.unitary(circuit_1) - cirq.unitary(circuit_2))))
```
returns
```
(0.7853981633974483, 0.7853981633974483, -0.2999999999999998)
(0.7853981633974482, 0.7853981633974482, 0.30099999999999993)
0.0009999999583333466
```
And thus we see that despite the fact the unitaries are very similar, the kak decomposition gives a seemingly random sign on the ZZ interaction coefficient. Is this a bug?

c-poole:
kak_canonicalize_vector now forces z>=0 when x=pi/4 case, addressing one of the symptons of #1647. There are likely additional issues that must be addressed before #1647 can be closed.

Part of #1647

### DIFF
--- a/cirq/linalg/decompositions.py
+++ b/cirq/linalg/decompositions.py
@@ -376,11 +376,16 @@ def kak_canonicalize_vector(x: float, y: float, z: float,
     Returns:
         The canonicalized decomposition, with vector coefficients (x2, y2, z2)
         satisfying:
+
             0 ≤ abs(z2) ≤ y2 ≤ x2 ≤ π/4
             if x2 = π/4, z2 >= 0
+
         Guarantees that the implied output matrix:
+
             g · (a1 ⊗ a0) · exp(i·(x2·XX + y2·YY + z2·ZZ)) · (b1 ⊗ b0)
+
         is approximately equal to the implied input matrix:
+
             exp(i·(x·XX + y·YY + z·ZZ))
     """
 
--- a/cirq/linalg/decompositions.py
+++ b/cirq/linalg/decompositions.py
@@ -363,20 +363,25 @@ class KakDecomposition:
             before)
 
 
-def kak_canonicalize_vector(x: float, y: float, z: float) -> KakDecomposition:
+def kak_canonicalize_vector(x: float, y: float, z: float,
+                            atol: float = 1e-8) -> KakDecomposition:
     """Canonicalizes an XX/YY/ZZ interaction by swap/negate/shift-ing axes.
 
     Args:
         x: The strength of the XX interaction.
         y: The strength of the YY interaction.
         z: The strength of the ZZ interaction.
+        atol: If any of the coefficients are less than atol, they will be
+        rounded down to zero
 
     Returns:
         The canonicalized decomposition, with vector coefficients (x2, y2, z2)
         satisfying:
 
-            0 ≤ abs(z2) ≤ y2 ≤ x2 ≤ π/4
-            z2 ≠ -π/4
+            0 ≤ z2 ≤ y2 ≤ x2 < π/2
+            x2 + y2 ≤ π/2
+            if z2 =0, kx ≤ π/4
+
 
         Guarantees that the implied output matrix:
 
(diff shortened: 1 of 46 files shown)


cirq/linalg/decompositions.py
```diff
@@ -363,20 +363,25 @@ def interaction_matrix(m: np.ndarray, c: float) -> np.ndarray:
...
 
     Returns:
         The canonicalized decomposition, with vector coefficients (x2, y2, z2)
         satisfying:
 
-            0 ≤ abs(z2) ≤ y2 ≤ x2 ≤ π/4
-            z2 ≠ -π/4
+            0 ≤ z2 ≤ y2 ≤ x2 < π/2
+            x2 + y2 ≤ π/2
+            if z2 =0, kx ≤ π/4
```
Strilanc: What's `k`?

Strilanc: I don't like this change to the canonical region.

1. It behaves surprisingly under small perturbations. For example, `XX**0.5 @ YY**0.4 @ ZZ**-epsilon` produces `(0.63, 0.63, tiny)` but then `XX**0.5 @ YY**0.4 @ ZZ**+epsilon` causes the X coefficient to jump up to `0.94`. Perturbing smallest interaction by a tiny amount caused the largest interaction to jump by 50%. Previously the only instability of this type was in the sign of the smallest interaction, and it only happened at the X=max plane. AFAIK you never saw magnitudes jumping around (bugs notwithstanding).

2. We have code in the code base that assumes "small interaction coefficient" is equivalent to "small two-qubit operation", and this code will break in the presence of values like pi/2 which actually correspond to small two-qubit operations.

3. The shape of the canonical region is no longer symmetric (with the exception of the X=max plane) about a plane passing through the origin.

    Current shape:

    image

    Proposed shape:

    image

    The complications of the new shape are particularly apparent in the fact that you needed an additional inequality in order to specify it.

Is this change to the region actually required? I would strongly prefer to stay with the existing region (except for the X=max plane where we were failing to canonicalize correctly, of course).

c-poole: The only reason I switched to this region is because the reference mentioned that there is a one-to-one correspondence between points in this set and equivalence classes in SU(4). It may also be possible to refactor the current region to maintain your "smallness" assumption and still achieve the one-to-one correspondence.

c-poole: A typo which will be fixed soon.

Strilanc: I agree that we want a one-to-one correspondence. But we can massage the shape of the region to be better.

c-poole: I've got a variation that changes our current region minimally. The only change is that if the XX coefficient = pi/4 within some tolerance, the ZZ term is forced to be positive. This change also makes @babbush's tests pass. I also verified that forcing the ZZ term to be negative works but arbitrarily chose positive since we have the freedom to choose.

c-poole:
This test takes substantially longer to run than would be acceptable to add to the pytest checks, but I just ran it on my machine and it passed. It scans the values of the xx and yy coefficients near where we might expect there to be issues based on the x=pi/4 special case and constructs a kak decomp with those coefficients with random local operations and global phase. It then perturbs the unitary of the resulting matrix and runs a decomp on the perturbed unitary, and finally checks that against the original unperturbed unitary. For each set of coefficient values, it does this 1000 times to give evidence that we didn't just get lucky with a specific choice of local operations.

```
import cirq
import numpy as np
from numpy.random import random as r


# Generate random local operations
def rl():
    return (cirq.testing.random_special_unitary(2),
            cirq.testing.random_special_unitary(2))


# 1e-9 is the default tolerance parameter for the canonicalize vector function
# so we look near x=pi/4, y=pi/4 with additional room on the order of 1e-9
e = 1.5*1e-9
for x in np.linspace(np.pi/4-e, np.pi/4+e, 5):
    for y in np.linspace(np.pi/4-e, np.pi/4+e, 5):
        for z in np.linspace(-np.pi/4, np.pi/4, 3):
            for trials in range(1000):

                # Get a kak decomp with desired interaction coefficients
                # and random local operations and global phase
                kak = cirq.KakDecomposition(global_phase=np.exp(np.pi*2j*r()),
                                            single_qubit_operations_before=rl(),
                                            interaction_coefficients=(x, y, z),
                                            single_qubit_operations_after=rl())

                # Get the unitary of the kak decomp and perturb each element
                # Perturbations are on the order of 1e-14 because larger
                # perturbations sometimes caused errors to be raised
                # in the decomposition code when it checks for conditions like
                # if a matrix is real or symmetric
                u = cirq.unitary(kak)
                for num in u:
                    u += np.random.random()*1e-14+np.random.random()*1j*1e-14

                # Get the kak decomp for the perturbed unitary
                newkak = cirq.kak_decomposition(u)

                # Assert that the unitary of the decomp of the perturbed matrix
                # is close to the unperturbed unitary
                assert np.allclose(cirq.unitary(newkak), cirq.unitary(kak))
```

cirq/linalg/decompositions.py
```diff
@@ -180,18 +169,14 @@ def kron_factor_4x4_to_2x2s(
         matrix: np.ndarray,
 ) -> Tuple[complex, np.ndarray, np.ndarray]:
     """Splits a 4x4 matrix U = kron(A, B) into A, B, and a global factor.
-
     Requires the matrix to be the kronecker product of two 2x2 unitaries.
     Requires the matrix to have a non-zero determinant.
     Giving an incorrect matrix will cause garbage output.
-
```
Strilanc: Don't remove these newlines. There should be a newline separating the sections of the documentation.

cirq/linalg/decompositions.py
```diff
@@ -340,7 +313,6 @@ def __repr__(self):
 
     def _unitary_(self):
         """Returns the decomposition's two-qubit unitary matrix.
-
```
Strilanc: (so basically revert everything up to this line)

cirq/linalg/decompositions.py
c-poole: That wasn't intended and honestly I'm not sure how I did it but I'll run through the files and undo it.  My primary concern with having (-pi/4+atol, +pi/4+atol] is that you are explicitly allowing two sets of coefficients which are equivalent up to local operations to exist in the canonical region. This seems like it could break code that tries to convert two operations that are equivalent up to local operations into each other. Granted, my choice to exclude x in (pi/4-atol, pi/4) and z<0 may cause problems in a similar way. My hope is that the test I posted provides evidence that this isn't the case, since operations that start in that region are tested.

PR #1688, opened 2019-05-30
Minor polish to QAOA example

kevinsung:
- removes function annotations for Python 2 compatibility
- puts factor of 2 on betas to match the paper convention


vtomole:
> removes function annotations for Python 2 compatibility

We are not supporting Python 2 anymore so we don't need to make our examples Python 2 compatible.

kevinsung:
Good point, but I still want to remove them here because I wasn't completely consistent within the same file the first time around.

PR #1690, opened 2019-05-30
Add __bool__ to cirq.PauliString

Strilanc:
Matches the behavior of other collections: empty is false, non-empty is true.

### DIFF
--- a/cirq/ops/pauli_string_test.py
+++ b/cirq/ops/pauli_string_test.py
@@ -616,3 +616,8 @@ def test_with_qubits():
     for q in new_qubits:
         assert new_pauli_string[q] == cirq.Pauli.by_index(q.x)
     assert new_pauli_string.coefficient == -1
+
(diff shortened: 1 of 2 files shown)


cirq/ops/pauli_string_test.py
```diff
@@ -616,3 +616,8 @@ def test_with_qubits():
     for q in new_qubits:
         assert new_pauli_string[q] == cirq.Pauli.by_index(q.x)
     assert new_pauli_string.coefficient == -1
+
+
+def test_bool():
+    assert not bool(cirq.PauliString({}))
+    assert bool(cirq.PauliString({a: cirq.X}))
```
viathor: I think you're missing `a = cirq.NamedQubit('a')` or something like this.

Strilanc: Oops

PR #1692, opened 2019-05-31
Add asyncio test methods

Strilanc:
These will be used in upcoming functionality for collecting samples in parallel.


cirq/testing/asynchronous.py
```diff
@@ -0,0 +1,106 @@
...
+        future: The future that should not yet be resolved.
+        timeout: The number of seconds to wait for the future. Make sure this is
+             a small value, because it holds up the passing test!
+
+    Raises:
+        AssertError: The future completed or failed within the timeout.
+    """
+    try:
+        asyncio.get_event_loop().run_until_complete(
+            asyncio.wait_for(asyncio.shield(future), timeout=0.001))
```
dabacon: pipe in variable timeout here

cirq/testing/asynchronous.py
```diff
@@ -0,0 +1,106 @@
...
+
+    Raises:
+        AssertError: The future did not complete in time, or did not contain
+            the expected result.
+    """
+    try:
+        actual = asyncio.get_event_loop().run_until_complete(
+            asyncio.wait_for(asyncio.shield(future), timeout=timeout))
+        if expected is not _just_return_result:
+            assert actual == expected, "<{!r}> != <{!r}> from <{!r}>".format(
```
maffoo: nit: shouldn't need the extra `<..>` here, since `repr` usually includes them for objects.

cirq/testing/asynchronous.py
maffoo: nit: could factor out this snippet into a helper method, since it appears in all of these tests:
```python
def _run_until_complete(future, timeout):
    return asyncio.get_event_loop().run_until_complete(
        asyncio.wait_for(asyncio.shield(future), timeout=timeout)
```

Strilanc: Done.

cirq/testing/asynchronous.py
Strilanc: Done.

cirq/testing/asynchronous.py
```diff
@@ -0,0 +1,106 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+from collections import Awaitable, Coroutine
+from typing import Any, Type, Union
+
+import pytest
+
+_just_return_result = ([],)  # type: Any
```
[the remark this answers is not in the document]
Strilanc: Done.

cirq/testing/asynchronous.py
Strilanc: Done.

Issue #1693, opened 2019-05-31
Should we bump our minimum python 3 version?


Strilanc:
Seeing as how we're dropping python 2 support, this seems like a natural time to do it.

I've run into a few issues recently w.r.t. things existing in the `typing` module or not. We could also start using features like data classes and attribute/variable type annotations.

vtomole:
+1 for bumping.

vtomole:
> data classes

So, versions `3.7+`

c-poole:
The biggest issue I see with doing something like this is whether or not we are limiting integration of cirq into projects that operate on earlier python 3 versions. We may want a release that is compatible with 3.5/3.6 so that it isn't a jump from 2.7, >=3.5 to 3.7 only

dabacon:
https://www.jetbrains.com/research/python-developers-survey-2018/ contains some info on version adoption

dabacon:
I'm not very familiar with how backward incompatible 3.x versions are.

FWIW: colab uses 3.6.7.

Strilanc:
We definitely don't want to become incompatible with colab.

vtomole:
Based on Dave's link, 54 percent of the users surveyed are on Python 3.6 vs 11% on 3.5.

Strilanc:
I think I've thought of a hacky solution to the 3.5-is-forever-confusingly-broken problem.

When we release the next package, we mark it as compatible with 3.5 (and even 2.7). But at import time we check if the version is 3.5, and if it is then we blow up with an error message saying "you either need to upgrade to python 3.5 or you need to install the old version of cirq==0.whatever".

This transforms a confusing error into a very loud and explained error, at the cost of doing the wrong thing for people who wanted `pip install cirq` to actually install the old version. But we keep telling people not to do that, so it may be acceptable.

terrillfrantz:
Folks, I am just a lurker to the thread at this point, but such a strong 
message is a good thing to do for those of us not using Cirq everyday, 
or are entirely new to it. Saves everyone--the noob and the development 
community--a lot of pain as you can well imagine.

PR #1695, opened 2019-06-04
Bump python version to 3.6 so that we may use new features

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/1693


Strilanc:
Replace the meat of `_version.py` with this code, which will allow us to properly warn people about dropping versions:

```python
import sys

if sys.version_info < (3, 6, 0):
    raise SystemError(
        "You installed the latest version of cirq but aren't on python 3.6+.\n"
        'To fix this error you need to either:\n'
        '\n'
        'A) Update to python 3.6 or later.\n'
        '- OR -\n'
        'B) Explicitly install an older deprecated-but-compatible version '
        'of cirq (e.g. "python -m pip install cirq==0.5.*")')


__version__ = "0.6.0.dev"
```

Strilanc:
The installation instructions need to be updated to talk about the new version.

vtomole:
@PTAL

cirq/_version.py
```diff
@@ -12,13 +12,14 @@
...
-"""Define version number here, read it from setup.py automatically, and warn users that
-the latest version of cirq uses python 3.6+"""
+"""Define version number here, read it from setup.py automatically,
+and warn users that the latest version of cirq uses python 3.6+"""
 
 import sys
 
 if sys.version_info < (3, 6, 0):
-    raise SystemError(
+    # TODO: Find the best way to cover this. Maybe mock sys.version_info
```
Strilanc: This does not need to be covered. Just mark it `# coverage: ignore`.

vtomole: Done.

PR #1696, opened 2019-06-04
Cache .benchmarks directory so performance checks may be compared with a previous run.

vtomole:
Part of: https://github.com/quantumlib/Cirq/issues/1124


vtomole:
What this does is compare the previous `.json` benchmarks that were generated to the current one. I've set the tolerance for test failure low at the moment ( Fail if Min is 10% slower for any test).

PTAL
@dabacon  @Strilanc

check/performance
```diff
@@ -0,0 +1,15 @@
...
+#
+# Usage:
+#     check/performance
+################################################################################
+
+# Get the working directory to the repo root.
+cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd $(git rev-parse --show-toplevel)
+
+pytest --benchmark-autosave --benchmark-only --benchmark-compare --benchmark-compare-fail=min:10%  \
```
viathor: IIUC, this makes the benchmark tests fail when minimum runtime of any test grows by more than 10%. Minimum statistic is very sensitive to occasional outliers. Let's make it more robust by using the mean instead.

PR #1697, opened 2019-06-04
Cross entropy benchmarking

XiaoMiQC:
Added experiment, test and example for doing cross entropy benchmarking on systems with >= 2 qubits.


cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,347 @@
...
+from typing import List, Set, Tuple, Sequence, Dict, Any
+
+from matplotlib import pyplot as plt
+from cirq import devices, ops, circuits, sim
+
+
+class CrossEntropyResult:
+    """Results from a cross-entropy benchmarking (XEB) experiment."""
+
+    def __init__(self, num_cycle_range: Sequence[int],
```
Strilanc: Change this pair of vectors into a `Dict[int, float]`, or into a `Sequence[CrossEntropyPair]` where `CrossEntropyPair = NamedTuple(...)`.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,347 @@
...
+        Args:
+            num_cycle_range: The different numbers of cycles (circuit depth) at
+                which the measurements were taken.
+            xeb_fidelities: The average XEB fidelity at each number of cycles.
+        """
+        self._num_cycles = num_cycle_range
+        self._xeb_fidelities = xeb_fidelities
+
+    @property
+    def data(self) -> Sequence[Tuple[int, float]]:
```
Strilanc: Give this tuple a name (`NamedTuple` from `typing`).

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+def cross_entropy_benchmarking(
+        sampler: sim.Sampler,
+        qubits: Sequence[devices.GridQubit],
+        two_qubit_gate: ops.gate_features.TwoQubitGate,
+        *,
+        num_circuits: int = 20,
+        repetitions: int = 1000,
+        num_cycle_range: Sequence[int] = range(2, 103, 10),
```
Strilanc: Rename to `cycles` or 'depths'; take `Union[int, Iterable[int]]` with both being flattened into a tuple of ints.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,347 @@
...
+        interaction_sequence: List[Set[Tuple[devices.GridQubit,
```
Strilanc: This is a pretty cumbersome argument to specify. I'm not sure how to make it better.

The way you separate out the two qubit gate is perhaps not necessary. This argument could be a `List[Set[Operation]]` and it would work just as well but be more flexible.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,347 @@
...
+                                             devices.GridQubit]]] = None,
+        use_tetrahedral_group: bool = False) -> CrossEntropyResult:
```
Strilanc: This should be replaced by a 'single qubit gates' argument.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,347 @@
...
+    r"""Cross-entropy benchmarking (XEB) of multiple qubits.
+
+    A total of M random circuits are generated, each of which is made of N
+    cycles where N = max(num_cycle_range). Every cycle contains randomly
```
Strilanc: Doesn't `N` iterate over the cycles, instead of being the max?

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,347 @@
...
+    generated single-qubit gates applied to each qubit, followed by two-qubit
+    gate(s) applied to one or more pairs of qubits. For each given number of
+    cycles n in num_cycle_range, the experiment performs the following:
+
+    1) Measure the bit-string probabilities P^(meas, m)_|...00>,
+    P^(meas, m)_|...01>, P^(meas, m)_|...10>, P^(meas, m)_|...11>... at the
+    end of circuit_mn for all m, where circuit_mn is the first n cycles of
```
Strilanc: We don't actually need it for all `m`. That only makes sense for small circuits where you can get enough samples to expect to estimate all the probabilities. In practice you may not even see some of the samples.

In fact, in some XEB-like tests (namely the HOG test; Heavy Output Generation), duplicate results are discarded to avoid the possibility of gaming the test. You don't want a broken machine that always returns 000000...0 to score highly because 000...0 happened to be above average.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,347 @@
...
+
+    2) Theoretically compute the expected bit-string probabilities
+    P^(exp, m)_|...00>,  P^(exp, m)_|...01>, P^(exp, m)_|...10>,
+    P^(exp, m)_|...11>... at the end of circuit_mn for all m.
+
+    3) Collect all measured probabilities into a single vector P_meas and
+    normalize P_meas such it sums to 1.
+
+    4) Collect all expected probabilities into a single vector P_exp. Set any
+    probabilities <1e-22 to 1e-22 to avoid singularities in log P_exp. Then
```
Strilanc: This is equivalent to omitting values that you didn't sample.

Equivalently, you could think of it as using Laplace's rule of succession. In the binary case your estimated probability is never 0/N, it's at least 1/(N+1).

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,347 @@
...
+
+    5) Calculate the "XEB fidelity" at this given circuit depth using the
+    following expressions:
+
+    s_incoherent = -np.sum(P_uniform * np.log(P_exp))
+    s_expected = -np.sum(P_exp * np.log(P_exp))
+    s_meas = -np.sum(P_meas * np.log(P_exp))
+    fidelity = (s_incoherent - s_meas) / (s_incoherent - s_expected)
+
+    where P_uniform = 1/(M * 2 ** num_qubits) is a constant probability
```
Strilanc: Why is M in this equation?

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,347 @@
...
+    expected from a uniform distribution, with num_qubits being the total
+    number of qubits.
+
+
+    Args:
+        sampler: The quantum engine or simulator to run the circuits.
+        qubits: The grid qubits included in the XEB experiment.
```
Strilanc: Do they have to be grid qubits?

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,347 @@
...
+    single_q_layers = []  # type: List[List[ops.OP_TREE]]
+    for i in range(num_layers):
+        rots_i = []  # type: List[ops.OP_TREE]
+        for j in range(num_qubits):
+            rots_i.extend([rot(qubits[j]) for rot in rot_ops[rand_nums[j, i]]])
+        single_q_layers.append(rots_i)
+    return single_q_layers
+
+
+def _interaction_sequence(qubits: Sequence[devices.GridQubit]
```
Strilanc: Rename to `_default_interaction_sequence`

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,347 @@
...
+    # probabilities obtained from a single trial.
+    probs_meas = {n: np.zeros((num_circuits, 2 ** num_qubits)) for n in
+                  num_cycle_range}
+    probs_exp = {n: np.zeros((num_circuits, 2 ** num_qubits)) for n in
+                 num_cycle_range}
+
+    for k in range(num_circuits):
+
+        # Generates one random XEB circuit with max(num_cycle_range) cycles.
+        # Then the first n cycles of the circuit are taken to generate
```
Strilanc: Is this safe to do? It doesn't introduce bad correlations into the results?

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Done. Changed to a sequence of NamedTuples.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Done.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Done.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: You are right. This is now corrected.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: This is now modified as suggested. The user would specify a list of sets of operations to be interleaved with the single-qubit gates. They can also specify nothing, in which case the circuits would just be single-qubit gates.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Done. Now, instead of specify a tetrahedral group, the user can choose to specify any arbitrary list of single-qubit gates to randomly choose from.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Since we now allow the user to specify the interleaved operations, indeed these qubits do not need to be grid qubits. I have modified the type hints to be `Qid` instead.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Done.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: This is what we do in experiment: use the first x cycles of a long circuit as short circuits. This gives the same fitted Pauli error rate as the case where all circuits are randomly chosen.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Great point! I have modified the doc-string to indicate that bit-strings not seen in experiment do not factor into the estimation for `s_meas`. I think no modification of the code is necessary (correct me if I'm wrong here), since p_meas for those bit-strings are 0 (therefore gives 0 in the summation anyways).

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: `P_exp` actually refers to the "expected" or theoretically computed probabilities here (`P_meas` is the experimental probability). I think setting `P_exp` to some arbitrarily small value is mostly to avoid the log throwing an error. We could have set it to 1e-25 or 1e-20 or any other really small number without affecting the value of the XEB fidelity.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: M is here because the vectors `P_exp`, `P_meas` and `P_uniform` include bit-string probabilities from all M random circuits, so to normalize each of them there is an additional division by M. I clarified this by including sentences in the doc-string stating this additional normalization.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,363 @@
...
+from cirq import devices, ops, circuits, sim
+
+CrossEntropyPair = NamedTuple('CrossEntropyPair', [('num_cycle', int),
+                                                   ('xeb_fidelity', float)])
+
+
+class CrossEntropyResult:
+    """Results from a cross-entropy benchmarking (XEB) experiment."""
+
+    def __init__(self, cross_entropy_pair: Sequence[CrossEntropyPair]):
```
Strilanc: pairs

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,363 @@
...
+        """
+        Args:
+            cross_entropy_pair: A sequence of NamedTuples, each of which
+                contains two fields: num_cycle which returns the circuit
+                depth as the number of cycles and xeb_fidelity which returns
+                the XEB fidelity after the give cycle number.
```
Strilanc: given

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,363 @@
...
+        """
+        self._data = cross_entropy_pair
+
+    @property
+    def data(self) -> Sequence[CrossEntropyPair]:
+        """Returns a sequence of NamedTuples, each of which contains a cycle
```
Strilanc: Summary should be one line, then perhaps more details in a paragraph.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,363 @@
...
+        number and the corresponding XEB fidelity.
+        """
+        return self._data
+
+    def plot(self, **plot_kwargs: Any) -> None:
+        """Plots the average XEB fidelity vs the number of cycles.
+
+        Args:
+            **plot_kwargs: Arguments to be passed to matplotlib.pyplot.plot.
```
Strilanc: Backticks (\`) around `matplotlib.pyplot.plot`

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,363 @@
...
+
+def cross_entropy_benchmarking(
+        sampler: sim.Sampler,
+        qubits: Sequence[ops.Qid],
+        *,
+        benchmark_ops: Sequence[Set[ops.OP_TREE]] = None,
+        num_circuits: int = 20,
+        repetitions: int = 1000,
+        cycles: Union[int, Iterable[int]] = range(2, 103, 10),
+        single_qubit_gates: List[List[ops.SingleQubitGate]] = None
```
Strilanc: It's not clear from the parameter name what the list nesting means. Make the name more descriptive.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,363 @@
...
+    Each circuit (circuit_m) from the M random circuits is further used to
+    generate a set of circuits {circuit_mn}, where circuit_mn is built from the
+    first n cycles of circuit_m. n spans all the values in cycles.
+
+    For each value n in cycles, the experiment performs the following:
+
+    1) Experimentally estimate the bit-string probabilities P^(meas, m)_|...00>,
+    P^(meas, m)_|...01>, P^(meas, m)_|...10>, P^(meas, m)_|...11>... at the
+    end of circuit_mn for all m. This is done by taking a total of R
+    projective measurements for each circuit_mn, histogram the collected
```
Strilanc: Is histogram a verb? You might mean binning.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,363 @@
...
+    2) Theoretically compute the expected bit-string probabilities
+    P^(exp, m)_|...00>,  P^(exp, m)_|...01>, P^(exp, m)_|...10>,
+    P^(exp, m)_|...11>... at the end of circuit_mn for all m and for all
+    possible bit-strings of the Hilbert space.
+
+    3) Collect measured probabilities of all M circuits into a single vector
+    P_meas and normalize P_meas by M such it sums to 1.
+
+    4) Collect all expected probabilities into a single vector P_exp. Set any
+    probabilities <1e-22 to 1e-22 to avoid singularities in log P_exp. Then
```
Strilanc: `>` -> `less than`

cirq/experiments/cross_entropy_benchmarking.py
Strilanc: Could you double-check with Sergio whether we still compute the logarithms of the probabilities? If we didn't, this would be an unnecessary value.

I don't like setting an arbitrary threshold like this in a statistical computation. If we're going to omit samples that didn't occur, we should just omit samples that didn't occur instead of doing it in a roundabout way.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,363 @@
...
+        num_d = len(benchmark_ops)
+    else:
+        num_d = 0
+    max_cycles = max(cycles)
+
+    if single_qubit_gates is None:
+        single_rots = _random_half_rotations(qubits, max_cycles)
+    else:
+        single_rots = _random_any_gates(qubits, single_qubit_gates, max_cycles)
+    all_circuits = []  # type: List[circuits.Circuit]
```
Strilanc: As of recently can write `all_circuits: List[circuits.Circuits] = []`.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,363 @@
...
+    for num_cycles in cycles:
+        circuit_exp = circuits.Circuit()
+        for i in range(num_cycles):
+            circuit_exp.append(single_rots[i])
+            if num_d > 0:
+                for op_set in benchmark_ops[i % num_d]:
+                    circuit_exp.append(op_set)
```
Strilanc: We should probably wrap these in an `ops.Moment` so that they don't get moved around. We want to exert control over the layout for XEB.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,363 @@
...
+    return all_circuits
+
+
+def _measure_prob_distribution(sampler: sim.Sampler, repetitions: int,
+                               qubits: Sequence[ops.Qid],
+                               circuit_list: List[circuits.Circuit]
+                               ) -> List[np.ndarray]:
+    all_probs = []  # type: List[np.ndarray]
+    num_states = 2 ** len(qubits)
+    for circuit_i in circuit_list:
```
Strilanc: What's the `_i` for? There's no variable `i` here.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,363 @@
...
+        for k, v in res_hist.items():
+            probs[k] = float(v) / float(repetitions)
+        all_probs.append(probs)
+    return all_probs
+
+
+def _xeb_fidelities(ideal_probs: Dict[int, np.ndarray],
+                    actual_probs: Dict[int, np.ndarray]) -> List[float]:
+    num_cycles = sorted(list(ideal_probs.keys()))
+    xeb_fidelity = []  # type: List[float]
```
Strilanc: This can be replaced by a simple list comprehension.

examples/cross_entropy_benchmarking_example.py
```diff
@@ -0,0 +1,31 @@
+import cirq
+
+
+def main():
+    # The device to run the experiment.
```
Strilanc: sampler, not device

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Done. Renamed to `cross_entropy_pairs`

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Done. Corrected.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Done.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Done.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: How does `scrambling_gates_per_cycle` sound? It's been changed to that.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Yes I meant binning. Changed the verb here.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Done.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: I have changed the XEB function to the linear version, which circumvents the issue of artificially clipping the probabilities to some small value. The code and doc-strings have all been modified to reflect this.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Great suggestion. All of the two-qubit ops are specified as `ops.Moment` now.

cirq/experiments/cross_entropy_benchmarking.py
Strilanc: It's not corrected.

cirq/experiments/cross_entropy_benchmarking.py
Strilanc: I think you may have not pushed your changes before saying done.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,363 @@
...
+            second cycle, CZ between q0 and q1 and CZ between q2 and q3 for
+            the third cycle and so on. If None, the circuits will consist only
+            of single-qubit gates.
+        num_circuits: The total number of random circuits to be used.
+        repetitions: The number of measurements for each circuit to estimate
+            the bit-string probabilities.
+        cycles: The different numbers of circuit layers in the XEB study.
+            Could be a single or a collection of values.
+        single_qubit_gates: If None (by default), the single-qubit
+            gates are chosen from X/2 (\pi/2 rotation around the X axis),
```
Strilanc: Use unicode character for pi, or use $ around latex.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Yes, still working on your last few comments. Will push the changes in a couple of hours.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Renamed to `circuit`.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Yes, changed the code here and now it is shorter.

examples/cross_entropy_benchmarking_example.py
XiaoMiQC: Corrected.

cirq/experiments/cross_entropy_benchmarking.py
XiaoMiQC: Done.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,374 @@
...
+            Y/2 ($\pi/2$ rotation around the Y axis) and (X + Y)/2 ($\pi/2$
+            rotation around an axis $\pi/4$ away from the X on the equator of
+            the Bloch sphere). Otherwise the single-qubit gates for each layer
+            are chosen from a list of possible choices (each choice is a list
+            of one or more single-qubit gates).
+
+    Returns:
+        A CrossEntropyResult object that stores and plots the result.
+    """
+    simulator = sim.Simulator()
```
dabacon: probably a simulator should be passed as a parameter and it defaults to this one.  That way you can override with a custom simulator.

XiaoMiQC: Thanks for the suggestion. It's done.

XiaoMiQC:
Problem with CLA. Close and reopen this PR.

Issue #1632, opened 2019-05-14
Partial trace doesn't work on state vector output


Strilanc:
I realize this is not really how the method was intended to use, but it felt natural as I was writing it.

```
import cirq
c = cirq.Circuit.from_ops(cirq.H.on_each(*cirq.LineQubit.range(3)))
r = cirq.Simulator().simulate(c)
m = cirq.partial_trace(
            r.final_simulator_state.state_vector.reshape((2, 2, 2)), [0])
```

causes

```
operand has more dimensions than subscripts given in einstein sum, but no '...' ellipsis provided to broadcast the extra dimensions.
```

peterse:
that reshape operation is incorrect. In general you have to promote the wavefunction to a density matrix with np.kron, then do the partial trace, then apply a "nearest kronecker product" factorization to get a wavefunction as the final form. Then this will give you a unique traced wf representation (mod global phase) IFF the circuit is completely unitary.

dabacon:
This would be a nice feature, but if we add support for this it should not be called just "partial trace" which is a well defined mathematical operation (what is described is a composition of outer product and partial trace).

peterse:
resolved in #1699 
for the record:
- `cirq.subwavefunction` attempts to factor a wavefunction into two pure states and return the pure wavefunction over the `kept_qubits` arg, or else fails
- `cirq.wavefunction_partial_trace_as_mixture` factors a wavefunction into a (possibly pure) mixture and returns a tuple representing that mixture.

PR #1699, opened 2019-06-06
Implement `subwavefunction` and `wavefunction_partial_trace`

peterse:
First pass at addressing #1632, to do a partial trace-like operation on a wavefunction

per @dabacon's suggestion, the method is kept separate from partial_trace because the process involved is  a composition of linalg operations instead of a single well-defined operation.

I feel like the user should be warned if they call `keep_qubits` in a way that returns a mixture (and therefore _not_ a wavefunction anymore) but I don't know how the library's standards for that sort of warning.


cirq/linalg/transformations.py
```diff
@@ -304,3 +305,47 @@ def partial_trace(tensor: np.ndarray,
...
+        return keep_rho
+
+    # FIXME: reasonable tolerance for purity check?
+    if approx_eq(np.trace(keep_rho @ keep_rho), 1):
+        w, v = np.linalg.eig(keep_rho)
+        # FIXME: reasonable tolerance for lone eigenvalue check?
+        return v[:, np.isclose(w, 1)].reshape((new_dim,))
+
+    # FIXME: not sure how to deal with warnings
+    print("Removing qubits from an entangled state resulted in mixture.")
```
Strilanc: The caller is either going to consider this to be an error, or they are going to want a sampled result from the mixture, or they are going to want the mixture (or density matrix or similar).

You should strongly distinguish between the use case where the caller expects the result to be pure, and the use case where the caller does not known. In fact these should be completely different methods with different names. One of them returns the single state or else fails, and the other always returns a mixture (which the caller can then inspect to see if it has length 1).

cirq/linalg/transformations.py
```diff
@@ -304,3 +305,47 @@ def partial_trace(tensor: np.ndarray,
...
+                  state).reshape((2, 2) * int(dim))
+    new_dim = 2**len(keep_indices)
+    keep_rho = partial_trace(rho, keep_indices).reshape((new_dim,) * 2)
+    if not any(keep_indices):
+        return keep_rho
+
+    # FIXME: reasonable tolerance for purity check?
+    if approx_eq(np.trace(keep_rho @ keep_rho), 1):
+        w, v = np.linalg.eig(keep_rho)
+        # FIXME: reasonable tolerance for lone eigenvalue check?
```
Strilanc: The success case of this method can be checked for significantly more efficiently than what is being done here. For each assignment of the other qubits, compute the 2-norm of the part of the array corresponding to that assignment, and then pick the one where the 2-norm was the largest. Normalize it to have a 2-norm of 1. This is the candidate wavefunction.

If the state factors into a "kept" and "not kept" part, the candidate wavefunction you just picked is equal to the kept part (modulo the fact that the global phase is not well defined). In order to verify that a candidate wavefunction is in fact a factor, you can dot product it against each other-assignment wavefunction and then add up the absolute values of the dot products. A result of 1 indicates perfect factoring.

In Quirk this is the strategy used by the amplitude display. The code is here: https://github.com/Strilanc/Quirk/blob/master/src/gates/AmplitudeDisplay.js#L65 . Unfortunately the code I linked is not easy to port over at all, because it is all tangled up with the way I go about executing it on the GPU.

If you create a circuit with an amplitude display and export the simulation data, it will include a "coherence measure" that is the value I was describing.

[image](https://algassert.com/quirk#circuit=%7B%22cols%22%3A%5B%5B%22H%22%5D%2C%5B%22%E2%80%A2%22%2C%22X%5E%E2%85%9B%22%5D%2C%5B%22Amps1%22%5D%5D%7D)


```
{
  [...]
 "displays": [
    [...]
    "serialized_id": "Amps1",
    "name": "Amplitude Display"
   },
   "data": {
    "coherence_measure": 0.9903004266906359,
    "superposition_phase_locked_state_index": 0,
    "ket": [
     {"r":0.7139325141906738,"i":0},
     {"r":0.6867600679397583,"i":0.13660506904125214}
    ],
    [...]
}
```

peterse: Oh cool this looks much faster! I think I follow but want to double check - by "each assignment of the other qubits" and "part of the array corresponding to that assignment" do you mean some index-masking scheme that would take in (say) a 2-qubit wf [sqrt(.5), sqrt(.5), 0, 0] and compute the 1-qubit wf's [1, 0] (qubit 0) and [.5, .5] (qubit 1)? At first glance it seems this would yield the same candidate wf's given an input of |+++> or sqrt(.5)*(|000>+|111>) the same way (though Quirk handles the mixed state outcome in the GHZ case which is a good sign)

Do you know if the "coherence measure" being computed is equal to the "purity" (Tr(rho*rho)) for the factored state? This is seems a necessary condition for the above factoring scheme to be complete.

cirq/linalg/transformations.py
peterse: Will do.

cirq/linalg/transformations.py
peterse: I might be implementing your algorithm incorrectly; the way I did it it fails on the the example (|00> + |11>)|+> state I mentioned above. I've pushed scratch code of my implementation so maybe you can see an issue there?

Strilanc: The coherence measure being computed is just some arbitrary thing that happens to equal 1 when pure; it doesn't correspond to any other measure AFAIK. Its only strength is that it is cheap to compute.

As an example of this not being a particularly ideal measure, it can vary based on operations performed on other qubits. For example, notice the text fading in and out as I move the QFT gate, corresponding to the measure going up and down:

Peek 2019-06-17 17-43

Strilanc: > by "each assignment of the other qubits" and "part of the array corresponding to that assignment" do you mean some index-masking scheme that would take in (say) a 2-qubit wf [sqrt(.5), sqrt(.5), 0, 0] and compute the 1-qubit wf's [1, 0] (qubit 0) and [.5, .5] (qubit 1)?

I mean this:

```python
import cirq
import math
import numpy as np
def _candidate_sub_wavefunction(kept_qubit_indices, wavefunction):
    return max(
        (
            cirq.slice_for_qubits_equal_to(kept_qubit_indices, k)
            for k in range(len(wavefunction) >> len(kept_qubit_indices))
        ),
        key=lambda c: np.linalg.norm(c, 2))
```

Strilanc: I mean this (I think; have not executed this code to test).

```python
import cirq
import math
import numpy as np
def _candidate_sub_wavefunction(kept_qubit_indices, wavefunction):
    # assumes wavefunction has shape (2,2,2,2,2,...)
    return max(
        (
            wavefunction[cirq.slice_for_qubits_equal_to(kept_qubit_indices, k)]
            for k in range(len(wavefunction) >> len(kept_qubit_indices))
        ),
        key=lambda c: np.linalg.norm(c, 2))
```

Strilanc: Note: edited the code after noticing a mistake.

cirq/linalg/transformations.py
```diff
@@ -304,3 +307,109 @@ def partial_trace(tensor: np.ndarray,
     left_indices = [keep_map[i] if i in keep_set else i for i in range(ndim)]
     right_indices = [ndim + i if i in keep_set else i for i in left_indices]
     return np.einsum(tensor, left_indices + right_indices)
+
+
+def wavefunction_partial_trace(wavefunction: np.ndarray,
+                               keep_indices: List[int],
+                               atol: Union[int, float] = 1e-8) -> Tuple:
```
Strilanc: Output type annotation doesn't specify what's in the tuple. Should probably be `List[Tuple[float, np.ndarray]]`

cirq/linalg/transformations.py
```diff
@@ -304,3 +307,109 @@ def partial_trace(tensor: np.ndarray,
...
+
+    States in the output mixture retain the shape (2,2,...).
+
+    Args:
+        wavefunction: A wavefunction to express over a qubit subset.
+        keep_indices: Which indices to express the wavefunction on.
+        atol: The tolerance for determining that a factored state is pure.
+
+    Returns:
+        A mixture representation of the partial trace of the wavefunction.
```
Strilanc: What representation is it? Describe it.

cirq/linalg/transformations.py
```diff
@@ -304,3 +307,109 @@ def partial_trace(tensor: np.ndarray,
...
+    """
+
+    tot_dims = 1 << len(wavefunction.shape)
+    keep_dims = 1 << len(keep_indices)
+    try:
+        # Attempt to do efficient state factoring.
+        state = keep_qubits(wavefunction, keep_indices).reshape(keep_dims, 1)
```
Strilanc: Modify `keep_qubits` to take an optional `default` argument similar to global methods like `cirq.unitary`. That way you can specify a default of `None` and avoid having an exception in an expected execution path.

cirq/linalg/transformations.py
```diff
@@ -304,3 +307,109 @@ def partial_trace(tensor: np.ndarray,
...
     return np.einsum(tensor, left_indices + right_indices)
+
+
+def wavefunction_partial_trace(wavefunction: np.ndarray,
+                               keep_indices: List[int],
+                               atol: Union[int, float] = 1e-8) -> Tuple:
+    """Computes a mixture representing the partial trace of a wavefunction.
+
+    The input wavfunction must have shape `(2,2,2,...)`. Attempt to factor the
+    wavefunction into a pure state over `keep_indices`. If this fails, a non-
```
Strilanc: I would be a bit vaguer here and just say that this method is more efficient for pure states, and describe that running time (e.g. it's `O(2^n)` where n is the number of qubits, whereas for mixed states it will I think be `O(2^n + 4^k)` or maybe even `+8^k` where k is the number of kept qubits.)

cirq/linalg/transformations.py
```diff
@@ -304,3 +307,109 @@ def partial_trace(tensor: np.ndarray,
...
+        state = keep_qubits(wavefunction, keep_indices).reshape(keep_dims, 1)
+        return ((1.0, state))
+    except ValueError:
+        # Fall back to a (non-unique) mixture representation.
+        rho = np.kron(np.conj(wavefunction.reshape(tot_dims, 1)).T,
+                      wavefunction.reshape(tot_dims, 1)).reshape((2, 2) * len(wavefunction.shape))
+        keep_rho = partial_trace(rho, keep_indices).reshape((keep_dims,) * 2)
+        w, v = np.linalg.eig(keep_rho)
+        mixture = tuple(zip(w, [vec.reshape((2,) * len(keep_indices)) for vec in v.T]))
+        return tuple(filter(lambda p: not approx_eq(p[0], 0.0), mixture))
```
Strilanc: Use a list comprehension instead of `filter`

cirq/linalg/transformations.py
```diff
@@ -304,3 +307,109 @@ def partial_trace(tensor: np.ndarray,
...
+        # Attempt to do efficient state factoring.
+        state = keep_qubits(wavefunction, keep_indices).reshape(keep_dims, 1)
+        return ((1.0, state))
+    except ValueError:
+        # Fall back to a (non-unique) mixture representation.
+        rho = np.kron(np.conj(wavefunction.reshape(tot_dims, 1)).T,
+                      wavefunction.reshape(tot_dims, 1)).reshape((2, 2) * len(wavefunction.shape))
+        keep_rho = partial_trace(rho, keep_indices).reshape((keep_dims,) * 2)
+        w, v = np.linalg.eig(keep_rho)
+        mixture = tuple(zip(w, [vec.reshape((2,) * len(keep_indices)) for vec in v.T]))
```
Strilanc: This is the right zip ordering? It looks like you're zipping for each item instead of zipping the weights with the vectors.

cirq/linalg/transformations.py
```diff
@@ -304,3 +307,109 @@ def partial_trace(tensor: np.ndarray,
...
+        return tuple(filter(lambda p: not approx_eq(p[0], 0.0), mixture))
+
+
+def keep_qubits(wavefunction: np.ndarray,
```
Strilanc: Rename this method to `subwavefunction`.

cirq/linalg/transformations.py
```diff
@@ -304,3 +307,109 @@ def partial_trace(tensor: np.ndarray,
...
+                keep_indices: List[int],
+                atol: Union[int, float] = 1e-8) -> np.ndarray:
```
Strilanc: Add a `*,` parameter before `atol`, forcing it to be specified by keyword

cirq/linalg/transformations.py
```diff
@@ -304,3 +307,109 @@ def partial_trace(tensor: np.ndarray,
...
+    """Return a representation of the given state over a subset of its qubits.
```
Strilanc: `Attempts to factor a wavefunction into two parts, and returns one of them.`

cirq/linalg/transformations.py
```diff
@@ -304,3 +307,109 @@ def partial_trace(tensor: np.ndarray,
...
+
+    The input wavfunction must have shape `(2,2,2,...)`.
+
+    A given wavefunction can be efficiently expressed over a subset of its
+    qubits by comparing magnitudes of unnormalized wavefunctions that remain
+    after each possible state over the qubits to be removed is sliced out of
```
Strilanc: You don't need to describe the algorithm in this much detail in the documentation. Instead, focus on the algebraic properties that the user needs to think about, such as the fact that the wavefunction must be of the form A tensor B where A is a pure state over the kept qubits and B is a pure state over the discarded qubits.

cirq/linalg/transformations.py
```diff
@@ -304,3 +307,109 @@ def partial_trace(tensor: np.ndarray,
...
+
+    Raises:
+        ValueError: if the wavefunction is not of the correct shape or the
+        indices are not a valid subset of the input wavefunction's indices, or
+        the result of factoring is not a pure state.
+    """
+
+    if not np.isclose(np.linalg.norm(wavefunction), 1):
+        raise ValueError("Input state must be normalized.")
+    if not len(set(keep_indices)) == len(keep_indices):
```
Strilanc: Use `!=` instead of `not ==`

cirq/linalg/transformations.py
```diff
@@ -304,3 +307,109 @@ def partial_trace(tensor: np.ndarray,
...
+        raise ValueError(
+            "keep_indices were {} but must be unique.".format(keep_indices))
+    n_qubits = len(wavefunction.shape)
+    if n_qubits < len(keep_indices) or \
```
Strilanc: Use parens for line continuations instead of `\`

Strilanc: Don't need the `n_qubits < len` check. It's implied by the distinctness + in range checks.

cirq/linalg/transformations.py
```diff
@@ -304,3 +307,109 @@ def partial_trace(tensor: np.ndarray,
...
+        for k in range(1 << len(other_qubits))
+    ]
+    # The coherence measure is computed using unnormalized candidates.
+    best_candidate = max(candidates, key=lambda c: np.linalg.norm(c, 2))
+    left = np.conj(best_candidate.reshape((keep_dims,))).T
+    coherence_measure = sum([
+        abs(np.dot(left, c.reshape((keep_dims,))))
+        for c in candidates
+    ])
+    print(coherence_measure)
```
Strilanc: Remove debug lines

cirq/linalg/transformations.py
```diff
@@ -304,3 +307,109 @@ def partial_trace(tensor: np.ndarray,
...
+            "keep_indices were {} but must be unique.".format(keep_indices))
+    n_qubits = len(wavefunction.shape)
+    if n_qubits < len(keep_indices) or \
+            any([ind >= n_qubits for ind in keep_indices]):
+        raise ValueError(
+            "keep_indices {} are an invalid subset of the input wavefunction.")
+
+    keep_dims = 1 << len(keep_indices)
+    other_qubits = sorted(set(range(n_qubits)) - set(keep_indices))
+    candidates = [
```
Strilanc: Use a generator `(...)` instead of a list `[...]` comprehension to save space.

cirq/linalg/transformations.py
```diff
@@ -304,3 +307,109 @@ def partial_trace(tensor: np.ndarray,
...
+        mixture = tuple(zip(w, [vec.reshape((2,) * len(keep_indices)) for vec in v.T]))
+        return tuple(filter(lambda p: not approx_eq(p[0], 0.0), mixture))
+
+
+def keep_qubits(wavefunction: np.ndarray,
+                keep_indices: List[int],
+                atol: Union[int, float] = 1e-8) -> np.ndarray:
+    """Return a representation of the given state over a subset of its qubits.
+
+    The input wavfunction must have shape `(2,2,2,...)`.
```
Strilanc: wavefunction

Strilanc: This method should also work on a shape like `(2^n,)`, since that is what is returned by simulators.

cirq/linalg/transformations_test.py
```diff
@@ -393,3 +393,169 @@ def test_partial_trace_invalid_inputs():
...
+        np.abs(cirq.keep_qubits(state, [0, 1])), a.reshape(2,2))
+    np.testing.assert_almost_equal(
+        np.abs(cirq.keep_qubits(state, [2, 3, 4])), b.reshape(2,2,2))
+
+    np.testing.assert_almost_equal(
+        np.abs(cirq.keep_qubits(state, [0, 1, 2, 3, 4])),
+        np.kron(a, b).reshape(2,2,2,2,2))
+
+
+# def test_keep_qubits():
```
Strilanc: Commented out test?

cirq/linalg/transformations_test.py
```diff
@@ -393,3 +393,169 @@ def test_partial_trace_invalid_inputs():
...
+            np.arange(16).reshape(2,2,2,2) / np.linalg.norm(np.arange(16)),
+            [1, 2, 2])
+    with pytest.raises(ValueError, match='invalid'):
+        cirq.keep_qubits(np.array([1,0,0,0]).reshape(2,2), [5])
+    with pytest.raises(ValueError, match='invalid'):
+        cirq.keep_qubits(np.array([1,0,0,0]).reshape(2,2), [0, 1, 2])
+
+
+def test_wavefunction_partial_trace_invalid_input():
+    # TODO
```
Strilanc: DO

cirq/linalg/transformations_test.py
```diff
@@ -393,3 +393,169 @@ def test_partial_trace_invalid_inputs():
...
+        assert mixtures_equal(mixture, truth)
+
+    truth = ((0.5, np.array([1, 0, 0, 0]).reshape(2,2)),
+             (0.5, np.array([0, 0, 0, 1]).reshape(2,2)))
+    for (q1, q2) in [(0, 1), (0, 2), (1, 2)]:
+        mixture = cirq.wavefunction_partial_trace(state, [q1, q2])
+        assert mixtures_equal(mixture, truth)
+
+
+if __name__ == "__main__":
```
Strilanc: Remove this block. Use `pytest` to execute tests.

cirq/linalg/transformations_test.py
```diff
@@ -393,3 +393,169 @@ def test_partial_trace_invalid_inputs():
...
         cirq.partial_trace(
             np.reshape(np.arange(2 * 2 * 2 * 2), (2,) * 4), [2])
+
+
+def test_keep_qubits():
+
+    bell00 = np.array([1, 0, 0, 1]) / np.sqrt(2)
+    plus_x = np.array([1, 1]) / np.sqrt(2)
+    state = np.kron(bell00, plus_x).reshape(2,2,2)
+    print("BELL", cirq.keep_qubits(state, [0, 1]))
```
Strilanc: Remove debug code

cirq/linalg/transformations_test.py
```diff
@@ -393,3 +393,169 @@ def test_partial_trace_invalid_inputs():
...
+#         np.abs(cirq.keep_qubits(state, [0, 1, 2, 3, 4])),
+#         np.kron(a, b).reshape(2,2,2,2,2))
+#     np.testing.assert_almost_equal(
+#         np.abs(cirq.keep_qubits(state, [0, 1, 5, 6, 7, 8])),
+#         np.kron(a, c).reshape(2,2,2,2,2,2))
+#     np.testing.assert_almost_equal(
+#         np.abs(cirq.keep_qubits(state, [2, 3, 4, 5, 6, 7, 8])),
+#         np.kron(b, c).reshape(2,2,2,2,2,2,2))
+
+
```
Strilanc: Write a test that uses the `cirq.testing.random_superposition` method to make two wavefunctions, then `np.kron` them together with some swaps, then see if the factoring splits them up correctly.

cirq/linalg/transformations.py
peterse: good catch

cirq/linalg/transformations.py
peterse: That might cause confusion with `cirq.partial_trace` which expects the user to reshape a density matrix to ((2,2)*n). If `cirq.subwavefunction` (fka keep_qubits) allows shape `(2^n,)`, partial trace probably should as well for consistency.

cirq/linalg/transformations_test.py
peterse: pushed by mistake..

cirq/linalg/transformations.py
peterse: candidates gets addressed twice; a generator would get exhausted

cirq/linalg/transformations.py
peterse: I've relabeled the variables for clarity

cirq/linalg/transformations_test.py
peterse: I've arranged for `wavefunction_partial_trace` to immediately call `subwavefunction`, and therefore rely on the latter's input validation. Is input validation testing for `wavefunction_partial_trace` still necessary?

peterse:
@Strilanc addressed most of your comments. A few remaining issues:

- `subwavefunction` and `wavefunction_partial_trace` now conform to the input wavefunction's shape. I think this is desirable since users likely expect a shape like `(2^n)` but most of the linalg tools require a shape like `(2, 2, ...)`. It feels like the same should be done for `partial_trace`. Also this would be better placed in a predicate - see #1750 
- I changed `subwavefunction` docstring to discuss outer products of kets, but I'm not sure what LaTeX packages your readthedocs is using. `\ket{\psi}` would compile in a nicer way at the expense of text readability that something like `|x>` has.

cirq/linalg/transformations.py
Strilanc: ah

Strilanc:
> now conform to the input wavefunction's shape

Great!

> I'm not sure what LaTeX packages your readthedocs is using.

You can use $|\psi\rangle$ and it should render correctly in the docs. You can use the `dev_tools/build-docs.sh` script to produce the html documentation output to check it and be sure.

cirq/linalg/transformations.py
```diff
@@ -304,3 +317,152 @@ def partial_trace(tensor: np.ndarray,
...
     right_indices = [ndim + i if i in keep_set else i for i in left_indices]
     return np.einsum(tensor, left_indices + right_indices)
+
+
+def wavefunction_partial_trace(
+        wavefunction: np.ndarray,
+        keep_indices: List[int],
+        *,
+        atol: Union[int, float] = 1e-8) -> Tuple[Tuple[float, np.ndarray]]:
+    """Attempts to factor a wavefunction into two parts and return one of them.
```
Strilanc: This is not a correct description of what the method does.

`Returns a wavefunction mixture equivalent to the result of only keeping some qubits.`

Or something like that

cirq/linalg/transformations.py
```diff
@@ -304,3 +317,152 @@ def partial_trace(tensor: np.ndarray,
...
+
+    The input wavefunction must have shape `(2, 2, ..., 2)` or `(2 ** N)` where
+    `wavefunction` is expressed over N qubits. Attempts to factor the
```
Strilanc: The optimization for the pure case should only be mentioned in passing at the end, because it's not something the user needs to understand in order to use the method.

cirq/linalg/transformations.py
```diff
@@ -304,3 +317,152 @@ def partial_trace(tensor: np.ndarray,
...
+    wavefunction into a pure state over `keep_indices`. If this fails, a non-
+    unique mixture can be computed by eigendecomposition on the partial trace of
+    the input's density matrix.
+
+    States in the output mixture will retain the same type of shape as the input
+    wavefunction, either `(2 ** k)` or `(2, 2, ..., 2)` where k is the number of
```
Strilanc: `(2,) * k` to be unambiguous.

cirq/linalg/transformations.py
```diff
@@ -304,3 +317,152 @@ def partial_trace(tensor: np.ndarray,
...
+        keep_indices: Which indices to express the wavefunction on.
+        atol: The tolerance for determining that a factored state is pure.
+
+    Returns:
+        A single-component mixture in which the factored wavefunction has
+        probability '1' if the factored state is pure, or else a mixture of the
+        default eigendecomposition of the mixed state's partial trace.
+
+    Raises:
+        ValueError: if the input wavefunction shape is neither `(2 ** N)` nor
```
Strilanc: `neither 2**n` -> `not an array whose length is a power of 2, or a tensor whose shape is of the form (2, 2, ..., 2).`

cirq/linalg/transformations.py
```diff
@@ -304,3 +317,152 @@ def partial_trace(tensor: np.ndarray,
...
+
+    # Attempt to do efficient state factoring.]
+    state = subwavefunction(wavefunction, keep_indices, default=None, atol=atol)
+    if state is not None:
+        return ((1.0, state),)
+
+    # Fall back to a (non-unique) mixture representation.
+    if wavefunction.shape == (wavefunction.size,):
+        return_shape = -1
+    elif wavefunction.shape == (2,) * int(np.log2(wavefunction.size)):
```
Strilanc: `elif all(e == 2 for e in wavefunction.shape):`

Or add 0.5 to the log output, to avoid the assumption that there is no floating point error in the log2 for powers of 2.

cirq/linalg/transformations.py
```diff
@@ -304,3 +317,152 @@ def partial_trace(tensor: np.ndarray,
     left_indices = [keep_map[i] if i in keep_set else i for i in range(ndim)]
     right_indices = [ndim + i if i in keep_set else i for i in left_indices]
     return np.einsum(tensor, left_indices + right_indices)
+
+
+def wavefunction_partial_trace(
```
Strilanc: This name is much better than it used to be, but it does suggest that the output is going to be a density matrix.

Maybe `wavefunction_partial_trace_as_mixture`? So long...
Maybe `subwavefunction_mixture`?

cirq/linalg/transformations.py
```diff
@@ -304,3 +317,152 @@ def partial_trace(tensor: np.ndarray,
...
+    r"""Return a representation of the given state over a subset of its qubits.
+
+    The input wavefunction must have shape `(2, 2, ..., 2)` or `(2 ** N)` where
+    `wavefunction` is expressed over N qubits. The returned array will retain
+    the same type of shape as the input wavefunction, either `(2 ** k)` or
+    `(2, 2, ..., 2)` where k is the number of qubits kept.
+
+    If a wavefunction |\psi> defined on N qubits is an outer product of kets
+    like  |\psi> = |x> \otimes |y>, and |x> is defined over the subset
+    `keep_indices` of  N qubits, then this method will factor |\psi> into |x>
```
Strilanc: `of k qubits`

cirq/linalg/transformations.py
```diff
@@ -304,3 +317,152 @@ def partial_trace(tensor: np.ndarray,
...
+    and |y> and return |x>.  Global phase of the input state will not
```
Strilanc: I would rephrase as

```
Note that $|x\rangle$ is not unique, because
$(e^{i \theta} |y\rangle) \otimes (|x\rangle) = (|y\rangle) \otimes (e^{i \theta} |x\rangle)$
. This method randomizes the global phase of $|x\rangle$ in order to avoid
accidental reliance on it.
```

cirq/linalg/transformations.py
```diff
@@ -304,3 +317,152 @@ def partial_trace(tensor: np.ndarray,
...
+    best_candidate = max(candidates, key=lambda c: np.linalg.norm(c, 2))
+    best_candidate = best_candidate / np.linalg.norm(best_candidate)
+    left = np.conj(best_candidate.reshape((keep_dims,))).T
+    coherence_measure = sum([
+            abs(np.dot(left, c.reshape((keep_dims,)))) ** 2
+            for c in candidates
+        ])
+
+    if approx_eq(coherence_measure, 1, atol=atol):
+        return best_candidate.reshape(return_shape)
```
Strilanc: Multiply the result by `np.exp(2j * np.pi * random.random())` to ensure users don't accidentally rely on the global phase of what we're returning, which will change unpredictably if we update the method.

cirq/linalg/transformations_test.py
Strilanc: Yes.

Justifying omitting a test by pointing at the current implementation of the code is very risky business. In a truly ideal world, the tests would be specified completely independent of the implementation so that they cross-check each other. Of course in practice you often use knowledge of the implementation to make the tests more effective, but I prefer to use that as a weapon to make the tests harder rather than a shortcut to make them easier.

cirq/linalg/transformations_test.py
```diff
@@ -393,3 +394,203 @@ def test_partial_trace_invalid_inputs():
...
+        cirq.wavefunction_partial_trace(state, [0, 1, 5, 6, 7, 8], atol=1e-8),
+        ((1.0, np.kron(a, c).reshape(2, 2, 2, 2, 2, 2)),))
+    assert mixtures_equal(
+        cirq.wavefunction_partial_trace(
+            state, [2, 3, 4, 5, 6, 7, 8], atol=1e-8),
+        ((1.0, np.kron(b, c).reshape(2, 2, 2, 2, 2, 2, 2)),))
+
+    # Return mixture will defer to numpy.linalg.eig's builtin tolerance.
+    assert mixtures_equal(
+        cirq.wavefunction_partial_trace(state, [0, 1], atol=1e-20),
```
Strilanc: You're using a strict tolerance in the call, but not in the check.

cirq/linalg/transformations.py
```diff
@@ -319,11 +319,11 @@ def partial_trace(tensor: np.ndarray,
     return np.einsum(tensor, left_indices + right_indices)
 
 
-def wavefunction_partial_trace(
-        wavefunction: np.ndarray,
-        keep_indices: List[int],
-        *,
-        atol: Union[int, float] = 1e-8) -> Tuple[Tuple[float, np.ndarray]]:
+def wavefunction_partial_trace(wavefunction: np.ndarray,
+                               keep_indices: List[int],
```
Strilanc: Rename `keep_indices` to `kept_qubit_indices`

cirq/linalg/transformations.py
peterse: cool, I like that idea.

cirq/linalg/transformations.py
peterse: I like how explicit `wavefunction_partial_trace_as_mixture` is

cirq/linalg/transformations_test.py
peterse: fixed.

peterse:
So I can't figure out what's up with mypy; do I need to explicitly cast the components of the mixture returned by `wavefunction_partial_trace_as_mixture` into `Tuple[Tuple[float, np.ndarray]]`? It currently returns `tuple([(float(p[0]), p[1]) for p in mixture if not approx_eq(p[0], 0.0)])` which seems fine because `p[1]` is definitely an np.array...

Similarly, its picking up a weird assignment error for `ret_shape = -1`. Not sure if there's an underspecified grep/regex thinking that's a return statement...?

Strilanc:
A `Tuple[X]` is a *single-element* tuple. If you want an arbitrary item count tuple you need `Tuple[X, ...]`. This explains the first mypy error:

`cirq/linalg/transformations.py:370: error: Incompatible return value type (got "Tuple[Tuple[float, Any], ...]", expected "Tuple[Tuple[float, Any]]")`

It's saying that the result is an arbitrary length tuple but you specified singleton.

For the second error, the issue is that in one branch of the `if` you are assigning an `int` (causing mypy to infer the variable's type as `int`) while in the other branch you are assigning a tuple. One solution is to pre-declare the value before the if-block like

```
        ret_shape: Union[int, Tuple[int, ...]]
        n_qubits = int(np.log2(wavefunction.size))
        if wavefunction.shape == (wavefunction.size,):
            ret_shape = -1
```

Strilanc:
*sigh* sometimes I think I should add CLA combat to CirqBot.

PR #1707, opened 2019-06-10
PauliSum

### LINKED ISSUE #1682: PauliString, LinearDict, QubitOperator
There are multiple schemes for representing operators that are sums of pauli strings. OpenFermion has `QubitOperator`. Cirq has `PauliString` and `LinearDict` but lacks a way to combine the two.

## Background
`QubitOperator` is a mapping (dictionary) from coefficients to terms, and each term is a tuple of single-qubit operators, which itself is a tuple of (`qubit`, `operator`). 

`LinearDict` is a mapping from coefficients to terms, but `PauliString` (the term in this case) has its own coefficient.

## Desired features
Users should be able to write something like `exp(1j * t * (cirq.X(a) * cirq.Y(b) + cirq.Z(a)))` and have it work. Users should be able to trivially convert between `QubitOperator` from OpenFermion.

## Proposal

Make a special `PauliLinearDict` that subclasses from `LinearDict`, teach it the rules of pauli multiplication and simplification. Instead of using full `PauliString`s as keys for the dict (which would result in redundant and inconsistent coefficients), I propose using only the frozenset of `(qubit, Pauli)` in the `PauliString` and dropping the coefficient. 

To avoid hairy issues of how the user indexes into `PauliLinearDict`, I suggest removing the ability for users to index into the dict and *only* support iterating over the terms in the sum. Under-the-hood we would re-construct `PauliString`s with coefficients and `yield` them.

xref #1439 #1299

mpharrigan:
Use a LinearDict under-the-hood to implement a sum of PauliStrings

fixes #1682


cirq/ops/linear_combinations.py
```diff
@@ -205,3 +207,54 @@ def extend(expansion: value.LinearDict[str],
...
+    add and subtract terms and then iterate over the resulting
+    (simplified) expression.
+
+    Under the hood, this class is backed by a LinearDict with coefficient-less
+    PauliStrings as keys. PauliStrings are reconstructed on-the-fly during
+    iteration.
+
+    """
+
+    def __init__(self, linear_dict):
```
Strilanc: Add a type annotation.

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +207,54 @@ def extend(expansion: value.LinearDict[str],
...
+        self._linear_dict = linear_dict
```
Strilanc: What if the user passes in a linear dict containing Pauli strings that have non-unit coefficients?

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +207,54 @@ def extend(expansion: value.LinearDict[str],
...
+
+    @classmethod
+    def from_pauli_strings(cls, terms: List[PauliString]):
```
Strilanc: Add output type annotation

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +207,54 @@ def extend(expansion: value.LinearDict[str],
...
+        terms = {pstring.unit(): pstring.coefficient for pstring in terms}
+        return cls(linear_dict=value.LinearDict(terms))
+
+    def __iter__(self):
+        for vec, coeff in self._linear_dict.items():
+            yield PauliString.from_unit(vec, coeff)
+
+    def __iadd__(self, other):
```
Strilanc: `__add__`, `__sub__`, `__mul__`, `__rmul__`, `__imul__`, and `__truediv__` should also be defined.

truediv and rmul only need to work in the case where the other value is a number, but the others should work for number or PauliString or PauliSum.

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +207,54 @@ def extend(expansion: value.LinearDict[str],
...
+        if isinstance(other, PauliString):
+            other = PauliSum.from_pauli_strings([other])
+
```
Strilanc: If other is a float, int, or complex, then it should become the coefficient of an empty Pauli product (i.e. of the identity).

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +207,54 @@ def extend(expansion: value.LinearDict[str],
...
+    def __isub__(self, other):
+        if isinstance(other, PauliString):
+            other = PauliSum.from_pauli_strings([other])
+        self._linear_dict -= other
+        return self
+
+    def __repr__(self) -> str:
+        return self._linear_dict.__repr__()
+
+    def __format__(self, format_spec: str) -> str:
```
Strilanc: I did not know this method existed! This is fantastic.

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +207,54 @@ def extend(expansion: value.LinearDict[str],
...
+        terms = [(PauliString.from_unit(v), self._linear_dict[v])
+                 for v in self._linear_dict.keys()]
+        return linear_dict._format_terms(terms=terms, format_spec=format_spec)
+
+    def __str__(self):
```
Strilanc: Add a `__repr__`.

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +207,54 @@ def extend(expansion: value.LinearDict[str],
...
+        self._linear_dict += other._linear_dict
+        return self
+
+    def __isub__(self, other):
+        if isinstance(other, PauliString):
+            other = PauliSum.from_pauli_strings([other])
+        self._linear_dict -= other
+        return self
+
+    def __repr__(self) -> str:
```
Strilanc: Implement `__eq__`, `__hash__`, `_approx_equal_`, and the new `_equal_up_to_global_phase_` (may not be merged yet). Note that you can use `@value.value_equality(approximate=true)` for the first three.

cirq/ops/linear_combinations_test.py
```diff
@@ -562,3 +562,20 @@ def test_linear_combination_of_operations_has_correct_pauli_expansion(
...
 def test_operation_expressions(expression, expected_result):
     assert_linear_combinations_are_equal(expression, expected_result)
+
+
+def test_pauli_sum_from_adding_strings():
+    q = cirq.LineQubit.range(2)
+    pstr1 = cirq.X(q[0]) * cirq.X(q[1])
+    pstr2 = cirq.Y(q[0]) * cirq.Y(q[1])
+    psum = pstr1 + pstr2
+    assert str(psum) == "1.000*X(0)*X(1)+1.000*Y(0)*Y(1)"
```
Strilanc: Once you have `__eq__` implemented, use it instead of these string comparisons.

cirq/ops/linear_combinations_test.py
```diff
@@ -562,3 +562,20 @@ def test_linear_combination_of_operations_has_correct_pauli_expansion(
 ))
 def test_operation_expressions(expression, expected_result):
     assert_linear_combinations_are_equal(expression, expected_result)
+
+
+def test_pauli_sum_from_adding_strings():
+    q = cirq.LineQubit.range(2)
+    pstr1 = cirq.X(q[0]) * cirq.X(q[1])
+    pstr2 = cirq.Y(q[0]) * cirq.Y(q[1])
+    psum = pstr1 + pstr2
```
Strilanc: Wait, `__add__` is inferred from `__iadd__`? Could you verify that you didn't mutate `pstr1` or `pstr2` by performing this addition?

cirq/ops/linear_combinations_test.py
```diff
@@ -562,3 +562,20 @@ def test_linear_combination_of_operations_has_correct_pauli_expansion(
...
+    assert str(psum) == "1.000*X(0)*X(1)+1.000*Y(0)*Y(1)"
+    assert list(psum) == [pstr1, pstr2]
+
+
+def test_pauli_sum_from_list_of_strings():
+    q = cirq.LineQubit.range(2)
+    pstr1 = cirq.X(q[0]) * cirq.X(q[1])
+    pstr2 = cirq.Y(q[0]) * cirq.Y(q[1])
+    psum = cirq.PauliSum.from_pauli_strings([pstr1, pstr2])
+    assert str(psum) == "1.000*X(0)*X(1)+1.000*Y(0)*Y(1)"
```
Strilanc: Need a bunch more tests. E.g. nothing is directly checking subtraction, formatting, iteration, raw initialization (though some of them happen to be hit indirectly as part of these tests).

cirq/ops/linear_combinations.py
mpharrigan: Copied this from LinearDict :)

cirq/ops/linear_combinations_test.py
mpharrigan: this is from PauliString.__add__

Strilanc: :man_facepalming:

cirq/ops/linear_combinations.py
mpharrigan: I had this sitting on a branch for a long time as inheriting from `LinearDict` so all of these got filled in. I got back to it today and ripped it out to favor composition and all the dundermethods were left behind in my haphazardness :)

mpharrigan:
Added some more things; will pick this up and add some more tests tomorrow

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +207,124 @@ def extend(expansion: value.LinearDict[str],
...
+        for qid, pauli in k:
+            if not isinstance(qid, raw_types.Qid):
+                return False
+            if not isinstance(pauli, pauli_gates.Pauli):
+                return False
+
+    return True
+
+@value.value_equality(approximate=True)
+class PauliSum:
```
kevinsung: Why not inherit from LinearCombinationOfOperations?

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +207,124 @@ def extend(expansion: value.LinearDict[str],
...
+                "PauliSum constructor takes a LinearDict[UnitPauliStringT]. "
+                "Consider using PauliSum.from_pauli_strings() or adding and "
+                "subtracting PauliStrings")
+        self._linear_dict = linear_dict
+
+    def _value_equality_values_(self):
+        return self._linear_dict
+
+    @classmethod
+    def from_pauli_strings(cls, terms: List[PauliString]) -> 'PauliSum':
```
Strilanc: Update type annotation to Union[PauliString, List[pauliString]]

cirq/ops/linear_combinations_test.py
```diff
@@ -562,3 +562,73 @@ def test_linear_combination_of_operations_has_correct_pauli_expansion(
...
+                                                     cirq.PauliString({}, 1.3)])
+
+
+def test_add_number_paulistring():
+    q = cirq.LineQubit.range(2)
+    pstr1 = cirq.X(q[0]) * cirq.X(q[1])
+    psum = pstr1 + 1.3
+    assert psum == cirq.PauliSum.from_pauli_strings([pstr1,
+                                                     cirq.PauliString({}, 1.3)])
+
```
Strilanc: Add a `__repr__` test using `cirq.testing.assert_equivalent_repr`.

cirq/ops/pauli_string.py
```diff
@@ -56,13 +58,26 @@ def from_single(qubit: raw_types.Qid,
     def coefficient(self) -> complex:
         return self._coefficient
 
+    def unit(self) -> UnitPauliStringT:
```
Strilanc: This should just return a `PauliString` with coefficient equal to 1, similar to how `abs(int)` returns an `int`.

cirq/ops/pauli_string.py
```diff
@@ -115,6 +130,24 @@ def __rmul__(self, other):
             return PauliString(self._qubit_pauli_map, self._coefficient * other)
         return NotImplemented
 
+    def __add__(self, other):
+        from cirq.ops.linear_combinations import PauliSum
```
Strilanc: This method should basically just be delegating to `PauliSum`. Just `return PauliSum.from(self).__add__(other)`.

cirq/ops/pauli_string.py
```diff
@@ -115,6 +130,24 @@ def __rmul__(self, other):
...
+        if isinstance(other, (int, float, complex)):
+            return PauliSum.from_pauli_strings(
+                [self, PauliString(coefficient=other)])
+        elif isinstance(other, PauliString):
+            return PauliSum.from_pauli_strings([self, other])
+        return other + self
+
+    def __sub__(self, other):
```
Strilanc: Delegate

cirq/ops/pauli_string.py
```diff
@@ -115,6 +130,24 @@ def __rmul__(self, other):
...
+        from cirq.ops.linear_combinations import PauliSum
```
Strilanc: Do we also need `__add__` and `__sub__` on Pauli gates X/Y/Z?

cirq/ops/linear_combinations.py
mpharrigan: see #1682 

We need the PauliStrings, -- which would otherwise serve as the key in a LinearDict -- to own their own coefficient

cirq/ops/pauli_string.py
mpharrigan: I need this for `PauliSum.from_pauli_strings` and it's used in `_value_equality_values`. Maybe we can give it a different name?

cirq/ops/pauli_string.py
mpharrigan: I added a test with single-qubit pauli gates which I expected to fail so then I could implement these, but lo and behold `SingleQubitPauliStringGateOperation` inherits from `PauliString` so it works already

cirq/ops/linear_combinations_test.py
mpharrigan: Added some more tests. Do you have any specific requests for more?

cirq/ops/linear_combinations.py
mpharrigan: done

cirq/ops/linear_combinations_test.py
mpharrigan: nifty! and done

cirq/ops/pauli_string.py
mpharrigan: done

cirq/ops/pauli_string.py
mpharrigan: done

cirq/ops/pauli_string.py
Strilanc: Given the current result type, this shouldn't be a public method. Users won't use it; it's too specific. It's an internal implementation detail.

If you need it in `from_pauli_strings`, define it in the pauli sum file instead of here. `_value_equality_values_` can be restored to its old form.

cirq/ops/linear_combinations_test.py
```diff
@@ -562,3 +562,101 @@ def test_linear_combination_of_operations_has_correct_pauli_expansion(
...
+    empty = cirq.PauliSum.from_pauli_strings([])
+    assert str(empty) == "0.000"
+
+
+def test_pauli_sum_repr():
+    q = cirq.LineQubit.range(2)
+    pstr1 = cirq.X(q[0]) * cirq.X(q[1])
+    pstr2 = cirq.Y(q[0]) * cirq.Y(q[1])
+    psum = pstr1 + pstr2
```
Strilanc: Include a `*2` on the second term, and also include a `+1` identity term, to exercise more cases.

cirq/ops/linear_combinations.py
mpharrigan: added validator

cirq/ops/linear_combinations_test.py
mpharrigan: done

cirq/ops/pauli_string.py
mpharrigan: done

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +209,132 @@ def extend(expansion: value.LinearDict[str],
...
+            -> 'PauliSum':
+        if isinstance(terms, PauliString):
+            terms = [terms]
+        termdict = defaultdict(lambda: 0)
+        for pstring in terms:
+            key = frozenset(pstring._qubit_pauli_map.items())
+            termdict[key] += pstring.coefficient
+        return cls(linear_dict=value.LinearDict(termdict))
+
+    def copy(self):
```
Strilanc: Output type annotation

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +209,132 @@ def extend(expansion: value.LinearDict[str],
...
+        if isinstance(other, (float, int, complex)):
+            other = PauliSum.from_pauli_strings(
+                [PauliString(coefficient=other)])
+        elif isinstance(other, PauliString):
+            other = PauliSum.from_pauli_strings([other])
+
+        self._linear_dict += other._linear_dict
+        return self
+
+    def __add__(self, other):
```
Strilanc: `__add__` should return `NotImplemented` for unrecognized types, and this should be tested.

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +209,132 @@ def extend(expansion: value.LinearDict[str],
...
+        result = self.copy()
+        result += other
+        return result
+
+    def __isub__(self, other):
+        if isinstance(other, PauliString):
```
Strilanc: Should return `NotImplemented` for unrecognized types, and this should be tested.

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +209,132 @@ def extend(expansion: value.LinearDict[str],
...
+            other = PauliSum.from_pauli_strings([other])
+        self._linear_dict -= other._linear_dict
+        return self
+
+    def __sub__(self, other):
+        result = self.copy()
```
Strilanc: Should return `NotImplemented` for unrecognized types, and this should be tested.

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +209,132 @@ def extend(expansion: value.LinearDict[str],
...
+        result -= other
+        return result
+
+    def __neg__(self):
+        factory = type(self)
+        return factory(-self._linear_dict)
+
+    def __imul__(self, a: value.Scalar):
+        self._linear_dict *= a
```
Strilanc: Should return `NotImplemented` for unrecognized types, and this should be tested.

Note that this means the type annotation is not correct. You can use `@overload` from typing to give the specific cases if you're feeling ambitious.

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +209,132 @@ def extend(expansion: value.LinearDict[str],
...
+        return self
+
+    def __mul__(self, a: value.Scalar):
```
Strilanc: Should return `NotImplemented` for unrecognized types, and this should be tested.

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +209,132 @@ def extend(expansion: value.LinearDict[str],
...
+        result = self.copy()
+        result *= a
+        return result
+
+    def __rmul__(self, a: value.Scalar):
+        return self.__mul__(a)
```
Strilanc: Should return `NotImplemented` for unrecognized types, and this should be tested.

cirq/ops/linear_combinations.py
```diff
@@ -205,3 +209,132 @@ def extend(expansion: value.LinearDict[str],
...
+
+    def __truediv__(self, a: value.Scalar):
```
Strilanc: Should return `NotImplemented` for unrecognized types, and this should be tested.

cirq/ops/linear_combinations.py
mpharrigan: why is the type annotation not correct?

mpharrigan: should I add the check to LinearDict, which this method dispatches to?

mpharrigan:
Added tests for all bad arithmetic operations. For multiplication and division, the straightforward dispatch to `LinearDict`'s operators means that it already throws a `TypeError` as appropriate

cirq/ops/linear_combinations.py
mpharrigan: done

cirq/ops/linear_combinations.py
mpharrigan: done

cirq/ops/linear_combinations.py
mpharrigan: done. Also fixed this so you can subtract a scalar

cirq/ops/linear_combinations.py
Strilanc: The type annotation is technically not correct because the method has a defined result when `a` is not a Scalar. Or... well, I guess I don't know the exact semantics assumed by mypy. Maybe it considereds returning NotImplemented to be an error where the type annotations ar enot supposed to apply.

mpharrigan: returning `NotImplemented` is a way for python itself to throw a `TypeError` after exhausting all possible options

Strilanc: If mypy is happy, I'm happy.

mpharrigan:
What's the timeline for requiring python 3.6? A test or two assumes dictionaries are ordered, which only happens in >py36

Strilanc:
Now that we've thought of a way to prevent old versions of python from being silently out of date when installing cirq, I think we can transition into 3.6 ASAP.

Are you aware you can run the format script with `--apply` to apply the changes instead of just viewing them?

mpharrigan:
running with `--apply` was the cause of 2a7641e. If I kept running the formatter with `--apply` it started to more aggressively mess up the formatting of the inputs.

Apparently I also have a mismatched version with travis, because the current format failure was put into place by my recent application of `--apply`

Strilanc:
The other reason it may be failing is if you are not synced with master. There's an open issue to make the incremental diff compute vs the merge base of master and the current branch instead of just vs master (which changes suddenly on travis as PRs get merged).

mpharrigan:
I've been merging in master periodically. I'm suspecting I have the wrong version of yapf

Strilanc:
Sync with master to get the fixed "check vs mergebase" format script. You then have a bunch of coverage errors to work through, and we can merge.

mpharrigan:
The tests are failing because it is assuming that `dict`s are ordered, and therefore the `PauliSum`s are ordered. If we are indeed switching to python 3.6+ soon, then it can wait. Otherwise I'll have to re-work the test and/or the code

Strilanc:
The switch has already happened. Sync with master.

mpharrigan:
Ah very good

PR #1709, opened 2019-06-11
Add qubit_map to SimulationTrialResult

mpharrigan:
xref https://github.com/quantumlib/Cirq/pull/1678#issuecomment-500966950

### DIFF
--- a/cirq/sim/simulator.py
+++ b/cirq/sim/simulator.py
@@ -534,5 +534,8 @@ class SimulationTrialResult:
         return (self.params, measurements, self._final_simulator_state)
 
     @property
-    def qubit_map(self):
+    def qubit_map(self) -> Dict[ops.Qid, int]:
+        """A map from Qid to index used to define the ordering of the basis in
+        the result.
+        """
         return self._final_simulator_state.qubit_map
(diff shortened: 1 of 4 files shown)


cirq/sim/simulator.py
```diff
@@ -532,3 +532,7 @@ def _value_equality_values_(self):
         measurements = {k: v.tolist() for k, v in
                         sorted(self.measurements.items())}
         return (self.params, measurements, self._final_simulator_state)
+
+    @property
+    def qubit_map(self):
```
Strilanc: Add a docstring explaining how this would be used, and also a return type annotation.

mpharrigan: done

mpharrigan:
ah, it would appear this breaks everything

mpharrigan:
So `StateVectorMixin` -- which `WaveFunctionTrialResult` mixes in -- tries to set `qubit_map` which resolves to `SimulationTrialResult`'s property. hmm

mpharrigan:
Ok, I may have fixed it. Please give it a look

dabacon:
Looks like CirqBot is having issues.

PR #1710, opened 2019-06-12
0.001 for Mean test fail is too sensitive. Change to 0.1

vtomole:


dabacon:
Is it possible to configure this per test?  My guess would be that some tests have different tolerances.

vtomole:
@viathor  I'm not sure, but https://travis-ci.com/quantumlib/Cirq/jobs/207496954 failed after the merge

```
Performance has regressed:
	test_example_runs_hello_line_perf (0001_dde1a77) - Field 'mean' has failed DifferenceRegressionCheck: 0.005218593 > 0.001000000
	test_example_runs_bcs_mean_field_perf (0001_dde1a77) - Field 'mean' has failed DifferenceRegressionCheck: 0.002496049 > 0.001000000
--------------------------------------------------------------------------------
ERROR: Performance has regressed.
```

Maybe 0.01 will do. It's seems like guesswork.

vtomole:
I'm going to leave this as is even though it seems conservative because other people's builds will fail with what is currently in the mainline.

viathor:
If we cannot specify this per-test as @dabacon suggested, maybe we can use percentages instead of seconds?

Also, a more rigorous way to arrive at the threshold would be to run the tests many times (20-100) and compute standard deviation of their runtimes. Then you can use [68-95-99.7 rule](https://en.wikipedia.org/wiki/68%E2%80%9395%E2%80%9399.7_rule) or something similar to decide on the threshold. WDYT?

vtomole:
I'm going to merge this so the test always passes. We can take this conversation to https://github.com/quantumlib/Cirq/issues/1124.

PR #1712, opened 2019-06-13
Speed up randomized bench-marking

### LINKED ISSUE #1476: examples test is slow
This test is taking almost 30s:
```
27.86s call     examples/examples_test.py::test_example_runs_qubit_characterizations
```

I think it is the single_qubit_randomized_benchmarking and two_qubit_randomized_benchamarking, which has default parameters that are pretty large.

c-poole:
Currently we generate the matrices for all 11520 elements of the 2 qubit Clifford group and in order to identify the inverse of a specific element we multiply it by all 11520 elements to find a candidate for the inverse. We have cirq.inverse() and should be using that instead. This shaved ~35 seconds off my pytest runs. Fixes #1476

### DIFF
--- a/cirq/experiments/qubit_characterizations.py
+++ b/cirq/experiments/qubit_characterizations.py
@@ -402,10 +402,10 @@ def two_qubit_state_tomography(sampler: sim.Sampler,
         prob_list = [results_hist[0], results_hist[1], results_hist[2]]
         return np.asarray(prob_list) / repetitions
 
-    sigma_0 = np.eye(2) / 2.0
-    sigma_1 = np.array([[0.0, 1.0], [1.0, 0.0]]) / 2.0
-    sigma_2 = np.array([[0.0, -1.0j], [1.0j, 0.0]]) / 2.0
-    sigma_3 = np.array([[1.0, 0.0], [0.0, -1.0]]) / 2.0
+    sigma_0 = np.eye(2) * 0.5
+    sigma_1 = np.array([[0.0, 1.0], [1.0, 0.0]]) * 0.5
+    sigma_2 = np.array([[0.0, -1.0j], [1.0j, 0.0]]) * 0.5
+    sigma_3 = np.array([[1.0, 0.0], [0.0, -1.0]]) * 0.5
     sigmas = [sigma_0, sigma_1, sigma_2, sigma_3]
 
     # Stores all 27 measured probabilities (P_00, P_01, P_10 after 9
--- a/cirq/experiments/qubit_characterizations.py
+++ b/cirq/experiments/qubit_characterizations.py
@@ -208,13 +208,12 @@ def single_qubit_randomized_benchmarking(
 
     cliffords = _single_qubit_cliffords()
     c1 = cliffords.c1_in_xy if use_xy_basis else cliffords.c1_in_xz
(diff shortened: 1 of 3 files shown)


c-poole:
As a point of comparison:

Test run-times for most recently merged PR vs. most recent commit on this PR as of when I wrote this comment:
(linux build) check/pytest-and-incremental-coverage 283.33s vs 238.18s
(windows build) check/pytest --benchmark-skip 244.04s vs 211.16s

cirq/experiments/qubit_characterizations.py
```diff
@@ -404,10 +402,10 @@ def _measurement(two_qubit_circuit: circuits.Circuit) -> np.ndarray:
         prob_list = [results_hist[0], results_hist[1], results_hist[2]]
         return np.asarray(prob_list) / repetitions
 
-    sigma_0 = np.eye(2) / 2.0
-    sigma_1 = np.array([[0.0, 1.0], [1.0, 0.0]]) / 2.0
-    sigma_2 = np.array([[0.0, -1.0j], [1.0j, 0.0]]) / 2.0
-    sigma_3 = np.array([[1.0, 0.0], [0.0, -1.0]]) / 2.0
+    sigma_0 = np.eye(2) * 0.5
```
dabacon: do these changes impact speed or anything?  Might move the scalar in front but that's a nit.

c-poole:
After the change in how the RB circuits are generated, it looks like the bottleneck for Rabi, RB, and Tomography is the execution/sampling of the circuits rather than circuit generation. I've reduced repetition values in tests and examples to amounts that still keep the generated plots looking fine but save another 20 seconds or so on my machine.

cirq/experiments/qubit_characterizations.py
c-poole: They technically do but practically speaking you only get gains if the np array is a reasonably large size. I made the division of constant -> multiplication by inverse changes mostly out of force of habit.

c-poole:
I think the new speeds of the tests mentioned in #1476 merit saying that this PR fixes #1476.

cirq/experiments/qubit_characterizations.py
```diff
@@ -455,57 +453,31 @@ def _indices_after_basis_rot(i: int, j: int) -> Tuple[int, Sequence[int],
...
+                           num_cfds: int,
+                           cliffords: Cliffords) -> circuits.Circuit:
     clifford_group_size = 11520
     idx_list = list(np.random.choice(clifford_group_size, num_cfds))
     circuit = circuits.Circuit()
     for idx in idx_list:
         circuit.append(_two_qubit_clifford(q_0, q_1, idx, cliffords))
-    inv_idx = _find_inv_matrix(protocols.unitary(circuit), cfd_matrices)
-    circuit.append(_two_qubit_clifford(q_0, q_1, inv_idx, cliffords))
+    circuit.append(protocols.inverse(circuit))
```
95-martin-orion: @c-poole, I have a question regarding this change. In my tests, applying `protocols.inverse` here creates an inverted circuit (all gates inverted in reverse order) rather than generating a single gate that inverts the circuit. This doubles the number of gates in the returned circuit - does this negatively affect the experiment?

95-martin-orion: Opened #2846 to address this.

PR #1715, opened 2019-06-13
[Ready for review] cache for faster Travis builds

c-poole:
This appears to shave ~2 minutes off installing contrib requirements on the linux builds. Unfortunately it doesn't appear to have an effect on the Windows build. I was unsuccessful in finding a way to make a cache option work for windows. I'm content to just merge the edit that speeds up the linux builds for now.


vtomole:
Nice! I didn't get the memo when working with caches for the performance tests.

Strilanc:
This doesn't seem to have changed the build times?

image

What exactly is it supposed to be caching?

c-poole:
It changes the build times on the Linux tests that install contrib. I tried a number of things but couldn't find a way to make simple caching work to speed up Windows. It seems like we would essentially just need to dedicate a directory to act like an artificial virtual environment for windows to speed up.

Strilanc:
With pip caching (pytest w/ incremental coverage):

image

Without caching:

image

It saves about 14 seconds on the linux-based builds. So I guess it's worth doing, though not *particularly* compelling.

.travis.yml
```diff
@@ -59,18 +60,18 @@ matrix:
...
+      - choco install python --version 3.7.3
+      - export PATH="/c/Python37:/c/Python37/Scripts:$PATH"
+      - python -m pip install --upgrade pip wheel
     install:
-    - python -m pip install -r requirements.txt
-    - python -m pip install -r cirq/contrib/contrib-requirements.txt
-    - python -m pip install -r dev_tools/conf/pip-list-dev-tools.txt
-    script: check/pytest --benchmark-skip
-    
+      - pip install -r requirements.txt
```
Strilanc: Is this change actually necessary for the caching to work? In general we want to avoid saying just `pip` instead of `python -m pip` because the latter is far more reliable. Sometimes `python` refers to a version of python different than what `pip` is targeting.

c-poole:
I can't seem to find the test that convinced me there was actually a speed gain here. I may have just misread a value and went down a rabbit hole trying to make caching speed things up. The difference you note in your post isn't really significant compared to the variations in build time even for the same .yaml file so I think it is probably best if this PR is just closed.

Strilanc:
This is such a typical interaction with CI systems.

"I did a lot of stuff, kept poking and proding getting things to work, but then I was right back where I started?"

PR #1716, opened 2019-06-13
Store float values in v2 Points proto

maffoo:
This aligns with the actual [Points](https://github.com/quantumlib/Cirq/blob/master/cirq/study/sweeps.py#L275) object, which can only have float values.

Before merging, we'd need to regenerate the `_pb2.py` files. @dabacon, are there instructions somewhere on how to do that?


dabacon:
With the proto compiler installed I believe it is just
```protoc -I=. --python_out=. cirq/api/google/v2/*.proto```

Issue #1156, opened 2018-11-14
Use protocol buffers for serialization instead of dicts


viathor:
When code generated from protocol buffer defs was removed in #825, a number of serialization/deserialization functions that used them (e.g. `sweep_to_proto()`) were changed to return/accept `dict` instead. We have recently undone the removal (see #1085).

Should also undo the serialization/deserialization change? Protocol buffers offer stronger type safety than `dict`.

maffoo:
I'd like to revisit this, because I'm currently dealing with some protobuf translation issues but want to get input before I put in too much work. I think it's nice to use the protobuf classes directly for some use cases (and we could even enable [stub generation](https://github.com/dropbox/mypy-protobuf) to get type checking of the messages). Also, there's an existing [`json_format`](https://github.com/protocolbuffers/protobuf/blob/master/python/google/protobuf/json_format.py) module that provides [`MessageToDict`](https://github.com/protocolbuffers/protobuf/blob/master/python/google/protobuf/json_format.py#L136) and [`ParseDict`](https://github.com/protocolbuffers/protobuf/blob/master/python/google/protobuf/json_format.py#L418) functions, so that we easily translate between proto messages and python dicts (or json strings) in the canonical format without writing any additional code. Thoughts?

Strilanc:
What do you see as the advantages of using the protobuf classes, instead of just dictionaries? These are serialized objects; we mostly interact with them in code that is converting into or out of the representation. Is that code noticeably difference if we use the protobuf classes instead of dicts?

I agree that outside of conversion code, when actually operating on the representation, you want more type safety than a dictionary.

maffoo:
One issue is that right now we are hand-rolling the serialization code to try to be compatible with the canonical json representation of protobuf message, but we differ in some cases. For example, protos represented as json messages can omit fields that have default values, but in `GridQubit` we currently [require](https://github.com/quantumlib/Cirq/blob/master/cirq/devices/grid_qubit.py#L152) that the `row` and `col` fields be present which does not agree with the canonical representation. That code is marked as deprecated, but it's symptomatic of trying to roll our own serialization code where we could just use the protobuf libraries.

I agree that from the perspective of a user there's almost no difference because as you say its just serialized data. But when implementing backend logic that receives values as protobuf messages, it's odd to have to convert to dicts first in order to be able to then deserialize into cirq objects. This backend logic does not even live in cirq, which is part of why I seem to be the only one running into issues around this :-)

[...]

PR #1717, opened 2019-06-14
Support converting to/from proto messages directly in v2 api.

maffoo:
Related to the discussion in #1156, this would switch the v2 (de)serializers to use proto message classes, with translation to/from json-compatible dicts using the standard `google.protobuf.json_format` package if desired. This way we can center the discussion around real code. A few notes:

- The protobuf semantics around missing vs. empty fields do not always align with what we've been doing in hand-written code. For example, there's no way to serialize an empty schedule to a protobuf and then deserialize it as an empty schedule, because the protobuf will contain neither a circuit nor schedule field, so we can't tell if this was intended to be an empty schedule or if neither circuit nor schedule was set (which should be an error). I've put in code to assume an empty schedule in this case because there was a test for serializing empty schedules, but I think it might be better to just disallow serializing empty schedules (to do anything you _at least_ need a measurement operation, so empty schedules are not very useful).

- The protos are defined with 32 bit float fields so we lose precision on some numbers. For tests we either need to account for this by checking approximate equality, or we need to use numbers that are exactly representable as 32-bit floats. I've opted for the latter for now to keep the diff smaller.

- The conversion to json format has some options like whether to include default-valued fields, whether to preserve the proto field names or convert to lowerCamelCase, and how to represent enums. The json_format parser for deserialization can handle any of these, but most of the tests were written assuming a specific json output, so I've set the options on `MessageToDict` to be closest to how the tests were written, again to keep the diff smaller.

### DIFF
--- a/cirq/google/op_deserializer.py
+++ b/cirq/google/op_deserializer.py
@@ -13,9 +13,12 @@
 # limitations under the License.
 
 from typing import Any, Callable, Dict, NamedTuple, Optional, Sequence
+
 import sympy
+from google.protobuf import json_format
 
 from cirq import devices, ops
+from cirq.api.google import v2
 from cirq.google import arg_func_langs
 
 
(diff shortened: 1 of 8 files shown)


cirq/google/op_deserializer.py
```diff
@@ -84,42 +87,48 @@ def __init__(self,
...
+                    elif which == 'string_value':
+                        value = str(arg_value.string_value)
+                elif which == 'symbol':
+                    value = sympy.Symbol(arg_proto.symbol)
 
             if value is None and arg.required:
                 raise ValueError(
                     'Could not get arg {} from arg_proto {}'.format(
-                        arg.serialized_name, args_proto_dict))
+                        arg.serialized_name, args.proto))
```
viathor: `args.proto` -> `proto.args`?
I think this is what mypy and pylint are crying about.

maffoo: Thanks; fixed.

Issue #1611, opened 2019-05-08
Non-parametrized sympy literals as numbers?


balopat:
````
 import cirq
 import sympy

 c = cirq.Circuit()
 q = cirq.NamedQubit('q')
 c.append(cirq.Rx(sympy.pi)(q))
 c.append(cirq.Ry(sympy.pi * 2 / sympy.cos(3))(q))   
 cirq.Simulator().run(c) 
 ````
 fails with `Can't simulate unknown operations that don't specify a _unitary_ method, a _decompose_ method, (_has_unitary_ + _apply_unitary_) methods,(_has_mixture_ + _mixture_) methods, or are measurements.: cirq.Rx(sympy.pi).on(cirq.NamedQubit('q'))` -- which could be led back to `if protocols.is_parameterized(rads)` returns `True ` for `sympy.Basic` objects. 

However, as sympy objects are acting most of the time as numbers - the printout of the circuit is

```
 q: ───Rx(pi)───Ry(2*pi/cos(3))───
```

Which is very nice vs the numpy version:

```
 q: ───Rx(π)───Ry(1.9800000000000004π)───
```

Where this came up: when importing QASM files (#1602), the printed circuits look significantly better when the angles read in symbolically. I did convert them to float of course but I feel like we are losing information there.  

Not sure what the implications on performance or functionality, but it looks like something worthwhile to think about as sympy expressions might be more precise/offer more functionality through symbolic manipulation.

What do you think?

c-poole:
I think we can get away with resolving this issue without too much fuss. Instead of returning `True` if we find a `sympy` object, return the `is_number` property of the object in the `is_parametarized` function.
`if isinstace(val, sympy.Basic):
    return val.is_number
`

c-poole:
I just tried my suggestion and surprisingly only one test failed. It was due to numpy complaining about not knowing how to multiply sympy constants with standard numpy data types. This can be patched over by making the appropriate numeric casts inside _apply_unitary_, and _unitary_ for each gate in the repository but that is a bit more fuss than I had originally anticipated. I just tried it with Rx and Ry and got your code snippet to run after adding a measurement to the circuit. I can take this issue if there is interest in this feature. @balopat, @Strilanc , @dabacon thoughts?

dabacon:
[...]

PR #1718, opened 2019-06-14
QASM importer: CX gate

balopat:
Adds parsing CX gate, which required the following:
- parse argument list for operation - mixed single or multi-qubit registers 
- validation of argument length
- mapping of qubits to `cirq.NamedQubit`s 
- mapping qasm operations to Cirq Operation 

This is the 2nd PR of the piecemeal reimplementation of #1602.

Next PRs planned after this: 
1. add U gate instruction with simple float parameters, parameter list validation 
2. add expression handling for parameters: functions, pi, arithmetic operators 
3. add other gates


cirq/contrib/qasm_import/_parser.py
```diff
@@ -11,11 +11,12 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import Dict, Optional
+from typing import Dict, Optional, List, Any, Iterable
 
 from ply import yacc
 
-from cirq import Circuit
+import cirq
+from cirq import Circuit, NamedQubit, CNOT
```
vtomole: `from cirq import CX` also works and it's more consitent with QASM using `cx`.

cirq/contrib/qasm_import/_parser_test.py
```diff
@@ -151,3 +152,108 @@ def test_syntax_error():
...
+
+    parsed_qasm = parser.parse(qasm)
+
+    assert parsed_qasm.supportedFormat is True
+    assert parsed_qasm.qelib1Include is False
+
+    ct.assert_same_circuits(parsed_qasm.circuit, expected_circuit)
+    assert parsed_qasm.qregs == {'q1': 2, 'q2': 2}
+
+    cirq.Simulator().simulate(parsed_qasm.circuit)
```
vtomole: Why the call to the simulator if we are not using the results from it?

cirq/contrib/qasm_import/_parser.py
```diff
@@ -35,6 +36,41 @@ def __init__(self, supported_format: bool, qelib1_include: bool,
...
+
+        for i in range(reg_size):
+            # because of mixed quantum register/qubit support there can
+            # be single qubit registers and longer qubit registers
+            # for single qubit registers qreg[0], for longer ones qreg[i]
+            qubits = [qreg[min(len(qreg) - 1, i)] for qreg in args]
+            yield self.cirq_gate.on(*qubits)
+
+
 class QasmParser(object):
```
vtomole: No need to inherit from `Object`: https://stackoverflow.com/a/1238632.

balopat:
Thank you for the review @vtomole!

cirq/contrib/qasm_import/_parser.py
balopat: sounds good!

cirq/contrib/qasm_import/_parser_test.py
balopat: I know ... it looks weird. I had this in because at one point as I was experimenting, I had a version where I returned sympy numbers instead of actual floats and even though circuit comparison succeeded - but the circuits were not "simulatable" - I wanted to guard against that, but probably this is not the right place to do so (I created an issue a while ago regarding that case https://github.com/quantumlib/Cirq/issues/1611). 
I took out this "non-assertion check" from here.

cirq/contrib/qasm_import/_parser.py
balopat: 👍

cirq/contrib/qasm_import/_parser_test.py
```diff
@@ -151,3 +154,106 @@ def test_syntax_error():
...
+def test_unknown_basic_gate():
+    qasm = """
+         OPENQASM 2.0;          
+         qreg q[2];
+         foobar q[0];
+    """
+    parser = QasmParser()
+
+    with pytest.raises(QasmException,
+                       match=r"""Unknown gate "foobar".* line 4.*forgot.*\?"""):
```
vtomole: Why does this say `line 4`? Shouldn't it be 3?

1: OPENQASM 2.0; 
2: qreg q[2];
3: foobar q[0];

cirq/contrib/qasm_import/_parser.py
```diff
@@ -35,7 +36,42 @@ def __init__(self, supported_format: bool, qelib1_include: bool,
         self.circuit = c
 
 
-class QasmParser(object):
+class QasmGate:
```
Strilanc: Don't call this a `Gate`; that has a very specific meaning in cirq that this class does not satisfy. Instead call it... maybe a statement? An invocation?

cirq/contrib/qasm_import/_parser.py
```diff
@@ -35,7 +36,42 @@ def __init__(self, supported_format: bool, qelib1_include: bool,
...
+        self.validate_args(args, lineno)
+        reg_size = 1
+        for reg in args:
+            if len(reg) > 1 and len(reg) != reg_size:
+                if reg_size == 1:
+                    reg_size = len(reg)
+                else:
+                    raise QasmException(
+                        "Non matching quantum registers of length {} "
+                        "at line {}".format([len(reg) for reg in args], lineno))
```
Strilanc: This will say line 0 when the lineno argument was not specified. Maybe it should default to something less confusing like "(not specified").

cirq/contrib/qasm_import/_parser.py
```diff
@@ -35,7 +36,42 @@ def __init__(self, supported_format: bool, qelib1_include: bool,
...
+
+        for i in range(reg_size):
+            # because of mixed quantum register/qubit support there can
+            # be single qubit registers and longer qubit registers
+            # for single qubit registers qreg[0], for longer ones qreg[i]
```
Strilanc: I don't understand what this comment is saying and how it relates to the code below.

cirq/contrib/qasm_import/_parser.py
```diff
@@ -35,7 +36,42 @@ def __init__(self, supported_format: bool, qelib1_include: bool,
...
+            qubits = [qreg[min(len(qreg) - 1, i)] for qreg in args]
```
Strilanc: Is this a correct interpretation of the input? It looks like it will replace malformed expressions like `q[999999999999999999999999999]` with `q[max_admissible_index]` instead of producing an error.

cirq/contrib/qasm_import/_parser_test.py
```diff
@@ -151,3 +154,106 @@ def test_syntax_error():
...
+                       match=(r"CX only takes 2 arg\(s\) "
+                              r"\(qubits and/or registers\)"
+                              r", got: 1, at line 4")):
+        parser.parse(qasm)
+
+
+def test_cx_gate_mismatched_registers():
+    qasm = """
+     OPENQASM 2.0;
+     qreg q1[2];
```
Strilanc: Add a test of an out-of-bounds access raising an error. Try too-large and too-small (negative).

cirq/contrib/qasm_import/_parser_test.py
balopat: there is a new line on the top. I will change that, I agree it's a bit confusing.

cirq/contrib/qasm_import/_parser_test.py
balopat: 👍

cirq/contrib/qasm_import/_parser.py
balopat: Sounds good! I like `QasmGateStatement`, will go with that.

cirq/contrib/qasm_import/_parser.py
balopat: Yeah this could be clearer. 

Maybe: 
```
# OpenQASM gates can be applied on single qubits and qubit registers. 
# We represent single qubits as registers of size 1. 
# Based on the OpenQASM spec (https://arxiv.org/abs/1707.03429), 
# single qubit arguments can be mixed with qubit registers.
# Given quantum registers of length n and single qubits are both used as arguments,
# we generate n `GateOperations` via iterating through each qubit of the registers
# 0 to n-1 and use the same one qubit from the "single-qubit registers" for each operation. 
```

Let's look at the example of two 2-qubit registers: 

```
qreg q1[2];
qreg q2[2]; 
```

We can do the following statements: 

```
CX q1, q2; 
```
Which means
```
         ┌──┐
q1_0: ────@─────
          │
q1_1: ────┼@────
          ││
q2_0: ────X┼────
           │
q2_1: ─────X────
         └──┘
```

While a single qubit mixed with a register: 
```
cx q1[0], q2;
```
Means:
```
q1_0: ───@───@───
         │   │
q2_0: ───X───┼───
             │
q2_1: ───────X───
```

balopat: At this point I might refactor this to make it more understandable though.

balopat: Also TODO for a new validation: the args shouldn't overlap in qubits ever i.e. these are invalid: 

```
CX q1[0], q1;
CX q2[1], q2[1];
CX q1, q1;
```

cirq/contrib/qasm_import/_parser.py
balopat: Because we check for it above, the registers will be either the same size or single qubits. I will replace this `min` logic with a more understandable loop. 

That validation for out of bounds is a real bug though, thank you for catching it. It should happen here at line 210 in `p_arg_bit` - currently a `q[999999999999999999999999999]` would create a `NamedQubit` with that name in a register without any issues. Will fix it.

cirq/contrib/qasm_import/_parser.py
Strilanc: I would describe this by talking about how we need to implement "broadcast semantics", which is a common thing from numpy.

cirq/contrib/qasm_import/_parser.py
balopat: Thanks for catching this. I do not expec `lineno` to be unspecfied. I will remove the default 0 value.

cirq/contrib/qasm_import/_parser.py
balopat: Interestingly cirq already has a validation here. I wonder if I should just catch errors on operation creation and throw expose them in the parser nicely so that user can see the line where it was thrown or should I write my own overlap checking logic here...

```
ValueError: Overlapping operations: (cirq.CNOT.on(cirq.NamedQubit('q1_1'), cirq.NamedQubit('q1_1')),)
```

cirq/contrib/qasm_import/_parser_test.py
vtomole: Catching the newline is a good test. Just leave a comment saying the newline is being tested too.

balopat:
Thanks for the reviews, @vtomole, @Strilanc ready for another look!

cirq/contrib/qasm_import/_parser.py
```diff
@@ -35,7 +36,56 @@ def __init__(self, supported_format: bool, qelib1_include: bool,
         self.circuit = c
 
 
-class QasmParser(object):
+class QasmGateStatement:
```
Strilanc: Could use a docstring

cirq/contrib/qasm_import/_parser.py
```diff
@@ -35,7 +36,56 @@ def __init__(self, supported_format: bool, qelib1_include: bool,
...
+
+        # OpenQASM gates can be applied on single qubits and qubit registers.
+        # We represent single qubits as registers of size 1.
+        # Based on the OpenQASM spec (https://arxiv.org/abs/1707.03429),
+        # single qubit arguments can be mixed with qubit registers.
+        # Given quantum registers of length reg_size and single qubits are both
+        # used as arguments, we generate reg_size GateOperations via iterating
+        # through each qubit of the registers 0 to n-1 and use the same one
+        # qubit from the "single-qubit registers" for each operation.
+        for i in range(reg_size):
```
Strilanc: I believe this whole loop is:

```
for qubits in functools.reduce(np.broadcast, reg):
    yield self.cirq_gate.on(*qubits)
```

But it may do the wrong thing for singleton lists.

balopat:
Thanks for the lgtm & merge - i'll address the nits separately

cirq/contrib/qasm_import/_parser.py
balopat: thank you for the suggestion - it totally works with an extra check for validation of uniqueness! Coming soon...

Issue #1093, opened 2018-10-31
Refactor PhasedXPowGate to inherit from EigenGate


Strilanc:
Note that it is necessary to override EigenGate's equality method, and methods such as `_with_exponent`, because they don't account for the phase parameter. For the most part, `PhasedXGate` will look quite a lot like `XPowGate` but with the phase parameter added.

dabacon:
I made an attempt at this and I really don't think it is worth doing.  One ends up having to override many of the EigenGate methods and the code becomes much less readable as one has to constantly flip up to EigenGate to understand it.

PR #1721, opened 2019-06-15
Fix PhasedXPowGate determinant discontinuity

### LINKED ISSUE #1719: Override of PhasedXPowGate causes determinant of gate to jump discontinuously
```
>>> np.linalg.det(cirq.unitary(cirq.PhasedXPowGate(phase_exponent=1.01, exponent=0.5)))
0.9999999999999998j
>>> np.linalg.det(cirq.unitary(cirq.PhasedXPowGate(phase_exponent=1., exponent=0.5)))
-0.9999999999999998j
>>> np.linalg.det(cirq.unitary(cirq.PhasedXPowGate(phase_exponent=0.99, exponent=0.5)))
0.9999999999999998j
```
and the gates also change badly
```
>>> cirq.unitary(cirq.PhasedXPowGate(phase_exponent=1.01, exponent=0.5))
array([[ 0.5       +0.5j       , -0.4840479 +0.51545866j],
       [-0.51545866+0.4840479j ,  0.5       +0.5j       ]])
>>> cirq.unitary(cirq.PhasedXPowGate(phase_exponent=1, exponent=0.5))
array([[0.5-0.5j, 0.5+0.5j],
       [0.5+0.5j, 0.5-0.5j]])
>>> cirq.unitary(cirq.PhasedXPowGate(phase_exponent=0.99, exponent=0.5))
array([[ 0.5       +0.5j       , -0.51545866+0.4840479j ],
       [-0.4840479 +0.51545866j,  0.5       +0.5j       ]])
```

These discontinuous jumps are likely coming from the __new__ override code.

c-poole:
Adjusts global_shift in the special cases of phase_exponent=1, phase_exponent=-0.5 to compensate for the effect of the negated exponent on global phase. This fix relies knowing that eigen_shifts of XPowGate and YPowGate are both [0, 1]. I don't anticipate these values changing, so I don't expect it to be a problem. I can make a version that doesn't rely on this but would need to call eigen_shifts to future proof if desired. Fixes #1719


c-poole:
I think removing the special casing in the constructor of PhasedXPowGate is fair. I think that rather than deleting it entirely, we may want to shove some special casing into the _decompose_ method in PhasedX or in single_qubit_matrix_to_phased_x_z. With this change (current commit as of this post) multiple tests that produce PhasedX gates that are equivalent to powers of X or Y but which don't get converted fail to produce "expected" circuits. I think we would prefer to preserve these expected circuits if possible.

c-poole:
Alternatively, we could commit to #1093 which would make the `eigen_spaces` of PhasedXPowGates continuous functions and may open the possibilities of cleaner special casing.

Strilanc:
EigenGate has the same kind of issue, where it's easier to specify some phase relationships than others. If we were going to port something over, I would choose what was done in PauliStringPhasor: just straight up separately specify the phases of the -1 and +1 eigenspaces, instead of ratios or products thereof.

mpharrigan:
A bigger question: why are these substitutions happening in `__new__` rather than e.g. in an optimization pass?

c-poole:
An optimization is typically used with the goal of reducing the cost of the circuit in some way, whereas are just special cases recognizing that an XPowGate/YPowGate with certain parameters is equivalent in any circuit to the requested PhasedXPowGate. It certainly makes circuit diagrams easier to read by special casing out things like Phased(0) with X and Phased(0.5) with Y. One of my primary concerns with just removing the special casing and changing nothing else is that our decomposition code then takes matrices like [[0, 1], [1, 0]], an obvious X gate, and outputs a Phased(1.0) gate. I intend to explore moving special casing into decomposition code as a way of resolving the currently failing decomposition tests rather than adjusting the expected output circuits.

mpharrigan:
You can imagine a case where the hardware supports a specific gate (like a PhasedXPowGate) and when you're compiling your circuit to a specific gateset for an actual device you'd like it to stay that way

c-poole:
For neutral atoms, PhasedXPowGate is precisely what we support, although the implementation of the gate is a direct rotation about the correct axis in the XY plane rather than a ZXZ sequence of rotations. When the phase angle works out so that the rotation caused by the gate is an X or Y axis rotation, we consistently report that as an X or Y rotation rather than specifying a phase angle. We don't view them as fundamentally different gates,  although others might. To us, X and Y are both special cases of the more generic PhasedX gate, and it is commonplace to simplify circuit diagrams in such cases.

Strilanc:
@mpharrigan you're probably right. Open an issue about it.

The original reasoning I had was basically this:

- If someone passes in an X gate, it should work
- If it breaks because PhasedX replaced itself with an X gate, then it wasn't going to work when I passed in an X gate and contained a bug. This is more likely to happen to a dev than the dev remembering to pass in X.
- Therefore replacing PhasedX with X will act as a bug detection mechanism during development.

Also, I wanted the diagrams to look prettier.

Strilanc:
I think the substitution in the constructor is indicative of a more general problem we have, where a consumer of a circuit wants to view it in a particular way but ends up having to deal with a lot of complexity in terms of the gates that exist. Is it a CZ or a ControlledOperation(c, Z(q)) or a ControlledGate(Z)(c, q)? This is why so many methods just fall back to "give the unitary".

Perhaps we need some sort of notion of a way of viewing gates, like `PhasedX.match(some_gate_or_operation)` returns a PhasedX if possible, and handles all the stuff you don't want to.

dabacon:
I agree with @Strilanc about wanting it to "just work".  However I also think that this is a case where "just work" can be wrong.  I like repeating myself so I will say "gate set".

Here is an example where it can go wrong:

Suppose the gates you can do are only PhasedXPowGate.  Then if you instantiate this for phase_exponent=0, it substitutes and instance of cirq.X.  But cirq.X is an XPowGate.  Now you can exponentiate that and you end up getting cirq.X**(exponent).  That's different that PhasedXPowGate raised to that exponent.  And if that's all thats in your gate set, you've gotten out of the gate set via something that if you didn't have the override would "do the right thing".  

I do like things "just working", but using __new__ to hide gate hierarchy seems dangerous (also it's definitely an "advanced python technique").

c-poole:
Based on the discussion, I have removed special casing that negates the exponent (this fixing the discontinuity) and elected to update expected circuit outputs for decompositions rather than fiddle with re-inserting special cases elsewhere. I think it would be appropriate to raise an issue to remove the `phase_exponent=0` and `phase_exponent=0.5` special cases and do that in a separate PR.

dabacon:
FWIW I  think my example is not quite correct as a problem.  Only if I had a discrete subset of the PhasedXPowGates would this be a problem (say pi and pi/2 pulses).

Issue #1657, opened 2019-05-17
_mixture_ should be able to return List[Tuple[float, Operation]] instead of List[Tuple[float, matrix]]


Strilanc:
Where it is required that the operations be unitary.

This is beneficial because it allows you to return mixtures of large operations, without having to generate their unitaries.

Fixing this mostly involves finding usages of mixture and making them account for this.

dabacon:
You mean Gate instead of Operation?

Strilanc:
Good point. A gate `_mixture_` would return gates, whereas an operation `_mixture_` would return operations. Then, in `GateOperation`, there would be a conversion.

c-poole:
Is this a duplicate of #1626? Also, I can give this a shot.

Strilanc:
Yup, it's a duplicate. Both by me. Four days apart. Wow.

PR #1722, opened 2019-06-17
Allow _mixture_ to contain gates/operations

c-poole:
Also refactor the _simulate_mixture method to make use of this fact. Closes https://github.com/quantumlib/Cirq/issues/1657


cirq/protocols/unitary.py
```diff
@@ -98,6 +99,9 @@ def unitary(val: Any,
     """
     from cirq import Gate, Operation  # HACK: Avoids circular dependencies.
 
+    if isinstance(val, np.ndarray) and linalg.is_unitary(val):
```
Strilanc: This is a pretty significant change to the semantics of cirq.unitary, not mentioned in the PR description.

I am very unsure about the "is unitary" check, because it is cubically more expensive than just reading the array. Also, it is technically stricter than what is allowed to come out of other things. Like, calling `cirq.unitary` on a malformed gate would cause no error, but `cirq.unitary(cirq.unitary(bad_gate))` would...?

I do like the idea of "How do I make a unitary thing?" being answered by "Just give the matrix as a numpy array and it will work".

cirq/ops/gate_operation.py
```diff
@@ -121,7 +121,12 @@ def _has_mixture_(self) -> bool:
         return protocols.has_mixture(self.gate)
 
     def _mixture_(self) -> Sequence[Tuple[float, Any]]:
-        return protocols.mixture(self.gate, NotImplemented)
+        mixture = protocols.mixture(self.gate, NotImplemented)
+        return ([(
+            p,
+            g(*self.qubits),
+        ) for p, g in mixture]
+                if mixture is not NotImplemented else NotImplemented)
```
Strilanc: Split the `if NotImplemented return NotImplemented` condition out into a statement, so that yapf doesn't do such a terrible job here.

cirq/ops/common_channels.py
```diff
@@ -146,7 +144,7 @@ def __init__(self, p) -> None:
         self._p = p
         self._delegate = AsymmetricDepolarizingChannel(p / 3, p / 3, p / 3)
 
-    def _mixture_(self) -> Sequence[Tuple[float, np.ndarray]]:
+    def _mixture_(self) -> Sequence[Tuple[float, raw_types.Gate]]:
```
Strilanc: We should re-order the tuple to be `[object, probability]` instead of `[probability, object]`, so it is compatible with `{A: p1, B: p2}.entries()`. You would never make a dictionary mapping probabilities to items (because duplicates), but vice versa is conceivable.

cirq/protocols/unitary.py
c-poole: Fair enough, I should have mentioned it in the PR description. I ran into the need to introduce this when I was modifying the mixture simulation to use apply_unitary methods for the gates/operations that were returned. My understanding was that a user might still end up defining a channel for themselves that returned `ndarray`s in the mixture. But calling `cirq.unitary` on a `ndarray` throws an error right now. I therefore added a check to see if the val I received IS a unitary and returned it back to the user in that case. The `is_unitary` check only gets executed if `val` is an `ndarray` because the `and` statement will short-circuit. If the user provides a malformed (non-unitary) matrix then `is_unitary` check returns `False` and the user would get the same error a user currently gets for passing in an `ndarray`. Admittedly that error message isn't appropriate and should instead be something along the lines of "The array you gave me isn't a unitary".

c-poole:
It isn't clear to me what you mean by "native" support.  `cirq.mixture` currently returns `Sequence[Tuple[float, Any]]`, so it "works" if you give it `Tuple`s with `Gate`s or `Operation`s. My edits adjust any `Gate` or `Operation` in the library that specifies mixture to return the appropriate `Tuple`, but it doesn't necessarily enforce that someone else can't. The language in the issue made it seem like returning `Gate`s or `Operation`s would be something you _could_ do, but not necessarily have to do. Your third and fourth paragraphs suggest "have to". Is this correct? If so I can try to implement what you've suggested. If not, It isn't clear to me why I haven't satisfied this condition.

> We need to modify `cirq.apply_channel` to take proper advantage of mixtures returning operations. The unitary matrix should never come into existence.

`apply_channel` wasn't touched because it doesn't attempt to call `_mixture_`. I also see from my edits in the `sparse_simulator` that I'm not actually making use of this potential speedup and so that definitely needs to be improved. I can make `cirq.apply_channel` fall back to `_mixture_` to avoid constructing unitaries if that works for you.

Strilanc:
> cirq.mixture currently returns Sequence[Tuple[float, Any]], so it "works" if you give it Tuples with Gates or Operations. My edits adjust any Gate or Operation in the library that specifies mixture to return the appropriate Tuple, but it doesn't necessarily enforce that someone else can't.

Ah I see. Yes, I was thinking in terms of having to return gates or operations. But I agree that being able to return just the unitary would be better (for gates), and that we may want to have mixtures of other things in the future. I am a bit worried about `has_mixture` returning True when the object is not a mixture of unitary operations or unitary gates; we need to handle this case correctly in `apply_mixture`, `channel`, etc.

> apply_channel wasn't touched because it doesn't attempt to call _mixture_. I also see from my edits in the sparse_simulator that I'm not actually making use of this potential speedup and so that definitely needs to be improved. I can make cirq.apply_channel fall back to _mixture_ to avoid constructing unitaries if that works for you.

Yes, I consider that to be key to closing the referenced issue. Otherwise I'd downgrade the "fixes" in the main message to "part of".

c-poole:
>  I am a bit worried about `has_mixture` returning True when the object is not a mixture of unitary operations or unitary gates; we need to handle this case correctly in `apply_mixture`, `channel`, etc.

We also have `has_mixture_channel` and `mixture_channel` which doesn't appear to be used anywhere except in tests. Perhaps we could use this or rename this to `has_mixture_of_unitaries` to special case the concept of "Mixture of unitaries" that is useful for simulation from the general case of "mixtures of objects"?

> Yes, I consider that to be key to closing the referenced issue. Otherwise I'd downgrade the "fixes"/"closes" in the main message to "part of".

I will work on making `apply_channel` utilize `mixture` as a special case then.

Strilanc:
Ah! `mixture_channel` *is* what you're calling `mixture_of_unitaries`.

I'll have to go through all this stuff trying to clean it up...

cirq/protocols/channel.py
```diff
@@ -130,7 +131,9 @@ def channel(val: Any,
     mixture_result = (
         NotImplemented if mixture_getter is None else mixture_getter())
     if mixture_result is not NotImplemented:
-        return tuple(np.sqrt(p) * u for p, u in mixture_result)
+        print("x")
```
Strilanc: Debug code present

c-poole: removed

c-poole:
I've modified `apply_channel` and the `SparseSimulator` to take advantage of the ability to return mixtures of Gates or Operations. For the `SparseSimulator` I directly changed the method that handles the mixture case and for`apply_channel` I added a strategy that attempts to utilize `_mixture_`. I think I have therefore addressed your concerns @Strilanc .

dabacon:
@Strilanc I believe this is waiting for a review from you?

dabacon:
@Strilanc  This looks like it has gone stale.  Especially given move to `_act_on_` and the deprecation of `mixture_channel` can this be resurrected or is it no longer needed?

Strilanc:
Yeah it's gone stale. Still a good idea... I'll try to harvest some of it and make a separate PR.

PR #1724, opened 2019-06-18
Remove pytest as dependency from asynchronous.py

### LINKED ISSUE #1723: cirq/testing/asychronous accidentally introduced a dependency on pytest
@cduck noticed this.

Pytest is only supposed to be a dev dependency, but `asynchronous.py` imports it in order to implement one of the helper testing methods. We should rewrite the code in question.

Also, we should have a trivial continuous integration check that does nothing except try to import cirq and print a circuit (similar to what the package verification script does). This could be snuck into one of the existing ones, just before installing test dependencies.

maffoo:
Fixes #1723

### DIFF
--- a/cirq/testing/asynchronous.py
+++ b/cirq/testing/asynchronous.py
@@ -13,10 +13,9 @@
 # limitations under the License.
 
 import asyncio
+import re
 from collections import Awaitable, Coroutine
 from typing import Any, Type, Union
-
-import pytest
 
 # A placeholder default value used to detect that callers did not specify an
 # 'expected' value argument in `assert_asyncio_will_have_result`, and so the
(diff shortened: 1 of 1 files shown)


maffoo:
Thanks @vtomole; I'm going to merge this so I can unblock building dev packages but we can revisit adding such a check in a future PR.

cirq/testing/asynchronous.py
```diff
@@ -106,7 +105,11 @@ def assert_asyncio_will_raise(
...
     try:
-        with pytest.raises(expected, match=match):
-            _run_loop_waiting_for(future, timeout)
+        _run_loop_waiting_for(future, timeout)
+    except expected as exc:
+        if match and not re.search(match, str(exc)):
+            assert False, "Pattern '{}' not found in '{}'".format(match, exc)
     except asyncio.TimeoutError:
         assert False, "Not done: {!r}".format(future)
+    else:
```
Strilanc: This works?!

maffoo: Yup :-) Can also have `else:` clause on for-loops, which will execute if the loop completes without `break`ing (that one is a bit obscure, but occasionally very useful).

PR #1725, opened 2019-06-18
Add script to generate protobuf files, including type stubs

maffoo:
This uses the grpcio-tools package, which bundles the protoc compiler. We also add the [mypy-protobuf package](https://github.com/dropbox/mypy-protobuf), which generates mypy stubs for the protobuf code, so that we get type checking of the code to build proto messages. The latter requires also updating the mypy version to handle the generated stubs.


dev_tools/build-protos.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 The Cirq Developers
```
dabacon: nit 2019

maffoo: Good catch :-) Fixed.

cirq/google/op_deserializer.py
```diff
@@ -115,12 +115,12 @@ def _args_from_proto(self, proto: v2.program_pb2.Operation
                 which = arg_proto.WhichOneof('arg')
                 if which == 'arg_value':
                     arg_value = arg_proto.arg_value
-                    which = arg_value.WhichOneof('arg_value')
-                    if which == 'float_value':
+                    which_val = arg_value.WhichOneof('arg_value')
```
Strilanc: Oh man... nice defensive fix

vtomole:
@Strilanc 

```
(base) vtomole@vtomole:~/Cirq$ dev_tools/build-protos.sh 
Writing mypy to cirq/api/google/v1/operations_pb2.pyi
Writing mypy to cirq/api/google/v1/params_pb2.pyi
Writing mypy to cirq/api/google/v1/program_pb2.pyi
Writing mypy to cirq/api/google/v2/metrics_pb2.pyi
Writing mypy to cirq/api/google/v2/program_pb2.pyi
Writing mypy to cirq/api/google/v2/result_pb2.pyi
Writing mypy to cirq/api/google/v2/run_context_pb2.pyi
(base) vagrant@vagrant:~/Cirq$ code cirq/api/google/v1/operations_pb2.pyi
```

cirq/google/op_deserializer.py
maffoo: I can't take credit; mypy caught the issue because the WhichOneof return values are typed as unions of literals, so it could tell that `which` was getting reassigned to a different type :-)

PR #1727, opened 2019-06-18
Add flexibility to serializers for cases where not all gates of given type can be serialized.

dabacon:
As an example, you might only be able to serialize cirq.XPowGate when it's exponent is 1 or 0.5 (pi or pi/2 pulses).


cirq/google/op_serializer.py
```diff
@@ -66,31 +66,44 @@ class GateOpSerializer:
...
     """
 
-    def __init__(self, *, gate_type: Type[Gate], serialized_gate_id: str,
-                 args: List[SerializingArg]):
+    def __init__(self,
+                 *,
+                 gate_type: Type[Gate],
+                 serialized_gate_id: str,
+                 args: List[SerializingArg],
+                 gate_predicate: Callable[[ops.Gate], bool] = None):
```
Strilanc: if you default to `= lambda _: True`, you wont need to deal with the `None` case.

cirq/google/op_serializer.py
```diff
@@ -66,31 +66,44 @@ class GateOpSerializer:
...
                 gate of the given gate type.
         """
         self.gate_type = gate_type
         self.serialized_gate_id = serialized_gate_id
         self.args = args
+        self.gate_predicate = gate_predicate
 
-    def to_proto_dict(self, op: ops.GateOperation) -> Dict:
-        return json_format.MessageToDict(self.to_proto(op),
+    def to_proto_dict(self, op: ops.GateOperation) -> Optional[Dict]:
```
Strilanc: I would rename the method to make it clearer that it is not guaranteed to return a serialization.

cirq/google/op_serializer.py
```diff
@@ -66,31 +66,44 @@ class GateOpSerializer:
...
+                 gate_type: Type[Gate],
+                 serialized_gate_id: str,
+                 args: List[SerializingArg],
+                 gate_predicate: Callable[[ops.Gate], bool] = None):
         """Construct the serializer.
 
         Args:
             gate_type: The type of the gate that is being serialized.
             serialized_gate_id: The string id of the gate when serialized.
+            gate_predicate: Sometimes a gate can only be serialized for
```
mpharrigan: Maybe a more descriptive name than `gate_predicate`? `can_serialize_predicate`? This is a `GateOpSerializer` so presumably everything has to do with gates

cirq/google/serializable_gate_set.py
```diff
@@ -13,14 +13,21 @@
...
+from typing import (cast, Dict, Iterable, Optional, Tuple, Type, TYPE_CHECKING,
+                    Union)
 
 from google.protobuf import json_format
 
 from cirq import circuits, devices, ops, schedules, value
 from cirq.api.google import v2
 from cirq.google import op_deserializer, op_serializer
 
+if TYPE_CHECKING:
```
mpharrigan: do we still have to do this in python 3?

cirq/google/op_serializer.py
dabacon: Not a fan of "maybe" methods, especially now with type declarations.  Suggestion?

cirq/google/serializable_gate_set.py
dabacon: nope, fixed.

cirq/google/op_serializer.py
dabacon: done

cirq/google/op_serializer.py
dabacon: done

dabacon:
I don't have better names for these methods, and think that the optional in the type signature is good enough for now.

PR #1732, opened 2019-06-19
Disable windows defender in travis config

dabacon:


dabacon:
Windows test took 6 min and 52 sec with defender disabled.

Previous tests took 16 min 22 sec, 14 min 41 sec, 15 min 9 sec, 15 min 30 sec.  So this looks like a significant savings.  

Now turning off a few other things which folks have suggested help.

Strilanc:
But Dave, what if someone hacks into the temporary virtual machine that lasts for all of 10 minutes? :roll_eyes:

Urrrggh, I can't believe this is why it was so slow.

vtomole:
Awesome! Wouldn't have thought of this.

PR #1733, opened 2019-06-19
[test] understand pip cache

dabacon:


dabacon:
So I was able to get pip caching working for windows....but it looks like a net zero.  It does speed up the pip install steps, but the time spent downloading from cache and uploading after done zeros this out.  For the full pytest coverage it actually seems to make things slower!  It seem that travis's cache isn't very "local" or something...it takes way more time to populate the cache than I would expect.

PR #1734, opened 2019-06-20
Add functions to convert Sweeps to/from v2 protos

### LINKED ISSUE #356: Cirq allows more zip and product that Engine does
I think we should probably make Engine more general.

maffoo:
Fixes #356

Not sure about the best location for this. For now I've put it in the v2 directory alongside the proto files, but happy to move if people have a better suggestion. I'll be adding some more utility functions such as dealing with the result format in the future.


maffoo:
Spoke to @Strilanc and he would prefer to have this in `cirq/google/` which is also where the other various (de)serializer classes live. Still not sure on the exact structure because I think it would be good to group all the v2 classes together somehow in anticipation of someday having v3, but I'll try something. Thoughts, @dabacon?

cirq/api/google/v2/util_test.py
```diff
@@ -0,0 +1,17 @@
...
+import cirq.api.google.v2.util as util
+
+
+@pytest.mark.parametrize('sweep', [
+    cirq.UnitSweep,
+    cirq.Linspace('a', 0, 10, 100),
+    cirq.Points('b', [1, 1.5, 2, 2.5, 3]),
+    cirq.Linspace('a', 0, 1, 5) * cirq.Linspace('b', 0, 1, 5),
+    cirq.Points('a', [1, 2, 3]) + cirq.Linspace('b', 0, 1, 3),
+    ])
```
dabacon: maybe add a case which tests the new cases we are covering?

dabacon:
So far proto type stuff has been going into cirq/google (besides the proto).  Maybe we should create a google/v2 and put that stuff there?  Sticking it with the protos is a bit strange, and may cause some problems upstream where keeping artifacts with the same build type in the same area is a best practice (i.e. putting python with proto defs is a pain often for people who want to write build rules that apply same rule to a full directory).  @Strilanc thoughts?

cirq/api/google/v2/util.py
```diff
@@ -0,0 +1,65 @@
...
+    elif isinstance(sweep, sweeps.Points):
+        msg.single_sweep.parameter_key = sweep.key
+        for point in sweep.points:
+            msg.single_sweep.points.points.append(point)
+    else:
+        raise ValueError(f'cannot convert to v2.Sweep proto: {sweep}')
+    return msg
+
+
+def sweep_from_proto(msg: run_context_pb2.Sweep) -> sweeps.Sweep:
```
Strilanc: Add a quick docstring

cirq/api/google/v2/util.py
```diff
@@ -0,0 +1,65 @@
+from typing import Optional
+
+from cirq.api.google.v2 import run_context_pb2
+from cirq.study import sweeps
+
+SweepFunction = run_context_pb2.SweepFunction
+
+
+def sweep_to_proto(
```
Strilanc: Add a quick docstring

cirq/api/google/v2/util_test.py
```diff
@@ -0,0 +1,17 @@
...
+def test_sweep_to_proto(sweep):
+    msg = util.sweep_to_proto(sweep)
+    deserialized = util.sweep_from_proto(msg)
+    assert deserialized == sweep
```
Strilanc: Add at least one test that verifies an intermediate form is actually a protobuf with the right content. I've seen these kind of round-trip-only tests let nasty bugs through in the past.

Strilanc:
Anywhere inside `cirq.google` is fine with me. Don't need to export to top level.

cirq/api/google/v2/util.py
maffoo: Added.

cirq/api/google/v2/util.py
maffoo: Added.

cirq/api/google/v2/util_test.py
maffoo: done.

PR #1737, opened 2019-06-21
[QASM import] U gate and simple parameter expressions

balopat:
Adds parsing U gate, which required the following:
- parse parameter list for operation - `pi`, natural and real numbers
- validation of parameter length
- parameter transformation - Cirq's `QasmUGate` takes the angles in half-turns and in different order

This is the 4th PR of the piecemeal reimplementation of #1602.

Next PRs planned after this: 
1. add expression handling for parameters: functions, arithmetic operators 
2. add other gates


cirq/contrib/qasm_import/_lexer_test.py
```diff
@@ -59,6 +61,36 @@ def test_valid_ids(identifier: str):
...
+    assert token.type == "NUMBER"
+    assert token.value == Number(number)
+
+
+def test_pi():
+    lexer = QasmLexer()
+    lexer.input('pi')
+    token = lexer.token()
+    assert token.type == "PI"
+    assert token.value == sympy.pi
```
Strilanc: Why is this being parsed as a symbolic constant instead of just a float?

cirq/contrib/qasm_import/_parser.py
```diff
@@ -111,7 +134,21 @@ def __init__(self):
...
+        QasmGateStatement(qasm_gate='CX',
+                          cirq_gate=CX,
+                          num_params=0,
+                          num_args=2),
+        'U':
+        QasmGateStatement(
+            qasm_gate='U',
+            num_params=3,
+            num_args=1,
+            # QasmUGate expects half turns and
```
Strilanc: Modify QasmUGate to have a gate order that matches the QASM order

cirq/contrib/qasm_import/_lexer_test.py
balopat: Originally I parsed parameters with symbolic expressions so that circuits can be more human readable when printed out. However, due to #1611 I had to convert them back to float at the very end (when applying parameters for the gate). I we see no chance for symbolic mode for circuits then I'm happy to switch over to floats.

PR #1741, opened 2019-06-24
Fix some shellcheck warnings in 'get to repo root' script code

Strilanc:
@fvkg mentioned this tool [shellcheck](https://www.shellcheck.net/). It produces a lot of useful warnings about our scripts, mostly related to undesired word splitting if strings happen to contain spaces.


fvkg:
This is a very ambitious crossover episode.

PR #1743, opened 2019-06-24
Add PauliString._unitary_ and cirq.GlobalPhaseOperation

Strilanc:
Found it was missing when trying to use PauliString as an Operation in unit tests for #1612.

`GlobalPhaseOperation` is a bit odd, because it applies to zero qubits. But it seems like a generally useful tool for making decompositions match the global phase of the original gate. We may need to update optimizers to fold them.


cirq/linalg/combinators.py
```diff
@@ -20,7 +20,7 @@
 import numpy as np
 
 
-def kron(*matrices: np.ndarray) -> np.ndarray:
+def kron(*matrices: Union[np.ndarray, complex, float]) -> np.ndarray:
```
dabacon: update args doc since i wouldn't guess a complex or float would be a "matrix"

cirq/ops/pauli_string.py
```diff
@@ -187,6 +190,25 @@ def __str__(self):
 
         return prefix + '*'.join(factors)
 
+    def _has_unitary_(self) -> bool:
```
dabacon: aside: does this point out we need another protocol for general objects that have a matrix but are not unitary

cirq/ops/global_phase_op.py
```diff
@@ -0,0 +1,64 @@
...
+            raise ValueError('Coefficient is not unitary: {!r}'.format(
+                coefficient))
+        self.coefficient = coefficient
+
+    @property
+    def qubits(self):
+        return ()
+
+    def with_qubits(self, *new_qubits):
+        assert not new_qubits
```
dabacon: add error message on assert

cirq/ops/pauli_string.py
Strilanc: Possibly. We could certainly add a `matrix` method to this class. In general I would like to have the protocols be driven by "I need that to be available to do X" rather than "I have a Y available and could provide it generically".

Strilanc:
> Why not include the global phase in the circuit and not as an operation?

Because the decompose method does not have access to the circuit, it can only return operations.

There have been enough occasions where I had the thought "Gee a global phase operation sure would be convenient for this" that I think it's worth including.

cirq/linalg/combinators.py
Strilanc: Done. Renamed to `factors` and updated the arg docs to mention scalars.

cirq/ops/global_phase_op.py
Strilanc: Done.

mpharrigan:
How do you want to draw this in the ascii diagrams? AFAICT it's currently omitted. It would be nice to show

Strilanc:
Perhaps a row of numbers below the circuit, included only if global phasing is actually present?

```
a:            ---@---
                 |
b:            ---X---

global phase:    0.5pi
```

mpharrigan:
That would work for me!

dabacon:
SGTM

Strilanc:
Added diagram notation

cirq/circuits/circuit.py
```diff
@@ -1710,11 +1717,34 @@ def _draw_moment_in_diagram(
         if x > max_x:
             max_x = x
 
+    global_phase = np.product([
```
dabacon: this now loops twice over all ops.  Why not just keep a running product in the first loop?

Strilanc: That loop is already extremely complicated, and I did not want to make it more so.

My rule of thumb is that when producing diagrams you are not in a time sensitive context, and so don't need to worry about micro-optimizing anything.

cirq/ops/global_phase_op_test.py
```diff
@@ -0,0 +1,50 @@
...
+
+    np.testing.assert_allclose(cirq.unitary(cirq.GlobalPhaseOperation(1j)),
+                               np.array([[1j]]),
+                               atol=1e-8)
+
+
+def test_diagram():
+    cirq.testing.assert_has_diagram(
+        cirq.Circuit.from_ops(cirq.X(cirq.LineQubit(0)),
+                              cirq.GlobalPhaseOperation(-1)), '0: ───X───')
```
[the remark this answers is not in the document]
Strilanc: I think this comment is stale? They are now in the diagrams.

cirq/circuits/circuit.py
```diff
@@ -1456,8 +1457,15 @@ def to_text_diagram_drawer(
         if qubit_namer is None:
             qubit_namer = lambda q: str(q) + ('' if transpose else ': ')
         diagram = TextDiagramDrawer()
+        diagram.write(0, 0, '')
         for q, i in qubit_map.items():
             diagram.write(0, i, qubit_namer(q))
+        if any(
+                isinstance(op, cirq.GlobalPhaseOperation)
```
mpharrigan: this formatting is ... interesting :)

cirq/ops/global_phase_op_test.py
```diff
@@ -0,0 +1,184 @@
...
+                ┌──┐
+0: ──────────────@─────
+                 │
+1: ──────────────┼@────
+                 ││
+10: ─────────────X┼────
+                  │
+11: ──────────────X────
+
+global phase:
```
mpharrigan: this empty line seems a little off. I understand why though: there are `GlobalPhaseOperation` so you need to have a line for `global phase` but you don't want to print a `1` for every moment that doesn't have a non-unity global phase

cirq/circuits/circuit.py
dabacon: If a loop is too complicated you should make the inner loop more readable by using other methods.

Issue #1689, opened 2019-05-30
Replace TrialResult data and query methods with a '.data' pandas.DataFrame attribute


Strilanc:
Pros:

- Less work for us.
- Easier to learn for users / user knowledge generalizes beyond cirq.
- The pandas API is already way better than ours w.r.t. handling data.
- run_sweep and run can have the same output type.

Cons:

- Less control over the format. E.g. we can't add sympy symbol support to the filtering.
- There's no longer metadata like "number of repetitions" associated with the result object. No natural place to put information such as "this is the random seed that was used".

@dabacon @mpharrigan @maffoo @babbush

maffoo:
How about if `TrialResult` "has a" dataframe rather than "is a" dataframe? Then we have a place to tack on extra metadata if needed or add some helper methods to do things like translate sympy queries into the data frame. Can also easily just get the data frame itself and discard the rest of the trial result if that's all you need.

Strilanc:
I think we should not add any of our own filtering or aggregation methods, because they will behave differently than the pandas ones and this will create opportunities for [mode errors](https://en.wikipedia.org/wiki/Mode_(user_interface)#Mode_errors).

I would prefer to just return the dataframe because it's one less thing to teach ("don't forget to put `.data`), but I agree that we may have to return a compound object that has a dataframe in addition to metadata.

viathor:
FWIW, complex numbers are supported in pandas:

```
In [1]: import numpy as np, pandas as pd

In [2]: pd.DataFrame(np.random.randn(2, 2) + 1j*np.random.randn(2, 2))
Out[2]: 
                                           0                                           1
0  (0.8085296905025485+0.21843211625055653j)    (-2.3254358862574604-0.676186251066216j)
1  (1.1482814917465722-0.13174530925455633j)  (-1.9747558058497554-0.34006501400069955j)
```

PR #1744, opened 2019-06-24
Add helpers for handling v2 result proto format

maffoo:


cirq/google/api/v2/results.py
```diff
@@ -0,0 +1,174 @@
...
+from cirq import study
+
+
+class Measurement:
+    """Info about a single measurement within a circuit or schedule."""
+
+    def __init__(self, key: str, qubits: List[devices.GridQubit], slot: int):
+        self.key = key
+        self.qubits = qubits
+        self.slot = slot
```
dabacon: suggest slot -> moment_index

cirq/google/api/v2/results.py
```diff
@@ -0,0 +1,174 @@
...
+
+from cirq.api.google.v2 import result_pb2
+from cirq import circuits
+from cirq import devices
+from cirq import ops
+from cirq import schedules
+from cirq import study
+
+
+class Measurement:
```
dabacon: make a named tupled?

dabacon: slight worry about the name....it will be only in Google namespace, but this could be confusing when people find "Measurement" here they might wonder how they use it in cirq core.

maffoo: done.

maffoo: I have the same hesitation, but couldn't think of a better name. Any suggestions?

Note that we could get rid of this entirely if `TrialResult` included the ids of the qubits that it applies to, which is probably a good idea anyway to make it more self-describing. I hope we can do that as part of the rework in #1689.

maffoo: Went with `MeasureInfo` for now, but happy to change that if we come up with a better name.

cirq/google/api/v2/results.py
maffoo: This is meant to be a more abstract notion of the location of the measurement in the program; it's moment index for circuits, but actual time for schedules. I've added a docstring to clarify. This is something we use internally when scheduling on hardware because we group together different measure ops that occupy the same "slot".

cirq/google/api/v2/results.py
```diff
@@ -0,0 +1,184 @@
...
+def _grid_qubits(op: ops.Operation) -> List[devices.GridQubit]:
+    if not all(isinstance(q, devices.GridQubit) for q in op.qubits):
+        raise ValueError('Expected GridQubits: {}'.format(op.qubits))
+    return cast(List[devices.GridQubit], list(op.qubits))
+
+
+def pack_bits(bits: np.ndarray) -> bytes:
+    """Pack bits given as a numpy array of bools into bytes."""
+    # Pad length to multiple of 8 if needed.
+    remainder = len(bits) % 8
```
Strilanc:
```
pad = -len(bits) % 8
if pad:
    bits = np.pad(bits, (0, pad), 'constant')
```

cirq/google/api/v2/results.py
```diff
@@ -0,0 +1,184 @@
...
+    """Unpack bits from a byte array into numpy array of bools."""
+    byte_arr = np.frombuffer(data, dtype='uint8').reshape((len(data), 1))
+    bits = np.unpackbits(byte_arr, axis=1)[:, ::-1].reshape(-1).astype(bool)
+    return bits[:repetitions]
+
+
+def results_to_proto(
+        trial_sweeps: Iterable[Iterable[study.TrialResult]],
+        measurements: List[MeasureInfo],
+        msg: Optional[result_pb2.Result] = None,
```
Strilanc: Nit: I would rename this to `out` and force keyword args on it, which is how numpy does it.

cirq/google/api/v2/results_test.py
```diff
@@ -0,0 +1,224 @@
...
+def test_pack_bits(reps):
+    data = np.random.randint(2, size=reps, dtype=bool)
+    packed = v2.pack_bits(data)
+    assert isinstance(packed, bytes)
+    assert len(packed) == (reps + 7) // 8
+    unpacked = v2.unpack_bits(packed, reps)
+    np.testing.assert_array_equal(unpacked, data)
+
+
+def q(row: int, col: int) -> cirq.GridQubit:
```
Strilanc: `q = cirq.GridQubit`

cirq/google/api/v2/results_test.py
```diff
@@ -0,0 +1,224 @@
...
+            [1, 0, 0],
+            [1, 0, 1],
+            [1, 1, 0],
+            [1, 1, 1],
+        ],
+                 dtype=bool))
+
+
+def test_results_from_proto_duplicate_qubit():
+    measurements = [v2.MeasureInfo('foo', [q(0, 0), q(0, 1), q(0, 2)], slot=0)]
```
Strilanc: You never pass anything except 0 into the `row` argument, so you can't have tested that part of the serialization.

cirq/google/api/v2/results.py
maffoo: Good idea. Done.

cirq/google/api/v2/results_test.py
maffoo: done.

cirq/google/api/v2/results.py
maffoo: nice trick. Done.

cirq/google/api/v2/results_test.py
maffoo: v2 serializes qubits as string qubit ids, so there's not a separate `row` field anymore.

maffoo: I went ahead and changed the tests to use some qubits on different rows. It's more realistic in any case :-)

PR #1745, opened 2019-06-25
[QASM import] measurement

balopat:
Adds measurement statements of qubits and quantum registers. 

This is the 5th PR of the piecemeal reimplementation of #1602.

### DIFF
--- a/cirq/contrib/qasm_import/_parser.py
+++ b/cirq/contrib/qasm_import/_parser.py
@@ -186,14 +186,16 @@ class QasmParser:
 
     # circuit : new_reg circuit
     #         | gate_op circuit
+    #         | measurement circuit
     #         | empty
 
     def p_circuit_reg(self, p):
         """circuit : new_reg circuit"""
         p[0] = self.circuit
 
-    def p_circuit_gate(self, p):
-        """circuit : gate_op circuit"""
+    def p_circuit_gate_or_measurement(self, p):
+        """circuit : gate_op circuit
+                   | measurement circuit"""
         self.circuit.insert(0, p[1])
         p[0] = self.circuit
 
@@ -309,6 +311,12 @@ class QasmParser:
                     self.qubits[arg_name] = NamedQubit(arg_name)
                 qubits.append(self.qubits[arg_name])
             p[0] = qubits
+        elif reg in self.cregs.keys():
+            keys = []
+            for num in range(self.cregs[reg]):
+                arg_name = self.make_name(num, reg)
+                keys.append(arg_name)
+            p[0] = keys
         else:
             raise QasmException('Undefined quantum/classical register "{}" '
                                 'at line {}'.format(reg, p.lineno(1)))
@@ -331,9 +339,29 @@ class QasmParser:
             if arg_name not in self.qubits.keys():
                 self.qubits[arg_name] = NamedQubit(arg_name)
             p[0] = [self.qubits[arg_name]]
+        elif reg in self.cregs.keys():
+            p[0] = [arg_name]
         else:
             raise QasmException('Undefined quantum register "{}" '
                                 'at line {}'.format(reg, p.lineno(1)))
+
+    # measurement operations
+    # measurement : MEASURE arg ARROW arg
+    def p_measurement(self, p):
+        """measurement : MEASURE arg ARROW arg ';'"""
+        qreg = p[2]
+        creg = p[4]
+
+        if len(qreg) != len(creg):
+            raise QasmException(
+                'mismatched register sizes {} -> {} for measurement '
+                'at line {}'.format(len(qreg), len(creg), p.lineno(1)))
+
+        measurements = []
+        for i in range(len(qreg)):
+            measurements.append(
+                cirq.MeasurementGate(num_qubits=1, key=creg[i]).on(qreg[i]))
+        p[0] = measurements
 
     def p_error(self, p):
         if p is None:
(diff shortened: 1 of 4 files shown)


cirq/contrib/qasm_import/_parser_test.py
```diff
@@ -360,3 +360,85 @@ def test_undefined_register_from_register_arg():
...
+
+    parsed_qasm = parser.parse(qasm)
+
+    assert parsed_qasm.supportedFormat is True
+    assert parsed_qasm.qelib1Include is True
+
+    ct.assert_same_circuits(parsed_qasm.circuit, expected_circuit)
+    assert parsed_qasm.qregs == {'q1': 2}
+    assert parsed_qasm.cregs == {'c1': 2}
+    cirq.Simulator().run(parsed_qasm.circuit)
```
vtomole: Why the call to `.run`?

cirq/contrib/qasm_import/_parser_test.py
```diff
@@ -360,3 +360,85 @@ def test_undefined_register_from_register_arg():
...
+
+    parsed_qasm = parser.parse(qasm)
+
+    assert parsed_qasm.supportedFormat is True
+    assert parsed_qasm.qelib1Include is True
+
+    ct.assert_same_circuits(parsed_qasm.circuit, expected_circuit)
+    assert parsed_qasm.qregs == {'q1': 3}
+    assert parsed_qasm.cregs == {'c1': 3}
+    cirq.Simulator().run(parsed_qasm.circuit)
```
vtomole: Why call `.run` here?

cirq/contrib/qasm_import/_parser_test.py
balopat: ahh, I forgot that one in again. Removing soon.

cirq/contrib/qasm_import/_parser_test.py
balopat: same as above. Thank you for catching these!

cirq/contrib/qasm_import/_parser.py
```diff
@@ -309,6 +311,12 @@ def p_arg_register(self, p):
                     self.qubits[arg_name] = NamedQubit(arg_name)
                 qubits.append(self.qubits[arg_name])
             p[0] = qubits
+        elif reg in self.cregs.keys():
+            keys = []
+            for num in range(self.cregs[reg]):
+                arg_name = self.make_name(num, reg)
+                keys.append(arg_name)
+            p[0] = keys
```
Strilanc: This looks like it could be a one-liner list comprehension.

cirq/contrib/qasm_import/_parser.py
```diff
@@ -331,10 +339,30 @@ def p_arg_bit(self, p):
             if arg_name not in self.qubits.keys():
                 self.qubits[arg_name] = NamedQubit(arg_name)
             p[0] = [self.qubits[arg_name]]
+        elif reg in self.cregs.keys():
+            p[0] = [arg_name]
         else:
             raise QasmException('Undefined quantum register "{}" '
                                 'at line {}'.format(reg, p.lineno(1)))
 
+    # measurement operations
```
Strilanc: Does anything check that the type of register is correct? E.g. that you don't measure classical into quantum.

cirq/contrib/qasm_import/_parser_test.py
```diff
@@ -360,3 +360,83 @@ def test_undefined_register_from_register_arg():
 
     with pytest.raises(QasmException, match=r"""Undefined.*register.*q.*"""):
         parser.parse(qasm)
+
+
+def test_measure_individual_bits():
+    qasm = """
```
Strilanc: Add a failure case where the output register is quantum, and another where the measured register is classical.

cirq/contrib/qasm_import/_parser.py
balopat: yeah that check was missing - solved it by handling quantum and classical registers separately

balopat:
@Strilanc @vtomole  /ready-for-re-review

cirq/contrib/qasm_import/_parser.py
```diff
@@ -284,56 +286,101 @@ def p_term(self, p):
...
+        """carg : ID '[' NATURAL_NUMBER ']' """
+        reg = p[1]
+        num = p[3]
+        arg_name = self.make_name(num, reg)
+        if reg not in self.cregs.keys():
+            raise QasmException('Undefined classical register "{}" '
+                                'at line {}'.format(reg, p.lineno(1)))
+
+        size = self.cregs[reg]
+        if num > size - 1:
```
Strilanc: `index >= size`

cirq/contrib/qasm_import/_parser.py
```diff
@@ -284,56 +286,101 @@ def p_term(self, p):
...
+                                'at line {}'.format(num, reg, size,
+                                                    p.lineno(1)))
+        if arg_name not in self.qubits.keys():
+            self.qubits[arg_name] = NamedQubit(arg_name)
+        p[0] = [self.qubits[arg_name]]
+
+    def p_classical_arg_bit(self, p):
+        """carg : ID '[' NATURAL_NUMBER ']' """
+        reg = p[1]
+        num = p[3]
```
Strilanc: Use `index` instead of `num`, because `num` suggests lengths or counts which will make comparisons look off-by-one (e.g. an index is out of range if >= whereas a length is out of range if >).

Issue #1687, opened 2019-05-30
Raise an error when attempting to sample a circuit with duplicate measurement keys


Strilanc:
This should be raising an error:

```python
import cirq
q = cirq.LineQubit(0)
c = cirq.Circuit.from_ops(
     cirq.measure(q, key='b'),
     cirq.measure(q, key='a'),
     cirq.measure(q, key='a'))
cirq.sample(c)
```

Instead it returns malformed data:

```
a=00
b=0
```

iamvamsikrishnad:
I'll take a stab at it. Can you please assign it to me.

iamvamsikrishnad:
@Strilanc : Can you please let me know if this sounds okay ?

Strilanc:
That is not sufficient, because operations may be appended into circuits dynamically via append, iadd, etc.

We can either enforce this at the circuit level by checking at all edit points or at the simulator level by checking at all entry points.

I do think we should check in Moment's constructor, as a sanity check.

iamvamsikrishnad:
Thanks @Strilanc 

I am almost finished with the coding. One question I have before submitting PR is:
how to deal with measurements with not key provided (as I see that the cirq.measure doesn't need key as mandatory arg). This creates two cases where all the measurements are provided without keys and the case where some measurements has keys and some don't.

iamvamsikrishnad:
https://github.com/quantumlib/Cirq/pull/1746/files

PR #1746, opened 2019-06-25
Check for duplicate measurement keys #1687

iamvamsikrishnad:


cirq/circuits/circuit.py
```diff
@@ -1841,3 +1850,27 @@ def _group_until_different(items: Iterable[TIn],
         Tuples containing the group key and item values.
     """
     return ((k, [value(i) for i in v]) for (k, v) in groupby(items, key))
+
+def _get_measurement_keys(
```
Strilanc: Add output type annotation

Strilanc: Rename to `_get_measurement_keys_and_verify_unique`

cirq/circuits/circuit.py
```diff
@@ -1841,3 +1850,27 @@ def _group_until_different(items: Iterable[TIn],
...
+        moment_or_operation_list: Union[ops.Moment, ops.Operation],
+        validate: bool = True):
+    seen = set()  # type: Set[str]
+    for moment_or_op in moment_or_operation_list:
```
Strilanc:
```
result = collections.Counter(
    protocols.measurement_key(op, default=None)
    for op in ops.flatten_op_tree(moment_or_operation_list)
)
result[None] = 0
duplicates = [k for k, v in result.entries() if v > 1]
if duplicates:
    raise ValueError('Duplicate measurement keys: {!r}'.format(duplicates)
return set(duplicates.keys())
```

cirq/circuits/circuit.py
```diff
@@ -884,6 +886,13 @@ def insert(
                                   preserve_moments=True),
             preserve_moments=True))
 
+        common_keys = _get_measurement_keys(
```
Strilanc: This is a very expensive check to perform every time there is an insertion... It iterates over every operation in the entire circuit, causing quadratic overhead.

If we're going to do this, we need to cache the set of used measurements and keep that cache up to date.

This is why I originally specified the idea in terms of the Simulator verifying that the circuit has no dupes; keeping the cache up to date will be a pain.

iamvamsikrishnad: Thanks for the review @Strilanc.

I weighed the current option against the simulator. I saw all the circuit/device constraints are checked when the circuit is being built and hence thought this check should also be there to maintain consistency. 

As you can see, I am not checking for measurement_key duplication within a moment in `_get_measurement_keys` method as that is already validated when constructing the moment. This is a slight cut in computation cost, but still I agree it is expensive. 

The other simple way to cut the computation cost is to cache the measurement keys at the moment level. I didn't do it because `Moment` class currently has no member variables like this.

iamvamsikrishnad: Please let me know what do you think!

iamvamsikrishnad: Gentle ping for review.

Strilanc: Iterating over the entire circuit every time we add an operation is too inefficient. This has to be fixed before we can merge.

iamvamsikrishnad: Do you prefer the idea of verifying in simulator compared to the proposal of adding a cache at the moment level ?

iamvamsikrishnad:
Since, this approach would lead to lot of cache updates and is not recommended, I will close this pull request and approach the identification of duplicates in the simulator path.

PR #1747, opened 2019-06-25
Refactor cirq.has_unitary into a list-of-strategies implementation

Strilanc:
- Also move it into its own file

### DIFF
--- a/cirq/protocols/has_unitary.py
+++ b/cirq/protocols/has_unitary.py
@@ -97,10 +97,8 @@ def has_unitary(val: Any) -> bool:
         Whether or not `val` has a unitary effect.
     """
     strats = [
-        _strat_has_unitary_from_has_unitary,
-        _strat_has_unitary_from_decompose,
-        _strat_has_unitary_from_apply_unitary,
-        _strat_has_unitary_from_unitary
+        _strat_has_unitary_from_has_unitary, _strat_has_unitary_from_decompose,
+        _strat_has_unitary_from_apply_unitary, _strat_has_unitary_from_unitary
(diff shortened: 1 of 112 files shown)


cirq/protocols/has_unitary.py
```diff
@@ -0,0 +1,182 @@
...
+    return False
+
+
+def _strat_has_unitary_from_has_unitary(val: Any) -> Optional[bool]:
+    """Attempts to infer a value's unitary-ness via its _has_unitary_ method."""
+    getter = getattr(val, '_has_unitary_', None)
+    result = NotImplemented if getter is None else getter()
+    if result is NotImplemented:
+        return None
+    return result
```
mpharrigan: Why is this function so convoluted? If there is no `_has_unitary_` then the flow is: `getter = None -> result = NotImplemented -> return = None` which seems roundabout.

Maybe: 
```python
if hasattr(val, '_has_unitary_'):
  result = getattr(val, '_has_unitary_')()
  if result is NotImplemented:
    return None
  return result
return None
```

The behavior would be different for classes that explicitly define a class attribute `_has_unitary_ = None` but I don't think that should ever happen (?)

Strilanc: That rewrite does two attribute lookups instead of one... but yes I agree it's clearer so I'll use it.

mpharrigan: > That rewrite does two attribute lookups instead of one

I knew someone would point this out :) Caring about performance while using `getattr` is a bit of an oxymoron though :)

PR #1749, opened 2019-06-25
Rename Circuit.apply_unitary_effect_to_state to Circuit.final_wavefunction

Strilanc:
- So that we're consistent with cirq.final_wavefunction
- Keeping Circuit's method because it provides more specialized options

### DIFF
--- a/cirq/circuits/circuit_test.py
+++ b/cirq/circuits/circuit_test.py
@@ -2024,85 +2024,85 @@ def test_apply_unitary_effect_to_state():
 
     # State ordering.
     cirq.testing.assert_allclose_up_to_global_phase(
-        cirq.Circuit.from_ops(cirq.X(a)**0.5).apply_unitary_effect_to_state(),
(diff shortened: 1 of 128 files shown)


cirq/circuits/circuit_test.py
```diff
@@ -2024,113 +2024,100 @@ def test_apply_unitary_effect_to_state():
...
-    cirq.testing.assert_allclose_up_to_global_phase(
-        cirq.Circuit.from_ops(cirq.X(a)**0.5).apply_unitary_effect_to_state(
-            initial_state=1),
-        np.array([1, 1j]) * np.sqrt(0.5),
-        atol=1e-8)
+    cirq.testing.assert_allclose_up_to_global_phase(cirq.Circuit.from_ops(
+        cirq.X(a)**0.5).final_wavefunction(initial_state=0),
+                                                    np.array([1j, 1]) *
+                                                    np.sqrt(0.5),
+                                                    atol=1e-8)
```
maffoo: yapf is so weird...

Issue #1701, opened 2019-06-07
cannot import name 'TYPE_CHECKING' in python 3.7


iamvamsikrishnad:
While using python-3.7 and running 
`python -m pip install -r dev_tools/conf/pip-list-dev-tools.txt`
throws an error:

> pylint 2.3.1 has requirement astroid<3,>=2.2.0, but you'll have astroid 2.1.0 which is incompatible.

Updating the requirement of astroid in pip-list-dev-tools.txt to 2.2.* makes it work.

vtomole:
Hey @iamvamsikrishnad, this was a compromise we had to make to be on a latest version of pylint while getting around this bug:
https://github.com/quantumlib/Cirq/blob/1e6436c136dbe94a62970fa64692d27ed8f919c8/dev_tools/conf/pip-list-dev-tools.txt#L21

iamvamsikrishnad:
Hi,
Thanks for the reply. I am not sure I totally follow you. 
Yes it (https://github.com/PyCQA/astroid/issues/650) is a issue in 3.5.2 and hence astroid was pinned to version 2.1.*, but that is breaking it for users on python 3.7. I believe we need to find a fix for both versions right ?

vtomole:
>  I believe we need to find a fix for both versions right ?

Yes, the fix is solving that issue that you linked which is still open. I've spent an hour or two playing around with astroid to try to fix that bug with no luck.

iamvamsikrishnad:
In that case, Can you please keep this issue open.

vtomole:
This issue is not a bug in cirq. It's a bug in astroid.

iamvamsikrishnad:
I'd leave it up to your judgement. 
I feel since the cirq installation is failing because of the dependencies, we should track it here as a cirq bug as well.

vtomole:
The cirq installation does't fail because of this. It throws a warning that the versions of astroid and pylint are not compatible. If i'm wrong about this let me know. Also, the title of this issue doesn't seem to match what we are discussing right now.

iamvamsikrishnad:
Sure.

PR #1754, opened 2019-06-26
Fixes type import in QASM parser

bryano:


cirq/contrib/qasm_import/_parser.py
```diff
@@ -22,6 +23,10 @@
 from cirq.contrib.qasm_import._lexer import QasmLexer
 from cirq.contrib.qasm_import.exception import QasmException
 
+if TYPE_CHECKING:
+    # pylint: disable=unused-import
+    from typing import Tuple
```
Strilanc: This is for the line

`for qubits in op_qubits:  # type: Tuple[cirq.Qid]`

?

That refers to the type of the loop variable? You should instead change the `type: np.broadcast` on the previous line. For example:

`op_qubits = cast(Sequence[Sequence[ops.Qid]], functools.reduce(np.broadcast, args))`

bryano: I didn't even look at how `Tuple` was used. I'm just trying to shut pylint up...

bryano: Done

bryano:
Is there a reason this file imports `cirq` at the top level rather than its submodules?

cirq/contrib/qasm_import/_parser.py
balopat: I'm wondering why pylint is reporting this as an error for you and not on Travis / or on my machine. 
Is it the right version? `dev_tools/conf/pip-list-dev-tools.txt` declares `pylint==2.3.*`

balopat:
> Is there a reason this file imports `cirq` at the top level rather than its submodules?

not really, it is simply an overlook on my end :)

cirq/contrib/qasm_import/_parser.py
bryano: I have pylint 2.3.1

Strilanc:
Oh sorry @bryano , I confused you with @balopat , which is why I was asking for a bunch of other stuff.

bryano:
Would be good to know why pylint's complaining on my machine and not others.

Strilanc:
Try `pip install --upgrade ./dev_tools/conf/pip-list-dev-tools.txt` and see if it persists.

bryano:
No difference (all those requirements were already satisfied).

balopat:
hmm, interesting, maybe it's something else. 
on my mac I have this setup: 
```
$ pylint --version
pylint 2.3.1
astroid 2.1.0
Python 3.6.4 (v3.6.4:d48ecebad5, Dec 18 2017, 21:07:28) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)]
```
What is your `pylint --version` output?

vtomole:
@balopat We are aware of that warning: https://github.com/quantumlib/Cirq/issues/1701

bryano:
Here's my pylint version info:
```
pylint 2.3.1
astroid 2.1.0
Python 3.6.8 |Anaconda, Inc.| (default, Dec 30 2018, 01:22:34) 
[GCC 7.3.0]
```

bryano:
What surprises me is not that pylint complained on my machine, but that it didn't on Travis. When not type checking, that was an actually unused import.

PR #1757, opened 2019-06-27
[QASM import] mathematical expressions for parameters

balopat:
Adds mathematical expressions that are supported by Quantum Experience. 
Also fixes the cirq imports: instead of directly importing cirq, we import `cirq.ops` now.

This is PR number 6 for the piecemeal reimplementation of #1602.

### DIFF
--- a/cirq/contrib/qasm_import/_parser.py
+++ b/cirq/contrib/qasm_import/_parser.py
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import functools
-from typing import Dict, Optional, List, Any, Iterable, Callable, Tuple, Union
+from typing import (Any, Callable, cast, Dict, Iterable, List, Optional,
+                    Sequence, Union)
+
 import numpy as np
 from ply import yacc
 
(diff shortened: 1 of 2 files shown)


cirq/contrib/qasm_import/_parser.py
```diff
@@ -12,13 +12,14 @@
...
 import functools
+import operator
 from typing import (Any, Callable, cast, Dict, Iterable, List, Optional,
                     Sequence, Union)
 
 import numpy as np
 from ply import yacc
 
-import cirq
+from cirq import ops
```
Strilanc: Merge with next line

cirq/contrib/qasm_import/_parser.py
```diff
@@ -132,7 +132,26 @@ def __init__(self):
         self.lexer = QasmLexer()
         self.supported_format = False
         self.parsedQasm = None  # type: Optional[Qasm]
-        self.qubits = {}  # type: Dict[str,cirq.NamedQubit]
+        self.qubits = {}  # type: Dict[str,ops.NamedQubit]
```
Strilanc: Weaken NamedQubit to `Qid` unless the name is really relevant

cirq/contrib/qasm_import/_parser_test.py
```diff
@@ -303,6 +303,71 @@ def test_u3_angles():
...
+
+    q0 = cirq.NamedQubit('q_0')
+
+    expected_circuit = Circuit()
+    expected_circuit.append(
+        QasmUGate(float(sympy.sympify(expr)) / np.pi, 2.0, 1 / 2.0)(q0))
+
+    parsed_qasm = parser.parse(qasm)
+
+    assert parsed_qasm.supportedFormat is True
```
Strilanc: Unless this value can be something besides true or false, just assert it instead of asserting is true

cirq/contrib/qasm_import/_parser_test.py
```diff
@@ -303,6 +303,71 @@ def test_u3_angles():
...
+    assert parsed_qasm.qelib1Include is False
```
Strilanc: Unless this value can be something besides true or false, just assert not it instead of asserting is false

Issue #1698, opened 2019-06-05
Circuit + OP_TREE


kevinsung:
Since `circuit += op_tree` works, I propose to let `circuit + op_tree` work too.

Strilanc:
Makes sense to me.

gauravMann:
I am willing to take this up. I have read the contributor's guide, and set up my local environment.

vtomole:
@gauravMann Go for it!

Strilanc:
Ah right, I forgot about the device inter-compatibility check. That's important to keep. The validation of moments in `other` can be delegated to `copy.__iadd__`, but the validation of moments in `self` would stay.

gauravMann:
Hi @Strilanc I apologize for the delay. 
Please review the pull request [https://github.com/quantumlib/Cirq/pull/1762](url)

PR #1762, opened 2019-06-28
Allow adding OP_TREEs into Circuit

gauravMann:
The method now delegates to `__iadd__` and `__init__` for validation of moments.

Fixes https://github.com/quantumlib/Cirq/issues/1698


gauravMann:
@Strilanc Please review the changes.
I removed the code `other.device = device` as it would have resulted in a mysterious bug.
 In doing `r = c1 + c2` I was changing c2's device.

Also the commit history got a little messy so I squashed all of them.

PR #1763, opened 2019-06-28
[Qasm import] rotation gates

balopat:
Adds support for `rx`, `ry` and `rz` gates in QASM import and as such, the first gates that are defined as part of the "qelib1.inc" file. 

This is PR number 7 for the piecemeal reimplementation of #1602.

4 more PRs are planned: 
- "special" - u3(=`U`), u2, u1 and id gates
- single qubit gates - x,y,z,h,t,tdg,s,sdg
- two qubit gates - cx (=`CX`),cy,cz,ch, swap
- three qubit gates - ccx, cswap

### DIFF
--- a/cirq/contrib/qasm_import/_parser.py
+++ b/cirq/contrib/qasm_import/_parser.py
@@ -287,12 +287,12 @@ class QasmParser:
 
     def _resolve_gate_operation(self, args: List[List[ops.Qid]], gate: str,
                                 p: Any, params: List[float]):
-        helper_text = (", did you forget to include qelib1.inc?"
-                       if not self.qelibinc else "")
         gate_set = (self.basic_gates if not self.qelibinc else self.qelib_gates)
         if gate not in gate_set.keys():
-            raise QasmException('Unknown gate "{}" at line {}{}'.format(
-                gate, p.lineno(1), helper_text))
+            msg = 'Unknown gate "{}" at line {}{}'.format(
+                gate, p.lineno(1), ", did you forget to include qelib1.inc?"
+                if not self.qelibinc else "")
+            raise QasmException(msg)
         p[0] = gate_set[gate].on(args=args, params=params, lineno=p.lineno(1))
 
     # params : parameter ',' params
--- a/cirq/contrib/qasm_import/_parser.py
+++ b/cirq/contrib/qasm_import/_parser.py
@@ -64,6 +64,9 @@ class QasmGateStatement:
         self.qasm_gate = qasm_gate
         self.cirq_gate = cirq_gate
         self.num_params = num_params
+
+        # at least one quantum argument is mandatory for gates to act on
+        assert num_args >= 1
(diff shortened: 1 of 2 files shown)


cirq/contrib/qasm_import/_parser.py
```diff
@@ -267,14 +287,13 @@ def p_gate_op_with_params(self, p):
...
                                 p: Any, params: List[float]):
-        if gate not in self.basic_gates.keys():
-            raise QasmException('Unknown gate "{}" at line {}, '
-                                'maybe you forgot to include '
-                                'the standard qelib1.inc?'.format(
-                                    gate, p.lineno(1)))
-        p[0] = self.basic_gates[gate].on(args=args,
-                                         params=params,
-                                         lineno=p.lineno(1))
+        helper_text = (", did you forget to include qelib1.inc?"
```
Strilanc: Don't compute error text on success branches. Move this inside the failure case.

(Why isn't it inlined into the message in the first place?)

cirq/contrib/qasm_import/_parser_test.py
```diff
@@ -368,39 +392,102 @@ def test_unknown_function():
...
+     {}(pi) q[0], q[1];     
+""".format(qasm_gate)
 
     parser = QasmParser()
 
-    with pytest.raises(QasmException,
-                       match=r"U takes 3 parameter\(s\).*got.*0.*line 3"):
+    with pytest.raises(
+            QasmException,
+            match=r".*{}.* takes 1 arg\(s\).*got.*2.*line 5".format(qasm_gate)):
```
Strilanc: Your matches are often very specific. No need to change that. Personally I make them vague (e.g. just `"takes 1 arg"`) so that it's easier to update the language.

cirq/contrib/qasm_import/_parser_test.py
```diff
@@ -368,39 +392,102 @@ def test_unknown_function():
         parser.parse(qasm)
 
 
-def test_u_gate_zero_params_error():
+rotation_gates = [
+    ('rx', cirq.Rx),
```
Strilanc: Are there any tests actually using these statements?

balopat: Yeah all three new tests are parameterized and using these

cirq/contrib/qasm_import/_parser.py
balopat: Hmm, yeah, this is pretty gnarly, I rushed it on my commute this morning. I'll move it inline to the error message.

cirq/contrib/qasm_import/_parser_test.py
balopat: Yeah it's a trade-off, here I wanted the data to show up - i.e lineno, num args accepted and actual arg number with some context to make the test understandable what I'm matching for. I could make the context smaller though. I'll think about it.

PR #1766, opened 2019-07-02
[QASM import] u3, u2, id gates

### LINKED ISSUE #1775: Qasm output should include identity gates

balopat:
Adds support for `u3`, `u2`, `id` gates in QASM import and improved testing by parametrized tests for arg and parameter error paths and expanded consistency check using the currently supported gates.  

This is PR number 8 for the piecemeal reimplementation of #1602.

3 more PRs are planned: 
- single qubit gates - x,y,z,h,t,tdg,s,sdg
- two qubit gates - cx (=`CX`),cy,cz,ch, swap
- three qubit gates - ccx, cswap

Fixes #1775.

### DIFF
--- a/cirq/circuits/qasm_output.py
+++ b/cirq/circuits/qasm_output.py
@@ -75,9 +75,9 @@ class QasmUGate(ops.SingleQubitGate):
 
     def __eq__(self, other):
         return isinstance(other, QasmUGate) and \
-               other.lmda == self.lmda and \
-               other.theta == self.theta and \
-               other.phi == self.phi
+               np.isclose(other.lmda, self.lmda) and \
+               np.isclose(other.theta, self.theta) and \
+               np.isclose(other.phi, self.phi)
 
     def __hash__(self):
         return hash((self.lmda, self.theta, self.phi))
--- a/cirq/circuits/qasm_output.py
+++ b/cirq/circuits/qasm_output.py
@@ -14,7 +14,7 @@
 
 """Utility classes for representing QASM."""
 
-from typing import Set  # pylint: disable=unused-import
+from typing import Set, Any  # pylint: disable=unused-import
 from typing import (
     Callable, Dict, Optional, Sequence, Tuple, Union
 )
@@ -25,6 +25,7 @@ import numpy as np
 from cirq import ops, linalg, protocols, value
 
 
+@value.value_equality(unhashable=True, approximate=True)
 class QasmUGate(ops.SingleQubitGate):
 
     def __init__(self, theta, phi, lmda) -> None:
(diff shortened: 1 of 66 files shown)


cirq/contrib/qasm_import/qasm_test.py
```diff
@@ -11,12 +11,29 @@
...
 
 # TODO as we grow the language, we'll add more complex examples here
 def test_consistency_with_qasm_output():
-    circuit1 = cirq.Circuit()
+    a, b, c, d = [cirq.NamedQubit('q_{}'.format(i)) for i in range(4)]
+    circuit1 = cirq.Circuit.from_ops(
+        cirq.Rx(np.pi / 2).on(a),
+        cirq.Ry(np.pi / 2).on(b),
+        cirq.Rz(np.pi / 2).on(b),
+        cirq.IdentityGate(1).on(c),
```
balopat: TODO: look into qasm output for Identity

balopat: Actually - `u3(0,0,0)` was the current output for a 1 qubit IdentityGate which is not bad, but for two qubits:

```
// Generated from Cirq v0.6.0.dev

OPENQASM 2.0;
include "qelib1.inc";


// Qubits: [q1, q2]
qreg q[2];


// Gate: I(2)
u3(pi*1.0,0,pi*0.5) q[0];
u3(0,0,0) q[1];
rx(pi*0.5) q[0];
cx q[0],q[1];
rx(pi*0.5) q[0];
ry(pi*0.5) q[1];
cx q[1],q[0];
rx(pi*-0.5) q[1];
rz(pi*0.5) q[1];
cx q[0],q[1];
u3(pi*1.0,0,pi*0.5) q[0];
u3(0,0,0) q[1];

```

balopat: with the new commits, the two qubit version is: 

```
// Generated from Cirq v0.6.0.dev

OPENQASM 2.0;
include "qelib1.inc";


// Qubits: [0, 1]
qreg q[2];


id q[0];
id q[1];
```

cirq/circuits/qasm_output.py
```diff
@@ -25,6 +25,7 @@
 from cirq import ops, linalg, protocols, value
 
 
+@value.value_equality(unhashable=True, approximate=True)
```
Strilanc: Why unhashable?

balopat: yeah, it is not strictly necessary, I can fix it in the next one. 
 
I am still unsure what is the use of a __hash__ function in the approximate equality case? In my understanding, checking equality on these classes is not useful due to the floating point nature of the underlying values. And having the default `_value_equality_hash` implementation is more expensive than a `None`...

Strilanc: approximate=True does not change the behavior of `__eq__`, it adds a method `_approx_eq_`.

PR #1770, opened 2019-07-02
Add 'no-else-return' to .pylintrc

vtomole:
Merge after https://github.com/quantumlib/Cirq/pull/1769 and https://github.com/quantumlib/Cirq/pull/1768

### DIFF
--- a/cirq/ops/clifford_gate.py
+++ b/cirq/ops/clifford_gate.py
@@ -317,7 +317,7 @@ class SingleQubitCliffordGate(gate_features.SingleQubitGate):
                 # Positive 90 degree rotation
                 output.append((pauli, 1))
             return output
-        elif num_whole == 0:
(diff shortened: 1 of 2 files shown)


Strilanc:
I'm not sure we should merge this. Do we consider this to be an important part of our style guide?

I think this is an example of enforcing a rule that doesn't *really* matter. Code isn't more error-prone if it has redundant else's, and it won't be more performant... it's just a bit less verbose. Note how all the other lint checks are either about potential mistakes or optimizations.

vtomole:
> it's just a bit less verbose

I think this is good. It's the ultimate `nit` ;)

vtomole:
@Strilanc Will merge or close after a response to my comment.

Strilanc:
We decided at the group meeting to keep linting about possible mistakes, instead of style issues.

Issue #933, opened 2018-09-18
Handle Qudits


dabacon:
Currently we use "QubitId" everywhere, but tying ourselves to two level systems seems bad.  How are we going to work with higher dimensional quantum systems, or boson or fermion (or heaven forbit para-fermion) based systems?

Strilanc:
Everything we do with operations should correspond to twiddling some kind of quantum control knob. A `Qnob` if you will...

singular-value:
I'm going to start making some changes in my local Cirq fork to support qudit handling (well specifically, I want qutrit simulation, but I'll plan to write qutrits as the d=3 case of qudits).

Do you think there will be a reasonable case to merge my (planned) local changes into Cirq? I can see a case for keeping everything in core Cirq in qubits (it could be irritating for beginners to need to be aware of qudits). But if you were definitely planning on building qudits into Cirq, I could do my best to do it exactly as you planned (e.g. rename Qubit to Qudit?). Let me know if you have any comments/suggestions. Thanks!

-Pranav Gokhale (PhD student at UChicago with Fred Chong)

Strilanc:
The main blocker for supporting qudits is the fact that `cirq.Operation` has a method called `qubits`, and I suspect there is code that assumes the matrix of an operation should have a width and height equal to `2**len(qubits)`.

One possible way to fix this would be to replace all matrices with tensors. This would replace the size assumption with a dimension-of-shape assumption (`len(cirq.unitary(op)) = 2*len(op.quwhatevers)`).

dabacon:
Suggested names for QubitId:

QuId
QId
Quid
Qid

The thing that will happen here that is odd is that methods that return "qubits" now return "quids" which is not really discoverable.  Also places where we pass in Qubits no take quids, which is also kind of odd.

Strilanc:
I vote `Qid`. Get your votes in now, everyone! @kevinsung @viathor @vtomole @maffoo @kevinsung @babbush @mrwojtek @Ashalynd @cduck @bryano @Ican'tremembereveryone'sscreennames

bryano:
What about QuditId? Or just Qudit?

The easy thing here is to have QubitId just be a special case of QuditId with the dimension attribute set to 2. Things could then be generalized to qudits gradually.

mrwojtek:
My vote goes to quid; it has a nice ring to it and one can make useful variations out of it easily: qubid, qutid, quqid, qupid, etc,. if ever needed.

[...]

PR #1771, opened 2019-07-03
[Qudits] Add support to cirq.unitary, cirq.apply_unitary, and Circuit

cduck:
For #933.  Part 2 after #1777.

- `cirq.unitary` and `cirq.apply_unitary` check for the `qid_shape` protocol and use properly sized numpy arrays.
- `Circuit` also checks for the `qid_shape` protocol for `Circuit.unitary` and related methods.
- `Circuit` now implements the `qid_shape` protocol.

Qudit features not in this PR are channels, mixtures, simulators, measurements, optimizers.


cirq/circuits/circuit.py
```diff
@@ -1379,17 +1401,20 @@ def apply_unitary_effect_to_state(
 
         qs = ops.QubitOrder.as_qubit_order(qubit_order).order_for(
             self.all_qubits().union(qubits_that_should_be_present))
-        n = len(qs)
+
+        # Force qubits to have dimension at least 2 for backwards compatibility.
+        qid_shape = self.max_qid_shape(qs, qid_min=2)
```
Strilanc: Do you have an example where this fails? I'm okay with this failing on user gates that didn't account for it, because the user will also be providing the shape that causes the failure.

cirq/circuits/circuit.py
```diff
@@ -1228,6 +1228,25 @@ def all_operations(self) -> Iterator[ops.Operation]:
         """
         return (op for moment in self for op in moment.operations)
 
+    def max_qid_shape(
```
Strilanc: This method needs a docstring.

Could you describe why we need this method, instead of just a shape method that takes a qid_order?

The `qid_min` argument seems unnecessary, since `np.max` exists.

cirq/protocols/apply_unitary.py
```diff
@@ -86,28 +90,60 @@ def __init__(self,
...
             axes: Which axes the unitary effect is being applied to (e.g. the
                 qubits that the gate is operating on).
+            qid_shape: A tuple of the dimensions of each qid in the state
+                `target_tensor`.
         """
+        self.axes = tuple(axes)
+        if qid_shape is None:
+            # Guess what the qid_shape would be for qubits.
+            min_num_qubits = max(self.axes, default=-1) + 1
+            qid_shape = (2,) * min_num_qubits  # Might be too short
```
Strilanc: Because users would be unlikely to make this particular value, we can make the argument required by the constructor. This will avoid this very bad case where the shape is wrong.

cirq/protocols/apply_unitary.py
```diff
@@ -86,28 +90,60 @@ def __init__(self,
...
+                               shape=qid_shape,
                                dtype=np.complex128)
-        return ApplyUnitaryArgs(state, np.empty_like(state), range(num_qubits))
-
-    def subspace_index(self, little_endian_bits_int: int
-                       ) -> Tuple[Union[slice, int, 'ellipsis'], ...]:
+        return ApplyUnitaryArgs(state, np.empty_like(state), range(num_qubits),
+                                qid_shape)
+
+    def subspace_index(
```
Strilanc: This method is specifically about "I want to give an integer index", so I would not add a tuple variant argument.

Instead of that, we should use the shape to pull apart the integer in a little endian fashion. For example `5 = 2*1 + 1*3` so 5 corresponds to [2, 1] when the shape is (*, 3).

We can separately consider adding an alternate method that scatters a tuple of indices over the axis indices.

cirq/protocols/apply_unitary.py
```diff
@@ -227,27 +264,48 @@ def apply_unitary(unitary_value: Any,
     if func is not None:
         result = func(args)
         if result is not NotImplemented and result is not None:
+            if result is args.target_tensor:
```
Strilanc: This is not sufficient. For example, it could fail if a user makes the mistake of saying `args.target_tensor *= -0.5`, which is something that HPowGate does.

I think what we need to do is slice down the tensor to the shape expected by the gate before passing it along, in order to guarantee that it is working in a subspace that it understands. Then we can separately handle the remainder using this approach.

cirq/protocols/apply_unitary.py
```diff
@@ -227,27 +264,48 @@ def apply_unitary(unitary_value: Any,
...
+            # copy the untouched entries from args.target_tensor into result.
+            op_shape = qid_shape_protocol.qid_shape(unitary_value,
+                                                    (2,) * len(args.axes))
+            if any(op_shape[i] < args.qid_shape[axis]
+                   for i, axis in enumerate(args.axes)):
+                # Copy extra entries from args.target_tensor into result.
+                for i in range(len(args.axes)):
+                    axis = args.axes[i]
+                    op_d = op_shape[i]
+                    axis_d = args.qid_shape[axis]
```
Strilanc: Make sure to only copy each place in the array once (e.g. don't copy (2:, :) and also (:, 2:) because this does (2:,2:) twice).

cirq/protocols/apply_unitary.py
```diff
@@ -227,27 +264,48 @@ def apply_unitary(unitary_value: Any,
...
             return result
 
     # Fallback to using the object's _unitary_ matrix.
     matrix = unitary(unitary_value, None)
     if matrix is not None:
-        # Special case for single-qubit operations.
-        if matrix.shape == (2, 2):
-            zero = args.subspace_index(0)
-            one = args.subspace_index(1)
+        # Special case for single-qubit operations.(2,) * (2 * len(args.axes))
```
Strilanc: What's the expression at the end of the comment?

cirq/protocols/apply_unitary.py
```diff
@@ -227,27 +264,48 @@ def apply_unitary(unitary_value: Any,
...
+        if len(args.axes) == 1:
```
Strilanc: This can be specific to the `(2,)` shape, because the tradeoff depends very much on the size of the matrix and the defualt einsum thing might already be winning by 3x3.

cirq/protocols/apply_unitary.py
```diff
@@ -304,17 +362,27 @@ def apply_unitaries(unitary_values: Iterable[Any],
...
     """
     if args is None:
-        args = ApplyUnitaryArgs.default(len(qubits))
+        # Default to 2 for backwards compatibility
+        max_shape_dict = {qubit: 2 for qubit in qubits}
+        unitary_values = tuple(unitary_values)
+        # Iterate over all unitary values to find the maximum qid_shape
+        for op in unitary_values:
+            for d, qubit in zip(qid_shape_protocol.qid_shape(op), op.qubits):
+                max_shape_dict[qubit] = max(d, max_shape_dict[qubit])
```
Strilanc: This looks like something that `cirq.qid_shape` should do for you. If you get an iterable of operations, take the max shape in the obvious way.

cirq/circuits/circuit.py
cduck: This happens when there is a qubit that has no gates operating on it (e.g. when explicitly specifying a list of qubits).  There are several test cases that fail when `qid_min=1`.

cirq/circuits/circuit.py
cduck: Because circuits don't really have a well-defined qubit order but everything else does.

Something like this?  `tuple(np.max([(1,2,3,4),(2,)*4], axis=0)) => array([2, 2, 3, 4])`

cirq/protocols/apply_unitary.py
cduck: I didn't realize that.  I can change it to `if len(args.axes) == 1 and matrix.shape == (2, 2):`.

cirq/protocols/apply_unitary.py
cduck: That's a typo.

cirq/protocols/apply_unitary.py
cduck: Removed tuple argument.

cirq/protocols/apply_unitary.py
cduck: Fixed with new args slicing method.

cirq/protocols/apply_unitary.py
cduck: Fixed with new args slicing method.

cduck:
@Strilanc please review.  I still need to add a few tests though.

cirq/linalg/states.py
```diff
@@ -38,3 +38,20 @@ def one_hot(*,
     result = np.zeros(shape=shape, dtype=dtype)
     result[index] = 1
     return result
+
+
+def eye_tensor(*, qid_shape: Tuple[int, ...],
```
Strilanc: Could rename from `qid_shape` to `input_shape` or `output_shape` or `half_shape`.

Don't force keyword args on the shape, to act more like numpy.

cirq/protocols/apply_unitary.py
```diff
@@ -85,28 +89,122 @@ def __init__(self,
...
+
+        Specify exactly one argument.
+
+        Args:
+            num_qubits: The number of qubits to make space for in the state.
+            qid_shape: The shape of the state, specifying the dimension of each
+                qid."""
+        if (num_qubits is None) == (qid_shape is None):
+            raise TypeError(
+                'Specify either the num_qubits or qid_shape argument.')
```
Strilanc: "exactly one of"

cirq/protocols/apply_unitary.py
```diff
@@ -85,28 +89,122 @@ def __init__(self,
...
+        if num_qubits is not None:
+            qid_shape = (2,) * num_qubits
+        assert qid_shape is not None, "Can't be None here"  # Satisfy mypy
```
Strilanc: You can use `typing.cast`

cirq/protocols/apply_unitary.py
```diff
@@ -85,28 +89,122 @@ def __init__(self,
...
+        num_qubits = len(qid_shape)
         state = linalg.one_hot(index=(0,) * num_qubits,
-                               shape=(2,) * num_qubits,
+                               shape=qid_shape,
                                dtype=np.complex128)
         return ApplyUnitaryArgs(state, np.empty_like(state), range(num_qubits))
 
-    def subspace_index(self, little_endian_bits_int: int
-                       ) -> Tuple[Union[slice, int, 'ellipsis'], ...]:
+    def with_transposed_and_sliced_tensors(
```
Strilanc: Name doesn't make it clear what the `indices` argument is. I would have expected it to be a permutation.

Maybe `with_right_aligned_axes`?

cirq/__init__.py
```diff
@@ -61,6 +61,7 @@
     dot,
     expand_matrix_in_orthogonal_basis,
     hilbert_schmidt_inner_product,
+    eye_tensor,
```
Strilanc: You can pull this into a separate PR while we dither over the details of this one.

cirq/protocols/apply_unitary.py
```diff
@@ -85,28 +89,122 @@ def __init__(self,
...
+            self, indices: Iterable[int],
+            slices: Iterable[Union[int, slice, Sequence[int]]]
```
Strilanc: Make this argument optional and merge this method with the "for_operation_with_qid_shape" method.

cirq/protocols/apply_unitary.py
```diff
@@ -85,28 +89,122 @@ def __init__(self,
                 dtype as the target tensor.
             axes: Which axes the unitary effect is being applied to (e.g. the
                 qubits that the gate is operating on).
+
         """
         self.target_tensor = target_tensor
         self.available_buffer = available_buffer
         self.axes = tuple(axes)
+        self._is_subspace = _is_subspace
```
Strilanc: Describe _is_subspace

cirq/protocols/apply_unitary.py
```diff
@@ -69,10 +70,13 @@ class ApplyUnitaryArgs:
...
-                 target_tensor: np.ndarray,
-                 available_buffer: np.ndarray,
-                 axes: Iterable[int]):
+    def __init__(
+            self,
+            target_tensor: np.ndarray,
+            available_buffer: np.ndarray,
+            axes: Iterable[int],
+            *,  # Force keyword args
+            _is_subspace: bool = False):
```
Strilanc: It doesn't make sense for a constructor argument to be private.

cirq/protocols/apply_unitary.py
```diff
@@ -407,3 +527,41 @@ def apply_unitaries(unitary_values: Iterable[Any],
...
+        sub_result: The result of calling an object's `_apply_unitary_` method
+            on `sub_args`.  A transposed subspace of the desired result.
+
+    Returns: The full result tensor after applying the unitary.  Either
+        `args.target_tensor` or `args.available_buffer`.
+    """
+    if sub_result is sub_args.target_tensor:
+        return args.target_tensor
+    if sub_result is sub_args.available_buffer:
+        if sub_args._is_subspace:
```
Strilanc: I think this information should be provided as an argument to this method, instead of side channel information that comes along via the args object. Perhaps `for_operation_with_qid_shape` should return a tuple with a boolean flag indicating this?

Basically this method should have an argument like `assume_not_subspace` or similar.

cirq/protocols/apply_unitary.py
cduck: If I merge them, I'll loose some of the functionality of this method because the shape argument of `for_operation_with_qid_shape` is converted to `slices=[slice(0, s) for s in shape]`.

cirq/linalg/states.py
cduck: Done.

cirq/protocols/apply_unitary.py
cduck: Done.

cirq/protocols/apply_unitary.py
cduck: Done.

cirq/protocols/apply_unitary.py
cduck: Removed it.

cirq/__init__.py
cduck: I won't bother unless you want it for something.

cirq/protocols/apply_unitary.py
cduck: Removed `with_transposed_and_sliced_tensors`.

cirq/protocols/apply_unitary.py
cduck: _is_subspace is no longer needed.

cirq/protocols/apply_unitary.py
cduck: _is_subspace is no longer needed.

cirq/protocols/apply_unitary.py
cduck: I realized that `is_subspace` can just be derived just from `args.target_tensor` and `sub_args.target_tensor`.

cirq/protocols/apply_unitary.py
```diff
@@ -85,28 +85,138 @@ def __init__(self,
...
+        ordered_axes = (*other_axes, *sub_axes)
+        # Transpose sub_axes to the end of the shape and slice them
+        target_tensor = self.target_tensor.transpose(*ordered_axes)[(...,
+                                                                     *slices)]
+        available_buffer = self.available_buffer.transpose(*ordered_axes)[(
+            ..., *slices)]
+        new_axes = range(len(other_axes), len(ordered_axes))
+        return ApplyUnitaryArgs(target_tensor, available_buffer, new_axes)
+
+    def recover_result_from_sub_result(self,
```
Strilanc: How about `incorporate_result`? We could then also include logic about swapping the target and the buffer, which is otherwise specified in a few places.

Does it make sense for this method to be on this type, exposed to users? We could move it into a private place.

cirq/protocols/apply_unitary.py
```diff
@@ -85,28 +85,138 @@ def __init__(self,
...
+        copies it back into `self.target_tensor` or `self.available_buffer` as
+        necessary to return the result of applying the unitary to the full args.
+
+        Args:
+            sub_args: A version of `self` with transposed and sliced views of
+                it's tensors.
+            sub_result: The result of calling an object's `_apply_unitary_`
+                method on `sub_args`.  A transposed subspace of the desired
+                result.
+            assume_full_space: If True, less copying is done and the output is
```
Strilanc: This parameter isn't used

cirq/protocols/apply_unitary.py
```diff
@@ -85,28 +85,138 @@ def __init__(self,
...
+        num_qubits = len(qid_shape)
         state = linalg.one_hot(index=(0,) * num_qubits,
-                               shape=(2,) * num_qubits,
+                               shape=qid_shape,
                                dtype=np.complex128)
         return ApplyUnitaryArgs(state, np.empty_like(state), range(num_qubits))
 
-    def subspace_index(self, little_endian_bits_int: int
-                       ) -> Tuple[Union[slice, int, 'ellipsis'], ...]:
+    def for_operation_with_qid_shape(self, indices: Iterable[int],
```
Strilanc: Does it make sense for this method to be exposed to users? Will they ever call it correctly? We could make it private.

cirq/protocols/apply_unitary.py
```diff
@@ -304,27 +421,36 @@ def _strat_apply_unitary_from_unitary(unitary_value: Any, args: ApplyUnitaryArgs
...
-                                             matrix, [zero, one],
-                                             out=args.available_buffer)
-
-    # General case via np.einsum.
-    return linalg.targeted_left_multiply(matrix.astype(
-        args.target_tensor.dtype).reshape((2,) * (2 * len(args.axes))),
-                                         args.target_tensor,
-                                         args.axes,
-                                         out=args.available_buffer)
+    val_qid_shape = qid_shape_protocol.qid_shape(unitary_value,
```
Strilanc: See if you can extract a method here.

cduck: For these 4 lines?

cirq/protocols/apply_unitary.py
cduck: Made private.

cirq/protocols/apply_unitary.py
cduck: Removed.

cirq/protocols/apply_unitary.py
cduck: Renamed and made private and global.  This can't swap the target and the buffer because if you recursively use sub-arguments, swapping the target and the buffer messes up the second call to this.

PR #1777, opened 2019-07-04
[Qudits] Add new protocols: cirq.num_qubits, cirq.qid_shape

cduck:
For #933.  Part 1.

- `Gate.num_qubits` is deprecated in favor of the new `cirq.num_qubits` protocol (implement `Gate._num_qubits_` or `Gate._qid_shape_` on new gate or operation subclasses).
- `qid_shape` is only needed for non-qubit gates (e.g. qutrits where each qutrit can be in the |0>, |1>, or |2> state) and returns a tuple describing the dimension of each qid it operates on.


cirq/protocols/qid_shape_protocol.py
```diff
@@ -0,0 +1,168 @@
...
+RaiseTypeErrorIfNotProvided = (0,)  # type: Any
+# Equal integers outside the range [-5, 256] aren't identically equal with `is`.
+RaiseTypeErrorIfNotProvidedInt = -2**512  # type: Any
+
+TDefault = TypeVar('TDefault')
+
+
+class SupportsQidShape(Protocol):
+    """A unitary, channel, mixture or other object that operates on a known
+    number qubits/qudits/qids, each with a specific dimension."""
```
Strilanc: each with a specific number of quantum levels

cirq/protocols/qid_shape_protocol.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def _qid_shape_(self) -> Union[Tuple[int, ...], NotImplementedType]:
+        """A tuple specifying the dimension of each qid that this object
+        operates on, e.g. (2, 2) for a two-qubit gate.
```
Strilanc: All the 2s make this an ambiguous example. (2, 2, 2) for a three qubit gate would be less ambiguous.

cirq/protocols/qid_shape_protocol.py
```diff
@@ -0,0 +1,168 @@
...
+        with respect to the list of qubits that the gate is applied to. For
+        operations, the tuple is ordered to match the list returned by its
+        `qubits` attribute.
+
+        Returns:
+            A unitary matrix describing this value, or NotImplemented if the
+            shape is unknown.
+        """
+
+    def _num_qubits_(self) -> Union[bool, NotImplementedType]:
```
Strilanc: Return type annotation is wrong

cirq/__init__.py
```diff
@@ -326,6 +328,7 @@
     SupportsQasm,
     SupportsQasmWithArgs,
     SupportsQasmWithArgsAndQubits,
+    SupportsQidShape,
```
Strilanc: Should be two imports since there's two new protocols, right?

cirq/ops/raw_types_test.py
```diff
@@ -111,3 +130,145 @@ def _decompose_(self, qubits):
...
 def test_gate_algebra(expression, expected_result):
     assert expression == expected_result
+
+
+def test_gate_shape():
+
+    class ShapeGate(cirq.Gate):
+
+        def _qid_shape_(self):
+            return (1, 2, 3, 4)
```
Strilanc: Nit: unnecessary parens

cirq/protocols/qid_shape_protocol.py
```diff
@@ -0,0 +1,168 @@
...
+# case. It is checked for using `is`, so it won't have a false positive if the
+# user provides a different (0,) value.
+RaiseTypeErrorIfNotProvided = (0,)  # type: Any
+# Equal integers outside the range [-5, 256] aren't identically equal with `is`.
+RaiseTypeErrorIfNotProvidedInt = -2**512  # type: Any
+
+TDefault = TypeVar('TDefault')
+
+
+class SupportsQidShape(Protocol):
```
Strilanc: Rename to `SupportsExplicitQidShape`. We'll do similar renamings for the other protocols.

Strilanc: Define `SupportsExplicitNumQubits`

cirq/__init__.py
cduck: I consider this a single protocol with two magic methods.  Should I split them into two protocols?

cirq/ops/raw_types_test.py
cduck: I think it's clearer with parentheses, especially for `return (2,)`.

cirq/protocols/qid_shape_protocol.py
```diff
@@ -0,0 +1,168 @@
...
+        `cirq.qid_shape` if `_qid_shape_` is not defined.  If this
+        method is not present, or returns NotImplemented, it will fallback
+        to using the length of `_qid_shape_`.
+
+        Returns:
+            An integer specifying the number of qubits.
+        """
+
+
+def qid_shape(val: Any, default: TDefault = RaiseTypeErrorIfNotProvided
```
Strilanc: We should do something reasonable for "circuit-esque" values such as Iterable[Operation]. If the input doesn't have an explicit method, but is iterable and only contains operations (perhaps we also need a qubits protocol to avoid the type specificity), then the `qubits` and `qid_shape` methods should do the thing you'd expect `Circuit` to do.

cirq/__init__.py
cduck: Fixed.

cirq/protocols/qid_shape_protocol.py
cduck: Done.

cirq/protocols/qid_shape_protocol.py
cduck: I implemented `cirq.max_qid_shape(op_tree)` to replace `Circuit.max_qid_shape()`.

cirq/_compat.py
```diff
@@ -56,15 +58,18 @@ def deprecated(*, deadline: str, fix: str) -> Callable[[Callable], Callable]:
...
+                qualname = (func.__qualname__
+                            if func_name is None else func_name)
                 logging.warning(
                     'DEPRECATION\n'
                     'The function %s was used but is deprecated.\n'
                     'It will be removed in cirq %s.\n'
-                    '%s\n', func.__qualname__, deadline, fix)
+                    '%s\n', qualname, deadline, fix)
 
             return func(*args, **kwargs)
```
Strilanc: I added some lines below which set the doc property and such. Is that the purpose of `wraps`? Please resolve the silent merge conflict.

cirq/ops/gate_operation.py
```diff
@@ -96,6 +96,12 @@ def _group_interchangeable_qubits(self) -> Tuple[
     def _value_equality_values_(self):
         return self.gate, self._group_interchangeable_qubits()
 
+    def _qid_shape_(self):
+        return protocols.qid_shape(self.gate)
+
+    def _num_qubits_(self):
+        return protocols.num_qubits(self.gate)
```
Strilanc: `len(self._qubits)`

cirq/ops/op_tree.py
```diff
@@ -128,3 +131,17 @@ def freeze_op_tree(root: OP_TREE) -> OP_TREE:
         all internal nodes are tuples instead of arbitrary iterables.
     """
     return transform_op_tree(root, iter_transformation=tuple)
+
+
+def max_qid_shape(op_tree: OP_TREE,
+                  qubit_order: QubitOrderOrList = QubitOrder.DEFAULT,
+                  qubits_that_should_be_present: Iterable[Qid] = (),
+                  default_level: int = 1) -> Tuple[int, ...]:
```
Strilanc: Docstring

cirq/ops/raw_types_test.py
```diff
@@ -111,3 +130,145 @@ def _decompose_(self, qubits):
...
+            return 3
+
+    class DeprecatedGate(cirq.Gate):
+
+        def num_qubits(self):
+            return 3
+
+    shape_gate = ShapeGate()
+    assert cirq.qid_shape(shape_gate) == (1, 2, 3, 4)
+    assert cirq.num_qubits(shape_gate) == 4
```
Strilanc: Also check the `.num_qubits()` property

cirq/protocols/qid_shape_protocol.py
```diff
@@ -0,0 +1,173 @@
...
+            shape is unknown.
+        """
+
+
+class SupportsExplicitNumQubits(Protocol):
+    """A unitary, channel, mixture or other object that operates on a known
+    number of qubits."""
+
+    def _num_qubits_(self) -> Union[int, NotImplementedType]:
+        """The number of qubits this object operates on.
```
Strilanc: Mention that qubits means qudits.

cirq/_compat.py
cduck: Yes.  `wraps` sets `__module__`, `__name__`, `__qualname__`, `__annotations__` and `__doc__` (https://docs.python.org/3/library/functools.html#functools.wraps).

I'll change the lines below to just modify the `__doc__` copied by `wraps`.

cirq/protocols/qid_shape_protocol.py
```diff
@@ -0,0 +1,173 @@
...
+RaiseTypeErrorIfNotProvidedInt = -2**512  # type: Any
+
+TDefault = TypeVar('TDefault')
+
+
+class SupportsExplicitQidShape(Protocol):
+    """A unitary, channel, mixture or other object that operates on a known
+    number qubits/qudits/qids, each with a specific number of quantum levels."""
+
+    def _qid_shape_(self) -> Union[Tuple[int, ...], NotImplementedType]:
```
Strilanc: Do you think we'll ever want the ability to have the input and output shapes be different? We could define a `QidShape` or `OpShape` class...

cirq/protocols/qid_shape_protocol.py
```diff
@@ -0,0 +1,173 @@
...
+        `cirq.qid_shape` if `_qid_shape_` is not defined.  If this
+        method is not present, or returns NotImplemented, it will fallback
+        to using the length of `_qid_shape_`.
+
+        Returns:
+            An integer specifying the number of qubits, qudits or qids.
+        """
+
+
+def qid_shape(val: Any, default: TDefault = RaiseTypeErrorIfNotProvided
```
Strilanc: What do you think of renaming this to `operation_shape`? Are there any cases where non-operations have this shape? I guess result objects... nevermind.

PR #1779, opened 2019-07-08
Use v2 protos in cirq.google.Engine

wcourtney:
Update cirq Engine to optionally use new proto definitions in cirq.api.google.v2 when communicating with the API.

### DIFF
--- a/cirq/google/api/v2/results.py
+++ b/cirq/google/api/v2/results.py
@@ -148,6 +148,9 @@ def results_from_proto(
     Args:
         msg: v2 Result message to convert.
         measurements: List of info about expected measurements in the program.
+            This may be used for custom ordering of the result. If no
+            measurement config is provided, then all results will be returned
+            in the order specified within the result.
 
     Returns:
         A list containing a list of trial results for each sweep.
@@ -162,8 +165,21 @@ def results_from_proto(
 
 def _trial_sweep_from_proto(
         msg: result_pb2.SweepResult,
-        measurements: Dict[str, MeasureInfo] = None,
+        measure_map: Dict[str, MeasureInfo] = None,
 ) -> List[study.TrialResult]:
+    """Converts a SweepResult proto into List of list of trial results.
+
+    Args:
+        msg: v2 Result message to convert.
+        measure_map: A mapping of measurement keys to a mesurement configuration
+            containing qubit ordering. If no measurement config is provided,
+            then all results will be returned in the order specified within the
+            result.
+
+    Returns:
+        A list containing a list of trial results for the sweep.
+    """
+
     trial_sweep: List[study.TrialResult] = []
     for pr in msg.parameterized_results:
         m_data: Dict[str, np.ndarray] = {}
@@ -175,10 +191,9 @@ def _trial_sweep_from_proto(
                 if qubit in qubit_results:
                     raise ValueError('qubit already exists: {}'.format(qubit))
                 qubit_results[qubit] = unpack_bits(qmr.results, msg.repetitions)
-            if measurements:
+            if measure_map:
                 ordered_results = [
-                    qubit_results[qubit]
-                    for qubit in measurements[mr.key].qubits
+                    qubit_results[qubit] for qubit in measure_map[mr.key].qubits
                 ]
             else:
                 ordered_results = list(qubit_results.values())
--- a/cirq/google/api/v2/results.py
+++ b/cirq/google/api/v2/results.py
@@ -168,15 +168,18 @@ def _trial_sweep_from_proto(
     for pr in msg.parameterized_results:
         m_data: Dict[str, np.ndarray] = {}
         for mr in pr.measurement_results:
-            qubit_results: OrderedDict[devices.GridQubit, np.ndarray] = OrderedDict()
+            qubit_results: OrderedDict[devices.GridQubit, np.
+                                       ndarray] = OrderedDict()
             for qmr in mr.qubit_measurement_results:
                 qubit = devices.GridQubit.from_proto_id(qmr.qubit.id)
                 if qubit in qubit_results:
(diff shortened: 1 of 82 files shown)


cirq/google/engine/engine.py
```diff
@@ -174,11 +186,12 @@ def __init__(self,
...
         self.default_gcs_prefix = default_gcs_prefix
+        self.proto_version = proto_version
 
         discovery_service_url = (
-            self.discovery_url if self.api_key is None else (
-                "%s&key=%s" % (self.discovery_url, urllib.parse.quote_plus(
-                               self.api_key))))
+            self.discovery_url if self.api_key is None else
+            ("%s&key=%s" %
+             (self.discovery_url, urllib.parse.quote_plus(self.api_key))))
```
maffoo: Consider breaking this up for readability:
```python
discovery_service_url = self.discovery_url
if self.api_key is not None:
    discovery_service_url += '&key={}'.format(self.api_key)
```

cirq/google/engine/engine.py
```diff
@@ -612,14 +691,13 @@ def _sweepable_to_sweeps(sweepable: Sweepable) -> List[Sweep]:
...
 
 
 def _resolver_to_sweep(resolver: ParamResolver) -> Sweep:
-    return Zip(*[Points(key, [value]) for key, value in
-                 resolver.param_dict.items()]) if len(
-        resolver.param_dict) else UnitSweep
+    return Zip(
+        *[Points(key, [value])
+          for key, value in resolver.param_dict.items()]) if len(
+              resolver.param_dict) else UnitSweep
```
maffoo: Could use an early return for readability:
```python
params = resolver.param_dict
if not params:
    return UnitSweep
return Zip(*[Points(k, [v]) for k, v in params.items()])
```

cirq/google/api/v2/results.py
```diff
@@ -138,10 +139,9 @@ def results_to_proto(
                     qmr.results = pack_bits(m_data[:, i])
     return out
 
-
 def results_from_proto(
         msg: result_pb2.Result,
-        measurements: List[MeasureInfo],
+        measurements: List[MeasureInfo] = None,
```
maffoo: I think instead of making this optional and relying on ordering, we should use the `api_v2.find_measurements` function to find measurements in the program before sending to the quantum engine, and store this in the `EngineJob`. Then when getting results the list of measurements can be passed in to `results_from_proto`.

cirq/google/engine/engine.py
```diff
@@ -204,6 +218,8 @@ def run(
             repetitions: The number of repetitions to simulate.
             priority: The priority to run at, 0-100.
             processor_ids: The engine processors to run against.
+            gate_set: The gate set used to serialize the circuit. The gate set
+                must be supported by the selected processor
```
dabacon: nit: "." at end of comment

cirq/google/engine/engine.py
```diff
@@ -392,11 +407,53 @@ def run_sweep(
...
+
+        program_dict = {}  # type: Dict[str, Any]
+        program_dict['@type'] = 'type.googleapis.com/' + descriptor.full_name
+        program_dict['parameter_sweeps'] = [
+            sweep_to_proto_dict(sweep, repetitions) for sweep in sweeps
+        ]
+        program_dict['operations'] = [
+            op for op in schedule_to_proto_dicts(schedule)
+        ]
+        return program_dict, None  # run context included in program
```
dabacon: is there a reason to do this for v1 protos?  It would be nice to always just the the run context.

cirq/google/api/v2/results.py
dabacon: Shouldn't all the info about the measurements be accessible from the results?

wcourtney: What is your concern with relying on the order in the result? If we want to support re-ordering, could it be a feature/utility rather than a requirement to parse a result?

cirq/google/engine/engine.py
wcourtney: We previously talked about making these changes separately, so I left it as-is. It should be straightforward, so I'll address this after attending to the other comments.

cirq/google/api/v2/results.py
maffoo: We want to be sure that the order in which results appear in `TrialResult` match the original measurement operation. The `measurements` arg on `results_from_proto` is here so we can check the qubit order in the results against the original measure ops, and it's as easy to reorder as to check so that's what it does. If we want to specify that the engine will not reorder (and then assume but not check this), then I'd suggest we remove this argument entirely rather than making it optional.

I think it would be good to include the qubit order in `TrialResult` itself to make the results more self-describing. That would alleviate some of my concern because we won't have so many assumptions about parallel ordering between various objects. I know `TrialResult` is getting an overhaul, so maybe we can include the qubit order as part of that change.

wcourtney: +1 to including qubit order details in the result.

I'd prefer to remove the `measurements` arg, but that would be a breaking change for existing external dependencies (I found one). My plan was to make this optional to continue support, then, if it was acceptable, remove all references, then clean up by removing the param. LMK if it's fine to just make the breaking change, otherwise I can add comments to that effect.

cirq/google/engine/engine.py
dabacon: SGTM.

wcourtney: Added. PTAL.

dabacon:
I'm fine with the None if we eventually migrate it away in external code.  @maffoo ?

cirq/google/api/v2/results.py
maffoo: I agree with @dabacon, making this optional for now is fine. We can remove in a future PR.

PR #1785, opened 2019-07-10
More verbose output when gate fails to be serialized

mpharrigan:


mpharrigan:
I'm reformatting under protest. I think the way I have it is much nicer

cirq/google/serializable_gate_set.py
```diff
@@ -127,7 +127,8 @@ def serialize_op(self,
                     proto_msg = serializer.to_proto(gate_op, msg)
                     if proto_msg is not None:
                         return proto_msg
-        raise ValueError('Cannot serialize op of type {}'.format(gate_type))
+        raise ValueError('Cannot serialize op {!r} of type {}'.format(
```
Strilanc: Nit: Generally you want to put reprs at the end of the error message, because they can be long.

PR #1786, opened 2019-07-10
Circuit optimization support for IdentityGate

mpharrigan:
- IdentityGate to support _phase_by_
 - DropIdentities optimization

### DIFF
--- a/cirq/__init__.py
+++ b/cirq/__init__.py
@@ -197,6 +197,7 @@ from cirq.ops import (
 from cirq.optimizers import (
     ConvertToCzAndSingleGates,
     DropEmptyMoments,
+    DropIdentities,
     DropNegligible,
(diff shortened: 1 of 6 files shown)


mpharrigan:
lemme add some tests

cirq/__init__.py
```diff
@@ -197,6 +197,7 @@
 from cirq.optimizers import (
     ConvertToCzAndSingleGates,
     DropEmptyMoments,
+    DropIdentities,
```
Strilanc: I think this should just be part of `DropNegligible`. Give `IdentityGate` a `def _trace_distance_bound_(self): return 0` and it should work.

dabacon:
Should we close this?  Identity now has trace distance 0.

mpharrigan:
Yes

PR #1788, opened 2019-07-10
Don't auto-replace PhasedXPowGate instances with XPowGate or YPowGate

Strilanc:


cirq/ops/phased_x_gate.py
```diff
@@ -211,5 +189,20 @@ def _canonical_exponent(self):
...
 
+    def _value_equality_values_cls_(self):
+        if self.phase_exponent == 0:
+            return common_gates.XPowGate
+        if self.phase_exponent == 0.5:
+            return common_gates.YPowGate
+        return PhasedXPowGate
+
     def _value_equality_values_(self):
+        if self.phase_exponent == 0:
```
dabacon: does this delegation really provide value?

mpharrigan:
1) This would make my life a lot easier :) so big +1 from me
2) I'm also mildly confused about the value equality special casing. Could you motivate this a bit more?

Strilanc:
I special cased the equality because not doing so broke many, many tests. I believe all those tests *could* be rewritten, but it was tricky to do so.

dabacon:
Let's open an issue then for this then, since I think it seems strickly not necessary.

PR #1790, opened 2019-07-10
 Refactor cirq.apply_unitary to use list-of-strategies approach

Strilanc:


cirq/protocols/apply_unitary.py
```diff
@@ -131,7 +130,7 @@ def subspace_index(self, little_endian_bits_int: int
                                                 little_endian_bits_int)
 
 
-class SupportsApplyUnitary(Protocol):
+class SupportsConsistentApplyUnitary(Protocol):
```
mpharrigan: The fact that this name change does not affect any other part of the code base makes me have to ask: where is this class actually used? Are things supposed to inherit from it. If so, why don't they? Is this class only for documentation?

Strilanc: Correct, it's only for documentation. Technically you could use it in mypy.

Issue #1289, opened 2019-01-10
Enable Entanglement Entropy Calculations in Cirq


karlunho:
R

@dabacon Can you take a look if you think this is something we should put in ?

michaelstreif:
Calculating the entanglement entropy gives insight into the mechanisms of the quantum calculations and is therefore an important quantity to find useful quantum algorithms. It moreover helps to estimate the required resources for classical simulations when using TNT.

I think this would be of help for research and simulations, but not really when using the real device. We already  have functions calculating the entanglement entropy, given the output of the cirq simulator.

dabacon:
I think adding support for this would be good.  It is definitely useful for simulations.  We already have the beginnings of denisty matrices appearing in Cirq.

dabacon:
This is implemented in `cirq.von_neumann_entropy`

PR #1791, opened 2019-07-10
Define von Neumann entropy

vtomole:
This is a response to: https://github.com/quantumlib/Cirq/pull/1753
Part of: https://github.com/quantumlib/Cirq/issues/1289


cirq/sim/density_matrix_utils.py
```diff
@@ -257,3 +257,16 @@ def _validate_indices(num_qubits: int, indices: List[int]) -> None:
...
+def von_neumann_entropy(density_matrix: np.ndarray) -> float:
+    """Calculates von Neumann entropy of density matrix in bits.
+    Args:
+        density_matrix: The density matrix.
+    Returns:
+        The calculated von Neumann entropy.
+    """
+    eigenvalues = np.linalg.eigvals(density_matrix)
+    if 0 in eigenvalues:
+        return 0
```
Strilanc: This is like saying that a die roll has no entropy because it can never roll a 7. You're not multiplying by zero when you see zero, you're adding zero when you see zero.

```
    def h(p: float):
        p = abs(p)
        if p == 0:
            return 0
        return -p * math.log2(p)

    return sum(h(p) for p in np.linalg.eigvals(density_matrix))
```

cirq/sim/density_matrix_utils.py
```diff
@@ -257,3 +257,16 @@ def _validate_indices(num_qubits: int, indices: List[int]) -> None:
...
+    return abs(np.sum(eigenvalues * np.log2(eigenvalues)))
```
Strilanc: You want to `abs` the eigenvalues before passing them through the `p log p` function.

cirq/sim/density_matrix_utils.py
vtomole: Makes sense. I haven't been able to find anything online that mentions this "nuance" on the definition of entropy. Everything I've found just gives the formula. How is one supposed to know what to do when an eigenvalue is 0? Any pointers? Thanks.

CCing @bryano cause i discussed this with him too.

cirq/sim/density_matrix_utils.py
bryano: The eigenvalues of the density matrix should be non-negative reals so you shouldn't have to take their absolute values.

Also, we should just negate the expression rather than take its absolute value.

cirq/sim/density_matrix_utils.py
bryano: Like Craig said in the sync, this is because the limit as p -> 0 of p log p is 0. One way of seeing this is to write p log p as log p / (1 / p) and apply L'Hospital's rule.

At an operational level, this also has to be the case because events of zero probability shouldn't affect the information content of the distribution.

bryano:
We could just use `scipy.stats.entropy`...

vtomole:
@bryano Scipy can't take complex eigenvalues :( https://github.com/scipy/scipy/blob/492e7f8e95ef23754b74731ca1af34a4e4288602/scipy/special/_convex_analysis.pxd#L8

cirq/sim/density_matrix_utils.py
Strilanc: If you perturb the density matrix (due to floating point error), you may perturb the eigenvalues off of the real line. This clamps them back.

vtomole:
@bryano Unfortunately, `scipy.stats.entropy` uses `log`  instead of `log2`.

CC: @Strilanc

vtomole:
@bryano That will work!

cirq/sim/density_matrix_utils_test.py
```diff
@@ -26,6 +26,37 @@ def assert_valid_density_matrix(matrix, num_qubits=1):
                                      dtype=matrix.dtype), matrix)
 
 
+def test_von_neumann_entropy():
```
bryano: Maybe test some maximally mixed states?

vtomole: Done.

cirq/sim/density_matrix_utils.py
```diff
@@ -257,3 +258,15 @@ def _validate_indices(num_qubits: int, indices: List[int]) -> None:
...
+
+def von_neumann_entropy(density_matrix: np.ndarray) -> float:
+    """Calculates von Neumann entropy of density matrix in bits.
+    Args:
+        density_matrix: The density matrix.
+    Returns:
+        The calculated von Neumann entropy.
+    """
+    eigenvalues = np.linalg.eigvals(density_matrix)
+    print(eigenvalues)
```
bryano: leftover debug line

cirq/__init__.py
```diff
@@ -249,6 +249,7 @@
     StepResult,
     to_valid_density_matrix,
     to_valid_state_vector,
+    von_neumann_entropy,
```
bryano: nit: alphabetical ordering

cirq/sim/density_matrix_utils.py
vtomole: Done.

cirq/__init__.py
vtomole: Done.

cirq/sim/density_matrix_utils.py
Strilanc: Still there.

Strilanc: Ah nevermind was viewing stale commit

cirq/sim/density_matrix_utils_test.py
```diff
@@ -26,6 +26,38 @@ def assert_valid_density_matrix(matrix, num_qubits=1):
...
+    # linear combination of the two with coefficients 0.1 and 0.9
+    res = cirq.testing.random_unitary(2)
+    first_column = res[:, 0]
+    first_density_matrix = 0.1 * np.outer(first_column, np.conj(first_column))
+    second_column = res[:, 1]
+    second_density_matrix = 0.9 * np.outer(second_column,
+                                           np.conj(second_column))
+    assert np.isclose(
+        cirq.von_neumann_entropy(first_density_matrix + second_density_matrix),
+        0.4689, 1e04)
```
Strilanc: Use explicit `atol=` for tolerance arguments.

cirq/sim/density_matrix_utils_test.py
```diff
@@ -26,6 +26,38 @@ def assert_valid_density_matrix(matrix, num_qubits=1):
...
+    # 3x3 state
+    assert np.isclose(
+        cirq.von_neumann_entropy(
+            np.array(
+                np.array([[0.5, 0.5j, 1], [-0.5j, 0.5, 0], [0.7, 0.4, 0.6]]))),
+        1.9, 1e01)
+    # 4X4 state
+    assert np.isclose(
+        cirq.von_neumann_entropy(
+            np.array([[0.5, 0.5j, 1, 3], [-0.5j, 0.5, 0, 4], [0.7, 0.4, 0.6, 5],
```
Strilanc: Use `# yapf: disable` and `# yapf: enable` to keep these matrices lined up.

cirq/sim/density_matrix_utils_test.py
```diff
@@ -26,6 +26,38 @@ def assert_valid_density_matrix(matrix, num_qubits=1):
...
+    first_column = res[:, 0]
+    first_density_matrix = 0.1 * np.outer(first_column, np.conj(first_column))
+    second_column = res[:, 1]
+    second_density_matrix = 0.9 * np.outer(second_column,
+                                           np.conj(second_column))
+    assert np.isclose(
+        cirq.von_neumann_entropy(first_density_matrix + second_density_matrix),
+        0.4689, 1e04)
+
+
```
Strilanc: Test the entropy of `diag([0, 0, 0.1, 0, 0.2, 0.3, 0.4, 0])`.

cirq/sim/density_matrix_utils_test.py
vtomole: Done.

cirq/sim/density_matrix_utils_test.py
vtomole: Done.

cirq/sim/density_matrix_utils_test.py
vtomole: Done.

vtomole:
@Strilanc Thanks for your patience on this and thanks @bryano for clearing things up for me.

Issue #1610, opened 2019-05-08
Apply noise model to circuit


dabacon:
I expected to be able to take a `NoiseModel` and apply it to a circuit.

```noise = cirq.ConstantQubitNoiseModel(cirq.depolarize(0.2))```
Then I'd like to be able to
```circuit.add_noise(noise)```
or 
```noise.apply_noise(circuit)```

**Comment by dabacon** [comment]
Of course there is some vagueness on the resulting circuit, for example when doing this for noisy_moment into which moments do you add the resulting op tree.

**Comment by dabacon** [comment]
Currently I do this via code like
```
noise = cirq.ConstantQubitNoiseModel(cirq.depolarize(0.2))
circuit = cirq.Circuit.from_ops(cirq.H(a), cirq.CNOT(a, b), cirq.measure(a, b))
print('Circuit with no noise:\n{}\n'.format(circuit))

noisy_circuit = cirq.Circuit()
for moment in circuit:
    noisy_circuit.append(noise.noisy_moment(moment, sorted(circuit.all_qubits())))
print('Circuit with noise:\n{}'.format(noisy_circuit))
```

```
Circuit with no noise:
a: ───H───@───M───
          │   │
b: ───────X───M───

Circuit with noise:
a: ───H────────D(0.2)───@───D(0.2)───M───D(0.2)───
                        │            │
b: ───D(0.2)────────────X───D(0.2)───M───D(0.2)───
```

**Comment by peterse** [comment]
one way to handle the vagueness could be to define `noise.apply_noise` with a conditional function kwarg to determine where the noise goes, e.g.
```
def apply_noise(self, circuit, cond = lambda x: True):
    noisy_circuit = cirq.Circuit()
    if hasattr(self, noisy_moment):
         for moment in circuit:
             if cond(moment):
                 noisy_circuit.append(noise.noisy_moment(moment, sorted(circuit.all_qubits())))
  . . . 
```
then you might do something like
```
noisy_moments_list = circuit.moments[3:5]
noise.apply_noise(circuit, lambda x: x in noisy_moments_list)
```
Alternatively, this might belong to a completely different class of NoiseModel where the user overrides `noisy_moment` with that conditional built in...

c-poole:
This vagueness stems at least in part from a lack of default behavior in `NoiseModel`. We have a clear hierarchy of potential methods to use:
`noisy_circuit` which can access the circuits sequence of moments and device
`noisy_moments` which only accesses a sequence of moments
[...]

PR #1792, opened 2019-07-10
**Proposal** implements `NoiseModel.apply_noise`

peterse:
**not ready for merge yet; please just review logic in apply_noise method**

After playing around with noise models a bit, I've implemented `apply_noise` with the following logic in mind:

1.  Users should be able to implement noise models on a circuit-, moment-, and operation- level of specification
2. Invoking separate models that each implement one of `noisy_moments`, `noisy_moment`, `noisy_operation` on a single circuit leads to ambiguity about whether the operations within a given circuit were the original operations intended by the user, or added as noise by one of the intermediate noise models
3. therefore, each `noisy_*` method of `NoiseModel` should append/insert gates that correspond to _different_ realizations of noise in the circuit.

This removes the need for abstraction in the NoiseModel baseclass.

In `noise_model_test.py` I have defined a few semi-realistic noise models that would justify the use of each `noisy_*` method, and also tried to show the difference between how noise models stack with / without the behavior described above.

@Strilanc @c-poole @dabacon Input on this implementation or suggestions are welcome

### DIFF
--- a/cirq/devices/noise_model.py
+++ b/cirq/devices/noise_model.py
@@ -15,6 +15,7 @@
 from typing import TYPE_CHECKING, Iterable, Sequence, Union
 
 from cirq import ops, value
+from cirq.circuits import circuit
 
 if TYPE_CHECKING:
     # pylint: disable=unused-import
@@ -119,6 +120,24 @@ class NoiseModel(metaclass=value.ABCMetaImplementAnyOneOf):
             An OP_TREE corresponding to the noisy operations implementing the
             noisy version of the given operation.
         """
+
+    def apply_noise(self, circuit: 'cirq.Circuit') -> 'cirq.Circuit':
+        """Compose a circuit augmenting the input with defined noise methods.
+
+        If multiple noise methods are defined, they will be inserted in order
+        of finest granularity i.e. `noisy_operation` operations before
+        `noisy_moment` operations before `noisy_moments` operations.
+
+        TODO
+        """
+
+        # Append to front of op trees within a queue to avoid in place changes
+        op_tree = self.noisy_moments(circuit)
+        for i, moment in enumerate(circuit):
+            op_tree[i] = self.noisy_moment(moment) + op_tree[i]
+            for op in moment:
+                op_tree[i].insert(0, self.noisy_operation(op))
+        return circuit.Circuit.from_ops(op_tree)
 
 
 @value.value_equality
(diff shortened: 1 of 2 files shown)


cirq/devices/noise_model.py
```diff
@@ -70,21 +51,8 @@ def noisy_moments(self, moments: 'Iterable[cirq.Moment]',
             A sequence of OP_TREEs, with the k'th tree corresponding to the
             noisy operations for the k'th moment.
         """
+        return list(moments)
```
Strilanc: Why was this change made? The purpose of the alternatives is that the user can implement any one of the methods and the rest will be derived for them. Now if the user implements one method the others will have no noise, which is a bug.

cirq/devices/noise_model.py
```diff
@@ -119,6 +76,25 @@ def noisy_operation(self, operation: 'cirq.Operation') -> 'cirq.OP_TREE':
             An OP_TREE corresponding to the noisy operations implementing the
             noisy version of the given operation.
         """
+        return operation
+
+    def apply_noise(self, input_circuit: 'cirq.Circuit') -> 'cirq.Circuit':
```
Strilanc: Rename this to `noisy_circuit_moments` and have it return `self.noisy_moments(circuit, sorted(circuit.all_qubits()))`.

We should avoid having a `noisy_circuit` method that returns Circuit, because it is incapable of maintaining the moment structure.

cirq/devices/noise_model.py
peterse: The intention here is that `noisy_operation` and `noisy_moment` generate different noise behaviors. For example, the `noise_model_test.MarkovianPairwiseCorrelatedNoise` is applied only at the moment level, affecting each operation in a given moment. The default `noisy_operation` function returns input operations unchanged because the class defining only `noisy_moment` implied that the user didn't have operation-specific noise in mind.

The reasoning here is that to apply different noise models at different levels of the operation-moment-moments hierarchy they must exist within a single model. Otherwise, subsequent noise models might apply their noise to operations that were introduced to previous noise models. This is what I'm demonstrating in `noise_model_test.test_composition_of_noise_models`

cirq/devices/noise_model.py
peterse: The goal here was to address #1610 - how to get `noisy_*` methods into a circuit.

The original default implementations of `noisy_moments`, `noisy_moment`, and `noisy_operation` have been wrapped up into this loop

cirq/devices/noise_model.py
peterse: of course, this is just the structure I'm proposing. There might be better ways to deal with applying multiple noise models without applying noise operations on top of noise operations.

peterse:
closing this and migrating feature suggestions to #1835

PR #1796, opened 2019-07-10
Change pytest.raises to use match instead of message.

kunalq:
`message` is deprecated in newer versions of pytest; update `pytest.raises` calls to use `match` instead. As far as I can tell, none of these are regular expressions and therefore do not need to be raw strings; however, I can be convinced otherwise.


Strilanc:
How did this break tests? Very strange... maybe message was doing nothing?

kunalq:
@Strilanc looks like message was not, in fact, doing anything. The exception text did not match the string; I've updated it to a regex. Also fixed formatting.

kunalq:
Also, getting type check errors that seem unrelated to this PR, but only when I run `./continuous-integration/check.sh`. If I run `./check/mypy`, I get no output.

```
cirq/contrib/acquaintance/permutation.py:38: error: Cannot determine type of 'SWAP'
cirq/contrib/acquaintance/permutation.py:92: error: Cannot determine type of 'SWAP'
cirq/contrib/acquaintance/permutation.py:112: error: Cannot determine type of 'SWAP'
cirq/ops/pauli_gates.py:89: error: Cannot determine type of '_num_qubits_' in base class 'Gate'
cirq/ops/pauli_gates.py:95: error: Cannot determine type of '_num_qubits_' in base class 'Gate'
cirq/ops/pauli_gates.py:101: error: Cannot determine type of '_num_qubits_' in base class 'Gate'
cirq/ops/linear_combinations.py:369: error: Module has no attribute "linear_dict"
Finished typecheck
```

They look [green on master](https://github.com/quantumlib/Cirq/runs/167986252), which suggests that I'm running a different version of a dev requirement perhaps? I validated that git log shows the right master branch (i.e. I can see [this commit](https://github.com/quantumlib/Cirq/commit/e73c9bef672e83143ab04e7f169988149055d630) in the logs).

Looking at it closely: it's having trouble inferring the type of this gate:
```python
class SwapPowGate(eigen_gate.EigenGate,
                  gate_features.TwoQubitGate,
                  gate_features.InterchangeableQubitsGate):
```

I haven't gone in-depth on mypy re: mixins, but I'm surprised that this is causing issues.

FWIW, I do have `mypy` version `0.701` installed, which matches the pip dev list entry.

kunalq:
@Strilanc what's the procedure for handling noise in the performance tests? Should I keep re-running until it passes? :)

vtomole:
Looks like there is a performance regression in `test_example_runs_bell_inequality_perf`. Reran and it's fine now. The performance check "might" be finicky. I'll create an issue if it happens again.

cduck:
This also happened in #1771.

kunalq:
Thanks, Victory!

PR #1797, opened 2019-07-10
s/v1/v2 for metrics proto.

wcourtney:
Updated package for metrics.proto from v1 to v2 and re-ran proto compiler on the directory, which updated some of the generated code for other protos as well.

### DIFF
--- a/cirq/api/google/v2/BUILD
+++ b/cirq/api/google/v2/BUILD
@@ -20,7 +20,7 @@ proto_library(
 
 proto_library(
     name = "run_context_proto",
-    srcs = ["run_contex.proto"],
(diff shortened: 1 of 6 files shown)


maffoo:
Note: we now have https://github.com/quantumlib/Cirq/blob/master/dev_tools/build-protos.sh for generating protobuf code. That script uses the version of protoc defined in the dev dependencies in the repo, so doesn't depend on protoc being installed on the user's machine, and also updates the mypy stubs. We should probably add a CI check that runs the build script and checks that there are no diffs.

dabacon:
Created #1801 for this

PR #1810, opened 2019-07-12
Implement TrialResult.__eq__ using pandas.DataFrame.equals

Strilanc:
- Can't use _value_equality_values_ because pandas.DataFrame.__eq__ broadcasts instead of returning a bool
- This is much more efficient than using the repr

I believe the test failures are due to having series within the table. So blocked by https://github.com/quantumlib/Cirq/pull/1811


cirq/study/trial_result.py
```diff
@@ -266,5 +265,7 @@ def _repr_pretty_(self, p: Any, cycle: bool) -> None:
     def __str__(self):
         return _keyed_repeated_bitstrings(self.data)
 
-    def _value_equality_values_(self):
-        return repr(self.data), self.params
+    def __eq__(self, other):
+        if not isinstance(other, type(self)):
+            return NotImplemented
+        return self.data.equals(other.data) and self.params == other.params
```
cduck: `data.equals` doesn't seem to be returning a bool like it should.

Strilanc: I believe this is because there are pd.Series items in the table. That's fixed in a different PR; this one is blocked.

PR #1811, opened 2019-07-12
Optimize TrialResult.data

Strilanc:
- Lazily compute the dataframe when it is needed
- Flatten bits into ints, instead of a series

### DIFF
--- a/cirq/study/trial_result.py
+++ b/cirq/study/trial_result.py
@@ -15,8 +15,8 @@
 """Defines trial results."""
 
 from typing import (
-    Iterable, Callable, Tuple, TypeVar, Dict, Any, TYPE_CHECKING, Union
-)
+    Iterable, Callable, Tuple, TypeVar, Dict, Any, TYPE_CHECKING, Union,
+    List, Optional)
 
 import collections
 import numpy as np
@@ -66,15 +66,34 @@ def _big_endian_int(bits: Iterable[Any]) -> int:
     return result
 
 
+def _big_endian_bits(val: int, bit_count: int) -> List[bool]:
+    """Returns the big-endian bits of an integer.
+
+    For example, 18 becomes [True, False, False, True, False] when `bit_count`
+    is five.
+
+    Args:
+        val: The integer to get bits from.
+        bit_count: The number of desired bits in the result.
+
+    Returns:
+        The bits.
+    """
+    return [bool(val & (1 << i)) for i in range(bit_count)[::-1]]
+
+
 def _bitstring(vals: Iterable[Any]) -> str:
     return ''.join('1' if v else '0' for v in vals)
 
 
-def _keyed_repeated_bitstrings(measurements: pd.DataFrame) -> str:
+def _keyed_repeated_bitstrings(vals: Dict[str, np.ndarray]
+                               ) -> str:
     keyed_bitstrings = []
-    for key in sorted(measurements.columns):
-        reps = pd.DataFrame(measurements[key].to_list())
-        all_bits = ', '.join(reps.apply(_bitstring, axis=0))
+    for key in sorted(vals.keys()):
+        reps = vals[key]
+        n = 0 if len(reps) == 0 else len(reps[0])
+        all_bits = ', '.join([_bitstring(reps[:, i])
+                              for i in range(n)])
         keyed_bitstrings.append('{}={}'.format(key, all_bits))
     return '\n'.join(keyed_bitstrings)
 
@@ -85,13 +104,6 @@ def _key_to_str(key: TMeasurementKey) -> str:
     if isinstance(key, ops.Qid):
         return str(key)
     return ','.join(str(q) for q in key)
-
-
-def _to_dict(measurements: pd.DataFrame) -> Dict[str, np.ndarray]:
-    repr_dict = {}
-    for key in sorted(measurements.columns):
-        repr_dict[key] = np.array(measurements[key].to_list())
-    return repr_dict
 
 
 @value.value_equality(unhashable=True)
@@ -126,19 +138,24 @@ class TrialResult:
                 measurements.
         """
         self.params = params
+        self._data: Optional[pd.DataFrame] = None
+        self._measurements = measurements
 
-        # Convert to a DataFrame with columns as measurement keys, rows as
-        # repetitions and a Series of measurements for a particular key and
-        # repetition as the value.
-        converted_dict = {}
-        for key, val in measurements.items():
-            converted_dict[key] = [pd.Series(m_vals) for m_vals in val]
-        self.data = pd.DataFrame(converted_dict)
+    @property
+    def data(self):
+        if self._data is None:
+            # Convert to a DataFrame with columns as measurement keys, rows as
+            # repetitions and a big endian integer for individual measurements.
+            converted_dict = {}
+            for key, val in self._measurements.items():
+                converted_dict[key] = [_big_endian_int(m_vals) for m_vals in
+                                       val]
+            self._data = pd.DataFrame(converted_dict)
+        return self._data
 
-    # Keep the old instance variables for test compatibility.
     @property
     def measurements(self) -> Dict[str, np.ndarray]:
-        return _to_dict(self.data)
+        return self._measurements
 
     @property
     def repetitions(self) -> int:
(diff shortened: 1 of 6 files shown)


smitsanghavi:
> * Flatten bits into ints, instead of a series
This would be incompatible with Qudits in general by forcing to binary. Is that not a concern?

Strilanc:
It is not incompatible, it's just changing the basis.

For example, suppose you have shape [2, 3, 4, 3, 2] and results [1, 2, 0, 1, 0]. Then the integer value should be

```
+0 * 1
+1 * 2
+0 * 2*3
+2 * 2*3*4
+1 *2*3*4*3
= 122
```

cirq/study/trial_result.py
```diff
@@ -66,15 +66,34 @@ def _big_endian_int(bits: Iterable[Any]) -> int:
...
+    Args:
+        val: The integer to get bits from.
+        bit_count: The number of desired bits in the result.
+
+    Returns:
+        The bits.
+    """
+    return [bool(val & (1 << i)) for i in range(bit_count)[::-1]]
+
+
```
cduck: This should be a utility method that works with arbitrary `qid_shape`.

cirq/study/trial_result.py
```diff
@@ -126,19 +138,24 @@ def __init__(
...
 
-        # Convert to a DataFrame with columns as measurement keys, rows as
-        # repetitions and a Series of measurements for a particular key and
-        # repetition as the value.
-        converted_dict = {}
-        for key, val in measurements.items():
-            converted_dict[key] = [pd.Series(m_vals) for m_vals in val]
-        self.data = pd.DataFrame(converted_dict)
+    @property
+    def data(self):
```
cduck: Return type: `pd.DataFrame`

cirq/study/trial_result.py
```diff
@@ -195,11 +212,14 @@ def multi_measurement_histogram(  # type: ignore
             A counter indicating how often measurements sampled various
             results.
         """
-        fixed_keys = [_key_to_str(key) for key in keys]
-        samples = self.data[fixed_keys]
+        fixed_keys = tuple(_key_to_str(key) for key in keys)
+        samples = zip(*[self.measurements[sub_key]
+                        for sub_key in fixed_keys])  # type: Iterable[Any]
```
cduck: *[...] -> *(...)

cirq/study/trial_result.py
```diff
@@ -66,15 +65,32 @@ def _big_endian_int(bits: Iterable[Any]) -> int:
     return result
 
 
+def _big_endian_bits(val: int, bit_count: int) -> List[bool]:
```
maffoo: I hate to reopen the endianness debate, but just curious what was the thinking to go with big-endian here? I think it at least merits duscussion, since we use little-endian bit order in internal code.

Strilanc: I chose big endian to satisfy the following invariant:

```python
def invariant(a: np.ndarray, ks: Tuple[int, ...]):
    b = a.reshape(a.size)
    k = bits_to_int(ks)
    assert a[ks] == b[k]
```

If `bits_to_int` is a little endian conversion, the assertion fails. If it is big endian, the assertion succeeds.

In other words, we're big endian because numpy is big endian. Later indices in a numpy array have smaller strides in the same way that later big endian digits have smaller weights.

maffoo: > Later indices in a numpy array have smaller strides in the same way that later big endian digits have smaller weights.
This is true only for "C-contiguous" arrays; "Fortran-contiguous" arrays go the other way, which you could get by doing
```python
b = a.reshape(a.size, order='F')
```
in your example. So numpy supports both in a fairly straightforward way, but I guess it's good to stick with the default.

Strilanc: You can also create similar invariants around merging indices for the result of np.kron.

If it comes down to it, we can add order='F' at some later date. For now this will do.

cirq/study/trial_result.py
Strilanc: Done. Extracted global utility methods taking a "base" argument that can be a list or an int.

cirq/study/trial_result.py
Strilanc: Done.

cirq/study/trial_result.py
Strilanc: Done.

cirq/value/digits.py
```diff
@@ -0,0 +1,195 @@
...
+        >>> big_endian_int_to_digits(11, base=[2, 3, 4])
+        [0, 1, 0, 2]
+
+        >>> big_endian_int_to_digits(18, bit_count=8)
+        [0, 0, 0, 1, 0, 0, 1, 0]
+
+        >>> big_endian_int_to_digits(18, bit_count=4)
+        [0, 0, 1, 0]
+
+        >>> big_endian_int_to_digits(-3, bit_count=4)
```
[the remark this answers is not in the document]
cduck: And this would throw 'Out of range'.

cirq/value/digits.py
```diff
@@ -0,0 +1,195 @@
...
+                'No digit count. Provide `digit_count` when base is an int.')
+        base = (base,) * digit_count
+    elif digit_count is None:
+        digit_count = len(base)
+
+    if len(base) != digit_count:
+        raise ValueError('Inconsistent digit count. len(base) != digit_count')
+
+    result = []
+    for b in base[::-1]:
```
cduck: `for b in reversed(base):`

cirq/value/digits_test.py
```diff
@@ -0,0 +1,61 @@
...
+                                       bit_count=8) == [0, 0, 0, 1, 0, 0, 1, 0]
+    assert cirq.big_endian_int_to_bits(18, bit_count=4) == [0, 0, 1, 0]
+    assert cirq.big_endian_int_to_bits(-3, bit_count=4) == [1, 1, 0, 1]
+
+
+def test_big_endian_int_to_digits():
+    with pytest.raises(ValueError, match='Inconsistent digit count'):
+        _ = cirq.big_endian_int_to_digits(0, base=[], digit_count=1)
+    with pytest.raises(ValueError, match='Inconsistent digit count'):
+        _ = cirq.big_endian_int_to_digits(0, base=[2, 3], digit_count=20)
```
cduck: Mypy doesn't complain about these?

Strilanc: I don't think overloads have to be exhaustive. These meet the signature of the general method..

cirq/value/digits.py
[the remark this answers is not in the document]
Strilanc: Fixed.

cirq/value/digits.py
```diff
@@ -0,0 +1,195 @@
...
+                             digit_count: Optional[int] = None,
+                             base: Union[int, Iterable[int]]) -> List[bool]:
+    """Separates an integer into big-endian digits.
+
+    Examples:
+        >>> big_endian_int_to_digits(11, digit_count=4, base=10)
+        [0, 0, 1, 1]
+
+        >>> big_endian_int_to_digits(11, digit_count=4, base=[2, 3, 4])
+        [0, 0, 1, 1]
```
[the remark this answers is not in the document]
Strilanc: Fixed.

cirq/value/digits.py
```diff
@@ -0,0 +1,195 @@
...
+
+        >>> big_endian_int_to_digits(11, digit_count=4, base=3)
+        [0, 1, 0, 2]
+
+        >>> big_endian_int_to_digits(11, base=[2, 3, 4])
+        [0, 1, 0, 2]
```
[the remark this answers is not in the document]
Strilanc: Fixed.

cirq/value/digits.py
Strilanc: Done.

cirq/value/digits_test.py
cduck: It should.

https://mypy.readthedocs.io/en/latest/more_types.html#function-overloading
> For example, mypy will understand that mouse_event(5, 25) will always have a return type of ClickEvent and will **report errors for calls like mouse_event(5, 25, 2)**.

cirq/study/trial_result.py
```diff
@@ -44,38 +43,19 @@ def _tuple_of_big_endian_int(bit_groups: Iterable[Any]) -> Tuple[int, ...]:
...
-def _keyed_repeated_bitstrings(measurements: pd.DataFrame) -> str:
+def _keyed_repeated_bitstrings(vals: Dict[str, np.ndarray]) -> str:
     keyed_bitstrings = []
-    for key in sorted(measurements.columns):
-        reps = pd.DataFrame(measurements[key].to_list())
-        all_bits = ', '.join(reps.apply(_bitstring, axis=0))
+    for key in sorted(vals.keys()):
+        reps = vals[key]
+        n = 0 if len(reps) == 0 else len(reps[0])
+        all_bits = ', '.join([_bitstring(reps[:, i]) for i in range(n)])
```
cduck: I missed this one earlier.  Brackets aren't needed.

cirq/value/digits_test.py
Strilanc: Mypy accepts the following:

```python
@overload
def ff(x: int, y: None) -> bool:
    pass

@overload
def ff(x: None, y: int) -> str:
    pass

def ff(x: Optional[int], y: Optional[int]) -> Union[str, bool]:
    pass


def g():
    a = ff(1, None)
    b = ff(None, 1)
    c = ff(None, None)
    d = ff(1, 1)
```

So I'm inclined to think it's fine.

PR #1814, opened 2019-07-15
TrialResult.measurements typo fix

mpharrigan:
How did this pass tests?


mpharrigan:
maybe `to_list` was added in a more recent version of pandas than the one I've got?

mpharrigan:
yes, It was added on 0.24.0 and I have 0.23.4

Strilanc:
Add a minimum version to the pandas line in requirements.txt

mpharrigan:
This seems to be subsumed by other changes to TrialResult

PR #1820, opened 2019-07-16
added swap_gate arg to complete acquaintance strategy

bryano:
Fixes omission of swap_gate arg, and adds tests to catch it.


cirq/contrib/acquaintance/permutation.py
```diff
@@ -234,15 +234,26 @@ def get_logical_operations(operations: ops.OP_TREE,
 
 class ExpandPermutationGates(optimizers.ExpandComposite):
     """Decomposes any permutation gates other SwapPermutationGate."""
```
dabacon: update doc since (which has a grammar bug btw) to describe that this sometimes keeps the SwapPermutationGates

cirq/contrib/acquaintance/permutation.py
```diff
@@ -234,15 +234,26 @@ def get_logical_operations(operations: ops.OP_TREE,
...
+            ]))
+        else:
+            self.no_decomp = lambda op: (not all([
+                isinstance(op, ops.GateOperation),
+                isinstance(op.gate, PermutationGate)
+            ]))
+
 
-expand_permutation_gates = ExpandPermutationGates()
+expand_permutation_gates = ExpandPermutationGates(keep_swap_permutations=True)
```
dabacon: style nit: constants should be CAPS_CASE

cirq/contrib/acquaintance/permutation.py
bryano: Done

cirq/contrib/acquaintance/permutation.py
bryano: Done

dabacon:
LGTM

PR #1821, opened 2019-07-16
Adds functionality for viewing and debugging swap networks

bryano:


cirq/contrib/acquaintance/permutation.py
```diff
@@ -86,6 +86,41 @@ def _circuit_diagram_info_(self, args: protocols.CircuitDiagramInfoArgs
         return wire_symbols
 
 
+class MappingDisplayGate(ops.Gate):
```
Strilanc: Docstring

cirq/contrib/acquaintance/permutation.py
```diff
@@ -86,6 +86,41 @@ def _circuit_diagram_info_(self, args: protocols.CircuitDiagramInfoArgs
...
+
+    def __init__(self, indices):
+        self.indices = tuple(indices)
+        self._num_qubits = len(self.indices)
+
+    def num_qubits(self):
```
Strilanc: annotate return type

cirq/contrib/acquaintance/permutation.py
```diff
@@ -199,3 +242,27 @@ def __init__(self):
...
+    permutation = {i: mapping[q] for i, q in enumerate(qubits)}
+    returning_permutation_op = LinearPermutationGate(n_qubits, permutation,
+                                                     swap_gate)(*qubits)
+    circuit.append(returning_permutation_op)
+
+
+def uses_consistent_swap_gate(circuit: circuits.Circuit,
+                              swap_gate: ops.Gate) -> bool:
+    for op in circuit.all_operations():
+        if (isinstance(op, ops.GateOperation) and
```
Strilanc:
```
gate = ops.op_gate_of_type(op, PermutationGate)
if gate is not None and gate.swap_gate != swap_gate:
    return False
```
```

File: cirq/contrib/acquaintance/permutation.py (hunk 1 above)
Comment (bryano) [reply cirq/contrib/acquaintance/permutation.py]: Good?

File: cirq/contrib/acquaintance/permutation.py [hunk 4]
```diff
@@ -199,3 +243,26 @@ def __init__(self):
...
+    permutation = {i: mapping[q] for i, q in enumerate(qubits)}
+    returning_permutation_op = LinearPermutationGate(n_qubits, permutation,
+                                                     swap_gate)(*qubits)
+    circuit.append(returning_permutation_op)
+
+
+def uses_consistent_swap_gate(circuit: circuits.Circuit,
+                              swap_gate: ops.Gate) -> bool:
+    for op in circuit.all_operations():
+        gate = ops.op_gate_of_type(op, Type[PermutationGate])
```
bryano: Any thoughts on how to do the typing here? Mypy doesn't like [instantiating a typevar with an abstract class](https://github.com/python/mypy/issues/5374).

bryano: I reverted so the tests pass, but if you see a way to use `op_gate_of_type` that mypy finds acceptable, let me know.

PR #1824, opened 2019-07-16
Add doctest to .travis.yml

Strilanc:
Was supposed to be in a previous PR, but appears to have been lost in a merge conflict with the removal of the perf tests.


cduck:
Consider moving it before or after the three pytest entries so it looks nicer in Travis.

PR #1825, opened 2019-07-16
Add --actually-quiet parameter to check/pytest and use it in continuous integration runs

Strilanc:
- This fixes an annoyance where error output is hidden waaaay down at the bottom of travis-ci logs

### DIFF
--- a/check/pytest
+++ b/check/pytest
@@ -4,9 +4,11 @@
 # Runs pytest on the repository.
 #
 # Usage:
-#     check/pytest [--flags] [file-paths-relative-to-repo-root]
+#     check/pytest [--actually_quiet] [--flags for pytest] [file-paths-relative-to-repo-root]
 #
-# You may specify additional flags or specific files to test. The file paths
+# The --actually_quiet argument filters out any progress output from pytest.
+#
+# You may specify pytest flags and specific files to test. The file paths
 # must be relative to the repository root. If no files are specified, everything
 # is tested.
 ################################################################################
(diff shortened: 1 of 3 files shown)


check/pytest
```diff
@@ -4,9 +4,11 @@
 # Runs pytest on the repository.
 #
 # Usage:
-#     check/pytest [--flags] [file-paths-relative-to-repo-root]
+#     check/pytest [--actually_quiet] [--flags for pytest] [file-paths-relative-to-repo-root]
```
cduck: The flag should probably have a dash, not an underscore to be consistent with other flags.  `--actually-quiet`.

check/pytest
```diff
@@ -15,4 +17,20 @@
...
+        PYTEST_ARGS+=("${arg}")
+    fi
+done
+
+if [ -z "${ACTUALLY_QUIET}" ]; then
+    pytest "${PYTEST_ARGS[@]}"
+else
+    # Filter out lines like "...F....x...      [ 42%]", with coloring.
+    pytest -q --color=yes "${PYTEST_ARGS[@]}" |
+        grep -Pv '^(.\[0m)?[\.FEsx]+(.\[36m)?\s+\[\s*\d+%\](.\[0m)?$'
```
cduck: This seems kind of brittle but nothing bad will happen if pytest updates and it breaks.

Strilanc: Yeah, I'm more worried about hiding real test output than I am about perfectly suppressing pytest's output.

Unfortunately it seems to have missed pytest's output on travis in the windows tests, so I need to tweak it at least a little.

check/pytest
Strilanc: Done.

check/pytest
Strilanc: I have no idea how to fix the windows thing. For now I will defer it. https://github.com/quantumlib/Cirq/issues/1826

PR #1827, opened 2019-07-16
When user removes automerge but leaves front_of_queue_automerge, cancel automerge

Strilanc:


dev_tools/auto_merge.py
```diff
@@ -18,8 +18,9 @@
 ACCESS_TOKEN_ENV_VARIABLE = 'CIRQ_BOT_GITHUB_ACCESS_TOKEN'
 
 POLLING_PERIOD = datetime.timedelta(seconds=10)
+USER_AUTO_MERGE_LABEL = 'front_of_queue_automerge'
```
cduck: Did you mean `'automerge'`?

Strilanc: ...yes

Strilanc: Fixed

Strilanc:
I was running the old version initially. Then it takes it ~45 seconds before it becomes confident that it's not a temporary thing.

PR #1828, opened 2019-07-16
Make resolve_parameters/is_parameterized recurse on tuples and lists

bryano:


cirq/protocols/resolve_parameters.py
```diff
@@ -88,10 +89,13 @@ def resolve_parameters(
     param_resolver = ParamResolver(param_resolver)
     if isinstance(val, sympy.Basic):
         return param_resolver.value_of(val)
+    if isinstance(val, Sequence):
+        return type(val)(resolve_parameters(e, param_resolver) for e in val)
```
Strilanc: This is not safe. Sequence types are not guaranteed to have a constructor that takes a generator. I think all a type has to do to be a sequence is have `__len__` and `__getitem__`.

I'd recommend having an explicit list of types we support, like just tuple and list and dict.

bryano: Okay, I changed it to list and tuples explicitly.

Is it clear to you how dicts should be resolved (i.e., just the keys, just the values, or both)? I think it's ambiguous enough that we should wait for a use case.

Strilanc: I agree.

cirq/protocols/resolve_parameters.py
```diff
@@ -88,10 +89,13 @@ def resolve_parameters(
...
+    if isinstance(val, (list, tuple)):
+        return type(val)(resolve_parameters(e, param_resolver) for e in val)
 
     getter = getattr(val, '_resolve_parameters_', None)
     result = NotImplemented if getter is None else getter(param_resolver)
 
     if result is not NotImplemented:
         return result
-    return val
+    else:
```
Strilanc: Victory just went around getting rid of this "return else return" thing.

bryano:
Added `is_parameterized`.

cirq/protocols/resolve_parameters_test.py
```diff
@@ -58,6 +58,18 @@ def _resolve_parameters_(self, resolver: ParamResolver):
     assert cirq.resolve_parameters(SimpleParameterSwitch('a'), d).parameter == 0
     assert cirq.resolve_parameters(sympy.Symbol('a'), r) == 0
 
+    a, b, c = tuple(sympy.Symbol(l) for l in 'abc')
+    x, y, z = 0, 4, 7
+    resolver = {a: x, b: y, c: z}
+    assert cirq.resolve_parameters((a, b, c), resolver) == (x, y, z)
```
Strilanc: Check the empty list corner case, and also that a list with no parameters is returned unchanged.

PR #1830, opened 2019-07-17
[QASM import] two and three qubit gates

balopat:
Adds support for two and three qubit gates in QASM import.
This is PR number 10 for the piecemeal reimplementation of #1602.

Next PR will add docs.

### DIFF
--- a/cirq/contrib/qasm_import/_parser.py
+++ b/cirq/contrib/qasm_import/_parser.py
@@ -278,7 +278,11 @@ class QasmParser:
                           num_params=0,
                           num_args=1,
                           cirq_gate=ops.T**-1),
-        **basic_gates
+    }
+
+    all_gates = {
+        **basic_gates,
+        **qelib_gates
     }
 
     tokens = QasmLexer.tokens
@@ -375,7 +379,7 @@ class QasmParser:
 
     def _resolve_gate_operation(self, args: List[List[ops.Qid]], gate: str,
                                 p: Any, params: List[float]):
-        gate_set = (self.basic_gates if not self.qelibinc else self.qelib_gates)
+        gate_set = (self.basic_gates if not self.qelibinc else self.all_gates)
         if gate not in gate_set.keys():
             msg = 'Unknown gate "{}" at line {}{}'.format(
                 gate, p.lineno(1), ", did you forget to include qelib1.inc?"
--- a/cirq/contrib/qasm_import/_parser.py
+++ b/cirq/contrib/qasm_import/_parser.py
@@ -280,10 +280,7 @@ class QasmParser:
                           cirq_gate=ops.T**-1),
     }
 
-    all_gates = {
-        **basic_gates,
-        **qelib_gates
-    }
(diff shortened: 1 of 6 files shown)


cirq/contrib/qasm_import/_parser.py
```diff
@@ -243,6 +278,7 @@ def __init__(self):
                           num_params=0,
                           num_args=1,
                           cirq_gate=ops.T**-1),
+        **basic_gates
```
Strilanc: These technically aren't qelib gates. Maybe you want to have an `all_gates`?

cirq/contrib/qasm_import/qasm_test.py
```diff
@@ -11,24 +11,77 @@
...
+
+    cirq.testing.assert_allclose_up_to_global_phase(cirq_unitary,
+                                                    qiskit_unitary,
+                                                    rtol=1e-8,
+                                                    atol=1e-8)
+
 
-    circuit2 = cirq.contrib.qasm_import.qasm.QasmCircuitParser().parse(qasm1)
-    ct.assert_same_circuits(circuit1, circuit2)
+def _reorder_indices_of_matrix(matrix: np.ndarray, new_order: List[int]):
```
balopat: this was a shameless copy paste from `cirq.circuits.qasm_output_test` - should I pull it out to a test utility in `cirq.testing`?

Strilanc: I was thinking something more along the lines of defining the method `assert_parsed_qasm_circuit_agrees_with_qiskit`, and yes sharing it with the output test. Note that this method should not be in `cirq.testing` because it has a dependency on qiskit.

balopat: Interesting. I would like to clarify the packages/dependency rules before I jump in: 
 
1. `cirq.testing` has `assert_qasm_is_consistent_with_unitary` in consistent_qasm.py which does refer to qiskit which file also contains a copy of `_reorder_indices_of_matrix` (it's also in cirq.circuits.qasm_output_test which has a `test_output_unitary_same_as_qiskit` test closer to what I wanted to do here)
2. I suspect that the rule of not calling contrib from cirq still stands for tests? In that case no qasm import functionality should be called from qasm output / consistency tests under cirq.testing or cirq.circuit

How about: 

1. `consistent_qasm.py` would be still the home for the qiskit related consistency checks. 
2. I would add a new, simpler assertion method `assert_qiskit_parsed_qasm_consistent_with_unitary(qasm, unitary)` that checks the passed in `unitary` against the qiskit parsed version of a passed in `qasm` string. This can then be used in `contrib.qasm_import.qasm_test` as well as in `qasm_output_test`

balopat: I added a commit implementing this idea:  https://github.com/quantumlib/Cirq/pull/1830/commits/462a069d3782796de41bb8624f48eed12dad374d Let me know what you think!

PR #1832, opened 2019-07-17
[Qudits] Add support to cirq.channel, cirq.apply_channel, and cirq.mixture

cduck:
For #933. Part 3 after #1771.

- `cirq.apply_channel` checks for the `qid_shape` protocol and use properly sized numpy arrays.
- `cirq.channel` and `cirq.mixture` should just work.

Qudit features not in this PR are simulators, measurements, optimizers.

### DIFF
--- a/cirq/protocols/apply_channel.py
+++ b/cirq/protocols/apply_channel.py
@@ -20,8 +20,8 @@ import numpy as np
 from typing_extensions import Protocol
 
 from cirq import linalg
-from cirq.protocols.apply_unitary import (
-    apply_unitary, ApplyUnitaryArgs, _incorporate_result_into_buffer)
+from cirq.protocols.apply_unitary import (apply_unitary, ApplyUnitaryArgs,
+                                          _incorporate_result_into_buffer)
 from cirq.protocols.channel import channel
 from cirq.protocols import qid_shape_protocol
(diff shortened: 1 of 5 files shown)


cduck:
I am still adding tests for non-qubit channels and mixtures.

cirq/protocols/apply_channel.py
```diff
@@ -324,3 +433,26 @@ def _apply_krauss_multi_qubit(krauss: Union[Tuple[Any], Sequence[Any]],
...
+        sub_args: A version of `args` with transposed and sliced views of
+            it's tensors.
+        sub_result: The result of calling an object's `_apply_channel_`
+            method on `sub_args`.  A transposed subspace of the desired
+            result.
+
+    Returns: The full result tensor after applying the channel.  Either
+        `args.target_tensor` or `args.out_buffer`.
+    """
+    return _incorporate_result_into_buffer(args.target_tensor, args.out_buffer,
```
Strilanc: Interesting that we don't have to use all the fields.

cduck: The documentation says that `_apply_channel_` isn't allowed to return either `auxiliary_buffer` as the result.

cduck:
This incorrectly extends channel matrices.

Issue #879, opened 2018-08-27
Optimize clifford/non-clifford matching optimization by calculating qubit states


Strilanc:
Currently, the optimization that removes non-Clifford operations and merges them does so by individually moving them through the circuit. If there are C cliffords and N non-cliffords, it takes N*C time.

What we should do instead is assign the k'th qubit an initial Pauli string of X_k for its X observable and Z_k for its Z observable. Then we iterate through the circuit, ignoring non-Cifford operations but modifying those Pauli strings with the Clifford operations, and keeping track of the string assigned to each qubit after each operation.

For example, start with a circuit:

```
-----@-----@-----
     |     |
--T--X--T--X--T--
```

Assign initial strings:

For example, start with a circuit:

```
(x0,z0)-----@-----@-----
            |     |
(x1,z1)--T--X--T--X--T--
```

Iterate through ignoring non-cliffords while tracking the strings:

```
(x0,z0)----(x0,z0)-@-----@-----
                   |     |
(x1,z1)--T-(x1,z1)-X--T--X--T--
```

But update based on Cliffords:

```
(x0,z0)----(x0,z0)-@-(x0,z0*x1)----@-----
                   |               |
(x1,z1)--T-(x1,z1)-X-(x1*z0,z1)-T--X--T--
```

Do the whole thing:

```
(x0,z0)----(x0,z0)-@-(x0,z0*x1)----@-(x0,z0)----
                   |                         |
(x1,z1)--T-(x1,z1)-X-(x1*z0,z1)-T-(x1*z0,z1)-X-(x1,z1)-T--
```

The non-clifford operations are now preceded by the Pauli string that they should be assigned. This took time O(C+N) instead of O(N*C). We also don't need to worry about finding where the Non-cliffords returns to, since they are still in place. All the commutes-or-not computations stay the same, and generally this should just make things faster.

daxfohl:
I'd like to try this one, but @Strilanc can you flesh out the final step a bit more? Plugging your original circuit into the existing optimizer we get

```
0: ───Y^-0.5───@───[X]^-0.25───@───Y^0.5───
               │               │
1: ────────────@───────────────@───Z^0.5───
```

I'm not quite seeing how to get from your final step above to this final representation.

(Also, I see that the first step of the optimizer calls `converted_gate_set`, which does *all* the work--the final circuit comes directly out of that, and the remaining optimizations don't change anything else. So here are you talking about a strategy that skips the call to `converted_gate_set`?)

[...]

PR #1834, opened 2019-07-17
Optimize recombination step in paulistring optimizer

balopat:
Before optimization: 

```bash
2019-07-16 09:16:20,584 - INFO - reading qasm file...
2019-07-16 09:16:25,140 - INFO - parsed qasm file: 13392 ops
2019-07-16 09:18:24,564 - INFO - optimized circuit: 1854 ops in 119.42398381233215 
2019-07-16 09:18:24,654 - INFO - stats: misses 74115, dupes: 6505723
```
which means that 98.87% of comparisons has been done at least twice.

Insight: we can  just sort the nodes of the frontier _once_, based on the best placement possible for a given paulistring instead of repeatedly reevaluating them. Why? 

1.) because the nodes of the frontier (rightmost nodes in the DAG) are commuting with each other, so they can be sorted independently of each other and each other's predecessors in the DAG and we can insert all the nodes in the frontier first.
2.)  sorting the nodes seems to guarantee the insertion indices to be sorted _as well_ , which is great, because then we don't have to worry about recalculating the best insertion spot for a node (as a previous insertion might have changed the circuit?) as I start with the furthest one, that leaves the circuit the same before, where the next node can come in...- why does sorting guarantee the insertion indices to be sorted? Because the best placements are where the paulistrings become of length one - and within those placements the insertion index is sorted. 

Thus, we can just insert the nodes in the sorted frontier to their best placements and only then draw the new frontier. 

This gives us: 

```bash
2019-07-16 10:00:49,046 - INFO - reading qasm file...
2019-07-16 10:00:53,691 - INFO - parsed qasm file: 13392 ops
2019-07-16 10:01:46,964 - INFO - optimized circuit: 1854 ops in 53.273059129714966 
2019-07-16 10:01:47,044 - INFO - stats: misses 843915, dupes: 31864
```

which is 3% duplication and >50% time reduction.

### DIFF
--- a/cirq/contrib/paulistring/recombine.py
+++ b/cirq/contrib/paulistring/recombine.py
@@ -80,10 +80,10 @@ def move_pauli_strings_into_circuit(circuit_left: Union[circuits.Circuit,
     while rightmost_nodes:
         # Pick the Pauli string that can be moved furthest through the Clifford
         # circuit
-        frontier = _sorted_best_string_placements(rightmost_nodes, output_ops)
+        placements = _sorted_best_string_placements(rightmost_nodes, output_ops)
         last_index = len(output_ops)
-        while frontier:
-            best_string_op, best_index, best_node = frontier.pop()
+        while placements:
+            best_string_op, best_index, best_node = placements.pop()
             # Place the best one into the output circuit
             assert best_index <= last_index, "{} >= {}, len: {}".format(
                 best_index, last_index, len(output_ops))
--- a/cirq/contrib/paulistring/recombine.py
+++ b/cirq/contrib/paulistring/recombine.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import (Any, Callable, Iterable, Iterator, Sequence, Tuple, Union,
+from typing import (Any, Callable, Iterable, Sequence, Tuple, Union,
                     cast, List)
 
 from cirq import ops, circuits
--- a/cirq/contrib/paulistring/recombine.py
+++ b/cirq/contrib/paulistring/recombine.py
@@ -12,8 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import (Any, Callable, Iterable, Sequence, Tuple, Union,
-                    cast, List)
+from typing import (Any, Callable, Iterable, Sequence, Tuple, Union, cast, List)
 
 from cirq import ops, circuits
 
@@ -85,8 +84,10 @@ def move_pauli_strings_into_circuit(circuit_left: Union[circuits.Circuit,
         while placements:
             best_string_op, best_index, best_node = placements.pop()
             # Place the best one into the output circuit
-            assert best_index <= last_index, "{} >= {}, len: {}".format(
-                best_index, last_index, len(output_ops))
+            assert (best_index <= last_index,
+                    "Unexpected insertion index order,"
+                    " {} >= {}, len: {}".format(best_index, last_index,
+                                                len(output_ops)))
 
             last_index = best_index
             output_ops.insert(best_index, best_string_op)
--- a/cirq/contrib/paulistring/recombine.py
+++ b/cirq/contrib/paulistring/recombine.py
@@ -77,13 +77,15 @@ def move_pauli_strings_into_circuit(circuit_left: Union[circuits.Circuit,
                        - set(before for before, _ in string_dag.edges()))
 
     while rightmost_nodes:
-        # Pick the Pauli string that can be moved furthest through the Clifford
-        # circuit
+        # Sort the pauli string placements based on paulistring length and
+        # furthest possible distance in circuit_right
         placements = _sorted_best_string_placements(rightmost_nodes, output_ops)
         last_index = len(output_ops)
         while placements:
+            # Pick the Pauli string that can be moved furthest through
+            # the Clifford circuit
             best_string_op, best_index, best_node = placements.pop()
-            # Place the best one into the output circuit
+
             assert (best_index <= last_index,
                     "Unexpected insertion index order,"
                     " {} >= {}, len: {}".format(best_index, last_index,
--- a/cirq/contrib/paulistring/recombine.py
+++ b/cirq/contrib/paulistring/recombine.py
@@ -86,10 +86,10 @@ def move_pauli_strings_into_circuit(circuit_left: Union[circuits.Circuit,
             # the Clifford circuit
             best_string_op, best_index, best_node = placements.pop()
 
-            assert (best_index <= last_index,
-                    "Unexpected insertion index order,"
-                    " {} >= {}, len: {}".format(best_index, last_index,
-                                                len(output_ops)))
+            assert best_index <= last_index, (
+                "Unexpected insertion index order,"
+                " {} >= {}, len: {}".format(best_index, last_index,
+                                            len(output_ops)))
 
             last_index = best_index
             output_ops.insert(best_index, best_string_op)
--- a/cirq/contrib/paulistring/recombine.py
+++ b/cirq/contrib/paulistring/recombine.py
@@ -58,7 +58,7 @@ def _sorted_best_string_placements(
 
         node_maxes.append(node_max)
 
-    return sorted(node_maxes, key=sort_key)
+    return sorted(node_maxes, key=sort_key, reverse=True)
 
 
 def move_pauli_strings_into_circuit(circuit_left: Union[circuits.Circuit,
@@ -81,10 +81,10 @@ def move_pauli_strings_into_circuit(circuit_left: Union[circuits.Circuit,
         # furthest possible distance in circuit_right
         placements = _sorted_best_string_placements(rightmost_nodes, output_ops)
         last_index = len(output_ops)
-        while placements:
-            # Pick the Pauli string that can be moved furthest through
-            # the Clifford circuit
-            best_string_op, best_index, best_node = placements.pop()
+
+        # Pick the Pauli string that can be moved furthest through
+        # the Clifford circuit
+        for best_string_op, best_index, best_node in placements:
 
             assert best_index <= last_index, (
                 "Unexpected insertion index order,"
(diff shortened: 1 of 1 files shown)


Strilanc:
Looks reasonable to me, but I'll defer to @cduck since he's most familiar with this code

balopat:
Thank you for the detailed explanation of the corner case. Yes, I was worried about that a little bit as well...but then all tests passed and the assertions did not fail even on large, diverse set of circuits :D I appreciate your thoughtful analysis! 
I'll have a look and see if I can reason about it - if nothing else, but out of curiosity. 

I really like that the idea in #879 is significantly more promising with O(N+C) over O(N*C). 
Just to confirm that I understand it correctly: we would use the (x,z) observables (implemented just simply as cirq.X and cirq.Z operations?) dragged across the circuit, affected by Cliffords to create markers for the non-Cliffords and that have the effect of the "dragging". Now we would still need to extract the non-cliffords in a DAG, merge the mergeable PaulStrings and remove the negligible ones. So some cases there is going to be less PauliStrings than markers. Which markers would merged PauliString nodes belong to? The one with a higher index I suppose as that contains the effect of the previous Cliffords.

cduck:
You have the right idea.  It's not clear to me exactly how the X and Z strings are tracked in #879.  Check with @Strilanc.

Yes, I'm pretty sure that placing a merged pauli string at the highest index of the original strings will work and is equivalent to the current code.

cirq/contrib/paulistring/recombine.py
```diff
@@ -50,13 +52,13 @@ def _possible_string_placements(
                 break
             string_op = string_op.pass_operations_over([out_op],
```
cduck: Please add a test that exercises this line when `out_op` is another pauli string placed earlier (this is the corner case I described).  This case didn't happen before this PR.

balopat: thank you for the analysis, this is cool, your derivation seems sound to me. 

And this argument holds even after merging PauliStrings with other PauliStrings that equal to it up to a coefficient - as that doesn't change the commutativity relations. 

I'm not sure I can cook up an interesting test case for this: 

Rephrasing your situation: 

- in the DAG the Paulistring A' has an ancestor B' that it doesn't commute with 
- A' and C' are in the frontier (commuting) 
- C' commutes with B as well

```
B' --- A'
       C'
```

In order to get the desired "conflict", we'll need to have: 
- A' has to be further than C' - so that it can be picked first from the frontier 
- B' has to be further in the circuit than C' - so the "conflict" arises

Thus: 
- A' gets inserted first 
- then C' is inserted 
- then B' needs to commute over C' to get to its place 

The only way the original single qubit paulistrings will commute is if they are equal up to a coefficient or they are on a different qubit. Having all of them in the same qubit leads to contradiction: C ~= A ~= B => [C,B]=0 => [C',B'] = 0 that contradicts our initial setup. Hence they have to be on separate qubits. 

```
...-----@-- B ------ A ----- 
...     |
...-C---@-----------------
```

To cover the line you want to cover with op being a paulistring, we need to make sure that as B' is passing through, it is a multiqubit PauliString that has a Pauli D' on the qubit where C lives and D' ~= C. Otherwise if the qubits are disjoint, we just simply skip the passing. 

Thinking...

Strilanc:
Assume A commutes with B, i.e. that [A, B] = AB - BA = 0

Then:

[UAU*, UBU*]
= UAU*UBU* - UBU*UAU*
= UABU* - UBAU*
= U(AB - BA)U* by distributivity
= U 0 U*
= 0

Therefore UAU* commutes with UBU*

cirq/contrib/paulistring/recombine.py
cduck: It was very hard to find a circuit to do this but I think I found one that does.  I only checked it by hand so I may have made a mistake.  The key was making [3] operate on another qubit that [2] didn't operate on and making [1] operate on another qubit that [2] and [3] don't operate on.

```
q0, q1, q2 = cirq.LineQubit.range(3)
strings = cirq.Circuit.from_ops(
    (cirq.Z(q1)*cirq.X(q2))**0.3,  # Operation [3]
    (cirq.X(q2))**0.2,  # Operation [2]
    (cirq.Z(q0)*cirq.X(q1))**0.1,  # Operation [1]
)
circuit = cirq.Circuit.from_ops(
    cirq.CNOT(q2, q1),
    cirq.SingleQubitCliffordGate.Z_sqrt(q2),
    cirq.SingleQubitCliffordGate.X_sqrt(q1),
    cirq.SingleQubitCliffordGate.X_sqrt(q1),
    cirq.SingleQubitCliffordGate.Y_sqrt(q0),
    cirq.CZ(q0, q1),
    cirq.SingleQubitCliffordGate.X_sqrt(q1),
    cirq.SingleQubitCliffordGate.Y_sqrt(q0),
    cirq.CZ(q1, q2),
    cirq.CX(q1, q0),
    cirq.SingleQubitCliffordGate.X_sqrt(q0),
    cirq.CNOT(q0, q2),
    cirq.SingleQubitCliffordGate.X_sqrt(q0),
    strategy=cirq.InsertStrategy.NEW_THEN_INLINE,
)
strings + circuit
```
```
0: ─────────────[Z]───────    ───────────────Y^0.5───@───Y^0.5───────X───X^0.5───@───X^0.5───
                │                                    │               │           │
1: ───[Z]───────[X]^0.1───    ───X───X^0.5───X^0.5───@───X^0.5───@───@───────────┼───────────
      │                          │                               │               │
2: ───[X]^0.3───X^0.2─────    ───@───Z^0.5───────────────────────@───────────────X───────────
```

- [1] is placed at the far end of the circuit on qubit 0.
- [2] is can only be placed on the left before the first CNOT.
- [3] commutes past [2] and becomes a single qubit pauli string right before the last X^0.5.

The circuit could probably be simplified.

balopat: I just had time to quickly try this out, will look into this deeper in the days. 

After running `dag = pauli_string_dag_from_circuit(c)` tried both the old and the new insertion algo, the new ops seem to get stuck at the front of the circuit: 

image

balopat:
> Assume A commutes with B, i.e. that [A, B] = AB - BA = 0
> 
> Then:
> 
> ```
> [UAU*, UBU*]
> = UAU*UBU* - UBU*UAU*
> = UABU* - UBAU*
> = U(AB - BA)U* by distributivity
> = U 0 U*
> = 0
> ```
> 
> Therefore UAU* commutes with UBU*. The commutativity relation of observables does not change as they are moved through operations in a circuit.

Thank you for this, yes -  @cduck's argument is similarly just using the same `U`. 
 If A and B are separated by one or more CZ operations, can that change the commutativity relations? And if that is the case, then order might matter in some cases - as even though A' and B' commute within the DAG, after being moved out from Ua and Ub, the finally placed B inserted earlier and the "partially moved" A might not. 

```
q0, q1, q2  = cirq.LineQubit.range(3)
c_orig = cirq.Circuit.from_ops(       
        ops.CZ(q0, q1),
        ops.CZ(q1, q2),  
        cirq.X(q1)**0.2,
        ops.CZ(q2, q0),
        cirq.X(q2)**0.1,       
        ops.CZ(q2, q0),
        cirq.X(q0)**0.3,       
)
```

image

Which, after inserting [3] and [2], [1] comes ... and steps over [2] - however this is where I'm a bit struggling to create a good test case - as in this case `(Z(0)*X(2))**0.1` qubits do not overlap with `(X(1))**0.2`...

```
0:    (Z(0)*Z(1)*X(2))**0.1 -> (Z(0)*Z(1)*X(2))**0.1 stepping over CZ(0, 1)
1:    (Z(0)*Z(1)*X(2))**0.1 -> (Z(0)*Z(1)*X(2))**0.1 stepping over CZ(1, 2)
2:    (Z(0)*Z(1)*X(2))**0.1 -> (Z(0)*X(2))**0.1 stepping over (X(1))**0.2
3:    (Z(0)*Z(1)*X(2))**0.1 -> (Z(0)*X(2))**0.1 stepping over CZ(2, 0)
4:    (Z(0)*Z(1)*X(2))**0.1 -> (X(2))**0.1
```

cirq/contrib/paulistring/recombine.py
balopat: I think it's because CNOT is not part of the gateset, CZ is.

cduck: Yes.  You can replace `CNOT(a, b)` with `[cirq.SingleQubitCliffordGate.Y_sqrt(b), CZ(a, b), cirq.SingleQubitCliffordGate.Y_nsqrt(b)]`.  (I may have the signs of the Y gates flipped.)

balopat: Ah yes, but they commute...which we check for before. i.e. the desired line won't be executed.
I commented with a simpler case that achieves the same, plus here's my derivation for the partial circuit matching - commutativity still holds even when we move the ops just part of the circuit: 

image

cduck: I didn't see that check.  I guess the code already handles this case.  That's what I wanted to trigger with the test.

I'll approve this once you address the other comment.

balopat: Done!

cirq/contrib/paulistring/recombine.py
```diff
@@ -75,22 +77,28 @@ def move_pauli_strings_into_circuit(circuit_left: Union[circuits.Circuit,
...
-            if len(string_dag.succ[pred_node]) <= 1)
-        string_dag.remove_node(best_node)
+        # Sort the pauli string placements based on paulistring length and
+        # furthest possible distance in circuit_right
+        placements = _sorted_best_string_placements(rightmost_nodes, output_ops)
+        last_index = len(output_ops)
+        while placements:
+            # Pick the Pauli string that can be moved furthest through
+            # the Clifford circuit
+            best_string_op, best_index, best_node = placements.pop()
```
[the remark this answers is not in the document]
balopat: done!

cirq/contrib/paulistring/recombine.py
balopat: Thank you again for the thorough review!

Issue #1760, opened 2019-06-27
Is anyone still using `continuous-integration/check.sh` ?


Strilanc:
In my own workflow I've switched over entirely to the scripts in the `check/` folder, which are significantly faster (if not quite as bullet proof in terms of agreeing with Travis-ci).

I ask because without maintenance these scripts gradually bit-rot and stop working. E.g. @cduck had some trouble with continuous-integration/check.sh .

vtomole:
I was about to make this issue.

CC: @dabacon

bryano:
I use `continuous-integration/simple_check.sh`.

dabacon:
I haven't been using it.  Do we have a script that runs all the relevant scripts in the check folder?

Strilanc:
No, there's not a "run all" script.

Because we now have continuous integration checks across platforms, it's not actually possible for one person to run all the checks anymore (without VMs anyways).

dabacon:
Yes I understand that.  But right now it is annoying that there isn't one thing I can run which gives me some idea that the build is going to fail or not.  Instead I run each of these by hand.
```
check/format-incremental --apply
check/pylint
check/mypy
check/pytest-changed-files
```
though sometimes I do the incremental coverage.   It would be nice to have a single script that does this

Strilanc:
We could definitely add an `all` script.

It would be a bit unusual for the `all` script to apply formatting changes instead of checking for them.

bryano:
@dabacon `simple_check.py` does the latter three things (with coverage). Its omission of the formatting I think is just because it hasn't been updated since formatting was introduced.

c-poole:
I've been using pytest on my own machine and relying on travis to tell me what else I need to fix.

dabacon:
Oops.  fat moused the button

PR #1839, opened 2019-07-17
Add check/all to run tests close to CI

dabacon:
Me bash foo is poor, so please tell me how to bash better.

Also made some choices about what is default, not sure if I got this correct.

#1760

### DIFF
--- a/check/all
+++ b/check/all
@@ -15,7 +15,11 @@
 #   doctest
 #
 # Usage:
-#     check/all [--only-changed-files] [--apply-format]
+#     check/all revision [BASE_REVISION] [--only-changed-files] [--apply-format]
+#
+# BASE_REVISION is forwarded to format-incremental and to the pytest checks
+# (pytest-and-incremental-coverage or pytest-changed-files). See those
+# checks for how to specify this value.
 #
 # If --only-change-files is specified, pytest-changed-files will be run
 # instead of pytest-and-incremental-coverage.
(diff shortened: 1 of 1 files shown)


check/all
```diff
@@ -0,0 +1,68 @@
...
+# Parse arguments.
+apply=0
+only_changed=0
+for arg in $@; do
+    if [[ "${arg}" == "--only-changed-files" ]]; then
+        only_change=1
+    elif [[ "${arg}" == "--apply-format-changes" ]]; then
+        apply=1
+    else
+        echo -e "\e[31mInvalid arguments. Expected [--only-change-files] [--apply-format].\e[0m" >&2
```
Strilanc: Inconsistent names

check/all
```diff
@@ -0,0 +1,68 @@
...
+
+echo "Running incremental format"
+if [ -z "${apply}" ]; then
+    check/format-incremental --apply
+else
+    check/format-incremental
+fi
+
+if [ -z "${only_changed}" ]; then
+    echo "Running pytest on changed files"
```
Strilanc: Add " (without incremental coverage)".

check/all
```diff
@@ -0,0 +1,68 @@
...
+################################################################################
+
+# Get the working directory to the repo root.
+cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "$(git rev-parse --show-toplevel)"
+
+# Parse arguments.
+apply=0
+only_changed=0
+for arg in $@; do
```
Strilanc: We need to allow the user to specify the revision the compare against. If there's one non-dash argument, it should be forwarded into the incremental scripts.

check/all
```diff
@@ -0,0 +1,68 @@
...
+
+echo "Running pylint"
+check/pylint
+
+echo "Running mypy"
+check/mypy
+
+echo "Running incremental format"
+if [ -z "${apply}" ]; then
+    check/format-incremental --apply
```
Strilanc: This could be `check/format-incremental "${apply_arg}"` with apply_arg being set to the flat or empty string.

check/all
dabacon: Done.  I never use anything besides the default :)

check/all
dabacon: fixed

check/all
dabacon: Done

check/all
dabacon: Done.

dabacon:
PTAL

check/all
```diff
@@ -0,0 +1,75 @@
...
+#
+# Defaults to the following tests:
+#   pylint
+#   mypy
+#   format-incremental
+#   pytest-and-incremental-coverage
+#   doctest
+#
+# Usage:
+#     check/all revision [BASE_REVISION] [--only-changed-files] [--apply-format]
```
c-poole: Should [--apply-format] be [--apply-format-changes]?

dabacon: fixed

PR #1841, opened 2019-07-18
Get processors and calibrations in the Engine

wcourtney:


cirq/google/engine/engine.py
```diff
@@ -619,6 +619,103 @@ def remove_job_labels(self, job_resource_name: str, label_keys: List[str]):
             fingerprint = job.get('labelFingerprint', '')
             self._set_job_labels(job_resource_name, new_labels, fingerprint)
 
+    def list_processors(self, project_id: str) -> List[Dict]:
+        """ Returns a list of Processors that the user has visibility to in the
```
dabacon: Nit, alignment of doc string should be against the """ and should just word wrap.  

Similar in other methods below.

cirq/google/engine/engine.py
```diff
@@ -619,6 +619,103 @@ def remove_job_labels(self, job_resource_name: str, label_keys: List[str]):
...
+            provided project. The names of these processors are used to identify
+            devices when scheduling jobs and gathering calibration metrics.
+
+        Params:
+            project_id: The ID of the Google Cloud project to check, e.g.
+                `my-project-123`
+
+        Returns:
+            A list of dictionaries containing the metadata of each processor.
```
dabacon: should we doc what is in these dictionaries?

cirq/google/engine/engine.py
```diff
@@ -619,6 +619,103 @@ def remove_job_labels(self, job_resource_name: str, label_keys: List[str]):
...
+        """
+        parent = 'projects/%s' % (project_id)
+        response = self.service.projects().processors().list(
+            parent=parent).execute()
+        return response['processors']
+
+    def get_latest_calibration(self,
+                               processor_name: str) -> Optional['Calibration']:
+        """ Returns metadata about a the latest known calibration run for a
```
dabacon: "a the" -> "the"

cirq/google/engine/engine.py
```diff
@@ -619,6 +619,103 @@ def remove_job_labels(self, job_resource_name: str, label_keys: List[str]):
...
+        processor, or None if there is no calibration available.
+
+        Params:
+            processor_name: A string of the form
+                `projects/project_id/processors/processor_id`.
+
+        Returns:
+            A dictionary containing the metadata.
```
dabacon: containing the callibration data?

cirq/google/engine/engine.py
```diff
@@ -619,6 +619,103 @@ def remove_job_labels(self, job_resource_name: str, label_keys: List[str]):
...
+
+
+class Calibration:
+    """A convenience wrapper for calibrations"""
+
+    def __init__(self, calibration: Dict) -> None:
+        self._time = int(calibration['timestampMs'])
+        self._metrics = calibration['metrics']
+
+    def get_timestamp(self) -> int:
```
dabacon: I'd make this a property and just call it timestamp.  Or alternatively in python form, just make it a attribute and doc that in the doc string.

cirq/google/engine/engine.py
```diff
@@ -619,6 +619,103 @@ def remove_job_labels(self, job_resource_name: str, label_keys: List[str]):
...
+            # (e.g. proto version of each value is {<type>: value}).
+            'values':
+            list(value
+                 for values in metric['values']
+                 for value in list(values.values()))
+        }
+                    for metric in self._metrics
+                    if metric['name'] == name)
+
+    def _values_from_metric(self, metric: Dict) -> List:
```
dabacon: unused

cirq/google/engine/engine.py
```diff
@@ -619,6 +619,103 @@ def remove_job_labels(self, job_resource_name: str, label_keys: List[str]):
...
+
+        Params:
+            name: the name of a metric referred to in the calibration. Valid
+            names can be found with the get_metric_names() method.
+
+        Returns:
+            A list of dictionaries containing pairs of targets and values for
+            the requested metric.
+        """
+        return list({
```
dabacon: my preference would be to write this as loops.  when these get multiple lines I find them harder to read.

wcourtney: Rewritten to use mutation + loops. WDYT?

cirq/google/engine/engine.py
wcourtney: I'm concerned about this getting out of sync, but it does seem useful to get a sense of what to expect so I made a pass at what's returned today. PTAL.

wcourtney: The service proto isn't published, so removing docs as discussed offline.

Issue #87, opened 2018-02-13
Make tight trace distance bounds


dabacon:
Currently gates that implement the trace_distance_bound do not so approximately.  For many of these we can explicitly calculate these bounds and we should do so.

vtomole:
In `class ExpZGate` for example

```
def trace_distance_bound(self):
        if isinstance(self.half_turns, value.Symbol):
            return 1
        return abs(self.half_turns) * 3.5
```
Where does the `3.5` come from? Is that what you mean by "approximately"? If so, how would it be calculated explicitly?

Strilanc:
The tight bound is `min_p trace_dist(p, U p U^-1)` where p is a (possibly multi qubit?) density matrix and U is the operation you are trying to bound. That is to say, find a state such that you can distinguish the before-operation state and after-operation states with high probability. The idea is that if the before/after states are nearly impossible to distinguish, then optimizations can provably drop the operation while maintaining a target error budget.

IIRC I assumed the worst case for Z**t was operating on a qubit from an EPR pair.

Strilanc:
Definitely not. That would assign a bound of 0 to the Z gate. You have to pick the worst case state.

Strilanc:
@bryano Oops, yes you're right. Fixed it.

vtomole:
@Strilanc What's the definition of a worst case state? An even mixture? `1/2*(|0><0| + |1><1|)`

Strilanc:
(Keeping in mind that in principle `p` may use up to twice as many qubits as the number of qubits that `U` acts on.)

vtomole:
`p` is needed to calculate the trace distance. I want a generic `p` to use in every case. That's what I'm trying to find. I don't see how that comes from `argmax_p trace_dist(p, U p U^-1)`.

Strilanc:
I suspect that the state where each qubit acted on by U is fully entangled with a qubit not acted on by U is always the worst case state, but I've never tried to prove it. @dabacon are you familiar with how to compute these values exactly without missing tricky corner cases? These kinds of distinguishability questions sometimes have tricky corner cases.

c-poole:
[...]

PR #1842, opened 2019-07-18
Calculate trace_distance_bound for unitaries

### LINKED ISSUE #1700: trace_distance_bound should fallback to unitary before giving up
This value can be calculated from the eigenvalues of a unitary matrix.

c-poole:
- Modifies the protocol trace_distance_bound to calculate the value from the unitary matrix if `_trace_distance_bound_` fails.
This is achieved by eigen-decomposing the unitary matrix and calculating the result directly from the eigenvalues. A special case where the result is a simple function of matrix elements is implemented for single qubit gates. The time cost for a single qubit gate is about ~5us and for a two qubit gate ~50us, doubling with each additional qubit.

- Adds a number of special case `_trace_distance_bound_` functions for gates with known unitaries to avoid needing to do the eigenvalue decomposition. Several of these methods look similar to the modification in the protocol but without the eigen-decomposing. There are also differences in factors of pi due to how eigenvalue information is stored in certain classes such as `eigen_gate`.

- Adds a `_trace_distance_bound_` function to `ParallelGateOperation` which extrapolates the value based on the value of `protocols.trace_distance_bound(self._gate)`.

- Fixes the `_trace_distance_bound_` method of `Controlled_Gate` and `Controlled_Operation` which didn't adequately account for the global phase on the sub gate.

Need to add tests but feel free to comment on implementation or ask questions. These methods should be accurately calculating the `trace_distance_bound` rather than grossly overestimating, and so this is part of #87 . I intend to do mixtures of unitaries as a follow up. Since #1700 only discusses unitaries, this also fixes #1700 .


c-poole:
The method used to perform the calculation is as follows:
Given a eigenvalues of the unitary matrix, find the angle of the minimum arc which contains every eigenvalue. 
If the arclength is greater than pi, return 1
else, return sin(arclength/2)

cirq/ops/controlled_operation.py
```diff
@@ -119,8 +119,22 @@ def _resolve_parameters_(self, resolver):
...
-        return protocols.trace_distance_bound(self.sub_operation)
+    def _trace_distance_bound_(self):
+        if self._is_parameterized_():
+            return 1
+        if protocols.has_unitary(self.sub_operation):
+            angles = np.sort(
+                np.append(
+                    np.angle(
+                        np.linalg.eigvals(protocols.unitary(
+                            self.sub_operation))), 0))
```
cduck: It's too bad that this needs the unitary even if sub_operation implementes `_trace_distance_bound_`.

cirq/ops/controlled_operation.py
```diff
@@ -119,8 +119,22 @@ def _resolve_parameters_(self, resolver):
...
+            maxim = 2 * np.pi + angles[0] - angles[-1]
+            for i in range(1, len(angles)):
+                maxim = max(maxim, angles[i] - angles[i - 1])
+            if maxim <= np.pi:
+                return 1
+            return np.sin(0.5 * maxim)
```
cduck: This might need to be a helper function to minimize code copying (maybe `trace_distance_from_eigen_shifts`).

I'm writing a class similar to ControlledOperation where I would also have to copy-paste this.

cirq/ops/controlled_operation.py
```diff
@@ -119,8 +119,22 @@ def _resolve_parameters_(self, resolver):
         new_sub_op = protocols.resolve_parameters(self.sub_operation, resolver)
         return ControlledOperation(self.controls, new_sub_op)
 
-    def _trace_distance_bound_(self) -> float:
-        return protocols.trace_distance_bound(self.sub_operation)
+    def _trace_distance_bound_(self):
```
cduck: Function return type annotation.

cirq/ops/controlled_gate.py
```diff
@@ -142,7 +142,20 @@ def _resolve_parameters_(self, param_resolver):
...
+                np.append(
+                    np.angle(np.linalg.eigvals(protocols.unitary(
+                        self.sub_gate))), 0))
+            maxim = 2 * np.pi + angles[0] - angles[-1]
+            for i in range(1, len(angles)):
+                maxim = max(maxim, angles[i] - angles[i - 1])
+            if maxim <= np.pi:
+                return 1
+            return np.sin(0.5 * maxim)
+        return 1
```
cduck: Code copying.  Same comments as in controlled_operation.py.

cirq/ops/common_gates_test.py
```diff
@@ -818,3 +818,32 @@ def test_rxyz_circuit_diagram():
...
+    assert cirq.trace_distance_bound(sswap) == 1
+    assert cirq.trace_distance_bound(siswap) == 1
+    assert cirq.trace_distance_bound(sh) == 1
+    assert cirq.trace_distance_bound(cirq.X) == 1
+    assert cirq.trace_distance_bound(cirq.Y**-1) == 1
+    assert cirq.trace_distance_bound(cirq.Z**0.5) == np.sin(np.pi / 4)
+    assert cirq.trace_distance_bound(cirq.H**0.25) == np.sin(np.pi / 8)
+    assert cirq.trace_distance_bound(cirq.CX**2) < 1.e-15
+    assert cirq.trace_distance_bound(cirq.CZ**(1 / 9)) == np.sin(np.pi / 18)
+    assert cirq.trace_distance_bound(cirq.SWAP**0.3) == np.sin(3 * np.pi / 20)
```
cduck: This is brittle.  Use approximate equality.

`3 * ... / 20` is confusing.  Use `0.3 * ... / 2`.

cirq/protocols/trace_distance_bound.py
```diff
@@ -57,6 +61,22 @@ def trace_distance_bound(val: Any) -> float:
     getter = getattr(val, '_trace_distance_bound_', None)
     result = NotImplemented if getter is None else getter()
 
-    if result is not NotImplemented and result < 1.0:
-        return result
+    if result is not NotImplemented:
+        return min(1.0, result)
+
+    if has_unitary(val):
+        u = unitary(val)
```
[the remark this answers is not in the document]
c-poole: Done

cirq/protocols/trace_distance_bound.py
```diff
@@ -57,6 +61,22 @@ def trace_distance_bound(val: Any) -> float:
...
+        if u.shape[0] == 2:
```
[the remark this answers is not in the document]
c-poole: Done

cirq/protocols/trace_distance_bound.py
```diff
@@ -57,6 +61,22 @@ def trace_distance_bound(val: Any) -> float:
     getter = getattr(val, '_trace_distance_bound_', None)
     result = NotImplemented if getter is None else getter()
 
-    if result is not NotImplemented and result < 1.0:
-        return result
+    if result is not NotImplemented:
+        return min(1.0, result)
+
+    if has_unitary(val):
```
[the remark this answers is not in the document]
c-poole: Done

cirq/protocols/trace_distance_bound.py
```diff
@@ -57,6 +61,22 @@ def trace_distance_bound(val: Any) -> float:
...
+            if squared <= 0:
+                return 0
+            return squared**0.5
+        angles = np.sort(np.angle(np.linalg.eigvals(u)))
+        maxim = 2 * np.pi + angles[0] - angles[-1]
+        for i in range(1, len(angles)):
+            maxim = max(maxim, angles[i] - angles[i - 1])
+        if maxim <= np.pi:
+            return 1
+        return np.sin(0.5 * maxim)
```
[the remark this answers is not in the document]
c-poole: Done

cirq/protocols/trace_distance_bound.py
```diff
@@ -57,6 +61,22 @@ def trace_distance_bound(val: Any) -> float:
...
+            squared = 1 - (0.5 * abs(u[0][0] + u[1][1]))**2
+            if squared <= 0:
+                return 0
+            return squared**0.5
+        angles = np.sort(np.angle(np.linalg.eigvals(u)))
+        maxim = 2 * np.pi + angles[0] - angles[-1]
+        for i in range(1, len(angles)):
+            maxim = max(maxim, angles[i] - angles[i - 1])
+        if maxim <= np.pi:
+            return 1
```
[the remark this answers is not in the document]
c-poole: Reformatted so all are floats

cirq/ops/parity_gates_test.py
```diff
@@ -267,3 +268,13 @@ def test_zz_diagrams():
...
+
+
+def test_trace_distance():
+    foo = sympy.Symbol('foo')
+    assert cirq.trace_distance_bound(cirq.XX**foo) == 1
+    assert cirq.trace_distance_bound(cirq.YY**foo) == 1
+    assert cirq.trace_distance_bound(cirq.ZZ**foo) == 1
+    assert cirq.trace_distance_bound(cirq.XX) == 1
+    assert cirq.trace_distance_bound(cirq.YY**0) < 1.e-15
+    assert cirq.trace_distance_bound(cirq.ZZ**2) < 1.e-15
```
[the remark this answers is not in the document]
c-poole: I added one non-extreme value here. These gates use the same method as the common gates where most of my non-extreme value tests were located.

cirq/ops/fsim_gate.py
```diff
@@ -79,6 +79,7 @@ def _unitary_(self) -> Optional[np.ndarray]:
             [0, 0, 0, c],
         ])
 
+
```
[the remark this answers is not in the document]
c-poole: good catch

cirq/ops/eigen_gate.py
```diff
@@ -293,11 +293,14 @@ def _value_equality_approximate_values_(self):
     def _trace_distance_bound_(self):
         if protocols.is_parameterized(self._exponent):
             return 1
-
-        angles = [half_turns for half_turns, _ in self._eigen_components()]
-        min_angle = min(angles)
-        max_angle = max(angles)
-        return abs((max_angle - min_angle) * self._exponent * 3.5)
+        angles = np.sort(self._eigen_shifts())
```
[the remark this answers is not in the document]
c-poole: This suggestion does fix the behavior

cirq/ops/controlled_operation.py
```diff
@@ -119,8 +119,22 @@ def _resolve_parameters_(self, resolver):
...
+        if self._is_parameterized_():
+            return 1
+        if protocols.has_unitary(self.sub_operation):
```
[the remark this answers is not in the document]
c-poole: Done

cirq/ops/controlled_operation.py
c-poole: It technically only needs the unitary of the sub_operation, so it isn't as expensive as it could have been. It is unfortunate though.

cirq/ops/controlled_operation.py
c-poole: Added annotations

cirq/ops/common_gates_test.py
c-poole: Done

cirq/ops/parallel_gate_operation_test.py
```diff
@@ -159,3 +159,13 @@ def test_equivalent_circuit():
...
+
+
+def test_trace_distance():
+    s = cirq.X**0.25
+    twoop = cirq.ParallelGateOperation(s, cirq.LineQubit.range(2))
+    threeop = cirq.ParallelGateOperation(s, cirq.LineQubit.range(3))
+    fourop = cirq.ParallelGateOperation(s, cirq.LineQubit.range(4))
+    assert cirq.trace_distance_bound(twoop) == np.sin(np.pi / 4)
+    assert cirq.trace_distance_bound(threeop) == np.sin(3 * np.pi / 8)
+    assert cirq.trace_distance_bound(fourop) == 1
```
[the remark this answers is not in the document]
c-poole: done

c-poole:
> I checked all your formulas are equivalent to `sin(arclength/2)` but is this equivalent to Wikipedia's definition? (https://en.wikipedia.org/wiki/Trace_distance)

Our definition of `trace_distance_bound` is the maximum possible trace distance between the input state and output state of a channel. This can be conveniently rephrased as the diamond norm distance between your channel and the identity channel (up to a factor of 2 depending on definitions). The diamond norm distance from a unitary channel and the identity channel is given by this sin(arclength/2) method.

> I don't think gates should return 1.0 from `_trace_distance_bound_` if they are parameterized. Return `None` instead to indicate that `cirq.trace_distance_bound` should not fallback to the unitary.

I'm don't agree. The purpose of the `_trace_distance_bound_` method is to provide a shortcut to calculating the trace distance if one is known. A parameterized gate can have a variable trace distance, and so the safest thing to return is 1.0. If we return 1.0 from the `_trace_distance_bound_`  method then the `unitary_strategy` won't be attempted.

> Write a function in cirq.testing like `assert_has_consistent_trace_distance_bound` that takes a value and verifies that `_trace_distance_bound_` is approximately greater than or equal to the bound calculated from its unitary. Add this function to `cirq.testing.consistent_protocols._assert_meets_standards_helper`.

Done

cduck:
> I'm don't agree. The purpose of the `_trace_distance_bound_` method is to provide a shortcut to calculating the trace distance if one is known. A parameterized gate can have a variable trace distance, and so the safest thing to return is 1.0. If we return 1.0 from the `_trace_distance_bound_` method then the `unitary_strategy` won't be attempted.

It's fine the way you have it but `cirq.unitary` does this by treating a return value of `None` different from `NotImplemented` (https://github.com/quantumlib/Cirq/blob/master/cirq/protocols/unitary.py#L124).

c-poole:
> It's fine the way you have it but `cirq.unitary` does this by treating a return value of `None` different from `NotImplemented` (https://github.com/quantumlib/Cirq/blob/master/cirq/protocols/unitary.py#L124).

In that case I will swap to a return value of `None` and refactor the `trace_distance_bound` protocol to a list of strategies in the same format as `cirq.unitary`. For now we only have two strategies but I hope to add more in the future.

cirq/ops/eigen_gate.py
```diff
@@ -290,17 +290,17 @@ def _value_equality_approximate_values_(self):
...
         return exponent, self._global_shift
 
-    def _trace_distance_bound_(self):
+    def _trace_distance_bound_(self) -> float:
         if protocols.is_parameterized(self._exponent):
-            return 1
-        angles = np.sort(self._eigen_shifts())
+            return 1.0
+        scaled_angles = np.array(self._eigen_shifts()) * self._exponent
+        angles = np.sort(scaled_angles - 2 * np.floor(scaled_angles * 0.5))
```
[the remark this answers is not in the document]
c-poole: It does, I'll go with that.

cirq/testing/consistent_protocols.py
```diff
@@ -92,6 +93,26 @@ def assert_eigen_shifts_is_consistent_with_eigen_components(
...
+    assert 0.0 <= val_from_trace <= 1.0
+    if u is not None:
+
+        class Unitary:
+
+            def __init__(self, unitary):
+                self._unitary = unitary
+
+            def _unitary_(self):
+                return self._unitary
```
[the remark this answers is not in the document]
c-poole: Done

cirq/testing/consistent_protocols.py
```diff
@@ -92,6 +93,26 @@ def assert_eigen_shifts_is_consistent_with_eigen_components(
     assert val._eigen_shifts() == [e[0] for e in val._eigen_components()]
 
 
+def assert_has_consistent_trace_distance_bound(val: Any,) -> None:
```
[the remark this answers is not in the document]
c-poole: good catch

cirq/ops/controlled_operation.py
c-poole: I went with a utility method in `trace_distance_bound` that the relevant files import.

cirq/ops/eigen_gate_test.py
```diff
@@ -210,7 +210,8 @@ def test_inverse():
 
 def test_trace_distance_bound():
     assert cirq.trace_distance_bound(CExpZinGate(0.001)) < 0.01
-    assert cirq.trace_distance_bound(CExpZinGate(sympy.Symbol('a'))) >= 1
+    assert cirq.trace_distance_bound(CExpZinGate(sympy.Symbol('a'))) == 1
+    assert cirq.trace_distance_bound(CExpZinGate(2)) == 1
```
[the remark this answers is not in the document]
c-poole: Added similar tests

cirq/ops/controlled_gate.py
c-poole: added utility method that controlled_gate imports to avoid the copying problem.

cirq/ops/pauli_string_phasor.py
```diff
@@ -131,6 +131,8 @@ def _circuit_diagram_info_(self, args: protocols.CircuitDiagramInfoArgs
                                                exponent=self.exponent_relative)
 
     def _trace_distance_bound_(self) -> float:
+        if self._num_qubits_() == 0:
```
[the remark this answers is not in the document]
c-poole: switched to len(self.qubits)

cirq/protocols/trace_distance_bound.py
```diff
@@ -83,7 +104,17 @@ def _strat_distance_from_unitary(val: any) -> Optional[float]:
...
+
+
+def trace_distance_from_angle_list(angle_list: Iterable[float]) -> float:
+    """ Given a list of arguments of the eigenvalues of a unitary matrix,
+    calculates the trace_distance_bound of the unitary effect.
+
+    The maximum provided angle should not exceed the minimum provided angle
+    by more than two pi.
+    """
+    angles = np.sort(angle_list)
```
cduck: `np.sort` only works on sequences.

cirq/protocols/trace_distance_bound.py
```diff
@@ -83,7 +104,17 @@ def _strat_distance_from_unitary(val: any) -> Optional[float]:
             return 0.0
         return squared**0.5
 
-    angles = np.sort(np.angle(np.linalg.eigvals(u)))
+    return trace_distance_from_angle_list(np.angle(np.linalg.eigvals(u)))
+
+
+def trace_distance_from_angle_list(angle_list: Iterable[float]) -> float:
+    """ Given a list of arguments of the eigenvalues of a unitary matrix,
```
cduck: Style nit: Leading space before `Given`.

cirq/protocols/trace_distance_bound.py
```diff
@@ -83,7 +104,17 @@ def _strat_distance_from_unitary(val: any) -> Optional[float]:
...
+    calculates the trace_distance_bound of the unitary effect.
```
cduck: Remove underscores or put in a code block.

cirq/protocols/trace_distance_bound.py
```diff
@@ -83,7 +104,17 @@ def _strat_distance_from_unitary(val: any) -> Optional[float]:
...
+
+    The maximum provided angle should not exceed the minimum provided angle
+    by more than two pi.
```
cduck: Use LaTeX or unicode for pi.  Most of the docstrings in Cirq use `π`.

cirq/ops/fsim_gate_test.py
```diff
@@ -217,9 +217,20 @@ def test_fsim_unitary():
     )
 
 
-def test_trace_distance():
-    foo = sympy.Symbol('foo')
-    assert cirq.trace_distance_bound(cirq.FSimGate(foo, foo)) == 1
-    assert cirq.trace_distance_bound(cirq.FSimGate(np.pi + 0.01, 0)) == 1
-    assert np.isclose(cirq.trace_distance_bound(cirq.FSimGate(0, np.pi / 2)),
-                      np.sin(np.pi / 4))
```
cduck: Did this get accidentally deleted?

cirq/protocols/trace_distance_bound.py
c-poole: I'll correct the type annotations accordingly

cirq/protocols/trace_distance_bound.py
c-poole: Fixed

cirq/protocols/trace_distance_bound.py
c-poole: Done

cirq/protocols/trace_distance_bound.py
c-poole: swapped to symbol

cirq/ops/fsim_gate_test.py
c-poole: I originally had some edits in fsim_gate that got removed and therefore these additional tests became unnecessary. I only recently noticed that while I was resolving a merge conflict.

cirq/protocols/__init__.py
```diff
@@ -84,10 +84,9 @@
     SupportsQasmWithArgs,
     SupportsQasmWithArgsAndQubits,
 )
-from cirq.protocols.trace_distance_bound import (
-    SupportsTraceDistanceBound,
-    trace_distance_bound,
-)
+from cirq.protocols.trace_distance_bound import (SupportsTraceDistanceBound,
+                                                 trace_distance_bound,
+                                                 trace_distance_from_angle_list)
```
cduck: Add a trailing comma to make yapf not do this.

cirq/protocols/trace_distance_bound.py
```diff
@@ -38,25 +40,85 @@ def _trace_distance_bound_(self) -> float:
...
-    before and after the effect.
+    and output.
+
+    This method attempts a number of strategies to calculate this value.
+
+    Strategy 1:
+        Use the effect's `_trace_distance_bound_` method.
+
+    Strategy 2:
+        If the effect is unitary, calculate the trace_distance_bound from the
```
cduck: Unnecessary underscores.

cirq/protocols/trace_distance_bound.py
```diff
@@ -38,25 +40,85 @@ def _trace_distance_bound_(self) -> float:
...
+
+    return trace_distance_from_angle_list(np.angle(np.linalg.eigvals(u)))
+
+
+def trace_distance_from_angle_list(angle_list: Sequence[float]) -> float:
+    """Given a list of arguments of the eigenvalues of a unitary matrix,
+    calculates the trace distance bound of the unitary effect.
+
+    The maximum provided angle should not exceed the minimum provided angle
+    by more than two π.
```
cduck: Nit: 2π

cirq/ops/eigen_gate_test.py
```diff
@@ -224,7 +224,7 @@ def _eigen_components(self):
                 (0, np.array([[1, 0], [0, 0]])),
                 (12, np.array([[0, 0], [0, 1]])),
             ]
-
+    assert E()._num_qubits_() == 1
```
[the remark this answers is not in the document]
c-poole: It's necessary for the coverage test to pass. The E gate is never asked how many qubits it has anywhere else since it is only defined here.

cirq/protocols/__init__.py
c-poole: done

cirq/protocols/trace_distance_bound.py
c-poole: fixed

cirq/protocols/trace_distance_bound.py
c-poole: done

cirq/ops/eigen_gate_test.py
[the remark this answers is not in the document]
cduck: Right.  You can use `# coverage: ignore` for cases like this.

PR #1851, opened 2019-07-19
Fix SupportsPhase documentation

kevinsung:


cirq/protocols/phase.py
```diff
@@ -32,8 +32,8 @@ def _phase_by_(self: Any, phase_turns: float, qubit_index: int):
 
         Specifically, returns an object with matrix P U P^-1 (up to global
         phase) where U is the given object's matrix and
-        P = Z(qubit_index)**(phase_turns/2). For example, an X gate phased by 90
-        degrees would be a Y gate.
+        P = Z(qubit_index)**(2 * phase_turns). For example, an X gate phased
```
dabacon: I'm not sure this is correct.
```
import cirq
q = cirq.NamedQubit('a')
print(cirq.phase_by(cirq.X(q), 0.25, 0))
print(cirq.Z(q)**(0.5) * cirq.X(q) * cirq.Z(q)**(-0.5))
```
yields
```
Y(a)
-X(a)
```

Maybe you are getting the `2 * phase turns` from the PhasedXGate definition?

kevinsung: There must be a bug in the multiplication of Paulis, since you can calculate by hand that Z**0.5 * X * Z**-0.5 is equal to Y. Indeed,
```
import cirq
q = cirq.NamedQubit('a')
c = cirq.Circuit.from_ops(cirq.Z(q)**-0.5, cirq.X(q), cirq.Z(q)**0.5)
print(cirq.unitary(c))
```
yields
```
[[0.+0.j 0.-1.j]
 [0.+1.j 0.+0.j]]
```

kevinsung: Resolved by #1896

PR #1852, opened 2019-07-19
Move api commands into api/v1 like v2

dabacon:
Note that I make Zip and Product visible.  Not sure if not having these was intentional or not.


dabacon:
Ping.

PR #1853, opened 2019-07-19
Add commutes protocol

### LINKED ISSUE #1125: Should we add a `cirq.commutes`/`_commutes_(other)` protocol?
Knowing commutation information can be helpful when working with circuits, and it seems likely that we will need it. For example, we may want a stronger version of "place at earliest possible position" where CZ gates being added actually commute over Z gates and other CZs in their way.

@dabacon opinions?

bryano:
Minimal version so far. Wanted to get some feedback.

Some thoughts:
- A lot of issues seemed to be caused by `values_equality` being the name of both a module and a method, so we should probably change at least one of those for `commutes`.
- Maybe `_commutes_with_` for the magic method? (I just went with convention in the initial issue for now.)
- I changed `cirq.commutes` to be the protocol, which delegates to `cirq.linalg.commutes` in the relevant cases. Shouldn't be a problem, but may cause confusion.
- Some classes already define a `commute_with` method. Should I add some underscores to that or add a new method that delegates to the existing one?
- If passed a `Gate` and `GateOperation` whose gate commutes with the `Gate`, what should be returned?

Fixes #1125


cirq/protocols/commutes.py
```diff
@@ -0,0 +1,92 @@
...
+        getter = getattr(a, '_commutes_', None)
+        if getter is None:
+            continue
+        val = getter(right_val, atol)
+        if val is not NotImplemented:
+            return val
+    return NotImplemented
+
+
+def _strat_commutes_from_matrix(left_val: Any, right_val: Any, atol: Union[int, float] = 1e-8
```
c-poole: This doesn't appear to take into consideration the possibility that `left_val` and `right_val` are operations that act on different qubits.

cirq/protocols/commutes.py
```diff
@@ -0,0 +1,92 @@
...
+        if result is not NotImplemented:
+            return result
+    return False
+
+
+def _strat_commutes_from_commutes(left_val: Any, right_val: Any, atol: Union[int, float] = 1e-8
+        ) -> Union[NotImplementedType, bool]:
+    """Attempts to determine commutativity via the objects' _commutes_ method."""
+    for a, b in [(left_val, right_val), (right_val, left_val)]:
+        getter = getattr(a, '_commutes_', None)
```
c-poole: +1 on changing to `_commutes_with_` and refactoring classes that already have this semantics.

cirq/protocols/commutes.py
c-poole: As an extension to this train of thought, I think that if one is a `Gate_Operation` and the other is a `Gate` then the protocol should treat it as a strictest case scenario where the `Gate` acts on the same qubits as the `Gate_Operation`.

bryano: The more I think about it, the less I think this general strategy is worth it. Makes more sense to have all the logic in the `_commutes_with_` method of the various objects. For example, qubit order using the unitary protocol is tricky with circuits.

cirq/protocols/commutes.py
```diff
@@ -0,0 +1,92 @@
...
+from cirq.protocols.unitary import unitary
+from cirq.type_workarounds import NotImplementedType
+
+
+def commutes(left_val: Any, right_val, atol: Union[int, float] = 1e-8) -> bool:
+    """Determines whether two values commute.
+
+    This is determined by any one of the following techniques:
+
+    - Either value has a `_commutes_` method that returns something besides
```
mpharrigan: note that left_val is considered first

cirq/protocols/commutes.py
```diff
@@ -0,0 +1,92 @@
...
+      NotImplemented. The return value is whatever the method returned.
+    - Each value is either a matrix or defines one via the unitary protocol.
```
mpharrigan: split this into two cases? 1. is a matrix 2. defines one via unitary protocol

cirq/protocols/commutes.py
```diff
@@ -0,0 +1,92 @@
...
+from cirq.type_workarounds import NotImplementedType
+
+
+def commutes(left_val: Any, right_val, atol: Union[int, float] = 1e-8) -> bool:
+    """Determines whether two values commute.
+
+    This is determined by any one of the following techniques:
+
+    - Either value has a `_commutes_` method that returns something besides
+      NotImplemented. The return value is whatever the method returned.
```
mpharrigan: "whatever the method returned" -> "the boolean result from this function" or something to emphasize that you can't return whatever you want in `_commutes_`

cirq/protocols/commutes.py
```diff
@@ -0,0 +1,92 @@
...
+
+    strats = [
+        _strat_commutes_from_commutes,
+        _strat_commutes_from_matrix,
+    ]
+    for strat in strats:
+        result = strat(left_val, right_val, atol)
+        if result is not NotImplemented:
+            return result
+    return False
```
mpharrigan: is this a sensible default? Why not raise a `NotImplementedError` or return `NotImplemented` all the way up to the caller? `cirq.unitary` doesn't return the identity matrix if it can't figure out the unitary representation of an object :)

cirq/protocols/commutes.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def _strat_commutes_from_commutes(left_val: Any, right_val: Any, atol: Union[int, float] = 1e-8
+        ) -> Union[NotImplementedType, bool]:
+    """Attempts to determine commutativity via the objects' _commutes_ method."""
+    for a, b in [(left_val, right_val), (right_val, left_val)]:
+        getter = getattr(a, '_commutes_', None)
+        if getter is None:
+            continue
+        val = getter(right_val, atol)
```
mpharrigan: this should be `b`

mpharrigan:
> Some classes already define a commute_with method. Should I add some underscores to that or add a new method that delegates to the existing one?

I'd advocate for adding underscores, keeping the old signature which dispatches to the underscore version with a deprecation warning for ~1 release cycle

cirq/protocols/commutes.py
bryano: I imagine this being used for circuit reordering, where what you want to know whether two things _definitely_ commute. (Note that this is the documented behavior.)

Do you have an example where false negatives would be problematic?

cirq/protocols/commutes.py
bryano: Why?

bryano:
> > Some classes already define a commute_with method. Should I add some underscores to that or add a new method that delegates to the existing one?
> 
> I'd advocate for adding underscores, keeping the old signature which dispatches to the underscore version with a deprecation warning for ~1 release cycle

Done

cirq/protocols/commutes.py
bryano: Done

cirq/protocols/commutes.py
bryano: Obsoleted (delegation to unitary moved to `Gate._commutes_with_`)

cirq/protocols/commutes.py
bryano: Done

cirq/protocols/commutes.py
bryano: Done

bryano:
Ready for re-review.

dabacon:
I am worried about making this a protocol.  This is because this has an n^2 problem, every gate needs to implement it for every other gate (for the fast version to work).  If we only partially implement it, then it can lead to surprising slow results.  

I feel like maybe because of this there is another object that one needs to use to be explicit about this. I think this may be another case where gate set helps.  Gate sets are contained versions of the n in n^2, and the gateset could be responsible for knowing this.

cirq/contrib/acquaintance/permutation.py
```diff
@@ -145,6 +145,14 @@ def __repr__(self):
     def _value_equality_values_(self):
         return (self.swap_gate,)
 
+    def _commutes_with_(self, other: Any, atol: Union[int, float] = 1e-8
+                       ) -> Union[bool, NotImplementedType]:
+        if (isinstance(other, ops.Gate) and
+                isinstance(other, ops.InterchangeableQubitsGate) and
+                protocols.num_qubits(other) == 2):
+            return True
```
Strilanc: Is this correct? Swap gates only commute if they are on the same qubits or completely disjoint qubits. There's also the issue that the given swap gate may be asymmetric, but it could be considered a contract of the class.

Add an "assert_commutes_with_is_consistent_with_unitary" testing method to `cirq.testing`, which validates that when `_commutes_with_` returns false or true that it agrees with the results of the fallback computation on unitaries.

cirq/ops/raw_types.py
```diff
@@ -319,6 +321,17 @@ def _qid_shape_(self) -> Tuple[int, ...]:
         (3, 3) for a 2-qutrit ternary gate.
         """
 
+    def _commutes_with_(self, other: Any, atol: Union[int, float] = 1e-8
+                       ) -> Union[bool, NotImplementedType]:
+        if (qid_shape_protocol.qid_shape(self, None) !=
+                qid_shape_protocol.qid_shape(other, None)):
+            return NotImplemented
+        self_unitary = protocols.unitary(self, None)
```
Strilanc: This fallback-to-unitary code should go into the global `cirq.commutes` method. If the objects both specify a qid_shape but it differs, they should not commute.

cirq/ops/raw_types.py
```diff
@@ -319,6 +321,17 @@ def _qid_shape_(self) -> Tuple[int, ...]:
         (3, 3) for a 2-qutrit ternary gate.
         """
 
+    def _commutes_with_(self, other: Any, atol: Union[int, float] = 1e-8
+                       ) -> Union[bool, NotImplementedType]:
+        if (qid_shape_protocol.qid_shape(self, None) !=
```
Strilanc: Note that `other` may be an operation, or some other object.

Whether or not two gates commute depends on whether or not they are applied to the same qubits, and even the order of those qubits. We need to account for this somehow.

Perhaps we should define a `_commutes_on_qubits_(self, qubits, other)` method. `GateOperation._commutes_` would check for the existence of this method, and forward along if appropriate.

cirq/ops/raw_types.py
```diff
@@ -319,6 +321,17 @@ def _qid_shape_(self) -> Tuple[int, ...]:
         (3, 3) for a 2-qutrit ternary gate.
         """
 
+    def _commutes_with_(self, other: Any, atol: Union[int, float] = 1e-8
```
Strilanc: I forget why we picked `_commutes_with_` instead of just `_commutes_`.

cirq/protocols/commutes.py
```diff
@@ -0,0 +1,104 @@
...
+
+from cirq import linalg
+from cirq.type_workarounds import NotImplementedType
+
+TDefault = TypeVar('TDefault')
+
+
+def commutes(left_val: Any,
+             right_val: Any,
+             atol: Union[int, float] = 1e-8,
```
Strilanc: Force keywords for `atol` and `default`.

cirq/protocols/commutes.py
```diff
@@ -0,0 +1,104 @@
...
+             default: TDefault = NotImplemented) -> Union[bool, TDefault]:
```
Strilanc: I think the default behavior should be to raise an exception, as with other protocols. It is too likely that someone will mistakenly write `if not cirq.commutes(a, b): print("does not commute therefore [MISTAKEN CONCLUSION]")`.

cirq/contrib/acquaintance/permutation.py
bryano: At this point, gate commutation checks assume the gates act on exactly the same set of qubits.

cirq/protocols/commutes.py
bryano: Any suggestion for the default value? It just needs to be something with a unique id, right?

bryano:
What should `commutes(A, B)` mean when `A` and `B` are both gates?

(1) They have the same `qid_shape` and commute when applied to the same qudits; indeterminate when different `qid_shape`.

(2) They commute for all possible qubits.

For example `CNOT` commutes with itself under (1) but not (2), while `Z` and `ZZ` commute under (2) but not (1).

cirq/ops/raw_types.py
bryano: Is there a difference between `_commutes_on_qubits_` and `_commutes_with_operation_`?

Strilanc:
Good question. I think I like (1) best. Specifically:

```
if qid_shape(a) != qid_shape(b):
    raise ValueError("Different sized gates.")
qs = cirq.LineQid(i, s) for i, s in enumerate(qid_shape(a))
return commutes(a(*qs), b(*qs))
```

cduck:
@Strilanc This is what `LineQid.for_qid_shape` is for:
```
qs = cirq.LineQid.for_qid_shape(qid_shape)
```

cirq/protocols/commutes.py
bryano: Done

cirq/ops/raw_types.py
bryano: Went with `_commutes_` everywhere.

cirq/ops/raw_types.py
bryano: This seems dangerous. For example, CNOT(a, b) and CNOT(b, a) have the same unitary. We could make sure that `GateOperation._commutes_` addresses this and that the `_commutes_` strategy precedes the unitary strategy in `cirq.commutes`, but that would be brittle.

cirq/ops/raw_types.py
bryano: Went with `_commutes_on_qids_`.

bryano:
@Strilanc PTAL

vtomole:
@Strilanc  We can't merge anything until https://github.com/quantumlib/Cirq/pull/2648 is merged.

PR #1854, opened 2019-07-19
fixes bug in Circuit.findall_operations_until_blocked

bryano:
Prior implementation double-counted operations because it iterated over qubits rather than moments.

### DIFF
--- a/cirq/circuits/circuit_test.py
+++ b/cirq/circuits/circuit_test.py
@@ -1183,6 +1183,16 @@ def test_findall_operations_until_blocked():
             start_frontier={a: idx},
             is_blocker=stop_if_h) == [(11, cirq.CZ.on(a,b))]
 
+    circuit = cirq.Circuit.from_ops(
+        [cirq.CZ(a, b), cirq.CZ(a, b),
+         cirq.CZ(b, c)])
+
+    start = {a: 0, b: 0}
+    is_blocker = lambda next_op: sorted(next_op.qubits) != [a, b]
+    assert (circuit.findall_operations_until_blocked(start, is_blocker) == [
+        (i, cirq.CZ(a, b)) for i in (0, 1)
+    ])
+
 
 def test_has_measurements():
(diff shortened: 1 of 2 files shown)


cirq/circuits/circuit_test.py
```diff
@@ -1183,6 +1183,16 @@ def test_findall_operations_until_blocked():
...
             is_blocker=stop_if_h) == [(11, cirq.CZ.on(a,b))]
 
+    circuit = cirq.Circuit.from_ops(
+        [cirq.CZ(a, b), cirq.CZ(a, b),
+         cirq.CZ(b, c)])
+
+    start = {a: 0, b: 0}
+    is_blocker = lambda next_op: sorted(next_op.qubits) != [a, b]
+    assert (circuit.findall_operations_until_blocked(start, is_blocker) == [
+        (i, cirq.CZ(a, b)) for i in (0, 1)
```
dstrain115: I think it may be better in tests to list the operations out, especially since there are only two of them.

bryano:
> Can you make sure to add more information about the bug in the final comment message, so it's clear what the bug is.

You mean commit message? Is what I wrote in PR description sufficient?

bryano:
What do you think about adding a flag argument that makes the inclusion of the moment indices optional? With the new implementation, the operations returned are guaranteed to be in "temporal" order.

dstrain115:
> What do you think about adding a flag argument that makes the inclusion of the moment indices optional? With the new implementation, the operations returned are guaranteed to be in "temporal" order.

How would that work exactly?  Wouldn't that change the return type?

bryano:
Yes, it would change the return type to `Union[List[Tuple[int, Operation]], List[Operation]]`. On further thought, I don't think it's worth it at this point. The user can just neglect the moment indices if they don't need them.

Just let me know about the commit message.

dstrain115:
Commit message is fine. (Maybe I missed it the first time).  Go ahead!

PR #1857, opened 2019-07-22
Add __str__ to DensityMatrixTrialResult

### LINKED ISSUE #1705: Create nice str representation for DensityMatrixTrialResult
It has a repr, but not a good str.

Should work similar to WaveFunctionTrialResult.

bradyb:
Closes #1705.


cirq/sim/density_matrix_simulator_test.py
```diff
@@ -656,3 +656,17 @@ def test_works_on_pauli_string():
     np.testing.assert_allclose(result.reshape(4, 4),
                                np.diag([0, 0, 0, 1]),
                                atol=1e-8)
+
+
+def test_str():
```
vtomole: Rename to `test_density_matrix_trial_result_str()`

cirq/sim/density_matrix_simulator_test.py
```diff
@@ -656,3 +656,17 @@ def test_works_on_pauli_string():
...
+    q0 = cirq.LineQubit(0)
+    final_simulator_state = cirq.DensityMatrixSimulatorState(
+        density_matrix=np.ones((2, 2)) * 0.5, qubit_map={q0: 0})
+    result = cirq.DensityMatrixTrialResult(
+        params=cirq.ParamResolver({}),
+        measurements={},
+        final_simulator_state=final_simulator_state)
+
+    assert str(result).startswith('measurements: (no measurements)\n'
```
vtomole: Why does this call `starts_with`?  Use `==` instead.

cirq/sim/density_matrix_simulator.py
```diff
@@ -620,6 +620,11 @@ def _value_equality_values_(self):
                         sorted(self.measurements.items())}
         return (self.params, measurements, self._final_simulator_state)
 
+    def __str__(self):
+        samples = super().__str__()
+        return 'measurements: {}\nfinal density matrix:\n{}'.format(
+            samples, str(self.final_density_matrix))
```
vtomole: I'm not sure if calling `str` on `final_density_matrix` is necessary here with the call to `.format` and all. What do you think @cduck?

cduck: `str` isn't needed because format does this for you when you specify `'{}'` or `'{!s}'`.  Use `'{!r}'` if you want the repr.

cirq/sim/density_matrix_simulator_test.py
bradyb: Done.

cirq/sim/density_matrix_simulator_test.py
bradyb: Done.

cirq/sim/density_matrix_simulator.py
bradyb: Cool, done.

PR #1863, opened 2019-07-23
Add pip package configurations for contrib and dev tools

### LINKED ISSUE #843: Split cirq packages into with/without contrib
Otherwise there's no way to easily pip install the contrib-requirements

cduck:
Fixes #843.

This allows users to easily install Cirq with the contrib dependencies:
```
$ pip install cirq[contrib]
```

To install Cirq in editable mode for development:
```
$ pip install -e .[dev,contrib]
```

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -89,9 +89,7 @@ See the previous section for instructions.
     ```bash
     mkvirtualenv cirq-py3 --python=/usr/bin/python3
     python -m pip install --upgrade pip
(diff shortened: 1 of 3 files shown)


c-poole:
We would probably benefit from a [dev] version of this that installs everything with a single pip command as well!

cduck:
Good point.

vtomole:
@Strilanc Oh right. We have a stable readthedocs.

PR #1871, opened 2019-07-25
Add direct support for 3-legged OAuth

wcourtney:
Support 3-legged flows across multiple user projects, by attributing requests to the resource project rather than the project that owns the OAuth client. This moves project specification to a required attribute of the Engine and is used as the single source of truth for consistency. Utilities related to the project were also updated to remove redundant specification of the project.


cirq/google/engine/engine.py
```diff
@@ -61,6 +62,20 @@ class ProtoVersion(enum.Enum):
 Program = Union[circuits.Circuit, Schedule]
 
 
+def user_project_header_request_builder(project_id):
+    """Provides a request builder that sets a user project header on engine
+    requests to allow using standard OAuth credentials.
+    """
+
+    def request_builder(*args, **kwargs):
+        from apiclient import http as apiclient_http
```
dabacon: is there a reason this import is here?  If so document, if not move up with other imports

cirq/google/engine/engine.py
```diff
@@ -61,6 +62,20 @@ class ProtoVersion(enum.Enum):
 Program = Union[circuits.Circuit, Schedule]
 
 
+def user_project_header_request_builder(project_id):
```
dabacon: annotate types

cirq/google/engine/engine.py
```diff
@@ -152,47 +156,43 @@ class Engine:
...
+                 project_id: str,
                  version: str = 'v1alpha1',
-                 default_project_id: Optional[str] = None,
                  discovery_url: Optional[str] = None,
                  default_gcs_prefix: Optional[str] = None,
                  proto_version: ProtoVersion = ProtoVersion.V1,
                  **kwargs) -> None:
         """Engine service client.
 
+        Requires project_id.
```
dabacon: Don't think you have to say this since any field without a default value will yell in python

cirq/google/engine/engine.py
dabacon: if we don't expected users to use this directly, make this "private" by leading "_"

cirq/google/engine/engine.py
```diff
@@ -152,47 +156,43 @@ class Engine:
...
-            **kwargs)
+        request_builder = user_project_header_request_builder(project_id)
+        service_args = {
+            'requestBuilder': request_builder,
+        }
+        service_args.update(kwargs)
+
+        self.service = discovery.build('' if discovery_url else 'quantum',
+                                       '' if discovery_url else version,
+                                       discoveryServiceUrl=self.discovery_url,
```
Strilanc: Just do `requestBuilder=request_builder, **kwargs`, instead of making a new dictionary and folding in the kwargs.

cirq/google/engine/engine.py
```diff
@@ -152,47 +156,43 @@ class Engine:
...
     def __init__(self,
-                 api_key: Optional[str] = None,
-                 api: str = 'quantum',
+                 project_id: str,
                  version: str = 'v1alpha1',
-                 default_project_id: Optional[str] = None,
                  discovery_url: Optional[str] = None,
                  default_gcs_prefix: Optional[str] = None,
                  proto_version: ProtoVersion = ProtoVersion.V1,
                  **kwargs) -> None:
```
Strilanc: Do we still need this kwargs argument? It strikes me as very "asking for what you want to make what you want" instead of "ask for what you need".

cirq/google/engine/engine.py
```diff
@@ -375,14 +354,19 @@ def run_sweep(
...
+            'name':
+            'projects/%s/programs/%s' % (
+                self.project_id,
+                job_config.program_id,
+            ),
+            'gcs_code_location': {
+                'uri': job_config.gcs_program
+            },
+            'code':
+            code,
```
Strilanc: The formatter wanted this on a new line?

cirq/google/engine/engine_test.py
```diff
@@ -180,11 +180,9 @@
 
 
 def test_repr():
-    v = cirq.google.JobConfig(project_id='my-project-id',
-                              program_id='my-program-id',
-                              job_id='my-job-id')
+    v = cirq.google.JobConfig(program_id='my-program-id', job_id='my-job-id')
 
-    assert repr(v) == ("JobConfig(project_id='my-project-id', "
+    assert repr(v) == ("JobConfig("
```
Strilanc: Use `cirq.testing.assert_equivalent_repr(v)` instead of asserting the specific details. You likely need to change the repr to say `cirq.google.JobConfig`.

cirq/google/engine/engine.py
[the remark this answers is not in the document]
wcourtney: I followed precedent I saw elsewhere, but I prefer it gone too. Done.

cirq/google/engine/engine.py
```diff
@@ -15,8 +15,9 @@
 """Classes for running against Google's Quantum Cloud Service.
 
 As an example, to run a circuit against the xmon simulator on the cloud,
-    engine = cirq.google.Engine(api_key='mysecretapikey')
-    options = cirq.google.JobConfig(project_id='my-project-id')
+    engine = cirq.google.Engine(
+        project-id='my-project-id')
```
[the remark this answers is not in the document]
wcourtney: Done: s/project-id/project_id

cirq/google/engine/engine.py
```diff
@@ -152,47 +156,43 @@ class Engine:
...
                  proto_version: ProtoVersion = ProtoVersion.V1,
                  **kwargs) -> None:
         """Engine service client.
 
+        Requires project_id.
+
         Args:
-            api_key: API key to use to retrieve discovery doc.
-            api: API name.
+            project_id: A project_id string of the Google Cloud Project to use.
```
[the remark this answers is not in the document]
wcourtney: Done.

cirq/google/engine/engine.py
wcourtney: I did it this way to allow the caller to override the requestBuilder. I think that the recommendation would break that ability since the requestBuilder argument would be specified twice. WDYT?

cirq/google/engine/engine.py
wcourtney: We use this today for alternative auth configurations that haven't been documented and I'm not sure we'll explicitly try to manage with our interface. We could point to the googleapiclient docs if that's useful?

cirq/google/engine/engine.py
wcourtney: Must have. I bumped it up again; we'll see what happens.

cirq/google/engine/engine.py
wcourtney: Moved up.

cirq/google/engine/engine.py
wcourtney: Done

cirq/google/engine/engine.py
wcourtney: Done

cirq/google/engine/engine.py
wcourtney: Should be private. Fixed

cirq/google/engine/engine.py
```diff
@@ -152,47 +156,43 @@ class Engine:
...
+                API interactions will be attributed to this project and any
+                resources created will be owned by the project.
             version: API version.
-            default_project_id: A fallback project_id to use when one isn't
-                specified in the JobConfig given to 'run' methods.
-                See JobConfig for more information on project_id.
-            discovery_url: Discovery url for the API. If not supplied, uses
-                Google's default api.googleapis.com endpoint.
+            discovery_url: Discovery url for the API to select a non-default
```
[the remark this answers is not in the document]
wcourtney: Made a pass at this. LMK what you think.

cirq/google/engine/engine_test.py
wcourtney: It turns out that this tests for equality of the object as evaluated from the repr, so I also had to provide equality on the JobConfig class as well. Done.

cirq/google/engine/engine.py
dabacon: I think we do need it but we should better document what it is.  Could you add a doc string.  Maybe also just call it service_args.  Also it could just be a dict and not a **kwargs, I guess.

wcourtney: Makes sense, I added a dict service_args and documented it.

cirq/google/engine/engine.py
wcourtney: Changed the format to add iff requestBuilder is not already defined on service_args.

PR #1872, opened 2019-07-25
Fix pytest.raises calls in permutation_test.py

maffoo:
Currently unable to validate a dev package build because these tests fail.

### DIFF
--- a/cirq/contrib/acquaintance/permutation_test.py
+++ b/cirq/contrib/acquaintance/permutation_test.py
@@ -46,14 +46,14 @@ def test_validate_permutation_errors():
     validate_permutation({})
 
     with pytest.raises(IndexError,
-                       message='key and value sets must be the same.'):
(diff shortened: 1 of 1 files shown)


cirq/contrib/acquaintance/permutation_test.py
```diff
@@ -46,14 +46,14 @@ def test_validate_permutation_errors():
     validate_permutation({})
 
     with pytest.raises(IndexError,
-                       message='key and value sets must be the same.'):
+                       match='key and value sets must be the same.'):
```
cduck: Escape the '.' in the regex expression.

maffoo: done.

maffoo:
The build succeeds, but `dev_tools/packaging/verify-published-package.sh` fails because it runs contrib tests.

PR #1874, opened 2019-07-25
Add type annotations in cirq/google/op_deserializer.py

Strilanc:


dabacon:
mypy says 
```cirq/google/op_deserializer.py:53: error: Too many positional arguments for "__new__" of "DeserializingArg"```

PR #1875, opened 2019-07-25
Add cirq.google.QuantumEngineSampler

Strilanc:
~~COMES AFTER https://github.com/quantumlib/Cirq/pull/1871 (currently conflicts with)~~


cirq/google/engine/engine_sampler.py
```diff
@@ -0,0 +1,59 @@
...
+    """A sampler that samples from real hardware on the quantum engine."""
+
+    def __init__(self,
+                 *,
+                 google_cloud_project_id: str,
+                 processor_id: Union[str, List[str]],
+                 gate_set: 'cirq.google.SerializableGateSet'):
+        """
+        Args:
+            google_cloud_project_id: The identifier of the google cloud project
```
wcourtney: Should this just be 'project_id' to match the existing identifiers in the Engine?

cirq/google/engine/engine_sampler.py
```diff
@@ -0,0 +1,59 @@
...
+                that quantum engine is being used in.
+            processor_id: String identifier, or list of string identifiers,
+                determining which processors may be used when sampling.
+            gate_set: Determines how to serialize circuits when requesting
+                samples.
+        """
+        self.google_cloud_project_id = google_cloud_project_id
+        self.processor = (
```
wcourtney: I think this should be plural since it's always a list.

cirq/google/engine/engine_sampler.py
```diff
@@ -0,0 +1,59 @@
...
+        return request
+
+    def run_sweep(
+            self,
+            program: Union['cirq.Circuit', 'cirq.Schedule'],
+            params: 'cirq.Sweepable',
+            repetitions: int = 1,
+    ) -> List['cirq.TrialResult']:
+
+        runner = engine.Engine(
```
wcourtney: What do you think of adding a **kwargs to the sampler that can be propagated to the Engine so that we can, e.g., support alternative auth mechanisms? If that's alright with you, then maybe drop the requestBuilder since it will be added in https://github.com/quantumlib/Cirq/pull/1871 and would be supported using **kwargs?

wcourtney: Consider making this be a private member of QuantumEngineSampler so that you don't need to retrieve the discovery doc each time you call run_sweep().

cirq/google/engine/engine_sampler.py
dabacon: link to google cloud documentation of what this id is

cirq/google/engine/engine_sampler.py
dabacon: why is the engine constructed for each run call?

cirq/google/engine/engine_sampler.py
```diff
@@ -0,0 +1,59 @@
...
+            proto_version=engine.ProtoVersion.V2,
```
dabacon: this means people can't update the proto version?  Think you want to propogate some kwargs**

cirq/google/engine/engine_sampler.py
Strilanc: Obsoleted by taking `engine` as an argument.

cirq/google/engine/engine_sampler.py
Strilanc: Done.

cirq/google/engine/engine_sampler.py
Strilanc: Obsoleted by taking engine as an argument to the init.

Strilanc: Fixed.

cirq/google/engine/engine_sampler.py
Strilanc: Obsoleted by taking engine as a constructor argument.

cirq/google/engine/engine_sampler.py
Strilanc: Obsoleted by taking engine as a constructor argument.

cirq/google/engine/engine_sampler.py
```diff
@@ -0,0 +1,44 @@
+from typing import Union, List, TYPE_CHECKING
```
dabacon: missing copyright

Issue #1728, opened 2019-06-18
Is python3-tk still a system requirement?


Strilanc:
We had this requirement because of matplotlib, but I just tried uninstalling python3-tk from my machine and then plotting stuff and it worked.

Note that we did switch to a more recent version of matplotlib: https://github.com/quantumlib/Cirq/pull/1684 . But in my testing the old version also worked..

vtomole:
tkinter is packaged with the Python 3+ installs. Seems like i wasn't the only one having problems reproducing this: https://stackoverflow.com/questions/25905540/importerror-no-module-named-tkinter.  I'll remove this as part of the "Remove Python 2 support" issue.

PR #1876, opened 2019-07-25
Restore python3-tk as an apt requirement

Strilanc:
- I needed to install this to get matplotlib working

### DIFF
--- a/apt-system-requirements.txt
+++ b/apt-system-requirements.txt
@@ -1,2 +1,3 @@
 texlive-latex-base
 latexmk
+python3-tk


dabacon:
Any idea why this changed from your last update? https://github.com/quantumlib/Cirq/issues/1728

Strilanc:
I'm not sure. Perhaps uninstalling python3-tk didn't *really* remove it, or perhaps I had installed some other dependency that was filling in the blank. In the future we should only remove this after testing that it works on a fresh virtual machine.

PR #1877, opened 2019-07-25
Document FSimGate.__init__'s parameters

Strilanc:


cirq/ops/fsim_gate.py
```diff
@@ -56,6 +56,13 @@ class FSimGate(gate_features.TwoQubitGate,
     """
 
     def __init__(self, theta: float, phi: float):
+        """
+        Args:
+            theta: Rotation angle, in radians, for the exchange interaction. Controls the inner 2x2 block of the
+                operation's unitary (the middle 2x2 block will equal `cirq.unitary(cirq.Rx(theta))`).
```
dabacon: What does this add that is not in the definition of the class?

The temrs "middle block" and "bottom 1x1 cell" aren't really clear terms in my opinion.

cirq/ops/fsim_gate.py
```diff
@@ -56,6 +56,13 @@ class FSimGate(gate_features.TwoQubitGate,
     """
 
     def __init__(self, theta: float, phi: float):
+        """
+        Args:
+            theta: Rotation angle, in radians, for the exchange interaction. Controls the inner 2x2 block of the
```
dabacon: not really true.  This is the term in from of  XX+YY.  The exchange interaction is XX+YY+ZZ

mpharrigan:
Something like: theta is the Rx angle in the |01> |10> subspace; phi is the CPhase angle

I think something like this would be helpful and add to the existing definition

cirq/ops/fsim_gate.py
Strilanc: It provides useful pop-up documentation when creating the class in pycharm. It places the definition directly adjacent to the use.

cirq/ops/fsim_gate.py
Strilanc: Fixed

cirq/ops/fsim_gate.py
```diff
@@ -56,6 +56,12 @@ class FSimGate(gate_features.TwoQubitGate,
     """
 
     def __init__(self, theta: float, phi: float):
+        """
+        Args:
+            theta: The strength of the XX+YY interaction, in radians.
```
dabacon: (XX+YY) / 2 ?

Strilanc: Maybe we should list which angles correspond to which operations, e.g. the angle at which 01 becomes 10 (phased) for the first time For now I'll merge this; feel free to improve it further.

PR #1880, opened 2019-07-26
JSON Protocol

mpharrigan:
Protocol to support serializing cirq objects

 - `cirq.to_json`
 - `cirq.read_json`

The former looks for a magic method `_json_dict_(self)` which returns a dictionary keyed by strings and whose values are either basic python types or are objects that support the SupportsJSON protocol. They dictionary must have an entry with key `"cirq_type"`, to be used during loading. The latter looks for a magic classmethod `_from_json_dict_(cls, **kwargs)` or falls back to calling the constructor.

`_json_dict_` is required for objects to support serialization, so `cirq.protocols.to_json_dict` is provided as convenience to implementers. You must only provide a list of attribute names which will be included in the dictionary.

During deserialization, the `"cirq_type"` value is passed to an ordered list of functions. Each function either *resolves* the string name into a python type or the next function is called. A default mapping from class name is provided as a dictionary. 

There is a test fixture that tests any object exposed at the top `cirq`-level import by roundtripping to JSON. There is a blacklist of classes for which it doesn't make sense to serialize, and another "xfail" list for objects that haven't been upgraded to support serialization yet, but there's no reason not for them to.

This pull request includes support for Qubits, Gates, Operations, Moments, Circuits, PauliString, and a couple of other things. I recommend reviewing and merging this PR which extends support for these high-priority objects and then working on the rest over time. Any new objects should implement the serialization protocol unless there's good reason not to. 

This pull request makes some small changes to objects, mostly adding `@property`s to provide a consistent expectation that arguments passed to the constructor are available under the same name as attributes/properties.


mpharrigan:
What is `args` in the above?

I'm not sure exactly what is different in your example vs. what I've implemented other than `GateOperation` being responsible for calling `_json_` on its children rather than having the encoder do it.

How does `from_json` know what type to construct?

cirq/protocols/json.py
```diff
@@ -0,0 +1,88 @@
...
+
+RaiseTypeErrorIfNotProvided = ([],)  # type: Any
+
+
+class SupportsJSON(Protocol):
+    """An object that can be turned into JSON dictionaries.
+
+    Returning `NotImplemented` or `None` means "don't know how to turn into
+    QASM". In that case fallbacks based on decomposition and known unitaries
+    will be used instead.
```
kevinsung: Docstring seems to be about QASM

mpharrigan: you caught me

cirq/protocols/json.py
```diff
@@ -0,0 +1,309 @@
...
+import json
+from typing import Union, Any, Dict, Optional, List, Callable, Type
+
+import numpy as np
+from typing_extensions import Protocol
+
+from cirq.type_workarounds import NotImplementedType
+
+
+# Note for developers:
```
Strilanc: We should probably have a test that checks that every public class is serializable.

cirq/protocols/json.py
```diff
@@ -0,0 +1,309 @@
...
+    `_from_json_dict_`, that will be called. Otherwise, the `cirq_type` key
+    will be popped from the dictionary and used as kwargs to the type's
+    constructor.
+    """
+
+    def _json_dict_(self) -> Union[None, NotImplementedType, Dict[Any, Any]]:
+        pass
+
+
+def to_json_dict(obj, attribute_names):
```
Strilanc: I'm a bit worried that we're straying into territory that should be solved at the python level instead of at a library level. I don't have a better suggestion, there's just an alarm going off in my head about it.

mpharrigan: `@dataclasses` can use the `asdict` helper. In general, I think hacking into the `__dict__` is a little too magic

cirq/protocols/json.py
mpharrigan: Sketched out an implementation of this in the latest commit. There are some classes that we probably don't want or need to make serializable and there are some that are lower priority that I don't want to block this PR from being merged

mpharrigan: one problem with the existing implementation is that it leaves out singletons whose class is private

mpharrigan: I reworked the logic to try anything that's not a function or module, with an explicit blacklist

mpharrigan:
The mypy errors are all false-positives; How do I ignore them?

Strilanc:
Often you use `cast` (from `typing`) to assert that the type is some particular thing.

The specific case you are seeing, where it talks about modules not having particular types, are annoying. I don't know a consistent way to solve them other than to binary search the PR for the line change that's causing it and then experiment with that one change to find out what makes mypy happy.

There is also `# type: ignore`, I think.

mpharrigan:
`Incompatible types in assignment (expression has type "List[Callable[[str], Type[Any]]]", variable has type "Optional[List[Callable[[str], Type[Any]]]]")`

How would I put a cast here? I would think that assigning a value of type `T` to a variable declared to be type `Optional[T]` would be completely fine.

`cirq/ops/pauli_string.py:530: error: Signature of "_from_json_dict_" incompatible with supertype "PauliString"`

I'm doing this on purpose to match the constructor (which does not have the same signature as its supertypes; but mypy doesn't seem to care about that).

cirq/protocols/json.py
mpharrigan: Synced offline with @Strilanc. His objection was to the concept of serialization and that it should be done at the python level. Python's built-in serialization is pickle which is too brittle for our case; python's `json` module in the standard library is what we exploit here and follow the general pattern for extending that library

cirq/protocols/__init__.py
```diff
@@ -54,6 +54,7 @@
 from cirq.protocols.inverse import (
     inverse,
 )
+from cirq.protocols.json import (to_json, read_json, to_json_dict)
```
dabacon: nit: follow others in file and one per line.

cirq/protocols/json.py
```diff
@@ -0,0 +1,187 @@
...
+    """Lazily import and build registry to avoid circular imports."""
+
+    def __init__(self):
+        self._crd = None
+
+    @property
+    def cirq_class_resolver_dictionary(self) -> Dict[str, Type]:
+        if self._crd is None:
+            import cirq
+            import cirq.ops.pauli_gates  # for mypy
```
dabacon: if this is for mypy can we put it behind an `if TYPE_CHECKING:`?

cirq/protocols/json.py
```diff
@@ -0,0 +1,187 @@
...
+    constructor.
+    """
+
+    def _json_dict_(self) -> Union[None, NotImplementedType, Dict[Any, Any]]:
+        pass
+
+
+def to_json_dict(obj, attribute_names):
+    d = {'cirq_type': obj.__class__.__name__}
+    for attr_name in attribute_names:
```
dabacon: I guess we want more control than just doing introspection on the attributes passed to the constructor? 

One thought is that you could use annotations to control this:
`def __init__(exponent: SerializableFloat):`

cirq/protocols/json.py
```diff
@@ -0,0 +1,187 @@
...
+    def __init__(self):
+        self._crd = None
+
+    @property
+    def cirq_class_resolver_dictionary(self) -> Dict[str, Type]:
+        if self._crd is None:
+            import cirq
+            import cirq.ops.pauli_gates  # for mypy
+            import cirq.devices.unconstrained_device  # for mypy
+            self._crd = {
```
dabacon: Probably a bad idea, but we could use `inspect`'s getmembers to construct this?

cirq/protocols/json.py
```diff
@@ -0,0 +1,95 @@
...
+TDefault = TypeVar('TDefault')
+
+RaiseTypeErrorIfNotProvided = ([],)  # type: Any
+
+
+class SupportsJSON(Protocol):
+    """An object that can be turned into JSON dictionaries.
+
+    Returning `NotImplemented` or `None` means "don't know how to turn into
+    QASM". In that case fallbacks based on decomposition and known unitaries
```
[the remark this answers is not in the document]
mpharrigan: this was from one of the earlier commits. Are you looking at the latest version of this PR?

cirq/protocols/__init__.py
mpharrigan: I did! the formatter put it like this

cirq/protocols/json.py
mpharrigan: done

cirq/protocols/json.py
mpharrigan: > doing introspection on the attributes passed to the constructor

This is a decent idea, but I'm worried it veers too far towards "magic". The current scheme makes it pretty easy to add JSON support for a class, but a developer must explicitly add something to indicate it supports JSON. If this part was automated there is a chance that a class will serialize badly. Right now, if you try to serialize something that a developer hasn't given a little bit of thought to, it will just refuse. I think that's a desirable thing

mpharrigan: The problem with the annotations in the constructor is it would introduce noise to users who might be entirely uninterested in serialization

cirq/protocols/json.py
mpharrigan: If you look through the history of commits, I used to just `import cirq; getattr(cirq, cirq_type_str)`. @Strilanc and I agreed that this was too open-ended

cirq/protocols/json.py
dabacon: That's true.  It's a very Java way to handle this problem :)

cirq/ops/common_gates.py
```diff
@@ -665,6 +679,12 @@ def _qasm_(self, args: protocols.QasmArgs,
     def _value_equality_values_(self):
         return self.num_qubits(),
 
+    def _json_dict_(self):
+        return {
+            'cirq_type': self.__class__.__name__,
+            'num_qubits': self._num_qubits
```
dabacon: should num_qubits be a property or an attribute?

cirq/protocols/json.py
```diff
@@ -0,0 +1,191 @@
...
+from typing import Union, Any, Dict, Optional, List, Callable, Type, cast, \
+    TYPE_CHECKING
+
+import numpy as np
+from typing_extensions import Protocol
+
+from cirq.type_workarounds import NotImplementedType
+
+if TYPE_CHECKING:
+    # mypy needs these explicitly imported
```
dabacon: Nit: I'd nix this comment because the if type checking makes this clear?

cirq/protocols/json.py
```diff
@@ -0,0 +1,191 @@
...
+
+def to_json_dict(obj, attribute_names):
+    d = {'cirq_type': obj.__class__.__name__}
+    for attr_name in attribute_names:
+        d[attr_name] = getattr(obj, attr_name)
+    return d
+
+
+class CirqEncoder(json.JSONEncoder):
+
```
dabacon: could use a small doc string

cirq/protocols/json.py
```diff
@@ -0,0 +1,191 @@
...
+
+RESOLVER_CACHE = _ResolverCache()
+
+
+def _cirq_class_resolver(cirq_type: str) -> Union[None, Type]:
+    return RESOLVER_CACHE.cirq_class_resolver_dictionary.get(cirq_type, None)
+
+
+DEFAULT_RESOLVERS = [
```
dabacon: What's the purpose of having this as a public constant?  So users can add to it their own resolvers?  If so this should be documented.

mpharrigan:
Very good point about parameterized circuits. Ideally, we could dispatch to a sympy-managed serialization format *but* the most promising candidate `sympy.srepr` requires calling `parse_expr` to load back in, which seems to just `eval` stuff which is a huge security risk.

We could support serialization of a number of common sympy objects using the same format as cirq objects I suppose. This might be a lot of work and/or only cover a subset of sympy objects

dabacon:
I think first would be to do the easy case of supporting symbols that are just Sybol strings.

cirq/ops/common_gates.py
mpharrigan: `num_qubits()` is a function. Most other gates do not take `num_qubits` as an argument to the constructor because most other gates are for a fixed number of qubits

mpharrigan: Changing `self._num_qubits` -> `self.num_qubits()` to be more obvious

cirq/protocols/json.py
mpharrigan: done

cirq/protocols/json.py
mpharrigan: done

cirq/protocols/json.py
mpharrigan: That's right. I've documented this use case with some sample code

mpharrigan:
@dabacon addressed concerns and added support for `sympy.Symbol`

cirq/ops/pauli_string.py
```diff
@@ -496,3 +518,13 @@ def __rmul__(self, other):
...
     def __neg__(self):
         return -self._as_pauli_string()
+
+    def _json_dict_(self):
+        return protocols.to_json_dict(self, ['pauli', 'qubit'])
+
+    @classmethod
+    def _from_json_dict_(  # type: ignore
+            cls, pauli: pauli_gates.Pauli, qubit: raw_types.Qid, **kwargs):
+        # Note, this method is required or else superclasses' deser
```
Strilanc: deser?

cirq/protocols/json.py
```diff
@@ -0,0 +1,234 @@
...
+    def __init__(self):
+        self._crd = None
+
+    @property
+    def cirq_class_resolver_dictionary(self) -> Dict[str, Type]:
+        if self._crd is None:
+            import cirq
+            self._crd = {
+                'CCXPowGate': cirq.CCXPowGate,
+                'CCZPowGate': cirq.CCZPowGate,
```
Strilanc: Is it necessary to repeat the names twice? You could have a list of the classes, and then `{a.__name__: a for a in ...}`

cirq/protocols/json.py
```diff
@@ -0,0 +1,234 @@
...
+    return RESOLVER_CACHE.cirq_class_resolver_dictionary.get(cirq_type, None)
+
+
+DEFAULT_RESOLVERS = [
+    _cirq_class_resolver,
+]
+"""A default list of 'resolver' functions for use in read_json.
+
+For more information about cirq_type resolution during deserialization
+please read the docstring for `read_json`.
```
Strilanc: `cirq.read_json` to be unambiguous

cirq/protocols/json.py
```diff
@@ -0,0 +1,234 @@
...
+
+        # TODO: More support for sympy
+        #       https://github.com/quantumlib/Cirq/issues/2014
+        if isinstance(o, sympy.Symbol):
+            return to_json_dict(o, ['name'], namespace='sympy')
+        return super().default(o)  # coverage: ignore
+
+
+def _cirq_object_hook(d, resolvers: List[Callable[[str], Union[None, Type]]]):
+    if 'cirq_type' in d:
```
Strilanc: Nit: invert into a guard clause `if not in d: return d`

cirq/protocols/json.py
```diff
@@ -0,0 +1,234 @@
...
+        if hasattr(cls, '_from_json_dict_'):
+            return cls._from_json_dict_(**d)
+
+        del d['cirq_type']
+        return cls(**d)
+
+    return d
+
+
+def to_json(obj: Any, file, *, indent=2, cls=CirqEncoder):
```
Strilanc: It *has* to go to a file? Maybe the method should be called `write_json` or `write_as_json`.

cirq/protocols/json.py
```diff
@@ -0,0 +1,234 @@
...
+    if isinstance(file, str):
+        with open(file, 'w') as actually_a_file:
+            return json.dump(obj, actually_a_file, indent=indent, cls=cls)
+
+    return json.dump(obj, file, indent=indent, cls=cls)
+
+
+def read_json(
```
Strilanc: This method name is not very clear. I would not expect `cirq.read_json` to return a circuit object just from reading the name. Maybe `read_from_json`?

cirq/protocols/json_test.py
```diff
@@ -0,0 +1,446 @@
...
+  "qubits": [
+    {
+      "cirq_type": "LineQubit",
+      "x": 5
+    }
+  ]
+}""")
+
+
+def test_op_roundtrip_filename(tmpdir):
```
Strilanc: ! `tmpdir` so convenient!

cirq/protocols/json_test.py
```diff
@@ -0,0 +1,446 @@
...
+    assert e.match("Could not resolve type 'MyCustomClass' "
+                   "during deserialization")
+
+
+QUBITS = cirq.LineQubit.range(5)
+Q0, Q1, Q2, Q3, Q4 = QUBITS
+
+TEST_OBJECTS = {
+    'CCNOT':
+    cirq.CCNOT,
```
Strilanc: yapf did this to you?

cirq/protocols/json_test.py
```diff
@@ -0,0 +1,446 @@
...
+    names = [
+        name for name, _ in inspect.getmembers(
+            cirq, lambda x: not (inspect.ismodule(x) or inspect.isfunction(x)))
+    ]
+    for name in NOT_YET_SERIALIZABLE:
+        assert name in names
+
+
+def test_mutually_exclusive_blacklist():
+    assert len(set(SHOULDNT_BE_SERIALIZED) & set(NOT_YET_SERIALIZABLE)) == 0
```
Strilanc: Make this error message significantly better. Anyone who adds a gate into cirq is going to see this exception. It should explain clearly what they need to do in order to fix it.

cirq/ops/pauli_string.py
mpharrigan: added the rest of the word

cirq/protocols/json.py
mpharrigan: done

cirq/protocols/json.py
mpharrigan: done

cirq/protocols/json.py
mpharrigan: to_json / read_json is modeled after pandas's function names

cirq/protocols/json.py
mpharrigan: it's like "read (a) book" :)

again, this is from pandas

cirq/protocols/json_test.py
mpharrigan: ya

cirq/protocols/json_test.py
mpharrigan: yes. are you asking about how it put the keys and values on separate lines but with the same indentation?

cirq/protocols/json_test.py
mpharrigan: the particular line you commented on developers will never see unless they are doing explicit json work. 

I've added a very cordial error message to `test_all_roundtrip`, which is where the error will be raised.

cirq/protocols/json.py
mpharrigan: - write implies you *have* to provide a file-like object, whereas this also works with a filename
 - json's method name "dump" is unpleasant (and also would imply needing a file-like object
 - pandas will typically be used in conjunction with this functionality. The first thing I do is add things I've saved to a dataframe

mpharrigan: I've realized that there's no docstring for this function. Adding now

cirq/protocols/json.py
Strilanc: In that eventuality, we can create the dictionary as `{easy_case} + {hard_stuff}` and have the best of both worlds. This is not a blocking change.

cirq/protocols/json.py
Strilanc: Okay

cirq/protocols/json.py
Strilanc: Okay

cirq/protocols/json_test.py
Strilanc: Nice fixed fix

PR #1886, opened 2019-07-26
Bump ply version

### LINKED ISSUE #1885: dev and contrib requirements inconsistent
I just rebuilt my virtualenv and noticed an error about inconsistent requirements:

```
ERROR: qiskit 0.6.1 has requirement ply>=3.10, but you'll have ply 3.4 which is incompatible.
```

It seems that our contrib requirements explicitly call for ply 3.4 while dev requirements call for qiskit 0.6.* and thus implicitly for ply >=3.10.

I checked whether the tests in qasm_import (the contrib package that uses ply) pass with 3.10 and they do. Time to bump ply to 3.10?

viathor:
Fixes #1885.

### DIFF
--- a/cirq/contrib/contrib-requirements.txt
+++ b/cirq/contrib/contrib-requirements.txt
@@ -1,6 +1,6 @@
 # Runtime requirements for contrib.
 
-ply==3.10
+ply>=3.4
 pylatex==1.3.*
 # Python 3.7 support for Tensorflow is currently only supported in the nightly build.
 tf-nightly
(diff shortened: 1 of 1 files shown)


cirq/contrib/contrib-requirements.txt
```diff
@@ -1,6 +1,6 @@
 # Runtime requirements for contrib.
 
-ply==3.4
+ply==3.10
```
balopat:
```suggestion
ply>=3.4
```

viathor:
PTAL, @balopat

PR #1889, opened 2019-07-26
Flip swap angle sign in iswap

### LINKED ISSUE #1882: Handedness between ISwapPowGate and FSim is different.
The ISwapPowGate and FSim gate rotate in different directions on the |01>, |10> subspace.  This feels a bit odd to me.

viathor:
Fixes #1882.


cirq/ops/common_gates.py
```diff
@@ -1081,14 +1081,12 @@ class ISwapPowGate(eigen_gate.EigenGate,
...
             (0, np.diag([1, 0, 0, 1])),
-            (+0.5, np.array([[0, 0, 0, 0],
-                             [0, 0.5, 0.5, 0],
-                             [0, 0.5, 0.5, 0],
-                             [0, 0, 0, 0]])),
-            (-0.5, np.array([[0, 0, 0, 0],
-                             [0, 0.5, -0.5, 0],
-                             [0, -0.5, 0.5, 0],
-                             [0, 0, 0, 0]])),
+            (-0.5,
```
vtomole: Make the arrays more readable by calling `#yapf disable` around these parts.

viathor: Done.

cirq/ops/common_gates_test.py
```diff
@@ -620,13 +620,12 @@ def test_iswap_str():
 
 
 def test_iswap_unitary():
```
vtomole: I'm seeing different answers on https://quantumcomputing.stackexchange.com/questions/2594/what-is-the-matrix-of-the-iswap-gate. Some people say it should be `+i` while some people argue that it doesn't matter if it's `+i` or `-i`. Could someone please clear this up for me?

viathor: There are two conflicting conventions. I think we have decided to use `-i` and already switched fSim in #1726. However, ISwap was missed in that PR and so @dabacon filed #1882. This PR fixes the inconsistency and completes our switch to the `-i` convention for both gates.

cirq/ops/common_gates.py
```diff
@@ -1060,46 +1060,48 @@ class ISwapPowGate(eigen_gate.EigenGate,
...
          [0, 0, 0, 1]]
 
     where:
 
         c = cos(π·t/2)
-        s = sin(π·t/2)
+        s = -i·sin(π·t/2)
 
     `cirq.ISWAP`, the swap gate that applies i to the |01> and |10> states,
     is an instance of this gate at exponent=1.
```
cduck: Is the sign of i wrong here now?  This was just changed from -1 a few days ago in #1848.

cirq/ops/common_gates_test.py
```diff
@@ -620,13 +620,12 @@ def test_iswap_str():
...
-    cirq.testing.assert_allclose_up_to_global_phase(
-        cirq.unitary(cirq.ISWAP),
-        np.array([[1, 0, 0, 0],
-                  [0, 0, 1j, 0],
-                  [0, 1j, 0, 0],
-                  [0, 0, 0, 1]]),
-        atol=1e-8)
+    cirq.testing.assert_allclose_up_to_global_phase(cirq.unitary(cirq.ISWAP),
+                                                    np.array([[1, 0, 0, 0],
+                                                              [0, 0, -1j, 0],
```
Strilanc: I think this is wrong. The definition of an iswap is to put i's along the anti-diagonal. I carefully looked into this and asked several people about it before putting this matrix into the test.

See this quantum stackexchange question: https://quantumcomputing.stackexchange.com/a/2596/119

cirq/ops/common_gates.py
dabacon: I think also you can add a trailing ","?  
`[[0, 0, 0, 0], 
 [0, 0.5, 0.5, 0],
 [0, 0.5, 0.5, 0],
 [0, 0.5, 0.5, 0],
 [0, 0, 0, 0],]`

cirq/ops/common_gates.py
dabacon: I think we want this to be consistent with FSim.  #1882

mpharrigan:
The only consistent thing is inconsistency

viathor:
We've decided to keep opposite sign conventions in iSWAP and fSim.

PR #1891, opened 2019-07-26
Add reference for iswap gate convention

Strilanc:
Part of #1882

I'm adding these references because I temporarily mistakenly approved a PR flipping the sign, and I want to beef up the defenses against that kind of mistake happening again.


cirq/ops/common_gates_test.py
```diff
@@ -622,6 +622,8 @@ def test_iswap_str():
 def test_iswap_unitary():
     cirq.testing.assert_allclose_up_to_global_phase(
         cirq.unitary(cirq.ISWAP),
+        # Reference for the iswap gate's matrix using +i instead of -i:
+        # https://quantumcomputing.stackexchange.com/questions/2594/
```
vtomole: Link to answer: https://quantumcomputing.stackexchange.com/a/2596/362

Strilanc: I purposefully linked to the question instead of the answer. Makes the link shorter and is more future-proof.

Strilanc:
Changing these kinds of conventions is a silent breaking change on our users. The iswap gate has been around for several versions now; we can't just change what it means.

Personally I don't mind so much if iswap is at fsim(-90deg) instead of fsim(+90deg).

Strilanc:
If we want to change this convention, we first need to deprecate iswap for one version then delete it for one version then re-introduce it with the new convention.

maffoo:
I tend to agree with @Strilanc. `ISWAP` as a gate is defined with +i and IMO it's worth sticking to that. That this corresponds to theta = -pi/2 in our definition of fsim is a straightforward consequence of the definitions and I would say it's not worth changing the definitions just to change that.

cirq/ops/common_gates.py
```diff
@@ -1076,6 +1076,10 @@ class ISwapPowGate(eigen_gate.EigenGate,
 
     `cirq.ISWAP`, the swap gate that applies i to the |01> and |10> states,
     is an instance of this gate at exponent=1.
+
+    References:
+        "What is the matrix of the iSwap gate?"
+        https://quantumcomputing.stackexchange.com/questions/2594/
```
dabacon: I guess it sort of makes sense to reference this.  But I'm not sure it is really helpful, I mean does going to that question really help the user?

vtomole: We'll be sending a user who asks why we are doing this to that link, so we might as well have it here for users who have that question.

dabacon: We don't do this for other gates (similarly they could be defined differently, and indeed the handedness of single qubit rotations is done very much using both conventions) so it seems strange to call it out.

vtomole: Good point. I assume all other gates are not "defined differently", meaning deviating from standards texts.

vtomole: > handedness of single qubit rotations is done very much using both conventions

Didn't understand what this meant before i posted my comment. I agree.

dabacon:
This needs a format fix.

cirq/ops/common_gates.py
Strilanc: I think the fact that this question came up for this gate, and not for other gates, justifies the citation. I would be happy to add similar references to the other gates, e.g. to particular pages of mike and ike.

PR #1894, opened 2019-07-28
Device specification proto

dstrain115:
- Add device specification proto for specifying devices
in cirq, including qubits, connections, timings, and gate sets.

- Also compiled protos but not sure if I did it correctly.


cirq/api/google/v2/device.proto
```diff
@@ -0,0 +1,81 @@
...
+    int64 gate_duration_picos = 4;
+}
+
+message GateArgument {
+    enum ArgumentType {
+       UNSPECIFIED = 0;
+       FLOAT = 1;
+       BOOLEAN = 2;
+       STRING = 3;
+    }
```
maffoo: In [program.proto](https://github.com/quantumlib/Cirq/blob/master/cirq/api/google/v2/program.proto#L147) we allow `RepeatedBoolean` as a possible argument type. It would seems to me we should try to keep the possible `ArgValue` variants specified there in sync with the types here.

cirq/api/google/v2/device.proto
```diff
@@ -0,0 +1,81 @@
...
+   string gate_set_name = 1;
+
+   repeated GateDefinition allowed_gates = 2;
+}
+
+message GateDefinition {
+
+    // A list of allowed ids for gates within the Program.
+    // Any programs with ids not in this list will be rejected.
+    string gate_name = 1;
```
maffoo: might want to call this `gate_id` to match https://github.com/quantumlib/Cirq/blob/master/cirq/api/google/v2/program.proto#L110.

cirq/api/google/v2/device.proto
```diff
@@ -0,0 +1,81 @@
...
+
+    // The number of qubits (targets) that can be specified for this gate.
+    // If unset or set to zero, any number of qubits is allowed.
+    int32 number_of_qubits = 2;
```
maffoo: If you have a 3-qubit gate, how would you specify which sets of qubits it can be applied to? For two qubit gates this is encoded in the neighbor information, but not clear how to do this for number_of_qubits > 2.

cirq/api/google/v2/device.proto
```diff
@@ -0,0 +1,81 @@
...
+}
+
+message ArgumentRange {
+   float minimum_value = 1;
+   float maximum_value = 2;
+}
+
+message QubitInformation {
+   // A list of qubit ids to which a 2-qubit operation is legal
+   repeated string neighbor_ids = 1;
```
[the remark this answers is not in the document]
dstrain115: Changed to Targets.  PTAL and let me know what you think.

cirq/api/google/v2/device.proto
dstrain115: Done.

cirq/api/google/v2/device.proto
dstrain115: Changed to id, since the gate part should be implied by the proto message.

cirq/api/google/v2/device.proto
dstrain115: Changed to Target

cirq/api/google/v2/device.proto
```diff
@@ -0,0 +1,96 @@
+syntax = "proto3";
+
+package cirq.api.google.v2;
+
+option java_package = "com.google.cirq.api.google.v2";
+option java_outer_classname = "CalibrationProto";
```
dabacon: DeviceProto

cirq/api/google/v2/device.proto
```diff
@@ -0,0 +1,96 @@
...
+    // This should correspond to the legal assignment
+    // of the Arg.arg oneof for this argument
+    ArgType type = 2;
+
+    // This should only be populated for type FLOAT.
+    // If not set, all float values are allowed.
+    repeated ArgumentRange allowed_ranges =3;
+}
+
+// Minimum and maximum should be inclusive.
```
dabacon: I think it is better to make one inclusive and the other exclusive, since these are often periodic in, say 2 pi, and making both inclusive means there is a double count (this can happen for other gates for other reasons, but that case is pretty common).

cirq/api/google/v2/device.proto
```diff
@@ -0,0 +1,96 @@
...
+    // map value of Operation.args
+    string name = 1;
+
+    // The type of the argument.
+    // This should correspond to the legal assignment
+    // of the Arg.arg oneof for this argument
+    ArgType type = 2;
+
+    // This should only be populated for type FLOAT.
+    // If not set, all float values are allowed.
```
dabacon: There are also cases where we have discrete values, i.e. you can only supply 0, 0.5, or 1 for this value.

cirq/api/google/v2/device.proto
```diff
@@ -0,0 +1,96 @@
...
+    repeated ArgumentRange allowed_ranges =3;
+}
```
dabacon: a worry (but not as big) is that ranges may not be general enough to capture actual region of allowed values, especially for multiple parameters.  Not sure there is a great we to capture this without a lot more complexity, though so probably can ignore for now.

cirq/api/google/v2/device.proto
```diff
@@ -0,0 +1,96 @@
...
+  repeated GateSet valid_gate_sets = 1;
+
+  // A list of allowed ids for qubits within the Program.
+  // Any programs with ids not in this list will be rejected.
+  // If empty, all qubit values are allowed (e.g. in a simulator)
+  repeated string valid_qubits = 2;
+
+  // A list of allowed targets for multi-qubit gates.
+  // For a typical grid layout, this will be a list of adjacent pairs
+  repeated Target valid_targets = 3;
```
dabacon: for gates that are not symmetric, does the order matter?

dabacon: I know we would like to keep this small but this might not be flexible enough.  For example we could have two qubit gates, plus measurements that are across one resonator and so couple multiple qubits.  Right now you can't specify that because Target isn't on the gate definition.

cirq/api/google/v2/device.proto
dstrain115: Done.

cirq/api/google/v2/device.proto
dstrain115: SGTM, changed documentation.

cirq/api/google/v2/device.proto
dstrain115: The plan for this is to have 3 allowed ranges, with max and min be the same value (0, 0.5, 1).

cirq/api/google/v2/device.proto
dstrain115: Acknowledged.  

I think we could move this into a oneof with more complicated semantics if necessary:
https://developers.google.com/protocol-buffers/docs/proto3#backwards-compatibility-issues

cirq/api/google/v2/device.proto
dstrain115: Hmmm.  What do you think about this idea?

We make the top-level target list the "default" list, but add an option to override the target list at the gate level.  If not specified at the gate-level, it defaults to the top-level list.  That way, for the typical case, it would be the same size, but would allow gate-specific target lists for more complicated gates.

What do you think?

(I have not put this into the proto yet)

cirq/api/google/v2/device.proto
```diff
@@ -46,13 +45,18 @@ message GateDefinition {
...
   // an operation of this gate
   repeated ArgDefinition valid_args = 3;
 
   // This defines the approximate amount of time for each gate,
   // specified as an integer number of picoseconds.
   int64 gate_duration_picos = 4;
+
+  // Valid targets that this gate can use.
+  // Values in this list correspond to the name of the TargetList
+  // If unset, all combinations using the number of qubits are allowed.
```
dabacon: maybe directly reference the field location?

cirq/api/google/v2/device.proto
```diff
@@ -87,6 +91,19 @@ message ArgumentRange {
...
+message TargetList {
+  // The name of the target list.
+  // This will be referenced in the GateDefintion to denote
+  // which targets are valid.
+  string name = 1;
+
+  // A list of targets that are valid
+  repeated Target targets = 2;
+}
+
```
dabacon: nit: extra \n

cirq/api/google/v2/device.proto
dstrain115: I have changed this to use a TargetList and have made is_symmetric an attribute on the TargetList.

dstrain115: I further refined this idea after we talked and then decided to make a proto called a TargetList with a list of targets.

The gates then refer to this list by name rather than re-list all targets.  This means that, for the typical case, you just list all the qubit pairs, then give it a name, like "adjacent_pairs" then all the gates say that their valid target is "adjacent_pairs".

What do you think?

cirq/api/google/v2/device.proto
dstrain115: Done.

cirq/api/google/v2/device.proto
dstrain115: Changed to reference number_of_qubits.

cirq/api/google/v2/device.proto
maffoo: Would be good to add a comment indicating this should be kept in sync with the `ArgValue` proto.

cirq/api/google/v2/device.proto
```diff
@@ -0,0 +1,116 @@
...
+// Minimum value is inclusive and maximum value is exclusive.
+// If minimum and maximum values are the same, only a single value is allowed.
+message ArgumentRange {
+  float minimum_value = 1;
+  float maximum_value = 2;
+}
+
+// A list of targets that are valid for a set of gates.
+// For instance, all qubit pairs that can be acted on by a 2-qubit gate
+message TargetList {
```
maffoo: nit: consider calling this `TargetSet` by analogy with `GateSet`.

cirq/api/google/v2/device.proto
```diff
@@ -0,0 +1,116 @@
...
+  // The name of the target list.
+  // This will be referenced in the GateDefinition to denote
+  // which targets are valid.
+  string name = 1;
+
+  // If set to true, any order of the targets is valid.
+  // If false, then the order of the targets matters and must match the order in
+  // the targets list.
```
maffoo: Might be good to clarify that this refers to the order of ids within each target listed in `targets`. The order of the entries in the `targets` field itself does not matter (I presume). Also, might it be useful to change this to be an enum instead, something like `TargetOrdering`? Could start with options that require the exact specified ordering or any permutation, but I could imagine other possibilities for multi-qubit gates in, say, ion traps.

cirq/api/google/v2/device.proto
```diff
@@ -0,0 +1,116 @@
...
+  repeated ArgDefinition valid_args = 3;
+
+  // This defines the approximate amount of time for each gate,
+  // specified as an integer number of picoseconds.
+  int64 gate_duration_picos = 4;
+
+  // Valid targets that this gate can use.
+  // Values in this list correspond to the name of the TargetList
+  // If unset, all combinations with number_of_qubits target are allowed.
+  repeated string valid_targets = 5;
```
maffoo: How would one specify targets for a gate that can take variable numbers of qubits, such as measurement gates, if you want to restrict to a subset of qubits?

cirq/api/google/v2/device.proto
dstrain115: Done.

cirq/api/google/v2/device.proto
dstrain115: Done.

cirq/api/google/v2/device.proto
dstrain115: Made this a SUBSET_PERMUTATION TargetOrdering

cirq/api/google/v2/device.proto
```diff
@@ -93,16 +94,31 @@ message ArgumentRange {
...
+    UNSPECIFIED = 0;
+
+    // Symmetric gates, any id order within each target is valid.
+    SYMMETRIC = 1;
+
+    // Asymmetric gates, the order of ids in each target is important.
+    // Only the order specified in each target is valid.
+    ASYMMETRIC = 2;
+
+    // All targets should be single qubits.  Any subset of these targets
```
maffoo: Might want to clarify here the difference between "targets" in this `TargetSet`, vs. "targets" of the gate, something like: "All targets in this TargetSet should be single qubits. The gate itself can be applied to any subset of these targets, in any order. This may be the case for measurement gates, for example, if we can measure any subset of qubits at once."

Issue #1850, opened 2019-07-19
Phase out SupportsPhase protocol


kevinsung:
The use case I have in mind is defining this protocol for SWAP. This is useful, for instance, for the EjectZ optimizer because one can pass a Z gate through a SWAP as follows:
```
0: ───Z^theta───×───
                │
1: ─────────────×───
```
=
```
0: ───×─────────────
      │
1: ───×───Z^theta───
```
So I think the `phase_by` function signature should be changed from
```
def phase_by(val: Any,
             phase_turns: float,
             qubit_index: int,
             default: TDefault = RaiseTypeErrorIfNotProvided):
```
to
```
def phase_by(val: Any,
             phase_turns: float,
             qubit_index: int=None,
             qubits: Sequence[cirq.Qid]=None,
             phased_qubit: cirq.Qid=None,
             default: TDefault = RaiseTypeErrorIfNotProvided):
```
Then `phase_by` for SWAP would return an `OP_TREE`.

mpharrigan:
We could also special case swaps in EjectZ for a quick and dirty solution

@Strilanc what do you think about @kevinsung 's proposed changes? also cc @dabacon

mpharrigan:
Initially: add in swap gate and fsim gate (with the one theta that works) support as hardcoded cases in EjectZ.

Later: delete phase_by protocol after hardcoding remaining cases.

dabacon:
I'm trying to reconcile the title of this issue with the description which appears to mean to augment the phase_by protocol.

What would be even better is if we could generalize this protocol for describing the logic behind these types of transforms.

kevinsung:
@dabacon I think what happened was that we discussed this issue during a Cirq sync and then decided to instead remove the SupportsPhase protocol using the plan in @mpharrigan's comment. Then I changed the title of this issue.

dstrain115:
Do we still have this protocol?  If we are still going to remove this, we should do it before 1.0.

dstrain115:
Adding back to triage/discuss since @tanujkhattar did not see a reason to deprecate it.

dabacon:
We can probably keep it.  It's an example of a very specific protocol tightly coupled to a particular transformer, which is not great.

dstrain115:
It was decided to keep this protocol after all.  The consensus in cirq cync was that this didn't really affect anyone negatively and did provide some small value, so that we would keep it in.

The above functionality to propagate Z's through swaps had already been added to EjectZ.

Closing this issue.

PR #1895, opened 2019-07-30
Eject Zs through SWAP

mpharrigan:
step one of #1850

### DIFF
--- a/cirq/optimizers/eject_z.py
+++ b/cirq/optimizers/eject_z.py
@@ -28,11 +28,15 @@ if TYPE_CHECKING:
 
 
 def _is_swaplike(op: ops.Operation):
-    if ops.op_gate_of_type(op, ops.SwapPowGate):
-        return op.gate.exponent == 1
+    gate = ops.op_gate_of_type(op, ops.SwapPowGate)
+    if gate:
+        return gate.exponent == 1
 
-    if ops.op_gate_of_type(op, ops.FSimGate):
-        return np.isclose(op.gate.theta, np.pi / 2)
+    gate = ops.op_gate_of_type(op, ops.FSimGate)
+    if gate:
+        return np.isclose(gate.theta, np.pi / 2)
+
+    return False
 
 
 class EjectZ():
--- a/cirq/optimizers/eject_z.py
+++ b/cirq/optimizers/eject_z.py
@@ -34,7 +34,7 @@ def _is_integer(n):
 def _is_swaplike(op: ops.Operation):
     gate1 = ops.op_gate_of_type(op, ops.SwapPowGate)
     if gate1:
-        return _is_integer((gate1.exponent - 1) / 2)
+        return gate1.exponent == 1
 
     gate2 = ops.op_gate_of_type(op, ops.ISwapPowGate)
     if gate2:
--- a/cirq/optimizers/eject_z.py
+++ b/cirq/optimizers/eject_z.py
@@ -91,7 +91,8 @@ class EjectZ():
 
                 if _is_swaplike(op):
                     a, b = op.qubits
-                    qubit_phase[a], qubit_phase[b] = qubit_phase[b], qubit_phase[a]
+                    qubit_phase[a], qubit_phase[b] = qubit_phase[
+                        b], qubit_phase[a]
                     continue
 
 
--- a/cirq/optimizers/eject_z.py
+++ b/cirq/optimizers/eject_z.py
@@ -40,7 +40,6 @@ def _is_swaplike(op: ops.Operation):
     if gate3:
         return np.isclose(np.abs(gate2.theta), np.pi / 2)
 
-
     return False
 
 
--- a/cirq/optimizers/eject_z.py
+++ b/cirq/optimizers/eject_z.py
@@ -32,9 +32,14 @@ def _is_swaplike(op: ops.Operation):
     if gate1:
         return gate1.exponent == 1
 
-    gate2 = ops.op_gate_of_type(op, ops.FSimGate)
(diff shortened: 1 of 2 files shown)


cirq/optimizers/eject_z.py
```diff
@@ -26,6 +27,18 @@
...
 
 
+def _is_swaplike(op: ops.Operation):
+    gate1 = ops.op_gate_of_type(op, ops.SwapPowGate)
+    if gate1:
+        return gate1.exponent == 1
+
+    gate2 = ops.op_gate_of_type(op, ops.FSimGate)
+    if gate2:
+        return np.isclose(gate2.theta, np.pi / 2)
```
dabacon: should this also include other angles that give swap like behavior?

kevinsung: Yeah, `-np.pi / 2` also gives swap-like behavior.

cirq/optimizers/eject_z.py
```diff
@@ -26,6 +27,18 @@
...
+
```
kevinsung: Also, `ISWAP` and its inverse.

cirq/optimizers/eject_z_test.py
```diff
@@ -261,3 +262,37 @@ def with_qubits(self, *new_qubits):
...
+def test_swap():
+    a, b = cirq.LineQubit.range(2)
+    original = cirq.Circuit.from_ops([cirq.Rz(.123).on(a), cirq.SWAP(a, b)])
+    optimized = original.copy()
+
+    cirq.EjectZ().optimize_circuit(optimized)
+    cirq.DropEmptyMoments().optimize_circuit(optimized)
+
+    assert optimized[0].operations == (cirq.SWAP(a, b),)
+    assert optimized[1].operations == (cirq.Z(b)**(.123 / np.pi),)
```
kevinsung: Why'd you switch from `Rz` to the exponentiated notation?

cirq/optimizers/eject_z_test.py
```diff
@@ -261,3 +262,37 @@ def with_qubits(self, *new_qubits):
...
+        cirq.FSimGate(theta=np.pi / 2, phi=.123).on(a, b)
+    ])
+    optimized = original.copy()
+
+    cirq.EjectZ().optimize_circuit(optimized)
+    cirq.DropEmptyMoments().optimize_circuit(optimized)
+
+    assert optimized[0].operations == (cirq.FSimGate(theta=np.pi / 2,
+                                                     phi=.123).on(a, b),)
+    assert optimized[1].operations == (cirq.Z(b)**(.123 / np.pi),)
```
kevinsung: Again, a bit confusing that you switched notation.

cirq/optimizers/eject_z.py
```diff
@@ -17,6 +17,7 @@
 from typing import Optional, cast, TYPE_CHECKING, Iterable
 from collections import defaultdict
 import sympy
+import numpy as np
```
dabacon: import numpy as np should be before import sympy

cirq/optimizers/eject_z.py
mpharrigan: done

cirq/optimizers/eject_z.py
mpharrigan: done

cirq/optimizers/eject_z_test.py
mpharrigan: No good reason. It should work though

cirq/optimizers/eject_z.py
mpharrigan: did this change, but my local pylint keeps giving me a `RecursionError` so I'm pushing and crossing my fingers that that was the only problem

mpharrigan:
PTAL

cirq/optimizers/eject_z.py
dabacon: should we also capture   + np.pi * an integer ?  I don't think any of our gates restrict their parameter ranges (probably they should?)

cirq/optimizers/eject_z.py
```diff
@@ -26,6 +27,22 @@
...
 
 
+def _is_swaplike(op: ops.Operation):
+    gate1 = ops.op_gate_of_type(op, ops.SwapPowGate)
+    if gate1:
+        return gate1.exponent == 1
+
+    gate2 = ops.op_gate_of_type(op, ops.ISwapPowGate)
+    if gate2:
+        return abs(gate2.exponent) == 1
```
kevinsung: Also -1

cirq/optimizers/eject_z_test.py
kevinsung: I think for readability you should keep it consistent.

cirq/optimizers/eject_z.py
mpharrigan: udpate iswap and fsim to do any multiple.

I couldn't do the same for SWAP because the unit test was reporting `cirq.SWAP(a, b)**n` was not equal to `(cirq.SWAP ** n).on(a,b)` which confused me. I just took it out for now unless you have an idea

kevinsung: That sounds like a bug; please open an issue for it. Add a TODO here to do SWAP.

dabacon:
Kevin wasn't the consistency, then we should get this in.

dabacon:
wants not wasn't

mpharrigan:
Update! There's actually a reason for the inconsistency. The `EjectZ` optimizer dumps the phase into exponentiated `Z`s instead of `Rz`s, ie the global phase is different

PR #1897, opened 2019-07-30
Document relationship between fSim, iSWAP and CPhase

### LINKED ISSUE #1882: Handedness between ISwapPowGate and FSim is different.

viathor:
Fixes #1882.

We've decided to accept the inconsistency, so the fix merely documents what is implemented. Also adds a unit test to confirm and improves existing documentation.

### DIFF
--- a/cirq/ops/fsim_gate.py
+++ b/cirq/ops/fsim_gate.py
@@ -63,12 +63,13 @@ class FSimGate(gate_features.TwoQubitGate,
     def __init__(self, theta: float, phi: float):
         """
         Args:
-            theta: Swap angle on the odd 1s subspace, in radians. Determined by
-                the strength and duration of the XX+YY interaction. Note: uses
-                opposite sign convention to the iSWAP gate.
+            theta: Swap angle on the span(|01⟩, |10⟩) subspace, in radians.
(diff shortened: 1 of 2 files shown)


cirq/ops/fsim_gate.py
```diff
@@ -53,14 +53,22 @@ class FSimGate(gate_features.TwoQubitGate,
...
+        FSimGate(θ, φ) = ISWAP**(-2θ/π) CZPowGate(exponent=-φ/π)
     """
 
     def __init__(self, theta: float, phi: float):
         """
         Args:
-            theta: The strength of the XX+YY interaction, in radians.
-            phi: CPhase angle in radians. Determines how much the |11⟩ state is
-                phased.
+            theta: Swap angle on the odd 1s subspace, in radians. Determined by
```
dabacon: nit: "odd 1s subspace" is a bit odd.  Suggest just saying "Swap angle on the span({|01>, |10>}) subspace".

viathor: It was odd indeed. I've evened it out.

PR #1902, opened 2019-07-31
 Filter out Identities from PauliString constructor

### LINKED ISSUE #1675: PauliString constructor should accept (but filter out) cirq.I (identity) values

mpharrigan:
fixes #1675

### DIFF
--- a/cirq/ops/pauli_string.py
+++ b/cirq/ops/pauli_string.py
@@ -44,8 +44,11 @@ class PauliString(raw_types.Operation):
                  coefficient: Union[int, float, complex] = 1) -> None:
         if qubit_pauli_map is None:
             qubit_pauli_map = {}
-        qubit_pauli_map = {q: p for q, p in qubit_pauli_map.items()
-                           if not isinstance(p, common_gates.IdentityGate)}
+        qubit_pauli_map = {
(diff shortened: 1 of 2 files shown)


cirq/ops/pauli_string.py
```diff
@@ -44,7 +44,12 @@ def __init__(self,
                  coefficient: Union[int, float, complex] = 1) -> None:
         if qubit_pauli_map is None:
             qubit_pauli_map = {}
-        for p in qubit_pauli_map.values():
+        qubit_pauli_map = {
+            q: p
+            for q, p in qubit_pauli_map.items()
+            if not isinstance(p, common_gates.IdentityGate)
+        }
+        for q, p in qubit_pauli_map.items():
```
viathor: If you restore `for p in qubit_pauli_map.values();` then linter will be happy :-)

mpharrigan: whoops! I changed this in an attempt to do the filtering within the loop but python (and everyone) is very unhappy if you remove things from a dictionary while iterating over it

viathor: Yeah, too many moving parts confuse the python interpreter.

PR #1903, opened 2019-08-01
Added Simon's Algorithm Example

Lucaman99:
I was able to finish implementing Simon's Algorithm with Cirq. I've tested it a decent amount, although there is still definitely a chance that there are bugs (specifically in the XOR Gaussian-Elimination part of the algorithm) that I have overlooked. Let me know if there are any changes that you want me to make, and I'll get them done ASAP.


Lucaman99:
I just discovered a bug, closing pull request until I fix it.

PR #1905, opened 2019-08-01
Reuse program

wcourtney:


cirq/google/engine/engine.py
```diff
@@ -16,8 +16,8 @@
 
 As an example, to run a circuit against the xmon simulator on the cloud,
     engine = cirq.google.Engine(project_id='my-project-id')
-    options = cirq.google.JobConfig(program_id='hello-world')
-    result = engine.run(options, circuit, repetitions=10)
+    program = engine.create_program(ciruit, 'my-experiment')
+    result = engine.run(program=program, repetitions=10)
```
dabacon: ah I was thinking we would use a different pattern 

```
engine = cirq.google.Engine(project_id='my-project-id')
program = engine.create_program(circuit, 'my-experiment')
program.run_sweeps(job_config=job_config, params=params0, repetitions=10, processor_ids=['gmon'])
program.run_sweeps(job_config=job_config, params=params1, repetitions=10, processor_ids=['gmon'])
...
```

cirq/google/engine/engine.py
```diff
@@ -213,7 +207,7 @@ def __init__(self,
     def run(
             self,
             *,  # Force keyword args.
-            program: Program,
+            program: Dict,
```
dabacon: I think we want to leave the ability to run directly from engine object which does create program, create job, return result.  But we want to add a run onto the program object so that it can be called repeatedly.  That way we don't break backwards compatibility.

wcourtney: Fixed, except that the program_id is removed from JobConfig, as discussed offline.

cirq/google/engine/engine.py
wcourtney: Updated to reflect this while keeping old Engine run() and run_sweep() methods. PTAL.

cirq/google/engine/engine.py
```diff
@@ -707,6 +726,103 @@ def get_metrics_by_name(self, name: str) -> List[Dict]:
...
+    Attributes:
+      name: The full resource name of the engine program.
+      code: A serialized version of the Circuit or Schedule
+    """
+
+    def __init__(self, program: Dict[str, Any], engine: Engine) -> None:
+        """A job submitted to the engine.
+
+        Args:
+            program: A full Program Dict from the Engine.
```
dabacon: I think you only use the program resource id?  I'd recommend just using that unless there is another use of the program dict

cirq/google/engine/engine.py
```diff
@@ -84,54 +99,37 @@ class JobConfig:
     scheduled or terminated programs executions. Program and job resources have
     string names. This object contains the information necessary to create a
     program and then create a job on Quantum Engine, hence running the program.
```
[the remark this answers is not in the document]
wcourtney: Updated and shortened this significantly. Moving away from GCS should eventually reduce this to only job_id, which could be later flattened.

cirq/google/engine/engine.py
wcourtney: My thought was that this could scale to do more without changing the interface, but fine by me to keep it simple for now. \Updated.

cirq/google/engine/engine.py
```diff
@@ -75,63 +82,51 @@ def request_builder(*args, **kwargs):
     return request_builder
 
 
+def _make_random_id(prefix: str, length: int = 6):
+    random_digits = [
+        random.choice(string.ascii_uppercase + string.digits) for _ in range(6)
```
dabacon: 6 should be length

cirq/google/engine/engine.py
```diff
@@ -336,57 +280,49 @@ def run_sweep(
         does not block until a result is returned.
 
         Args:
-            program: The Circuit or Schedule to execute. If a circuit is
-                provided, a moment by moment schedule will be used.
-            job_config: Configures the names of programs and jobs.
+            program: A Quantum Engine-wrapped Circuit or Schedule object. This
+              may be generated with create_program() or get_program().
+            job_config: Configures optional job parameters.
```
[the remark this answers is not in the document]
dabacon: I don't see it?

cirq/google/engine/engine.py
```diff
@@ -324,6 +267,7 @@ def run_sweep(
             self,
             *,  # Force keyword args.
             program: Program,
```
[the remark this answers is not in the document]
dabacon: Oh got it.

dabacon:
Also you have a mypy failure
```
$ check/mypy
cirq/google/engine/engine.py:434: error: Argument 2 to "_serialize_program" of "Engine" has incompatible type "Optional[SerializableGateSet]"; expected "SerializableGateSet"
The command "check/mypy" exited with 1.
```

cirq/google/engine/engine.py
wcourtney: Thanks

cirq/google/engine/engine.py
```diff
@@ -405,65 +340,145 @@ def run_sweep(
...
+            program_dict = {}  # type: Dict[str, Any]
+            program_dict['@type'] = TYPE_PREFIX + program_descriptor.full_name
+            program_dict['operations'] = [
+                op for op in api_v1.schedule_to_proto_dicts(schedule)
+            ]
+            return program_dict
+        elif self.proto_version == ProtoVersion.V2:
+            if isinstance(program, Schedule):
+                program.device.validate_schedule(program)
+            program = gate_set.serialize(program)
```
dabacon: need to assert that gate_set is not None.  May need to cast it to SerializableGateSet

PR #1911, opened 2019-08-03
Efficient integer powers of Pauli combinations

### LINKED ISSUE #1037: Powering vector of pauli components

viathor:
Fixes #1037. This is one of the last useful bits mined from #1299.

Follow-up: use the new algorithm in other classes.

### DIFF
--- a/cirq/linalg/operator_spaces_test.py
+++ b/cirq/linalg/operator_spaces_test.py
@@ -232,3 +232,64 @@ def test_expand_is_inverse_of_reconstruct(m1, basis):
     c2 = cirq.expand_matrix_in_orthogonal_basis(m2, basis)
     assert np.allclose(m1, m2)
     assert c1 == c2
+
+
+@pytest.mark.parametrize(
+    'coefficients,exponent',
+    itertools.product((
+        (0, 0, 0, 0),
+        (-1, 0, 0, 0),
+        (0.5, 0, 0, 0),
+        (0.5j, 0, 0, 0),
+        (1, 0, 0, 0),
+        (2, 0, 0, 0),
+        (0, -1, 0, 0),
+        (0, 0.5, 0, 0),
+        (0, 0.5j, 0, 0),
+        (0, 1, 0, 0),
+        (0, 2, 0, 0),
+        (0, 0, -1, 0),
+        (0, 0, 0.5, 0),
+        (0, 0, 0.5j, 0),
+        (0, 0, 1, 0),
+        (0, 0, 2, 0),
+        (0, 0, 0, -1),
+        (0, 0, 0, 0.5),
+        (0, 0, 0, 0.5j),
+        (0, 0, 0, 1),
+        (0, 0, 0, 2),
+        (0, -1, 0, -1),
+        (0, 1, 0, 1j),
+        (0, 0.5, 0, 0.5),
+        (0, 0.5j, 0, 0.5j),
+        (0, 0.5, 0, 0.5j),
+        (0, 1, 0, 1),
+        (0, 2, 0, 2),
+        (0, 0.5, 0.5, 0.5),
+        (0, 1, 1, 1),
+        (0, 1.1j, 0.5 - 0.4j, 0.9),
+        (0.7j, 1.1j, 0.5 - 0.4j, 0.9),
+        (0.25, 0.25, 0.25, 0.25),
+        (0.25j, 0.25j, 0.25j, 0.25j),
+        (0.4, 0, 0.5, 0),
+        (1, 2, 3, 4),
+        (-1, -2, -3, -4),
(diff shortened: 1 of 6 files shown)


viathor:
Seems like github ate the question? :-)
@dabacon

I guess the question might be about the inelegant split of the algorithm into two cases. Note that the cases can be merged. Specifically, one can cancel `v` between numerator and denominator and obtain an expression that works in all cases, whether `v` is zero or not. However, this yields a sum with O(exponent) terms whereas the present form is O(1). Ultimately, I decided to make this fast at the expense of elegance...

BTW: the correct handling of near-zero `v` matters since the algorithm is supposed to work for complex combinations, too. The resulting operators aren't Hermitian, but I think they will arise in practice, e.g. as the creation and annihilation operators of second quantization. In fact, these two provide an example where `v` ends up being zero!

dabacon:
Doh.  Question was what is the speed of that test.  I'm worried that the product plus explicit calculation will make it a dog.

viathor:
Ah, good question. On my workstation the additional 480 tests increased the runtime of operator_spaces_test.py by mere 0.55s. This modest cost is probably due to the fact that the calculation, even though explicit, is really O(1) time :-)

Let me know if you'd like me to remove some of the cases nonetheless.

```
(cirq) 14:11:13 viathor@ubik:~/src/Cirq$ git checkout master
Switched to branch 'master'
Your branch is ahead of 'origin/master' by 11 commits.
  (use "git push" to publish your local commits)
(cirq) 14:11:16 viathor@ubik:~/src/Cirq$ time pytest cirq/linalg/operator_spaces_test.py 
====================================================================== test session starts ======================================================================
platform linux -- Python 3.6.6, pytest-3.8.2, py-1.8.0, pluggy-0.12.0
benchmark: 3.2.2 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
plugins: cov-2.5.1, benchmark-3.2.2
collected 94 items                                                                                                                                              

cirq/linalg/operator_spaces_test.py ..............................................................................................                        [100%]

=================================================================== 94 passed in 2.58 seconds ===================================================================

real    0m3.166s
user    0m2.863s
sys     0m0.998s
(cirq) 14:11:22 viathor@ubik:~/src/Cirq$ git checkout powering_paulis 
Switched to branch 'powering_paulis'
Your branch is up to date with 'origin/powering_paulis'.
(cirq) 14:11:32 viathor@ubik:~/src/Cirq$ time pytest cirq/linalg/operator_spaces_test.py 
====================================================================== test session starts ======================================================================
platform linux -- Python 3.6.6, pytest-3.8.2, py-1.8.0, pluggy-0.12.0
benchmark: 3.2.2 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
plugins: cov-2.5.1, benchmark-3.2.2
collected 574 items                                                                                                                                             

cirq/linalg/operator_spaces_test.py ..................................................................................................................... [ 20%]
......................................................................................................................................................... [ 47%]
......................................................................................................................................................... [ 73%]
.......................................................................................................................................................   [100%]

================================================================== 574 passed in 3.13 seconds ===================================================================

real    0m3.729s
user    0m3.431s
sys     0m1.107s
(cirq) 14:11:38 viathor@ubik:~/src/Cirq$ 
```

viathor:
Oh, I guess you referred to the `matrix_power()` call in the test. I imagine this is fast since all the matrices are 2x2.

cirq/linalg/operator_spaces_test.py
```diff
@@ -232,3 +232,64 @@ def test_expand_is_inverse_of_reconstruct(m1, basis):
...
+        (0.7j, 1.1j, 0.5 - 0.4j, 0.9),
+        (0.25, 0.25, 0.25, 0.25),
+        (0.25j, 0.25j, 0.25j, 0.25j),
+        (0.4, 0, 0.5, 0),
+        (1, 2, 3, 4),
+        (-1, -2, -3, -4),
+        (-1, -2, 3, 4),
+        (1j, 2j, 3j, 4j),
+        (1j, 2j, 3, 4),
+    ), (0, 1, 2, 3, 4, 5, 10, 11, 20, 21, 100, 101)),
```
dabacon: maybe get rid of 10, 11, 20, 21?  don't quite see how those would pass when 100, 101 would succeed.

viathor: Done.

PR #1913, opened 2019-08-05
Make moment equality independent of ordering of operations

### LINKED ISSUE #1912: Circuit equality with moment order
When I run the following code the assertion fails.
```
d1 = cirq.GridQubit(4,4)
d2 = cirq.GridQubit(5,5)
m1 = cirq.GridQubit(5,4)
m2 =cirq.GridQubit(4,5)

circuit= cirq.Circuit()
circuit.append(cirq.I(m1))
circuit.append(cirq.I(m2))
circuit.append(cirq.Z(d1).controlled_by(m2))
circuit.append(cirq.X(d1).controlled_by(m1))#
circuit.append(cirq.Z(d2).controlled_by(m2))#
circuit.append(cirq.X(d2).controlled_by(m1))
circuit.append(cirq.Circuit.from_ops(cirq.measure(m1), cirq.measure(m2)))

should_be = cirq.Circuit()
should_be.append(cirq.I(m1))
should_be.append(cirq.I(m2))
should_be.append(cirq.X(d1).controlled_by(m1))
should_be.append(cirq.Z(d1).controlled_by(m2))#
should_be.append(cirq.X(d2).controlled_by(m1))#
should_be.append(cirq.Z(d2).controlled_by(m2))
should_be.append(cirq.Circuit.from_ops(cirq.measure(m1), cirq.measure(m2)))

assert circuit == should_be
```
These circuits are the same just that the order of the elements in the moment with the # is different.
```
(4, 4): ───────X───Z───────────
               │   │
(4, 5): ───I───┼───@───@───M───
               │       │
(5, 4): ───I───@───@───┼───M───
                   │   │
(5, 5): ───────────X───Z───────


(4, 4): ───────X───Z───────────
               │   │
(4, 5): ───I───┼───@───@───M───
               │       │
(5, 4): ───I───@───@───┼───M───
                   │   │
(5, 5): ───────────X───Z───────
```
I think this is because the moment is stored as a tuple of operations so the order matters.

c-poole:
Fixes #1912 .

### DIFF
--- a/cirq/ops/moment_test.py
+++ b/cirq/ops/moment_test.py
@@ -59,10 +59,6 @@ def test_equality():
     eq.add_equality_group(Moment([cirq.X(b)]))
     eq.add_equality_group(Moment([cirq.Y(a)]))
 
-    # Equality depends on order.
-    eq.add_equality_group(Moment([cirq.X(a), cirq.X(b)]))
-    eq.add_equality_group(Moment([cirq.X(b), cirq.X(a)]))
-
     # Two qubit gates.
     eq.make_equality_group(lambda: Moment([cirq.CZ(c, d)]))
     eq.make_equality_group(lambda: Moment([cirq.CZ(a, c)]))
--- a/cirq/ops/moment.py
+++ b/cirq/ops/moment.py
@@ -112,7 +112,7 @@ class Moment:
     def __eq__(self, other):
         if not isinstance(other, type(self)):
             return NotImplemented
-        return self.operations == other.operations
+        return set(self.operations) == set(other.operations)
 
     def _approx_eq_(self, other: Any, atol: Union[int, float]) -> bool:
         """See `cirq.protocols.SupportsApproximateEquality`."""


c-poole:
I agree that _approx_eq_ should also have this behavior and now that I'm looking back at this __hash__ will need it as well. I'm not sure if I share your concern about the size of the sets. This is a change at the moment level, and the number of operations in any given moment shouldn't exceed the number of qubits in the system you want to simulate by a large amount. GlobalPhase gates or other types of zero qubit operations are the only reason I can think of why the number of operations might exceed the number of qubits. Practically speaking this limits the size of the sets being created. 

The current implementation is has space cost of 2n (n for each of two sets) and time cost of 3n (2n for creation of two sets and n for comparison).
We could also iterate over `other.operations` to check if each element is in `set(self.operations)` which has a space cost of n and time cost of 2n.  
If space is still a concern we can avoid the sets altogether and just check if each element of `other.operations` is in `self.operations` for a n^2 cost in time.

How do these other two options sound?

cduck:
Don't use the n^2 option.  Since Moments are immutable, you could lazily create and store `frozenset(self.operations)` so it isn't recalculated for every comparison and hash.

What is your opinion about having two moments that compare equal but contain operations in different orders?

bryano:
What about ensuring that the operations are sorted at construction time?

c-poole:
I think two moments with an identical set of operations but with those operations in an otherwise different order should compare equal. Having the operations stored as a list inside Moments is convenient for allowing Moments to be part of an op_tree, but I think the ordering of operations in a Moment is arbitrary and therefore unimportant for comparisons. Two users could make circuits that look identical (with identical string representations even) but be confused about why their circuits are not actually equal do to the order in which operations are inserted while making the circuit. I don't think this should be the case. 
I am a bit concerned about the implication of having two Moments compare equal but having different repr() and str() calls. We may be able to patch that up by having str() and repr() use alphanumeric ordering to make sure equal Moments have equal repr() and str() calls. If there is a convenient ordering we could place on Operations we could also just sort the list at construction time.

cduck:
A good ordering for operations in this context would be this sort key: `lambda op: op.qubits`.

c-poole:
That sounds good. I'll see how it goes.

Strilanc:
In the past I proposed a change like this, but I also modified the moment data structure to store a set. This made @dabacon balk because it made it harder to work with moments (e.g. no indexing).

I like this variant of the change better, because you get most of the benefit without paying the runtime costs. (Most equating of moments happens in tests, I believe, and so it's okay for it to be a bit slower.) So I'm fine with merging it. 👍

cirq/ops/moment_test.py
```diff
@@ -59,10 +59,6 @@ def test_equality():
     eq.add_equality_group(Moment([cirq.X(b)]))
     eq.add_equality_group(Moment([cirq.Y(a)]))
 
-    # Equality depends on order.
-    eq.add_equality_group(Moment([cirq.X(a), cirq.X(b)]))
```
dabacon: think the test should be modified to add_equality_groups with these two instead of deleting (test that the order doesn't matter).

dabacon:
Ah was going to add the test and merge, but didn't notice the comment about the repr.

Strilanc:
I would just recreate the set for the equality each time, but it's fine to cache it.

Don't sort the operations in the constructor. We want users to be able to get at the operations in the order they passed in.

dabacon:
If we don't want to change the repr, and don't want to cache, then this is good to merge.

cduck:
This changes needs to be added to approximate equality as well before merge.

c-poole:
We seem to have a situation where `A == B` returns True when A and B are lists but `set(A) == set(B)` returns False. I think this is because sets use hashes of their elements to determine equality, and there are instances of `GateOperation`s that compare equal but do not have equal hashes.

c-poole:
Given the sympy hashing issue, there are two options I see us having for resolving this.

1) The n^2 option which is by far the simplest to implement and not terribly expensive for circuits that have numbers of qubits that can actually be simulated.

2) The n*Log(n) option which requires imposing an ordering on operations and comparing sorted lists of operations. Operations could inherit the ordering of their qubits as Casey suggested, but this wouldn't work as a strict total ordering since two Operations A and B that act on the same qubits but are otherwise not equal (for instance a different gate) break property that exactly one of A < B, A == B, A > B are True. Moments forbid overlapping operations, so this doesn't impose a problem for this specific use case, but having an ordering that behaves this way still bothers me. Do we expect Moments to get large enough for the performance difference between these two options to merit this trade-off?

Alternatively, if there is an obvious option that doesn't involve hashing that I've missed, I'm all ears. What do people think? @cduck , @dabacon , @Strilanc

c-poole:
For some reason I imagined I would be adding this ordering to operations generically rather than using it as a key for this specific purpose. The sorted operation should work as a key for sorted so there shouldn't be any issue.

PR #1914, opened 2019-08-06
adds missing methods to Circuit.__doc__

bryano:


cirq/circuits/circuit.py
```diff
@@ -59,11 +59,15 @@ class Circuit:
...
+        findall_operations_between
         findall_operations_until_blocked
         findall_operations_with_gate_type
+        reachable_frontier_from
+        has_measurements
         are_all_matches_terminal
         are_all_measurements_terminal
         unitary
+        final_wavefunction
         apply_unitary_effect_to_state
```
viathor: Remove `apply_unitary_effect_to_state()`. It is deprecated and users should use `final_wavefunction()` instead.

bryano: Done

PR #1916, opened 2019-08-07
Improves semantics of Circuit.findall_operations_until_blocked

bryano:
Now takes into account blocking operations on qubits in the 'light cone' of the start frontier.

I found the previous semantics quite surprising. See the added tests for examples of the new behavior.

Only one existing test had to be changed, so it seems like this shouldn't be too much of a breaking change.


cirq/circuits/circuit.py
```diff
@@ -666,16 +666,41 @@ def findall_operations_until_blocked(
...
-        operations that involve that qubit.  Operations are only considered
-        to be blocking the qubits that they operate on, so a blocking operation
-        that does not operate on any qubit in the starting frontier is not
-        actually considered blocking.  See `reachable_frontier_from` for a more
-        in depth example of reachable states.
+        Finds all operations until a blocking operation is hit.
+
+        An operation is considered blocking if
+
+        a) It is in the 'light cone' of start_frontier.
```
dstrain115: What does 'light cone' of start_frontier mean?

I think these semantics are pretty weird, but maybe I am not understanding it correctly.

cirq/circuits/circuit.py
```diff
@@ -666,16 +666,41 @@ def findall_operations_until_blocked(
...
+
+        Every qubit acted on by a blocking operation is thereafter itself
+        blocked.
+
+
+        The notion of reachability here differs from that in
+        reachable_frontier_from in two respects:
+
+        1) An operation is not considered blocking only because it is in a
+            moment before the start_frontier of one of the qubits on which it
```
dstrain115: This statement needs to be rephrased.  I've read it a few times now, and I am still confused as to what it means.  I would flip it around to make it a more direct statement.  Something like:

- An operation before the start_moment of one of the qubits is not considered blocking.

cirq/circuits/circuit_test.py
```diff
@@ -1137,61 +1154,166 @@ def test_findall_operations_until_blocked():
...
         cirq.H(a))
+    expected_diagram = """
+0: ───H───@───────────────────────────────────────@───H───
+          │                                       │
+1: ───────@───H───@───────────────────────@───H───@───────
+                  │                       │
+2: ───────────────@───H───@───────@───H───@───────────────
+                          │       │
+3: ───────────────────────@───H───@───────────────────────
+""".strip()
```
dstrain115: Nice, I like this approach much better than specifying it as a comment!

cirq/circuits/circuit.py
bryano: That's not quite true. If an operation acts on two qubits in a moment that is before the start frontier of one but after the start frontier of the other, then it might be considered blocking.

cirq/circuits/circuit.py
bryano: Basically, if you create the circuit DAG (using the default `can_reorder`), I think of the "light cone" of an operation as the successors of that node in the DAG. The light cone of a position (qubit, moment_index) is the set of positions at which operations would be in the light cone of an operation at the first position.

bryano: With the current implementation (in master), for

the circuit
```
0: ───ZZ────────────
      │
1: ───ZZ───ZZ───Z───
           │
2: ───XX───ZZ───────
      │
3: ───XX────────────
```

`start_frontier[q] = 0` for all four qubits and `is_blocker = lambda op: isinstance(op.gate, cirq.XXPowGate)`, 

`findall_all_operations_until_blocked` returns all the Z/ZZ gates.

That seems surprising, because all but the first one "should" be blocked by the XX.

dabacon:
@Strilanc could you take a look as this is the method you wrote.

vtomole:
Ping @bryano

bryano:
@Strilanc PTAL

bryano:
@Strilanc Could you help me out with the docs error? I don't understand what the problem is.

bryano:
@Strilanc ping

Issue #1900, opened 2019-07-30
extras_require doesn't work with wildcard requirements


dabacon:
In setup.py we now have extras_require to only install some dependencies when requested.  Like
```
pip install cirq[contrib]
```
However this appears to not work when we use this setup.py to create a wheel and upload it to pypi.  It yells about being unable to support a version when publishing the package.
```
HTTPError: 400 Client Error: Invalid value for requires_dist. Error: Invalid version: '1.3.*' for url: https://upload.pypi.org/legacy/
```

cduck:
I was able to reproduce the error.  Somehow it works for this package: https://pypi.org/project/RelStorage/#files

dabacon:
Fixed in #1935

PR #1917, opened 2019-08-07
Roll back extras_require

dabacon:
See #1900   extras_require doesn't seem to work well with wildcard requirements on pypi.  

We should roll this back until we understand if there is a better solution since this blocks easy dev releases.


docs/install.md
```diff
@@ -34,13 +34,7 @@ That way you have control over when a breaking change affects you.
...
-3. (Optional) install other dependencies.
-
-    Install dependencies of features in `cirq.contrib`.
-
-    ```bash
-    python -m pip install cirq[contrib]
-    ```
+3. (Optional) install system dependencies that pip can't handle.
 
     Install system dependencies that pip can't handle.
```
cduck: Delete this line.

docs/install.md
```diff
@@ -34,13 +34,7 @@ That way you have control over when a breaking change affects you.
...
```
cduck:
```suggestion
```

docs/install.md
dabacon: done

PR #1925, opened 2019-08-08
Remove whitespace in density simulator test

dstrain115:
- Numpy does not have a consistent whitespace across
versions, so this test can fail in other contexts

### DIFF
--- a/cirq/sim/density_matrix_simulator_test.py
+++ b/cirq/sim/density_matrix_simulator_test.py
@@ -667,6 +667,9 @@ def test_density_matrix_trial_result_str():
         measurements={},
         final_simulator_state=final_simulator_state)
 
-    assert str(result) == ('measurements: (no measurements)\n'
-                           'final density matrix:\n'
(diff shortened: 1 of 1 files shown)


cirq/sim/density_matrix_simulator_test.py
```diff
@@ -667,6 +667,9 @@ def test_density_matrix_trial_result_str():
         measurements={},
         final_simulator_state=final_simulator_state)
 
-    assert str(result) == ('measurements: (no measurements)\n'
-                           'final density matrix:\n'
-                           '[[0.5 0.5]\n [0.5 0.5]]')
+    # numpy varies whitespace in its representation for differnt versions
```
vtomole:
```suggestion
    # numpy varies whitespace in its representation for different versions
```

dstrain115:
> Which numpy version should i use to reproduce this?

Looks like the one that I was using was 1.13.3, which does violate cirq's requirements.txt, but such is life.

I am not quite sure when they changed the whitespace for ndarrays.

PR #1927, opened 2019-08-08
Add Clifford simulator

BryceKobrin:


dabacon:
@BryceKobrin thanks for submitting this, will try to get to a review soon.  If you have any questions about the CLA let me know.

BryceKobrin:
@dabacon thanks for following up. I've submitted the CLA based on my @berkeley.edu account since I no longer have access to my LDAP. Let me know if there's anything else I need to do.

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+            state: A CliffordState
+            measurements: A dictionary from measurement gate key to measurement
+                results, ordered by the qubits that the measurement operates on.
+            qubit_map: A map from the Qubits in the Circuit to the the index
+                of this qubit for a canonical ordering. This canonical ordering
+                is used to define the state vector (see the state_vector()
+                method).
+        """
+        self.measurements = measurements
+        # self.qubit_map = qubit_map
```
dabacon: remove commented out code

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+            measurements.append(self.state.perform_measurement(qubits,collapse_wavefunction=False))
+        
+        return np.array(measurements,dtype=bool)
+
+class CliffordState():
+    """A state of the Clifford simulation. 
+
+    The state is stored using two complementary representations: Anderson's tableaux form
+    and Bravyi's CH-form. The tableaux keeps track of the stabilizer operations, while the
+    CH-form allows access to the full wavefunction (including phase).
```
dabacon: It feels strange to keep track of both, especially for a simulator where we do care about efficiency.  I think you should pass which of these you want into the simulator constructor and then use that to decide which state to use.  You probably will end up with two state classes, which is ok, especially since they support different ways to get information about the simulation out of them.

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+            state = self.copy()
+        
+        for qubit in qubits:
+            result = state.tableau._measure(self.qubit_map[qubit])
+            state.CH_form.project_Z(self.qubit_map[qubit],result)
+            results.append(result)
+        
+        return results
+
+class CliffordTableau():
```
Strilanc: Split this class out into a separate file.

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+            self.zs[p,:] = False
+
+            self.zs[p,q] = True
+
+            self.rs[p] = bool(np.random.randint(2))
+
+            return int(self.rs[p])
+
+
+class CH_Form():
```
Strilanc: Split this class out into a separate file, and give it a more descriptive name.

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+    '''
```
Strilanc: Nit: always use double quotes instead of single quotes for docstrings.

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+    A representation of stabilizer states using the CH form,
+
+        |psi> = omega U_C U_H |s>
```
Strilanc: You can use dollar signs to make this show up as latex in the docs.

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+
+    This representation keeps track of overall phase.
+
+    See Bravyi et al, 2016 for details.
```
Strilanc: Give a more explicit reference, including URL and title.

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+        self.omega = 1
+
+        def bits(s):
+            while s > 0:
+                yield s & 1
+                s >>= 1
+
+        # Apply X for every non-zero element of initial_state
+        for (i,val) in enumerate(bits(initial_state)):
+            if val:
```
Strilanc: `if (initial_state >> i) & 1:`

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+    '''
+    A representation of stabilizer states using the CH form,
+
+        |psi> = omega U_C U_H |s>
+
+    This representation keeps track of overall phase.
+
+    See Bravyi et al, 2016 for details.
+    '''
+    def __init__(self,num_qubits,initial_state=0):
```
Strilanc: The constructor should allow initializing all aspects of the state directly (e.g. take each of the matrices as arguments and do some basic sanity checks on them w.r.t their size). Use a static factory method to produce the zero state.

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+        copy.gamma = self.gamma.copy()
+        copy.v = self.v.copy()
+        copy.s = self.s.copy()
+        copy.omega = self.omega
+
+        return copy
+    
+    def __repr__(self):
+        ''' Return the wavefunction representation of the state. '''
+        return cirq.dirac_notation(self.to_numpy())
```
Strilanc: The repr has to be an evaluable python expression that returns an instance equal to the original class.

(E.g. after modifying the constructor this would just be passing the current instance's attributes' reprs into the constructor)

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+        string += "G:\n{}\n".format(np.array(self.G,dtype=int))
+        string += "F:\n{}\n".format(np.array(self.F,dtype=int))
+        string += "M:\n{}\n".format(np.array(self.M,dtype=int))
+        string += "gamma: {}\n".format(np.array(self.gamma,dtype=int))
+        string += "v: {}\n".format(np.array(self.v,dtype=int))
+        string += "s: {}\n".format(np.array(self.s,dtype=int))
+        
+        return string
+    
+    def _int2arr(self,x):
```
Strilanc: `cirq.big_endian_int_to_bits`

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+        i = self.n-1
+        while x > 0:
+            if x & 1:
+                arr[i] = True
+            i -= 1
+            x >>= 1
+        
+        return arr
+        
+    def inner_product(self,x):
```
Strilanc: This name is not very clear.

Add type annotations.

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+            
+        mu = sum(x * self.gamma)
+        
+        u = np.zeros(self.n,dtype=bool)
+        for p in range(self.n):
+            if x[p]:
+                u ^= self.F[p,:]
+                mu += 2*(sum(self.M[p,:] & u) % 2)
+                
+        return self.omega*2**(-sum(self.v)/2)*1j**mu*(-1)**sum(self.v & u & self.s)*np.all(self.v | (u==self.s))
```
Strilanc: Split this into a few lines.

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+
+    def wave_function(self):
+        wf = np.zeros(2**self.n,dtype=complex)
+        
+        for x in range(2**self.n):
+            wf[x] = self.inner_product(x)
+            
+        return wf
+    
+    def _S(self,q,right=False):
```
Strilanc: What's `right`? Is it whether or not to do an inverse S?

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+    
+    def _update_sum(self,t,u,delta=0,alpha=0):
+        ''' Implements the transformation (Proposition 4 in Bravyi et al)
+
+                i^alpha U_H (|t> + i^delta |u>) = omega W_C W_H |s'>
+        '''
+        if np.all(t == u):
+            self.s = t
+            self.omega *= 1/np.sqrt(2)*(-1)**alpha*(1+1j**delta)
+        else:
```
Strilanc: Add `return` to the if block so the else block can be unindented

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+                
+            (omega,a,b,c) = self._H_decompose(self.v[q],y[q],z[q],delta)
+            
+            self.s = y
+            self.s[q] = c
+            self.omega *= (-1)**alpha*omega
+            
+            if a:
+                self._S(q,right=True)
+            self.v[q] ^= b ^ self.v[q]
```
Strilanc: Isn't this just `self.v[q] = b`? Or maybe `self.v[q][:] = b`

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+
+    def _H_decompose(self,v,y,z,delta):
```
Strilanc: This method doesn't use self; it can be moved out of the class

cirq/sim/clifford_simulator.py
```diff
@@ -0,0 +1,749 @@
...
+                omega = (-1)**(c & y)
+            else:
+                omega = 1/np.sqrt(2)*(1+1j**delta)
+                b = True
+                a = True
+                c = not ((delta >> 1) ^ y)
+
+        return omega,a,b,c
+
+    def to_numpy(self):
```
Strilanc: Document, add type annotations, give a better name (e.g. `to_state_vector`)

cirq/sim/clifford_simulator_test.py
```diff
@@ -0,0 +1,359 @@
...
+    (q0,q1) = (cirq.LineQubit(0),cirq.LineQubit(1))
+    circuit = cirq.Circuit()
+
+    np.random.seed(2)
+
+    for _ in range(10):
+        x = np.random.randint(7)
+
+        if x == 0:
+            circuit.append(cirq.X(np.random.choice((q0,q1))))
```
Strilanc: Make a list of gates and `random.choice` out of it, instea dof manually indexing like this,

cirq/sim/clifford_simulator_test.py
```diff
@@ -0,0 +1,359 @@
...
+        elif x == 6:
+            circuit.append(cirq.CZ(q0,q1))
+
+    circuit.append(cirq.measure(q0))
+    result = cirq.CliffordSimulator().run(circuit, repetitions=100)
+
+    assert sum(result.measurements['0'])[0] < 80
+    assert sum(result.measurements['0'])[0] > 20
+
+def test_clifford_circuit_3():
```
Strilanc: Is this just a copy paste of the previous test? Use `pytest.mark.parameterize` t omerge them.

cirq/sim/clifford_simulator.py
BryceKobrin: This is a great point, and I agree that decoupling the two representations would be a more elegant solution. However there are two caveats. 

1) I'm not sure how easy it is to perform measurements with Bravyi's representation alone (the current implementation relies on the tableaux for performing measurements). Unless we can figure this out, the Bravyi-based simulation would be limited to unitary operators only. 

2) We can drop the Bravyi representation entirely without too much loss of functionality -- the only thing missing would be the ability to get wavefunction amplitudes in the computational basis. But this functionality is necessary if one wants to improve the efficiency of non-Clifford operations (https://arxiv.org/abs/1808.00128).

cirq/sim/clifford_simulator.py
vtomole: Let's keep this name because it's what the authors call it and it's as descriptive as they get: https://arxiv.org/abs/1808.00128.

vtomole:
(@Strilanc, @dabacon) @BryceKobrin didn't sign the CLA. We can still use this, correct?

cirq/sim/clifford_simulator.py
vtomole: We have been defining methods that don't use `self` everywhere in the codebase. How is this the exception?

PS: There is a `no-self-use` setting in pylint that catches this. We could turn it on and fix all of them.

cirq/sim/clifford_simulator.py
vtomole: Done.

cirq/sim/clifford_simulator_test.py
vtomole: Done.

cirq/sim/clifford_simulator.py
vtomole: Done.

cirq/sim/clifford_simulator.py
vtomole: Split into 2 for now.

cirq/sim/clifford_simulator.py
vtomole: Done.

cirq/sim/clifford_simulator.py
vtomole: Done.

cirq/sim/clifford_simulator.py
vtomole: Done.

cirq/sim/clifford_simulator.py
vtomole: Done.

cirq/sim/clifford_simulator.py
vtomole: Renaming to `to_state_vector` but not documenting because the name documents it.

cirq/sim/clifford_simulator.py
vtomole: Changed to `inner_product_of_state_and_x`.

cirq/sim/clifford_simulator_test.py
vtomole: Done.

cirq/sim/clifford_simulator.py
vtomole: @dabacon Let's get this in and address decoupling the representations here: https://github.com/quantumlib/Cirq/issues/2123.

cirq/sim/clifford_simulator.py
vtomole: These matrices?

```python
self.G = np.eye(self.n, dtype=bool)
self.F = np.eye(self.n, dtype=bool)
self.M = np.zeros((self.n, self.n), dtype=bool)
self.gamma = np.zeros(self.n, dtype=int)
self.v = np.zeros(self.n, dtype=bool)
self.s = np.zeros(self.n, dtype=bool)
```
If so, why would we want to pass them in if they are all initialized by `n`?
> Use a static factory method to produce the zero state.

Please explain further.

cirq/sim/clifford_simulator.py
vtomole: Done.

cirq/sim/clifford_simulator.py
vtomole: On second thought, let's put `CH_Form` into a different PR.

PR #1928, opened 2019-08-09
Add SimulatesAmplitudes simulator interface

### LINKED ISSUE #1922: SimulatesAmplitudes simulator interface
Some simulators (such as those based on contracting tensor networks) compute individual amplitudes of bitstrings without storing the entire wavefunction.

kevinsung:
Fixes #1922


cirq/sim/simulator.py
```diff
@@ -177,6 +177,74 @@ def compute_samples_displays_sweep(
         return compute_displays_results
 
 
+class SimulatesAmplitudes(metaclass=abc.ABCMeta):
```
dabacon: add to documentation at top of class

cirq/sim/simulator.py
```diff
@@ -177,6 +177,74 @@ def compute_samples_displays_sweep(
...
+            param_resolver: 'study.ParamResolverOrSimilarType' = None,
+            qubit_order: ops.QubitOrderOrList = ops.QubitOrder.DEFAULT,
+    ) -> List[complex]:
+        """Computes the desired amplitudes.
+
+        The initial state is assumed to be the all zeros state.
+
+        Args:
+            program: The circuit or schedule to simulate.
+            bitstrings: The bitstrings whose amplitudes are desired, input
```
dabacon: @cduck can you comment on how this would play with your qudit work?

cduck: This is fine as it can easily be changed from bools to ints.

I think this API could be improved by making it match more closely with how other simulators return measurement results.  E.g. outputs=[[mesurement key 'a'=0b100110, key 'b'=5, ...], ...]

Is it possible for this kind of simulator to not measure some qubits at the end?

kevinsung: Usually with this kind of simulator qubits are not measured at all.

cirq/sim/wave_function_simulator.py
```diff
@@ -24,7 +24,8 @@
 from cirq.sim import simulator, wave_function
 
 
-class SimulatesIntermediateWaveFunction(simulator.SimulatesIntermediateState,
+class SimulatesIntermediateWaveFunction(simulator.SimulatesAmplitudes,
+                                        simulator.SimulatesIntermediateState,
```
dabacon: Definitely this is valid, but it kind of defeats the purpose of the interface as an indication that it calculates the amplitude efficiently?  I worry that this encourages using this type of simulator for just amplitudes.

kevinsung: I think this is convenient and it allows people to start using the interface immediately with the sparse simulator.

cirq/sim/simulator.py
dabacon: ok.

cirq/sim/simulator.py
kevinsung: Done.

PR #1929, opened 2019-08-09
Add ResetChannel

### LINKED ISSUE #1774: Add Reset channel
The reset channel is equivalent to performing an unobserved measurement which then controls a bit flip onto the targeted qubit. It is also equivalent to tracing out a qubit and then swapping a fresh 0 qubit into the system.

andbe91:
Closes #1774

### DIFF
--- a/cirq/ops/common_channels.py
+++ b/cirq/ops/common_channels.py
@@ -447,6 +447,94 @@ def amplitude_damp(gamma: float) -> AmplitudeDampingChannel:
 
 
 @value.value_equality
+class ResetChannel(gate_features.SingleQubitGate):
+    """Reset a qubit back to its $|0\rangle$ state.
+
+    The reset channel is equivalent to performing an unobserved measurement 
+    which then controls a bit flip onto the targeted qubit.
+    """
+
+    def __init__(self) -> None:
+        r"""The reset channel.
+
+        Construct a channel that resets the qubit.
+
+        This channel evolves a density matrix as follows:
+
+            $$
+            \rho \rightarrow M_0 \rho M_0^\dagger + M_1 \rho M_1^\dagger
+            $$
+
+        With:
+
+            $$
+            \begin{aligned}
+            M_0 =& \begin{bmatrix}
+                    1 & 0  \\
+                    0 & 0
+                  \end{bmatrix}
+            \\
+            M_1 =& \begin{bmatrix}
+                    0 & 1 \\
+                    0 & 0
+                  \end{bmatrix}
+            \end{aligned}
+            $$
+
+        """
+        self._delegate = AmplitudeDampingChannel(1.0)
+
+    def _channel_(self) -> Iterable[np.ndarray]:
+        return self._delegate._channel_()
+
+    def _has_channel_(self) -> bool:
+        return True
+
+    def _value_equality_values_(self):
+        return 1
+
+    def __repr__(self) -> str:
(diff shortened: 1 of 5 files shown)


cirq/ops/common_channels.py
```diff
@@ -446,6 +446,94 @@ def amplitude_damp(gamma: float) -> AmplitudeDampingChannel:
...
+
+    def __str__(self) -> str:
+        return 'reset()'
+
+    def _circuit_diagram_info_(self,
+                               args: protocols.CircuitDiagramInfoArgs) -> str:
+        return 'R'
+
+
+def reset() -> ResetChannel:
```
cduck: `reset` should be an instance of `ResetChannel` because there are no parameters: `reset = ResetChannel()`

Instead of this usage:
```
circuit.append(cirq.reset()(qubit))
```

I would prefer this usage:
```
circuit.append(cirq.reset(qubit))
```

You could move this docstring to the class docstring.

cirq/ops/common_channels.py
```diff
@@ -446,6 +446,94 @@ def amplitude_damp(gamma: float) -> AmplitudeDampingChannel:
...
+        self._delegate = AmplitudeDampingChannel(1.0)
+
+    def _channel_(self) -> Iterable[np.ndarray]:
+        return self._delegate._channel_()
+
+    def _has_channel_(self) -> bool:
+        return True
+
+    def _value_equality_values_(self):
+        return 1
```
cduck: `return None` or `return ()` might make it more clear that there are no values to compare.

cirq/ops/common_channels.py
```diff
@@ -446,6 +446,94 @@ def amplitude_damp(gamma: float) -> AmplitudeDampingChannel:
     return AmplitudeDampingChannel(gamma)
 
 
+@value.value_equality
+class ResetChannel(gate_features.SingleQubitGate):
+    """Reset a qubit back to its $|0>$ state.
```
cduck: Math mode won't change how |0> looks.  Use `|0⟩`.

cirq/ops/common_channels.py
andbe91: Done.
Is it enough to have the density matrix evolution in the `__init__` docstring, or do you want me to also add it to the class docstring?

cirq/ops/common_channels.py
andbe91: Done.

cirq/ops/common_channels.py
andbe91: Done.

cirq/ops/common_channels.py
```diff
@@ -446,6 +446,67 @@ def amplitude_damp(gamma: float) -> AmplitudeDampingChannel:
     return AmplitudeDampingChannel(gamma)
 
 
+@value.value_equality
+class ResetChannel(gate_features.SingleQubitGate):
```
dabacon: It's not clear that there isn't a need for a more general reset channel that resets to states different that |0>.  This is often used in QEC.

andbe91: I can implement this in a following PR. Would resetting to 0, 1, +, -, etc be enough or are you envisioning reseting to arbitrary single qubit states? For full control I guess one would like to be able to specify that a measured 0 resets to X and a measured 1 goes to Y, where X and Y are some arbitrary states.

PR #1935, opened 2019-08-12
Add back in extras_require by using compatible versions

dabacon:
Instead of using `==v.w.*` this uses `~=v.w` the compatibility operator.  Compatibility is `==v.* >=v.w`.  This is nearly the same thing with some sublties about prelease versions that shouldn't hit us.

I've tested this by pushing to test pypi and prod pypi and it works.

### DIFF
--- a/dev_tools/conf/pip-list-dev-tools.txt
+++ b/dev_tools/conf/pip-list-dev-tools.txt
@@ -24,4 +24,4 @@ sphinx_rtd_theme
 sphinx-markdown-tables
 
 # Need to pin pylint's parser to 2.1 instead of 2.2 until https://github.com/PyCQA/astroid/issues/650 is fixed
-astroid==2.1.0
+astroid~=2.1.0
--- a/dev_tools/conf/pip-list-dev-tools.txt
+++ b/dev_tools/conf/pip-list-dev-tools.txt
@@ -1,27 +1,27 @@
 # For testing and analyzing code.
-mypy==0.701.*
-pylint==2.3.*
-pytest==3.8.*
-pytest-cov==2.5.*
-pytest-benchmark==3.2.*
-yapf==0.27.*
+mypy~=0.701
+pylint~=2.3
+pytest~=3.8
+pytest-cov~=2.5
+pytest-benchmark~=3.2
+yapf~=0.27
 
 # For generating protobufs
-grpcio-tools==1.21.*
+grpcio-tools~=1.21
 mypy-protobuf==1.10
 
 # For uploading packages to pypi.
 twine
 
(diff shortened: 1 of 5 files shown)


cduck:
Something went wrong with Travis and crashed pylint.

dev_tools/conf/pip-list-dev-tools.txt
```diff
@@ -1,27 +1,27 @@
...
-pytest==3.8.*
-pytest-cov==2.5.*
-pytest-benchmark==3.2.*
-yapf==0.27.*
+mypy~=0.701
+pylint~=2.3
+pytest~=3.8
+pytest-cov~=2.5
+pytest-benchmark~=3.2
+yapf~=0.27
```
cduck: I think you need to do `yapf~=0.27.0` instead (https://stackoverflow.com/questions/39590187/in-requirements-txt-what-does-tilde-equals-mean).  Same for most of the others.

dev_tools/conf/pip-list-dev-tools.txt
```diff
@@ -1,27 +1,27 @@
...
 twine
 
 # For verifying behavior of qasm output.
-qiskit==0.6.*
+qiskit~=0.6
 
 # For generating documentation.
 pypandoc
-recommonmark >= 0.4.0
+recommonmark ~= 0.4.0
```
cduck: This limits the allowed version to [0.4.0, 0.5) instead of [0.4.0, inf).  Is this OK for this package?

dev_tools/conf/pip-list-dev-tools.txt
dabacon: done

dev_tools/conf/pip-list-dev-tools.txt
dabacon: moved it back, though I think `>=` is considered dangerous.

dev_tools/conf/pip-list-dev-tools.txt
```diff
@@ -24,4 +24,4 @@ sphinx_rtd_theme
 sphinx-markdown-tables
 
 # Need to pin pylint's parser to 2.1 instead of 2.2 until https://github.com/PyCQA/astroid/issues/650 is fixed
-astroid==2.1.*
+astroid==2.1.0
```
cduck: '~='?

dabacon:
PTAL

dev_tools/conf/pip-list-dev-tools.txt
dabacon: fixed

PR #1937, opened 2019-08-12
Fix PhasedXPowGate circuit diagram

### LINKED ISSUE #1933: ValueError for PhasedXPowGate with integer phase_exponent
`PhasedXPowGate._circuit_diagram_info_` contains a format string that throws a `ValueError` when `phase_exponent` is an `int`.  I know the type annotation specifies `float` but mypy allows int.

```
c = cirq.Circuit.from_ops(
    cirq.PhasedXPowGate(phase_exponent=1)(cirq.LineQubit(0)),
)
print(c)
```
Output:
```
...
~/docs/cirq/cirq/ops/phased_x_gate.py in _circuit_diagram_info_(self, args)
    154         else:
    155             s = 'PhasedX({{:.{}}})'.format(args.precision).format(
--> 156                 self.phase_exponent)
    157         return protocols.CircuitDiagramInfo(
    158             wire_symbols=(s,),

ValueError: Precision not allowed in integer format specifier
```

cduck:
For integer `phase_exponent`.

Fixes #1933.

### DIFF
--- a/cirq/ops/phased_x_gate.py
+++ b/cirq/ops/phased_x_gate.py
@@ -152,7 +152,7 @@ class PhasedXPowGate(gate_features.SingleQubitGate):
                 args.precision is None):
             s = 'PhasedX({})'.format(self.phase_exponent)
         else:
-            s = 'PhasedX({{:.{}}})'.format(args.precision).format(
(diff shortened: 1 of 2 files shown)


dabacon:
Thanks for the fix!

PR #1939, opened 2019-08-13
TESTING: Github Actions

vtomole:
Background: https://developer.github.com/actions/

Adopting Github actions to replace Travis will require a `Cirq/.github/workflows/pythonpackage.yml`. 

This could make implementing nightly builds like https://github.com/quantumlib/Cirq/pull/1938 easier.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -1,0 +1,27 @@
+name: Python package
+
+on: [push]
+
+jobs:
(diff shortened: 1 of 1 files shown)


vtomole:
After some reading (https://knapsackpro.com/ci_comparisons/github-actions/vs/travis-ci), I don't see how this would give us an advantage over Travis.

CC: @Strilanc @dabacon @cduck

PR #1941, opened 2019-08-13
Cleanup: remove instances of abstractmethod for properties

dabacon:
@abstractproperty is deprecated after python 3.3

### DIFF
--- a/cirq/contrib/acquaintance/executor.py
+++ b/cirq/contrib/acquaintance/executor.py
@@ -40,14 +40,16 @@ class ExecutionStrategy(metaclass=abc.ABCMeta):
 
     keep_acquaintance = False
 
-    @abc.abstractproperty
+    @abc.abstractmethod
+    @property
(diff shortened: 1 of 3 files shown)


cirq/contrib/acquaintance/executor.py
```diff
@@ -40,14 +40,16 @@ class ExecutionStrategy(metaclass=abc.ABCMeta):
 
     keep_acquaintance = False
 
-    @abc.abstractproperty
+    @abc.abstractmethod
+    @property
```
cduck: Unfortunately the order matters.  https://docs.python.org/3/library/abc.html#abc.abstractmethod

dabacon: doh

cduck:
Is there a pylint check to make sure we don't use deprecated functions like this?

dabacon:
I don't see a pylint check for things like this.

Issue #1930, opened 2019-08-09
Unwinding Symbolic formulas and parameters?


dabacon:
Quite accidentally I ran an optimization pass which ended up working correctly but changing a parameter to a formula in sympy, like Symbol('a') + 1.23.

It occurred to me that if we allow this, then we could support a flow where you
a) do compilations that support formulas in parameters
b) create parameter sweeps
c) unwind the sweeps and then calculate the values of the formulas in the parameters. I think this could have a problem where you have different formulas for the same parameter or parameters entering multiple times, but I don't see a fundamental problem with this.
d) now you have some parameters and some parameterized circuits which you can send off to a simulator or a quantum computer!

cduck:
I did the same thing while decomposing a gate into e.g. [X**(gate_param+0.5), ...].  This would be useful.

It turns out that many gate's `_circuit_diagram_info_` crash if the gate parameter is a formula.

cduck:
@dabacon Am I missing any features not in the referencing PRs?

dabacon:
I think this is great!  The one thing I don't see in the ParamFlattener is the case where teh formula is the same and a small helper to do this all in one pass.

cduck:
What do you mean by "the case where the formula is the same"?

dabacon:
The case where `Symbol('a') + 1` appears in multiple gates.

cduck:
Yes.  That is already handled in #1963.

I just found that this doesn't work as expected when two gates have `a+1` and `a+1.0` (#1997).

PR #1944, opened 2019-08-14
Change instance checks for sympy.Symbol to sympy.Basic

cduck:
For #1930.

Improves support for sympy formulas as gate parameters.  Before this, printing the circuit diagram for circuits containing certain gates with formula parameters would crash.

### DIFF
--- a/cirq/google/api/v1/programs.py
+++ b/cirq/google/api/v1/programs.py
@@ -350,7 +350,7 @@ def _parameterized_value_from_proto_dict(message: Dict
 def _parameterized_value_to_proto_dict(param: Union[sympy.Basic, float]
                                       ) -> Dict:
     out = {}  # type: Dict
-    if isinstance(param, sympy.Basic):
+    if isinstance(param, sympy.Symbol):
         out['parameter_key'] = str(param.free_symbols.pop())
     else:
         out['raw'] = float(param)
--- a/cirq/google/api/v1/programs.py
+++ b/cirq/google/api/v1/programs.py
@@ -350,7 +350,7 @@ def _parameterized_value_from_proto_dict(message: Dict
 def _parameterized_value_to_proto_dict(param: Union[sympy.Basic, float]
                                       ) -> Dict:
     out = {}  # type: Dict
-    if isinstance(param, sympy.Symbol):
+    if isinstance(param, sympy.Basic):
         out['parameter_key'] = str(param.free_symbols.pop())
     else:
         out['raw'] = float(param)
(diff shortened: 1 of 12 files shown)


cirq/google/api/v1/programs.py
```diff
@@ -350,7 +350,7 @@ def _parameterized_value_from_proto_dict(message: Dict
 def _parameterized_value_to_proto_dict(param: Union[sympy.Basic, float]
                                       ) -> Dict:
     out = {}  # type: Dict
-    if isinstance(param, sympy.Symbol):
+    if isinstance(param, sympy.Basic):
```
dabacon: will the result in a formula trying to be passed in to parameter key?  for the proto stuff we only really support string symbols right now.

cduck: Oops.  I was find-replacing too fast.

cduck: Fixed.

cirq/ops/phased_x_gate.py
```diff
@@ -148,7 +148,7 @@ def _circuit_diagram_info_(self, args: protocols.CircuitDiagramInfoArgs
                                ) -> protocols.CircuitDiagramInfo:
         """See `cirq.SupportsCircuitDiagramInfo`."""
 
-        if (isinstance(self.phase_exponent, sympy.Symbol) or
+        if (isinstance(self.phase_exponent, sympy.Basic) or
```
dabacon: add an example of this to the test_diagram test case.

cduck:
I moved set-dev-version.sh to it's own Travis job because I don't think before deploy will run unless the condition is already met.  (https://docs.travis-ci.com/user/deployment/pypi/#running-commands-before-and-after-release)  This has the benefit that every job won't try push its own dev release (each with a different timestamp).

It seems that `*` characters need to be quoted otherwise Travis doesn't run.

@dabacon PTAL.

cduck:
(Dismissed review for wrong PR)

cduck:
@dabacon PTAL.

cirq/optimizers/eject_phased_paulis_test.py
```diff
@@ -310,28 +315,26 @@ def test_blocked_by_unknown_and_symbols():
...
+        [cirq.X(a)],
+        [cirq.CZ(a, b)**sym],
+        [cirq.X(a)],
+    ),
+                     expected=quick_circuit(
+                         [cirq.X(a)],
+                         [cirq.CZ(a, b)**sym],
+                         [cirq.X(a)],
+                     ),
+                     compare_unitaries=False)
```
cduck: I'm not sure what yapf did here.

PR #1947, opened 2019-08-14
Fix Travis deploy condition

cduck:
Follow up to #1938.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,28 +8,36 @@ branches:
   - master
 before_install:
 - pip install --upgrade pip setuptools wheel
+stages:
+- test
+- name: deploy
+  if: branch = master
 matrix:
   include:
-  - os: linux
+  - stage: test
+    os: linux
     env: NAME=mypy
     python: '3.6'
     install:
     - cat dev_tools/conf/pip-list-dev-tools.txt | grep mypy | xargs pip install
     script: check/mypy
-  - os: linux
+  - stage: test
+    os: linux
     env: NAME=pylint
     python: '3.6'
     install:
     - cat dev_tools/conf/pip-list-dev-tools.txt | grep "pylint\|astroid" | grep -v
       "#" | xargs pip install
     script: check/pylint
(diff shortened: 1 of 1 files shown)


dabacon:
For some reason Travis did not pick this up.

cduck:
I moved set-dev-version.sh to it's own Travis job because I don't think before deploy will run unless the condition is already met.  (https://docs.travis-ci.com/user/deployment/pypi/#running-commands-before-and-after-release)  This has the benefit that every job won't try push its own dev release (each with a different timestamp).

It seems that `*` characters need to be quoted otherwise Travis doesn't run.

@dabacon PTAL.

cduck:
I think the condition is working now.

.travis.yml
```diff
@@ -60,22 +70,28 @@ matrix:
...
     env: NAME=doctest
     python: '3.6'
     install:
     - pip install -r requirements.txt
     - pip install -r cirq/contrib/contrib-requirements.txt
     script: check/doctest -q
+  - stage: deploy
+    os: linux
+    env: NAME=dev-release
+    python: '3.7'
```
dabacon: why 3.7?  I think we say we support python 3.6

dabacon:
One comment about python version.

.travis.yml
cduck: I just picked one.  This would just be the python version that runs setup.py.

dabacon: ah k

cduck:
Deploy ran but failed:
```
Uploading distributions to https://test.pypi.org/simple/

Uploading cirq-0.6.0.dev0-py3-none-any.whl

100%|████████████████████████████████████████| 851k/851k [00:00<00:00, 2.27MB/s]

NOTE: Try --verbose to see response content.

HTTPError: 405 Client Error: Method Not Allowed for url: https://test.pypi.org/simple/

No stash entries found.

PyPI upload failed.

failed to deploy
```

PR #1953, opened 2019-08-15
Fix Travis dev-release URL and version

cduck:
Follow up to #1947.

- Changes the PyPI server to https://test.pypi.org/legacy/.
- Exports CIRQ_DEV_VERSION so setup.py can see it.


cduck:
This will need to be merged to test it.  Travis won't deploy to PyPI from a PR.

cduck:
It seems to have worked!  https://test.pypi.org/project/cirq-dev/#history

dabacon:
Next steps:  We should add a warning to the text on the pypi page saying that this is a dev release.  We should switch over to pushing to the non-test version of pypi.

PR #1956, opened 2019-08-15
Add warning to readme for dev version on pypi

dabacon:


dabacon:
Want to verify this first before moving over to prod pypi.

setup.py
```diff
@@ -22,18 +22,22 @@
...
 
+# If CIRQ_DEV_VERSION is set then we use cirq-dev as the name of the package
+# and update the version to this value.
+if 'CIRQ_DEV_VERSION' in os.environ:
+    name = 'cirq-dev'
+    __version__ = os.environ['CIRQ_DEV_VERSION']
+    long_description = (
+        "**This is a development version of Cirq and may be "
+        "unstable.**\n\n**For the latest stable release of Cirq "
+        "see ** `here <https://pypi.org/project/cirq>`__.\n" + long_description)
```
cduck: I'm not familiar with reStructuredText but should this be a single underscore?
```suggestion
        "see ** `here <https://pypi.org/project/cirq>`_.\n\n" + long_description)
```

dabacon: Double underscore makes it an anonymous hyperlink, which is what we want.  With the single underscore you can reference it.  https://stackoverflow.com/questions/27420317/restructured-text-rst-http-link

cduck:
HTTPError: 400 Client Error: The description failed to render in the default format of reStructuredText. See https://test.pypi.org/help/#description-content-type for more information. for url: https://test.pypi.org/legacy/

setup.py
cduck: Apparently the extra space matters.  Also you needed an extra newline.

PR #1959, opened 2019-08-15
Add pandoc to dev apt-list

### LINKED ISSUE #1920: Pandoc missing
When trying to build the docs on a fresh linux machine I got an error message saying that pandoc (the command line tool) was missing. Perhaps add this to the dev requirements [dev_tools/conf/apt-list-dev-tools.txt](https://github.com/quantumlib/Cirq/blob/master/dev_tools/conf/apt-list-dev-tools.txt)?

dabacon:
Fixes #1920 

Tested that removing pandoc mails buid-docs.sh fail and that installing after made it succeed.


dabacon:
Think we need to clean up the warnings.  I haven't looked at that in a long time.

PR #1960, opened 2019-08-15
Cross entropy benchmarking

XiaoMiQC:


=2.2.3
```diff
@@ -0,0 +1,11 @@
+Collecting matplotlib
```
vtomole: Accidental commit. Please remove.

XiaoMiQC: Done.

cirq/experiments/__init__.py
```diff
@@ -11,3 +11,8 @@
     two_qubit_randomized_benchmarking,
     two_qubit_state_tomography,
 )
+
+from cirq.experiments.cross_entropy_benchmarking import (
+    cross_entropy_benchmarking,
```
vtomole: Alphabetical: `build_entangling_layers` should come first.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -0,0 +1,377 @@
...
+    @property
+    def data(self) -> Sequence[CrossEntropyPair]:
+        """Returns a sequence of CrossEntropyPairs.
+
+        Each CrossEntropyPair is a NamedTuple that contains a cycle number and
+        the corresponding XEB fidelity.
+        """
+        return self._data
+
+    # coverage: ignore
```
vtomole: Why the coverage ignore?

examples/cross_entropy_benchmarking_example.py
```diff
@@ -0,0 +1,38 @@
+# coverage: ignore
```
vtomole: Why are we ignoring coverage here?

vtomole: Add more documentation like our other examples. A reference to the paper where this comes from e.t.c. https://github.com/quantumlib/Cirq/blob/master/examples/bcs_mean_field.py is a good example.

dabacon:
@vtomole this was a borked merge that had been previously reviewed. @XiaoMiQC has to move on to other things, so I'm going to merge this and then open a pr to address your comments.

Issue #1819, opened 2019-07-16
Nightly dev builds


dabacon:
Releasing dev builds to pypi is currently done manually and in an ad hoc manner.  We could move to nightly builds.

Strilanc:
Sounds reasonable. Any suggestions on what would actually be producing the builds? Some kind of cron job running on Travis?

balopat:
We could use Cloud Build on GCP as well.

dabacon:
I'm looking into this.

Current idea is that we will do dev builds to a separate cirq-dev pypi package on every commit to master.  We can use travis's deploy functionality to do this I think.

dabacon:
We now have builds that deploy on every successful new merge.  These are pushed to "cirq-dev" which is owned by "quantumlibdev".

vtomole:
YAY!!

vtomole:
Look for a new way of doing this.

dabacon:
Think we can close this again since we are back doing this?  Or was there some other way we wanted to look to do this?

vtomole:
I think it's good! :+1:

PR #1961, opened 2019-08-15
Start using prod pypi for dev releases

dabacon:
Should finish off #1819


dabacon:
Tought about this a bit, and decided I don't like the security of this.  So I changed password of quantumlib on pypi and we will use a separate account quantumlibdev for this release.

PR #1963, opened 2019-08-16
Add parameter expression flattening

### LINKED ISSUE #1930: Unwinding Symbolic formulas and parameters?

cduck:
Fixes #1930.

Adds methods that `flatten`, `flatten_with_sweep`, and `flatten_with_params` to flatten parameterized circuits and transform sweeps to match.


cduck:
Pylint didn't catch that error locally for some reason.

cirq/study/flattener.py
```diff
@@ -0,0 +1,275 @@
...
+            value: The sympy.Symbol, sympy expression, name, or float to resolve
+                to a unique symbol or float.
+
+        Returns:
+            The unique symbol or value of the parameter as resolved by this
+            resolver.
+        """
+        if isinstance(value, (int, float)):
+            return value
+        if isinstance(value, str):
```
dabacon: this one isn't documented.

cirq/study/flattener.py
```diff
@@ -0,0 +1,275 @@
...
+            value = sympy.Symbol(value)
+        out = self.param_dict.get(value, None)
+        if out is not None:
+            return out
+        # Create a new symbol
+        symbol = sympy.Symbol(self._next_name())
+        # Ensure the symbol hasn't already been used
```
dabacon: I'd consider moving this into _next_name since then it makes the guarantee of _next_name that it is unique.

cirq/study/flattener.py
```diff
@@ -0,0 +1,275 @@
...
+        """Initializes a new ParamFlattener.
+
+        Args:
+            param_dict: A default initial mapping from some parameter names,
+                symbols, or expressions to other symbols or values.  Only sympy
+                expressions and symbols not specified in `param_dict` will be
+                flattened.
+            new_param_names: An iterator of parameter names to replace each
+                expression with.  It may be infinite but there must be no
+                repetitions.  By default the parameter names are x0, x1, x2, ...
```
dabacon: An alternative default would be to use the str() on the sympy formula.  That would make the mapping that this potentially more readable.

cirq/study/flattener_test.py
```diff
@@ -0,0 +1,168 @@
+# Copyright 2018 The Cirq Developers
```
dabacon: nit: 2019

cduck:
Sympy already handles the equality of formulas.  It seems to canonical expressions as it goes so x+1+1 => x+2.  I couldn't find any counterexamples using arithmetic operations.

ParamFlattener is mutable because of the way it is used to trick objects implementing `_resolve_parameters_` into saying what parameters those are.  This class could be made private and hidden behind helper methods if you're worried.

cirq/study/flattener.py
```diff
@@ -0,0 +1,275 @@
...
+        >>> list(new_sweep)  #doctest: +NORMALIZE_WHITESPACE
+        [cirq.ParamResolver(OrderedDict([('x0', 0.0), ('x1', 1.0)])),
+         cirq.ParamResolver(OrderedDict([('x0', 0.25), ('x1', 0.75)])),
+         cirq.ParamResolver(OrderedDict([('x0', 0.5), ('x1', 0.5)])),
+         cirq.ParamResolver(OrderedDict([('x0', 0.75), ('x1', 0.25)]))]
+
+        >>> print(cirq.resolve_parameters(c_flat, list(new_sweep)[1]))
+        0: ───X^0.25───X^0.75───
+        >>> print(cirq.resolve_parameters(circuit, list(sweep)[1])) # Equivalent
+        0: ───X^0.25───X^0.75───
```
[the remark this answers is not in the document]
cduck: I thought about it but I don't like functions that return more than one thing.  This may be better than using a mutable object though.

cirq/study/flattener.py
```diff
@@ -0,0 +1,275 @@
...
+import itertools
+import sympy
+
+from cirq.study.resolver import ParamResolver, ParamResolverOrSimilarType
+from cirq.study.sweeps import Sweep, Params
+from cirq.study.sweepable import to_sweep
+from cirq.protocols.resolve_parameters import resolve_parameters
+
+
+class ParamFlattener(ParamResolver):
```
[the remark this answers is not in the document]
cduck: It should be fine but this is needed to match the current API.  The only thing in common with `ParamResolver` is the `value_of` method.  It only needs to be able to take the place of the object passed to `_resolve_parameters_`.

cirq/study/flattener.py
```diff
@@ -0,0 +1,275 @@
...
+
+        >>> print(cirq.resolve_parameters(c_flat, list(new_sweep)[1]))
+        0: ───X^0.25───X^0.75───
+        >>> print(cirq.resolve_parameters(circuit, list(sweep)[1])) # Equivalent
+        0: ───X^0.25───X^0.75───
+    """
+
+    def __new__(cls, *args, **kwargs):
+        """Disables the behavior of `ParamResolver.__new__`."""
+        return super().__new__(cls)
```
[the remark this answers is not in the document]
cduck: `ParamResolver` used `__new__` to implement "make sure this resolver-like object is a ParamResolver".  This function should have been moved to a helper called `to_param_resolver`.

cirq/study/flattener.py
cduck: Yes but I'm worried about symbols with names that look like expressions.  `repr(sympy.Symbol('x + 1') + 1) == 'x + 1 + 1'`.  The circuit diagrams would look identical but have different meanings.

cirq/protocols/resolve_parameters.py
```diff
@@ -82,11 +82,11 @@ def resolve_parameters(
         If `val` has no `_resolve_parameters_` method or if it returns
         NotImplemented, `val` itself is returned.
     """
-    if not param_resolver:
+    from cirq import ParamResolver, ParamFlattener  # HACK: break cycle.
```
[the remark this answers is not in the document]
cduck: Done.

cirq/study/flattener.py
```diff
@@ -0,0 +1,275 @@
...
+
+        >>> new_sweep = flattener.transform_sweep(sweep)
+        >>> list(new_sweep)  #doctest: +NORMALIZE_WHITESPACE
+        [cirq.ParamResolver(OrderedDict([('x0', 0.0), ('x1', 1.0)])),
+         cirq.ParamResolver(OrderedDict([('x0', 0.25), ('x1', 0.75)])),
+         cirq.ParamResolver(OrderedDict([('x0', 0.5), ('x1', 0.5)])),
+         cirq.ParamResolver(OrderedDict([('x0', 0.75), ('x1', 0.25)]))]
+
+        >>> print(cirq.resolve_parameters(c_flat, list(new_sweep)[1]))
+        0: ───X^0.25───X^0.75───
```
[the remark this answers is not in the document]
cduck: Done.

cirq/study/flattener.py
```diff
@@ -0,0 +1,275 @@
...
+# limitations under the License.
+"""Resolves symbolic expressions to unique symbols."""
+
+from typing import (overload, Any, Dict, Iterable, Iterator, List, Optional,
+                    Union)
+
+import itertools
+import sympy
+
+from cirq.study.resolver import ParamResolver, ParamResolverOrSimilarType
```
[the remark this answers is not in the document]
cduck: Done.

cirq/study/flattener.py
cduck: Done.

cirq/study/flattener.py
cduck: Done.

cirq/study/flattener_test.py
cduck: Done.

cirq/study/flattener.py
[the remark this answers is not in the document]
cduck: Done.

cirq/study/flattener.py
cduck: Done.  `Used '<{}>'.format(symbol)` to avoid ambiguity.

cduck:
PTAL.  I restructured it so that nothing mutable is public and changed the parameter naming to used `<str(symbol)>`.

cirq/study/flatten_expressions.py
```diff
@@ -0,0 +1,368 @@
...
+        >>> expr_map
+        cirq.ExpressionMap({a/4: <a/4>, 1 - a/2: <1 - a/2>})
+
+        >>> new_sweep = expr_map.transform_sweep(sweep)
+        >>> list(new_sweep)  #doctest: +NORMALIZE_WHITESPACE
+        [cirq.ParamResolver(OrderedDict([('<a/4>', 0.0), ('<1 - a/2>', 1.0)])),
+         cirq.ParamResolver(OrderedDict([('<a/4>', 0.25), ('<1 - a/2>', 0.5)])),
+         cirq.ParamResolver(OrderedDict([('<a/4>', 0.5), ('<1 - a/2>', 0.0)])),
+         cirq.ParamResolver(OrderedDict([('<a/4>', 0.75), ('<1 - a/2>', -0.5)])\
+)]
```
cduck: This line was 1 char over 80.  I'm not sure how else to wrap it so that doctest succeeds.

cirq/study/flatten_expressions.py
```diff
@@ -0,0 +1,368 @@
...
+    type.
+
+    Use `ExpressionMap.transform_sweep` or `ExpressionMap.transform_params`
+    after `flatten`.  Use `flatten_with_sweep` or `flatten_with_params` instead
+    of `flatten` as a shortcut.
+
+    Args:
+        val: The value to copy with substituted parameters.
+
+    Example:
```
dabacon: It seems to have the args and return value separated by the example, is this something being enforced by our doc test?

cirq/study/flatten_expressions.py
```diff
@@ -0,0 +1,368 @@
...
+    expr_map = ExpressionMap(flattener.param_dict)
+    return val_flat, expr_map
+
+
+def flatten_with_sweep(val: Any,
+                       sweep: Union[sweeps.Sweep, List[resolver.ParamResolver]]
+                      ) -> Tuple[Any, sweeps.Sweep]:
+    """Creates a copy of `val` with any symbols or expressions replaced with
+    new symbols.  `val` can be a `Circuit`, `Gate`, `Operation`, or other
+    type.  Also creates a transformed `Sweep` that resolves the new value.
```
dstrain115: Same comment here and below.  I would recommend more descriptive text to this and the following function.

cirq/study/flatten_expressions.py
cduck: No.  The doc test just looks for '>>>' and following lines.  The example should come before Args though.

cirq/study/flatten_expressions.py
```diff
@@ -0,0 +1,368 @@
...
+
+        Args:
+            param_dict: A default initial mapping from some parameter names,
+                symbols, or expressions to other symbols or values.  Only sympy
+                expressions and symbols not specified in `param_dict` will be
+                flattened.
+            get_param_name: A callback function that returns a new parameter
+                name for a given sympy expression or symbol.  If this function
+                returns the same value for two different expressions, `'_#'` is
+                appended to the name to avoid name collision wher `#` is the
```
dstrain115: wher => where

cirq/study/flatten_expressions.py
```diff
@@ -0,0 +1,368 @@
...
+            symbol = sympy.Symbol('{}_{}'.format(name, collision))
+        return symbol
+
+    def value_of(self, value: Union[sympy.Basic, float, str]
+                ) -> Union[sympy.Basic, float]:
+        """Resolves a symbol or expression to a new symbol unique to that value.
+
+        If value is a float, returns it.
+        If value is a str, treat it as a symbol with that name and continue.
+        If this `_ParamFlattener` was initialized with a `param_dict` and
```
dstrain115: I would get rid of the "was initialized with a param_dict", since this returns param_dict[value] if it has been added by a previous call.

cirq/study/flatten_expressions.py
```diff
@@ -0,0 +1,368 @@
...
+        `transform_params` after `flatten`.
+
+        Args:
+            val: The value to copy with substituted parameters.
+        """
+        return protocols.resolve_parameters(val, self)
+
+
+class ExpressionMap(dict):
+    """A dictionary where the keys are sympy expressions and symbols and the
```
dstrain115: Add comma after the second "and" or split it into two sentences.

cirq/study/flatten_expressions.py
```diff
@@ -0,0 +1,368 @@
...
+    values are sympy symbols.
+
+    This is returned by `cirq.flatten`.  See `ExpressionMap.transform_sweep` and
+    `ExpressionMap.transform_params`.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """Initialized the `ExpressionMap`.  The arguments are the same as the
```
dstrain115: Initialized => Initializes

cirq/study/flatten_expressions_test.py
```diff
@@ -0,0 +1,182 @@
...
+    assert flattener.value_of(sympy.Symbol('c')) == 5
+    # Twice
+    assert (flattener.value_of(sympy.Symbol('c') / 2 +
+                               1) == sympy.Symbol('<c/2 + 1>'))
+    assert (flattener.value_of(sympy.Symbol('c') / 2 +
+                               1) == sympy.Symbol('<c/2 + 1>'))
+    # Collisions
+    assert (flattener.value_of(
+        sympy.Symbol('c') /
+        sympy.Symbol('2 + 1')) == sympy.Symbol('<c/2 + 1>_1'))
```
dstrain115: This is a rather weird use case, since c/(2+1) is different than c/2 +1.  I'm not sure about this behavior.

In any case, this seems like something we would not want to explicitly test.

dstrain115: But this test is testing collisions.  It's not clear that we want to enforce via testing that 'c'/'2+1' should collide with 'c'/2+1.  The two seem like different expressions to me.

I would actually qualify this as surprising behavior, since it is really a collision of the symbol of '2+1' with 2+1

cduck: This is a sympy issue.  Ideally `str(symbol1) == str(symbol2)` implies `symbol1 == symbol2` and it does if your symbol names aren't unusual.  Would you prefer that we add checks that `symbol == sympy.sympify(symbol.name)` to avoid this?

cirq/study/flatten_expressions.py
```diff
@@ -0,0 +1,368 @@
...
+import sympy
+
+from cirq import protocols, value
+from cirq.study import resolver, sweeps, sweepable
+
+
+def flatten(val: Any) -> Tuple[Any, 'ExpressionMap']:
+    """Creates a copy of `val` with any symbols or expressions replaced with
+    new symbols.  `val` can be a `Circuit`, `Gate`, `Operation`, or other
+    type.
```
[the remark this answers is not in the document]
cduck: Done.

cirq/study/flatten_expressions.py
cduck: Done.

cirq/study/flatten_expressions.py
cduck: Done.

cirq/study/flatten_expressions.py
cduck: Done.

cirq/study/flatten_expressions.py
cduck: Done.

cirq/study/flatten_expressions.py
cduck: Done.

cirq/study/flatten_expressions.py
cduck: Fixed.

cduck:
@dstrain115 PTAL.

Issue #1204, opened 2018-11-29
Fix cirq.contrib.tpu not showing up in documentation


Strilanc:
I have no idea why sphynx is omitting this particular package and not others, but it is. We need to figure out why and fix it.

Strilanc:
Definitely! Go for it!

Ashalynd:
I have figured what was the problem there in my case (python 3.7 on mac os): 

1) cirq.contrib.tpu depends on tensorflow
2) `pip install tensorflow` doesn't work out of the box for python 3.7 (see https://github.com/tensorflow/tensorflow/issues/20444 for more background and workarounds)
3) as stated on [tf page](https://www.tensorflow.org/install/pip), installation with pip is currently only supported for Python 3.4 - 3.6

the following workaround helped in my case:

pip install --upgrade https://storage.googleapis.com/tensorflow/mac/cpu/tensorflow-1.12.0-py3-none-any.whl

(I verified with `pip search tensorflow` that the latest version was 1.12.0)

andbe91:
It looks like tensorflow is now supported by python 3.7. I have verified that it works to build the docs on python 3.7 on Linux after `pip install tensorflow`. Should we change the contrib requirements to tensorflow instead of tf-nightly?

vtomole:
@andbe91 Yes. Please open a pull request doing so.

vtomole:
@dabacon How long does it take for latest: https://cirq.readthedocs.io/en/latest/api.html to update once merged? The problem is still persisting.

andbe91:
Oh, it was actually not working with tf-nighlty then. That probably means that the problem is elsewhere.

andbe91:
I wonder if the issue is that `.readthedocs.yml` reads `requirements_file: dev_tools/conf/pip-list-dev-tools.txt`, i.e. it doesn't even try to install the contrib requirements. However, that should also mean that `requirements.txt` doesn't get installed, and then nothing should work, so I'm not sure what is going on here.

vtomole:
List of builds: https://readthedocs.org/projects/cirq/builds/

andbe91:
Thanks. The logs confirm that the contrib requirements are not installed. I'm no expert on the config of readthedocs, but it looks like there is `.readthedocs.yml` and settings on readthedocs.org itself. Where do we want to configure this?

vtomole:
Are you referring to https://github.com/quantumlib/Cirq/blob/master/.readthedocs.yml?

[...]

PR #1964, opened 2019-08-16
Revert tf_nightly requirement to `tensorflow` because tensorflow now support python 3.7

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/1204


andbe91:
I confirmed that this now builds the documentation successfully on a fresh python 3.7 environment after pip installing requirements.txt, dev_tools/conf/pip-list-dev-tools.txt and cirq/contrib/contrib_requirements.txt.

PR #1969, opened 2019-08-16
Add `install: requirements` section in .readthedocs.yml

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/1204

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,3 +4,7 @@ build:
 python:
   version: 3.6
   pip_install: true
+
+  install:
+      - requirements: contrib/contrib-requirements.txt
+      - requirements: requirements.txt


vtomole:
@andbe91 I got this suggestion from: https://docs.readthedocs.io/en/stable/config-file/v2.html#requirements-file

andbe91:
Is it possible to test it before merging?

vtomole:
@andbe91 I'm currently trying to spin up a local server to test this.

.readthedocs.yml
```diff
@@ -2,5 +2,7 @@ requirements_file: dev_tools/conf/pip-list-dev-tools.txt
 build:
   image: latest
 python:
-  version: 3.6
+  version: 3.7
   pip_install: true
+  extras_requirements:
```
andbe91: This should be `extra_requirements`. See [https://docs.readthedocs.io/en/stable/config-file/v1.html#python-extra-requirements](https://docs.readthedocs.io/en/stable/config-file/v1.html#python-extra-requirements)

.readthedocs.yml
```diff
@@ -2,5 +2,7 @@ requirements_file: dev_tools/conf/pip-list-dev-tools.txt
...
+      - contrib
```
andbe91: If this works, you can also add `dev` here instead of specifying it above.

.readthedocs.yml
```diff
@@ -2,5 +2,7 @@ requirements_file: dev_tools/conf/pip-list-dev-tools.txt
 build:
   image: latest
 python:
-  version: 3.6
+  version: 3.7
```
Strilanc: Why the version bump?

.readthedocs.yml
dabacon: dev would be dev-env now

.readthedocs.yml
```diff
@@ -2,5 +2,8 @@ requirements_file: dev_tools/conf/pip-list-dev-tools.txt
 build:
```
andbe91: `requirements_file: dev_tools/conf/pip-list-dev-tools.txt` shouldn't be needed with the addition of `dev_env` to `extra_requirements`

.readthedocs.yml
andbe91: Still 3.7

andbe91:
LGTM if it builds for you locally

PR #1974, opened 2019-08-16
Support symbols in EjectPhasedPaulis optimizer

cduck:
For #1930.  Independent of #1963.

### DIFF
--- a/cirq/optimizers/decompositions_test.py
+++ b/cirq/optimizers/decompositions_test.py
@@ -49,7 +49,10 @@ def test_is_negligible_turn():
     assert not cirq.is_negligible_turn(-0.5, 1e-5)
     assert not cirq.is_negligible_turn(0.5, 1e-5)
     assert not cirq.is_negligible_turn(4.5, 1e-5)
+    # Variable sympy expression
     assert not cirq.is_negligible_turn(sympy.Symbol('a'), 1e-5)
+    assert not cirq.is_negligible_turn(sympy.Symbol('a') + 1, 1e-5)
+    assert not cirq.is_negligible_turn(sympy.Symbol('a') * 1e-10, 1e-5)
     # Sympy expression for a constant
     assert cirq.is_negligible_turn(sympy.Symbol('a') * 0 + 3 + 1e-6, 1e-5)
     assert not cirq.is_negligible_turn(sympy.Symbol('a') * 0 + 1.5 - 1e-6, 1e-5)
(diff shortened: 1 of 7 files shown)


cirq/optimizers/decompositions_test.py
```diff
@@ -31,6 +32,29 @@ def assert_gates_implement_unitary(gates: Sequence[cirq.SingleQubitGate],
...
+    assert cirq.is_negligible_turn(3 + 1e-6, 1e-5)
+    assert not cirq.is_negligible_turn(1e-4, 1e-5)
+    assert not cirq.is_negligible_turn(-1e-4, 1e-5)
+    assert not cirq.is_negligible_turn(0.5, 1e-5)
+    assert not cirq.is_negligible_turn(-0.5, 1e-5)
+    assert not cirq.is_negligible_turn(0.5, 1e-5)
+    assert not cirq.is_negligible_turn(4.5, 1e-5)
+    assert not cirq.is_negligible_turn(sympy.Symbol('a'), 1e-5)
+    # Sympy expression for a constant
+    assert cirq.is_negligible_turn(sympy.Symbol('a') * 0 + 3 + 1e-6, 1e-5)
```
dabacon: it feels strange to me that this is true, since 'a' can take on some value which is not near an integer.

cirq/optimizers/eject_phased_paulis.py
```diff
@@ -25,10 +25,13 @@
     from typing import Dict, List
 
 
+TVal = Union[float, sympy.Basic]
```
dabacon: we use this type quite a bit, consider moving it somewhere common, maybe cirq.value

cirq/optimizers/decompositions_test.py
cduck: a is multiplied by zero so it doesn't matter.

dabacon: could you add the case where a symbol appears in a formula, not just alone 
`assert not cirq.is_negligible_turn(sympy.Symbol('a') + 1 - 1e-6, 1e-5)`

cduck: Done.

cirq/optimizers/eject_phased_paulis.py
cduck: I don't want to do that in this PR.

dabacon: SG

Issue #1901, opened 2019-07-30
Expectation protocol


kevinsung:
PauliSum, LinearCombinationOfGates, and Circuit could all implement a protocol
```
def _expectation_(self, state: Union[int, np.ndarray]) -> complex
```

peterse:
Do you have in mind what the function that calls this protocol look like? 

I was thinking of a single `expectation` protocol that would dispatch to different `_expectation_*`  methods defined for wavefunctions, density matrices, or bitstring arrays. Then, for example,  `expectation(cirq.X(cirq.LineQubit(0)), wf)`, would end up using a method like `PauliString._expectation_from_wavefunction`.

Right now, if we were to put this method into `PauliString` it would just be a thin wrapper for `cirq.pauli_string_expectation(self).value_derived_from_wavefunction`. However, it feels like the methods of `display.ApproxPauliStringExpectation` and `display.PauliStringExpectation` should live inside `PauliString`.

peterse:
@Strilanc @mpharrigan does this seem like a reasonable protocol?

mpharrigan:
xref #1805 

@peterse : this seems like a weird form of dispatch. Why not `wf.expectation()`? 

@kevinsung : What is state? How would the protocol work on `Circuit`s? Circuits are unitary but generally not Hermitian

peterse:
@mpharrigan I don't think there's a dedicated class for wavefunctions/dm's at this point.

Another alternative is `PauliString.expectation( ... )` though the natural choice of defining `PaulliSum.expectation` in terms of `PauliString.expectation` would lead to a lot of redundant input validating/dispatch.

kevinsung:
@mpharrigan `state` would be the wavefunction. Now that you mention it, in fact none of the objects I mentioned (PauliSum, LinearCombinationOfGates, and Circuit) are guaranteed to be Hermitian. I was thinking that the expectation for any object A would nevertheless be defined as ⟨Ѱ|A|Ѱ〉, but I guess this might not be conventional?

peterse:
resolved with #1975

Issue #1921, opened 2019-08-07
Proposal: Remove mixture in favor of mixture_channel


c-poole:
If `cirq.mixture` is allowed to return mixtures of various types of concepts, then any code calling this method needs to sort through what it received before knowing how to use it. Rather than having one `mixture` protocol designed to handle the general case of "probabilistic mixtures of foo", we should have independent protocols for each "foo". We only currently use `mixture` to describe a mixture of unitaries, and as such the current functionality of `mixture` should be handed over to `mixture_channel`. If we wanted to represent other probabilistic mixtures such as mixed states, we could do that with a separate unambiguous protocol such as `mixture_of_states` or the like. This impacts #1722 and should be decided on before that PR is merged.

Strilanc:
I think that `mixture_channel` may currently return the mixture *wrapped to look like a channel* (i.e. a combination of matrices A_k such that sum A_k * A_k^-1 == Identity). This is a distinct concept from "tell me the mixture of this object", because it is more of a conversion instead of a getter.

I agree that we should refactor to have one type of mixture, and that this single type of mixture is a probabilistic combination of unitaries, and that this means `mixture` and `mixture_channel` need to be tweaked to make this all more clear.

Strilanc:
This is obsoleted by `mixture_channel` being deprecated and `mixture` taking up its functionality.

PR #1975, opened 2019-08-18
first pass at an `expectation` function

### LINKED ISSUE #1805: Move Pauli string expectation value calculation into PauliString then delete the associated display

peterse:
My thoughts for how to implement #1901 

Not sure if supporting general mixtures in this method is appropriate, but it could be done easily.

I went light on the `PauliString.expectation` tests since refactoring `display.py` (see #1805) will shift all the existing expectation methods in `display_test.py` into `pauli_string_test.py` . I can also resolve #1805 in this PR if thats not too much to review.


cirq/protocols/.gitignore
```diff
@@ -0,0 +1,2 @@
+*params.npy
+*bias.npy
```
mpharrigan: what are these?

cirq/ops/linear_combinations.py
```diff
@@ -315,6 +316,67 @@ def copy(self) -> 'PauliSum':
         factory = type(self)
         return factory(self._linear_dict.copy())
 
+    def expectation(self,
```
mpharrigan: I think this should just delegate to `sum(p.expectation(state) for p in self)`

cirq/protocols/.gitignore
peterse: no idea how that got there; removed.

peterse:
> I think it would make sense to fold in the resolution of #1805 here

I've gone ahead and removed the `PauliStringExpectation` display and associated tests and usages.

peterse:
> I think this should just delegate to `sum(p.expectation(state) for p in self)`

Now that I've fixed the state input validation, calls to `PauliString.expectation` will need to check if `state` is a valid wavefunction or density matrix. This validation will add up if the user calls `PauliSum.expectation` on a pauli sum with many terms.

On a related note, I want to use `to_valid_state_vector` and `to_valid_density_matrix` for input state validation but `cirq.ops` can't import from `cirq.sim` due to circular dependency. Is there any specific reason `wave_function.py` and `density_matrix_utils.py` couldn't be moved to `linalg`? @Strilanc

mpharrigan:
IMO, I'd expect to find `wave_function.py` and `density_matrix_utils.py` with simulator stuff. I see the argument for linalg, but basically everything in quantum computing is linear algebra so it makes sense to have the distinction between low-level, utility linalg routines and simulation-related linear algebra functions. Maybe just those validation functions can be lifted out into `linalg`? 

Alternatively, you can break the circular import by doing a local import within the function

cirq/ops/linear_combinations.py
```diff
@@ -315,6 +318,63 @@ def copy(self) -> 'PauliSum':
...
+            NotImplementedError if any term in this PauliSum is non-Hermitian.
+            ValueError if the input is a state with a size that is not a power
+            of 2, or a shape that is neither `(2 ** n,)` or `(2 ** n, 2 ** n)`.
+        """
+        if any([abs(p.coefficient.imag) > 0.0001 for p in self]):
+            raise NotImplementedError(
+                "Cannot compute expectation value of a non-Hermitian "
+                "PauliString <{}>. Coefficient must be real.".format(self))
+
+        # TODO: add support for mixtures.
```
mpharrigan: do, or open an issue and link here

peterse: From discussion at the 8/21 sync and #1921 mixtures of states won't exist in cirq for the time being so I've removed this TODO

cirq/ops/linear_combinations.py
```diff
@@ -315,6 +316,71 @@ def copy(self) -> 'PauliSum':
...
+            ValueError if the shape of `state` is neither `(2 ** n,)`
+                or `(2 ** n, 2 ** n)`.
+        """
+        if any([abs(p.coefficient.imag) > 0.0001 for p in self]):
+            raise NotImplementedError(
+                "Cannot compute expectation value of a non-Hermitian "
+                "PauliString <{}>. Coefficient must be real.".format(self))
+
+        if qubit_map is None:
+            qubit_map = {q: i for i, q in enumerate(self.qubits)}
```
Strilanc: Inferring this value may cause user error. You run a simulation, take the output, pass it into the expectation... oops, the pauli sum had only a subset of the qubits.

At the very least this should be limited to the case where the number of qubits in the state is equal to the number of qubits in the operator.

cirq/ops/linear_combinations.py
```diff
@@ -315,6 +316,71 @@ def copy(self) -> 'PauliSum':
...
+        else:
+            if not isinstance(qubit_map, dict) or not all(
+                isinstance(k, raw_types.Qid) and isinstance(v, int) for k, v in qubit_map.items()):
+                raise TypeError("Input qubit map must be a valid mapping from "
+                                "Qubit ID's to integer indices.")
+
+        # `state` must support a map over all qubits in this PauliSum.
+        size = state.size
+        if state.shape == (size,):
```
Strilanc: `if len(state.shape) == 1`

cirq/ops/linear_combinations.py
```diff
@@ -315,6 +316,71 @@ def copy(self) -> 'PauliSum':
...
+            num_qubits = size.bit_length() - 1
+            if len(set(qubit_map.keys())) > num_qubits:
```
Strilanc: Having additional qubits in the state should be allowed. Expectation operators don't have to cover the entire state.

cirq/ops/linear_combinations.py
```diff
@@ -315,6 +316,71 @@ def copy(self) -> 'PauliSum':
         factory = type(self)
         return factory(self._linear_dict.copy())
 
+    def expectation(self,
```
Strilanc: I think the user should be required to declare the kind of state they are passing in, instead of relying on the method to infer it from the shape. Otherwise state vector tensors and density matrix tensors will be indistinguishable.

Split the expectation method into `state_vector_expectation` and `density_matrix_expectation` or similar.

peterse:
Changes based on reviews:
 - PauliString and PauliSum each have  `expectation_from_density_matrix`  and `expectation_from_wavefunction` methods. These accept either common shape for the input states (e.g.  `(2 ** n, )` or `(2, 2, ..., 2)` for wavefunctions).
 - `qubit_map` mapping pauli qubits to state indices is now a required arg with no default. I'll do a followup PR exploring kwarg options.

cirq/ops/linear_combinations.py
```diff
@@ -315,6 +316,90 @@ def copy(self) -> 'PauliSum':
         factory = type(self)
         return factory(self._linear_dict.copy())
 
+    def expectation_from_wavefunction(
+            self,
+            state: np.ndarray,
+            qubit_map: Dict[raw_types.Qid, int]) -> float:
+        """Evaluate the expectation of this PauliSum given a wavefunction.
+
+        See `PauliString.expectation_from_wavefunction`.
```
Strilanc: Copy over details about parameters and return type.

cirq/ops/linear_combinations.py
```diff
@@ -315,6 +316,90 @@ def copy(self) -> 'PauliSum':
...
+        """
+        if any([abs(p.coefficient.imag) > 0.0001 for p in self]):
```
Strilanc: Nit: redundant `[` and `]`

cirq/ops/linear_combinations.py
```diff
@@ -315,6 +316,90 @@ def copy(self) -> 'PauliSum':
...
+            raise NotImplementedError(
+                "Cannot compute expectation value of a non-Hermitian "
```
Strilanc: Is this actually the case? I'd expect the expectation to be just be scaled by the complex value.

It also seems like this would be handled fine from within PaulSitring.

@dabacon opinion on this? Is the concept of the expectation well defined when there's imaginary terms? Is it better to have guard rails?

cirq/ops/linear_combinations.py
```diff
@@ -315,6 +316,90 @@ def copy(self) -> 'PauliSum':
...
+                "PauliString <{}>. Coefficient must be real.".format(self))
+
+        if not isinstance(qubit_map, dict) or not all(
```
Strilanc: Don't require that it be a `dict` (e.g. a `defaultdict` would be fine). I think `isinstance(x, Mapping)` from `typing` works more generally.

cirq/ops/linear_combinations.py
```diff
@@ -315,6 +316,90 @@ def copy(self) -> 'PauliSum':
...
+            isinstance(k, raw_types.Qid) and isinstance(v, int) for k, v in qubit_map.items()):
+            raise TypeError("Input qubit map must be a valid mapping from "
+                            "Qubit ID's to integer indices.")
+
+        if set(qubit_map.keys()) < set(self.qubits):
```
Strilanc: You want `not set(self.qubits) >= qubit_map.keys()`. For example, suppose the sets are disjoint.

Add tests exercising this case for A>B, A<B, and mutual-subset-but-different-extra-items.

cirq/ops/linear_combinations.py
```diff
@@ -315,6 +316,90 @@ def copy(self) -> 'PauliSum':
...
+            raise TypeError("Input state dtype must be np.complex64 or "
+                            "np.complex128")
+
+        size = state.size
+        num_qubits = size.bit_length() - 1
+        if len(state.shape) == 1 or state.shape == (2,) * num_qubits:
+            # HACK: avoid circular import
+            from cirq.sim.wave_function import validate_normalized_state
+            validate_normalized_state(state, num_qubits, dtype=state.dtype)
+            return sum([p._expectation_from_wavefunction(state, qubit_map) for p in self])
```
Strilanc: Nit: unneeded square brackets.

cirq/ops/linear_combinations.py
```diff
@@ -315,6 +316,90 @@ def copy(self) -> 'PauliSum':
...
+
+        # FIXME: Avoid enforce specific complex type. This is necessary to
+        # prevent an `apply_unitary` bug (Issue #2041).
+        if state.dtype.kind != 'c':
+            raise TypeError("Input state dtype must be np.complex64 or "
+                            "np.complex128")
+
+        size = state.size
+        num_qubits = size.bit_length() - 1
+        if len(state.shape) == 1 or state.shape == (2,) * num_qubits:
```
Strilanc: Invert if statement to "if bad then throw" guard statement style.

cirq/ops/linear_combinations.py
```diff
@@ -315,6 +316,90 @@ def copy(self) -> 'PauliSum':
...
+
+    def expectation_from_density_matrix(self,
+                    state: np.ndarray,
+                    qubit_map: Optional[Dict[raw_types.Qid, int]]=None
+                    ) -> float:
+        """Evaluate the expectation of this PauliSum given a density matrix.
+
+        See `PauliString.expectation_from_density_matrix`.
+        """
+        if any([abs(p.coefficient.imag) > 0.0001 for p in self]):
```
Strilanc: Square brackets

cirq/ops/linear_combinations.py
```diff
@@ -315,6 +316,90 @@ def copy(self) -> 'PauliSum':
...
+
+        size = state.size
+        num_qubits = int(np.sqrt(size)).bit_length() - 1
+        dim = int(np.sqrt(size))
+        if state.shape == (dim, dim) or state.shape == (2, 2) * num_qubits:
+            # HACK: avoid circular import
+            from cirq.sim.density_matrix_utils import to_valid_density_matrix
+            # Do not enforce reshaping if the state all axes are dimension 2.
+            _ = to_valid_density_matrix(state.reshape(dim, dim), num_qubits, dtype=state.dtype)
+            return sum([p._expectation_from_density_matrix(state, qubit_map) for p in self])
```
Strilanc: Square brackets

cirq/ops/linear_combinations.py
```diff
@@ -315,6 +316,90 @@ def copy(self) -> 'PauliSum':
...
+            raise NotImplementedError(
+                "Cannot compute expectation value of a non-Hermitian "
+                "PauliString <{}>. Coefficient must be real.".format(self))
+
+        if not isinstance(qubit_map, dict) or not all(
+            isinstance(k, raw_types.Qid) and isinstance(v, int) for k, v in qubit_map.items()):
+            raise TypeError("Input qubit map must be a valid mapping from "
+                            "Qubit ID's to integer indices.")
+
+        if set(qubit_map.keys()) < set(self.qubits):
```
Strilanc: The body of this method looks *extremely* similar to the wavefunction one. Can you extra a common private method that is parameterized by an argument of `sub=PauliString.expectation_from_density_matrix` or `sub=PauliString.expectation_from_wavefunction`?

cirq/ops/pauli_string.py
```diff
@@ -240,6 +239,189 @@ def _apply_unitary_(self, args: protocols.ApplyUnitaryArgs):
...
+        if len(state.shape) == 1 or state.shape == (2,) * num_qubits:
+            # HACK: avoid circular import
+            from cirq.sim.wave_function import validate_normalized_state
+            validate_normalized_state(state, num_qubits, dtype=state.dtype)
+            return self._expectation_from_wavefunction(state, qubit_map)
+
+        raise ValueError("Input array does not represent a wavefunction with "
+                         "shape `(2 ** n,)` or `(2, 2, ..., 2)`.")
+
+    def _expectation_from_wavefunction(self,
```
Strilanc: Give this a better name, e.g. add `_without_validation`.

cirq/ops/pauli_string.py
```diff
@@ -240,6 +239,189 @@ def _apply_unitary_(self, args: protocols.ApplyUnitaryArgs):
...
+            qubit_map: A map from all qubits used in this PauliString to the
+            indices of the qubits that `state` is defined over.
+
+        Returns:
+            The expectation value of the input state.
+
+        Raises:
+            NotImplementedError if this PauliString is non-Hermitian.
+        """
+        if abs(self.coefficient.imag) > 0.0001:
```
Strilanc: Most of this validation is common to all four of these expectation methods. It could be extracted into a utility method `_validate_qubit_mapping`.

You could also check that the integers being provided make sense, given the size of the state vector or density matrix.

cirq/ops/linear_combinations.py
peterse: Yea I agree that non-Hermitian operators should work fine if allowed, so the error message isn't quite correct.

My main arguments for disallowing non-Hermitian operators is that the usual definition for variance of a Hermitian operator B (Var[B] = <B**2> - <B>**2) breaks down. For example, the operator B=iX has Var[B] = i for the state |0> using that standard formula.

The accepted answer at his [link ](https://physics.stackexchange.com/questions/82613/why-are-only-real-things-measurable) shows that non-Hermitian operator expectations don't add any outcomes that couldn't have been accomplished with Hermitain operators, though that doesn't mean they couldn't be convenient for some purpose.

cirq/ops/linear_combinations.py
peterse: the issue with that is that Paulisum is making calls directly to the PauliString methods that have no input validation so that `state` only ever gets validated once (in the `PauliSum.expectation_from_*` method). So the `sum(p._expectation_from_*(state, qubit_map) for p in self)` return value has to be preceded by state-specific input validation that would be nastier to parametrize over.

The best that a common method could do (after adding a `_validate_qubit_mapping` function) is to provide Hermiticity and dtype validation which isn't worth an additional function def imo

cirq/ops/pauli_string.py
peterse: done

cirq/ops/linear_combinations.py
peterse: done.

cirq/ops/pauli_string.py
```diff
@@ -240,6 +239,187 @@ def _apply_unitary_(self, args: protocols.ApplyUnitaryArgs):
...
+            num_qubits = state.shape[0].bit_length() - 1
+            state = np.reshape(state, (2,) * num_qubits)
+
+        ket = np.copy(state)
+        for qubit, pauli in self.items():
+            buffer = np.empty(ket.shape, dtype=state.dtype)
+            args = protocols.ApplyUnitaryArgs(target_tensor=ket,
+                                              available_buffer=buffer,
+                                              axes=(qubit_map[qubit],))
+            ket = protocols.apply_unitary(pauli, args)
```
trevormccrt: I like how you did this - O(n) expectation values without writing any new code.

trevormccrt: what if we want to calculate the expectation value of something hermetian but non-unitary (For example a projector on the zero state of a qubit)?

peterse: Credit for that code actually goes to Matt.

Non-unitary is fine (PauliSums will be non-unitary on general)

trevormccrt: i suppose that can be expressed as (I-Z)/2. I guess anything expressible as a pauli_string will have an expectation value calculable via unitary evolution - correct me if i'm wrong though.

peterse: yep. For these methods, the expectation is purely postprocessing on a circuit output.

trevormccrt:
what's the timeline on getting this in? I'd like to be able to use it in code for another google product i'm working on.

examples/hhl.py
```diff
@@ -257,18 +257,22 @@ def hhl_circuit(A, C, t, register_size, *input_prep_gates):
...
             cirq.PauliString({ancilla: cirq.Z}),
+            num_samples=5000,
             key='a'
         ),
         cirq.pauli_string_expectation(
             cirq.PauliString({memory: cirq.X}),
+            num_samples=5000,
             key='x'
         ),
         cirq.pauli_string_expectation(
```
trevormccrt: i think you forgot to update this call?

cirq/sim/density_matrix_simulator_test.py
```diff
@@ -586,17 +528,17 @@ def test_compute_samples_displays(dtype):
         cirq.H(b),
         cirq.X(c),
         cirq.H(c),
-        cirq.pauli_string_expectation(
+        cirq.approx_pauli_string_expectation(
```
trevormccrt: this won't pass (missing num_samples arg)

examples/hhl.py
peterse: done

cirq/sim/density_matrix_simulator_test.py
peterse: fixed

cirq/ops/linear_combinations.py
kevinsung: By the way, it can have imaginary coefficients and still be Hermitian (if the Hermitian-conjugated term is also in there). I lean toward handling the non-Hermitian case but also want to see if @dabacon has an opinion.

trevormccrt:
@peterse you want to take care of the broken tests?

cirq/ops/linear_combinations.py
dabacon: Making it just Hermitian make sense when thinking of this as "expectation value of an observable in quantum theory".  For non-Hermitian operators, sometimes you can think of them as "observables" with complex eignevalues.  But sometimes you end up with non-Hermitian operators that don't a complete set of eigenvectors than span the space, so then it is really hard to interpret this as an observable. 

That said, it definitely makes sense to define <x|A|x> for a non-hermitian operator.  My suggestion would be to allow this via an explicit parameter.  i.e. normally it requires Hermitian and throws an error that says if you want to calculate for a non-hermitian operator, set the parameter appropriate.

PR #1982, opened 2019-08-19
Add EngineProgram to QuantumEngineSampler

dabacon:
Note that I chose not to validate that the `EngineProgram` was created from the same `Engine` as the sampler.  We could add this.

Alternatively we could add another type of `Sampler`.

### DIFF
--- a/cirq/google/engine/engine_sampler.py
+++ b/cirq/google/engine/engine_sampler.py
@@ -23,7 +23,7 @@ if TYPE_CHECKING:
 
 
 class QuantumEngineSampler(work.Sampler):
-    """A sampler that samples from processor managed by the Quantum Engine.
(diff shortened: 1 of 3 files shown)


dabacon:
Fixes #1982

cirq/google/engine/engine_sampler.py
```diff
@@ -12,18 +12,21 @@
...
+from cirq.google import engine
 
 if TYPE_CHECKING:
     import cirq
 
 
+
 class QuantumEngineSampler(work.Sampler):
-    """A sampler that samples from real hardware on the quantum engine.
+    """A sampler that samples from processor managed by the Quantum Engine.
```
wcourtney: Nit: either make 'processor' plural or add an article.

dabacon: fixed

PR #1988, opened 2019-08-20
Make engine calibration a dict and add heatmap

dabacon:
Also move over to pattern of importing module, not classes.


cirq/google/engine/engine.py
```diff
@@ -694,8 +700,25 @@ def get_calibration(self, calibration_name: str) -> 'Calibration':
...
+    to the metrics values.
+
+    Calibrations act just like a python dictionary. For example you can get
+    a list of all of the metric names using
+
+        `calibration.keys()`
+
+    and query a single value by looking up the name by index:
+
+        `calibration['t1'`
```
wcourtney: Add closing bracket to example.

cirq/google/engine/engine.py
```diff
@@ -704,35 +727,65 @@ class Calibration:
...
+        self._metric_dict = self._compute_metric_dict(calibration['metrics'])
+
+    def _compute_metric_dict(
+            self, metrics: Dict
+    ) -> Dict[str, Dict[Tuple[devices.GridQubit, ...], Any]]:
+        results: Dict[str, Dict[Tuple[devices.
+                                      GridQubit, ...], Any]] = defaultdict(dict)
+        for metric in metrics:
+            name = metric['name']
+            # Flatten the values a list, removing keys containing type names
```
wcourtney: "Flatten the values into a list"?

cirq/google/engine/engine.py
```diff
@@ -704,35 +727,65 @@ class Calibration:
...
+                results[name][()] = flat_values
+        return results
 
-    def get_metrics_by_name(self, name: str) -> List[Dict]:
-        """ Get a filtered list of metrics matching the provided name.
+    def __getitem__(self, key: str) -> Dict[Tuple[devices.GridQubit, ...], Any]:
+        """Supports getting calibrations by index.
 
-        Values are grouped into a flat list, grouped by target.
+        Calibration may be accessed by key
```
wcourtney: Add punctuation.

cirq/google/engine/engine.py
```diff
@@ -704,35 +727,65 @@ class Calibration:
...
 
-        Params:
-            name: the name of a metric referred to in the calibration. Valid
-            names can be found with the get_metric_names() method.
+            calibration['t1']
```
wcourtney: Surround code with backticks.

cirq/google/engine/engine.py
```diff
@@ -694,8 +700,25 @@ def get_calibration(self, calibration_name: str) -> 'Calibration':
...
 
-class Calibration:
-    """A convenience wrapper for calibrations
+class Calibration(abc.Mapping):
+    """A convenience wrapper for calibrations that acts like a dictionary.
+
+    Calibrations act as dictionaries whose keys are the names of the metric,
+    and whose values are the metric values.  The metric values themselves are
+    represented as a dictionary.  These metric value dictionaries have
+    keys that are tuples of `cirq.GridQubit`s and values that are lists of the
```
wcourtney: What do you think of using a frozenset of GridQubits for the key instead of a tuple since order shouldn't change the match?

cirq/google/engine/engine_test.py
```diff
@@ -148,34 +149,39 @@
         '1562544000021',
         'metrics': [{
             'name': 'xeb',
-            'targets': ['q0_0', 'q0_1'],
+            'targets': ['0_0', '0_1'],
             'values': [{
```
wcourtney: The current format returned today is qX_Y, but it shouldn't matter. If we'd like to abstract/simplify these, should we collapse to a single value, e.g. "['0', '1']"?

cirq/google/engine/engine_test.py
```diff
@@ -148,34 +149,39 @@
...
             }]
         }, {
             'name': 't1',
-            'targets': ['q1_0'],
+            'targets': ['1_0'],
             'values': [{
                 'doubleVal': 505
             }]
+        }, {
+            'name': 'temp',
```
wcourtney: Consider naming this something like "globalMetric" to clarify its usage?

cirq/google/engine/engine.py
dabacon: I'm not sure that the order won't matter.  For example if you have a directional gate (like a cnot that goes from a to b), then you might have metrics about that gate whose order differs (a to b versus b to a).

cirq/google/engine/engine.py
dabacon: done

cirq/google/engine/engine.py
dabacon: done

cirq/google/engine/engine.py
dabacon: done

cirq/google/engine/engine_test.py
dabacon: Oh.. They shouldn't do that.  It should use the canonicalization defined in GridQubit!

cirq/google/engine/engine_test.py
dabacon: done

cirq/google/engine/engine_test.py
dabacon: Added code to handle first and will file bug.

PR #1990, opened 2019-08-20
Add TParamVal alias

### LINKED ISSUE #1987: Add type alias for circuit parameter value Union[float, sympy.Basic]
_Originally posted by @dabacon in https://github.com/quantumlib/Cirq/pull/1974#discussion_r314924054_

Add something like `TParamVal = Union[float, sympy.Basic]` under `cirq.value` and refactor all uses of `Union[float, sympy.Basic]` and `Union[sympy.Basic, float]`.

andbe91:
Closes #1987 

Replaced all `Union[float, sympy.Basic]` and `Union[sympy.Basic, float]` I could find.

### DIFF
--- a/cirq/google/api/v1/programs.py
+++ b/cirq/google/api/v1/programs.py
@@ -13,13 +13,13 @@
 # limitations under the License.
 import json
 from typing import (Any, cast, Dict, Iterable, Optional, Sequence, Tuple,
-                    TYPE_CHECKING, Union)
+                    TYPE_CHECKING)
 import numpy as np
 import sympy
 
 from cirq import devices, ops, protocols
 from cirq.schedules import Schedule, ScheduledOperation
-from cirq.value import Timestamp
+from cirq.value import Timestamp, type_alias
 
 if TYPE_CHECKING:
     from cirq.google import xmon_device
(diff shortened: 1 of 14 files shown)


cirq/google/api/v1/programs.py
```diff
@@ -13,13 +13,13 @@
...
 from typing import (Any, cast, Dict, Iterable, Optional, Sequence, Tuple,
-                    TYPE_CHECKING, Union)
+                    TYPE_CHECKING)
 import numpy as np
 import sympy
 
 from cirq import devices, ops, protocols
 from cirq.schedules import Schedule, ScheduledOperation
-from cirq.value import Timestamp
+from cirq.value import Timestamp, type_alias
```
dabacon: Since you're not touching code in `cirq.value` I think it is preferable to put this up in the cirq imports: `from cirq import devices, ops, protocols, value`.  This also make the name `value.TParamVal` which is slightly nicer that `type_alias.TParamVal`.  

Note we've not been great about enforcing this pattern...even this file has some imports that I don't think should be done this way (all the classes).

andbe91: Done! I changed it for all instances.

Issue #1978, opened 2019-08-18
Delete xmon simulator


dabacon:
It has only given advantage for large qubit number, and only on certain machines.  Also it is tied to a gate set which is a bad coupling.

dabacon:
@Strilanc agree?

Strilanc:
Agree

dabacon:
Long live multiprocessing.

PR #1993, opened 2019-08-20
Delete XmonSimulator

dabacon:
This simulator has only delivered marginal improvement over cirq.Simulator(), and then only on some machines.  

Functionality about only supporting some gates will come with the addition of gateset to device.


dabacon:
Fixes #1978

mpharrigan:
RIP

cirq/devices/gate_set.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class GateSpec():
+
+    def __init__(self, constructor, params):
+        pass
+
+    def canonical
+
+class GateSet():
```
mpharrigan: :eyes:

dabacon: ha yeah bad merge.

PR #2003, opened 2019-08-21
Simplify the type of ValueMap in cirq/vis/heatmap.py.

pingyeh:
The `QubitCorrdinate` type defined 3 lines above works fine.


cirq/vis/heatmap.py
```diff
@@ -41,8 +41,7 @@
 QubitCoordinate = Union[Tuple[int, int], grid_qubit.GridQubit]
 
 # The value map is qubit coordinate -> a type that supports float conversion.
-ValueMap = Union[Dict[Union[Tuple[int, int]], SupportsFloat],
-                 Dict[grid_qubit.GridQubit, SupportsFloat]]
```
dabacon: this fails the mypy check on the code I checked in that depends on heatmap
```
cirq/google/engine/engine.py:791: error: Argument 1 to "Heatmap" has incompatible type "Dict[GridQubit, Any]"; expected "Dict[Union[Tuple[int, int], GridQubit], SupportsFloat]"
The command "check/mypy" exited with 1.
```
I believe the reason is that dict keys are co-variant, as is union.  Union would have to be contravariant for this to work?

pingyeh: `Dict` is invariant per [mypy doc](https://mypy.readthedocs.io/en/latest/generics.html#variance-of-generics). `Union` is covariant. Maybe that's why mypy doesn't allow a `Union` as a key.

I've reverted to the previous form (`Union` on the outside) but removed the `Union` on the `Tuple`. Mypy passed on my desktop.

PR #2005, opened 2019-08-21
Fix the quantum teleportation example

pingyeh:
The current example code shows errors in teleportation like the following output.

```
$ python examples/quantum_teleportation.py 
Circuit:
0: ───X^0.78───Y^0.341───@───H───M───────@───
                         │       │       │
1: ───H────────@─────────X───────M───@───┼───
               │                     │   │
2: ────────────X─────────────────────X───@───

Bloch Sphere of Message After Random X and Y Gates:
x:  -0.0 y:  -0.6368 z:  -0.771

Bloch Sphere of Qubit 2 at Final State:
x:  -0.6763 y:  -0.6368 z:  -0.3703
```

This PR fixes it, and raises a `ValueError` when the final state is not close to the expected state.


pingyeh:
> Create a case where `ValueError` is thrown so that it can be covered by coverage.

I've moved the success check into `example_test.py`. The pytest coverage passes now.

PR #2008, opened 2019-08-22
Update extras_requires to a contrib and a dev_env which includes dev plus contrib

dabacon:
This was confusing because we also have cirq-dev pypi.


dabacon:
How about developer_env?

vtomole:
How about about `python -m pip install -e .[developer_deps,contrib_deps]`? That way we are saying that this installs cirq along it's contrib and dev requirements.

vtomole:
Ah! Let's get more opinions in. @Strilanc @cduck @dstrain115

c-poole:
I like `developer_env` but think we could probably even shorten that to `dev_env`. The only two install options I have desired is a vanilla `pip install cirq` which delivers the package as intended for a user, and an option that one-stop installs everything I need for development work. Asking the user to opt into developer dependencies and contrib dependencies would confuse me if I didn't know what they meant from experience. One sounds like dependencies I need to do development work on cirq and the other sounds like the dependencies I need to contribute to cirq. Adding these options individually isn't much more convenient for a user than having them run pip install on the separate text files as our development instructions currently suggest. Having a `pip install cirq[dev_env]` that folds the three pip instructions into one is all I think we need.

dabacon:
For the developer requirements I think we need to make clear that they are optional (hence "env") because otherwise I think they could be considered linking and we could run into some licensing issues.  I think there might be cases where dev only makes sense...for example in deployment situations where you also want to be able to run tests.

cduck:
@c-poole The purpose of the [contrib] option is so that *users* can easily install and use functionality under `cirq.contrib`.  When a user installs cirq with pip, they don't have easy access to the contrib requirements file.

vtomole:
@c-poole . I agree with `pip install cirq[dev_env]` installing things that are not "Cirq official"; like contrib.

@dabacon 
> For the developer requirements I think we need to make clear that they are optional

Contrib is optional

c-poole:
If we want a [contrib] option for users then I agree that makes sense. What I am more opposed to is having a single [dev] option that only installs the dev_tools requirements and not the contrib requirements. If we want an option that installs both, as well as options that individually install either, I'm game for that.

Strilanc:
I agree with Cody. An important part of a development environment is the ability to run the tests in contrib.

dabacon:
OK moved to `cirq[contrib]` and `cirq[dev_env]`.  The later installs both contrib and dev requirements.

dabacon:
Everyone (or majority) happy?

setup.py
```diff
@@ -63,7 +63,7 @@
       install_requires=requirements,
       extras_require={
           'contrib': contrib_requirements,
-          'dev': dev_requirements,
+          'dev_eng': dev_requirements + contrib_requirements,
```
vtomole:
```suggestion
          'dev_env': dev_requirements + contrib_requirements,
```

PR #2012, opened 2019-08-22
Make simulators fail when all parameters are not resolved

### LINKED ISSUE #1515: Give a descriptive error when asked to simulate a circuit with unresolved parameters
I think the reason https://github.com/quantumlib/Cirq/issues/1514 was reported as a bug is because the current error that results from this mistake is so bad that it looks like a bug in cirq. Both cirq.Simulator and cirq.google.XmonSimulator should check if `cirq.is_parameterized(circuit)` before continuing into the numerical simulation, and if so then raise an exception describing the problem.

This error should raise both when attempting to `run` a parameterized circuit (e.g. you forgot to use `run_sweep`) and when sweeping and you forget a parameter.

dabacon:
Fixes #1515

An alternative is to add flag into resolve protocol which says require everything to be resolved.


vtomole:
Looks like https://github.com/quantumlib/Cirq/pull/1590 addresses the same thing but @Strilanc hasn't gotten back to @davemc84  on it.

dabacon:
Because duplicating something twice is not really a big deal, and this isn't a really useful public class (circuit already supports the method, the duplication is in using the same error message)

dabacon:
#1590 is very brittle and uses circuit info.

PR #2013, opened 2019-08-22
Update development docs for check/all script

dabacon:
Follow up to #

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -154,7 +154,7 @@ A more convenient way to run checks is to via the scripts in the [check/](https:
 ./check/all [BASE_REVISION] [--only-changed-files] [--apply-format-changes]
 ```
 
-The above scripts are convenient and reasonably fast, but they often won't exactly match the results computed by the continuous integration builds run on travis.
+The above scripts are convenient and reasonably fast, but they often won't exactly match the results computed by the continuous integration builds run on T ravis.
 For example, you may be running an older version of `pylint` or `numpy`. If you need to test against the actual continuous integration check, open up a pull request.
 For this pull request you may want to mark it as `[Testing]` so that it is not reviewed.
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -155,7 +155,7 @@ If you want to run all of these tests, you can use the [check/all](https://githu
 ```
 
 This script will run all relevant checks within those clean environments. When using this script, you can
-run a subset of the checks using the ```--only-changed-files``` flag to only pytest against changed files
+run a subset of the checks using the ```--only-changed-files``` flag to only pytest against changed files,
 and the ```--apply-format-changes-only``` flag to fix formatting problems automatically.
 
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -147,16 +147,16 @@ A more convenient way to run checks is to via the scripts in the [check/](https:
...
 For example, you may be running an older version of `pylint` or `numpy`.
-In order to run a check that is significantly more likely to agree with the travis builds, you can use the [continuous-integration/check.sh](https://github.com/quantumlib/Cirq/blob/master/continuous-integration/check.sh) script:
+In order to run a check that is significantly more likely to agree with the travis builds, you can use the [check/all](https://github.com/quantumlib/Cirq/blob/master/check/all) script:
 
 ```bash
-./continuous-integration/check.sh
+./check/all
 ```
 
 This script will create (temporary) virtual environments, do a fresh install of all relevant dependencies and run all relevant checks within those clean environments.
```
vtomole: > This script will create (temporary) virtual environments

Is this still true? I did a quick check of `all` and i don't see anything that may point to new envs being spun up.

dabacon: You are correct.  Updated.

docs/development.md
```diff
@@ -147,16 +147,16 @@ A more convenient way to run checks is to via the scripts in the [check/](https:
 
 The above scripts are convenient and reasonably fast, but they often won't exactly match the results computed by the continuous integration builds run on travis.
```
vtomole: > they often won't exactly match the results computed by the continuous integration builds run on travis.

Sure, but `all` won't either since `all` is just running all the previous scripts so we don't have to run each one by one?

vtomole: I know you didn't touch these, but would be good to fix here.

dabacon: Updated entire section.

docs/development.md
```diff
@@ -143,20 +143,20 @@ A more convenient way to run checks is to via the scripts in the [check/](https:
...
+
+# Run all of the above tests. Which pytest is run is set by the --only-changed-files.
+./check/all [BASE_REVISION] [--only-changed-files] [--apply-format-changes]
 ```
 
-This script will create (temporary) virtual environments, do a fresh install of all relevant dependencies and run all relevant checks within those clean environments.
-Note that creating the virtual environments takes time, and prevents some caching mechanisms from working, so `continuous-integration/check.sh` is significantly slower than the simpler check scripts.
-When using this script, you can run a subset of the checks using the ```--only``` flag.
-This flag value can be `pylint`, `typecheck`, `pytest`, or `incremental-coverage`.
+The above scripts are convenient and reasonably fast, but they often won't exactly match the results computed by the continuous integration builds run on travis.
```
vtomole:
```suggestion
The above scripts are convenient and reasonably fast, but they often won't exactly match the results computed by the continuous integration builds run on T ravis.
```
nit

Issue #159, opened 2018-02-26
make sure random seeding is consistent


dabacon:
in particular we need to make sure random seeding works properly especially for multiprocessing simulations.

zoltanegyed:
Python uses the Mersenne Twister as the core generator. This will give a reproducible sequence given a seed, but it is not sufficient for this purpose for 2 reasons:

1. We need to expose the setting of a seed more clearly for simulations, including possibly passing them on to other processes (not a single seed for all such processes, but the ability to specify them separately).

2. Users may wish to plug in totally different random generation, such as low discrepancy sequences.

dabacon:
Note that right now all our randomness for simulation is coming from numpy.

Agree with 1.

For 2 we should likely add something that does the right delegation to numpy in cirq.  That way we are explicit about this.

dabacon:
Since we deleted the multiprocessing simulator, this now just depends on the numpy random seed.  We should add this to documentation, and then close.

dabacon:
Or better yet allow it to be set in the simulator.

PR #2017, opened 2019-08-23
Add seed to simulators

dabacon:
My only worry is that changing seeds between calls on numpy can bork things.

### DIFF
--- a/cirq/sim/density_matrix_simulator.py
+++ b/cirq/sim/density_matrix_simulator.py
@@ -118,13 +118,17 @@ class DensityMatrixSimulator(simulator.SimulatesSamples,
     def __init__(self,
                  *,
                  dtype: Type[np.number] = np.complex64,
-                 noise: devices.NoiseModel = devices.NO_NOISE):
+                 noise: devices.NoiseModel = devices.NO_NOISE,
+                 seed: Optional[int] = None):
         """Density matrix simulator.
 
          Args:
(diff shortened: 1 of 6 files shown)


vtomole:
What do you mean by `bork`?

dabacon:
https://www.merriam-webster.com/dictionary/bork  "to cause (something, such as an electronic device) to stop working properly"   So was just saying that someone could mess with numpy in between calls to the simulator (especially step) and cause issue (either by calling for more random numbers or less likely setting the seed).

dabacon:
Fixed #159

mpharrigan:
I really like allowing the user to pass in a `np.random.RandomState` object so the user has complete control over the sequence of random numbers.

```python
class Simulator:
  def __init__(self, things, rs=None):
    if not isinstance(rs, np.random.RandomState):
      rs = np.random.RandomState(rs)
    self.rs = rs

  def simulate(self):
    i = self.rs.uniform(size=(10,3))
    ...
```

By default, this will seed according to dev/urandom *or* let you pass an integer seed *or* let you share a `RandomState` instantiation

```python
def main():
  my_rs = np.random.RandomState()
  simulator = Simulator(rs=my_rs)
  measurements = simulator.simulate() #  advances the random state
  my_rs.shuffle(measurements)  # uses the same (advanced) random state
  second_measurements = simulator.simulate() 
```

is completely reproducible

dabacon:
Good suggestion.  Create #2020

PR #2023, opened 2019-08-26
Update publish dev pypi script to use cirq-dev

dabacon:
I'm keeping the set-dev-version script in the case that we can go back to using travis's deploy.

### DIFF
--- a/dev_tools/packaging/generate-dev-version-id.sh
+++ b/dev_tools/packaging/generate-dev-version-id.sh
@@ -15,17 +15,21 @@
 # limitations under the License.
 
 ################################################################################
-# If cirq's version is a dev release, this script can be used to set
-# the CIRQ_DEV_VERSION variable to a string that is the dev version
-# plus an increasing date string.
+# This script prints a new dev version id if the current cirq version
+# is a dev release.
 #
-# This is used by travis during deployment and by published-dev-package.sh.
+# If version in Cirq's _version.py file contains 'dev' this
+# prints a dev version appended by an id given by the time the
+# script was run. This can be used to ensure later releases
+# have increasing numerical release numbers.
 #
-# The variable can be set using
+# This script is used by publish-dev-package and can also be
+# used to set a bash variable for use by travis ci to deploy
+# a new dev version on successful merge.
 #
-#     export CIRQ_DEV_VERSION=`dev_tools/packaging/set-dev-version.sh`
-#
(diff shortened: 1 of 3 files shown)


dabacon:
@Strilanc if you look at this I can do a follow up turning back on dev deploys.

dev_tools/packaging/publish-dev-package.sh
```diff
@@ -105,6 +107,9 @@ cd "$(git rev-parse --show-toplevel)"
 tmp_package_dir=$(mktemp -d "/tmp/publish-dev-package_package.XXXXXXXXXXXXXXXX")
 trap "{ rm -rf ${tmp_package_dir}; }" EXIT
 
+# Configure to push to cirq-dev and not cirq.
+export CIRQ_DEV_VERSION=`dev_tools/packaging/set-dev-version.sh`
```
Strilanc: `export CIRQ_DEV_VERSION=$(dev_tools/packaging/set-dev-version.sh)` to be consistent with the rest of the file

dev_tools/packaging/set-dev-version.sh
```diff
@@ -19,49 +19,32 @@
 # the CIRQ_DEV_VERSION variable to a string that is the dev version
```
Strilanc: Update this description to say that the script prints a new dev version. Also rename the script to `pick-dev-version` or `generate-dev-version-id` or similar to make this clearer.

dev_tools/packaging/publish-dev-package.sh
dabacon: done

dev_tools/packaging/set-dev-version.sh
dabacon: done

dev_tools/packaging/generate-dev-version-id.sh
```diff
@@ -0,0 +1,54 @@
...
+ACTUAL_VERSION=`echo $ACTUAL_VERSION_LINE | cut -d'"' -f 2`
+
+if [[ ${ACTUAL_VERSION} == *"dev" ]]; then
+  echo "${ACTUAL_VERSION}$(date "+%Y%m%d%H%M%S")"
+else
+  echo "Version doesn't end in dev: ${ACTUAL_VERSION_LINE}" >&2
+  exit 1
+fi
+
+exit 0
```
Strilanc: `exit 0` is redundant at the end of a script.

dabacon: done

PR #2037, opened 2019-08-28
Fix circular import issues and document inter-module dependency ordering

### LINKED ISSUE #2035: Should we start using `import cirq` at top level inside cirq?
Now that we are using python 3, importing is much lazier and it's possible for us in many situations to just `import cirq` instead of individual modules.

Pros:

- Much simpler to import functionality. Don't have to remember where things are.
- Internal code will look more like external code.

Cons:

- Further obscures the dependencies between files and modules. If you think the dependency graph is already a mess *now*... well...
- Might complicate unit testing when using mocks. Or might make it easier. Unclear.
- Programmer has to understand when the imported value is used eagerly (e.g. as a base class or unquoted type annotation) vs lazily (e.g. in the body of a method that's not run at import time). 
    - Note that the eager-vs-lazy determination is technically a global property. A method in one module may be invoked during the initialization of another module (e.g. `X = XPowGate()` would invoke `Gate.__init__`).

cduck:
Fixes #2035.

- Require top-level inter-module imports to follow a defined dependency order.  This order is defined at the top of `cirq/__init__.py`.
- Make cirq.protocols lazily loaded so any module can `from cirq import protocols` and protocol code can import any other module at import time.
- Add a test (cirq/import_test.py) to check for circular dependencies.
- Add a tool (dev_tools/run_import_test.py) to debug circular dependencies.


cduck:
@Strilanc PTAL.

cduck:
@Strilanc PTAL.

Strilanc:
I stand corrected about the python3 thing.

cduck:
Changed it to sys.executable.

cirq/__init__.py
```diff
@@ -12,6 +12,50 @@
...
+# Begin dependency order list of sub-modules.
+from cirq import (
+    # Low level
+    _version,
+    _compat,
+    type_workarounds,
+)
+with _import.delay_import('cirq.protocols'):
+    from cirq import (
+        # Core
```
mpharrigan: this and below both have heading `Core`

mpharrigan:
does this play nice with IDEs?

cirq/__init__.py
cduck: That's intended.  They are in the same group.  These are just names for how I think of the modules.

cduck:
Yes, it appears to work fine.  This is because everything is still imported normally and static analysis ignores things like hooking the import system.

PR #2043, opened 2019-08-29
Polish type annotations in cirq/circuits/

Strilanc:
- This is a mechanical change produced by the following steps:

1. Apply the following regex replacement:

    ```
    (: |-> |List\[|Iterable\[|Sequence\[|Optional\[|Callable\[\[)(circuits|ops|protocols)\.([a-zA-Z_]+\b)
    $1'cirq.$3'
    ```

2. Add 'import cirq' to files that mypy complains about

3. Autoformat


dabacon:
So what is our rule.  If there is a dependency on the module in the code (not annotations) then we use that in the type annotation, otherwise we use cirq?

Strilanc:
We may also transition to using `cirq.bla` within function code, but that introduces a lot of complications w.r.t. what is allowed to run during init. For now I would like to only transition the type annotations, because they affect what users see in documentation and autocomplete results.

dabacon:
Ah yeah so type annotations we should always aim for `cirq` or `cirq.google` our top level modules.

cirq/circuits/circuit.py
```diff
@@ -1005,8 +996,8 @@ def _pick_inserted_ops_moment_indices(operations: Sequence[ops.Operation],
     def _push_frontier(self,
                        early_frontier: Dict[ops.Qid, int],
                        late_frontier: Dict[ops.Qid, int],
```
dabacon: fix these too?

PR #2056, opened 2019-08-31
[Qudits] Add support to IdentityGate, MeasurementGate, several protocols

cduck:
Part of #933.

Support added:
- IdentityGate and Measurement can be applied to qudits.
- Protocols apply_channel, pauli_expansion, phase_by, and qasm either support qudits or don't crash when given qudit values.
- `testing.assert_implements_consistent_protocols()` correctly tests qudit values.

### DIFF
--- a/cirq/ops/common_gates.py
+++ b/cirq/ops/common_gates.py
@@ -439,7 +439,7 @@ class MeasurementGate(raw_types.Gate):
     """
 
     def __init__(self,
-                 num_qubits: int,
+                 num_qubits: Optional[int] = None,
                  key: str = '',
                  invert_mask: Tuple[bool, ...] = (),
                  qid_shape: Tuple[int, ...] = None) -> None:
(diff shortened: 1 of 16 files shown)


cirq/ops/common_gates.py
```diff
@@ -438,13 +438,11 @@ class MeasurementGate(raw_types.Gate):
     of measurements.
     """
 
-    def num_qubits(self) -> int:
-        return self._num_qubits
-
     def __init__(self,
                  num_qubits: int,
```
Strilanc: You can make this argument default to `None`, so that it's not necessayr to specify it if qid shape is specified.

cirq/ops/common_gates_test.py
```diff
@@ -610,6 +680,44 @@ def test_measurement_channel():
                        [0, 0, 0, 0],
                        [0, 0, 0, 0],
                        [0, 0, 0, 1]])))
+    np.testing.assert_allclose(
+            cirq.channel(cirq.MeasurementGate(2, qid_shape=(2, 3))),
+            (np.array([[1, 0, 0, 0, 0, 0],
```
Strilanc: `np.diag(cirq.one_hot` would shorten this up.

cirq/ops/common_gates.py
cduck: Added.  It makes the init code a bit more complicated.

cirq/ops/common_gates_test.py
cduck: Shortened.

PR #2060, opened 2019-09-02
Break cyclic dependence and remove remaining lazy imports from protocols

andbe91:
Follow-up to #2059 
By moving `_try_decompose_into_operations_and_qubits` to `decompose`, the cycling dependence in `unitary/has_unitary/apply_unitary` can be removed.

@cduck PTAL

### DIFF
--- a/cirq/protocols/decompose.py
+++ b/cirq/protocols/decompose.py
@@ -12,12 +12,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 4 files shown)


cirq/protocols/decompose.py
```diff
@@ -12,13 +12,25 @@
...
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    overload,
+    Sequence,
+    Tuple,
+    TypeVar,
+    Union,
```
cduck: Yapf might use fewer lines if you remove the trailing comma.

Issue #1873, opened 2019-07-25
Move api protobuf definitions from cirq.api.google to cirq.google.api


maffoo:
It's rather confusing to have both of these packages, so I propose we restructure. In particular, I would suggest that we move the protobuf definitions for v1 and v2 into `cirq.google.api.v1.protos` and `cirq.google.api.v2.protos`, respectively. We can reesport the `*_pb2` packages in the `__init__.py` for the `cirq.google.api.vN` packages, so that both protos and various helper functions can be used with a single import. One question would be whether we also want to change the package declarations in the proto files, since they would not match the python package location anymore. Thoughts?

dabacon:
I agree it would be nice to move these.  Need to be careful in doing this with downstream dependencies.

dstrain115:
Oh dear.  This is going to be painful.

I propose copying the protos from cirq.api.google to cirq.google.api so that there are two copies of all the protos temporarily.

Once everyone has switched over to using the new package name, then we can get rid of the old one.

This will cause some annoyance, since we will need to update two copies of protos for a little while, but it hopefully won't break anything.

What do people think?

dstrain115:
The list of things that need to change packages that I know about are:

- cirq
- cirq_internal
- pyle
- google3

I think that most other projects should be using the public interfaces of cirq rather than using the api directly.

PR #2073, opened 2019-09-04
Add a new copy of protos in cirq.google.api

dstrain115:
- We will eventually get rid of cirq.api.google.
- This will coalesce our two api.google directories together.


maffoo:
Ref: #1873. In that issue, you mentioned that these copies will have to coexist until everyone has switched over to the new package name. Do we have a place to track all the changes that will need to happen?

cirq/google/api/v1/__init__.py
```diff
@@ -13,6 +13,11 @@
 # limitations under the License.
 """Data format v1 for google api."""
 
+from cirq.api.google.v1 import (
```
dabacon: was this mean to not be google.api?

dabacon:
Do we think this will take long enough that we should add a test that just compares the files in case someone updates one and not other?

dstrain115:
> Do we think this will take long enough that we should add a test that just compares the files in case someone updates one and not other?

I really hope that it doesn't last that long that we need this.  A few weeks hopefully?  I will try to move things as quick as possible.  Can you and Craig just keep an eye out for changes to these files for now?

PR #2078, opened 2019-09-05
Add decompose for ZZPowGate

dstrain115:
- Testing is done by preexisting consistency checks.
- Thanks to Craig for telling me the decomposition.


cirq/ops/parity_gates.py
```diff
@@ -20,7 +20,7 @@
 
 from cirq import protocols
 from cirq._compat import proper_repr
-from cirq.ops import gate_features, eigen_gate
+from cirq.ops import gate_features, eigen_gate, ZPowGate, CZPowGate
```
cduck: Prefer importing the submodule that defines these.

cirq/ops/parity_gates_test.py
```diff
@@ -29,7 +29,7 @@
 )
 def test_eigen_gates_consistent_protocols(eigen_gate_type):
     cirq.testing.assert_eigengate_implements_consistent_protocols(
-            eigen_gate_type)
+        eigen_gate_type, ignoring_global_phase=True)
```
cduck: Can you add a global phase gate to the decomposition instead?

cirq/ops/parity_gates.py
dstrain115: Done.

cirq/ops/parity_gates_test.py
dstrain115: Done.

cirq/ops/parity_gates.py
```diff
@@ -159,6 +160,14 @@ class ZZPowGate(eigen_gate.EigenGate,
         where w = e^{i \pi t} and '.' means '0'.
     """
 
+    def _decompose_(self, qubits):
+        # See eigen_gate.EigenGate for more info about the global phase
+        phase = np.exp(1j * np.pi * self.global_shift * self.exponent)
+        yield GlobalPhaseOperation(coefficient=phase)
```
cduck: This global phase is really multiplied by the exponent?

This could also just be the global phase on the CZPowGate.

Strilanc: Could you double check that putting the wrong global phase causes a test failure, before submitting?

cirq/ops/parity_gates.py
```diff
@@ -21,7 +21,8 @@
 from cirq import protocols
 from cirq._compat import proper_repr
 from cirq.ops import gate_features, eigen_gate
-from cirq.ops.common_gates import _rads_func_symbol
+from cirq.ops.common_gates import _rads_func_symbol, ZPowGate, CZPowGate
+from cirq.ops.global_phase_op import GlobalPhaseOperation
```
cduck: You didn't change these imports to `... import common_gates, global_phase_op`.

cirq/ops/parity_gates.py
dstrain115: It definitely does.  That's why I was over at your desk complaining about global phase.  Turns out I just had the wrong phase that didn't take into account global_shift

dstrain115: I am using the information from here:
https://github.com/quantumlib/Cirq/blob/master/cirq/ops/eigen_gate.py#L99

If you don't multiply by the exponent, the test fails for gates that have global_shift.

cirq/ops/parity_gates.py
dstrain115: Changed.  Is this what you mean now?

cduck: Yes.  Thanks.

cirq/ops/parity_gates.py
cduck: @Strilanc `assert_eigengate_implements_consistent_protocols` checks several global_shift values.

cduck: Yes.  My suggestion was to set the global_shift argument to CZPowGate to e.g. `-global_shift/2` instead of using a separate GlobalPhaseOperation.  It's fine the way it is.

dstrain115: Oh! Now that I understand what you mean, I think that's a great idea.  I have gone and updated the global phase to be a ghobal_shift on the CZPowGate.

PR #2079, opened 2019-09-05
[Qudits] Fix qudit result with invert mask in simulators

cduck:
Part of #933.

Where the invert mask is True, switches the measurements of |0> and |1> without modifying |2> and above.

Also fix `str(TrialResult)` for qudits.

### DIFF
--- a/cirq/sim/density_matrix_simulator.py
+++ b/cirq/sim/density_matrix_simulator.py
@@ -290,8 +290,10 @@ class DensityMatrixSimulator(simulator.SimulatesSamples,
                             indices,
                             qid_shape=qid_shape,
                             out=state.tensor)
-                        corrected = [bit ^ mask for bit, mask in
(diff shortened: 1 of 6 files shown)


cirq/sim/density_matrix_simulator.py
```diff
@@ -290,8 +290,10 @@ def keep(potential_op: ops.Operation) -> bool:
                             indices,
                             qid_shape=qid_shape,
                             out=state.tensor)
-                        corrected = [bit ^ mask for bit, mask in
-                                     zip(bits, invert_mask)]
+                        corrected = [
+                            bit ^ (bit < 2 and mask)
```
Strilanc: Where does the `2` come from? This looks like it can't possibly be correct.

cduck: If (the measured value is 0 or 1) and invert_mask, then toggle.

Issue #2061, opened 2019-09-02
Engine doesn't serialize lists of dicts of params


dabacon:
Passing in a list of dicts to a param sweep doesn't serialize to the v2 protos
```
params=[{'a': 1, 'b': 2}, {'a':2, 'b':3}]
```

dstrain115:
Being solved by PR #2080

dstrain115:
You should now be able to serialize flattened sweeps and lists of dicts, as long as you make them a ListSweep.

e.g.  params=ListSweep([{'a': 1, 'b': 2}, {'a':2, 'b':3}])

PR #2080, opened 2019-09-06
Allowing lists of dicts and flattened sweeps to be serialized

dstrain115:
- Created cirq.study.dict_list_to_sweep to convert a list of dicts into a sweep.
- Transform _TransformedSweep into a real sweep so that it can be serialized

Address issue #2061


cirq/study/sweepable.py
```diff
@@ -78,6 +78,26 @@ def to_sweep(sweep_or_resolver_list: Union['Sweep', ParamResolverOrSimilarType,
         'Unexpected sweep-like value: {}'.format(sweep_or_resolver_list))
 
 
+def dict_list_to_sweep(dict_list: Iterable[Dict[str, float]]):
+    """Converts a list of dicts into a Zipped sweep
+
+    Args:
+        dict_list: this must be a sequence of dicts.  Each dict must have the
```
dabacon: nit: sequence -> iterable

In python sequences have`__len__` and `__getitem__` while iterables have `__iter__`.  Your type is correct as you only use iterating over the list.

cirq/study/flatten_expressions.py
```diff
@@ -337,7 +337,10 @@ def transform_sweep(self,
             sweep: The sweep to transform.
         """
         sweep = sweepable.to_sweep(sweep)
```
dabacon: Is there a reason the functionality of  `dict_list_to_sweep` isn't in `sweepable.to_sweep`.  When I read the `to_sweep` code it looks to me like it is trying to do the conversion you are doing (maybe it is failing), but that we aren't supporting ListSweep in api.v2.sweeps.

cirq/study/sweepable.py
dstrain115: Removed this function.

cirq/study/flatten_expressions.py
dstrain115: I moved this functionality into api.v2.sweeps.  Can you take another look?

cirq/study/flatten_expressions.py
```diff
@@ -337,7 +337,10 @@ def transform_sweep(self,
...
-        return _TransformedSweep(sweep, dict(self))
+        return sweeps.ListSweep([
+            dict(param_tuples) for param_tuples in _TransformedSweep(
```
cduck: Delete the `_TransformedSweep` class and move the needed code from `_TransformedSweep.param_tuples` to here.

cirq/study/resolver.py
```diff
@@ -80,6 +80,9 @@ def value_of(self,
                 return float(v)
         return value
 
+    def __iter__(self):
+        return self.param_dict.__iter__()
+
```
cduck: What's this for?

cirq/study/sweepable.py
```diff
@@ -14,7 +14,7 @@
 
 """Defines which types are Sweepable."""
 
-from typing import cast, Iterable, List, Union
+from typing import cast, Dict, Iterable, List, Union
```
cduck: Pylint didn't catch that `Dict` is not used?

cirq/google/api/v2/sweeps.py
```diff
@@ -54,6 +54,17 @@ def sweep_to_proto(
         out.single_sweep.parameter_key = sweep.key
         for point in sweep.points:
             out.single_sweep.points.points.append(point)
```
cduck: This isn't part of the PR but I noticed this.  It should use `extend` instead of a loop.

dstrain115: Done.

cirq/study/sweepable.py
dstrain115: Done.

cirq/study/resolver.py
dstrain115: I have to iterate across the keys of param resolvers in order to transform them into zips in the proto.  Basically, I have to transpose {'a':1,'b':2},{'a':3,'b':4} into a zip of 'a' {1,3} zipped with b {2,4}.  In order to do that, I need to iterate across the keys of the param resolver.

cirq/study/flatten_expressions.py
dstrain115: Done.  Collapsed it into one function.

cirq/study/flatten_expressions.py
```diff
@@ -13,8 +13,7 @@
 # limitations under the License.
 """Resolves symbolic expressions to unique symbols."""
 
-from typing import (overload, Any, Callable, Dict, List, Optional,
-                    Tuple, Union)
+from typing import (overload, Any, Callable, Dict, List, Optional, Tuple, Union)
```
cduck: Nit: Unnecessary parentheses.  Yapf can't remove them because it doesn't modify the token stream.

cirq/study/resolver.py
```diff
@@ -80,6 +80,9 @@ def value_of(self,
                 return float(v)
         return value
 
+    def __iter__(self):
+        return self.param_dict.__iter__()
```
maffoo: nit: can do `return iter(self.param_dict)`

dstrain115:
Adding a trivial comment to try to make CLA bot restart.

PR #2086, opened 2019-09-09
cirq.google.api is the new cirq.api.google

dstrain115:
- Exterminate all cirq.api.google references
- Welcome to our new cirq.google.api


dstrain115:
> Are we going to delete the `cirq/api/google` directory in the future?

Yes.  Do you foresee a problem with that?

PR #2089, opened 2019-09-10
[Qudits] Allow argument of Gate._decompose_ to not just be 'qubits'

cduck:
Part of #933.

This is a minor API change.
- `def _decompose_(self, qids):` will now work.
- `def _decompose_(self, **kwargs):` will no longer work.
- `def _decompose_(self, something_else=5, qubits=None):` will no longer work.

### DIFF
--- a/cirq/protocols/decompose.py
+++ b/cirq/protocols/decompose.py
@@ -312,10 +312,10 @@ def decompose_once(val: Any,
             `val`.
 
     Returns:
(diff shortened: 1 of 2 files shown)


cduck:
Yes.  When writing a qutrit gate, it's not natural to write qubits for the argument and it was surprising to me that the protocol didn't use a positional argument.

Issue #1973, opened 2019-08-16
Why does `ControlledGate` have a notion of qubits


mpharrigan:
`ControlledGate.control_qubits` is a tuple of `Qid` (or `None`s). I thought "gates" don't have qubits and that that is reserved for `Operation`s.

c-poole:
`ControlledGate` seems to use the `control_qubits` property primarily as a way of helping it interpret how to use a qubit list that has been provided to it by methods like `_decompose_`, `_on_`, and such. It is primarily being used decide which qubits in the provided list are controls and which are targets using a (seemingly?) undocumented convention that control qubits come earlier in the list. The flexibility that this provides is that you can specify a subset of the control qubits initially and use the resulting `ControlledGate` class as a `Gate` or `Operation` generator. I'm not sure if there is a need for this degree of flexibility in generation operations though. It seems like `ControlledGate` is redundant if we have `ControlledOperation`.

dabacon:
This is to have controlled gates that are associated with control on some fixed qubits.  I agree that it is confusing and argued against it in the original PR.  It adds complexity, violates the usual case that gates do not have qubits, and leads to code that is only slightly nicer in the case that it is used.

Strilanc:
> I thought "gates" don't have qubits and that that is reserved for Operations.

It would be more accurate to say that a gate has some (but not necessarily only) unspecified qubits. A gate is a function that takes qubits and returns an operation.

dabacon:
> It would be more accurate to say that a gate has some (but not necessarily only) unspecified qubits. A gate is a callable, plus convenience methods, that takes qubits and returns an operation.

That's what's at debate.  I think that including qubits in the constructor is conceptually confusing. 
Maybe a solution is to hide these sort of qubits in a class constructor (so they are only set if you come in from `ControlledGate.controlled_from_fixed_qubits()`)

The other issue with this pattern is that it breaks any code that reasons "the only qubits this gate acts on are those that are operated on by the `on` method".  There are other cases where `Operation` breaks this as well, and I worry a lot about these (I would bet there is already broken code for these types of operations)

mpharrigan:
If you can have partially specified qubits, I'm led to ask why bother having a difference between `Gate` and `Operation` at all? There could be one object that has any of its qubits specified or not. If all of them are there, it would be similar to an `Operation`. If none of them are there, it would be similar to (most) `Gate`s. 

[...]

PR #2090, opened 2019-09-10
Remove prespecified control qubits from ControlledGate.

smitsanghavi:
Also rename controlled_by to control to allow for no qubits to be specified in case of operating on Gates.


smitsanghavi:
#1973

cirq/ops/controlled_gate.py
```diff
@@ -86,109 +75,81 @@ def __init__(
...
-                    'Control values <{!r}> outside of range for qubit '
-                    '<{!r}>.'.format(val, q))
+                    'Control values <{!r}> outside of range for control qubit '
+                    'number <{!r}>.'.format(val, i))
 
         # Flatten nested ControlledGates.
         if isinstance(sub_gate, ControlledGate):
             self.sub_gate = sub_gate.sub_gate  # type: ignore
-            self.control_qubits += sub_gate.control_qubits
+            self.num_controls += sub_gate.num_controls
```
cduck: You don't need self.num_controls.  Just use len(control_qid_shape).

cirq/ops/controlled_gate.py
```diff
@@ -86,109 +75,81 @@ def __init__(
...
             self.control_values += sub_gate.control_values
             self.control_qid_shape += sub_gate.control_qid_shape
         else:
             self.sub_gate = sub_gate
 
-    def num_controls(self) -> int:
-        return len(self.control_qubits)
-
```
cduck: Put this back.  I would avoid changing the API from .num_controls() to .num_controls.

cirq/ops/controlled_gate.py
```diff
@@ -86,109 +75,81 @@ def __init__(
...
-                continue
-            q = next(iter_qubits)
+        if len(qubits) < self.num_qubits():
+            raise ValueError('Not all qubits specified.')
+        for q, dimension in zip(qubits, self.control_qid_shape):
             if q.dimension != dimension:
                 raise ValueError(
                     "Control qids have dimensions that don't match this gate's "
                     "control_qid_shape <{!r}>.  Qids are <{!r}>".format(
                         self.control_qid_shape, qubits))
```
cduck: The above code can probably be replaced with `super().validate_args(qubits)`.

cirq/ops/controlled_gate_test.py
```diff
@@ -118,72 +102,56 @@ def test_init():
...
         cirq.ControlledGate(cirq.Z, num_controls=1, control_values=(1, 0))
     with pytest.raises(ValueError,
                        match=r'len\(control_qid_shape\) != num_controls'):
         cirq.ControlledGate(cirq.Z, num_controls=1, control_qid_shape=(2, 2))
-    with pytest.raises(ValueError, match='qubit dimension .*does not match'):
-        cirq.ControlledGate(cirq.Z, [p], control_qid_shape=(3,))
     with pytest.raises(ValueError, match='Control values .*outside of range'):
-        cirq.ControlledGate(cirq.Z, [p], control_values=[2])
-    with pytest.raises(ValueError, match='Control values .*outside of range'):
-        cirq.ControlledGate(cirq.Z, [p], control_values=[(1, -1)])
```
cduck: Convert this test instead of deleting.

cirq/ops/controlled_gate_test.py
```diff
@@ -272,54 +207,35 @@ def test_validate_args():
...
-    eq.add_equality_group(SCSCH)
     eq.add_equality_group(cirq.ControlledGate(cirq.H))
     eq.add_equality_group(cirq.ControlledGate(cirq.X))
     eq.add_equality_group(cirq.X)
-    eq.add_equality_group(
-        cirq.ControlledGate(cirq.H, [q, p3], control_values=[1, (0, 2)]),
-        cirq.ControlledGate(cirq.H, [p3, q], control_values=[(2, 0), 1]))
-    eq.add_equality_group(
-        cirq.ControlledGate(cirq.H, [q, p3], control_values=[1, 0]),
-        cirq.ControlledGate(cirq.H, [p3, q], control_values=[0, 1]))
```
cduck: Keep these two groups by specifying the control qid shape instead.

cirq/ops/raw_types.py
```diff
@@ -262,21 +262,14 @@ def __pow__(self, power):
     def __call__(self, *args, **kwargs):
         return self.on(*args, **kwargs)
 
-    def controlled_by(self,
-                      *control_qubits: Qid,
-                      control_values: Optional[Sequence[
-                          Union[int, Collection[int]]]] = None) -> 'Gate':
+    def control(self,
+                control_values: Optional[Sequence[
+                    Union[int, Collection[int]]]] = None) -> 'Gate':
```
cduck: I would add a num_controls argument and a qid shape argument.

cirq/ops/raw_types.py
```diff
@@ -262,21 +262,14 @@ def __pow__(self, power):
...
         """Returns a controlled version of this gate.
-
-        Args:
-            control_qubits: Optional qubits to control the gate by.
```
cduck: Document the arguments.  (It looks like I forgot to document control_values.)

cirq/ops/raw_types.py
```diff
@@ -262,21 +262,14 @@ def __pow__(self, power):
...
         """
         # Avoids circular import.
         from cirq.ops import ControlledGate
-        if len(control_qubits) == 0:
-            return self
```
cduck: Change the condition to check the length of control_values.

cirq/ops/raw_types.py
```diff
@@ -362,10 +355,10 @@ def transform_qubits(self, func: Callable[[Qid], Qid]) -> 'Operation':
         """
         return self.with_qubits(*(func(q) for q in self.qubits))
 
-    def controlled_by(self,
-                      *control_qubits: Qid,
-                      control_values: Optional[Sequence[
-                          Union[int, Collection[int]]]] = None) -> 'Operation':
+    def control(self,
```
cduck: I don't know if it makes sense to rename Operation.controlled_by.  This will break code that doesn't use partially specified ControlledGates.

cirq/protocols/control.py
```diff
@@ -33,7 +33,7 @@ def control(controllee: Union['cirq.Gate', op_tree.OP_TREE],
     """Returns a Controlled version of the given value, if defined.
 
     Controllees define how to be controlled by defining a method
-    controlled_by(self, control_qubits). Note that the method may return
+    control(self, control_qubits). Note that the method may return
```
cduck: This is incorrect.  ControlledGate.control doesn't have this signature.

I think Gates can no longer support this protocol.

cirq/ops/controlled_gate_test.py
```diff
@@ -489,21 +333,10 @@ def test_extrapolatable_effect():
...
 def test_reversible():
-    assert (cirq.inverse(cirq.ControlledGate(cirq.S)) ==
-            cirq.ControlledGate(cirq.S**-1))
-    assert (cirq.inverse(cirq.ControlledGate(cirq.S, [q])) ==
-            cirq.ControlledGate(cirq.S**-1, [q]))
-    assert (cirq.inverse(cirq.ControlledGate(cirq.S,
-                                             [q3])) == cirq.ControlledGate(
-                                                 cirq.S**-1, [q3]))
+    assert (cirq.inverse(cirq.ControlledGate(cirq.S)) == cirq.ControlledGate(
+        cirq.S**-1))
```
cduck: Add tests where the control is a qutrit or the control value is specified.

cirq/ops/controlled_gate_test.py
```diff
@@ -642,32 +471,20 @@ def test_repr():
     cirq.testing.assert_equivalent_repr(
         cirq.ControlledGate(cirq.Z, num_controls=2))
     cirq.testing.assert_equivalent_repr(
-        cirq.ControlledGate(cirq.Y, control_qubits=[cirq.LineQubit(1)]))
+        cirq.ControlledGate(cirq.Y, num_controls=1))
```
cduck: You can delete this assert.  It's the same as the second one.

cirq/ops/controlled_gate.py
```diff
@@ -39,7 +38,6 @@ def __init__(
 
         Args:
             sub_gate: The gate to add a control qubit to.
-            control_qubits: The qubits that would act as controls.
             num_controls: Total number of control qubits.
             control_values: For which control qubit values to apply the sub
                 gate.  A sequence the same length as `control_qubits` where each
```
mpharrigan: "sequence the same length as `control_qubits`" no longer makes sense

cirq/ops/controlled_gate.py
smitsanghavi: Removed validate_args altogether.

cirq/ops/raw_types.py
smitsanghavi: Documented. Also added documentation to the Operation controlled_by and TODOs for adding these arguments to that and the control protocol.

cirq/ops/raw_types.py
smitsanghavi: I think the default should be to create a ControlledGate with one control qubit.

cirq/protocols/control.py
smitsanghavi: Removed Gates from the signature

smitsanghavi:
Reverted the breaking changes to Operation.controlled_by and the control protocol (except removing Gate from control protocol signature, for now)

cirq/ops/raw_types.py
```diff
@@ -263,21 +263,29 @@ def __pow__(self, power):
     def __call__(self, *args, **kwargs):
         return self.on(*args, **kwargs)
 
-    def controlled_by(self,
-                      *control_qubits: Qid,
-                      control_values: Optional[Sequence[
-                          Union[int, Collection[int]]]] = None) -> 'Gate':
+    def control(self,
```
Strilanc: Renamed this to `controlled`

cirq/ops/raw_types.py
```diff
@@ -371,6 +379,14 @@ def controlled_by(self,
...
         Args:
             control_qubits: Qubits to control the operation by. Required.
+            control_values: For which control qubit values to apply the
+                operation.  A sequence of the same length as `control_qubits`
+                where each entry is an integer (or set of integers)
+                corresponding to the qubit value (or set of possible values)
+                where that control is enabled.  When all controls are enabled,
+                the operation is applied.  If unspecified, control values
+                default to 1.
+            TODO: add control_qid_shape
```
Strilanc: Do the todo.

cirq/protocols/control.py
```diff
@@ -27,7 +27,7 @@
 TDefault = TypeVar('TDefault')
 
 
-def control(controllee: Union['cirq.Gate', op_tree.OP_TREE],
+def control(controllee: op_tree.OP_TREE,
```
Strilanc: I think we should delete this entire file (`cirq/protocols/control.py`). With gate removed, this protocol is effectively dominated by the `controlled_by` method on operation. Repeating across lists isn't enough of an "add on" to justify keeping it.

Now is also the perfect time to do it, since the last release didn't have it; it will avoid a breaking change.

smitsanghavi: I was planning to add support for gates in the subsequent PR that addressed the TODO in this file. It would throw error if the control qubits argument is specified when the operand is a Gate. But it would still be able to use the Gate.controlled() method with control_values and control_qid_shape. Does that sound like something worth adding or should I continue with deleting the file?

cirq/ops/raw_types.py
smitsanghavi: Given that I reverted all changes to Operation and control protocol from this PR, I figured it would be better to address in a subsequent one along with the control protocol changes.
But if we want to delete control protocol as per the comment below, I'll do this TODO here itself.

cirq/protocols/control.py
Strilanc: I would continue with deleting it. However, that can be a separate PR.

Issue #2075, opened 2019-09-04
QuantumEngine takes a list of processor_ids?


mpharrigan:
When using `QuantumEngine`

>             processor_ids: The engine processors that should be candidates
                to run the program. Only one of these will be scheduled for
                execution.

What's the intended use case here? Can you figure out which processor actually ran your job?  In the NISQ context you typically want to tailor a circuit to a device.

A related question: what's the utility of `Circuit.device`? Is it respected/validated when using `QuantumEngine`?

dabacon:
The multiple processor_id case is for example when you are testing, and there are multiple processors, and you don't care which one it runs on.  This can increase throughput for you.  I agree for most NISQy business you want to specify the processor

dabacon:
Circuit.device is not used at all in engine right now.

We are adding the ability to down device specifications, at which point that could be used.  You can always add your device to a circuit and it will validate as you build the circuit.

dabacon:
In some places where is says processor_ids you can also pass in a single str.  I don't really like that.  Maybe we should just be more explicit here and have both processor_id and processor_ids, you gotta specify one and not both.

mpharrigan:
My personal opinion is that the cognitive overhead of giving a list outweighs the (IMO marginal) utility of the current api.

Overhead: Users following the intro documentation / notebooks / examples will be surprised by the pluralized argument and visually cluttered `('xmonsim',),` argument, expecting a single target. Furthermore, I don't think it's obvious whether the process will run on *any* or *all* of the targets without reading the docstring

Marginal utility: NISQy business concerns, per above.

I would advocate for hiding multi-processor_ids in most APIs; maybe only exposing it on e.g. `Engine.create_job` which most users won't be using directly.

Strilanc:
I suggest adding a `processor_id` argument, which takes a string and throws an error if specified at the same time as `processor_ids`. That fixes the "tricky to learn" issue.

dabacon:
Another question is whether we want default values of these specified on Engine.  I'm sure everyone is tired of typing `rainbow` by now?

mpharrigan:
For forward compatibility and general readability, I'd strongly suggest not having any default. I was fooled once by the current default

dabacon:
Current state is that we still have multiple ids, but we no longer have places where a single string is acceptable.  

There is a good case to not have a default.  
There is a good case for having a `processor_id`.

[...]

PR #2096, opened 2019-09-11
Add `sampler` method to `cirq.google.Engine`

### LINKED ISSUE #2095: Proposal: Merge QuantumEngineSampler and Engine
Often when writing code that uses Engine you really want a Sampler so that you can easily swap in a simulator. It seems unwieldy to have both these classes around and one kind of has to look around or be told about QuantumEngineSampler to realize it exists.

kevinsung:
Fixes #2095

### DIFF
--- a/cirq/google/engine/engine.py
+++ b/cirq/google/engine/engine.py
@@ -697,6 +697,14 @@ class Engine:
     def sampler(self, processor_id: Union[str, List[str]],
                 gate_set: serializable_gate_set.SerializableGateSet
                ) -> engine_sampler.QuantumEngineSampler:
+        """Returns a sampler backed by the engine.
+
+        Args:
+            processor_id: String identifier, or list of string identifiers,
+                determining which processors may be used when sampling.
+            gate_set: Determines how to serialize circuits when requesting
+                samples.
+        """
         return engine_sampler.QuantumEngineSampler(engine=self,
                                                    processor_id=processor_id,
                                                    gate_set=gate_set)
--- a/cirq/google/engine/engine.py
+++ b/cirq/google/engine/engine.py
@@ -694,9 +694,9 @@ class Engine:
             name=calibration_name).execute()
         return calibration.Calibration(response['data']['data'])
 
-    def sampler(self,
-                processor_id: Union[str, List[str]],
+    def sampler(self, processor_id: Union[str, List[str]],
                 gate_set: serializable_gate_set.SerializableGateSet
                ) -> engine_sampler.QuantumEngineSampler:
-        return engine_sampler.QuantumEngineSampler(
-            engine=self, processor_id=processor_id, gate_set=gate_set)
+        return engine_sampler.QuantumEngineSampler(engine=self,
+                                                   processor_id=processor_id,
+                                                   gate_set=gate_set)
--- a/cirq/google/engine/engine.py
+++ b/cirq/google/engine/engine.py
@@ -41,7 +41,8 @@ from cirq.api.google import v1, v2
 from cirq.google import gate_sets, serializable_gate_set
 from cirq.google.api import v1 as api_v1
 from cirq.google.api import v2 as api_v2
-from cirq.google.engine import calibration, engine_job, engine_program
+from cirq.google.engine import (
+        calibration, engine_job, engine_program, engine_sampler)
 
 gcs_prefix_pattern = re.compile('gs://[a-z0-9._/-]+')
 TYPE_PREFIX = 'type.googleapis.com/'
@@ -692,3 +693,11 @@ class Engine:
         response = self.service.projects().processors().calibrations().get(
             name=calibration_name).execute()
         return calibration.Calibration(response['data']['data'])
+
+    def sampler(self,
+                processor_id: Union[str, List[str]],
+                gate_set: serializable_gate_set.SerializableGateSet
+                ) -> engine_sampler.QuantumEngineSampler:
+        return engine_sampler.QuantumEngineSampler(engine=self,
+                                                   processor_id=processor_id,
+                                                   gate_set=gate_set)
--- a/cirq/google/engine/engine.py
+++ b/cirq/google/engine/engine.py
@@ -41,8 +41,8 @@ from cirq.api.google import v1, v2
 from cirq.google import gate_sets, serializable_gate_set
 from cirq.google.api import v1 as api_v1
 from cirq.google.api import v2 as api_v2
-from cirq.google.engine import (
-        calibration, engine_job, engine_program, engine_sampler)
+from cirq.google.engine import (calibration, engine_job, engine_program,
+                                engine_sampler)
 
 gcs_prefix_pattern = re.compile('gs://[a-z0-9._/-]+')
 TYPE_PREFIX = 'type.googleapis.com/'
@@ -697,7 +697,6 @@ class Engine:
     def sampler(self,
                 processor_id: Union[str, List[str]],
                 gate_set: serializable_gate_set.SerializableGateSet
-                ) -> engine_sampler.QuantumEngineSampler:
-        return engine_sampler.QuantumEngineSampler(engine=self,
-                                                   processor_id=processor_id,
-                                                   gate_set=gate_set)
+               ) -> engine_sampler.QuantumEngineSampler:
+        return engine_sampler.QuantumEngineSampler(
+            engine=self, processor_id=processor_id, gate_set=gate_set)
--- a/cirq/google/engine/engine_test.py
+++ b/cirq/google/engine/engine_test.py
@@ -992,30 +992,40 @@ def test_sampler(build):
     programs = service.projects().programs()
     jobs = programs.jobs()
     programs.create().execute.return_value = {
-        'name': 'projects/project-id/programs/test'}
+        'name': 'projects/project-id/programs/test'
+    }
     jobs.create().execute.return_value = {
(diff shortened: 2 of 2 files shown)


cirq/google/engine/engine.py
```diff
@@ -263,18 +255,13 @@ def program_as_schedule(self, program: TProgram) -> schedules.Schedule:
...
-            priority: int = 500,
-            processor_ids: Sequence[str] = ('xmonsim',),
-            gate_set: serializable_gate_set.SerializableGateSet = gate_sets.XMON
-    ) -> engine_job.EngineJob:
+    def run_sweep(self,
+                  program: Union[circuits.Circuit, schedules.Schedule],
+                  params: study.Sweepable,
+                  repetitions: int = 1,
+                  program_id: Optional[str] = None,
+                  job_config: Optional[JobConfig] = None
```
kevinsung: Is it okay to have the additional keyword arguments `program_id` and `job_config` even though they are not present in `Sampler.run_sweep`? I don't think it makes sense to set these in `__init__` because if they are both set to non-None values then they need to be changed for every run.

cirq/google/engine/engine_sampler_test.py
```diff
@@ -1,47 +0,0 @@
...
-    params = [cirq.ParamResolver({'a': 1})]
-    sampler.run_sweep(circuit, params, 5)
-    engine.run_sweep.assert_called_with(gate_set=cg.XMON,
-                                        params=params,
-                                        processor_ids=['tmp'],
-                                        program=circuit,
-                                        repetitions=5)
-
-
-def test_run_engine_program():
```
Strilanc: Move these tests into the `engine_test.py` file.

cirq/google/engine/engine.py
```diff
@@ -155,13 +155,18 @@ class Engine:
...
+    def __init__(
+            self,
+            project_id: str,
+            version: Optional[str] = None,
+            discovery_url: Optional[str] = None,
+            default_gcs_prefix: Optional[str] = None,
+            proto_version: ProtoVersion = ProtoVersion.V1,
+            service_args: Optional[Dict] = None,
+            priority: int = 50,
+            processor_ids: Sequence[str] = ('xmonsim',),
```
Strilanc: You need to document these additional parameters in the docstring. Explain that they are default values used if more specific ones are not specified on an individual call.

cirq/google/engine/engine.py
```diff
@@ -206,49 +217,29 @@ def __init__(self,
...
+            gate_set: Optional[serializable_gate_set.SerializableGateSet] = None
+    ) -> 'Engine':
+        return Engine(project_id = project_id or self.project_id,
+                      version = version,
+                      discovery_url = discovery_url or self.discovery_url,
+                      default_gcs_prefix = (default_gcs_prefix
+                                            or self.default_gcs_prefix),
+                      proto_version = proto_version or self.proto_version,
+                      service_args = service_args,
+                      priority = priority or self.priority,
```
Strilanc: What if they set `priority` to 0?

cirq/google/engine/engine.py
```diff
@@ -206,49 +217,29 @@ def __init__(self,
...
+                      processor_ids = processor_ids or self.processor_ids,
```
Strilanc: What if they set `processor_ids` to the empty list?

cirq/google/engine/engine.py
```diff
@@ -206,49 +217,29 @@ def __init__(self,
                 discoveryServiceUrl=self.discovery_url,
                 **service_args)
 
-    def run(
+    def copy(
```
Strilanc: Calling `copy` is going to cause another call to the discovery URL service. That seems surprising.

cirq/google/engine/engine.py
Strilanc: I would prefer for us to stick to the interface exactly... @dabacon thoughts?

maffoo: I would also prefer to stick to the interface; it would be good to check whether mypy (and abc and...) still considers a valid "implementation" if there are extra params like this.

I've stumbled on this as well with some internal implementations of `Sampler` where we need extra per-call information that is implementation-specific. Could we add an extra argument in the `Sampler` interface to allow passing extra info which implementations can use as needed?

Strilanc: Wouldn't that harm the interchangeability? E.g. one of the really nice things about EngineSampler is that I can substitute `cirq.Simulator` with absolutely no other changes. If EngineSampler took some additional arguments, that would not be the case.

kevinsung:
@Strilanc if we take your suggestion to keep QuantumEngineSampler as a private class, an instance of which is returned by Engine.processor, should we keep the `run` and `run_sweep` methods of Engine? It seems ugly to have two different ways of doing the same thing.

mpharrigan:
`cirq.Circuit` should know its own (optional) `program_id`. `JobConfig` attributes can be (optionally) set upon instantiation of the engine-sampler. 

@Strilanc : 
(1): I generally, agree; see suggestion above
(2): I propose: `QuantumEngineSampler` "has a" `Engine`.
```python
class QuantumEngineSampler:
  def __init__(self, processor_id, engine=None):
    if engine is None:
      engine = Engine()  # with sensible defaults
    self.engine = engine
    self.processor_id = processor_id
```

(3): I agree

Strilanc:
> cirq.Circuit should know its own (optional) program_id

Adding metadata to circuits is an interesting idea. How would we validate that the program id was actually correct? How does metadata change when the circuit changes, e.g under concatenation?

> (2): I propose: QuantumEngineSampler "has a" Engine.

Yes, that's how it already was.

kevinsung:
The format script `check/format-incremental` isn't working for me. Even though Travis says I need reformatting, running the script locally tells me that I don't need formatting.

mpharrigan:
> How would we validate that the program id was actually correct? How does metadata change when the circuit changes, e.g under concatenation?

I don't know what constitutes a correct program id, but the metadata's behavior under circuit modification is a good question. This could be a good time to seriously give it some thought, as circuits already have a `device` field that tags along and I'd suspect that there are some operations either in cirq or in the wild that don't preserver that field like one would expect

> Yes, that's how it already was.

Ah very good. I'm mainly pushing back against a chained construction like `sampler = cirq.google.Engine(project_id=********).processor(id='bristlecone', gate_set=XMON)` instead of `cirq.EngineSampler(project_id, processor_id, gate_set)` which would behave closer to `cirq.Simulator(dtype)` or `cirq.DensityMatrixSimulator()`

Strilanc:
If an EngineSampler "has a" Engine, then it should take an engine in its constructor. Otherwise unit testing becomes harder, because you have to intercept the calls being used inside the constructor to make the engine. This naturally leads to the chaining method style, or to making an engine and then passing it in to the sampler.

mpharrigan:
I misspoke (mistyped) by including `project_id` in the list of arguments. Specifically what I'd like to see is as sketched out in my previous comment (https://github.com/quantumlib/Cirq/pull/2096#issuecomment-531320912) 

In ordinary usage of cirq, users shouldn't have to be bothered about what `Engine` instance they've got

Strilanc:
@mpharrigan I don't think we can provide sensible defaults for the engine object, because it has a project id and also a gate set both of which are likely to change from person to person and from time to time. I like the flow of "tell me about auth" and then "tell me about where you want to run". I also like that the `sampler` method pushes the users towards understanding that we have some kind of generic concept for sampling.

mpharrigan:
1. I thought project_id is usually available as an environment variable
2. Why is gateset a part of engine and not the sampler? That doesn't seem like a part of the auth flow

cirq/google/engine/engine.py
```diff
@@ -692,3 +693,10 @@ def get_calibration(self, calibration_name: str) -> calibration.Calibration:
         response = self.service.projects().processors().calibrations().get(
             name=calibration_name).execute()
         return calibration.Calibration(response['data']['data'])
+
+    def sampler(self, processor_id: Union[str, List[str]],
```
mpharrigan: I'm going to continue to object to accepting a list of processor_id's. Please document the (surprising!) semantics in this wrapper method if people are adamant that we keep it

Strilanc:
> Why is gateset a part of engine and not the sampler? That doesn't seem like a part of the auth flow

That's a good point. I think we should move it. But in a separate PR.

> I thought project_id is usually available as an environment variable

I also think project_id works well as an environment variable. There is code somewhere that does that IIRC. But I do expect to confuse more new users with the instruction "set this environment variable" as opposed to "paste this code and tweak the string". I think it is totally reasonable to have the user say `Engine().sampler(...)` instead of `QuantumEngineSampler(...)`.

cirq/google/engine/engine.py
kevinsung: I copied documentation from QuantumEngineSampler but you may object that it is too vague. It's the best I can do though because I don't actually know the semantics (i.e., which processor gets executed? Do they get tried one at a time from the first one? I don't know. Perhaps @Strilanc can answer).

Strilanc:
I agree that we should replace processor_ids with a single string processor_id. It can happen to take a list and have it work, but for now the default workflow is to specify one specific processor.

cirq/google/engine/engine.py
dabacon: If many processors are supplied it will schedule on the first available.  

See #2075  for a discussion of this, I think we decided there to have two separate args?

Issue #1890, opened 2019-07-26
Plotting code should not call `show`


maffoo:
Calling `show` in some cases leads to warning messages, for example if using an inline backend in colab:
```
/home/maffoo/.virtualenvs/3.6/lib/python3.6/site-packages/matplotlib/figure.py:445: UserWarning: Matplotlib is currently using module://ipykernel.pylab.backend_inline, which is a non-GUI backend, so cannot show the figure.
  % get_backend())
```
We could try to detect which backend is running, but I think that way lies pain. Better IMO would be to just not call show, since in common environments (colab, ipython with matplotlib integration) it's not needed.

Strilanc:
The reason I wanted these calls included is because, without them, the plot method fails to show a plot when running from the command line, from python's interpreter, and from within pycharm.

We could perhaps have `plot` and `show_plot`? Seems confusing...

Intercept the warning message and dump it? Seems dangerous...

Detect the backend and react appropriately? Sounds like massive jank...

Do we have data on how often users are using each environment?

maffoo:
I would bet that most users will be in a notebook environment (colab or jupyter) or ipython, but I don't have data to say what the ratios are. We certainly have lots of colab examples for introductory purposes and it seems to me that emitting warnings for plots in those examples is not great. If there's a way to detect whether the matplotlib backend is a "GUI" backend and do it robustly, then we could have a helper method like `try_show_plot` that we would call instead.

viathor:
FWIW, on the commandline you can just append `; plt.show()`...

Strilanc:
I agree that we want to fix the warning in notebook environments, but I'd strongly prefer to not require different code in different environments to make the plots pop up. I want to be able to say "call .plot to see a plot" in the documentation, instead of "if you are in a notebook environment such as ipython or jupyer call .plot to see a plot. If you are not, then you also need to import matplotlib.pyplot as plt and then call plt.show()".

dabacon:
We now have both patterns in Cirq.  Probably we should make a decision on this.

mpharrigan:
Coming here from @cduck 's xref on #1992. Library code should never call `show`

dabacon:
I agree that we shouldn't call show by default.  We could have a `show` bool parameter, so that people who really want to call show would do `plot(show=True)`.  I do think we should optimize the plotting for common environments, which means colab / ipython type setups, which also argues against calling show.

dabacon:
[...]

PR #2097, opened 2019-09-11
Add a cirq.plot() function and SupportsPlot protocol

pingyeh:
This implements what I proposed in https://github.com/quantumlib/Cirq/issues/1890#issuecomment-524569028.
 - A `SupportsPlot` protocol containing a single `_plot_` method. Think of it as a graphic version of `__str__`. It should only plot on the axes given in the arg, and never call `show()`.
 - A `cirq.plot` function that calls `_plot_` with an axes instance. Think of it as a graphic version of `print`. It creates a new figure and axes if no axes is given, passes the axes to `_plot_` for rendering, and calls `show`. It is intended for interactive use.

I migrated `Heatmap`, `CrossEntropyResult`, `RabiResult` and `RandomizedBenchMarkResult` to `SupportsPlot` protocol in this PR. `TomographyResult` looks a bit more complicated so I left a TODO.

Two demo codes and plots are shown below.

heatmap

rabi_oscillation

### DIFF
--- a/cirq/__init__.py
+++ b/cirq/__init__.py
@@ -420,7 +420,9 @@ from cirq.neutral_atoms import (
 )
 
 from cirq.vis import (
-    Heatmap,)
+    Heatmap,
+    plot
+)
 
 from cirq.work import (
     CircuitSampleJob,
--- a/cirq/__init__.py
+++ b/cirq/__init__.py
@@ -419,10 +419,7 @@ from cirq.neutral_atoms import (
     NeutralAtomDevice,
 )
 
-from cirq.vis import (
-    Heatmap,
-    plot
-)
+from cirq.vis import (Heatmap, plot)
 
 from cirq.work import (
     CircuitSampleJob,
--- a/cirq/experiments/cross_entropy_benchmarking.py
+++ b/cirq/experiments/cross_entropy_benchmarking.py
@@ -31,20 +31,30 @@ class CrossEntropyResult:
         return self._data
 
     def plot(self, **plot_kwargs: Any) -> None:
-        """Plots the average XEB fidelity vs the number of cycles.
+        """Plots mean XEB fidelity vs number of cycles on a new figure.
 
         Args:
             **plot_kwargs: Arguments to be passed to 'matplotlib.pyplot.plot'.
         """
-        num_cycles = [d.num_cycle for d in self._data]
-        fidelities = [d.xeb_fidelity for d in self._data]
+        # TODO(pingyeh): deprecate this in favor of the SupportsPlot protocol.
         fig = plt.figure()
         ax = plt.gca()
-        ax.set_ylim([0, 1.1])
-        plt.plot(num_cycles, fidelities, 'ro-', figure=fig, **plot_kwargs)
-        plt.xlabel('Number of Cycles', figure=fig)
-        plt.ylabel('XEB Fidelity', figure=fig)
+        self._plot_(ax, **plot_kwargs)
         fig.show(warn=False)
+
+    def _plot_(self, ax: plt.Axes, **plot_kwargs: Any) -> None:
+        """Plots mean XEB fidelity vs number of cycles onto ax.
+
+        Args:
+            ax: the axes to plot onto.
+            **plot_kwargs: Arguments to be passed to 'ax.plot'.
+        """
+        num_cycles = [d.num_cycle for d in self._data]
+        fidelities = [d.xeb_fidelity for d in self._data]
+        ax.set_ylim([0, 1.1])
+        ax.plot(num_cycles, fidelities, 'ro-', **plot_kwargs)
+        ax.set_xlabel('Number of Cycles')
+        ax.set_ylabel('XEB Fidelity')
 
 
 def cross_entropy_benchmarking(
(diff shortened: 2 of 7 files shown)


Strilanc:
What is the advantage of this over just calling `object.plot()`?

For example, the advantage of `cirq.unitary` is that it can fallback to other methods (such as decompose) and works on native python types such as a list (of operations). The advantage of `cirq.approx_equal` is all of the fallback logic around trying both objects and using `__eq__`, which would be tedious to add to every object.

pingyeh:
As described in my proposal https://github.com/quantumlib/Cirq/issues/1890#issuecomment-524569028, this design solves the tension between 2 usecases:
1. People want to show figures in interactive sessions, with less keystrokes if possible.
2. People don't want to see figures popping up in tests.

This design makes `_plot_` method plot onto the Axes passed in, so there is no figures popping up in tests. The `cirq.plot` function provides a easy shorthand to show the graph in interactive sessions while re-using the logic in `_plot_` methods.

In this design, `_plot_` is a library which should never call `show`. I think it is a better solution than `show(warn=False)`. 

@dabacon, @mpharrigan, what do you think?

Strilanc:
But couldn't that also be solved by having the object have a `plot` method that took the axes? I don't see the user's incentive to have the method global.

pingyeh:
You have to do 2 things for a `plot` method to work both as a library and in interactive sessions:
1. Create an axes if one is not passed in.
2. Call `show` with `warn=False` so it doesn't cause warnings in tests.

(1) is repeated logic in every `plot` method. The global `cirq.plot` method put that logic in one place. I like the DRY perspective of that.

(2) is likely not the best solution, either. We just should not call `show` in library codes.

mpharrigan:
If someone doesn't pass in an `ax`, they're presumably already using matplotlib's stateful mode. Cirq would only use object-y methods. The translation between the two would happen with the one call to `ax = plt.gca()`. 

Off the top of my head, I thought of these libraries which offer plotting, and they follow this idiom:

https://networkx.github.io/documentation/stable/_modules/networkx/drawing/nx_pylab.html#draw
https://github.com/mwaskom/seaborn/blob/c73055b2a9d9830c6fbbace07127c370389d04dd/seaborn/categorical.py#L2953

Strilanc:
Ping, are you aware of libraries that offer plotting following a different idiom? I would strongly prefer to follow existing patterns when possible.

Strilanc:
For now I think the verdict is that we keep plotting as a per-object method. If we find more things that have to be common to the plotting methods, such as e.g. supporting writing to a file in some standard way, then it will make sense to extract a common method.

I am particularly wary of the fact that plotting is something that is important outside of cirq, and so cirq cannot really impose a standard way to do it. The benefits would need to be pretty large to override this (e.g. as they are with approx equal).

pingyeh:
Let me bring your attention to the [Coding Styles](https://matplotlib.org/faq/usage_faq.html#coding-styles) recommended by matplotlib.

It recommends that the figure and axes be created via functions in the `plt` module, then use the object-oriented codes in user's plotting functions, where an `Axes` object, data to be plotted, and keyword arguments (for tuning the look of the plot) are passed in. No calls to `plt` functions from the user function.

We all like to group data to be plotted into classes. The user's plotting functions become a plotting method of the class, so the data args are unnecessary. What remains is exactly the `SupportsPlot` protocol in this PR.

@Strilanc seems to focus on the `cirq.plot`, which is not my focus. I can remove it from the PR. The key is the `SupportsPlot` protocol and removing `plot.show` in plotting methods.

I'm open to naming plot methods `_plot_` or `plot`. The args and behavior are what I really care about.

Strilanc:
Yes, please remove `cirq.plot`. But note that there is no purpose in defining a `SupportsPlot` protocol if there's no corresponding method consuming it, so the protocol should also be removed.

How about we add a `show=False` argument to the plot methods?

pingyeh:
Do you mean something like this?
```python
    def plot(self, ax, show=False, **kwargs):
        ...
        if show:
            ax.get_figure().show()
```
No, I don't want to. Library codes should not call `show()`, as discussed in #1890 and indicated in the matplotlib coding style.

If we agree that all plot methods should take (an `Axes`, a keyword args) and not call `show`, I can drop this PR and start migrating the existing `plot` methods and their call sites throughout the codebase.

mpharrigan:
> Let me bring your attention to the Coding Styles recommended by matplotlib.

TIL! Thanks @pingyeh. I still don't think anyone follows this though... They have a lot of gall to suggest we all "do as they say" :) For fun, I was quickly able to find an instance within matplotlib that uses `if ax is None: ax = gca()`: https://matplotlib.org/3.1.1/_modules/matplotlib/pyplot.html#colorbar . Hilariously, the documentation for this parameter is: "Parent axes from which space for a new colorbar axes will be stolen". Makes me chuckle how far from their own advice it is

pingyeh:
It is a function in `pyplot`, so it's supposed to do that.

We certainly do not follow recommendation blindly. But this recommendation is something I agree with quite strongly for reasons I wrote above, e.g., lack of side effects, easier to test, etc.

Strilanc:
Please drop the global plot method. Write the class instance plot method according to the best practices you outlined.

pingyeh:
> Please drop the global plot method. Write the class instance plot method according to the best practices you outlined.

Sent #2286 for review.

Strilanc:
Obsoleted by the introduction of conventions around plotting methods.

PR #2100, opened 2019-09-12
[Qudits] Add documentation

### LINKED ISSUE #933: Handle Qudits

cduck:
Final part of #933.  Closes #933.

Adds a Qudits section to readthedocs.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -309,10 +309,12 @@ operations, and other types.
     mixture
     mixture_channel
     mul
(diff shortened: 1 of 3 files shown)


mpharrigan:
Congrats @cduck !

Issue #2103, opened 2019-09-12
Allow NoiseModel to substitute and/or add qubits


cduck:
When DensityMatrixSimulator is run with a noise model, the simulator assumes that the noise model only returns operations on the same qubits and crashes otherwise.

There are two main use cases for this:
- Simulating noise caused by leakage to the |2> state.  The noise model needs to return operations on qutrits corresponding to the input circuit's qubits.
- Simulating entanglement with the environment.  The noise model needs to add extra qubit(s) to represent the environment.

I propose adding a method to NoiseModel named something like `substituted_system_qubits` that takes an ordered list of the circuit qubits and returns a list of the qubits the NoiseModel plans to output.  Simulators will need to be updated to account for this.

Related to #933 and #1610.

balopat:
Is the main motivation behind updating the qubits "online" during simulation, instead creating space for them upfront to optimize resource usage? Or is it more of a user experience concern, that you'll have to now add the extra qubits upfront when constructing the circuit?

cduck:
This isn't about updating qubits "online".  The simulators currently take an ideal circuit and a noise model, then simulates the noisy circuit.  This issue here is the density simulator assumes the qubits in the ideal circuit are the same as the noisy circuit.  This limits the transformations a `NoiseModel` can perform on a circuit.

My current workaround for the two use cases I mentioned is to first apply the noise model, then give the noisy circuit to the simulator.

verult:
Marking as after-1.0 assuming only method additions are necessary. cc @95-martin-orion

mhucka:
Closing because of age (6 years since first openened) and prioritization. If this is in error, someone please feel free to reopen it.

PR #2109, opened 2019-09-13
Add Circuit.with_noise

### LINKED ISSUE #1610: Apply noise model to circuit

cduck:
Fixes #1610.

### DIFF
--- a/cirq/circuits/circuit_test.py
+++ b/cirq/circuits/circuit_test.py
@@ -3238,3 +3238,42 @@ def test_json_dict():
         'moments': [cirq.Moment([cirq.CNOT(q0, q1)])],
         'device': cirq.UNCONSTRAINED_DEVICE,
     }
+
+
+def test_with_noise():
+
+    class Noise(cirq.NoiseModel):
+
+        def noisy_operation(self, operation):
+            yield operation
+            if cirq.LineQubit(0) in operation.qubits:
+                yield cirq.H(cirq.LineQubit(0))
+
+    q0, q1 = cirq.LineQubit.range(2)
+    c = cirq.Circuit.from_ops(
+        cirq.X(q0),
+        cirq.Y(q1),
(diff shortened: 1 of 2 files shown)


cirq/circuits/circuit_test.py
```diff
@@ -3238,3 +3238,42 @@ def test_json_dict():
...
+        cirq.Moment([
+            cirq.X(q0),
+        ]),
+        cirq.Moment([
+            cirq.H(q0),
+        ]),
+    ])
+    c_noisy = c.with_noise(Noise())
+    assert str(c_noisy) == str(c_expected)
+    assert c_noisy == c_expected
```
dstrain115: Why are we testing both string equality and equality?

Strilanc: Probably leftover debugging code.

mpharrigan:
Should noisy simulators be modified to use this functionality?

cduck:
@mpharrigan Ideally, yes.  There are two challenges that need to be resolved to do that:
- The intermediate results simulator needs to know what came from the original moments.
- What do you do if the user specifies a qubit order and the noise model substitutes the qubits (#2103)?

cirq/circuits/circuit_test.py
cduck: Fixed.

cduck:
@googlebot Committed from my computer instead of the web editor.

Strilanc:
> Should noisy simulators be modified to use this functionality?

No, because the noisy simulators sometimes have to preserve information about moment structure in a way that this method does not.

PR #2110, opened 2019-09-13
Add device specification proto support

dstrain115:
Initial draft of a SerializableDevice object that can support translation between the device proto in cirq.google.api.v2 and the cirq Device object.

The idea is that we can take a device proto (from either a text file or from an API) and quickly turn it into a Device object that can be used to verify circuits.  It would make it easier to share devices and/or pass them around.

There are a bunch of features still missing, but I wanted to get some early feedback on it to see if people like the approach.


mpharrigan:
Can you motivate why we need a separate class for this? In contrast to a class factory method like `Device.from_device_proto()` that deserializes and constructs an ordinary `cirq.Device`?

dstrain115:
> Can you motivate why we need a separate class for this? In contrast to a class factory method like `Device.from_device_proto()` that deserializes and constructs an ordinary `cirq.Device`?

I am not sure what the correct place is for this, so I am open to suggestions.

However, Device is an abstract class.  Specific devices are supposed to implement it.  So, adding a concrete device class to it would be awkward.  I think you would still need another class for it, but you could have Device.from_device_proto() hook into it.

Also, this is somewhat Google-specific, so I think it belongs in the Google directory, since it is specific to Google protos.  You could imagine other protos, specs, or serializations (such as the JSON serialization) using this for other devices.

mpharrigan:
> Device is an abstract class

Ah, that is a good reason indeed.

I'm mildly uncomfortable with the `Device` class hierarchy already. It's strange that there is a leap from a completely abstract `Device` class to e.g. `XmonDevice` which is *very* specific (limited set of supported gates, experimental constraints part of the validation logic, ...). I guess I just don't expect the *description* of a device to be encoded in a bunch of overriding *methods*. In contrast, the device proto (and  this new class) function as a lovely bag of data that can support all sorts of devices subject to some basic assumptions. So in that sense, I like this a lot.

It's strange that this device class is named after its serializablity while all the other subclasses are named after physics.

I would (personally) love to see a base `Device` class that is not abstract,, but which is modeled loosley off of the `DeviceSpecification` proto. Then you could just have a factory method to create it from an actual proto. Subclasses could still override the validation functions for more complicated constraints (ie those that can't be expressed via a `GateSet`)

dstrain115:
> It's strange that this device class is named after its serializablity while all the other subclasses are named after physics.

I don't totally disagree with you, but I think that's the purpose of this class: to provide basic functionality of device constraints without respect to physics or experiment.

I disagree about Device being a concrete class.  Hard-coding assumptions about the device will erode the utility of this as a base type.

mpharrigan:
I guess I just don't see the utility of the abstract base class as currently written. Instead of asking "is this operation ok", I typically want to know what are all the gates you can do? what different gatesets do you support? how many qubits do you have? what is their connectivity? 

As far as I can tell, in practice people rarely use the device class for its validation features delegating to e.g. proto serialization for gateset validation

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,136 @@
...
+
+        Args:
+            proto: A proto describing the qubits on the device, as well as the
+                supported gates and timing information.
+            gate_set: A SerializableGateSet that can translate the gate_ids
+                into cirq Gates.
+        """
+
+        self.qubits = self._qubits_from_ids(proto.valid_qubits)
+        self.allowed_targets: Dict[str, Set[Tuple]] = dict()
```
mpharrigan: `Set[Tuple[cirq.Qid]]]`?

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,136 @@
...
+
+class SerializableDevice(devices.Device):
+    """Device object generated from a device specification proto.
+
+    Given a device specification proto and a gate_set to translate the
+    serialized gate_ids to cirq Gates, this will generate a Device that can
+    verify operations and circuits for the hardware specified by the device.
+    """
+
+    def __init__(self, proto: device_pb2.DeviceSpecification,
```
mpharrigan: Regardless of the thread on the main discussion, I think you should have the constructor take the python types, i.e. qubits, allowed_targets, duration, target_sets and then factor out the protobuf stuff into a classmethod.

Among other benefits, this would let you document the class attributes which are currently undocumented. For example, I can't really tell what `self.target_sets` is going to be.

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,136 @@
...
+        for gate_type in gate_set.supported_gate_types():
+            for serializer in gate_set.serializers[gate_type]:
+                gate_id = serializer.serialized_gate_id
+                if gate_id not in gate_defs:
+                    raise ValueError(f'Serializer has {gate_id} which is not ' +
+                                     'supported by the device specification')
+                gate_picos = gate_defs[gate_id].gate_duration_picos
+                self.durations[gate_type] = Duration(picos=gate_picos)
+                self.target_sets[gate_type] = gate_defs[gate_id].valid_targets
+        print(self.durations)
```
mpharrigan: remove

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def _qubits_from_ids(self, id_list):
+        """Translates a list of ids in proto format e.g. '4_3'
+        into cirq.GridQubit objects"""
+        # TODO(dstrain): Add support for non-grid qubits
```
mpharrigan: do or open a follow up issue

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,136 @@
...
+        for target in ts.targets:
+            qid_list = self._qubits_from_ids(target.ids)
+            target_set.add(tuple(qid_list))
+            if ts.target_ordering == device_pb2.TargetSet.SYMMETRIC:
+                qid_list.reverse()
+                target_set.add(tuple(qid_list))
+        return target_set
+
+    def _find_operation_type(self, op_key: 'cirq.Operation',
+                             type_dict: Dict[Any, Any]):
```
mpharrigan: can this type annotation be made more specific? the docstring says the keys have `Gate` type

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,136 @@
...
+            raise ValueError(
+                f'Operation does not use valid qubit target: {operation}.')
+
+    def validate_scheduled_operation(
+            self, schedule: 'cirq.Schedule',
+            scheduled_operation: 'cirq.ScheduledOperation') -> None:
+        pass
+
+    def validate_schedule(self, schedule: 'cirq.Schedule') -> None:
+        pass
```
mpharrigan: `raise NotImplementedError`?

cirq/google/serializable_device.py
dstrain115: Done

cirq/google/serializable_device.py
dstrain115: Ok, let me think about the best way to do this.  This is still in progress.

cirq/google/serializable_device.py
dstrain115: Done.

cirq/google/serializable_device.py
dstrain115: Done.

cirq/google/serializable_device.py
dstrain115: Done.

cirq/google/serializable_device.py
dstrain115: I think that would render the class useless, since every schedule would be considered invalid then.  I will add this in a follow-up PR unless I can finish while people are still reviewing.

mpharrigan: got it

cirq/google/known_devices.py
```diff
@@ -15,10 +15,15 @@
 from typing import Dict, List, Set, Tuple
 
 from cirq.devices import GridQubit
+from cirq.google import gate_sets, serializable_gate_set
+from cirq.google.api.v2 import device_pb2
 from cirq.google.xmon_device import XmonDevice
 from cirq.value import Duration
 
 
+TARGET_SET_NAME = "grid"
```
Strilanc: Better name.

cirq/google/known_devices.py
```diff
@@ -47,6 +52,50 @@ def _parse_device(s: str) -> Tuple[List[GridQubit], Dict[str, Set[GridQubit]]]:
...
+    qubits, _ = _parse_device(s)
+    spec = device_pb2.DeviceSpecification()
+    grid_targets = spec.valid_targets.add()
+    grid_targets.name = TARGET_SET_NAME
+    grid_targets.target_ordering = device_pb2.TargetSet.SYMMETRIC
+    qubit_set = frozenset(qubits)
+    neighbor_set: Set[Tuple] = set()
+    spec.valid_qubits.extend([q.proto_id() for q in qubits])
+    for q in qubits:
+        for neighbor in [q + (0, 1), q + (1, 0), q + (-1, 0), q + (0, -1)]:
```
Strilanc: Extract this into a method on `GridQubit`. Also on `LineQubit`.

cirq/google/serializable_device.py
dstrain115: This has been done.

cirq/google/known_devices.py
dstrain115: Done.

cirq/google/known_devices.py
dstrain115: Done.

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,198 @@
...
+
+    Given a device specification proto and a gate_set to translate the
+    serialized gate_ids to cirq Gates, this will generate a Device that can
+    verify operations and circuits for the hardware specified by the device.
+    """
+
+    def __init__(self,
+                 qubits: List['cirq.Qid'],
+                 durations: Dict[Type, Duration],
+                 target_sets: Dict[Type, Set[Tuple['cirq.Qid', ...]]],
```
mpharrigan: thanks for the refactor! Can you document these arguments? Is `Type` the correct type for the keys?

dstrain115: Done.  Added docs and changed to Type['cirq.Gate'] which I think is more specific.

cirq/devices/line_qubit.py
```diff
@@ -40,6 +40,18 @@ def is_adjacent(self, other: ops.Qid) -> bool:
         """Determines if two qubits are adjacent line qubits."""
         return isinstance(other, _BaseLineQid) and abs(self.x - other.x) == 1
 
+    def neighbors(self, qids: Iterable[ops.Qid] = None) -> List['_BaseLineQid']:
+        """Returns qubits that are potential neighbors to this GridQubit
```
kevinsung: Not a GridQubit.

dstrain115: Fixed.

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,220 @@
...
+
+            # A permutation gate can have any combination of qubits
+            valid_qubits = self._find_operation_type(operation,
+                                                     self.target_sets)
+            if valid_qubits == set():
+                # All qubits are valid
+                return
+            for q in operation.qubits:
+                for qubit_tuple in valid_qubits:
+                    if q not in qubit_tuple:
```
maffoo: IIUC this raises an error if there is _any_ qubit tuple that does not contain `q`. I would think we'd instead want something like:
```python
valid_target_sets = self._find_operation_type(...)

# flatten into one set; each target _should_ have only one qubit, per device.proto
valid_qubits = {q for target in valid_target_sets for q in target}

if not all(q in valid_qubits for q in operation.qubits):
    raise ValueError(...)
```

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,220 @@
...
+                        raise ValueError(
+                            'Operation does not use valid qubits: ' +
+                            f'{operation}.')
+            return
+
+        if (len(operation.qubits) > 1):
```
maffoo: nit: unneccessary parens

cirq/google/known_devices.py
```diff
@@ -47,6 +54,64 @@ def _parse_device(s: str) -> Tuple[List[GridQubit], Dict[str, Set[GridQubit]]]:
...
+
+    grid_targets = spec.valid_targets.add()
+    grid_targets.name = _2_QUBIT_TARGET_SET
+    grid_targets.target_ordering = device_pb2.TargetSet.SYMMETRIC
+    qubit_set = frozenset(qubits)
+    neighbor_set: Set[Tuple] = set()
+    spec.valid_qubits.extend([q.proto_id() for q in qubits])
+    for q in qubits:
+        for neighbor in q.neighbors(qubit_set):
+            if tuple((neighbor, q)) not in neighbor_set:
```
maffoo: nit: use literal tuple
```python
if (neighbor, q) not in neighbor_set:
```

cirq/google/known_devices.py
```diff
@@ -47,6 +54,64 @@ def _parse_device(s: str) -> Tuple[List[GridQubit], Dict[str, Set[GridQubit]]]:
...
+    meas_targets = spec.valid_targets.add()
+    meas_targets.name = _MEAS_TARGET_SET
+    meas_targets.target_ordering = device_pb2.TargetSet.SUBSET_PERMUTATION
+
+    grid_targets = spec.valid_targets.add()
+    grid_targets.name = _2_QUBIT_TARGET_SET
+    grid_targets.target_ordering = device_pb2.TargetSet.SYMMETRIC
+    qubit_set = frozenset(qubits)
+    neighbor_set: Set[Tuple] = set()
+    spec.valid_qubits.extend([q.proto_id() for q in qubits])
```
maffoo: nit: don't need square brackets; `extend` can take a generator expression

cirq/google/known_devices.py
```diff
@@ -47,6 +54,64 @@ def _parse_device(s: str) -> Tuple[List[GridQubit], Dict[str, Set[GridQubit]]]:
...
+    for q in qubits:
+        for neighbor in q.neighbors(qubit_set):
+            if tuple((neighbor, q)) not in neighbor_set:
+                # Don't add pairs twice
+                new_target = grid_targets.targets.add()
+                new_target.ids.extend([q.proto_id(), neighbor.proto_id()])
+                neighbor_set.add(tuple((q, neighbor)))
```
maffoo:
```python
neighbor_set.add((q, neighbor))
```

cirq/google/known_devices.py
```diff
@@ -47,6 +54,64 @@ def _parse_device(s: str) -> Tuple[List[GridQubit], Dict[str, Set[GridQubit]]]:
...
+                    # Only add each type once
+                    continue
+
+                gate_ids.add(gate_id)
+                gate = gs_proto.valid_gates.add()
+                gate.id = gate_id
+                if gate_type == MeasurementGate:
+                    gate.valid_targets.extend([_MEAS_TARGET_SET])
+                else:
+                    gate.valid_targets.extend([_2_QUBIT_TARGET_SET])
```
maffoo: Should only do this for 2-qubit gates, not single-qubit gates. Can we tell the gate arity from `SerializableGateSet`?

cirq/google/known_devices.py
```diff
@@ -47,6 +54,64 @@ def _parse_device(s: str) -> Tuple[List[GridQubit], Dict[str, Set[GridQubit]]]:
...
+                if durations is not None and gate.id in durations:
+                    gate.gate_duration_picos = durations[gate.id]
+                # TODO: add valid args and number of qubits
```
maffoo: I wonder if it would be better to just generate `.pbtxt` versions of the descriptions for existing known devices and check those it, rather than try to generate these from grids where we don't have all the needed information (like gate arity and args).

cirq/devices/grid_qubit.py
```diff
@@ -50,6 +50,18 @@ def is_adjacent(self, other: ops.Qid) -> bool:
         return (isinstance(other, GridQubit) and
                 abs(self.row - other.row) + abs(self.col - other.col) == 1)
 
+    def neighbors(self, qids: Iterable[ops.Qid] = None) -> List['GridQubit']:
```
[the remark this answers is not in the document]
dstrain115: Done

cirq/devices/grid_qubit_test.py
```diff
@@ -114,6 +114,21 @@ def test_grid_qubit_is_adjacent():
     assert not cirq.GridQubit(500, 999).is_adjacent(cirq.GridQubit(5034, 999))
 
 
+def test_grid_qubit_neighbors():
+    expected = set([
+        cirq.GridQubit(1, 2),
+        cirq.GridQubit(2, 1),
+        cirq.GridQubit(0, 1),
+        cirq.GridQubit(1, 0)
+    ])
```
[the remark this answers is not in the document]
dstrain115: Done

cirq/google/known_devices_test.py
```diff
@@ -0,0 +1,204 @@
...
+def test_foxtail_qubits():
+    expected_qubits = []
+    for i in range(0, 2):
+        for j in range(0, 11):
+            expected_qubits.append(GridQubit(i, j))
+    assert set(expected_qubits) == known_devices.Foxtail.qubits
+
+
+def test_foxtail_device_proto():
+    assert str(known_devices.FOXTAIL_PROTO) == """valid_gate_sets {
```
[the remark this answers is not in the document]
dstrain115: Done

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,220 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Device object for converting from device specification protos"""
+from typing import Any, cast, Dict, List, Set, Tuple, Type, TYPE_CHECKING
+
+from cirq import devices, ops
+from cirq.google import serializable_gate_set
+from cirq.google.api.v2 import device_pb2
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,220 @@
...
+            permutation_gates: A list of types that act on all permutations
+                of the qubit targets.  (e.g. measurement gates)
+        """
+        self.qubits = qubits
+        self.durations = durations
+        self.target_sets = target_sets
+        self.is_permutation_gate: Dict[Type['cirq.Gate'], bool] = dict()
+        if permutation_gates is not None:
+            for gate in permutation_gates:
+                self.is_permutation_gate[gate] = True
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/known_devices.py
dstrain115: Done.

cirq/google/known_devices.py
dstrain115: Done.

cirq/google/known_devices.py
dstrain115: Done.

cirq/google/known_devices.py
dstrain115: I have done that by checking whether it is a subclass of SingleQubitGate.

cirq/google/known_devices.py
dstrain115: Yeah, that's my plan.  ... However, how are we going to generate the .pbtxt versions?

The goal is that this (or something similar to it, will help to generate the .pbtxt version.  I certainly don't want to do it by hand.

I think, with the serializers and the gate set, we have the necessary information (for the use cases we have anyway).w

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def _find_operation_type(self, op_key: 'cirq.Operation',
+                             type_dict: Dict[Type['cirq.Gate'], Any]) -> Any:
+        """Finds the type (or a compatible type) of an operation from within
+        a dictionary with keys of Gate type.
+
+        Returns:
+             the value corresponding to that key or None if no type matches
+        """
+        gate_op = cast(ops.GateOperation, op_key)
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,220 @@
...
+    def _create_target_set(ts: device_pb2.TargetSet
+                          ) -> Set[Tuple['cirq.Qid', ...]]:
+        """Transform a TargetSet proto into a set of qubit tuples"""
+        target_set = set()
+        for target in ts.targets:
+            qid_list = SerializableDevice._qubits_from_ids(target.ids)
+            target_set.add(tuple(qid_list))
+            if ts.target_ordering == device_pb2.TargetSet.SYMMETRIC:
+                qid_list.reverse()
+                target_set.add(tuple(qid_list))
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,220 @@
...
+                if gate_id not in gate_defs:
+                    raise ValueError(f'Serializer has {gate_id} which is not ' +
+                                     'supported by the device specification')
+                gate_ts = gate_defs[gate_id].valid_targets
+                which_are_permutations = [t in permutation_ids for t in gate_ts]
+                if any(which_are_permutations):
+                    if not all(which_are_permutations):
+                        raise NotImplementedError('Cannot currently mix ' +
+                                                  'SUBSET_PERMUTATION with ' +
+                                                  'other target sets')
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,220 @@
...
+        durations: Dict[Type['cirq.Gate'], Duration] = dict()
+        target_sets: Dict[Type['cirq.Gate'], Set[
+            Tuple['cirq.Qid', ...]]] = dict()
+        permutation_gates: List[Type['cirq.Gate']] = list()
+        for gate_type in gate_set.supported_gate_types():
+            for serializer in gate_set.serializers[gate_type]:
+                gate_id = serializer.serialized_gate_id
+                if gate_id not in gate_defs:
+                    raise ValueError(f'Serializer has {gate_id} which is not ' +
+                                     'supported by the device specification')
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,220 @@
...
+                supported gates and timing information.
+            gate_set: A SerializableGateSet that can translate the gate_ids
+                into cirq Gates.
+        """
+
+        # Store target sets, since they are refered to by name later
+        allowed_targets: Dict[str, Set[Tuple['cirq.Qid', ...]]] = dict()
+        permutation_ids: List[str] = list()
+        for ts in proto.valid_targets:
+            allowed_targets[ts.name] = SerializableDevice._create_target_set(ts)
```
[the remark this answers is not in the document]
dstrain115: I changed the staticmethods that call other methods in the class, but kept static ones that are self-contained.  Let me know if you think we should convert all of them.

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,220 @@
...
+
+        Args:
+            proto: A proto describing the qubits on the device, as well as the
+                supported gates and timing information.
+            gate_set: A SerializableGateSet that can translate the gate_ids
+                into cirq Gates.
+        """
+
+        # Store target sets, since they are refered to by name later
+        allowed_targets: Dict[str, Set[Tuple['cirq.Qid', ...]]] = dict()
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,220 @@
...
+        target_set = set()
+        for target in ts.targets:
+            qid_list = SerializableDevice._qubits_from_ids(target.ids)
+            target_set.add(tuple(qid_list))
+            if ts.target_ordering == device_pb2.TargetSet.SYMMETRIC:
+                qid_list.reverse()
+                target_set.add(tuple(qid_list))
+        return target_set
+
+    def _find_operation_type(self, op_key: 'cirq.Operation',
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,220 @@
...
+        for q in operation.qubits:
+            if q not in self.qubits:
+                raise ValueError('Qubit not on device: {!r}'.format(q))
+
+        if self._find_operation_type(operation,
+                                     self.is_permutation_gate) is not None:
+
+            # A permutation gate can have any combination of qubits
+            valid_qubits = self._find_operation_type(operation,
+                                                     self.target_sets)
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,220 @@
...
+            if isinstance(gate_key, type_key):
+                return type_dict[type_key]
+        return None
+
+    def duration_of(self, operation: 'cirq.Operation') -> Duration:
+        duration = self._find_operation_type(operation, self.durations)
+        if duration is not None:
+            return duration
+        raise ValueError(
+            f'Operation {operation} does not have a duration listed')
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/serializable_device.py
dstrain115: Done.

cirq/google/serializable_device.py
dstrain115: Done.

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,238 @@
...
+            target_sets=target_sets,
+            permutation_gates=permutation_gates,
+        )
+
+    @staticmethod
+    def _qid_from_str(id_str: str) -> 'cirq.Qid':
+        try:
+            return devices.GridQubit.from_proto_id(id_str)
+        except ValueError:
+            return ops.NamedQubit(id_str)
```
Strilanc: This fallback behavior can create some very confusing issues. For example, suppose I put a LineQubit in somewhere and it gets turned into "1". When it gets parsed here, I'll get a NamedQubit("1") != LineQubit(1) out.

mpharrigan: now you understand why my json schema is so verbose :)

dstrain115: I'm okay with this behavior.  LineQubits are probably not going to be serialized as part of Google's devices (and this is in the google directory).  This is further reinforced by the fact that LineQubit's don't have proto_id() or from_proto_id() functions.

I did put pydoc with description that explicitly only supports GridQubit and NamedQubit.

cirq/devices/line_qubit.py
```diff
@@ -40,6 +40,19 @@ def is_adjacent(self, other: ops.Qid) -> bool:
         """Determines if two qubits are adjacent line qubits."""
         return isinstance(other, _BaseLineQid) and abs(self.x - other.x) == 1
 
+    def neighbors(self, qids: Optional[Iterable[ops.Qid]] = None
+                 ) -> Set['_BaseLineQid']:
```
maffoo: This is fine for now, but could use a self-type to make this more precise. Something like:
```python
T = TypeVar('T', bound='_BaseLineQid', covariant=True)

class _BaseLineQid(ops.Qid):
    ...
    def neighbors(self: T, ...) -> Set[T]:
        ...
```

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,245 @@
...
+                    target_sets[gate_type] = set()
+                for target_set_name in gate_defs[gate_id].valid_targets:
+                    target_sets[gate_type] |= allowed_targets[target_set_name]
+        for gate_type in target_sets:
+            gate_definitions[gate_type] = _GateDefinition(
+                duration=durations[gate_type],
+                target_set=target_sets[gate_type],
+                is_permutation=(permutation_gates is not None and
+                                gate_type in permutation_gates),
+                number_of_qubits=gate_defs[gate_id].number_of_qubits)
```
maffoo: It looks like you should be able to construct all the `_GateDefinition` objects just from the protos, and put them in a map indexed by gate_id. Then loop over serializers and create a new map indexed by the cirq gate type. This should avoid the need for the `durations`, `target_sets`, and `permutation_gates` temporaries.

dstrain115: Not sure how much clearer it is (maybe slightly simpler), but it's done.

cirq/google/serializable_device.py
```diff
@@ -0,0 +1,245 @@
...
+            # A permutation gate can have any combination of qubits
+
+            if gate_def.target_set == set():
+                # All qubits are valid
+                return
+
+            # flattens all qubits in all targets into one set
+            flattened_qubits = {
+                q for qubit_tuple in gate_def.target_set for q in qubit_tuple
+            }
```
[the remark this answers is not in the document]
dstrain115: Good idea.  Done.

cirq/google/known_devices.py
```diff
@@ -80,6 +176,17 @@ def _json_dict_(self):
     qubits=_parse_device(_FOXTAIL_GRID)[0])
 
 
+# Duration dict in picoseconds
+_DURATIONS_FOR_XMON = {
+    'exp_11': 50000,
+    'exp_w': 20000,
+    'exp_z': 0,
+    'meas': 1000000,
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/known_devices.py
```diff
@@ -47,6 +54,95 @@ def _parse_device(s: str) -> Tuple[List[GridQubit], Dict[str, Set[GridQubit]]]:
     return qubits, measurement_lines
 
 
+def create_device_proto_from_diagram(
+        s: str,
+        gate_set: serializable_gate_set.SerializableGateSet = None,
+        durations: Dict[str, int] = None,
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/known_devices_test.py
```diff
@@ -11,10 +11,224 @@
...
+def test_foxtail_qubits():
+    expected_qubits = []
+    for i in range(0, 2):
+        for j in range(0, 11):
+            expected_qubits.append(cirq.GridQubit(i, j))
+    assert set(expected_qubits) == cirq.google.known_devices.Foxtail.qubits
+
+
+def test_foxtail_device_proto():
+    print(str(cirq.google.known_devices.FOXTAIL_PROTO))
```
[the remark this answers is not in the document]
dstrain115: Done.

dabacon:
> It's strange that this device class is named after its serializablity while all the other subclasses are named after physics.

This is completely legacy and I think XmonDevice should be removed and replaced by a better structure (including this one).

Issue #2111, opened 2019-09-15
Implement Quantum Volume algorithm


KevinVillela:
Spoke with @mpharrigan about this. It'd be nice to have an implementation of the Quantum Volume algorithm described at https://arxiv.org/abs/1811.12926. Note that OpenQASM has the first part of this, [creating the model circuit](https://github.com/Qiskit/openqasm/blob/5fb61893661261b7211942dc04ee85f8fd845f2a/benchmarks/quantum_volume/quantum_volume.py).

If someone could assign this issue to me, I'd greatly appreciate it :)

vtomole:
The referring pull request has been merged. Please reopen if it's not done yet @mpharrigan @KevinVillela .

vtomole:
Sorry! Only the `HOG` portion is done.

KevinVillela:
Yup, I'm working on the next part now :)

KevinVillela:
I believe this is done now, anyone can feel free to re-open and assign to me if they see something I missed. Thanks!

Issue #2020, opened 2019-08-23
Use np.random.RandomState for simulator seeds.


dabacon:
see comments in #2017

mpharrigan:
See discussion on https://github.com/quantumlib/Cirq/pull/2112#discussion_r325297470 for examples of this idiom in scikit-learn and pandas

PR #2112, opened 2019-09-16
Implement Quantum Volume HOG portion

KevinVillela:
See #2111 for details. After this, we will implement the model circuit on our devices and run them to calculate their Quantum Volume.

First PR so let me know what I messed up on. Don't hold back :)


KevinVillela:
@googlebot I signed it!

KevinVillela:
@googlebot I fixed it.

KevinVillela:
@googlebot I consent

dev_tools/profiling/quantum_volume.py
```diff
@@ -0,0 +1,144 @@
...
+
+    Args:
+        num_qubits: The number of qubits in the generated circuit.
+        depth: The number of layers in the circuit.
+
+    Returns:
+      The generated circuit.
+    """
+    # Setup the circuit and its qubits.
+    qubits = [cirq.GridQubit(0, i) for i in range(num_qubits)]
```
mpharrigan: ~cirq.LineQubit.range(num_qubits)`

dev_tools/profiling/quantum_volume.py
```diff
@@ -0,0 +1,144 @@
...
+
+    # For each layer.
+    for _ in range(depth):
+        # Generate uniformly random permutation Pj of [0...n-1]
+        perm = np.random.permutation(num_qubits)
+
+        # For each consecutive pair in Pj, generate Haar random SU(4)
+        # Decompose each SU(4) into CNOT + SU(2) and add to Ci
+        for k in range(math.floor(num_qubits / 2)):
+            permuted_qubits = [int(perm[2 * k]), int(perm[2 * k + 1])]
```
mpharrigan: `permuted_indices` (since they're not `cirq.Qid` objects)

dev_tools/profiling/quantum_volume.py
```diff
@@ -0,0 +1,144 @@
...
+      The generated circuit.
+    """
+    # Setup the circuit and its qubits.
+    qubits = [cirq.GridQubit(0, i) for i in range(num_qubits)]
+    circuit = cirq.Circuit()
+
+    # For each layer.
+    for _ in range(depth):
+        # Generate uniformly random permutation Pj of [0...n-1]
+        perm = np.random.permutation(num_qubits)
```
mpharrigan: This function should let you supply a random seed to replicability. The best way (IMO) of doing this is with the following pattern:

```python
def my_function(..., rs=None):
  if not isinstance(rs, np.random.RandomState):
    rs = np.random.RandomState(rs)
  ...
  perm = rs.permutation(num_qubits)
```

This lets you pass either (1) a numeric seed, (2) an existing `RandomState` so you can compose this function with other stochastic functions (3) `None`, which will seed according to dev/urandom

dev_tools/profiling/quantum_volume.py
```diff
@@ -0,0 +1,144 @@
...
+
+        # For each consecutive pair in Pj, generate Haar random SU(4)
+        # Decompose each SU(4) into CNOT + SU(2) and add to Ci
+        for k in range(math.floor(num_qubits / 2)):
+            permuted_qubits = [int(perm[2 * k]), int(perm[2 * k + 1])]
+            special_unitary = cirq.testing.random_special_unitary(4)
+            kak_unitary = cirq.unitary(cirq.kak_decomposition(special_unitary))
```
mpharrigan: This line seems unnecessary. You take the unitary, decompose it, but then turn it right back into the original unitary.

dev_tools/profiling/quantum_volume.py
```diff
@@ -0,0 +1,144 @@
...
+
+            # Convert the decomposed unitary to Cirq operations and add them to
+            # the circuit.
+            ops = cirq.two_qubit_matrix_to_operations(
+                qubits[permuted_qubits[0]], qubits[permuted_qubits[1]],
+                kak_unitary, False)
```
mpharrigan: use a keyword argument for `False` so the reader (i.e. me) knows what flag you're setting

dev_tools/profiling/quantum_volume.py
```diff
@@ -0,0 +1,144 @@
...
+            circuit.append(ops)
+
+    # Measure all of the qubits at the end of the circuit.
+    circuit.append([cirq.measure(qubit) for qubit in qubits])
```
mpharrigan: `cirq.measure(*qubits)` will create one measurement operation on all qubits

dev_tools/profiling/quantum_volume.py
```diff
@@ -0,0 +1,144 @@
...
+        seed: A seed to pass to the simulator.
+
+    Returns:
+        A list containing all of the heavy bit-string results.
+    """
+    # Run the simulation 100 times, storing the a dict from resulting bit-string
+    # to the number of times that bit-string came up.
+    results: DefaultDict[str, int] = defaultdict(int)
+    for _ in range(100):
+        simulator = cirq.Simulator(seed=seed)
```
mpharrigan: You re-seed each time! This is likely not what you want at all

dev_tools/profiling/quantum_volume.py
```diff
@@ -0,0 +1,144 @@
...
+        circuit: The circuit to classically simulate.
+        seed: A seed to pass to the simulator.
+
+    Returns:
+        A list containing all of the heavy bit-string results.
+    """
+    # Run the simulation 100 times, storing the a dict from resulting bit-string
+    # to the number of times that bit-string came up.
+    results: DefaultDict[str, int] = defaultdict(int)
+    for _ in range(100):
```
mpharrigan: `simulator.run` takes a repetitions argument which will be *much* faster than this for-loop.

dev_tools/profiling/quantum_volume.py
```diff
@@ -0,0 +1,144 @@
...
+        simulator = cirq.Simulator(seed=seed)
+        result = simulator.run(circuit)
+        bit_string = "".join(str(i) for i in result.data.iloc[0])
+        results[bit_string] += 1
+
+    # Compute the median probability of the output bit-strings.
+    median = np.median(list(results.values()))
```
mpharrigan: This is not the median you want. 

Calculating hog requires one call to `simulator.simulate()` (which can only be done classically with exponential cost). This will give you a vector of 2**n_qubits probabilities, of which you can find the median. 

*Then* you can call `sampler.run()` for a number of repetitions. For each bitstring you sample, you can look up its true probability and figure out if it's higher than the true median

dev_tools/profiling/quantum_volume.py
KevinVillela: Done.

dev_tools/profiling/quantum_volume.py
KevinVillela: D'oh, and two_qubit_matrix_to_operations decomposes it as well :) Done.

dev_tools/profiling/quantum_volume.py
KevinVillela: Done.

dev_tools/profiling/quantum_volume.py
KevinVillela: Nice! Done.

dev_tools/profiling/quantum_volume.py
KevinVillela: Ah of course, thanks for catching that. Done, PTAL.

dev_tools/profiling/quantum_volume.py
KevinVillela: Not used anymore, but I'll keep it in mind for later, thanks!

dev_tools/profiling/quantum_volume.py
KevinVillela: Woops, done.

dev_tools/profiling/quantum_volume.py
KevinVillela: Done, had to modify a few other files for full replicability.

KevinVillela:
Thanks for the thorough review, I believe I have addressed all comments, PTAL :)

examples/quantum_volume.py
```diff
@@ -0,0 +1,153 @@
...
+    consider the quantum volume test passed for that size.
+
+    Args:
+        num_qubits: The number of qubits for the circuit.
+        depth: The number of gate layers to generate.
+        num_repetitions: The number of times to run the algorithm.
+    """
+    for _ in range(num_repetitions):
+        model_circuit = generate_model_circuit(num_qubits, depth)
+        print(model_circuit)
```
Strilanc: Leftover debug code?

examples/quantum_volume.py
```diff
@@ -0,0 +1,153 @@
...
+    # Compute the median probability of the output bit-strings.
+    median = statistics.median(results.state_vector())
+
+    # The output wave function is a vector from the result value (big-endian) to
+    # the probability of that bit-string. Compute a format string converts the
+    # given index to the corresponding qubit string.
+    format_str = '{0:0%sb}' % len(circuit.all_qubits())
+    # Return all of the bit-strings that have a probability greater than the
+    # median.
+    return ([
```
Strilanc: Nit: unnecessary parens

Strilanc: Is the quantum computer allowed to count the same bit string multiple times in this test?

examples/quantum_volume.py
```diff
@@ -0,0 +1,153 @@
...
+            special_unitary = cirq.testing.random_special_unitary(4)
+
+            # Convert the decomposed unitary to Cirq operations and add them to
+            # the circuit.
+            ops = cirq.two_qubit_matrix_to_operations(
+                qubits[permuted_indices[0]],
+                qubits[permuted_indices[1]],
+                special_unitary,
+                allow_partial_czs=False)
+            circuit.append(ops)
```
Strilanc: You can directly append the function output into the circuit.

examples/quantum_volume.py
```diff
@@ -0,0 +1,153 @@
...
+    parser = argparse.ArgumentParser('Quantum volume benchmark.')
+    parser.add_argument('--num_qubits',
+                        default=4,
+                        type=int,
+                        help='The number of circuit qubits to benchmark.')
+    parser.add_argument('--depth',
+                        default=4,
+                        type=int,
+                        help='SU(4) circuit depth.')
+    parser.add_argument(
```
Strilanc: If you are going to have a seed, take it as an argument.

examples/quantum_volume.py
```diff
@@ -0,0 +1,153 @@
...
+    # For each layer.
+    for _ in range(depth):
+        # Generate uniformly random permutation Pj of [0...n-1]
+        perm = rs.permutation(num_qubits)
+
+        # For each consecutive pair in Pj, generate Haar random SU(4)
+        # Decompose each SU(4) into CNOT + SU(2) and add to Ci
+        for k in range(math.floor(num_qubits / 2)):
+            permuted_indices = [int(perm[2 * k]), int(perm[2 * k + 1])]
+            special_unitary = cirq.testing.random_special_unitary(4)
```
Strilanc: If you are going to have a seed, pass it along. And add a test asserting that the behavior is consistent.

examples/quantum_volume.py
```diff
@@ -0,0 +1,153 @@
...
+        rs = np.random.RandomState(rs)
+
+    # For each layer.
+    for _ in range(depth):
+        # Generate uniformly random permutation Pj of [0...n-1]
+        perm = rs.permutation(num_qubits)
+
+        # For each consecutive pair in Pj, generate Haar random SU(4)
+        # Decompose each SU(4) into CNOT + SU(2) and add to Ci
+        for k in range(math.floor(num_qubits / 2)):
```
Strilanc: `for k in range(0, num_qubits - 1, 2)`

examples/quantum_volume.py
```diff
@@ -0,0 +1,153 @@
...
+    median probability of being generated.
+
+    Args:
+        circuit: The circuit to classically simulate.
+
+    Returns:
+        A list containing all of the heavy bit-string results.
+    """
+    # Classically compute the probabilities of each output bit-string through
+    # simulation.
```
Strilanc: So this example is testing this concept of quantum volume in the abstract, without any geometric constraints on what qubits can talk to each other, yeah? Not sure if there's a nice way to fix that.

Strilanc:
Thanks for the PR! I looked it over and left comments.

examples/quantum_volume.py
```diff
@@ -0,0 +1,153 @@
...
+import sys
+from typing import List, cast
+
+import numpy as np
+
+import cirq
+
+
+def generate_model_circuit(num_qubits: int, depth: int,
+                           rs=None) -> cirq.Circuit:
```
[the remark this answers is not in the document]
mpharrigan: Is there a place where we can put code that is not library functionality per se but is more advanced than an introductory example?

We definitely want seeding for this project

examples/quantum_volume.py
```diff
@@ -0,0 +1,153 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tool to run the Quantum Volume benchmark defined by IBM in
+https://arxiv.org/abs/1811.12926."""
+
+import argparse
+import math
+import statistics
+import sys
+from typing import List, cast
```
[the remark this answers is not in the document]
mpharrigan: Same question as below: Is there a place where we can put code that is not library functionality per se but is more advanced than an introductory example?

cirq/testing/lin_alg_utils.py
```diff
@@ -36,11 +36,12 @@ def random_superposition(dim: int) -> np.ndarray:
     return state_vector
 
 
-def random_unitary(dim: int) -> np.ndarray:
+def random_unitary(dim: int, rs=None) -> np.ndarray:
```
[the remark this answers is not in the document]
mpharrigan: How about we change it to `random_state` but still allow it to do double duty. This is how it's done in scikit-learn (e.g. https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html) and I like it quite a bit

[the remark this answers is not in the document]
mpharrigan: and pandas, e.g. https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sample.html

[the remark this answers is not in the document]
mpharrigan: related: https://github.com/quantumlib/Cirq/issues/2020

examples/quantum_volume.py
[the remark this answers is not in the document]
vtomole: We could sub-directory the examples into `beginner` and `advanced`.

cirq/testing/lin_alg_utils.py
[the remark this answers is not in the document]
KevinVillela: Changed to random_state per comments below.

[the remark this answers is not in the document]
KevinVillela: Done.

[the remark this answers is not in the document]
KevinVillela: Changed to random_state for now, I also think that's cleaner than having two args and requiring only one to be passed in.

examples/quantum_volume.py
```diff
@@ -0,0 +1,153 @@
+# Copyright 2019 The Cirq Developers
```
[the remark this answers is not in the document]
KevinVillela: Done.

examples/quantum_volume.py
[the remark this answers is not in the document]
KevinVillela: Done, and kept the type annotations - I think they help with readability and I'd argue they make sense in this new advanced dir. WDYT?

examples/quantum_volume.py
KevinVillela: Done.

examples/quantum_volume.py
KevinVillela: Done.

examples/quantum_volume.py
KevinVillela: Woops, good catch. Passed along, and there is now a test verifying this behavior.

cirq/testing/lin_alg_utils.py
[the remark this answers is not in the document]
KevinVillela: Done.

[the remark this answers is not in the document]
KevinVillela: Done.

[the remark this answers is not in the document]
KevinVillela: Went with @mpharrigan's suggestion above.

examples/quantum_volume.py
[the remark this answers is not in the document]
KevinVillela: Done.

examples/quantum_volume.py
KevinVillela: Done.

examples/quantum_volume.py
KevinVillela: Yup, this sort of gap is lamented in https://arxiv.org/pdf/1902.08091.pdf: "the circuit model allows multi-qubit gates to act on any qubits without restriction"... The next step, after Heavy Output Generation, is to compile this ideal model circuit into a circuit that can run our architecture.

KevinVillela:
Thanks @mpharrigan, @Strilanc, and @vtomole for the comments! Made the requested changes.

cirq/testing/lin_alg_utils.py
```diff
@@ -36,11 +36,35 @@ def random_superposition(dim: int) -> np.ndarray:
     return state_vector
 
 
-def random_unitary(dim: int) -> np.ndarray:
+def get_seeded_state(random_state: Union[int, np.random.RandomState]
```
Strilanc: I don't think this method belongs in `cirq.testing`. The names also aren't really descriptive of what's happening.

Instead of doing this, just have methods take an `Optional[np.random.RandomState]` instead of an `int`. It doesn't really make sense to pass an integer seed into `random_unitary`, that would be like seeding `randint`. You seed the generator, not the value-generating method.

Furthermore, when no `np.random.RandomState` is provided you should not be defaulting to one seeded by `None`. This will result in every call to `random_unitary` producing a new RandomState. If they happen to be seeding from the clock by default, this could result in all random unitaries being equal over the span of a millisecond. You should instead be falling back to unseeded methods.

Basically, the details of how the randomness are seeded have to be handled on a method by method basis, so you can't extract out the logic like this.

examples/advanced/quantum_volume.py
```diff
@@ -0,0 +1,147 @@
+"""Tool to run the Quantum Volume benchmark defined by IBM in
+https://arxiv.org/abs/1811.12926."""
```
Strilanc: Include usage instructions and example output.

examples/advanced/quantum_volume.py
```diff
@@ -0,0 +1,147 @@
...
+    permutations followed by random two-qubit gates that are sampled from the
+    Haar measure on SU(r).
+
+    Args:
+        num_qubits: The number of qubits in the generated circuit.
+        depth: The number of layers in the circuit.
+        random_state: A way to seed the RandomState.
+
+    Returns:
+      The generated circuit.
```
Strilanc: Four spaces

examples/advanced/quantum_volume.py
```diff
@@ -0,0 +1,147 @@
...
+    # Compute the median probability of the output bit-strings.
+    median = statistics.median(results.state_vector())
+
+    # The output wave function is a vector from the result value (big-endian) to
+    # the probability of that bit-string. Compute a format string converts the
+    # given index to the corresponding qubit string.
+    format_str = '{0:0%sb}' % len(circuit.all_qubits())
+    # Return all of the bit-strings that have a probability greater than the
+    # median.
+    return ([
```
Strilanc: Still unnecessary parens

cirq/testing/lin_alg_utils.py
Strilanc: I see that @mpharrigan asked for the seed+state double duty because it matches what happens in scikit and pandas. I guess that's fine. But it is important to not constant produce new generators when getting None; that's not what happens in the other libraries I think.

KevinVillela: Done, thanks for the explanation! How's this?

examples/advanced/quantum_volume.py
KevinVillela: Done, although this will be updated as the program is completed.

examples/advanced/quantum_volume.py
KevinVillela: Done.

examples/advanced/quantum_volume.py
KevinVillela: Done.

KevinVillela:
'Nother round :)

examples/quantum_volume.py
mpharrigan: FWIW the lack of geometric constraints is supposed to be an attractive feature of the quantum volume. It allows "fair" comparison among quantum computers with different connectivities. That is, you get bonus points if your hardware is more connected.

In practice, actual algorithms have more structure which can be exploited so qvol is very pessimistic

cirq/testing/lin_alg_utils.py
mpharrigan: @Strilanc 's concerns are good ones. I don't think the latest changes address the crux of it though. We want to use the global state if an existing state is omitted. Like this person:

https://stackoverflow.com/questions/41985484/how-to-access-numpy-default-global-random-number-generator

You would think this would be easier....

Unlike the poster, we probably won't have to worry about pickling these functions so you can probably get away with

```python
if random_state is None:
  random_state = np.random
```

or you can use the answer of pulling out `random_state = numpy.random.random.__self__`.

I don't know which is less objectionable in the eyes of the python gods.

examples/advanced/quantum_volume.py
```diff
@@ -0,0 +1,156 @@
...
+    median = statistics.median(results.state_vector())
+
+    # The output wave function is a vector from the result value (big-endian) to
+    # the probability of that bit-string. Compute a format string converts the
+    # given index to the corresponding qubit string.
+    format_str = '{0:0%sb}' % len(circuit.all_qubits())
+    # Return all of the bit-strings that have a probability greater than the
+    # median.
+    return [
+        format_str.format(idx)
```
mpharrigan: We (or at least I) generally don't like turning bit-strings into character-strings. Typically we leave them as 2d numpy integer arrays

```python
hog_bitstrings = np.array([
  [1, 0, 1, 1],
  [0, 0, 0, 0],
  ...
])
```

In this case, it might be more useful to leave them as integers (ie `idx`) to make the set-testing easier when you draw samples and want to know if they're in the HOG set. Up to you

examples/advanced/quantum_volume.py
```diff
@@ -0,0 +1,156 @@
...
+        for idx, prob in enumerate(results.state_vector())
+        if prob > median
```
mpharrigan: np.abs(amp**2)

examples/advanced/quantum_volume.py
```diff
@@ -0,0 +1,156 @@
...
+def generate_model_circuit(num_qubits: int,
+                           depth: int,
+                           *,
+                           random_state: Optional[np.random.RandomState] = None
+                          ) -> cirq.Circuit:
+    """Generates a model circuit with the given number of qubits and depth.
+
+    The generated circuit consists of `depth` layers of random qubit
+    permutations followed by random two-qubit gates that are sampled from the
+    Haar measure on SU(r).
```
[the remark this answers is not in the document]
KevinVillela: Done.

examples/advanced/quantum_volume.py
```diff
@@ -0,0 +1,156 @@
...
+    Returns:
+        A list containing all of the heavy bit-string results.
+    """
+    # Classically compute the probabilities of each output bit-string through
+    # simulation.
+    simulator = cirq.Simulator()
+    results = cast(cirq.WaveFunctionTrialResult,
+                   simulator.simulate(program=circuit))
+    # Compute the median probability of the output bit-strings.
+    median = statistics.median(results.state_vector())
```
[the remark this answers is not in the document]
KevinVillela: Makes sense, thanks for the explanation! My test actually should have caught this, it indeed has a negative amplitude in it. After this change, that value shows up in the test.

examples/advanced/quantum_volume.py
KevinVillela: Sure, makes sense. Done.

examples/advanced/quantum_volume.py
KevinVillela: Done.

cirq/testing/lin_alg_utils.py
KevinVillela: Gotcha. Messing with __self__ looks a bit ugly, I went with the first option.

KevinVillela:
Thanks all, sorry for the errors with the numpy RNG stuff, I understand what's going on now. PTAL.

Strilanc:
Let me get a snapshot of the PR state for debugging before you merge

mpharrigan:
For sure

Strilanc:
This is so strange, the reviews reply has no `"state": "APPROVED"` anywhere. It doesn't include any comment mentioning "handling the RNG stuff", which was your approving comment.

See:
https://api.github.com/repos/quantumlib/cirq/pulls/2112/reviews
https://api.github.com/repos/quantumlib/cirq/pulls/2112

Strilanc:
It's as if the review results are just missing a bunch of relevant data.

Strilanc:
Maybe it's truncating the results?

Very strangely, there are comments under 

https://api.github.com/repos/quantumlib/cirq/issues/2112/comments

that don't appear under

https://api.github.com/repos/quantumlib/cirq/pull/2112/comments

such as the CLA bot comments. This may be because the "pull" comments are actually review comments, whereas the "issue" comments are "normal" comments.

Strilanc:
I was able to manually find Matt's approval in the api by copy pasting the comment id from the HTML source ( https://api.github.com/repos/quantumlib/cirq/pulls/2112/reviews/290232722 ) but it's still not in the list of review comments

Strilanc:
Okay I think I've learned as much as I can. I've reported it as a bug to github; they'll at least tell me the correct thing to do if I'm in the wrong. Feel free to manually merge.

KevinVillela:
Hooray, thanks everyone for your patience and comments!

Issue #2113, opened 2019-09-16
Factor out XEB fidelity estimator out of XEB experiment code


viathor:
XEB experiment code is monolithic. In particular, it tightly couples generation of random circuits and fidelity estimation. The latter is potentially useful on its own (e.g. when RQCs have been generated and executed offline), so should be factored out for re-use.

Also, there is more than one formula for estimating fidelity, e.g. on a few qubits one can compute cross-entropies directly while on many qubits one has to resort to formulas based on some assumptions about the shape of the output distribution. Therefore, it makes sense to implement fidelity estimation separate from RQC generation and overall XEB orchestration.

mpharrigan:
See https://github.com/quantumlib/ReCirq/issues/85

MichaelBroughton:
Is this still an issue @viathor, @mpharrigan  ? Is there a plan here ?

augustehirth:
I agree with this and am glad to work on it if there's consensus, as I'm already working on some other XEB tutorial related improvements.

dstrain115:
@NoureldinYosri refactored this code and it is now modular.  This is now resolved.

PR #2114, opened 2019-09-16
XEB fidelity estimator for large circuits

viathor:
First part of #2113.


cirq/experiments/__init__.py
```diff
@@ -16,3 +16,6 @@
     build_entangling_layers,
     cross_entropy_benchmarking,
 )
+
+from cirq.experiments.fidelity_estimation import (
+    estimate_circuit_fidelity,)
```
Strilanc: This name needs to be more specific to the technique being used, as there are many many ways that this could be done. For example, whether or not you are using a logarithmic score (e.g. average Bayes points lost) vs a linear score.

cirq/experiments/fidelity_estimation.py
```diff
@@ -0,0 +1,51 @@
...
+from cirq.ops import Qid
+from cirq.sim import Simulator, WaveFunctionTrialResult
+
+
+def estimate_circuit_fidelity(circuit: Circuit, qubit_order: Sequence[Qid],
+                              bitstrings: Sequence[int]) -> float:
+    """Computes fidelity estimate from one circuit using linear XEB estimator.
+
+    Args:
+        circuit: Random quantum circuit which has been executed on quantum
```
Strilanc: Describe in more detail what the conditions on this circuit are. "Random" is not specific.

cirq/experiments/fidelity_estimation.py
```diff
@@ -0,0 +1,51 @@
...
+            processor under test
+        qubit_order: Qubit order used to construct bitstrings from measurements
```
Strilanc: Can this be defaulted to the default order?

cirq/experiments/fidelity_estimation.py
```diff
@@ -0,0 +1,51 @@
...
+        bitstrings: Results of terminal all-qubit measurements performed after
```
Strilanc: Describe how the bits are packed into the ints. Little endian? Big endian? We use big endian in a lot of places..

cirq/experiments/fidelity_estimation.py
```diff
@@ -0,0 +1,51 @@
...
+            each circuit execution
+    Returns:
+        Estimate of circuit fidelity.
```
Strilanc: Describe in more detail what the returned number means and how to use it.

cirq/experiments/fidelity_estimation.py
```diff
@@ -0,0 +1,51 @@
...
+            bitstrings is inconsistent with the number of qubits.
+    """
+    dim = 2**len(qubit_order)
+
+    if set(qubit_order) != circuit.all_qubits():
+        raise ValueError(
+            f'Inconsistent qubits: circuit has {circuit.all_qubits()}, '
+            f'qubit order is {qubit_order}')
+    for bitstring in bitstrings:
+        if bitstring < 0 or dim <= bitstring:
```
Strilanc: `if not 0 <= bitstring < dim`

cirq/experiments/fidelity_estimation.py
```diff
@@ -0,0 +1,51 @@
...
+        qubit_order: Qubit order used to construct bitstrings from measurements
+        bitstrings: Results of terminal all-qubit measurements performed after
+            each circuit execution
+    Returns:
+        Estimate of circuit fidelity.
+    Raises:
+        ValueError: Circuit is inconsistent with qubit order or one of the
+            bitstrings is inconsistent with the number of qubits.
+    """
+    dim = 2**len(qubit_order)
```
Strilanc: @cduck would want this to be qudit compatible :wink:

cirq/experiments/fidelity_estimation.py
Strilanc: This is technically a set, right? Repetitions are not allowed because they can drastically increase the variance of the test.

cirq/experiments/fidelity_estimation.py
```diff
@@ -0,0 +1,51 @@
...
+        if bitstring < 0 or dim <= bitstring:
+            raise ValueError(
+                f'Bitstring {bitstring} could not have been observed '
+                f'on {len(qubit_order)} qubits.')
+
+    simulator = Simulator()
+    result = cast(WaveFunctionTrialResult,
+                  simulator.simulate(circuit, qubit_order=qubit_order))
+    output_probabilities = np.abs(result.final_state)**2
+    assert 1 - 1e-4 < np.sum(output_probabilities) < 1 + 1e-4
```
Strilanc: Take an atol parameter if you're going to have a number like this.

cirq/experiments/fidelity_estimation.py
```diff
@@ -0,0 +1,51 @@
...
+            f'Inconsistent qubits: circuit has {circuit.all_qubits()}, '
+            f'qubit order is {qubit_order}')
+    for bitstring in bitstrings:
+        if bitstring < 0 or dim <= bitstring:
+            raise ValueError(
+                f'Bitstring {bitstring} could not have been observed '
+                f'on {len(qubit_order)} qubits.')
+
+    simulator = Simulator()
+    result = cast(WaveFunctionTrialResult,
```
Strilanc: I would use `cirq.sim.final_wavefunction` for this. That method will (in the future) intelligently choose an appropriate simulator for the circuit.

cirq/experiments/fidelity_estimation_test.py
```diff
@@ -0,0 +1,59 @@
...
+def sample_noisy_bitstrings(circuit: cirq.Circuit, depolarization: float,
+                            n_samples: int) -> Sequence[int]:
+    assert 0 <= depolarization <= 1
+    n_incoherent = int(depolarization * n_samples)
+    n_coherent = n_samples - n_incoherent
+    incoherent_samples = np.random.randint(2, size=(n_incoherent, 2))
+    if n_coherent > 0:
+        sim = cirq.Simulator()
+        r = sim.run(circuit, repetitions=n_coherent)
+        coherent_samples = r.measurements['0,1']
```
Strilanc: I think `r.data['0,1']` has the ints that you want.

cirq/experiments/fidelity_estimation.py
```diff
@@ -0,0 +1,51 @@
...
+
+import numpy as np
+
+from cirq.circuits import Circuit
+from cirq.ops import Qid
+from cirq.sim import Simulator, WaveFunctionTrialResult
+
+
+def estimate_circuit_fidelity(circuit: Circuit, qubit_order: Sequence[Qid],
+                              bitstrings: Sequence[int]) -> float:
```
kevinsung: I think the bitstrings should be input in the same format as the output of running a circuit on a sampler (simulator or Engine), that is, as a numpy array. That way we don't need to perform any conversions. Also, I don't think `qubit_order` is necessary because the measurement gates in the circuit already determine a qubit order so we can assume that the bitstrings use the same order.

cirq/experiments/fidelity_estimation.py
```diff
@@ -0,0 +1,51 @@
...
+            f'qubit order is {qubit_order}')
+    for bitstring in bitstrings:
+        if bitstring < 0 or dim <= bitstring:
+            raise ValueError(
+                f'Bitstring {bitstring} could not have been observed '
+                f'on {len(qubit_order)} qubits.')
+
+    simulator = Simulator()
+    result = cast(WaveFunctionTrialResult,
+                  simulator.simulate(circuit, qubit_order=qubit_order))
```
kevinsung: If the bitstrings are in a numpy array as I suggest above, then you can use `simulator.compute_amplitudes`.

cirq/experiments/__init__.py
viathor: Renamed to `compute_linear_xeb_fidelity()`.

cirq/experiments/fidelity_estimation.py
viathor: Added comment explaining assumptions. In particular, with regards to circuit randomness, the comment explains that what is needed is the Porter-Thomas output distribution. Also, added a few words about how this assumption can be verified in practice and a reference to a paper with more details.

cirq/experiments/fidelity_estimation.py
viathor: Sure, done.

cirq/experiments/fidelity_estimation.py

cirq/experiments/fidelity_estimation.py
viathor: Done. Also added a reference.

cirq/experiments/fidelity_estimation.py

cirq/experiments/fidelity_estimation.py
viathor: Oops, good catch!

cirq/experiments/fidelity_estimation.py
viathor: Repetitions are in fact allowed. In fact, they will happen for circuits with few qubits in which case the actual frequency really matters for the estimate (if you see high-probability bitstrings often fidelity estimate will be higher).

cirq/experiments/fidelity_estimation.py
viathor: I don't think it's actually necessary. I only introduced it initially to double-check I obtained the wavefunction correctly.

cirq/experiments/fidelity_estimation.py

cirq/experiments/fidelity_estimation_test.py
viathor: Cool, thanks!

cirq/experiments/fidelity_estimation.py
viathor: Re numpy array: Done.

Re qubit_order: I don't think I can use `cirq.final_wavefunction()` with measurement gates - it'll collapse the wave function.

cirq/experiments/fidelity_estimation.py
viathor: Another comment suggested `cirq.final_wavefunction()` which plays well with the way we specify bitstrings here. That said, `compute_amplitudes()` does have an advantage since in future it may be able to compute a proper subset of amplitudes which would allow this code to work with (slightly) larger number of qubits. Then again, for convenience (compact output) and compatibility with other code I'd like to represent measurements as int arrays not as 2-D bool arrays...

What do y'all think about changing the interface (2D bool -> 1D int for measurements)? @kevinsung @Strilanc

kevinsung: I think the bitstrings input to `compute_amplitudes` (as well as your `estimate_circuit_fidelity`) should be of the same format as the bitstrings in `TrialResult.measurements`. Doesn't that better satisfy your "compatibility with other code" criterion?

cirq/experiments/fidelity_estimation.py
kevinsung: I meant a 2-d numpy array of bools, to match with the output of `Sampler` and how bitstrings are stored in `TrialResult.measurements`.

Strilanc: Personally I strongly prefer the bit packing into integers, as is done in `TrialResults.data`. We should continue moving away from the double numpy array format, which I think has been confusing to users.

cirq/experiments/fidelity_estimation.py
```diff
@@ -0,0 +1,70 @@
...
+from typing import Iterable, Tuple, Set
+
+import numpy as np
+
+from cirq.circuits import Circuit
+from cirq.ops import QubitOrder, QubitOrderOrList
+from cirq.sim import final_wavefunction
+
+
+def compute_linear_xeb_fidelity(
```
Strilanc: The word `compute` is redundant; drop it. Just `linear_xeb_fidelity`.

cirq/experiments/fidelity_estimation.py
```diff
@@ -0,0 +1,70 @@
...
+    See https://arxiv.org/abs/1608.00263 for more details.
+
+    Args:
+        circuit: Random quantum circuit which has been executed on quantum
+            processor under test
+        bitstrings: Results of terminal all-qubit measurements performed after
+            each circuit execution as integer array where each integer is
+            formed from measured qubit values according to qubit_order from
+            most to least significant qubit, i.e. in the order consistent with
+            cirq.final_wavefunction().
```
Strilanc: Use back ticks when referring to a method like `cirq.final_wavefunction`. Omit the parens.

cirq/experiments/fidelity_estimation.py
```diff
@@ -0,0 +1,70 @@
...
+    many circuits.
+
+    See https://arxiv.org/abs/1608.00263 for more details.
+
+    Args:
+        circuit: Random quantum circuit which has been executed on quantum
+            processor under test
+        bitstrings: Results of terminal all-qubit measurements performed after
+            each circuit execution as integer array where each integer is
+            formed from measured qubit values according to qubit_order from
```
Strilanc: Back ticks around `qubit_order`

cirq/experiments/fidelity_estimation_test.py
```diff
@@ -0,0 +1,89 @@
...
+
+import cirq
+
+MEASUREMENT_KEY = 'm'
+
+
+def sample_noisy_bitstrings(circuit: cirq.Circuit,
+                            qubit_order: Sequence[cirq.Qid],
+                            depolarization: float,
+                            n_samples: int) -> np.ndarray:
```
Strilanc: `repetitions` to be consistent with rest of cirq

cirq/experiments/fidelity_estimation_test.py
```diff
@@ -0,0 +1,89 @@
...
+    assert np.isclose(estimated_fidelity, expected_fidelity, atol=0.09)
+
+    np.random.set_state(prng_state)
+
+
+def test_compute_linear_xeb_fidelity_invalid_qubits():
+    q0, q1, q2 = cirq.LineQubit.range(3)
+    circuit = cirq.Circuit.from_ops(cirq.H(q0), cirq.CNOT(q0, q1),
+                                    cirq.measure(q0, q1, key=MEASUREMENT_KEY))
+    bitstrings = sample_noisy_bitstrings(circuit, (q0, q1, q2), 0.9, 10)
```
Strilanc: Why isn't this blowing up due to inserting a measurement with a duplicate key inside the sampling method?

cirq/experiments/fidelity_estimation_test.py
```diff
@@ -0,0 +1,89 @@
...
+def sample_noisy_bitstrings(circuit: cirq.Circuit,
+                            qubit_order: Sequence[cirq.Qid],
+                            depolarization: float,
+                            n_samples: int) -> np.ndarray:
+    assert 0 <= depolarization <= 1
+    dim = np.product(circuit.qid_shape())
+    n_incoherent = int(depolarization * n_samples)
+    n_coherent = n_samples - n_incoherent
+    incoherent_samples = np.random.randint(dim, size=n_incoherent)
+    circuit_with_measurements = cirq.Circuit.from_ops(
```
Strilanc: Append with `circuit + cirq.measure(...)`

cirq/experiments/fidelity_estimation_test.py
```diff
@@ -0,0 +1,89 @@
...
+        circuit, cirq.measure(*qubit_order, key=MEASUREMENT_KEY))
+    if n_coherent > 0:
+        sim = cirq.Simulator()
+        r = sim.run(circuit_with_measurements, repetitions=n_coherent)
```
Strilanc: just `cirq.sample` will do here.

cirq/experiments/fidelity_estimation_test.py
```diff
@@ -0,0 +1,89 @@
...
+                            depolarization: float,
+                            n_samples: int) -> np.ndarray:
+    assert 0 <= depolarization <= 1
+    dim = np.product(circuit.qid_shape())
+    n_incoherent = int(depolarization * n_samples)
+    n_coherent = n_samples - n_incoherent
+    incoherent_samples = np.random.randint(dim, size=n_incoherent)
+    circuit_with_measurements = cirq.Circuit.from_ops(
+        circuit, cirq.measure(*qubit_order, key=MEASUREMENT_KEY))
+    if n_coherent > 0:
```
Strilanc: I think this optimize belongs inside `sim.run` instead of outside of it. Also this is test code; don't worry about it.

cirq/experiments/fidelity_estimation_test.py
```diff
@@ -0,0 +1,89 @@
+import itertools
+
+from typing import Sequence
+
+import numpy as np
+import pytest
+
+import cirq
+
+MEASUREMENT_KEY = 'm'
```
Strilanc: Can you inline this value? It's a bit silly to make it long instead of short...

cirq/experiments/fidelity_estimation.py
Strilanc: Or e.g. `linear_xeb_fidelity_of_samples`

cirq/experiments/fidelity_estimation.py
kevinsung: I see now that the way you have it is compatible with `TrialResult.data`, so I'm fine with it actually. I agree that we should change the interface of `compute_amplitudes` to match with this.

cirq/experiments/fidelity_estimation.py
kevinsung: Ok in that case you should change the input type back to Sequence[int].

viathor: Yeah, I also prefer 1-D integer array. They render in a much more readable way on screen and make direct access and iteration simpler for most use-cases.

I'll keep int array format in this PR and will make changes elsewhere to be consistent with this representation.

cirq/experiments/fidelity_estimation.py
viathor: Agree with dropping "compute". I wouldn't append "of_samples" since it's equally redundant: samples are the only type of input that XEB fidelity estimator supports (as is explained in the comments).

cirq/experiments/fidelity_estimation_test.py
viathor: Filed #2152.

viathor: Also, removed duplicate measurements.

cirq/experiments/fidelity_estimation_test.py
viathor: Filed #2153.

cirq/experiments/fidelity_estimation_test.py
viathor: Filed #2154.

cirq/experiments/fidelity_estimation.py
```diff
@@ -0,0 +1,70 @@
...
+import numpy as np
+
+from cirq.circuits import Circuit
+from cirq.ops import QubitOrder, QubitOrderOrList
+from cirq.sim import final_wavefunction
+
+
+def linear_xeb_fidelity(
+        circuit: Circuit,
+        bitstrings: np.ndarray,
```
kevinsung: Change back to Sequence[int].

cirq/experiments/fidelity_estimation.py
viathor: FTR: #2151.

cirq/experiments/fidelity_estimation.py
viathor: FTR: #2151.

PR #2115, opened 2019-09-16
Add google protos to package

dstrain115:


dstrain115:
Oops, hold on.  This is not correct.  Let me recreate this.

PR #2117, opened 2019-09-17
Add `cirq.scatter_plot_normalized_kak_interaction_coefficients`

Strilanc:
This PR also:

- Marks `plt.show()` lines as not needing to be covered
- Fixes a deprecation warning w.r.t. `np.asscalar`
- Allows `plt.show()` to be called in doctested example code
- Allows `cirq.kak_decompose` to be called on objects with a `_unitary_` method, such as `cirq.Circuit`.

Plotting KAK interaction coefficients visually is something I've found really handy for understanding two qubit operation decomposition. I put together this flexible method for doing so.

Example:

```python
import cirq
import numpy as np

ax = None
for y in np.linspace(0, 0.5, 10):
    circuits = [
        cirq.Circuit.from_ops(
            cirq.CZ(a, b)**0.5,
            cirq.X(a)**y, cirq.X(b)**x,
            cirq.CZ(a, b)**0.5,
            cirq.X(a)**x, cirq.X(b)**y,
            cirq.CZ(a, b) ** 0.5,
        )
        for x in np.linspace(0, 1, 100)
    ]

    ax = cirq.scatter_plot_normalized_kak_interaction_coefficients(
        circuits,
        s=1,
        ax=ax,
        include_frame=ax is None,
        label=f'y={y:0.2f}')
ax.legend()
plt.show()
```

image

image

### DIFF
--- a/cirq/linalg/decompositions.py
+++ b/cirq/linalg/decompositions.py
@@ -595,8 +595,8 @@ def scatter_plot_normalized_kak_interaction_coefficients(
         ...         include_frame=ax is None,
         ...         label=f'y={y:0.2f}')
         >>> _ = ax.legend()
-        >>> # import matplotlib.pyplot as plt
-        >>> # plt.show()
+        >>> import matplotlib.pyplot as plt
+        >>> plt.show()
     """
     if ax is None:
         fig = plt.figure()
--- a/cirq/linalg/decompositions.py
+++ b/cirq/linalg/decompositions.py
@@ -526,13 +526,14 @@ class KakDecomposition:
             before)
 
 
-def scatter_plot_normalized_kak_interaction_coefficients(interactions: Iterable[
-        Union[np.ndarray, 'cirq.SupportsUnitary', 'KakDecomposition']],
-                                                         *,
-                                                         include_frame: bool = True,
-                                                         ax: Optional[plt.Axes] = None,
-                                                         show: bool = False,
-                                                         **kwargs):
+def scatter_plot_normalized_kak_interaction_coefficients(
+        interactions: Iterable[
+            Union[np.ndarray, 'cirq.SupportsUnitary', 'KakDecomposition']],
+        *,
+        include_frame: bool = True,
+        ax: Optional[plt.Axes] = None,
+        show: bool = False,
+        **kwargs):
     """Plots the interaction coefficients of many two-qubit operations.
 
     Plots:
--- a/cirq/linalg/decompositions.py
+++ b/cirq/linalg/decompositions.py
@@ -628,12 +628,7 @@ def scatter_plot_normalized_kak_interaction_coefficients(
 
     points = []
     for obj in interactions:
-        if isinstance(obj, KakDecomposition):
-            kak = obj
-        elif isinstance(obj, np.ndarray) and obj.shape == (4, 4):
-            kak = kak_decomposition(obj)
-        else:
-            kak = kak_decomposition(protocols.unitary(obj))
+        kak = kak_decomposition(obj)
         normalized = np.array(kak.interaction_coefficients) * 4 / np.pi
         points.append(normalized)
 
@@ -762,13 +757,15 @@ def kak_canonicalize_vector(x: float, y: float, z: float,
 
 
 def kak_decomposition(
-        mat: np.ndarray,
+        unitary_object: Union[np.ndarray, 'cirq.SupportsUnitary'],
         rtol: float = 1e-5,
         atol: float = 1e-8) -> KakDecomposition:
     """Decomposes a 2-qubit unitary into 1-qubit ops and XX/YY/ZZ interactions.
 
     Args:
-        mat: The 4x4 unitary matrix to decompose.
+        unitary_object: The value to decompose. Can either be a 4x4 unitary
+            matrix, or an object that has a 4x4 unitary matrix (via the
+            `cirq.SupportsUnitary` protocol).
         rtol: Per-matrix-entry relative tolerance on equality.
         atol: Per-matrix-entry absolute tolerance on equality.
 
@@ -787,6 +784,13 @@ def kak_decomposition(
         'An Introduction to Cartan's KAK Decomposition for QC Programmers'
         https://arxiv.org/abs/quant-ph/0507171
     """
+    if isinstance(unitary_object, KakDecomposition):
+        return unitary_object
+    if isinstance(unitary_object, np.ndarray):
+        mat = unitary_object
+    else:
+        mat = protocols.unitary(unitary_object)
+
     magic = np.array([[1, 0, 0, 1j],
                       [0, 1j, 1, 0],
                       [0, 1j, -1, 0],
--- a/cirq/linalg/decompositions.py
+++ b/cirq/linalg/decompositions.py
@@ -15,13 +15,25 @@
 
 """Utility methods for breaking matrices into useful pieces."""
 
-from typing import Callable, List, Set, Tuple, TypeVar, Union
+from typing import (
+    Callable,
+    List,
+    Set,
+    Tuple,
+    TypeVar,
+    Union,
+    Iterable,
+    Optional,
+)
 
 import math
 import cmath
 import numpy as np
 
-from cirq import value
+import matplotlib.pyplot as plt
+
+import cirq
+from cirq import value, protocols
 from cirq._compat import proper_repr
 from cirq.linalg import combinators, diagonalize, predicates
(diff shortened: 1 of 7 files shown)


cirq/linalg/decompositions.py
```diff
@@ -514,6 +528,120 @@ def interaction_matrix(m: np.ndarray, c: float) -> np.ndarray:
...
+            (0, 0, 0),
+            (1, 0, 0),
+            (1, 1, 0),
+            (0, 0, 0),
+        ]
+        ax.plot(*coord_transform(envelope), c='black', linewidth=1)
+
+    points = []
+    for obj in interactions:
+        kak = kak_decomposition(obj)
```
pingyeh: I haven't found time to read the reference paper on KAK decomposition, but I take it that it's some logic that transforms the object into a KakDecomposition whose `interaction_coefficients` are to be plotted in a 3D axes. That sounds like mixing business logic (the transformation) with presentation (the plotting).

How about making this a method of the KakDecomposition class?

cirq/linalg/decompositions.py
```diff
@@ -667,6 +796,13 @@ def kak_decomposition(
         'An Introduction to Cartan's KAK Decomposition for QC Programmers'
         https://arxiv.org/abs/quant-ph/0507171
     """
+    if isinstance(unitary_object, KakDecomposition):
+        return unitary_object
+    if isinstance(unitary_object, np.ndarray):
+        mat = unitary_object
```
viathor: This special-casing to make numpy arrays work in a context that expects a unitary seems potentially useful elsewhere. Have you considered moving this to the unitary protocol?

cirq/linalg/decompositions.py
```diff
@@ -514,6 +528,120 @@ def interaction_matrix(m: np.ndarray, c: float) -> np.ndarray:
...
+        The wireframe includes lines along the surface of the space at z=0.
+
+    Args:
+        interactions: An iterable of two qubit unitary interactions. Each
+            interaction can be specified as a raw 4x4 unitary matrix, or an
+            object with a 4x4 unitary matrix according to `cirq.unitary` (
+            (e.g. `cirq.CZ` or a `cirq.KakDecomposition` or a `cirq.Circuit`
+            over two qubits).
+        include_frame: Determines whether or not to draw the kak space
+            wireframe. If not set. Defaults to `True`.
```
viathor: nit: "If not set, defaults to `True`." (one sentence, two sub-clauses).

Alternatively, remove this since it duplicates what is already stated in function signature.

cirq/linalg/decompositions.py
```diff
@@ -514,6 +528,120 @@ def interaction_matrix(m: np.ndarray, c: float) -> np.ndarray:
...
+        kwargs: Arguments forwarded into the call to `scatter` that plots the
+            points. Working arguments include color `c='blue'`, scale `s=2`,
+            labelling `label="theta=pi/4"`, etc. For reference see the
+            `matplotlib.pyplot.scatter` documentation:
+            https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.scatter.html
+
+    Returns:
+        The matplotlib 3d axes object that was plotted into.
+
+    Examples:
```
viathor: nit: "Example" (singular)

cirq/linalg/decompositions.py
```diff
@@ -514,6 +528,120 @@ def interaction_matrix(m: np.ndarray, c: float) -> np.ndarray:
...
+    """
+    if ax is None:
+        fig = plt.figure()
+        ax = fig.add_subplot(projection='3d')
+
+    def coord_transform(
+            pts: Iterable[Tuple[float, float, float]]
+    ) -> Tuple[Iterable[float], Iterable[float], Iterable[float]]:
+        xs, ys, zs = zip(*pts)
+        return xs, zs, ys
```
viathor: This just wraps the `zip(*argument)` expression which is hardly enough logic to justify a function.

OTOH, inlining it leads to one of the idiomatic applications of `zip()`: transposing a 2-layer collection:

```python
ax.plot(*zip(*envelope), ...)
```

cirq/linalg/decompositions.py
viathor: I think this is fine as-is. AFAIU, the purpose of `scatter_plot_normalized_kak_interaction_coefficients()`  is to help in interactive work where one repeatedly examines the coefficients of different variations of a unitary. Splitting out the decomposition and presentation would force the user to call two functions instead of one.

Also, this particular presentation is tailored to the specifics of the decomposition.

cirq/linalg/decompositions.py
```diff
@@ -514,6 +528,120 @@ def interaction_matrix(m: np.ndarray, c: float) -> np.ndarray:
...
+            (1, 0, 0),
+            (0, 0, 0),
+            (1, 1, -1),
+            (1, 0, 0),
+            (0, 0, 0),
+            (1, 0, 0),
+            (1, 1, 0),
+            (0, 0, 0),
+        ]
+        ax.plot(*coord_transform(envelope), c='black', linewidth=1)
```
viathor: Could you add labels to the points that correspond to well-known gates like CNOT/CZ, iSWAP, SWAP etc? I think this would be helpful to new users and people who use cirq to learn about KAK.

cirq/linalg/decompositions.py
Strilanc: > That sounds like mixing business logic (the transformation) with presentation (the plotting).

We already have support for getting the interaction coefficients and matplotlib already has methods for making scatter plots. This method puts the two together in a convenient way. Factoring it into two methods would produce no changes to the code base, because the two pieces already exist.

Strilanc: > How about making this a method of the KakDecomposition class?

I originally did, but it had to be a static method (because we're talking about a *list* of values instead of a single value) and also this made it less clear that you could include things that were not predecomposed in the list.

cirq/linalg/decompositions.py
Strilanc: I have considered it. One thing that worries me about it is what to do if the array is not unitary. Do we need to check?

cirq/linalg/decompositions.py
Strilanc: This is the standard name for the section, similar to Args when only one argument.

cirq/linalg/decompositions.py
Strilanc: This method also transposes the Y and Z coordinates, to orient the plot correctly.

cirq/linalg/decompositions.py
Strilanc: I think putting text at these points would clutter the plot. They are already highlighted in the sense that they are at corners or intersections of blue lines. The SWAP is the top left = top right corner. CZ is the bottom of the vertical triangle. ISWAP is the middle of the top of the vertical triangle.

I'll see if I can draw something i nthe doc string.

Strilanc: How's this

```
        The space is a prism with the identity at the origin, a crease along
        y=z=0 leading to the CZ/CNOT at x=1 and a vertical triangular face that
        contains the iswap at x=y=1,z=0 and the swap at x=y=z=1:

                                 (x=1,y=1,z=0)   
                             swap___iswap___swap (x=1,y=1,z=-1)
                               _/\    |    /
                             _/   \   |   /
                           _/      \  |  /
                         _/         \ | /
                       _/            \|/
        (x=0,y=0,z=0) I---------------CZ (x=1,y=0,z=0)
```

cirq/linalg/decompositions.py
viathor: Currently, the protocol raises an exception when the argument fails to satisfy our requirements, so I'd do the same in the new case.

Minor caveat: Existing cases justify its use of the `TypeError` exception. The present case fits `ValueError` better.

cirq/linalg/decompositions.py
viathor: Oops, TIL!

cirq/linalg/decompositions.py
viathor: Missed this. Keep as is.

cirq/linalg/decompositions.py
viathor: How about using colored points at the vertices and adding a legend for the color mapping? You'd have to somehow avoid the color collisions with the input data, though.

viathor: The ASCII art also works (didn't see it before posting previous comment).

PR #2127, opened 2019-09-19
Add cirq.ArithmeticOperation helper class

Strilanc:
This is a helper class to make it easier to implement arithmetic operations. These operations are not NISQ-y, but people have asked if cirq can do them efficiently a few times in the past. The answer was that it's *possible*, by implementing an `_apply_unitary_` operation, but it's not so convenient to write that method. Now we have a convenient way to do it.

Here is how you would implement an adder using this class:

```python
class Add(cirq.ArithmeticOperation):
    def __init__(self, target_register, input_register):
        self.target_register = target_register
        self.input_register = input_register

    def registers(self):
        return self.target_register, self.input_register

    def with_registers(self, *new_registers):
        return Add(*new_registers)

    def apply(self, target_value, input_value):
        return target_value + input_value
```

The fact that the output is an int instead of a list of two ints, and also that the output may be larger than the register range, is all fixed up automatically. int gets wrapped into List[int], List[int] gets padded up to max length assuming on changes in later registers, and each result gets modded into range. I originally tried a version with all these things explicit, but it was generally quite a pain to have to constantly deal with sizing and remembering to list unchanged values.

A fun feature is that, because the input register is not modified (phase kickback notwithstanding), the input register can be set to a list of qubits *or* to a classical constant:

With a constant:

```python
cirq.unitary(
    Add(target_register=cirq.LineQubit.range(2),
        input_register=1))
```

```
[[0, 0, 0, 1],
 [1, 0, 0, 0],
 [0, 1, 0, 0],
 [0, 0, 1, 0]]
```

With another register:

```python
c = cirq.Circuit.from_ops(
   cirq.X(cirq.LineQubit(3)),
   cirq.X(cirq.LineQubit(2)),
   cirq.X(cirq.LineQubit(6)),
   cirq.measure(*cirq.LineQubit.range(4, 8), key='before:in'),
   cirq.measure(*cirq.LineQubit.range(4), key='before:out'),

   Add(target_register=cirq.LineQubit.range(4),
       input_register=cirq.LineQubit.range(4, 8)),

   cirq.measure(*cirq.LineQubit.range(4, 8), key='after:in'),
   cirq.measure(*cirq.LineQubit.range(4), key='after:out'),
)
print(cirq.sample(c).data)
```

```
   before:in  before:out  after:in  after:out
0          2           3         2          5
```

### DIFF
--- a/cirq/ops/arithmetic_operation.py
+++ b/cirq/ops/arithmetic_operation.py
@@ -15,12 +15,14 @@
 
 import abc
 import itertools
-from typing import Union, Iterable, List, Sequence, cast
+from typing import Union, Iterable, List, Sequence, cast, TYPE_CHECKING
 
 import numpy as np
 
-import cirq
 from cirq.ops.raw_types import Operation
+
+if TYPE_CHECKING:
+    import cirq
 
 
 class ArithmeticOperation(Operation, metaclass=abc.ABCMeta):
@@ -32,7 +34,8 @@ class ArithmeticOperation(Operation, metaclass=abc.ABCMeta):
     This class handles the details of ensuring that the scaling of implementing
     the operation is O(2^n) instead of O(4^n) where n is the number of qubits
     being acted on, by implementing an `_apply_unitary_` function in terms of
-    the registers and the apply function of the child class.
+    the registers and the apply function of the child class. It also handles the
+    boilerplate of implementing the `qubits` and `with_qubits` methods.
 
     Examples:
         >>> class Add(cirq.ArithmeticOperation):
--- a/cirq/ops/arithmetic_operation.py
+++ b/cirq/ops/arithmetic_operation.py
@@ -1,0 +1,245 @@
+# Copyright 2018 The Cirq Developers
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Helper class for implementing classical arithmetic operations."""
+
+import abc
+import itertools
+from typing import Union, Iterable, List, Sequence, cast
+
+import numpy as np
+
+import cirq
+from cirq.ops.raw_types import Operation
+
+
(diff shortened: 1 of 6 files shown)


cirq/ops/arithmetic_operation.py
```diff
@@ -0,0 +1,248 @@
...
+        operation (their value must remain fixed when passed to `apply`).
+
+        Registers are big endian. The first qubit is the most significant, the
+        last qubit is the 1s qubit, the before last qubit is the 2s qubit, etc.
+
+        Returns:
+            A list of constants and qubit groups that the operation will act
+            upon.
+        """
+        raise NotImplementedError()
```
mpharrigan: technically unnecessary; redundant with `abc.abstractmethod`.

cirq/ops/arithmetic_operation.py
```diff
@@ -0,0 +1,248 @@
...
+
+    @abc.abstractmethod
+    def with_registers(self, *new_registers: Union[int, Sequence['cirq.Qid']]
```
mpharrigan: why would one use this instead of just calling the constructor again?

cirq/ops/arithmetic_operation.py
```diff
@@ -0,0 +1,248 @@
...
+        1. The `apply` method is permitted to return values that have more bits
+            than the registers they will be stored into. The extra bits are
+            simply dropped. For example, if the value 5 is returned for a 2
+            qubit register then 5 % 2**2 = 1 will be used instead. Negative
+            values are also permitted. For example, for a 3 qubit register the
+            value -2 becomes -2 % 2**3 = 6.
+        2. When the value of the last `k` registers is not changed by the
+            operation, the `apply` method is permitted to omit these values
+            from the result. That is to say, when the length of the output is
+            less than the length of the input, it is padded up to the intended
```
mpharrigan: I'm mildly confused about the description of this feature; especially where it talks about "padding".

Is this an example: If I had a "summation" gate that takes `k` registers and puts the sum in the first register, my `apply` method would only return one number. Where is padding happening? How is this point (2) different than the next one (3)?

cirq/ops/arithmetic_operation.py
Strilanc: Child classes often have parameters other than the registers in their constructor. For example: a boolean switch determining whether to increment or decrement, or metadata about the source of the gate. They may also follow conventions such as enforcing keywords.

cirq/ops/arithmetic_operation.py
Strilanc: If you have two modified registers and three input registers, you can return a pair of values instead of a quintuplet of values. It's true that this can be seen as a generalization of (3), but the in the code they refer to separate steps.

PR #2130, opened 2019-09-19
Make Duration JSON serializable

kevinsung:

PR #2131, opened 2019-09-19
Make _NamedConstantXmonDevice JSON serializable

kevinsung:


mpharrigan:
Looks like this roped in the `Duration` changes

cirq/google/known_devices.py
```diff
@@ -61,6 +61,16 @@ def __init__(self, constant: str, **kwargs) -> None:
...
 
+    def _json_dict_(self):
+        return {
+            'cirq_type': self.__class__.__name__,
+            'constant': self._repr,
+            'measurement_duration': self._measurement_duration,
+            'exp_w_duration': self._exp_w_duration,
+            'exp_11_duration': self._exp_z_duration,
+            'qubits': sorted(self.qubits)
+        }
```
mpharrigan: When I went through with the original serialization PR, I took the opportunity to make slight modifications to classes to make all these classes follow more of a "dumb-container-of-data" model. The certification of a class is the ability to use `protocols.obj_to_dict_helper`. 

I think we should strongly consider

1. renaming `self._repr` to `self.constant`
2. exposing `_measurement_duration`, `_exp_w_duration`, `_exp_z_duration` as properties on `XmonDevice` *or* just rename those properties to not have the initial underscore.

@Strilanc : thoughts?

As a final note, I don't think we'd be able to get all the way because for json you have to do the conversion from qubits being a `frozenset` to a list

cirq/google/known_devices_test.py
```diff
@@ -0,0 +1,33 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import cirq
+
+
+def test_json_dict():
```
mpharrigan: also enable a roundtrip test in `protocols.json_test`

kevinsung: Done.

kevinsung:
The Duration change is necessary because this class's JSON dict includes Durations; that's why this comes after #2130.

cirq/protocols/json_test.py
```diff
@@ -444,7 +446,14 @@ def test_mutually_exclusive_blacklist():
 ]
 
 
-@pytest.mark.parametrize('cirq_type,cls', _get_all_public_classes())
+def _roundtrip_test_classes():
+    for cirq_type, cls in _get_all_public_classes():
+        yield cirq_type, cls
+    yield ('_NamedConstantXmonDevice',
+           cirq.google.known_devices._NamedConstantXmonDevice)
```
mpharrigan: Interesting... so these things aren't exposed at the top-level namespace because `cirq.google` is not "flattened". 

Instead of this other function, you could add the yield to the end of `_get_all_public_classes`. There's already a couple ad-hoc items there. 

The real solution would be: refactor `_get_public_classes` to take a module object to extract public classes from and then chain together `_get_public_classes(cirq) + _get_public_classes(cirq.google)`. You'd also have to go through and add the resultant test failures to `SHOULDNT_BE_SERIALIZED` or `NOT_YET_SERIALIZABLE`.

If you want, you can do the second thing as part of this PR. Otherwise, open an issue and link to it in a comment with the ad-hoc insertion of `_NamedConstantXmonDevice`

mpharrigan: As a final thing: the actual objects returned by this inspecting function would be `Foxtail` and `Bristlecone` instead of `_NamedConstantXmonDevice`. This should still provide full test coverage of what we care about

kevinsung: That still wouldn't get `_NamedConstantXmonDevice` because it is not public in cirq.google (and shouldn't be), though, so we'd still need the ad-hoc insertion, right?

mpharrigan: Right, only `Foxtail` and `Bristlecone` would show up, but those would take care of testing `_NamedConstantXMonDevice`. 

It's like how we test for `UNCONSTRAINED_DEVICE` which is a singleton of type `_UnconstrainedDevice`

kevinsung: I've refactored the tests. I still have this other function to handle the ad-hoc cases like complex, np.ndarray, etc.

PR #2132, opened 2019-09-19
Add ability to slice sweeps

### LINKED ISSUE #2106: Add __getitem__[start:stop] slicing support to parameter sweeps
If we have a sweepable that is really big it is useful to break it up into smaller sweeps.  We should add functionality which does this slicing.

MichaelBroughton:
Fixes #2106 .
Since we are producing iterators, I found pythons `itertools.islice` will only allow for positive slicing/access indices. If we are OK producing and using lists I could quickly change things to allow for negative indices as well (the code would also be way shorter)


cirq/study/sweeps.py
```diff
@@ -101,6 +101,28 @@ def __iter__(self) -> Iterator[resolver.ParamResolver]:
...
+            single_item = True
+        if not isinstance(val, slice):
+            raise TypeError(
+                'Sweep indices must be either int or slices, not {}'\
+                    .format(type(val)))
+        if (val.start is not None and val.start < 0) or \
+           (val.stop is not None and val.stop < 0) or \
+           (val.step is not None and val.step < 0):
+            raise ValueError(
+                'Cannot slice/access sweeps using negative indices.')
```
maffoo: I think we would want to support as much of the standard python slicing syntax as we can. Generic iterables do not have a known length, so `itertools.islice` can't support negative indexes, but sweeps _do_ have know length so we should be able to handle these.

cirq/study/sweeps.py
```diff
@@ -101,6 +101,28 @@ def __iter__(self) -> Iterator[resolver.ParamResolver]:
         for params in self.param_tuples():
             yield resolver.ParamResolver(collections.OrderedDict(params))
 
+    def __getitem__(self, val) -> Union[Params, Iterator[Params]]:
```
maffoo: I think if we slice we should get back a `Sweep`, not just an `Iterator[Params]`. This is more consistent with other types, for example slicing a list gives a list. Also, getting a single index should return `resolver.ParamResolver` since if you just iterate over the sweep you get `Iterator[resolver.ParamResolver]` (see `__init__` above). For the latter, I think you just need to pass `self` instead of `self.param_tuples()` as the first arg to `itertools.islice`.

Also, note that you can use `an @overload` decorator to make the types more precise here so that mypy knows getting a single index returns one type, while getting a slice returns a Sweep (for reference, see how we do this in [`Circuit`](https://github.com/quantumlib/Cirq/blob/3bb71d11042543e26d3e6e4d0dc784f03eef4a5d/cirq/circuits/circuit.py#L177)).

cirq/study/sweeps.py
MichaelBroughton: I've updated the code to support this, let me know what you think now.

cirq/study/sweeps.py
MichaelBroughton: I'm returning a sweep now, but I'm not exactly sure about this @overload, can you give me a more concrete example ?

cirq/study/sweeps.py
```diff
@@ -101,6 +102,38 @@ def __iter__(self) -> Iterator[resolver.ParamResolver]:
         for params in self.param_tuples():
             yield resolver.ParamResolver(collections.OrderedDict(params))
 
+    # pylint: disable=function-redefined
+    @overload
+    def __getitem__(self, val: int) -> resolver.ParamResolver:
+        pass
+
+    # pylint: disable=function-redefined
```
maffoo: remove this and instead re-enable the rule after the `__getitem__` implementation.

cirq/study/sweeps.py
```diff
@@ -101,6 +102,38 @@ def __iter__(self) -> Iterator[resolver.ParamResolver]:
...
+            if val < 0:
+                val += len(self)
+            val = slice(val, val + 1, None)
+        if not isinstance(val, slice):
+            raise TypeError(
+                'Sweep indices must be either int or slices, not {}'\
+                    .format(type(val)))
+
+        inds_map: Dict[int, int] = {
+            val: i for i, val in enumerate(range(len(self))[val])
```
maffoo: use a different name for the loop variable, maybe do `i` -> `slice_i`, `val` -> `sweep_i`. Something like that.

cirq/study/sweeps.py
```diff
@@ -101,6 +102,38 @@ def __iter__(self) -> Iterator[resolver.ParamResolver]:
...
+        }
+        results = [resolver.ParamResolver()] * len(inds_map.keys())
```
maffoo: Can just do `len(inds_map)`.

cirq/study/sweeps.py
```diff
@@ -101,6 +102,38 @@ def __iter__(self) -> Iterator[resolver.ParamResolver]:
...
+    # pylint: disable=function-redefined
+    @overload
+    def __getitem__(self, val: slice) -> 'Sweep':
+        pass
+
+    def __getitem__(self, val):
+        if isinstance(val, int):
+            if val < 0:
+                val += len(self)
+            val = slice(val, val + 1, None)
```
maffoo: can just do
```python
return next(itertools.islice(self, val))
```

cirq/study/sweeps.py
```diff
@@ -101,6 +102,38 @@ def __iter__(self) -> Iterator[resolver.ParamResolver]:
...
+        if not isinstance(val, slice):
+            raise TypeError(
+                'Sweep indices must be either int or slices, not {}'\
```
maffoo: nit: don't need trailing slash

cirq/study/sweeps.py
```diff
@@ -101,6 +102,38 @@ def __iter__(self) -> Iterator[resolver.ParamResolver]:
...
+        inds_map: Dict[int, int] = {
+            val: i for i, val in enumerate(range(len(self))[val])
+        }
+        results = [resolver.ParamResolver()] * len(inds_map.keys())
+        for i, item in enumerate(self):
+            if i in inds_map:
+                results[inds_map[i]] = item
+
+        if len(results) == 1:
+            return results[0]
```
maffoo: not needed.

cirq/study/sweeps_test.py
```diff
@@ -74,6 +74,48 @@ def test_product():
...
+    sweep_reversed = sweep[::-1]
+    assert list(sweep) == list(reversed(list(sweep_reversed)))
+
+
+def test_access_sweep():
+    sweep = cirq.Points('a', [1, 2, 3]) * cirq.Points('b', [4, 5, 6, 7])
+
+    last_elem = sweep[-1]
+    assert last_elem.__repr__(
+    ) == 'cirq.ParamResolver(OrderedDict([(\'a\', 3), (\'b\', 7)]))'
```
maffoo: nit: just compare equality to a cirq.ParamResolver instance. Also below.

cirq/study/sweeps_test.py
```diff
@@ -74,6 +74,48 @@ def test_product():
...
+        _ = sweep['junk']
+
+
+def test_slice_sweep():
+    sweep = cirq.Points('a', [1, 2, 3]) * cirq.Points('b', [4, 5, 6, 7])
+
+    first_two = sweep[:2]
+    assert list(first_two.param_tuples())[0] == (('a', 1), ('b', 4))
+    assert list(first_two.param_tuples())[1] == (('a', 1), ('b', 5))
+    assert len(list(first_two)) == 2
```
maffoo: Maybe add a test case for a slice of length 1 (should give sweep of len 1, not param resolver).

cirq/study/sweeps.py
maffoo: Actually, that suggestion is also not quite right. Imagine sweep were length 10 and you asked for `sweep[-20]`; the code above would return the 0th element, but should raise an `IndexError`. I think we need to a bounds check:
```python
if isinstance(val, int):
    if val < -len(self) or val >= len(self):
        raise IndexError(f'sweep index out of range: {val}')
    if val < 0:
        val += len(self)
    return next(itertools.islice(self, val))
```
It might be nice to make a local variable `n = len(self)` at the top of the function to avoid calling `len(self)` multiple times.

cirq/study/sweeps.py
MichaelBroughton: done.

cirq/study/sweeps.py
MichaelBroughton: done.

cirq/study/sweeps.py
MichaelBroughton: fixed.

cirq/study/sweeps.py
MichaelBroughton: changed.

cirq/study/sweeps.py
MichaelBroughton: removed.

cirq/study/sweeps.py
MichaelBroughton: removed.

cirq/study/sweeps_test.py
MichaelBroughton: fixed.

cirq/study/sweeps_test.py
MichaelBroughton: added.

cirq/study/sweeps.py
maffoo: Looks like the fix does not reflect my follow-up comment about bounds checking. Should fix that to ensure we raise `IndexError` for out-of-bounds indexes, instead of returning the zeroth element or raising `StopIteration`.

PR #2136, opened 2019-09-20
Define cirq.Moment.__pow__

Strilanc:
- A moment has an powered version when each of its operations has a powered version
- Rewrite circuit inverse to delegate to moment inverse
- Move `cirq/protocols/inverse.py` to `cirq/protocols/inverse_protocol.py` to fix a mypy complaint


mpharrigan:
Can you comment on some example use-cases of distributing an exponent to every operation in the moment?

Strilanc:
Well, it was used in the circuit inverse code.

Mostly I'm including it because it feels right for it to act this way. It makes it more cohesive with the rest of the library. It should also have a unitary and stuff like that.

Strilanc:
@c-poole Have you looked more into what is causing the Windows flakes?

Issue #1787, opened 2019-07-10
Syntactic sugar for cirq.Circuit.from_ops (?)


mpharrigan:
When using cirq, you'll typically want to create a circuit. As far as I can tell, the most common way of doing that is with `cirq.Circuit.from_ops`. Since this is so common, it *might* be nice to have a shorter alias. Maybe `cirq.ascircuit(...)` by analogy to `np.asarray`. What do you think?

bryano:
Why don't we just allow arbitrary op_trees in the `Circuit` constructor?

Strilanc:
Now that we have the convention about moments going in without being reflowed, `from_ops` is in fact a strict superset of the constructor. Allowing `cirq.Circuit(OP_TREE)` could be reasonable.

PR #2137, opened 2019-09-20
[WIP] Refactor Circuit.__init__ to accept an OP_TREE and use it instead of from_ops

Strilanc:
- Deprecate cirq.Circuit.from_ops(*ops) in favor of cirq.Circuit(*ops)
- Replace all usage of cirq.Circuit.from_ops with cirq.Circuit (except in the from_ops test
- Added a `strategy` parameter to cirq.Circuit.__init__

BREAKING CHANGES:
- Circuit(moments=) no longer works
- Circuit(moments, device) will now cause a "that's not an operation" error; have to say Circuit(moments, device=device)
- e.g. I had to fix about ~10 tests which had become incompatible

I'd like comments/approval from @dabacon, @mpharrigan, and @maffoo on this change

Fixes https://github.com/quantumlib/Cirq/issues/1787

This change is to reduce boilerplate, especially in example code.
Note that, because moments are guaranteed to insert while ignoring the insertion strategy, all existing calls will return the same circuit (or else fail because they specified moments= or didn't specify device=)


Strilanc:
I made the mistake of doing the replace-all changes at the same time as the core changes. But I only did the format changes in a second commit, so the first commit will give an easier-on-the-eyes diff.

mpharrigan:
Well I opened the original ticket asking for this, so I'm on board :) 

We might want to leave `from_ops` as a deprecated method for a couple of releases since it is *very* widely used

Strilanc:
This PR in fact does deprecate that method.

(You mean you want to increase the deadline version?)

dstrain115:
I am not against this change, but it will definitely generate a lot of churn.

Correct me if I am wrong, but this basically changes both the major ways you instantiate a Circuit, so pretty much everyone will need to modify their code since it will either be broken or deprecated.

I am in favor of a more gentle deprecation (maybe change deprecation deadline to v0.8 and also keep moments around for another cycle?)

mpharrigan:
> (You mean you want to increase the deadline version?)

yes

Strilanc:
`cirq.it` is certainly "qute". It might be over-optimizing for this one use case. It will also make the examples less readable to new users.

Strilanc:
I've added `deprecated_parameter` and used it to rewrite old calls that use a positional device or the moments keyword into the new signature. This change is no longer a breaking change, but rather just a deprecating change.

Strilanc:
I've also bumped the deprecation version to 0.8

Strilanc:
Example of how the deprecation plays out:

```
In [5]: cirq.Circuit(moments=[cirq.Moment([cirq.X(a)])])                                                                           
```

```
WARNING:root:DEPRECATION
The moments keyword parameter of cirq.Circuit was used but is deprecated.
It will be removed in cirq v0.8.
Pass contents positionally or using the "contents=" keyword.

Out[5]: 0: ───X───
```

```
In [6]: cirq.Circuit(moments=[cirq.Moment([cirq.X(a)])])                                                                           
```

```
Out[6]: 0: ───X───
```

(no output results the second time)

Strilanc:
I will split this PR into multiple pieces if we decide to merge it.

First piece: define deprecated parameter
Second piece: rewrite cirq.Circuit and deprecate its parameters
Third piece: deprecate from_ops and replace all usage

Strilanc:
Thanks everyone for commenting (@dabacon did via other channels). Moving forward on this now.

Strilanc:
Obsoleted by https://github.com/quantumlib/Cirq/pull/2181

PR #2140, opened 2019-09-20
Add OpTree protocol

maffoo:
Defining OP_TREE using a protocol lets us create a recursive type, which is not possible with type aliases alone. This means downstream operations can preserve more type information, which is good but reveals a problem with `flatten_op_tree` returning a Union type due to the `preserve_moments` option. To work around this I've split it up into two functions instead. Another possibility would be to remove the `preserve_moments` option on `flatten_op_tree` and just add a second function for the non-default case (which is a tiny minority of callsites).

### DIFF
--- a/cirq/ops/op_tree.py
+++ b/cirq/ops/op_tree.py
@@ -15,37 +15,43 @@
 """A recursive type describing trees of operations, and utility methods for it.
 """
 
-from typing import Any, Callable, Iterable, Union
+from typing import Callable, Iterable, Iterator, NoReturn, Union
+from typing_extensions import Protocol
 
 from cirq.ops.moment import Moment
 from cirq.ops.raw_types import Operation
 
 
-OP_TREE = Union[Operation, Iterable[Any]]
-"""The recursive type consumed by circuit builder methods.
+class OpTree(Protocol):
+    """The recursive type consumed by circuit builder methods.
 
-An OP_TREE is a contract, not a class. The basic idea is that, if the input can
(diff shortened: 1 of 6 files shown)


maffoo:
To clarify, this is not a "protocol" in the cirq sense, but just using a mypy `Protocol` to define `OP_TREE` instead of a type alias, so that we can describe the recursive type withouy falling back to `Any`.

cirq/ops/op_tree.py
```diff
@@ -15,37 +15,43 @@
...
-- A generator yielding operations is an OP_TREE.
+    For example:
+    - An Operation is an OP_TREE all by itself.
+    - A list of operations is an OP_TREE.
+    - A list of tuples of operations is an OP_TREE.
+    - A list with a mix of operations and lists of operations is an OP_TREE.
+    - A generator yielding operations is an OP_TREE.
 
-Note: once mypy has support for recursive types we can define this as:
+    Note: once mypy supports recursive types this could be defined as an alias:
```
mpharrigan: conservation of passive voice :)

maffoo:
@Strilanc, how do you feel about a new `flatten_to_ops` function as I've done here, vs just removing the `preserve_moments` parameter on the existing `flatten_op_tree` function? I like the idea of removing the parameter; it's technically a breaking change but I doubt there's much outside usage (nothing like `Circuit.from_ops`, say).

Strilanc:
I liked that change. We can deprecate the old method at a later time if we want.

PR #2145, opened 2019-09-22
Delete repeated `Product` import in `study/__init__.py`

nathanshammah:
Remove a repeated import for the `Product` class defined in `study/sweeps.py`.


nathanshammah:
@googlebot I signed it

Issue #2107, opened 2019-09-12
Engine should using sweepable slices to keep job sizes down


dabacon:
Follow up after #2106

Engine should take a repetition x parameter sweep size that it wants to break things into, and if the job is bigger than this break the job into multiple jobs of this size (or smaller).  

One issue here is how to name these jobs.

MichaelBroughton:
I can take this on. I'll plan to solve it in two steps:
1. Add a function to combine two trial results together (provided they have same params and qubits measured).
2. Add a new class EngineJobBatch which will manage the small enginejobs.

How does that sound ?

balopat:
@MichaelBroughton are you still interested in this? 
1.) still sounds useful 
2.) we might be able to get away with just implementing this in the methods of Engine, EngineProgram, adding the extra param `max_job_size` to it and then handle returning of multiple jobs and the merging of the TrialResults

dstrain115:
@wcourtney Did we add a maximum batch size to the client at some point?  Do we still need this feature?

dstrain115:
@wcourtney With streaming, is this still a valid issue?  Can we close it?

wcourtney:
The issue describes a solution without a problem it's trying to solve. I'd guess the problem is that a large job could block other users in an open-swim mode, but we now interleave executions in a way related to sweeps, so that shouldn't be an issue.

PR #2146, opened 2019-09-22
TrialResult addition

MichaelBroughton:
Implemented the addition operation for TrialResults. Only valid when measurement keys and parameter settings are the same. This will make combining smaller jobs as a part of #2107 easier.


maffoo:
Note that we get 1 trial result _per point_ in a sweep, that is, running a sweep returns a `List[cirq.TrialResult]`. We'd only need to split and then recombine individual trial results if you request so many reps that a single trial exceeds job limits, say you ask to run a single circuit many millions of times. In typical cases, the "recombining" step will just be concatenating lists of trials results, one list from each chunk.

That said, I guess it's fine to have a recombination operation for individual trial results to handle this case, but I'd suggest not using `__add__` for this. `TrialResult` is a thin wrapper around a numpy array (or dataframe) of measurement results, and for numpy arrays addition is element-wise, rather than doing concatenation like here. If we extend `TrialResult` to capture data from a sweep (so we don't need to wrap it in a list as we do now), I think we should avoid calling this concatenation operation "addition". Numpy uses "vstack" for this; I think "concatenate" would also be reasonable.

cirq/study/trial_result.py
```diff
@@ -271,3 +271,25 @@ def __eq__(self, other):
...
             return NotImplemented
         return self.data.equals(other.data) and self.params == other.params
+
+    def __add__(self, other: 'TrialResult') -> 'TrialResult':
+        if not isinstance(other, type(self)):
+            return NotImplemented
+        if self.params == other.params and self.measurements.keys(
+        ) == other.measurements.keys():
+            all_measurements: Dict[str, np.ndarray] = {}
+            for key in other.measurements:
```
Strilanc:
```
def _measurement_shape(self):
    return self.params, {k: v.shape[1] for k, v in self.measurements.items()}
```

```
def __add__(self, other: 'cirq.TrialResult') -> 'cirq.TrialResult':
    if self._result_shape() != other._result_shape():
        raise ValueError(...)
    all_measurements = {
        k: np.append(v1, v2, axis=0) 
        for k, v1, _, v2 in zip(self.measurements.items(), 
                                other.measurements.items())
    }
    return ...
```

Strilanc: On second thought, that zip might massively backfire due to ordering.

I wish there was a built-in dictionary pair zip.

cirq/study/trial_result.py
```diff
@@ -271,3 +271,21 @@ def __eq__(self, other):
...
             return NotImplemented
         return self.data.equals(other.data) and self.params == other.params
+
+    def _measurement_shape(self):
+        return self.params, {k: v.shape[1] for k, v in self.measurement.items()}
+
+    def __add__(self, other: 'cirq.TrialResult') -> 'cirq.TrialResult':
+        if not isinstance(other, type(self)):
+            return NotImplemented
+        if self._measurement_shape() != other._measurement_shape():
```
Strilanc: I think you've got the comparison backwards here.

Keep it like it is, but move the exception case to be conditional instead of the correct case. (Make it a guard statement.)

Strilanc:
@maffoo I think it's acceptable to define addition to mean concatenation here. I would personally never have expected adding trial results to do a pointwise addition instead of a concatenation.

cirq/study/trial_result.py
MichaelBroughton: Oops, I found this comment too late, is it OK the way I have it now or would you prefer me to revert back ?

cirq/study/trial_result.py
```diff
@@ -273,12 +273,14 @@ def __eq__(self, other):
...
-        return self.params, {k: v.shape[1] for k, v in self.measurement.items()}
+        return self.params, {
+            k: v.shape[1] for k, v in self.measurements.items()
+        }
 
     def __add__(self, other: 'cirq.TrialResult') -> 'cirq.TrialResult':
         if not isinstance(other, type(self)):
             return NotImplemented
-        if self._measurement_shape() != other._measurement_shape():
+        if self._measurement_shape() == other._measurement_shape():
```
Strilanc: Flip this into a guard statement. "If something is bad then fail", not "if something is good then do the whole thing" followed by a raise at the end.

PR #2149, opened 2019-09-23
Add atol parameter to wavefunction and density matrix utils

### LINKED ISSUE #2148: to_valid_density_matrix is too picky
The density matrix result of `sim.simulate()` regularly fails the assertions in `to_valid_density_matrix`. I had to weaken the tolerances for my code (which worked just fine under the previous implementation that just didn't check)

cc #1901

kevinsung:
Fixes #2148

### DIFF
--- a/cirq/ops/pauli_string.py
+++ b/cirq/ops/pauli_string.py
@@ -255,7 +255,7 @@ class PauliString(raw_types.Operation):
     def expectation_from_wavefunction(self,
                                       state: np.ndarray,
                                       qubit_map: Mapping[raw_types.Qid, int],
-                                      atol: float = 1e-8) -> float:
+                                      atol: float = 1e-7) -> float:
         r"""Evaluate the expectation of this PauliString given a wavefunction.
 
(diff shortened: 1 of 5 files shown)


mpharrigan:
Without looking at it too deeply, I found the default of 1e-8 to be too strict in practice for reasonably sized density matrices

mpharrigan:
Can you include this for the trace and hermitian tests as well; I had to loosen all of them for my usage

kevinsung:
Are you using complex64 or complex128 for your density matrices? I'd expect 1e-8 to work for complex128 but not for complex64.

mpharrigan:
I use whatever `cirq.DensityMatrixSimulator()` gives me

kevinsung:
@mpharrigan I've changed the default to 1e-7. Have you found that to be good enough for your applications? I think we should definitely have a default that works with the default settings of the simulators.

mpharrigan:
This seems to work for me with some brief testing

PR #2155, opened 2019-09-23
Allow random circuit generator to be seeded

kevinsung:


cirq/testing/random_circuit.py
```diff
@@ -72,15 +76,23 @@ def random_circuit(qubits: Union[Sequence[ops.Qid], int],
...
+
     moments: List[ops.Moment] = []
     for _ in range(n_moments):
         operations = []
-        free_qubits = set(q for q in qubits)
+        free_qubits = set(qubits)
         while len(free_qubits) >= max_arity:
-            gate, arity = choice(tuple(gate_domain.items()))
-            op_qubits = sample(free_qubits, arity)
+            gate, arity = tuple(gate_domain.items())[prng.randint(
```
bryano: Seems wasteful to keep recomputing the tuple and length, but probably doesn't matter.

bryano:
LGTM after coverage addressed.

cirq/testing/random_circuit.py
kevinsung: Done.

cirq/testing/random_circuit_test.py
```diff
@@ -67,3 +68,31 @@ def test_random_circuit(n_qubits: Union[int, Sequence[cirq.Qid]],
...
+                              random_state=seed)
+    circuit2 = random_circuit(qubits=20,
+                              n_moments=20,
+                              op_density=0.7,
+                              random_state=seed)
+    assert circuit1 == circuit2
+
+
+@pytest.mark.parametrize('seed', [randint(0, 2**32) for _ in range(10)])
+def test_random_circuit_reproducible_with_random_state(seed):
```
bryano: Given that the parameters of these tests are the same, they should probably just be combined. Something like
```for s in (seed, np.random.RandomState(seed)):```

Comment (kevinsung) [reply cirq/testing/random_circuit_test.py]: I've combined the tests.

File: cirq/testing/random_circuit.py [hunk 3]
```diff
@@ -72,15 +76,24 @@ def random_circuit(qubits: Union[Sequence[ops.Qid], int],
     if n_qubits < 1:
         raise ValueError('At least one qubit must be specified.')
 
+    if isinstance(random_state, np.random.RandomState):
+        prng = random_state
+    else:
+        prng = np.random.RandomState(random_state)
```
mpharrigan:
```python
if random_state is None:
  prng = np.random
```

It's important not to change the default behavior of using the global random number generator. As @Strilanc  pointed out in #2112 , you could potentially get into a situation where repeated calls to `np.random.RandomState(None)` will give you less-than-random random states. For example, if the default seed is the time, if you call it fast enough you'd seed with the same values

kevinsung: Good point. I'll fix this.

PR #2156, opened 2019-09-23
Update errors from Engine API

wcourtney:
- Retry 500 and 503 errors
- Retry ConnectionResetErrors
- Parse HttpError messages to make them more readable to users.


cirq/google/engine/engine.py
```diff
@@ -388,6 +395,31 @@ def _infer_gcs_results(self, job_config: JobConfig) -> None:
...
+                # Raise RuntimeError for exceptions that are not retryable.
+                # Otherwise, pass through to retry.
+                if not err.get('code') in RETRYABLE_ERROR_CODES:
+                    raise RuntimeError(message) from None
+
+            current_delay *= 2
+            if current_delay > self.max_retry_delay:
+                raise Exception('Reached max retry attempts for error: ' +
+                                message)
+            print(message)
```
Strilanc: Where is this being printed? This method may have been called inside a script, it's not appropriate to print like this.

At the very least the output should go to stderr instead of stdout, but I would also add a parameter to control whether or not any printed progress is shown at all.

cirq/google/engine/engine.py
```diff
@@ -388,6 +395,31 @@ def _infer_gcs_results(self, job_config: JobConfig) -> None:
...
+                return request.execute()
+            except ConnectionResetError:
+                message = "Lost connection to the engine."
+            except HttpError as rawErr:
+                err = json.loads(rawErr.content).get('error')
+                message = err.get('message')
+                # Raise RuntimeError for exceptions that are not retryable.
+                # Otherwise, pass through to retry.
+                if not err.get('code') in RETRYABLE_ERROR_CODES:
+                    raise RuntimeError(message) from None
```
Strilanc: Why `from None` instead of `from rawErr`?

Strilanc: `RuntimeError` is an error "that doesn't fall into any category". Can we raise something more specific to the problem?

cirq/google/engine/engine.py
```diff
@@ -388,6 +395,31 @@ def _infer_gcs_results(self, job_config: JobConfig) -> None:
...
+
+            current_delay *= 2
+            if current_delay > self.max_retry_delay:
+                raise Exception('Reached max retry attempts for error: ' +
```
Strilanc: Do not raise `Exception`, raise an error of appropriate type.

cirq/google/engine/engine.py
```diff
@@ -388,6 +395,31 @@ def _infer_gcs_results(self, job_config: JobConfig) -> None:
...
+    def _make_request(self, request: HttpRequest) -> Dict:
+        RETRYABLE_ERROR_CODES = [500, 503]
+        current_delay = 0.1  #100ms
+
+        while True:
+            try:
+                return request.execute()
+            except ConnectionResetError:
+                message = "Lost connection to the engine."
+            except HttpError as rawErr:
```
dabacon: style: raw_err

cirq/google/engine/engine.py
dabacon: style nit: prefer '{}'.format().

cirq/google/engine/engine.py
wcourtney: I was avoiding the "another exception occurred while handling..." messaging. Agreed that raising from rawErr is useful and updated.

cirq/google/engine/engine.py
wcourtney: Updated to TimeoutError

cirq/google/engine/engine.py
```diff
@@ -166,7 +167,8 @@ def __init__(self,
                  discovery_url: Optional[str] = None,
                  default_gcs_prefix: Optional[str] = None,
                  proto_version: ProtoVersion = ProtoVersion.V1,
-                 service_args: Optional[Dict] = None) -> None:
+                 service_args: Optional[Dict] = None,
+                 verbose: Optional[bool] = True) -> None:
```
dabacon: no need to make optional, can just be True or False.

dabacon: add doc string for verbose

cirq/google/engine/engine_test.py
```diff
@@ -14,6 +14,7 @@
 
 """Tests for engine."""
 import base64
+import copy
```
dabacon: not sure why this was added

wcourtney: I added this for json.dumps for generating the error content, but I can just make a string. \Removed.

cirq/google/engine/engine.py
dabacon: I'd create a new type of exception that makes it possible for people to catch exceptions of this form and not more general runtime errors than can be used.  EngineException is probably good.

cirq/google/engine/engine.py
```diff
@@ -388,6 +399,32 @@ def _infer_gcs_results(self, job_config: JobConfig) -> None:
             job_config.gcs_results = '{}jobs/{}'.format(job_config.gcs_prefix,
                                                         job_config.job_id)
 
+    def _make_request(self, request: HttpRequest) -> Dict:
+        RETRYABLE_ERROR_CODES = [500, 503]
```
dabacon: nit: since this isn't a global constant this should just be lower case snake case

cirq/google/engine/engine.py
wcourtney: SG, I added an EngineException and am returning that instead.

dabacon:
You've got some uncovered lines

cirq/google/engine/engine_test.py
```diff
@@ -1140,3 +1141,46 @@ def test_sampler(build):
...
+def test_api_doesnt_retry_404_errors(build):
+    service = mock.Mock()
+    build.return_value = service
+    getProgram = service.projects().programs().get()
+    content = '{"error": {"message": "not found", "code": 404}}'.encode('utf-8')
+    getProgram.execute.side_effect = HttpError(mock.Mock(), content)
+    engine = cg.Engine(project_id='project-id')
+    with pytest.raises(cg.engine.engine.EngineException, match='not found'):
+        engine.get_program('foo')
+        assert getProgram.execute.call_count == 1
```
dabacon: this needs to be un-indented (also two more below)

wcourtney: Fixed, thanks!

PR #2157, opened 2019-09-23
Cleanup unused type imports

viathor:
Found by pylint.


bryano:
Why didn't pylint catch this in the original PR?

vtomole:
Unused imports are caught. We don't error on them though.

PR #2158, opened 2019-09-24
Change compute_amplitudes input format from 2-d array of bools to 1-d array of ints

### LINKED ISSUE #2151: Change input to SimulatesAmplitudes.compute_amplitudes from 2-d numpy array to 1-d int sequence
@viathor are you doing this in #2114 ? If not I can do it.

kevinsung:
Fixes #2151


cirq/sim/simulator.py
```diff
@@ -191,20 +192,21 @@ class SimulatesAmplitudes(metaclass=abc.ABCMeta):
     def compute_amplitudes(
             self,
             program: Union[circuits.Circuit, schedules.Schedule],
-            bitstrings: np.ndarray,
+            bitstrings: Sequence[int],
```
Strilanc: This is a silent breaking change. Add code to detect the wrong type being passed in and raise an appropriate exception.

cirq/sim/simulator.py
```diff
@@ -221,20 +223,21 @@ def compute_amplitudes(
...
         Args:
             program: The circuit or schedule to simulate.
             bitstrings: The bitstrings whose amplitudes are desired, input
-                as a two-dimensional array of bools. The first dimension
-                indexes the bitstrings and the second dimension indexes
-                the bits within a bitstring.
+                as an integer array where each integer is formed from measured
+                qubit values according to `qubit_order` from most to least
+                significant qubit, i.e. in the order consistent with
+                `cirq.final_wavefunction`.
```
Strilanc: We shouldn't keep referring to final wavefunction for this. Just describe what the ordering is in more detail here.

cirq/sim/wave_function_simulator.py
```diff
@@ -164,27 +165,23 @@ def compute_displays_sweep(
...
                    program.to_circuit())
-
         trial_results = self.simulate_sweep(circuit, params, qubit_order)
 
-        qid_shape = circuit.qid_shape(qubit_order=qubit_order)
-        amplitude_indices = [
-            value.big_endian_digits_to_int(bitstring, base=qid_shape)
-            for bitstring in bitstrings
-        ]
+        if isinstance(bitstrings, tuple):
```
Strilanc: What about all the other possible types of sequence? Why is tuple being handled specially?

cirq/sim/simulator.py
kevinsung: I've added the code to `SimulatesIntermediateWaveFunction.compute_amplitudes_sweep`.

cirq/sim/simulator.py
kevinsung: I've changed the wording slightly.

cirq/sim/wave_function_simulator.py
kevinsung: 1-dimensional tuples are handled specially by numpy when used as an array index: https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#advanced-indexing

"Advanced indexing is triggered when the selection object, obj, is a non-tuple sequence object ..."

Strilanc: Comment that in the code.

kevinsung: Done.

kevinsung: Done.

cirq/sim/wave_function_simulator.py
```diff
@@ -164,27 +165,30 @@ def compute_displays_sweep(
...
+            bitstrings = list(bitstrings)
 
         all_amplitudes = []
         for trial_result in trial_results:
             # mypy doesn't know that this trial result has a final_state
             # attribute
             final_state = trial_result.final_state  # type: ignore
-            amplitudes = [final_state[index] for index in amplitude_indices]
+            amplitudes = final_state[bitstrings]
```
Strilanc: You can use `cast(type, value)` to convince mypy that it's a particular kind of trial result that does have a final state

kevinsung: Done.

PR #2162, opened 2019-09-24
Add _compat.deprecate_parameter

Strilanc:
- Add _compat_test.py and unit test methods in _compat
- Fix proper_repr not preserving the dtype of numpy arrays


cirq/_compat.py
```diff
@@ -83,3 +85,64 @@ def decorated_func(*args, **kwargs) -> Any:
...
+            nonlocal used
+            if match(args, kwargs):
+                if rewrite is not None:
+                    args, kwargs = rewrite(args, kwargs)
+
+                if not used:
+                    used = True
+                    qualname = (func.__qualname__
+                                if func_name is None else func_name)
+                    logging.warning(
```
mpharrigan: Why `logging.warning` and not `warnings.warn`? If I understand correctly, this will almost never be printed unless the user's script calls `logging.basicConfig(level=logging.WARNING)`

mpharrigan: `warning.warn("text", DeprecationWarning)` is (sometimes) correctly parsed by pycharm and shows the deprecated function with a strikethrough font and includes the message. I don't know if it's possible to trigger this behavior for deprecated parameters (I know it works for functions) but that would be really nice

Strilanc: Seems to work by default for me.

```
$ python -c "import logging; logging.warning('TEST')"
WARNING:root:TEST
```

Also works in the python and ipython repls by default.

mpharrigan: ok so it does show up, but pycharm doesn't cross it out. I think `warnings.warn("message", DeprecationWarning)` is more common. For example, `logging.warn` uses this form :)

cirq/_compat.py
```diff
@@ -83,3 +85,64 @@ def decorated_func(*args, **kwargs) -> Any:
...
+        deadline: The version where the parameter will be deleted (e.g. "v0.7").
+        fix: A complete sentence describing what the user should be using
+            instead of this particular function (e.g. "Use cos instead.")
+        func_name: How to refer to the function.
+            Defaults to `func.__qualname__`.
+        parameter_desc: The name and type of the parameter being deprecated,
+            e.g. "janky_count" or "janky_count keyword" or
+            "positional janky_count".
+        match: Detects if the deprecated parameter is present in the
+            args/kwargs.
```
dstrain115: I think this could be explained better.  I had to go to the tests to understand how this is supposed to work.  I totally didn't understand that the match parameter was the triggering condition for the parameter.  Maybe add some examples?

cirq/_compat.py
Strilanc: For me, `python -c "import warnings; warnings.warn('x', DeprecationWarning)"` is the one that produces no output.

cirq/linalg/decompositions_test.py
```diff
@@ -310,22 +310,23 @@ def test_kak_repr():
...
-        np.array([[0j, (1+0j)], [(1+0j), 0j]]),
-        np.array([[0j, -1j], [1j, 0j]]),
+        np.array([[0j, (1+0j)], [(1+0j), 0j]], dtype=np.complex128),
+        np.array([[0j, -1j], [1j, 0j]], dtype=np.complex128),
     ),
     single_qubit_operations_after=(
-        np.array([[1.0, 0.0], [0.0, 1.0]]),
-        np.array([[(1+0j), 0j], [0j, (-1+0j)]]),
+        np.array([[1.0, 0.0], [0.0, 1.0]], dtype=np.float64),
+        np.array([[(1+0j), 0j], [0j, (-1+0j)]], dtype=np.complex128),
```
dstrain115: What's up with these kak changes?  They seem unrelated.

cirq/_compat.py
Strilanc: Won't that form result in a warning every time you call the method, instead of only the first time?

```
    def warn(self, msg, *args, **kwargs):
        warnings.warn("The 'warn' method is deprecated, "
            "use 'warning' instead", DeprecationWarning, 2)
        self.warning(msg, *args, **kwargs)
```

Strilanc: When I use `warnings.warn` in the function, pycharm doesn't start crossing out deprecated functions. I think there's too much separation between the usage inside of the decorator and the function being deprecated by the decorator.

cirq/linalg/decompositions_test.py
Strilanc: It's due to the proper_repr fix for numpy arrays, caught by adding tests that it wasn't including the dtype

cirq/_compat.py
Strilanc: Done

cirq/_compat.py
Strilanc: I'm going to leave it as is for now. Feel free to do a follow-up that uses `warnings.warn` instead, if you can get pycharm to underline the deprecated methods. That would be really useful.

mpharrigan: I'm no expert, nor do I really want to become one; but `DeprecationWarning`s get printed only once and there's some complicated behavior for when they get turned on or off. IIRC there's some heuristic that tries to guess if you're running the code as a "developer" or "user"  and it only shows it to users. I'm fine with logging. I don't think pycharm is sophisticated enough to figure out deprecations from decorators

PR #2168, opened 2019-09-24
Replace usage of cirq.Circuit.from_ops(...) with cirq.Circuit(...) in circuits,work,dev_tools

Strilanc:
Part of #1787

This should be nothing but basically a search-replace of `Circuit.from_ops(` with `Circuit(`, followed by formatting, followed by tiny manual tweaks.


mpharrigan:
Is it worth leaving one usage of `from_ops` in to check that the deprecated behavior is still functional?

Strilanc:
I left some tests in `circuit_tests.py`. A future change renames the test to include `deprecated`.

Strilanc:
@mpharrigan I take it back, I accidentally replaced those from_ops also. Fixed it now.

PR #2171, opened 2019-09-24
Replace usage of cirq.Circuit.from_ops(...) with cirq.Circuit(...) in documentation

Strilanc:
Part of #1787

This should be nothing but basically a search-replace of Circuit.from_ops( with Circuit(, followed by formatting, followed by tiny manual tweaks.

### DIFF
--- a/docs/circuits.md
+++ b/docs/circuits.md
@@ -324,8 +324,8 @@ A very nice pattern emerges from this structure: define
 or `Operation` parameters.
 
 Another useful method is to construct a `Circuit` fully formed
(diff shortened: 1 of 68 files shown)


docs/circuits.md
```diff
@@ -328,7 +328,7 @@ from an ``OP_TREE`` via the static method `Circuit.from_ops`
 (which takes an insertion strategy as a parameter):
```
MichaelBroughton: You missed a from_ops in this file. I can't comment on the line, but it's right above here.

Strilanc: That's a pretty significant one; the example didn't match the description anymore

PR #2173, opened 2019-09-25
Phased ISWAP gate

viathor:
Slight generalization of OpenFermion-cirq's YXXY, which is going away.


cirq/ops/phased_iswap_gate.py
```diff
@@ -0,0 +1,197 @@
...
+        c = cos(π·t/2)
+        s = sin(π·t/2)
+        f = exp(2πi·p)
+
+    and star indicates complex conjugate.
+    """
+
+    def __init__(self,
+                 *,
+                 phase_exponent: Union[float, sympy.Symbol] = 1 / 4,
```
Strilanc: It defaults to the fourth root of the iswap?

cirq/ops/phased_iswap_gate.py
```diff
@@ -0,0 +1,197 @@
...
+                 exponent: Union[float, sympy.Symbol] = 1.0,
+                 global_shift: float = 0.0) -> None:
```
Strilanc: Drop global shift. It's been troublesome in eigen gate. It's also a bit ambiguous whether it applies to the inner 2x2 matrix or the entire matrix.

cirq/ops/phased_iswap_gate.py
```diff
@@ -0,0 +1,197 @@
...
+
+import cirq
+from cirq import value, protocols
+from cirq._compat import proper_repr
+from cirq.ops import gate_features, raw_types, op_tree, common_gates
+from cirq.value import type_alias
+
+
+@value.value_equality(manual_cls=True)
+class PhasedISwapPowGate(common_gates.ISwapPowGate, gate_features.TwoQubitGate):
```
Strilanc: I am worried about this parent-child relationship. PhasedISwapPowGate cannot actually be substituted into all places that expect an ISwapPowGate. For example, I could have code that says "if it's an iswap then it commutes with XX", but this is actually not true of a phased iswap pow gate.

cirq/ops/phased_iswap_gate.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def _apply_unitary_(self, args: 'protocols.ApplyUnitaryArgs'
+                       ) -> Optional[np.ndarray]:
+        if protocols.is_parameterized(self):
+            return NotImplemented
+        if self.exponent != 1:
+            return NotImplemented
+
+        phase = np.exp(1j * np.pi * self.phase_exponent)
+        zo = args.subspace_index(0b01)
```
Strilanc: You can use `cirq.linalg.apply_matrix_to_slices` to operate on the 2x2 inner block. It will be simpler and faster.

cirq/ops/phased_iswap_gate.py
viathor: I considered and implemented two other approaches:
 * composition: putting an ISwapPowGate into `self._iswap`,
 * no direct dependency on ISwapPowGate.

In all cases, I'd like PhasedISwapPowGate to be an EigenGate since it gets us lots of things for free and makes sense conceptually (eigendecomposition is easy(*) for this gate). However, a consequence of this is that we have to store the exponent in `self._exponent`. So when you go with composition then you duplicate it between `self._iswap._exponent` and `self._exponent`.

When you remove direct dependency on ISwapPowGate, then you either end up duplicating its eigen-projectors or you end up accessing private method via `ISwapPowGate._eigen_components`.

WDYT about making `EigenGate._eigen_components` public? I think this would be the most elegant solution: doesn't break the is-a relationship (as the code in this PR does), doesn't duplicate anything and it does in fact make sense that EigenGates expose their eigendecomposition publically in some form.

----
(*) In general, if you know eigendecomposition of A then you can easily derive eigendecomposition of BAB^-1. Since conjugation is a similarity transform, eigenvalues don't change. If P is a projector on an eigenspace of A, then BPB^-1 is a projector on the corresponding eigenspace of BAB^-1.

Strilanc: This still wouldn't address the problem that a phased iswap pow gate *isn't an iswap gate*. It is outside of the reference class established by the parent. It violates the Liskov substitution principle. I cannot safely pass this child class into any method that would be expecting the parent class. In my opinion this is a very serious problem.

viathor: I agree. Done.

cirq/ops/phased_iswap_gate.py
```diff
@@ -0,0 +1,202 @@
...
+            args.append(f'global_shift={global_shift}')
+        arg_string = ', '.join(args)
+        return f'cirq.PhasedISwapPowGate({arg_string})'
+
+
+# The ISWAP gate conjugated by T⊗T^-1.
+#
+# Matrix of GivensRotation^t is
+#
+#     [[1, 0, 0, 0],
```
Strilanc: How is this equation affected by the global shift?

cirq/ops/phased_iswap_gate.py
viathor: Quarter of the phase, i.e. the T gate (or its inverse) on each qubit. This is the most interesting case since it gives half a full circuit phase difference between off-diagonal elements of the inner block (i.e. YXXY gate).

Explained in comment.

cirq/ops/phased_iswap_gate.py
viathor: Done.

cirq/ops/phased_iswap_gate.py
viathor: Unfortunately, `apply_matrix_to_slices` cannot apply the matrix in-place (and in our case, we're just multiplying in some phase into each row and column - something easily done in place as it is done currently).

cirq/ops/phased_iswap_gate.py
viathor: Well, it'd gain phase, but we've removed global shift, so this is now moot.

cirq/ops/phased_iswap_gate.py
```diff
@@ -0,0 +1,193 @@
...
+# where
+#
+#     c = cos(π·t/2),
+#     s = sin(π·t/2).
+#
+# Equivalently, it may be defined as
+#
+#     GivensRotation^t ≡ exp(-i π t (Y⊗X - X⊗Y) / 4)
+#
+GivensRotation = PhasedISwapPowGate()
```
Strilanc: Should this be a function that takes a rotation angle in radians? I've noticed the theory team prefers functions like Rx which take radians.

cirq/ops/phased_iswap_gate.py
```diff
@@ -0,0 +1,193 @@
...
+        if self.exponent != 1:
+            exponent = proper_repr(self.exponent)
+            args.append(f'exponent={exponent}')
+        arg_string = ', '.join(args)
+        return f'cirq.PhasedISwapPowGate({arg_string})'
+
+
+# The ISWAP gate conjugated by T⊗T^-1.
+#
+# Matrix of GivensRotation^t is
```
Strilanc: Distinguish between the linear algebra meaning and the chemistry meaning, if applicable.

dabacon:
Slight aside: I think at the Cirq Cync there should be a discussion about gates.  We are adding gates in a sort of adhoc manner and I worry that we are creating legacy bags of gates that really should all be part of a framework for our different types of gates.

cirq/ops/phased_iswap_gate.py
viathor: Good catch, done.

cirq/ops/phased_iswap_gate.py
viathor: Done.

viathor:
@dabacon I very much agree with your concern. Therefore, in adding this specific gate (which in a different form already exists in OpenFermion-cirq and needs to be moved to clean unwanted dependency) I tried to make the new gate fit into cirq's existing gate repertoire as tightly as possible. Specifically, its relationship to ISwapPowGate is modelled on PhasedXPowGate's relationship to XPowGate. This BTW is the source of the name.

Strilanc:
Personally, I am in favor of adding lots and lots of gates because it will give us a reference to use when designing frameworks that can encode them all. I want custom gates created by clients to integrate more seamlessly into the library, and I assume the number of such custom gates is effectively gigantic.

viathor:
Problems that accompany a Cambrian explosion of gates are easier to deal with if we have some structure in place to represent relationships between gates. This is one reason to look for ways to represent things like tensor products, compositions, groups etc.

Even without going this far, gate compilation needs at least to be able to check gate equivalence. What I have seen in practice is people use `cirq.op_gate_isinstance`, but if we are to allow proliferation of gates then we should discourage gate equivalence tests based on types and instead offer a tool to check gate equivalence as members of the unitary group.

Strilanc:
I love to have something better to replace `isinstance` with.

viathor:
Filed #2182.

cirq/ops/phased_iswap_gate.py
Strilanc: Current code (arbitrary exponent): 3.9s
Current code (exponent=1): 0.47s
Apply to slices code (arbitrary exponent): 0.41s to 0.48s

New code:

```python
    def _apply_unitary_(self, args: 'protocols.ApplyUnitaryArgs'
                       ) -> Optional[np.ndarray]:
        if protocols.is_parameterized(self):
            return NotImplemented

        c = np.cos(np.pi * self._exponent / 2)
        s = np.sin(np.pi * self._exponent / 2)
        f = np.exp(2j * np.pi * self._phase_exponent)
        matrix = np.array([
            [c, 1j*s*f],
            [1j*s*f.conjugate(), c]
        ])

        zo = args.subspace_index(0b01)
        oz = args.subspace_index(0b10)
        linalg.apply_matrix_to_slices(
            args.target_tensor,
            matrix,
            [oz, zo],
            out=args.available_buffer)
        return args.available_buffer
```

Test case:

```python
import cirq

qubits = cirq.LineQubit.range(18)
c = cirq.Circuit(
    cirq.PhasedISwapPowGate(exponent=0.1432, phase_exponent=0.6542).on(a, b)
    for a in qubits
    for b in qubits
    if a != b
)

import time
t0 = time.monotonic()
cirq.final_wavefunction(c)
t1 = time.monotonic()
print(t1 - t0)
```

cirq/ops/phased_iswap_gate.py
```diff
@@ -0,0 +1,203 @@
...
+        return args.target_tensor
+
+    def _decompose_(self, qubits: Sequence[raw_types.Qid]) -> op_tree.OP_TREE:
+        if len(qubits) != 2:
+            raise ValueError(f'Expected two qubits, got {len(qubits)}')
+        a, b = qubits
+
+        yield cirq.Z(a)**self.phase_exponent
+        yield cirq.Z(b)**-self.phase_exponent
+        yield from protocols.decompose_once_with_qubits(self._iswap, qubits)
```
Strilanc: You can yield the iswap instead of its decomposition.

cirq/ops/phased_iswap_gate.py
```diff
@@ -0,0 +1,203 @@
...
+#
+#     c = cos(a),
+#     s = sin(a).
+#
+# The matrix is a Givens rotation in the numerical linear algebra sense
+# acting on the subspace spanned by the |01⟩ and |10⟩ states.
+#
+# The gate is also equivalent to the ISWAP conjugated by T^-1 ⊗ T.
+#
+def GivensRotation(angle_rads: value.TParamVal) -> PhasedISwapPowGate:
```
Strilanc: Conflicted about the casing here. Per PEP should be `givens_rotation`... Hmm.

viathor: We kind of violated the PEP with gate factories such as `Rz`. I think we did that to make those not stand out among other gates, e.g. in a circuit or decomposition. I agree the longer name here makes the PEP violation more visible, but at the same time `givens_rotation` would stand out among other gates even more. Compare:

```python
yield Rx(np.pi/2).on(a)
yield Ry(np.pi/2).on(b)
yield GivensRotation(np.pi/8).on(a, b)
yield CNOT(a, b)
```

and

```python
yield Rx(np.pi/2).on(a)
yield Ry(np.pi/2).on(b)
yield givens_rotation(np.pi/8).on(a, b)
yield CNOT(a, b)
```

I don't have a strong preference, but I think the former looks better.

Strilanc: Yes, that is the source of the "Hmmm". Leave it as is.

cirq/ops/phased_iswap_gate.py
viathor: Very cool! Done.

PR #2175, opened 2019-09-25
Added QRW Tutorial

Lucaman99:
The other branch refused to cooperate, so here is a new one that should work.


Strilanc:
Hahaha, it was a good try. It looks like there's actually three email addresses in play? For example, the commits say that "Jack Ceroni" authored it but separately "Jack Ceroni" submitted it, but neither has your favicon on it so they aren't associated with your github account.

PR #2177, opened 2019-09-25
Fix ISWAP trace_distance_bound

### LINKED ISSUE #2172: Squared ISWAP fails trace distance bound test
```
In [1]: import cirq

In [2]: cirq.testing.consistent_protocols.assert_has_consistent_trace_distance_bound(cirq.ISWAP**2)
AssertionError                            Traceback (most recent call last)
<ipython-input-2-9d7942dfa2d6> in <module>
----> 1 cirq.testing.consistent_protocols.assert_has_consistent_trace_distance_bound(cirq.ISWAP**2)

~/src/Cirq/cirq/testing/consistent_protocols.py in assert_has_consistent_trace_distance_bound(val)
    109 
    110         assert val_from_trace >= val_from_unitary or np.isclose(
--> 111             val_from_trace, val_from_unitary)
    112 
    113 

AssertionError: 

In [3]: cirq.trace_distance_bound(cirq.ISWAP**2)
Out[3]: 1.2246467991473532e-16
```

I guess we really care about this near identity, but still the protocol explicitly says "Underestimates are not permitted.", so we should fix it in case it's relied on in future.

Other exponents in [1.0029..2.9971] also fail the check.

c-poole:
There was a special case that was applied in error to ISWAP that has now been removed. Added tests to verify that problem this fixes #2172.

### DIFF
--- a/cirq/ops/common_gates.py
+++ b/cirq/ops/common_gates.py
@@ -1287,11 +1287,6 @@ class ISwapPowGate(eigen_gate.EigenGate,
                              [0, 0, 0, 0]])),
         ]
 
-    def _trace_distance_bound_(self) -> Optional[float]:
-        if self._is_parameterized_():
-            return None
-        return abs(np.sin(self._exponent * 0.5 * np.pi))
-
     def _decompose_(self, qubits):
         a, b = qubits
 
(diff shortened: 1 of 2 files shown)


Strilanc:
The test failed on the old code, right?

viathor:
@Strilanc Yeah, the new test fails on master.

PR #2181, opened 2019-09-25
Deprecate `cirq.Circuit.from_ops`

### LINKED ISSUE #1787: Syntactic sugar for cirq.Circuit.from_ops (?)

Strilanc:
- Replace all remaining usage with `cirq.Circuit(*ops)`
- Test deprecation of other things on Circuit (and make a few fixes)

Fixes #1787


cirq/circuits/circuit.py
```diff
@@ -95,7 +95,7 @@ class Circuit:
 
     @deprecated_parameter(
         deadline='v0.8',
-        fix='Pass contents positionally or using the "contents=" keyword.',
+        fix='Pass circuit contents positionally (without a keyword).',
```
mpharrigan: explain?

Strilanc: This was an oversight on my part when writing the original message. It turns out you can't pass `*args` arguments by name.

mpharrigan: oh duh. For some reason my brain kept reading it as `**contents` in the signature

PR #2184, opened 2019-09-25
Fix lack-of-aliasing bug in ArithmeticOperation

### LINKED ISSUE #2160: ArithmeticOperation apply unitary assumes reshape always returns a view
This is not true. In some cases reshape returns a copy. This can result in undefined data going into the result, because the workspace buffer is never initialized.

The safe way to reshape and get an error if it's not a mutable view is to make an identical view `view = buffer[...]` and then *set* its shape `view.shape = shape`.

The code needs to be fixed to work in the case where reshape would fail, and a test needs to be added to trigger this case. I ran into this as part of https://github.com/quantumlib/Cirq/pull/2137 .

Repro:

```
def test_reshape_referencing():
    class Op1(cirq.ArithmeticOperation):
        def apply(self, *register_values: int):
            return register_values[0] + 1

        def registers(self):
            return [cirq.LineQubit.range(2)[::-1]]

        def with_registers(self, *new_registers):
            raise NotImplementedError()

    state = np.ones(4, dtype=np.complex64) / 2
    output = cirq.final_wavefunction(
        cirq.Circuit.from_ops(Op1()),
        initial_state=state)
    np.testing.assert_allclose(state, output)
```

Results in:

```
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       Mismatch: 100%
E       Max absolute difference: 0.5
E       Max relative difference: inf
E        x: array([0.5+0.j, 0.5+0.j, 0.5+0.j, 0.5+0.j], dtype=complex64)
E        y: array([0.+0.j, 1.+0.j, 0.+0.j, 0.+0.j], dtype=complex64)
```

But any permutation of the all-plus initial state should be the all-plus state.

Strilanc:
- Also add standard test `_assert_apply_unitary_works_when_axes_transposed` for all gates and operations
- Also delete stale "qubit_count" arguments from standard test battery methods (obsoleted by qid_shape protocol)
- Remove unused `assert_eigen_gate_has_consistent_apply_unitary`

Fixes #2160


Strilanc:
I think it's because the axes are always in order in the battery of tests. I'll modify this to add an out-of-order axes case to the test, where the first axis is moved to the end and is not contiguous with the rest of the axes.

Strilanc:
@viathor PTAL. I added a test method to the standard battery of tests we use on gates. I checked that it in fact fails on the old arithmetic gate code.

cirq/testing/circuit_compare.py
```diff
@@ -305,6 +308,75 @@ def assert_has_consistent_apply_unitary(
...
+    random.shuffle(permutation)
+    transposed_shape = [0] * padded_n
+    for i in range(padded_n):
+        transposed_shape[permutation[i]] = padded_shape[i]
+
+    # Prepare input states.
+    in_order_input = lin_alg_utils.random_superposition(size).reshape(
+        padded_shape)
+    in_order_input = np.array(range(1, size+1), dtype=np.complex128).reshape(
+        padded_shape)
```
viathor: You probably meant to remove this line.

cirq/testing/circuit_compare.py
```diff
@@ -305,6 +308,75 @@ def assert_has_consistent_apply_unitary(
...
+    out_of_order_input = np.empty(shape=transposed_shape, dtype=np.complex128)
+    out_of_order_input.transpose(permutation)[...] = in_order_input
+
+    # Apply to in-order and out-of-order axes.
+    in_order_output = protocols.apply_unitary(
+        val,
+        protocols.ApplyUnitaryArgs(
+            in_order_input,
+            np.empty_like(in_order_input),
```
viathor: nit: use keyword args for target_tensor and available_buffer for consistency with the call below (or remove the keywords in the code below).

cirq/testing/circuit_compare.py
```diff
@@ -305,6 +308,75 @@ def assert_has_consistent_apply_unitary(
                                    atol=atol)
 
 
+def _assert_apply_unitary_works_when_axes_transposed(
```
viathor: This is a nice addition to the battery. Could you add tests for this assert? We do this for a number of other existing ones.

cirq/testing/circuit_compare.py
```diff
@@ -305,6 +308,75 @@ def assert_has_consistent_apply_unitary(
...
+        val: The operation, gate, or other unitary object to test.
+        atol: Absolute error tolerance.
+    """
+
+    # Only test custom apply unitary methods.
+    if not hasattr(val, '_apply_unitary_') or not protocols.has_unitary(val):
+        return
+
+    # Pick sizes and shapes.
+    shape = protocols.qid_shape(val)
```
viathor: This fails for schedules whose unitaries we appear to be testing, too. Either skip or make it work in this case(*). I think this is why the tests are failing.

(*) For example, by adding `_num_qubits_` or `_qid_shape_` to `Schedule`.

cirq/testing/circuit_compare.py
Strilanc: Added a failure triggering test.

cirq/testing/circuit_compare.py
Strilanc: I have expanded the PR to delete all the "qubit count" parameters out of the testing methods, and have them all use the qid_shape protocol instead. I fixed all test objects which did not specify this method. The tests predated the qid shape protocol existing, but now that it does we should of course use it.

cirq/testing/circuit_compare.py
Strilanc: Done.

cirq/testing/circuit_compare.py
Strilanc: Good catch. Done.

cirq/testing/circuit_compare_test.py
```diff
@@ -543,3 +554,38 @@ class NoProtocol:
...
+            view[1, ...] *= 1j
+            view[2, ...] *= -1
+            view[3, ...] *= -1j
+            return args.target_tensor
+
+        def _num_qubits_(self):
+            return 2
+
+    assert cirq.has_unitary(BadOp())
+    bad_op = BadOp()
```
viathor: nit: no need to instantiate two, re-use.

cirq/testing/circuit_compare_test.py
```diff
@@ -543,3 +554,38 @@ class NoProtocol:
...
+
+    # Appears to work.
+    np.testing.assert_allclose(cirq.unitary(bad_op), np.diag([1, 1j, -1, -1j]))
+    # But fails the more discerning test.
+    with pytest.raises(AssertionError,
+                       match='acted differently on out-of-order axes'):
+        for _ in range(100):  # Axis orders chosen at random. Brute force a hit.
```
viathor: This test should be deterministic. As is, it risks accidental random test failures and adds a bit to the creep of tests taking longer than necessary.

AFAICT, it's just a matter of finding and setting the seed that triggers the assert failure.

(Also, it's worth using getstate() and setstate() around the logic running with a fixed seed.)

Strilanc: A quick numerical check indicates that an individual run will fail about 90% of the time. That suggests the chance 100 sequential checks failing is around 1 in 10^100. To my tastes that might as well be certain. In fact it's very likely that there is no failing random seed and it really *is* certain.

This is far less brittle than picking a seed and running one check, because picking a seed exposes us to the risk of the internals of the testing method changing how they generate the randomness.

viathor: Fair enough.

cirq/testing/circuit_compare_test.py
Strilanc: Done.

Issue #2180, opened 2019-09-25
PauliString should accept List[Operation] in constructor


JulianSKelly:
Usage is a bit awkward that we need a dict in the constructor. Specifically:

```python
q = cirq.NamedQubit('q0')
pauli_string = cirq.PauliString({q: cirq.X})
```
But this doesn't work
```python
q = cirq.NamedQubit('q0')
pauli_string=cirq.PauliString([cirq.X(q)])
```

JulianSKelly:
I'm going to give fixing this a shot for my own edification.

Strilanc:
Assigned it to you.

JulianSKelly:
@Strilanc what do you think about `from_single` accepting a single `Operation` argument over what it takes now:

https://github.com/quantumlib/Cirq/blob/116fbf5480c40a04af64aad4cb006db3adb5ffac/cirq/ops/pauli_string.py#L57-L61

Strilanc:
I think that this change to the constructor should deprecate that method. If you're feeling ambitious, `from cirq._compat import deprecate` and decorate `from_single` with it.

Strilanc:
And you'd set "fix" to like "call `cirq.PauliString([op])` instead"

PR #2186, opened 2019-09-26
Add List[op] to PauliString constructor, deprecate from_single

JulianSKelly:
Fixes: #2180


Strilanc:
Looks like there's a call to from_single left in the code base. It's in contrib/paulistirng/clifford_optimize.py ; that's why the contrib-less tests passed. Only the first invocation of the method triggers the logging, and the contrib test happens to run before the deprecated test.

Strilanc:
Obsoleted by https://github.com/quantumlib/Cirq/pull/2232

PR #2138, opened 2019-09-20
[WIP] Improve circuit export to Quirk URLs

Strilanc:
This is working towards making it possible to take a Quirk URL and produce a simulable `cirq.Circuit` instance. This involves defining quite a lot of gates; I've tried to do it in a minimalist fashion.

To give a sense, each arithmetic gate is taking up about one line (with a decent amount of machinery hiding to make this possible):

```python
    yield from reg_arithmetic_family("^=A", lambda x, a: x ^ a)
    yield from reg_arithmetic_family("+cntA", lambda x, a: x + popcnt(a))
    yield from reg_arithmetic_family("-cntA", lambda x, a: x - popcnt(a))
    yield from reg_arithmetic_family(
        "Flip<A", lambda x, a: a - x - 1 if x < a else x)
    yield from reg_arithmetic_family(
        "*AmodR", lambda x, a, r: (x * a) % r
        if x < r and modular_multiplicative_inverse(a, r) else x)
```

These gates will not be exposed at the top level. They only exist to make parsing possible.

PR #2188, opened 2019-09-26
Define cirq.contrib.quirk.quirk_url_to_circuit

Strilanc:
- Support single qubit rotation gates
- Support controls
- Support qubit permutations
- Define the 'Cell' type, which represents an entry in Quirk's grid (including operation modifiers and displays)
- Include assert_url_to_circuit_returns testing utility method
- Basic formula string and matrix string parsing

Other operations coming in later PRs

Pulled out of https://github.com/quantumlib/Cirq/pull/2138


Strilanc:
This is a gigantic 1600 line PR. I cut it down as much as I could without it being so trivial that it wasn't clear how the pieces would fit together.

cirq/contrib/quirk/cells/parse.py
```diff
@@ -0,0 +1,94 @@
...
+    "⅛": "(1/8)",
+    "⅜": "(3/8)",
+    "⅝": "(5/8)",
+    "⅞": "(7/8)",
+    "⅑": "(1/9)",
+    "⅒": "(1/10)",
+}
+
+
+def expand_unicode_fractions(text: str) -> str:
```
viathor: nit: private `_expand_unicode_fractions`?

cirq/contrib/quirk/cells/parse.py
```diff
@@ -0,0 +1,94 @@
...
+    for k, v in UNICODE_FRACTIONS.items():
+        text = text.replace('√' + k, f'(sqrt{v})')
+        text = text.replace(k, v)
+    return text
+
+
+def parse_matrix(text: str) -> np.ndarray:
```
viathor: Remove dead code? (Fine to leave if you're anticipating usage soon.)

cirq/contrib/quirk/cells/parse.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def _parse_complex(text: str) -> complex:
+    try:
+        if (text.endswith('i') and len(text) > 1 and not text.endswith('+i') and
+                not text.endswith('-i')):
+            text = text[:-1] + '*i'
+        expr = sympy.parsing.sympy_parser.parse_expr(text)
+        return complex(expr.subs({'i': 1j}))
+    except Exception as ex:
```
viathor: Is there a better, more concrete exception type we can use here (or multiple types)?

cirq/contrib/quirk/cells/parse.py
```diff
@@ -0,0 +1,94 @@
...
+    return np.array([[_parse_complex(c) for c in row.split(',')] for row in rows
+                    ])
+
+
+def _parse_complex(text: str) -> complex:
+    try:
+        if (text.endswith('i') and len(text) > 1 and not text.endswith('+i') and
+                not text.endswith('-i')):
+            text = text[:-1] + '*i'
+        expr = sympy.parsing.sympy_parser.parse_expr(text)
```
viathor: nit: `parse_expr(text)` since you're using import-from statement above

cirq/contrib/quirk/cells/parse.py
```diff
@@ -0,0 +1,94 @@
...
+        return complex(expr.subs({'i': 1j}))
```
viathor: You don't need to make the substitution. Sympy [supports](https://docs.sympy.org/latest/_modules/sympy/parsing/sympy_parser.html) imaginary unit using capital i. Something like this should work and also avoids the need for preprocessing by the conditional statement above (l.87-89):

```python
return complex(parse_expr(text.replace('+i', '+I').replace('-i', '-I').replace('i', '*I'))) 
```

(That said, I really think we should reconsider using such a versatile parser, so this is probably moot.)

cirq/contrib/quirk/cells/parse.py
viathor: Note that using this parser means that a lot of math expressions will work, which may be more than you're expecting. Examples: '(-1)**(1/2)', '100!', 'cos(3.14159265)'. Some expressions are inadvertently broken by the manipulation that makes 'i' the imaginary unit.

More interestingly, you can trick it easily into performing heavy computations by passing something like '2**1000000000000', so it opens up a way to a potential DoS attack once this ends up on a codepath serving some externally controlled request.

It'd be safer to use a dumber, more targeted tool instead.

cirq/contrib/quirk/cells/qubit_permutation_cells.py
```diff
@@ -0,0 +1,121 @@
...
+    CellMaker,
+    ExplicitOperationsCell,
+)
+
+if TYPE_CHECKING:
+    import cirq
+
+
+@value.value_equality
+class QuirkQubitPermutationGate(ops.Gate):
```
viathor: I'd prefer a generic `PermutationGate`. See other comment.

cirq/contrib/quirk/cells/single_qubit_rotation_cells.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def _formula_gate(identifier: str, default_formula: str,
+                  gate_func: Callable[[Union[sympy.Symbol, float]], 'cirq.Gate']
+                 ) -> CellMaker:
+    return CellMaker(identifier=identifier,
+                     size=gate_func(0).num_qubits(),
+                     maker=lambda args: ExplicitOperationsCell([
+                         gate_func(
+                             parse_formula(default_formula if args.value is None
```
viathor: It'd be good to define explicitly what we intend to support here. This would help in choosing or implementing a safe way to parse these formulae (see other comment).

cirq/contrib/quirk/cells/cell.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def controlled_by(self, qubit: 'cirq.Qid'):
+        return ExplicitOperationsCell(
+            [op.controlled_by(qubit) for op in self._operations],
+            self._basis_change)
+
+
+CELL_SIZES = range(1, 17)
+
+CellMakerArgs = NamedTuple('CellMakerArgs', [
```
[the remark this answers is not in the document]
Strilanc: Before I had this value, I found myself adding parameters over time and having to go back and rewrite the signatures of a dozen lambdas methods. It's true that it is only produced in one place, but it is consumed in many places.

cirq/contrib/quirk/cells/parse.py
Strilanc: Unfortunately, sympy seems to throw a variety of exceptions when parsing.

cirq/contrib/quirk/__init__.py
```diff
@@ -14,4 +14,11 @@
 
 """Converts cirq circuits into quirk circuits."""
 
-from cirq.contrib.quirk.export_to_quirk import circuit_to_quirk_url
+from cirq.contrib.quirk.cells import (
+    QuirkQubitPermutationGate,)
```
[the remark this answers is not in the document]
Strilanc: I intend to expose all Gate classes to the outside world and to give them reprs, so that printing out circuit reprs and pasting them into test cases continues to work.

I agree that it could be fused with the permutation gate from contrib and moved into `ops`, but I think that belongs in a later PR.

The reason the gate is carrying its quirk id is because I later intend to allow round tripping of the gates.

cirq/contrib/quirk/cells/parse.py
Strilanc: I've completely rewritten this method. It now passes the fact that 'i' maps to 1j in via a dictionary to the parse method. It also restricts the available functions and constants to the set available in cirq. It furthermore unifies both the complex and formula parsing through the same code.

cirq/contrib/quirk/cells/parse.py
Strilanc: I've completely rewritten this method. It now passes the fact that 'i' maps to 1j in via a dictionary to the parse method. It also restricts the available functions and constants to the set available in cirq. It furthermore unifies both the complex and formula parsing through the same code.

cirq/contrib/quirk/cells/qubit_permutation_cells.py
Strilanc: Acknowledged.

cirq/contrib/quirk/cells/single_qubit_rotation_cells.py
Strilanc: Basically the idealized goal is to exactly match what Quirk does.

cirq/contrib/quirk/cells/parse.py
Strilanc: Done.

cirq/contrib/quirk/cells/parse.py
Strilanc: Anticipating usage soon.

cirq/contrib/quirk/cells/parse.py
Strilanc: Done.

cirq/contrib/quirk/cells/control_cells_test.py
```diff
@@ -0,0 +1,165 @@
...
+            cirq.Y(qs[4])**-0.5,
+            cirq.X(qs[5])**-0.5,
+            cirq.X(qs[6])**0.5,
+            cirq.X(qs[0]).controlled_by(*qs[1:7]),
+            cirq.Z(qs[7]).controlled_by(*qs[1:7]),
+            cirq.X(qs[2]),
+            cirq.Y(qs[3])**-0.5,
+            cirq.Y(qs[4])**0.5,
+            cirq.X(qs[5])**0.5,
+            cirq.X(qs[6])**-0.5,
```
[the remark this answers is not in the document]
Strilanc: Done.

cirq/contrib/quirk/cells/cell_test.py
```diff
@@ -0,0 +1,44 @@
...
+from cirq.contrib.quirk.cells.cell import Cell, ExplicitOperationsCell
+
+
+def test_cell_defaults():
+    c = Cell()
+    assert c.operations() == ()
+    assert c.basis_change() == ()
+    assert c.controlled_by(cirq.LineQubit(0)) is c
+    x = []
+    c.modify_column([])
```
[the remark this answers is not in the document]
Strilanc: Oh wow, good catch.

cirq/contrib/quirk/url_to_circuit.py
```diff
@@ -0,0 +1,121 @@
...
+    # Parse column json into cells.
+    registry = {
+        entry.identifier: entry for entry in generate_all_quirk_cell_makers()
+    }
+    parsed_cols: List[List[Optional[Cell]]] = []
+    for i, col in enumerate(cols):
+        parsed_cols.append(parse_col_cells(registry, i, col))
+
+    # Apply column modifiers (controls and inputs).
+    for c in parsed_cols:
```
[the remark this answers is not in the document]
Strilanc: Done.

cirq/contrib/quirk/cells/control_cells.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def basis_change(self) -> 'cirq.OP_TREE':
+        yield from self._basis_change
+
+        # Temporarily move the ZZZ..Z parity observable onto a single qubit.
+        for q in self.qubits[1:]:
+            yield ops.CNOT(q, self.qubits[0])
+
+
+def generate_all_control_cell_makers():
```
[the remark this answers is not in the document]
Strilanc: Done.

cirq/contrib/quirk/cells/cell.py
```diff
@@ -0,0 +1,131 @@
...
+
+class Cell:
+    """A gate, operation, display, operation modifier, etc from Quirk.
+
+    Represents something that can go into a column in Quirk, and supports the
+    operations ultimately necessary to transform a grid of these cells into a
+    `cirq.Circuit`.
+    """
+
+    def controlled_by(self, qubit: 'cirq.Qid'):
```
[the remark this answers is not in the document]
Strilanc: Done.

cirq/contrib/quirk/cells/all_cells.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2018 The Cirq Developers
```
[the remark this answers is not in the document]
Strilanc: Done.

cirq/contrib/quirk/cells/parse.py
viathor: I think the code will still allow expressions such as '2**1000000000000' into sympy opening up DoS vulnerability.

In general, I find the approach based on fortifying a versatile symbolic evaluator using regular expressions to be risky. A safer approach would be to explicit list a limited number of expressions we'd like to support. Isn't it feasible?

cirq/contrib/quirk/__init__.py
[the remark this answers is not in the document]
viathor: Optional: File an issue or leave a TODO so we don't forget to merge the gates.

cirq/contrib/quirk/cells/parse.py
```diff
@@ -0,0 +1,132 @@
+# Copyright 2018 The Cirq Developers
```
viathor: s/2018/2019/

cirq/contrib/quirk/cells/cell.py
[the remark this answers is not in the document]
viathor: Makes sense.

cirq/contrib/quirk/cells/cell.py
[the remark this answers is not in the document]
viathor: Return type annotation still missing.

Strilanc:
Ah, I added the return type to the wrong spot.

I think we have to settle for restricted-value sympy parsing for now, unless we want to duplicate a lot of, uh, quirky, parsing code from Quirk.

cirq/contrib/quirk/__init__.py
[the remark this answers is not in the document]
Strilanc: Acknowledged. I'm quite unsure how to merge the gates at the moment (one is more of a utility to make permutation gates whereas the other is an arbitrary permutation).

cirq/contrib/quirk/cells/cell.py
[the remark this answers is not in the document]
Strilanc: Done.

Strilanc:
Opened https://github.com/quantumlib/Cirq/issues/2259

PR #2197, opened 2019-09-26
Update the spelling error in sparse_simulator.py

cjcarey:
Error was at L344


cjcarey:
@googlebot I signed it!

PR #2205, opened 2019-09-26
Extricate the swaps

viathor:
Because common_gates.py has two many gates.

With the exception of some tests extended with a few cases for consistency and coverage, this PR only moves code around.


vtomole:
I agree with making `common_gates.py` have less lines; but Swap is a "common gate" to me. What defines what goes under `common_gates` and what doesn't?

Strilanc:
@vtomole The issue is that we shouldn't have a file called "common gates" in the first place, because it guarantees the file will be too large.

viathor:
Size is one thing, ease of navigation is another: cirq/ops will be easier to navigate for new users if filenames reflect their content better. "common_gates" is an arbitrary catch-all and should go away.

mpharrigan:
fyi this actually breaks [pytket](https://github.com/CQCL/pytket)'s cirq intergration because they import specifically from `common_gates`. Typically, any module that doesn't start with a `_` is fair game.

Can we add (back) in compatibility imports with a deprecation warning?

Strilanc:
Is there a way to do that? Having a `property` method at the module level doesn't work. How do we intercept the access?

PR #2206, opened 2019-09-26
Delete control protocol.

smitsanghavi:
Also remove a TODO from Operation.controlled_by(). The TODO was invalid since Operation always has the qubits and does not need qid_shapes separately.

Fixes https://github.com/quantumlib/Cirq/issues/1973


smitsanghavi:
#1973

PR #2207, opened 2019-09-27
makes random_circuit reproducible across runs

bryano:


cirq/testing/random_circuit_test.py
```diff
@@ -72,24 +72,31 @@ def test_random_circuit(n_qubits: Union[int, Sequence[cirq.Qid]],
...
-    assert circuit1 == circuit2
+    seeds = (seed, np.random.RandomState(seed))
+    circuits = [
+        random_circuit(qubits=20,
+                       n_moments=20,
+                       op_density=0.7,
+                       random_state=seed) for seed in seeds for _ in range(2)
+    ]
+    eq = cirq.testing.EqualsTester()
+    eq.add_equality_group(circuits)
```
kevinsung: This call is incorrect; it should be
```
eq.add_equality_group(*circuits)
```
Fixing this reveals that the test is incorrect because it passes the same RandomState object twice in a row, which results in different circuits. That's why in the original test, the line
```
prng = np.random.RandomState(seed)
```
is performed twice.

bryano: Fixed. Two bugs makes a right :)

mpharrigan:
the test failure looks suspicious

bryano:
@mpharrigan That bug is fixed in #2223

PR #2222, opened 2019-09-27
[Incomplete] Breaking down jobs.

MichaelBroughton:
In an effort to tackle the second part of #2107 , I introduced a new class for batches of jobs along with some unpolished logic for how it would operate. Since writing the tests would take a lot of time I was wondering if I could get a quick look over from @dabacon or @wcourtney and see if this is roughly what was in mind as a solution to the issue before I go ahead with writing them all. The `results` method has an identical return type as the regular `EngineJob`.

### DIFF
--- a/cirq/google/__init__.py
+++ b/cirq/google/__init__.py
@@ -38,6 +38,7 @@ from cirq.google.engine import (
     Engine,
     engine_from_environment,
     EngineJob,
+    EngineJobBatch,
     EngineProgram,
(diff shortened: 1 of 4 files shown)


balopat:
@MichaelBroughton is this something you're still working on?

MichaelBroughton:
Nope. Feel free to close, this is basically irellevant now after Doug put in batching all the way through.

PR #2223, opened 2019-09-27
[Fixes #2174] makes get_logical_operations and is_valid_routing more robust

### LINKED ISSUE #2174: contrib/test_route_circuit_via_unitaries flaked
Happened in this build:

https://github.com/quantumlib/Cirq/pull/2165/checks?check_run_id=235027498

This is the build log from travis:

```
_____ test_route_circuit_via_unitaries[circuit9-device_graph9-greedy-True] _____
circuit = cirq.Circuit([
    cirq.Moment(operations=[
        cirq.S.on(cirq.NamedQubit('1')),
        cirq.ISWAP.on(cirq.NamedQ...t('0'), cirq.NamedQubit('3')),
    ]),
    cirq.Moment(operations=[
        cirq.Y.on(cirq.NamedQubit('2')),
    ]),
])
device_graph = <networkx.classes.graph.Graph object at 0x7fd4cf8e76a0>
algo = 'greedy', make_bad = True
    @pytest.mark.parametrize(
        'circuit,device_graph,algo,make_bad', [(cirq.testing.random_circuit(
            4, 8, 0.5), ccr.get_grid_device_graph(3, 2), 'greedy', make_bad)
                                               for make_bad in (False, True)
                                               for _ in range(5)] +
        [(cirq.Circuit(), ccr.get_grid_device_graph(3, 2), 'greedy', False)])
    def test_route_circuit_via_unitaries(circuit, device_graph, algo, make_bad):
        swap_network = ccr.route_circuit(circuit, device_graph, algo_name=algo)
    
        logical_qubits = sorted(circuit.all_qubits())
        if len(logical_qubits) < 2:
            return
        reverse_mapping = {l: p for p, l in swap_network.initial_mapping.items()}
        physical_qubits = [reverse_mapping[l] for l in logical_qubits]
        physical_qubits += list(set(device_graph).difference(physical_qubits))
        n_unused_qubits = len(physical_qubits) - len(logical_qubits)
    
        if make_bad:
            swap_network.circuit += [cirq.CNOT(*physical_qubits[:2])]
        cca.return_to_initial_mapping(swap_network.circuit)
    
        logical_unitary = circuit.unitary(qubit_order=logical_qubits)
        logical_unitary = np.kron(logical_unitary, np.eye(1 << n_unused_qubits))
        physical_unitary = swap_network.circuit.unitary(qubit_order=physical_qubits)
    
>       assert ccr.is_valid_routing(circuit, swap_network) == (not make_bad)
cirq/contrib/routing/router_test.py:64: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
cirq/contrib/routing/utils.py:76: in is_valid_routing
    return cca.is_topologically_sorted(circuit_dag, logical_operations, equals)
cirq/contrib/acquaintance/topological_sort.py:63: in is_topologically_sorted
    ops.flatten_op_tree(operations)):
cirq/ops/op_tree.py:89: in flatten_to_ops
    for subtree in root:
cirq/contrib/acquaintance/permutation.py:234: in get_logical_operations
    yield op.transform_qubits(mapping.__getitem__)
cirq/ops/raw_types.py:364: in transform_qubits
cirq/ops/gate_operation.py:54: in with_qubits
    return self.gate.on(*new_qubits)
cirq/ops/common_gates.py:1131: in on
    return super().on(*args)
cirq/ops/raw_types.py:200: in on
    return gate_operation.GateOperation(self, list(qubits))
cirq/ops/gate_operation.py:39: in __init__
    gate.validate_args(qubits)
cirq/ops/raw_types.py:190: in validate_args
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
val = cirq.CNOT, qubits = [cirq.NamedQubit('0'), cirq.NamedQubit('0')]
    def _validate_qid_shape(val: Any, qubits: Sequence[Qid]) -> None:
        """Helper function to validate qubits for gates and operations.
    
        Raises:
            ValueError: The operation had qids that don't match it's qid shape.
        """
        qid_shape = protocols.qid_shape(val)
        if len(qubits) != len(qid_shape):
            raise ValueError('Wrong number of qubits for <{!r}>. '
                             'Expected {} qubits but got <{!r}>.'.format(
                                 val, len(qid_shape), qubits))
        if any(qid.dimension != dimension
               for qid, dimension in zip(qubits, qid_shape)):
            raise ValueError('Wrong shape of qids for <{!r}>. '
                             'Expected {} but got {} <{!r}>.'.format(
                                 val, qid_shape,
                                 tuple(qid.dimension for qid in qubits), qubits))
        if len(set(qubits)) != len(qubits):
            raise ValueError('Duplicate qids for <{!r}>. '
                             'Expected unique qids but got <{!r}>.'.format(
>                                val, qubits))
E           ValueError: Duplicate qids for <cirq.CNOT>. Expected unique qids but got <[cirq.NamedQubit('0'), cirq.NamedQubit('0')]>.
cirq/ops/raw_types.py:452: ValueError
----------- coverage: platform linux, python 3.6.7-final-0 -----------
Coverage annotated source written next to source
=============================== warnings summary ===============================
/home/travis/build/quantumlib/Cirq/cirq/ops/arithmetic_operation.py:224: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
  dst[outputs] = src[inputs]
-- Docs: https://docs.pytest.org/en/latest/warnings.html
1 failed, 7902 passed, 26 skipped, 72 xfailed, 1 warnings in 274.27 seconds
Finished pytest
Running incremental-coverage
All touched lines covered
Finished incremental-coverage
Failed.
```

bryano:
also adds SwapNetwork.__str__


bryano:
For good measure, I ran the test 10k times, and it always passed. (Previously, would fail at least once in a hundred runs.)

cirq/contrib/acquaintance/permutation.py
```diff
@@ -224,14 +227,30 @@ def update_mapping(mapping: Dict[ops.Qid, LogicalIndex],
...
+    logical_ops = []
     for op in cast(Iterable['cirq.Operation'], ops.flatten_op_tree(operations)):
         if (isinstance(op, ops.GateOperation) and
                 isinstance(op.gate, PermutationGate)):
             op.gate.update_mapping(mapping, op.qubits)
         else:
-            yield op.transform_qubits(mapping.__getitem__)
+            for q in op.qubits:
+                if mapping.get(q) is None:
+                    return None
```
kevinsung: I feel like this should raise an error instead of returning None.

bryano: That was my original impulse as well, but this allows `is_valid_routing` to handle it without catching exceptions.

bryano: I could add a "strict" flag that throw an error in such cases.

I also thought about returning a list the same length as `operations` but with `None` where there is a logical operation on an unmapped qubit, but I didn't see a use case for such precision.

kevinsung: I would raise an error and have `is_valid_routing` handle an exception.

bryano: Why? `None` indicates that there isn't a well-defined answer. It's also not iterable, so a caller that doesn't account for that return value would get an error anyway.

kevinsung: I think that returning None is surprising behavior. I think that in most cases a user calling this function expects to get something as the output and it's better to give a helpful error message if they got the input wrong.

bryano: Done

PR #2224, opened 2019-09-27
Support receiving both api.google and google.api protos

dstrain115:
- Make sure that engine class can receive both versions of the proto.


cirq/google/engine/engine.py
```diff
@@ -520,11 +520,16 @@ def get_job_results(self,
             parent=job_resource_name).execute()
         result = response['result']
         result_type = result['@type'][len(TYPE_PREFIX):]
-
         if result_type == 'cirq.api.google.v1.Result':
             return self._get_job_results_v1(response['result'])
```
dabacon: nit: these could use result instead of response['result']

PR #2225, opened 2019-09-27
Utilities for visualizing routing

mpharrigan:
I found these to be helpful when playing with the routing code


mpharrigan:
cc @KevinVillela might be useful

cirq/contrib/routing/__init__.py
```diff
@@ -18,10 +18,12 @@
...
     get_grid_device_graph,
     gridqubits_to_graph_device,
+    nx_qubit_layout,
 )
 from cirq.contrib.routing.router import (route_circuit, ROUTERS)
 from cirq.contrib.routing.swap_network import SwapNetwork
 from cirq.contrib.routing.utils import (
     is_valid_routing,
     ops_are_consistent_with_device_graph,
+    get_circuit_connectivity,
```
bryano: nit: alphabetize

cirq/contrib/routing/device_test.py
```diff
@@ -31,3 +33,22 @@ def test_get_linear_device_graph(n_qubits):
     assert sorted(graph) == cirq.LineQubit.range(n_qubits)
     assert len(graph.edges()) == n_qubits - 1
     assert all(abs(a.x - b.x) == 1 for a, b in graph.edges())
+
+
+def test_nx_qubit_layout():
+    foxtail_graph = ccr.xmon_device_to_graph(cirq.google.Foxtail)
+    pos = ccr.nx_qubit_layout(foxtail_graph)
```
bryano:
```suggestion
    pos = ccr.nx_qubit_layout(foxtail_graph)
    assert len(pos) == len(foxtail_graph)
```

cirq/contrib/routing/device.py
```diff
@@ -47,3 +47,30 @@ def gridqubits_to_graph_device(qubits: Iterable[cirq.GridQubit]):
 
 def _manhattan_distance(qubit1: cirq.GridQubit, qubit2: cirq.GridQubit) -> int:
     return abs(qubit1.row - qubit2.row) + abs(qubit1.col - qubit2.col)
+
+
+def nx_qubit_layout(graph: nx.Graph) \
```
bryano: This can just take in an iterable of qubits (which our device graphs happen to be).

cirq/contrib/routing/utils.py
```diff
@@ -74,3 +77,20 @@ def is_valid_routing(
                                                    can_reorder=can_reorder)
     logical_operations = swap_network.get_logical_operations()
     return cca.is_topologically_sorted(circuit_dag, logical_operations, equals)
+
+
+def get_circuit_connectivity(circuit: 'cirq.Circuit') -> nx.Graph:
```
bryano: It would be helpful to annotate the edges with their multiplicity and/or depth in the circuit (i.e., which time slice they would be in), but that can wait for a later PR.

cirq/contrib/routing/device.py
```diff
@@ -47,3 +47,30 @@ def gridqubits_to_graph_device(qubits: Iterable[cirq.GridQubit]):
...
+    positioned in a line.
+
+    >>> import cirq.contrib.routing as ccr
+    >>> import networkx as nx
+    >>> g = ccr.xmon_device_to_graph(cirq.google.Foxtail)
+    >>> pos = ccr.nx_qubit_layout(g)
+    >>> nx.draw_networkx(g, pos=pos)
+
+    """
+    pos = {
```
Strilanc: Use inline type annotations now that we're python 3.6

`pos: Dict[...] = {}`

cirq/contrib/routing/device.py
```diff
@@ -47,3 +47,30 @@ def gridqubits_to_graph_device(qubits: Iterable[cirq.GridQubit]):
 
 def _manhattan_distance(qubit1: cirq.GridQubit, qubit2: cirq.GridQubit) -> int:
     return abs(qubit1.row - qubit2.row) + abs(qubit1.col - qubit2.col)
+
+
+def nx_qubit_layout(graph: nx.Graph) \
+        -> Dict[Union[cirq.GridQubit, cirq.LineQubit], Tuple[int, int]]:
```
Strilanc: You should take arbitrary Qid. Map any non-grid non-line ones to the Y axis by sorted index or something like that.

cirq/contrib/routing/device.py
```diff
@@ -47,3 +47,30 @@ def gridqubits_to_graph_device(qubits: Iterable[cirq.GridQubit]):
...
+    >>> nx.draw_networkx(g, pos=pos)
+
+    """
+    pos = {
+    }  # type: Dict[Union[cirq.GridQubit, cirq.LineQubit], Tuple[int, int]]
+    for node in graph.nodes:
+        if isinstance(node, cirq.GridQubit):
+            pos[node] = (node.col, -node.row)
+        elif isinstance(node, cirq.LineQubit):
+            pos[node] = (node.x, 0)
```
Strilanc: If there's a mix of line qubits and grid qubits, this could cause some very confusing overlap. A simple workaround is to set the Y coordinate to `0.5`.

cirq/contrib/routing/__init__.py
```diff
@@ -18,10 +18,12 @@
     get_linear_device_graph,
     get_grid_device_graph,
     gridqubits_to_graph_device,
+    nx_qubit_layout,
```
Strilanc: This name is not descriptive enough.

cirq/contrib/routing/device.py
```diff
@@ -47,3 +47,30 @@ def gridqubits_to_graph_device(qubits: Iterable[cirq.GridQubit]):
...
+
+    This can be used in place of nx.spring_layout or other networkx layouts.
+    GridQubits are positioned according to their row/col. LineQubits are
+    positioned in a line.
+
+    >>> import cirq.contrib.routing as ccr
+    >>> import networkx as nx
+    >>> g = ccr.xmon_device_to_graph(cirq.google.Foxtail)
+    >>> pos = ccr.nx_qubit_layout(g)
+    >>> nx.draw_networkx(g, pos=pos)
```
Strilanc: This code will be executed by doctest

cirq/contrib/routing/device_test.py
mpharrigan: added locally

cirq/contrib/routing/device.py
mpharrigan: This is to match the other networkx layout functions

cirq/contrib/routing/utils.py
mpharrigan: yeah out of scope. It would take some thought on how to present that information visually

cirq/contrib/routing/device.py
mpharrigan: nice

cirq/contrib/routing/device.py
mpharrigan: nice

cirq/contrib/routing/__init__.py
mpharrigan: I was attempting to make it look like networkx

```python
g = nx.Graph(...)
pos = nx.spring_layout(g)
pos = nx.planar_layout(g)
pos = nx_qubit_layout(g)
```

cirq/contrib/routing/device.py
mpharrigan: done

cirq/contrib/routing/device.py
mpharrigan: is that a problem?

mpharrigan:
@Strilanc PTAL

cirq/contrib/routing/device.py
Strilanc: It depends if draw_networkx makes a file in some random folder or not

bryano: It doesn't [look like](https://networkx.github.io/documentation/stable/_modules/networkx/drawing/nx_pylab.html#draw_networkx) `draw_networkx` writes to any files.

Issue #2196, opened 2019-09-26
Remove conditional in fidelity_elimination_test after #2154


cjcarey:
https://github.com/quantumlib/Cirq/blob/46d72e867abb091e4c305ed46b05f2178c8bf251/cirq/experiments/fidelity_estimation_test.py#L22

viathor:
Oops, should be https://github.com/quantumlib/Cirq/issues/2154.

PR #2227, opened 2019-09-27
Allow repetitions=0 in cirq.sample

### LINKED ISSUE #2154: Support repetitions=0 in cirq.sample and related utilities
Currently, setting `repetitions` to 0 raises

```
ValueError: operands could not be broadcast together with shapes (0,) (5,)
```

viathor:
Fixes #2154 and #2196.

### DIFF
--- a/cirq/study/trial_result.py
+++ b/cirq/study/trial_result.py
@@ -108,7 +108,10 @@ class TrialResult:
                 converted_dict[key] = [
                     value.big_endian_bits_to_int(m_vals) for m_vals in val
                 ]
-            self._data = pd.DataFrame(converted_dict, dtype=int)
+            # Note that when a numpy array is produced from this data frame,
+            # Pandas will try to use np.int64 as dtype, but will upgrade to
+            # object if any value is too large to fit.
+            self._data = pd.DataFrame(converted_dict, dtype=np.int64)
         return self._data
 
     @staticmethod
--- a/cirq/study/trial_result.py
+++ b/cirq/study/trial_result.py
@@ -108,7 +108,7 @@ class TrialResult:
                 converted_dict[key] = [
                     value.big_endian_bits_to_int(m_vals) for m_vals in val
                 ]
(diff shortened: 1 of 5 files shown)


cirq/study/trial_result.py
```diff
@@ -108,7 +108,7 @@ def data(self) -> pd.DataFrame:
                 converted_dict[key] = [
                     value.big_endian_bits_to_int(m_vals) for m_vals in val
                 ]
-            self._data = pd.DataFrame(converted_dict)
+            self._data = pd.DataFrame(converted_dict, dtype=np.int64)
```
vtomole: Could there be any problems in the future from hard coding this to int64?

kevinsung: Doesn't this limit us to 63 qubits?

viathor: Pandas is actually smart about this. It'll use the suggested type when possible, but if you pass values that are too large, it'll upgrade to `dtype='object'`:

```
In [2]: pd.DataFrame({'a': []}, dtype=np.int64).to_numpy().dtype
Out[2]: dtype('int64')

In [3]: pd.DataFrame({'a': [1, 2]}, dtype=np.int64).to_numpy().dtype
Out[3]: dtype('int64')

In [4]: pd.DataFrame({'a': [1, 2**64+1]}, dtype=np.int64).to_numpy().dtype
Out[4]: dtype('O')
```

I've changed to `int` since it is less misleading.

viathor: See other comment.

viathor:
I think the Windows failures were actually caused by my change from dtype=np.int64 to dtype=int:

```
E       AssertionError: Attributes are different
E       
E       Attribute "dtype" are different
E       [left]:  int32
E       [right]: int64
```

This suggests that specifying `int` allows pandas to use 32-integers which our tests don't accept.

I switched back to dtype=np.int64 and added a comment explaining that pandas will use dtype=object when np.int64 turns out insufficient.

Issue #1835, opened 2019-07-17
Make noise models composable


Strilanc:
A major problem with NoiseModel as it is currently defined is that there is no way to define a compose(noise1, noise2) operation that works in the way the user would expect.

In general there may not be a way to compose two arbitrary noise models in a user-expected way, but certainly why they e.g. apply to non-overlapping operations or talk about orthogonal concepts of noise, it would be useful to be able to just compose.

Strilanc:
@mpharrigan @peterse @cduck could you post examples of code where you needed this?

mpharrigan:
```python
def _moment_is_measurements(moment: cirq.Moment) -> bool:
    """
    True/False if moment contains all measurement gates.

    If a moment is a mixture of measurement and non-measurement gates
    this will throw a ValueError.
    """
    is_measurements = None
    for gate in moment:
        if cirq.is_measurement(gate):
            if is_measurements is None:
                is_measurements = True
            elif is_measurements:
                # Check that *all* are measurements
                pass
            else:
                raise ValueError("Moment must be all measurements "
                                 "or all operations.")
        else:
            if is_measurements is None:
                is_measurements = False
            elif not is_measurements:
                # Check that all are *not* measurements
                pass
            else:
                raise ValueError("Moment must be all measurements "
                                 "or all operations.")
    return is_measurements


class SingleQubitDepolarzingWithReadoutNoiseModel(NoiseModel):
    """Applies noise to each qubit individually at the end of every moment.

    Args:
        depol_prob: depolarizing probability
        bitflip_prob: probability of a bitflip during readout
    """

    def __init__(self, depol_prob: float, bitflip_prob: float):
        cirq.value.validate_probability(depol_prob, 'depol prob')
        cirq.value.validate_probability(bitflip_prob, 'bitflip prob')
        self.qubit_noise_gate = cirq.DepolarizingChannel(depol_prob)
        self.readout_noise_gate = cirq.BitFlipChannel(bitflip_prob)

    def noisy_moment(self, moment: cirq.Moment, system_qubits: Sequence[cirq.Qid]):
        if _moment_is_measurements(moment):
            return [
                cirq.Moment(self.readout_noise_gate(q) for q in system_qubits),
                moment,
            ]
        else:
            return [
                moment,
                cirq.Moment(self.qubit_noise_gate(q) for q in system_qubits),
            ]


class SingleQubitDepolarzingWithDampedReadoutNoiseModel(NoiseModel):
[...]

PR #2229, opened 2019-09-27
Depolarizing noise models

mpharrigan:
Until #1835 is figured out, these are really helpful

originally by @ncrubin


cirq/contrib/noise_models/noise_models.py
```diff
@@ -0,0 +1,152 @@
...
+                raise ValueError("Moment must be all measurements "
+                                 "or all operations.")
+    return is_measurements
+
+
+class DepolarizingNoiseModel(devices.NoiseModel):
+    """Applies depolarizing noise to each qubit individually at the end of
+    every moment.
+
+    If a circuit contains measurements, they be in moments that don't
```
Strilanc: grammar

cirq/contrib/noise_models/noise_models.py
```diff
@@ -0,0 +1,152 @@
...
+        ]
+
+
+class DepolarizingWithReadoutNoiseModel(devices.NoiseModel):
+    """DepolarizingNoiseModel with probabilistic bit flips preceding
+    measurement.
+
+    This simulates readout error.
+
+    If a circuit contains measurements, they be in moments that don't
```
Strilanc: grammar

cirq/contrib/noise_models/noise_models.py
```diff
@@ -0,0 +1,152 @@
...
+    def noisy_moment(self, moment: 'cirq.Moment',
+                     system_qubits: Sequence['cirq.Qid']):
+        if _moment_is_measurements(moment):
+            return [
+                ops.Moment(self.readout_noise_gate(q) for q in system_qubits),
+                moment,
+            ]
+        return [
+            moment,
+            ops.Moment(self.qubit_noise_gate(q) for q in system_qubits),
```
Strilanc: Why after instead of before like in the other one?

cirq/contrib/noise_models/noise_models.py
```diff
@@ -0,0 +1,152 @@
...
+    """DepolarizingWithReadoutNoiseModel with T1 decay preceding
+    measurement.
+
+    This simulates asymmetric readout error. The noise is structured
+    so the T1 decay is applied, then the readout bitflip, then measurement.
+
+    If a circuit contains measurements, they be in moments that don't
+    also contain gates.
+
+    Args:
```
Strilanc: Does listing the arguments here instead of on `__init__` work? Will they show up in pycharm's autocomplete?

cirq/contrib/noise_models/noise_models.py
```diff
@@ -0,0 +1,152 @@
...
+        depol_prob: Depolarizing probability.
+        bitflip_prob: Probability of a bit-flip during measurement.
+        decay_prob: Probability of T1 decay during measurement.
```
Strilanc: How does this probability combine with the bitflip probability? They don't commute with each other.

For example, if I set both probabilities to 1, will the measurement always report true or will it always report false?

cirq/contrib/noise_models/noise_models.py
```diff
@@ -0,0 +1,152 @@
...
+
+if TYPE_CHECKING:
+    import cirq
+
+
+def _moment_is_measurements(moment: 'cirq.Moment') -> bool:
+    """Whether the moment is nothing but measurement gates.
+
+    If a moment is a mixture of measurement and non-measurement gates
+    this will throw a ValueError.
```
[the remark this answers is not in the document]
mpharrigan: done

cirq/contrib/noise_models/noise_models.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    is_measurements = None
```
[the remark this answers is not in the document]
mpharrigan: awesome

cirq/contrib/noise_models/noise_models.py
mpharrigan: I was trying to be hip.

Jk, good catch

cirq/contrib/noise_models/noise_models.py
mpharrigan: I've done some investigating. Pep257 says to put constructor docs in `__init__` but numpy (and presumably others) put it all in the class.

Pycharm autocomplete pulls the names from the function signature and not the docstring, so it still works. Pycharm "quick documentation" *does* correctly parse the Args documentation from class docstrings. 

I can put this on `__init__` if that's what the rest of cirq does

cirq/contrib/noise_models/noise_models.py
mpharrigan: I think you recently (read: very recently) changed it to noise models would keep measurements terminal. Some code expects terminal measurements.

Perhaps this is an unnecessary complication now that you're proposing putting all noise gates before

cirq/contrib/noise_models/noise_models.py
mpharrigan: This is addressed a bit in the body of the docstring

>    This simulates asymmetric readout error. The noise is structured
>    so the T1 decay is applied, then the readout bitflip, then measurement.

I think this is more physical. I imagine the bitflip is more of a classification error and the t1 decay is something physically going wrong in measurement. If that matches reality, then the current order makes sense.

mpharrigan: added to Args as well

cirq/contrib/noise_models/noise_models.py
Strilanc: Yeah, that's what the rest of cirq happens to do.

PR #2230, opened 2019-09-27
Add common serializers to cirq

dstrain115:
Adds 3 sets of serializers to cirq:

    - SINGLE_QUBIT_SERIALIZERS: A list of GateOpSerializer for single qubit
        rotations using cirq Gates.
    - MEASUREMENT_SERIALIZER:  Single GateOpSerializer for the measurement gate
    - SINGLE_QUBIT_HALF_PI_SERIALIZERS: A list of GateOpSerializer for single qubit
        rotations confined to half-pi increments using cirq Gates.


cirq/google/__init__.py
```diff
@@ -30,6 +30,15 @@
     xmon_op_from_proto_dict,
 )
 
+from cirq.google.common_serializers import (
+    MEASUREMENT_DESERIALIZER,
```
Strilanc: I wouldn't flatten these to the module level, as I wouldn't expect end users to use it.

dstrain115: removed.

Issue #1672, opened 2019-05-24
Serializing code shouldn't be in core Cirq: remove proto methods from GridQubit


dabacon:
For example we currently have proto calls in GridQubit.  

For old v1 protos we move this to serializing/deserializing methods that consume and produce objects.  We should make sure to do this for all objects, not just gates.

dstrain115:
#2231 fixed this.  Closing.

PR #2231, opened 2019-09-27
Move proto methods out of GridQubit

dabacon:


dabacon:
Fixes  #1672

cirq/google/api/v2/program.py
```diff
@@ -0,0 +1,31 @@
...
+from typing import TYPE_CHECKING
+
+from cirq import devices
+
+if TYPE_CHECKING:
+    import cirq
+
+
+def qubit_to_proto_id(q: devices.GridQubit) -> str:
+    """Return a proto id for a `cirq.GridQubit`"""
```
dstrain115: This will be gross if we need to extend this beyond GridQubit.  Can we make this support other Qubit types or rename it to grid_qubit_to_proto?

cirq/google/api/v2/program.py
```diff
@@ -0,0 +1,31 @@
...
+    return '{}_{}'.format(q.row, q.col)
+
+
+def qubit_from_proto_id(proto_id: str) -> 'cirq.GridQubit':
+    """Parse a proto id string to a `cirq.GirdQubit`"""
+    row, col = proto_id.split('_')
```
dstrain115: Can we have better error handling here, so it is clear what happened if you pass it an id that is not a grid qubit?

dabacon:
#2110 should go in first and then I will rework.

cirq/google/api/v2/program.py
dabacon: Without guidance on what we are doing for other types it doesn't make sense to add it.  But I guess you were thinking we would also have NameQubits?

cirq/google/api/v2/program.py
dabacon: move it to per type.

cirq/google/api/v2/program.py
dabacon: added for different qubit types

cirq/google/api/v1/programs.py
```diff
@@ -345,6 +352,15 @@ def raise_missing_fields(gate_name: str):
     raise ValueError('invalid operation: {}'.format(proto_dict))
 
 
+def _qubit_from_proto_dict(proto_dict):
+    """Proto dict must have 'row' and 'col' keys."""
+    # TODO: Deprecate v1 proto method.
```
dstrain115: Can we link to a person or issue here?

dabacon:
PTAL

cirq/google/api/v1/programs.py
dabacon: I'm not sure the todo makes any sense.  We do want to deprecate v1 protos.  But that's a much bigger task than this single location.  Removed.

PR #2232, opened 2019-09-27
Introduce `cirq.PAULI_STRING_LIKE` type

Strilanc:
- A `PAULI_STRING_LIKE` is a number, pauli op, identity op, mapping from qubits to pauli gates, or recursive iterable thereof
- Generalize `PauliString`'s constructor to take a `PAULI_STRING_LIKE`
- Add private `_MutablePauliString` type for aggregating products efficiently
- Tweak `op_gate_isinstance/of_type` to take `Any` instead of just `Operation`
- Add explicit keyword arguments to calls to `PauliString` constructor outside tests, as these do not go through recursive type checks
- Generalize `phased_pauli_product` to accept an `IdentityGate` argument
- Simplified `PauliString.__mul__` using new constructor
- Generalized `PauliString.__mul__/__rmul__` to accept IdentityGate
- Documented `PauliString.__init__` with examples

Fixes: #2180
Obsoletes https://github.com/quantumlib/Cirq/pull/2186


JulianSKelly:
Nerd sniping a core developer into fixing your problems = impact?

cirq/ops/pauli_string.py
```diff
@@ -22,43 +23,85 @@
...
+                      pauli_gates, clifford_gate, pauli_interaction_gate)
+
+if TYPE_CHECKING:
+    import cirq
+
+# A value that can be unambiguously converted into a `cirq.PauliString`.
+PAULI_STRING_LIKE = Union[
+    complex, 'cirq.OP_TREE',
+    Mapping['cirq.Qid', Union['cirq.Pauli', 'cirq.IdentityGate']],
+    Iterable,  # of PAULI_STRING_LIKE, but mypy doesn't do recursive types yet.
```
maffoo: Can do recursion with protocols if you want to preserve more type info, as in #2140.

cirq/ops/pauli_string.py
```diff
@@ -22,43 +23,85 @@
...
+            1j*Z(0)
+        """
+        if qubit_pauli_map is not None:
+            for v in qubit_pauli_map.values():
+                if not isinstance(v, pauli_gates.Pauli):
+                    raise TypeError(f'{v} is not a Pauli')
+
+        p = _MutablePauliString(coef=complex(coefficient),
+                                paulis=dict(qubit_pauli_map or {}))
+        p.inline_times_pauli_string_like(contents)
```
mpharrigan: seems weird to use both `qubit_pauli_map` and `contents`. You can imagine wanting to deprecate / remove `qubit_pauli_map` completely in the future and having everything put in `contents`. These semantics would make that change very painful

cirq/ops/pauli_string.py
```diff
@@ -111,38 +154,33 @@ def get(self, key: raw_types.Qid, default=None):
...
-                terms, self.coefficient * other.coefficient * extra_phase)
-        return NotImplemented
+    def __mul__(self, other) -> 'PauliString':
+        if not (isinstance(other, (PauliString, numbers.Number)) or
+                gate_operation.op_gate_isinstance(other,
+                                                  common_gates.IdentityGate)):
+            return NotImplemented
 
-    def __rmul__(self, other):
+        return PauliString(other,
```
mpharrigan: again, it seems very odd to rely on the (imo) strange constructor semantics to implement multiplication

cirq/ops/pauli_string.py
Strilanc: There are two reasons for leaving it present.

1. It is slightly faster and more predictable because it involves less type inspection shenanigans.
2. If you remove it, you are quite a large risk of creating an infinite recursion loop (e.g. if you accidentally create an X operation in the initializer, you'll infinitely recurse due to the call to `PauliString.__init__` inside `SingleQubitPauliStringGateOperation.__init__`).

Strilanc: I disagree that these changes make later deprecation change more painful. We can deprecate the argument in the same way we'd deprecate any other, with an auto-rewrite that puts it at the start of the contents list.

Strilanc: The use of the constructor for implicit multiplying is a consequence of the idea "I should be able to pass in a list of Pauli operations and it understands what I mean" + closure.

cirq/ops/pauli_string.py
Strilanc: This implementation also kinda caught me by surprise. I tried a few different ways of writing it (e.g. having a module level method that both `__init__` and `__mul__` delegate to) but found this to be the clearest.

PR #2234, opened 2019-09-27
Show yellow text when lines had to be reformatted

Strilanc:
- This makes it clearer at a glance that something happened instead of nothing (shown in green)


vtomole:
Cirq bot cookie expiration?

PR #2236, opened 2019-09-27
Reorder travis tests

dabacon:
This reorders the travis tests from longest to shortest in duration.  Because travis runs with 5 concurrent jobs that pick up test in this order putting a long test like build docs or windows later can lead to slower overall testing time.


Strilanc:
I kinda like that the format check is short finishes first, because it's the one that fails the most.

PR #2239, opened 2019-09-29
Update protobuf version

### LINKED ISSUE #2238: Protobuf version compatibility with Bazel
Downstream projects that use Bazel versions 0.22 or greater are only compatible with protobuf versions after 3.6.1.2.  This is due to use of a constant that the newer versions of Bazel consider deprecated (for details, see discussion at https://github.com/bazelbuild/bazel/issues/5827).  So I propose updating to protobuf version 3.7.0, the first minor version of protobuf with correct compatibility.

zaqqwerty:
Resolves #2238


maffoo:
We should also update the protobuf version in [`requirements.txt`](https://github.com/quantumlib/Cirq/blob/master/requirements.txt#L8) to match what's in the bazel build. Also note that we have a script [`dev_tools/build-protos.sh`](https://github.com/quantumlib/Cirq/blob/master/dev_tools/build-protos.sh) to build protos that uses the `grpcio-tools` python library, which bundles `protoc`. I think this is typically what we've been running to regenerate protos and check them in. We have the version of that library pinned in the devtools [requirements](https://github.com/quantumlib/Cirq/blob/master/dev_tools/conf/pip-list-dev-tools.txt#L10), though it's not immediately clear what version of protoc that bundles. It looks like we never set up CI jobs to ensure that the checked-in generated code is up to date, but we should set that up (also need to update docs to describe how we're handling protos, because there's potential confusion about whether to use `check/bazel` or `dev_tools/build-protos.sh`.

zaqqwerty:
Thank you for the comments.  I had indeed used `dev_tools/build-protos.sh` to regenerate the protos.  I find that the `grpcio-tools` version in the devtools requirements, 1.21.0, points at protobuf version 3.7.0 ([third party tools](https://github.com/grpc/grpc/tree/v1.21.0/third_party) -> [protobuf submodule](https://github.com/protocolbuffers/protobuf/tree/582743bf40c5d3639a70f98f183914a2c0cd0680)).  I updated `requirements` accordingly.

I find that the bazel machinery in master is actually not functional, I am working on remedying this here.

Strilanc:
I think we used to check that the bazel build worked, but then it kept breaking for reasons out of our control (e.g. the link where the protobuf installer was located stopped working) so I removed the check. If you can get it working consistently, that would be awesome.

zaqqwerty:
Made some further changes:
- Updated the WORKSPACE and BUILD files with the relevant rules for building version 3.8.0 protos
- Added the script `check/build-protos` to bazel build both the python and C++ proto rules, and added a stage to the travis build to call the script
- Updated the grpcio-tools version to 1.24.0, which uses proto version 3.8.0;  ran `/dev_tools/build-protos.sh ` and checked in the updated protos
- Added a section to `docs/development` describing how to approach these proto tools

Strilanc:
This looks good to me. @maffoo you agree? If so automerge it.

PR #2241, opened 2019-09-30
Fix initial state possibly being modified in simulation

### LINKED ISSUE #2240: Simulator modifies initial state
It seems that sparse_simulator modifies the initial_state argument in place during simulation. Is this a bug or a feature?

dabacon:
Fixes #2240

### DIFF
--- a/cirq/sim/sparse_simulator_test.py
+++ b/cirq/sim/sparse_simulator_test.py
@@ -622,9 +622,11 @@ def test_does_not_modify_initial_state():
     circuit = cirq.Circuit(InPlaceUnitary()(q0))
 
     initial_state = np.array([1, 0], dtype=np.complex64)
-    _ = simulator.simulate(circuit, initial_state=initial_state)
+    result = simulator.simulate(circuit, initial_state=initial_state)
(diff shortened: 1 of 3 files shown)


cirq/sim/sparse_simulator_test.py
```diff
@@ -605,6 +605,28 @@ def _apply_unitary_(self, args: cirq.ApplyUnitaryArgs):
...
+            args.target_tensor[0], args.target_tensor[1] = (
+                args.target_tensor[1], args.target_tensor[0])
+            return args.target_tensor
+
+    circuit = cirq.Circuit(InPlaceUnitary()(q0))
+
+    initial_state = np.array([1, 0], dtype=np.complex64)
+    _ = simulator.simulate(circuit, initial_state=initial_state)
+    np.testing.assert_array_almost_equal(np.array([1, 0], dtype=np.complex64),
+                                         initial_state)
```
viathor: Optional nit: I'd add a second assert on the result (ignored on l.625) to ensure simulate had any effect at all (if it didn't have an effect then even when simulate was making a copy the current assert might have succeed).

dabacon: done

PR #2242, opened 2019-09-30
Fix tensorflow version since tensorflow 2.0 was just released

dabacon:
`pip install tensorflow` now gives tensorflow 2.0 which will break us.


vtomole:
@dabacon What broke?

dabacon:
`tf.placeholder` no longer exists.  2.0 is a breaking change (hence major version bump).   They just released: https://venturebeat.com/2019/09/30/google-launches-tensorflow-2-0-with-tighter-keras-integration/  

I noticed this while doing other changes.

Issue #2176, opened 2019-09-25
Doing noisy simulation is unintuitive


mpharrigan:
To set up the most basic noisy simulation, you need three jargony-named things:

```python
cirq.DensityMatrixSimulator(noise=cirq.ConstantQubitNoiseModel(qubit_noise_gate=cirq.DepolarizingChannel(p=0.005)))
```

Opening this based on real user research :) Despite my hints, the user could not find the correct incantation to find a depolarizing noise model. One thing they tried is using `AQTNoiseModel`.

Strilanc:
I would say it's one jargony-named thing:

```
cirq.sample(
    circuit,
    noise=cirq.ConstantQubitNoiseModel(cirq.depolarize(p=0.005)))
```

Maybe what we could do is permit stuff like:

```
cirq.sample(
    circuit,
    noise=cirq.depolarize(p=0.005))
```

with the understanding that when the noise model is a single qubit operation it automatically gets wrapped?

mpharrigan:
`cirq.sample` isn't a good choice for testing code that uses a `Sampler` that you want to run via quantum engine

mpharrigan:
I think it would ease a lot of confusion if there was something named `DepolarizingNoiseModel`

Strilanc:
I see, you don't just want "the most basic noise simulation" you want to make a noisy sampler specifically.

Auto-wrapping single qubit operations into noise models would still make sense in that context.

We could generalize `cirq.Simulator()` to include density matrix simulation, and have a `with_noise(noise)` method on it.

Strilanc:
Creating a noise model type for every noisy channel operation type would be quite a lot of duplication. Having `cirq.depolarize(p)` play double duty as a de-facto noise model would be superior IMO.

Strilanc:
I'm liking the idea of generalizing simulator more and more. I'll bring it up at the meeting.

dabacon:
This now exists for single qubit noise.

PR #2244, opened 2019-09-30
Add `NOISE_MODEL_LIKE`

Strilanc:
- A `cirq.NOISE_MODEL_LIKE` is a `cirq.NoiseModel`, `None`, or a single-qubit gate
- Methods that took `cirq.NoiseModel` now take `cirq.NOISE_MODEL_LIKE`
- None is auto-converted into `cirq.NO_NOISE`
- Single qubit gates auto-converted into `cirq.ConstantQubitNoiseModel`

For example, you can now say: `cirq.sample(circuit, noise=cirq.depolarize(0.01))`

Part of https://github.com/quantumlib/Cirq/issues/2176


mpharrigan:
Does switching the order change simulation results?

Strilanc:
Yes, it can change simulation results. For example, for the circuit "X then measure", it means there is two hits of depolarizing noise instead of one before the measurement.

mpharrigan:
I'd consider this a [BREAKING CHANGE] and a particularly insidious one at that. Numerical studies will silently start giving different results when re-run on a newer cirq

Strilanc:
I've reverted the ordering change. Now only contains the generalization of things accepted as noise.

PR #2246, opened 2019-10-01
Add measure/scalar/swap/display/qft support to import-from-quirk

Strilanc:


Strilanc:
This is ready for review now @viathor

cirq/contrib/quirk/cells/frequency_space_cells.py
```diff
@@ -0,0 +1,51 @@
+# Copyright 2018 The Cirq Developers
```
viathor: nit: s/2018/2019/

(10x throughout)

cirq/contrib/quirk/cells/ignored_cells.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def reg_ignored_family(identifier_prefix: str) -> Iterator[CellMaker]:
+    yield reg_ignored_gate(identifier_prefix)
+    for i in CELL_SIZES:
+        yield reg_ignored_gate(identifier_prefix + str(i))
+
+
+def reg_ignored_gate(identifier: str) -> CellMaker:
+    return CellMaker(identifier, 0, lambda _: None)
```
viathor: nit: The purpose of these cell makers isn't obvious, so maybe add a comment somewhere in the file?

cirq/contrib/quirk/cells/frequency_space_cells.py
```diff
@@ -0,0 +1,51 @@
...
+def generate_all_frequency_space_cell_makers() -> Iterator[CellMaker]:
+    # Frequency space.
+    yield from reg_family("QFT", lambda n: cirq.QuantumFourierTransformGate(n))
+    yield from reg_family(
+        "QFT†", lambda n: cirq.inverse(cirq.QuantumFourierTransformGate(n)))
+    yield from reg_family(
+        "PhaseGradient", lambda n: cirq.PhaseGradientGate(num_qubits=n,
+                                                          exponent=0.5))
+    yield from reg_family(
+        "PhaseUngradient", lambda n: cirq.PhaseGradientGate(num_qubits=n,
```
viathor: nit: This name is weird. "PhaseGradient†" would be both clearer and consistent with the QFT cell above.

cirq/contrib/quirk/cells/measurement_cells.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def reg_measurement(identifier: str,
+                    basis_change: Optional['cirq.Gate'] = None) -> CellMaker:
+    return CellMaker(
+        identifier=identifier,
+        size=1,
+        maker=lambda args: ExplicitOperationsCell(
+            [ops.measure(*args.qubits, key=f'row={args.row},col={args.col}')],
+            basis_change=cast(Iterable['cirq.Operation'], [
```
viathor: Can we get rid of the cast? Gate.on() does have return type annotation, so it must be something else.

cirq/contrib/quirk/cells/scalar_cells.py
```diff
@@ -0,0 +1,34 @@
...
+if TYPE_CHECKING:
+    import cirq
+
+
+def generate_all_scalar_cell_makers() -> Iterator[CellMaker]:
+    yield _scalar("NeGate", ops.GlobalPhaseOperation(-1))
+    yield _scalar("i", ops.GlobalPhaseOperation(1j))
+    yield _scalar("-i", ops.GlobalPhaseOperation(-1j))
+    yield _scalar("√i", ops.GlobalPhaseOperation(1j**0.5))
+    yield _scalar("√-i", ops.GlobalPhaseOperation((-1j)**0.5))
```
viathor: Optional: This looks incomplete: -sqrt(i) and -sqrt(-i) are missing

cirq/contrib/quirk/cells/frequency_space_cells.py
Strilanc: Done.

cirq/contrib/quirk/cells/frequency_space_cells.py
Strilanc: Too late to change it now, this is the id used in Quirk.
(I was not consistent over time with the ids, in many many ways.)

cirq/contrib/quirk/cells/ignored_cells.py
Strilanc: Added a comment.

cirq/contrib/quirk/cells/measurement_cells.py
Strilanc: The problem is actually with the "x if y else z". mypy infers a too-specific type from x, and then complains about z.

cirq/contrib/quirk/cells/scalar_cells.py
Strilanc: Those aren't gates in Quirk. This is a pretty common pattern: everyone talks about T gates and S gates, but not really the "ST" gate.

cirq/contrib/quirk/cells/frequency_space_cells.py
Strilanc: Done.

cirq/contrib/quirk/cells/frequency_space_cells.py
Strilanc: This is forced on the Quirk side. I was not consistent with gate ids over time, in many ways.

cirq/contrib/quirk/cells/ignored_cells.py
Strilanc: Done.

cirq/contrib/quirk/cells/measurement_cells.py
Strilanc: The issue is the ternary expression. Mypy infers a too-specific type from the true-case value.

cirq/contrib/quirk/cells/scalar_cells.py
Strilanc: Those aren't Quirk gates. This is a common pattern: there's S and there's T, but not a combined ST.

Strilanc:
Hm. Thought github lost the comments the first time.

PR #2251, opened 2019-10-01
Give repr to contrib.routing.SwapNetwork, add error message test for unmapped qubits

Strilanc:


cirq/contrib/routing/router_test.py
```diff
@@ -87,3 +87,36 @@ def test_router_bad_args():
...
+    swap = cirq.contrib.acquaintance.SwapPermutationGate()
+
+    # Works before swap.
+    swap_network = cirq.contrib.routing.SwapNetwork(cirq.Circuit(cirq.CZ(a, t)),
+                                                    {
+                                                        a: a,
+                                                        b: b
+                                                    })
+    with pytest.raises(ValueError, match='acts on unmapped qubit'):
+        _ = list(swap_network.get_logical_operations())
```
bryano: Is the mock assignment for pylint?

cirq/contrib/acquaintance/permutation.py
```diff
@@ -220,9 +220,11 @@ def update_mapping(mapping: Dict[ops.Qid, LogicalIndex],
         operations: The operations to update according to.
     """
     for op in ops.flatten_op_tree(operations):
-        if (isinstance(op, ops.GateOperation) and
-            isinstance(op.gate, PermutationGate)):
-            op.gate.update_mapping(mapping, op.qubits)
+        # Ignoring type warning about passing an abstract type into Type[...].
```
bryano: Maybe link to the [issue](https://github.com/python/mypy/issues/5374) here?

cirq/contrib/acquaintance/permutation.py
```diff
@@ -220,9 +220,11 @@ def update_mapping(mapping: Dict[ops.Qid, LogicalIndex],
...
+        gate = ops.op_gate_of_type(op, PermutationGate)  # type: ignore
```
bryano: Is this just a style preference?

cirq/contrib/acquaintance/permutation.py
```diff
@@ -220,9 +220,11 @@ def update_mapping(mapping: Dict[ops.Qid, LogicalIndex],
...
+        if gate is not None:
+            # Ignoring type warning about op.qubits not being a tuple.
+            gate.update_mapping(mapping, op.qubits)  # type: ignore
```
bryano: Why does mypy think `op.qubits` has type `Union[Tuple[Qid, ...], FrozenSet[Qid]]`? Shouldn't it be a tuple?

cirq/contrib/routing/router_test.py
Strilanc: Yes, sometimes pylint complains that the result of an expression is unused. So I got into the habit of assigning to `_` as a way of saying "I know I'm throwing this away".

cirq/contrib/acquaintance/permutation.py
Strilanc: I can revert it if you prefer. It's just much shorter, and it's why we defined these methods.

cirq/contrib/acquaintance/permutation.py
Strilanc: I honestly don't know. My guess is that someone in the code base a qubits method returns a FrozenSet[Qid] and mypy is inferring the return type from that, but I wasn't able to find where.

bryano: It seems like a silly thing to spend too much time on, but this seems like a case where mypy might actually be catching an bug, so maybe we shouldn't override it so quickly.

cirq/contrib/acquaintance/permutation.py
bryano: I don't have a preference. IIRC, this is what I originally wrote, and changed it to avoid the type issues.

cirq/contrib/routing/router_test.py
bryano: Maybe
```
for op in swap_network.get_logical_operations():
    pass # coverage: ignore
```

It's fine as-is if you have a preference.

bryano:
I don't want to hold this up on the mypy thing; if you think it's not worth the trouble, that's fine.

LGTM after adding the link to the issue.

bryano:
@Strilanc ping

bryano:
I thought it was good modulo getting the tests passed.

balopat:
@Strilanc - can you have a look at this why it's failing and merge it?

Strilanc:
Closed as obsolete

PR #2253, opened 2019-10-02
Added __repr__ tests

### LINKED ISSUE #998: Test cirq.testing.assert_equivalent_repr on every class with a __repr__ exposed by cirq
Comes after https://github.com/quantumlib/Cirq/pull/997 (which defines the `cirq.testing.assert_equivalent_repr` method)

Basically the goal here is, for each place you can find `def __repr__(` in the project, there should be an associated test that calls `cirq.testing.assert_equivalent_repr(a, b, c) ` on a few test values.

Hancockan:
Searched for classes with __repr__ methods and added tests where missing.

Fixes #998


Hancockan:
@vtomole Just completed the format check.

Issue #2147, opened 2019-09-23
Move `Cirq/cirq/vis/examples/` to `Cirq/examples`


vtomole:

jshede:
assign to me

vtomole:
Closed by: https://github.com/quantumlib/Cirq/pull/2255

PR #2255, opened 2019-10-02
#2147 Move `Cirq/cirq/vis/examples/` to `Cirq/examples`

jshede:
Updated the location of the bristlecone_heatmap_exmaples to be in sync with the other examples.


jshede:
@googlebot I signed it!

PR #2258, opened 2019-10-03
Add support for arithmetic gates to import-from-quirk

Strilanc:
- Add QuirkArithmeticOperation
- Add repr-able QuirkArithmeticLambda
- Add ArithmeticCell
- Define every arithmetic operation in Quirk in terms of ArithmeticCell

Example repr:

```python
cirq.contrib.quirk.QuirkArithmeticOperation(
    '+=AB3',
    operation=cirq.contrib.quirk.QuirkArithmeticLambda('lambda x, a, b: x + a * b'),
    target=[cirq.LineQubit(0), cirq.LineQubit(1), cirq.LineQubit(2)],
    inputs=[
        3,
        (cirq.LineQubit(3), cirq.LineQubit(4)),
    ],
)
```


Strilanc:
@viathor This one is ready for review now

cirq/contrib/quirk/cells/arithmetic_cells.py
```diff
@@ -0,0 +1,308 @@
...
+
+    def __init__(self, code: str):
+        """
+        Args:
+            code: A string containing a python lambda expression using only
+                values built into python or cirq. The lambda expression is
+                passed as a string to guarantee it has a `repr`.
+        """
+        self.code = code
+        self._func: Callable = eval(code, {'cirq': cirq}, {})
```
viathor: It is very dangerous to allow input to be interpreted as code. Can we not do that and instead implement the functionality we actually need here? Perhaps we can simply take a lambda directly, rather than taking its string form.

To make this more concrete and entertaining, here's how to delete files:

```python
payload = "__import__('os').system('rm -rf /path/to/victims/files')"
cirq.contrib.quirk.QuirkArithmeticLambda(payload)
```

cirq/contrib/quirk/cells/cell.py
```diff
@@ -89,6 +105,18 @@ def modify_column(self, column: List[Optional['Cell']]) -> None:
             Nothing. The `column` argument is mutated in place.
         """
 
+    def persistent_modifiers(self) -> Dict[str, Callable[['Cell'], 'Cell']]:
+        """Overridable modifications to apply to the rest of the circuit.
+
+        Persistent modifiers apply to all cells in the same column, not just to
+        future columns.
```
viathor: This comment is confusing. Does a persistent modifier apply to future columns or the same column as `self` is in? Also, what about past columns?

cirq/contrib/quirk/cells/arithmetic_cells.py
```diff
@@ -0,0 +1,308 @@
...
+                    Tuple, Any, cast, List)
+
+import cirq
+from cirq import ops, value
+from cirq.contrib.quirk.cells.cell import Cell, CellMaker, CELL_SIZES
+
+
+@value.value_equality
+class QuirkArithmeticOperation(ops.ArithmeticOperation):
+    """Applies an arithmetic to a target and some inputs.
```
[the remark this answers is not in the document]
Strilanc: Done.

cirq/contrib/quirk/cells/arithmetic_cells.py
```diff
@@ -0,0 +1,308 @@
...
+
+    Uses quirk-specific implicit effects like assuming that the presence of an
+    'r' input implies modular arithmetic (meaning that e.g. values larger than
+    the modulus will not be affected).
```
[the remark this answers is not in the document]
Strilanc: Done.

cirq/contrib/quirk/cells/arithmetic_cells.py
```diff
@@ -0,0 +1,308 @@
...
+    """
+
+    def __init__(self, identifier: str,
+                 operation: 'cirq.contrib.quirk.QuirkArithmeticLambda',
+                 target: Sequence['cirq.Qid'],
+                 inputs: Sequence[Optional[Union[Sequence['cirq.Qid'], int]]]):
```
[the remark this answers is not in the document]
Strilanc: Done.

cirq/contrib/quirk/cells/arithmetic_cells.py
```diff
@@ -0,0 +1,308 @@
...
+        Args:
+            identifier: The quirk identifier string for this operation.
+            operation: A repr-able lambda that determines the new value of the
+                target based on its current value and the values of the inputs.
+            target: The target qubit register.
+            inputs: Qubit registers (or classical constants) that
+                determine what happens to the target.
+        """
+        if operation.is_modular:
+            r = cast(Union[Sequence['cirq.Qid'], int], inputs[-1])
```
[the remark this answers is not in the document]
Strilanc: That used to be the way it worked. It made things more complicated, because this parameter had to be fed through several places in the code. For example, it ended up being in the repr shown to users. The specialization of the meaning of r is a quirk convention that it makes sense to reflect in the implementation.

cirq/contrib/quirk/cells/arithmetic_cells.py
```diff
@@ -0,0 +1,308 @@
...
+            assert r is not None
```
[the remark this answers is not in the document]
Strilanc: I fixed the `inputs` parameter allowing None in its type. This assertion became a `cast` in `ArithmeticCell` which is guaranteed to succeed because it is preceded by a check that all inputs are present.

cirq/contrib/quirk/cells/arithmetic_cells.py
Strilanc: I don't see how the attack surface is larger than taking a normal callable. With a normal callable you can also trivially delete files. But we don't say that python's `map` function is insecure because of this.

Every string that goes through this method is a string literal in cirq. There is no situation where these strings are constructed by pasting together bits of other strings. Well, actually, in tests this does happen in the sense that you call the repr and then eval it. But calling eval inside the eval isn't any *more* problematic.

If an attacker modifies the repr to include a bad string argument here then they can do bad things. But in that case the user already pasted a string into their python interpreter from an attacker. The attacker could just directly delete files instead of going through the roundabout method of involving this file.

Do you know of a situation where this causes a problem that wouldn't also be present for a raw Callable?

cirq/contrib/quirk/cells/cell.py
Strilanc: Rewritten.

cirq/contrib/quirk/cells/arithmetic_cells.py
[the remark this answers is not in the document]
viathor: Makes sense.

cirq/contrib/quirk/cells/arithmetic_cells.py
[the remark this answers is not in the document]
viathor: Outside of quantum computing, modular arithmetic would just force the value larger or equal to the modulus into the internal 0..modulus-1. In quantum computing we cannot do this due to the reversible nature of unitaries. So instead on values in the internal modulus..2**size-1 one needs another mapping from modulus..2**size-1 into modulus..2**size-1. The easiest is to just use the identity on this interval.

cirq/contrib/quirk/cells/arithmetic_cells.py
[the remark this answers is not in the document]
viathor: Cool.

cirq/contrib/quirk/cells/arithmetic_cells.py
viathor: Use of `eval` here is dangerous and unnecessary.

It's dangerous, because even if you have taken sufficient care to ensure all strings that hide python code are literals today, we have no mechanism to ensure this remains true in future. Also, anything built on top of this by our users is likely to incur a serious security vulnerability.

It's unnecessary, because if we're only concerned with a finite set of lambda expressions, we can easily just list them in a lookup table (and as a minor bonus we'll recover the ability to type-check the code currently hidden in strings).

Strilanc: I switched the implementation to use a global hidden table for the operations.

cirq/contrib/quirk/cells/arithmetic_cells.py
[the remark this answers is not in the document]
Strilanc: Oh, you were referring to the behavior of the operation not to the usage of r. Rewrote it.

cirq/contrib/quirk/cells/arithmetic_cells.py
```diff
@@ -0,0 +1,316 @@
...
+                 inputs: Sequence[Union[None, Sequence['cirq.Qid'], int]]):
+        self.identifier = identifier
+        self.target = target
+        self.inputs = inputs
+
+    @property
+    def operation(self):
+        return ARITHMETIC_OP_TABLE[self.identifier]
+
+    def with_input(self, letter, register):
```
viathor: Missing type annotations, also applied to `operation`.

cirq/contrib/quirk/cells/arithmetic_cells.py
```diff
@@ -0,0 +1,316 @@
...
+
+    # Misc.
+    yield from _arithmetic_family("+cntA", lambda x, a: x + _popcnt(a))
+    yield from _arithmetic_family("-cntA", lambda x, a: x - _popcnt(a))
+    yield from _arithmetic_family("^=A", lambda x, a: x ^ a)
+    yield from _arithmetic_family(
+        "Flip<A", lambda x, a: a - x - 1 if x < a else x)
+
+    # Multiplication.
+    yield from _arithmetic_family("*A", lambda x, a: x * a if a & 1 else x)
```
viathor: This is surprising. Why don't we perform multiplication by a when a is even?

cirq/contrib/quirk/cells/arithmetic_cells.py
```diff
@@ -0,0 +1,316 @@
...
+    """Returns `a**-1` if `a` has a multiplicative inverse, else 1."""
+    if m == 0:
+        return 1
+    gcd, x, _ = _extended_gcd(a % m, m)
+    if gcd != 1:
+        return 1
+    return x % m
+
+
+def _popcnt(a: int) -> int:
```
viathor: Optional nit: `_hamming` is a more informative name for this function.

cirq/contrib/quirk/cells/input_cells.py
```diff
@@ -0,0 +1,75 @@
...
+class InputCell(Cell):
+    """A modifier that provides a quantum input to gates in the same column."""
+
+    def __init__(self, qubits: Iterable['cirq.Qid'], letter: str):
+        self.qubits = tuple(qubits)
+        self.letter = letter
+
+    def modify_column(self, column: List[Optional['Cell']]):
+        for i in range(len(column)):
+            gate = column[i]
```
viathor: s/gate/cell/ is a better name

cirq/contrib/quirk/cells/arithmetic_cells.py
Strilanc: Because that's what Quirk does and we have to match the behavior.

Quirk does it because multiplication by 2 is not a reversible operation on a fixed width binary register. The alternative I considered was to replace factors of 2 by a left rotate, but I rejected that because in the modular case it require factoring the modulus and factoring was too expensive to do while remaining performant.

cirq/contrib/quirk/cells/arithmetic_cells.py
Strilanc: Acknowledged, but popcnt is also common terminology.

cirq/contrib/quirk/cells/input_cells.py
Strilanc: Done.

PR #2264, opened 2019-10-03
Add qubits and extra_recognized parameters to quirk_url_to_circuit

Strilanc:
I needed these as soon as I started playing with it.


cirq/contrib/quirk/url_to_circuit.py
```diff
@@ -13,22 +13,86 @@
...
+        *,
+        qubits: Optional[Sequence['cirq.Qid']] = None,
+        extra_recognized: Iterable['cirq.contrib.quirk.cells.CellMaker'] = ()
+) -> 'cirq.Circuit':
+    """Parses a Cirq circuit out of a Quirk URL.
+
+    Args:
+        quirk_url: The URL of a bookmarked Quirk circuit. It is not required
+            that the domain be "algassert.com/quirk". The only important part of
+            the URL is the fragment (the part after the #).
```
viathor: Waiting for the day when an RFC defines a URL namespace for compact specification of quantum circuits :-)

cirq/contrib/quirk/url_to_circuit.py
```diff
@@ -13,22 +13,86 @@
...
+        qubits: Qubits to use in the circuit. The length of the list must be
+            at least the number of qubits in the Quirk circuit (including unused
+            qubits). The maximum number of qubits in a Quirk circuit is 16.
```
viathor: Question (no action needed): This seems low. What would it take to increase it?

cirq/contrib/quirk/url_to_circuit.py
```diff
@@ -13,22 +13,86 @@
...
+            used to parse URLs that come from a modified version of Quirk that
+            includes gates that Quirk doesn't define. Each entry must be a
+            `cirq.contrib.quirk.cells.CellMaker`, which is a `NamedTuple` with
+            an `identifier` (the string that identifies the gate type), a `size`
+            (the height of the operation; the number of qubits it covers), and
+            a `maker` function which takes a
+            `cirq.contrib.quirk.cells.CellMakerArgs` and returns a
+            `cirq.Operation` or a `cirq.contrib.quirk.cells.Cell`. The cell is
+            more flexible (it can modify other cells in the same column before
+            producing operations).
```
viathor: Move the explanation of the structure of cell makers to where `CellMaker` is defined. When we modify the definition of `CellMaker`, we're unlikely to remember this text needs to be updated, too.

cirq/contrib/quirk/url_to_circuit.py
```diff
@@ -77,16 +142,31 @@ def quirk_url_to_circuit(quirk_url: str) -> 'cirq.Circuit':
...
-        body = cirq.Circuit(
+        body = circuits.Circuit(
             cell.operations() for cell in col if cell is not None)
         result += basis_change
         result += body
         result += basis_change**-1
 
+    # Remap qubits if requested.
+    if qubits is not None:
+        qs = cast(Sequence['cirq.Qid'], qubits)
```
viathor: IIRC, mypy is able to make the inference within a conditional with `is not None`. Is this not the case?

cirq/circuits/circuit.py
```diff
@@ -467,6 +467,31 @@ def prev_moment_operating_on(self,
...
 
+    def transform_qubits(self,
+                         func: Callable[['cirq.Qid'], 'cirq.Qid'],
+                         *,
+                         new_device: 'cirq.Device' = None) -> 'cirq.Circuit':
+        """Returns the same circuit, but with different qubits.
+
+        Note that this method does essentially the same thing as
+        `cirq.Circuit.with_device`. It is included regardless because there are
+        also `transform_qubits` methods on `cirq.Operation` and `cirq.Moment`.
```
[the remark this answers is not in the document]
Strilanc: I think there's a clear conceptual distinction between the two, they just happen to overlap in functionality due to convenience extras here. In any case this was part of the stuff that was in #2263; it wasn't supposed to be in the diff here.

cirq/contrib/quirk/cells/cell.py
```diff
@@ -127,5 +127,7 @@ def controlled_by(self, qubit: 'cirq.Qid') -> 'ExplicitOperationsCell':
 CellMaker = NamedTuple('CellMaker', [
     ('identifier', str),
     ('size', int),
-    ('maker', Callable[[CellMakerArgs], Optional[Cell]]),
+    ('maker',
+     Callable[[CellMakerArgs],
+              Union[None, 'Cell', 'cirq.Operation']]),
```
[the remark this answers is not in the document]
Strilanc: It's still needed for operations that include a basis change. This happens for the measurement gates. That may change in the future.

cirq/contrib/quirk/url_to_circuit.py
```diff
@@ -13,22 +13,86 @@
...
-def quirk_url_to_circuit(quirk_url: str) -> 'cirq.Circuit':
+if TYPE_CHECKING:
+    import cirq
+
+
+def quirk_url_to_circuit(
+        quirk_url: str,
+        *,
+        qubits: Optional[Sequence['cirq.Qid']] = None,
+        extra_recognized: Iterable['cirq.contrib.quirk.cells.CellMaker'] = ()
```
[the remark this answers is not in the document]
Strilanc: Renamed to `extra_cell_makers`.

cirq/contrib/quirk/url_to_circuit.py
Strilanc: Acknowledged.

cirq/contrib/quirk/url_to_circuit.py
Strilanc: When I increase the limit in Quirk, I observe instabilities like running out of video memory and just plain returning the wrong answers. Even at 16 I already have to do gymnastics to avoid precision issues with the frag coord input to the shaders. Basically it would require a lot of careful thinking about what guarantees WebGL can provide.

cirq/contrib/quirk/url_to_circuit.py
Strilanc: Done.

cirq/contrib/quirk/url_to_circuit.py
Strilanc: The issue is that qubits is used inside the lambda, which involves capturing and deferred execution semantics. Mypy does not make the inference.

cirq/contrib/quirk/url_to_circuit.py
viathor: Interesting, thanks!

cirq/contrib/quirk/url_to_circuit.py
```diff
@@ -13,22 +13,79 @@
...
+        ...     'http://algassert.com/quirk#circuit={"cols":[["H"],["•","X"]]}',
+        ...     qubits=[cirq.NamedQubit('Alice'), cirq.NamedQubit('Bob')]
+        ... ))
+        Alice: ───H───@───
+                      │
+        Bob: ─────────X───
+
+        >>> print(cirq.contrib.quirk.quirk_url_to_circuit(
+        ...     'http://algassert.com/quirk#circuit={"cols":[["iswap"]]}',
+        ...     extra_recognized=[
```
viathor: `extra_cell_makers`

cirq/contrib/quirk/url_to_circuit_test.py
```diff
@@ -80,3 +80,37 @@ def test_parse_not_supported_yet():
...
+            cirq.H(a),
+            cirq.X(c).controlled_by(a))
+
+    with pytest.raises(IndexError, match="qubits specified"):
+        _ = quirk_url_to_circuit(
+            'http://algassert.com/quirk#circuit={"cols":[["H"],["•","X"]]}',
+            qubits=[cirq.GridQubit(0, 0)])
+
+
+def test_extra_recognized():
```
viathor: `test test_extra_cell_makers`

PR #2266, opened 2019-10-04
Add compiling (deconstruction, routing, optimizing) to quantum volume algorithm

KevinVillela:
This should be the last PR before we are able to run it on a real device. The QV program now not only prints out the results, but also returns a QuantumVolumeResult that contains all of the relevant information for further analysis.

This PR also contains some additions to the existing QV Jupyter notebook that adds scatter plot points for the routed and the optimized results of running on the given simulator. Finally, it adds a new Jupyter notebook which plots the expected HOG probability vs a series of device error rates for a given depth/number of qubits.


examples/advanced/quantum_volume.py
```diff
@@ -1,24 +1,32 @@
...
-
-Usage:
-    python examples/advanced/quantum_volume.py \
-        --num_qubits=4 --depth=4 --num_repetitions=1 [--seed=int]
+https://arxiv.org/abs/1811.12926. By default, this runs on the Bristlecone
+device.
 
+Usage: python examples/advanced/quantum_volume.py \
+         --num_qubits=4 --depth=4 --num_repetitions=1 [--seed=int]
+nn
```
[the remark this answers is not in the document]
KevinVillela: Grr the only complaint I have about the Mac butterfly keyboard is that it doesn't register ctrl/alt sometimes... Done.

examples/advanced/quantum_volume.py
```diff
@@ -136,7 +153,86 @@ def sample_heavy_set(circuit: cirq.Circuit,
...
+            'model_circuits', 'heavy_sets', 'compiled_circuits',
+            'sampler_results'
+        ])
+
+
+def main(*,
+         num_qubits: int,
+         depth: int,
+         num_repetitions: int,
+         seed: int,
```
mpharrigan: missing a docstring for seed

examples/advanced/quantum_volume.py
```diff
@@ -136,7 +153,86 @@ def sample_heavy_set(circuit: cirq.Circuit,
...
+         device: cirq.google.xmon_device.XmonDevice,
```
mpharrigan: missing docstring for device

examples/advanced/quantum_volume.py
```diff
@@ -151,19 +247,36 @@ def main(num_qubits: int, depth: int, num_repetitions: int, seed: int):
         num_qubits: The number of qubits for the circuit.
         depth: The number of gate layers to generate.
         num_repetitions: The number of times to run the algorithm.
+        samplers: The samplers to run the algorithm on.
+        deconstruct: An optional function to deconstruct the model circuit's
+            gates down to the targer devices gate set.
```
mpharrigan: targer -> target

examples/advanced/quantum_volume.py
```diff
@@ -151,19 +247,36 @@ def main(num_qubits: int, depth: int, num_repetitions: int, seed: int):
         num_qubits: The number of qubits for the circuit.
         depth: The number of gate layers to generate.
         num_repetitions: The number of times to run the algorithm.
+        samplers: The samplers to run the algorithm on.
+        deconstruct: An optional function to deconstruct the model circuit's
```
mpharrigan: hmm

1. I think the name `deconstruct` is unexpected. Maybe "compiler"?
2. What's the difference between this and `optimize`? 
3. You could consider also taking a callable (or similar) to control which routing algorithm is used

examples/advanced/quantum_volume.py
```diff
@@ -136,7 +153,86 @@ def sample_heavy_set(circuit: cirq.Circuit,
...
+         num_repetitions: int,
+         seed: int,
+         device: cirq.google.xmon_device.XmonDevice,
+         samplers: List[cirq.Sampler],
+         deconstruct: PointOptimizer = None,
+         optimize: Callable[[cirq.Circuit], cirq.Circuit] = None
+        ) -> QuantumVolumeResult:
     """Run the quantum volume algorithm.
 
     The Quantum Volume benchmark is fairly straightforward. This algorithm will
```
mpharrigan: can we remove the line "The Quantum Volume benchmark is fairly straightforward"? At best it doesn't really add anything and at worst it could discourage a user who doesn't find it straightforward :)

examples/advanced/quantum_volume.py
```diff
@@ -136,7 +153,86 @@ def sample_heavy_set(circuit: cirq.Circuit,
...
+    """
+
+    def __init__(self):
+        # All of the model circuits used.
+        self.model_circuits: [cirq.Circuit] = []
+        # All of the heavy sets computed from the give model circuits.
+        self.heavy_sets: [List[int]] = []
+        # All of the model circuits after being compiled.
+        self.compiled_circuits: [cirq.Circuit] = []
+        # A map from each sampler index top a list of its HOG probabilities.
```
mpharrigan: top -> to

examples/advanced/quantum_volume.py
```diff
@@ -136,7 +153,86 @@ def sample_heavy_set(circuit: cirq.Circuit,
...
+        self.sampler_results: Dict[int, [float]] = defaultdict(list)
```
mpharrigan: `List[float]`

examples/advanced/quantum_volume.py
```diff
@@ -136,7 +153,86 @@ def sample_heavy_set(circuit: cirq.Circuit,
...
+    # Quantum Volume is intended to test those as well, we allow this to be
+    # passed in. This optimizer is not allowed to change the order of the
+    # qubits.
+    if optimize:
+        compiled_circuit = optimize(compiled_circuit)
+
+    return (compiled_circuit, swap_network.final_mapping())
+
+
+class QuantumVolumeResult:
```
mpharrigan: I'd recommend "transposing" your classes.

What does this mean? Imagine your data as a table with each repetition being a row and each (model_circuit, heavy_set, ...) being a column. `QuantumVolumeResult` is a list (essentially) of columns. You'll find it more easy to wrangle data if instead you have a list of rows, where each row can be represented with a custom class

```python
@dataclass
class result:
  model_circuit
  heavy_set
  compiled_circuit
  sampler_result

def main():
  results = []
  for _ in range(reps):
    results.append(result(
      model_circuit=...
    ))
```

Bonus: now your (data)class can be immutable. 

Bonus: your invariant that all your current lists have to be the same length is trivially enforced

examples/advanced/quantum_volume.py
```diff
@@ -151,19 +247,36 @@ def main(num_qubits: int, depth: int, num_repetitions: int, seed: int):
...
+            gates down to the targer devices gate set.
+        optimize: An optional function to optimize the circuit's gates
+            after routing.
     """
-    for _ in range(num_repetitions):
+    result = QuantumVolumeResult()
+    for repetition in range(num_repetitions):
+        print(f"Repetition {repetition + 1}")
         model_circuit = generate_model_circuit(
             num_qubits, depth, random_state=np.random.RandomState(seed))
```
mpharrigan: eek! bad seeding. Create one random state at the start. Every circuit will be the same otherwise

examples/advanced/quantum_volume.py
```diff
@@ -151,19 +247,36 @@ def main(num_qubits: int, depth: int, num_repetitions: int, seed: int):
...
-              f"{sample_heavy_set(model_circuit, heavy_set)}")
-        noisy = cirq.DensityMatrixSimulator(noise=cirq.ConstantQubitNoiseModel(
-            qubit_noise_gate=cirq.DepolarizingChannel(p=0.005)))
-        print(f"Noisy simulation probability: "
-              f"{sample_heavy_set(model_circuit, heavy_set, sampler=noisy)}")
-        # TODO(villela): Implement model circuit and run it.
+        result.heavy_sets.append(heavy_set)
+        print(f"  Heavy Set: {heavy_set}")
+
+        [compiled_circuit, mapping] = compile_circuit(model_circuit,
```
mpharrigan: nit: brackets unnecessary

examples/advanced/quantum_volume.py
```diff
@@ -151,19 +247,36 @@ def main(num_qubits: int, depth: int, num_repetitions: int, seed: int):
...
+                                                      device=device,
+                                                      deconstruct=deconstruct,
+                                                      optimize=optimize)
+        result.compiled_circuits.append(compiled_circuit)
+        for idx, sampler in enumerate(samplers):
+            prob = sample_heavy_set(compiled_circuit,
```
mpharrigan: outside the scope of this PR, but:

This is a weird loop to have because it indicates that your functions aren't very composable. Prefer separating your script into two "tasks": 1. problem generation 2. problem execution. You can imagine wanting to do all the hard hog computation up front (once) and then collect data using various samplers on a whim, perhaps at regular time intervals. Imagine we'd want to run this benchmark once a week to track performance of a device over time

examples/advanced/quantum_volume.py
```diff
@@ -191,4 +304,14 @@ def parse_arguments(args):
 
 
 if __name__ == '__main__':
-    main(**parse_arguments(sys.argv[1:]))
+    device = cirq.google.Bristlecone
```
mpharrigan: put this code in a function. I'd suggest: this code goes in `main()` and you rename `main` to something like `calculate_quantum_volume` (or something meaningful)

reason: you're actually declaring global variables here. PyCharm tells me that your usage of the variable `optimize` in other functions shadows this outer definition which is probably fine in practice but could be bad

examples/advanced/quantum_volume.py
```diff
@@ -136,7 +153,86 @@ def sample_heavy_set(circuit: cirq.Circuit,
...
+                                     algo_name='greedy')
+    compiled_circuit = swap_network.circuit
+    # Step 3: Optimize. The paper uses various optimization techniques - because
+    # Quantum Volume is intended to test those as well, we allow this to be
+    # passed in. This optimizer is not allowed to change the order of the
+    # qubits.
+    if optimize:
+        compiled_circuit = optimize(compiled_circuit)
+
+    return (compiled_circuit, swap_network.final_mapping())
```
mpharrigan: nit: redundant parentheses

examples/advanced/quantum_volume.ipynb
```diff
@@ -50,38 +54,68 @@
...
+    "        # Run the given samplers over the model, compiled, and optimized circuits.\n",
     "        for sampler in samplers:\n",
     "            probability = quantum_volume.sample_heavy_set(\n",
     "                model_circuit, heavy_set, sampler=sampler['sampler'])\n",
-    "            sampler['probabilities'][depth - depths[-1]] += probability\n",
+    "            sampler['probabilities'][depth] += probability\n",
     "            print(f\"        {sampler['label']} HOG probability: {probability}\")\n",
     "\n",
-    "        # TODO: The next steps are to compile the model circuit and run it on various devices.\n",
+    "            compiled_probability = quantum_volume.sample_heavy_set(\n",
```
mpharrigan: outside the scope of this PR:

If you refactor the code into a problem generation task and data collection task (as suggested in another comment), you can probably simplify the notebook and just call those functions. Obviously this is a balance between making the notebook instructive (i.e. not a complete black box) and concise

examples/advanced/quantum_volume.ipynb
```diff
@@ -98,23 +132,34 @@
...
+    "    axs.scatter([d + idx / 10 for d in depths],\n",
+    "                sampler['compiled-probabilities'].values(),\n",
+    "                marker=\"s\",\n",
+    "                c=f\"{sampler['color']}\",\n",
+    "                label=f\"{sampler['label']} Compiled\")\n",
+    "    axs.scatter([d + idx / 10 for d in depths],\n",
+    "                sampler['optimized-probabilities'].values(),\n",
+    "                marker=\"o\",\n",
+    "                c=f\"{sampler['color']}\",\n",
+    "                label=f\"{sampler['label']} Optimized\")\n",
```
mpharrigan: language is getting dicey: 

"optimized" vs. "compiled" is a distinction that isn't meaningful to me. A verbose but clear distinction could be between "routed" and "gate compiled".

Right now you've split up your "gate compiled" step into a pre-routing step and a post-routing step. I've been playing around with the code a bit and I found it to be quite nice to have the model circuit emit `TwoQubitMatrixGate`s instead of decomposed-into-cz. Then your gate compilation can happen all at once (after routing)

examples/advanced/quantum_volume.py
KevinVillela: Ah sure, that makes sense. Is it OK if I do that in a PR right after this one, so this doesn't get too big?

examples/advanced/quantum_volume.py
KevinVillela: Ah, good to know, thanks! Didn't realize it, but in retrospect it's pretty clear that this is in the global scope.

examples/advanced/quantum_volume.py
KevinVillela: Done

examples/advanced/quantum_volume.py
KevinVillela: Done

examples/advanced/quantum_volume.py
KevinVillela: Done

examples/advanced/quantum_volume.py
KevinVillela: Done

examples/advanced/quantum_volume.py
KevinVillela: 1. Changed to `compiler`.
2. The difference is that the `deconstruct` (now `compiler`) step doesn't do any optimizations (e.g. combining of gates) - instead, it just decomposes the model gates down into a gate set supported by the device. `optimize` actually elides and combines gates after routing is complete.
3. Done.

examples/advanced/quantum_volume.py
KevinVillela: Haha yea, I still remember the lesson of "don't use the word 'obvious' in a paper" from one of my college profs... Looks like I forgot that here. Done.

examples/advanced/quantum_volume.ipynb
KevinVillela: Yea I struggled with how much to reproduce in here to make it instructional. I agree that it could be cleaned up a bit - as above, I'd prefer to do it in the next PR if that's OK.

examples/advanced/quantum_volume.ipynb
KevinVillela: Nice, love the `TwoQubitMatrixGate`s, the code is much simpler now!

examples/advanced/quantum_volume.py
KevinVillela: Done.

examples/advanced/quantum_volume.py
KevinVillela: I think this would need to be a List[List[float]], since each element is a list of all the output probabilities. Done.

examples/advanced/quantum_volume.py
KevinVillela: Done.

examples/advanced/quantum_volume.py
KevinVillela: Done.

examples/advanced/quantum_volume.py
KevinVillela: Ah that's cleaner. I was thinking that in the end I wanted to print out this one large object, but having something that behaves as a bunch of rows will be nicer.

KevinVillela:
Thanks for the in-depth review, I've responded to all the comments - PTAL at your leisure :)

KevinVillela:
I had to update the travis config to get the tests to run on Python 3.7 in Mac and Linux (it was already the case for Windows), which is required to support dataclasses. That means that this benchmark now requires Python 3.7 to run as well. Let me know if that's not OK, and also if you want me to move the Travis changes to a different CL.

examples/advanced/quantum_volume.py
mpharrigan: of course

mpharrigan:
Sorry for the delay. Looks great now! 

I think @Strilanc will yell at us if we bump the python version here. Add `dataclasses` to the requirements. It's available as a backport for python 3.6 (and the package is a no-op for 3.7+)

KevinVillela:
Craig has been nothing but nice to me, I'd hate to anger him with a .travis.yml update :) Added to the requirements, tests seem to pass now. Thanks!

PR #2272, opened 2019-10-07
Create DensePauliString and MutableDensePauliString

Strilanc:
This class achieves a few purposes.

1. (For @dabacon) It is the "gate" for PauliString.
2. (For @harryputterman) It doesn't require you to specify qubits.
3. It should have better performance than `cirq.PauliString` because operations are vectorized over numpy arrays and don't involve dictionaries.

Example usage:

```
>>> print(cirq.DensePauliString('XXIY'))
+XXIY

>>> print(cirq.MutableDensePauliString('IZII', coefficient=-1))
-IZII (mutable)

>>> print(cirq.DensePauliString([0, 1, 2, 3],
...                             coefficient=sympy.Symbol('t')))
t*IXZY

>>> a = cirq.DensePauliString('IXYZ')
>>> b = cirq.DensePauliString('XXZZ')
>>> print(a * b)
1j*XIXI
```

I was particularly happy to find a compact way to compute the phase of the product of two Paulis. If I=0, X=1, Z=2, Y=3 then the following computes the `t` in `1j**t` that is the complex coefficient resulting from multiplying two Paulis:

```python
    t = lhs * (rhs != 0)
    t -= rhs * (lhs != 0)
    t += 1
    t %= 3
    t -= 1
```

The basic idea is to force all (0, k) and (k, 0) to (0, 0) so that you can safely do the "is the cycle advancing or reversing" modulo 3 check on the remaining inputs.


cirq/ops/pauli_string.py
```diff
@@ -164,6 +164,16 @@ def __mul__(self, other) -> 'PauliString':
                            qubit_pauli_map=self._qubit_pauli_map,
                            coefficient=self.coefficient)
 
+    @property
+    def gate(self) -> 'cirq.DensePauliString':
+        order: List[Optional[pauli_gates.Pauli]] = [
+            None, pauli_gates.X, pauli_gates.Z, pauli_gates.Y
```
mpharrigan: You're generally not supposed to have mutable default function arguments (like list)

mpharrigan:
I think people will get tripped up by the order being X, Z, Y

cirq/ops/pauli_string.py
Strilanc: That's not an argument, it's a local variable with a type declaration.

mpharrigan: ah, my brain got confused

mpharrigan: this type annotation seems to hurt rather than help readability, imo

Strilanc:
I re-ordered the indices.

PR #2275, opened 2019-10-07
Vectorized implementation of KAK decomposition.

dkafri:
Adds a `kak_vector` method that computes the interaction coefficients (but not the single qubit unitaries) of two-qubit gates a hundred times faster than `kak_decomposition`.


cirq/linalg/decompositions_test.py
```diff
@@ -18,6 +18,7 @@
 import pytest
 
 import cirq
+from cirq.linalg.decompositions import kak_vector
```
dkafri: I didn't see how to run the test without this line. I tried adding `kak_vector` to some `__init__.py` files but that didn't work.

cirq/linalg/decompositions.py
```diff
@@ -854,3 +857,98 @@ def kak_decomposition(unitary_object: Union[np.ndarray, 'cirq.SupportsUnitary'],
         global_phase=g * inner_cannon.global_phase,
         single_qubit_operations_before=(b1, b0),
         single_qubit_operations_after=(a1, a0))
+
+
+def kak_vector(unitary: np.ndarray):
```
Strilanc: Argument type is `Union[Iterable[np.ndarray], np.ndarray]`

cirq/linalg/decompositions.py
```diff
@@ -854,3 +857,98 @@ def kak_decomposition(unitary_object: Union[np.ndarray, 'cirq.SupportsUnitary'],
...
+    UB = np.einsum('...ab,...bc,...cd', MAGIC_CONJ_T, unitary, MAGIC)
+
+    m = np.einsum('...ab,...cb', UB, UB)
+
+    evals, _ = np.linalg.eig(m)
+
+    # The algorithm in the appendix mentioned above is slightly incorrect in
+    # that it only works for elements of SU(4). A phase correction must be
+    # added to deal with U(4).
+    phases = np.log(-1j * np.linalg.det(unitary)).imag + np.pi / 2
```
Strilanc: Could you benchmark the speed of this method applied to 1000 random unitaries (from `cirq.testing.random_unitary(4)`) and compare it to the speed of the existing decomposition method? So basically:

```
import time
unitaries = [cirq.testing.random_unitary(4) for _ in range(1000)]
t0 = time.monotonic()
_ = kak_vector(unitaries)
t1 = time.monotonic()
print(t1 - t0, 'secs')
```

cirq/linalg/decompositions.py
```diff
@@ -854,3 +857,98 @@ def kak_decomposition(unitary_object: Union[np.ndarray, 'cirq.SupportsUnitary'],
...
+    $$ A = \exp( i * \sum_{j=x,y,z} k_j \sigma_{j,0}\sigma{j,1} ) $$
+
+    where $(k_x,k_y,k_z)$ is in the Weyl chamber, a tetrahedron with corners
+    $$(0,0,0), (\pi/2,0,0), (\pi/4,pi/4,0), (\pi/4,\pi/4,\pi/4)$$
+
+    The vector entries are ordered such that
+    $$ \pi/2 - k_y \geq k_x \geq k_y \geq z \geq 0 $$
+
+    Args:
+        unitary: Shape (...,4,4) unitary numpy.ndarray.
```
Strilanc: Expand this description.

```
unitary: A unitary matrix, or a multi-dimensional array of unitary matrices, specified as a numpy array or iterable of numpy arrays. Must have shape (..., 4, 4), where the last two axes are for the unitary matrix and other axes are for broadcasting the kak vector computation.
```

cirq/linalg/decompositions.py
```diff
@@ -854,3 +857,98 @@ def kak_decomposition(unitary_object: Union[np.ndarray, 'cirq.SupportsUnitary'],
...
+
+    Returns:
+        The KAK vector of the unitary, output shape (...,3).
```
Strilanc:
```
The KAK vector of the given unitary or unitaries. The output shape is the same as the input shape, except the two unitary matrix axes are replaced by the kak vector axis (i.e. the output has shape `unitary.shape[:-2] + (,3)`).
```

cirq/linalg/decompositions.py
```diff
@@ -854,3 +857,98 @@ def kak_decomposition(unitary_object: Union[np.ndarray, 'cirq.SupportsUnitary'],
...
+
+    References:
```
Strilanc: Add an `Examples` section

```
Examples:
    >>> cirq.kak_vector(np.eye(4))
    [0, 0, 0]

    >>> cirq.kak_vector([cirq.unitary(cirq.CZ), cirq.unitary(cirq.ISWAP)])
    [[1, 0, 0], [1, 1, 0]]
```

Be sure to use the `>>> line` then `result` style, as that way the example will be tested by doctest.

cirq/linalg/decompositions_test.py
Strilanc: You need to include `kak_vector` in the big import lists within `cirq/linalg/__init__.py` and `cirq/__init__.py`.

cirq/linalg/decompositions.py
Strilanc: Add a `kak_vector` entry in `api.rst` right after `kak_decomposition`, so that it gets included in documentation. Rendered output looks like this currently (note sigma spelled instead of symbol):

image

cirq/linalg/decompositions.py
```diff
@@ -854,3 +857,98 @@ def kak_decomposition(unitary_object: Union[np.ndarray, 'cirq.SupportsUnitary'],
...
+
+    $$ U = k_l A k_r $$
+    where $k_l, k_r$ are single qubit (local) unitaries and
+    $$ A = \exp( i * \sum_{j=x,y,z} k_j \sigma_{j,0}\sigma{j,1} ) $$
+
+    where $(k_x,k_y,k_z)$ is in the Weyl chamber, a tetrahedron with corners
+    $$(0,0,0), (\pi/2,0,0), (\pi/4,pi/4,0), (\pi/4,\pi/4,\pi/4)$$
+
+    The vector entries are ordered such that
+    $$ \pi/2 - k_y \geq k_x \geq k_y \geq z \geq 0 $$
```
Strilanc: The canonicalization we use is actually 

```
0 ≤ abs(z2) ≤ y2 ≤ x2 ≤ π/4
if x2 = π/4, z2 >= 0
```

This canonicalization has the benefit that operations that are near the identity are near (0, 0, 0) instead of possibly being near (pi/2, 0, 0). And the only place where there's a discontinuity vs the 3d euclidean distance is on the x=max face, where z can flip sign from small perturbations to the unitary matrix.

cirq/linalg/decompositions_test.py
```diff
@@ -517,3 +518,119 @@ def test_scatter_plot_normalized_kak_interaction_coefficients():
...
+
+def test_KAK_vector_local_invariants_random_input():
+    actual = _local_invariants_from_kak(kak_vector(_random_unitaries))
+    expected = _local_invariants_from_kak(_kak_vecs)
+
+    np.testing.assert_almost_equal(actual, expected)
+
+
+def test_kak_vector_on_weyl_chamber_face():
+    # unitaries with KAK vectors from I to ISWAP
```
Strilanc: You can use `cirq.unitary(cirq.KakDecomposition(...))` as a convenient way to generate unitaries with known interaction coefficients.

cirq/linalg/decompositions_test.py
```diff
@@ -517,3 +518,119 @@ def test_scatter_plot_normalized_kak_interaction_coefficients():
...
+    np.testing.assert_almost_equal(actual, expected)
+
+
+@pytest.mark.parametrize('unitary,expected',
+                         ((np.eye(4), (0, 0, 0)), (SWAP, [np.pi / 4] * 3),
+                          (SWAP * 1j, [np.pi / 4] * 3),
+                          (CNOT, [np.pi / 4, 0, 0]), (CZ, [np.pi / 4, 0, 0]),
+                          (CZ @ SWAP, [np.pi / 4, np.pi / 4, 0]),
+                          (np.kron(X, X), (0, 0, 0))))
+def test_KAK_vector_weyl_chamber_vertices(unitary, expected):
```
Strilanc: You also need tests of failure cases, such as the user passing in incorrectly shaped data. `with pytest.raise(ValueError, match="bad shape"): ...`

cirq/linalg/decompositions.py
Strilanc: Nevermind this. I did a cursory check and got this:

New method: `0.15s` to do 10K instances.
Old method: `5.2s` to do 10K instances.
So basically a 30x speedup.

```
    import time
    unitaries = [cirq.testing.random_unitary(4) for _ in range(10000)]
    t0 = time.monotonic()
    _ = kak_vector(unitaries)
    t1 = time.monotonic()
    print(t1 - t0, 'secs')
    t0 = time.monotonic()
    for u in unitaries:
        _ = cirq.kak_decomposition(u)
    t1 = time.monotonic()
    print(t1 - t0, 'secs')
```

cirq/linalg/decompositions.py
dkafri: How do I check the rendered output?

vtomole: Build the docs locally by running https://github.com/quantumlib/Cirq/blob/master/dev_tools/build-docs.sh.

Strilanc: Currently there isn't a fast convenient way other than to invoke `./dev_tools/build-docs.sh` from the repo root.

cirq/linalg/decompositions_test.py
dkafri: Show I add checks for unitarity like in kak_decomposition? I would need to vectorize those as well...

cirq/linalg/decompositions_test.py
dkafri: Done.

cirq/linalg/decompositions.py
```diff
@@ -636,13 +636,16 @@ def coord_transform(
...
+    if not isinstance(interactions, np.ndarray):
+        interactions = [
+            a if isinstance(a, np.ndarray) else protocols.unitary(a)
+            for a in interactions
+        ]
+
+    points = kak_vector(interactions) * 4 / np.pi
 
-    ax.scatter(*coord_transform(points), **kwargs)
+    ax.scatter(*np.moveaxis(points[..., (0, 2, 1)], -1, 0), **kwargs)
```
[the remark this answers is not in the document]
dkafri: Done.

cirq/linalg/decompositions.py
```diff
@@ -854,3 +857,98 @@ def kak_decomposition(unitary_object: Union[np.ndarray, 'cirq.SupportsUnitary'],
         global_phase=g * inner_cannon.global_phase,
         single_qubit_operations_before=(b1, b0),
         single_qubit_operations_after=(a1, a0))
+
+
+def kak_vector(unitary: np.ndarray):
+    r"""Compute the KAK vectors of a two qubit unitary.
```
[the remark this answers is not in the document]
dkafri: Done.

cirq/linalg/decompositions.py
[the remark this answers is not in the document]
dkafri: Done.

dkafri: Done.

cirq/linalg/decompositions.py
dkafri: Done.

cirq/linalg/decompositions.py
dkafri: Done.

cirq/linalg/decompositions.py
dkafri: Done.

cirq/linalg/decompositions.py
dkafri: [This](https://screenshot.googleplex.com/zoupWMO1opG) is what I get when I try to build. I previously ran `pip install pypandoc` and it seems to have installed.

cirq/linalg/decompositions.py
dkafri: Changed the canonicalization to match the one above.

cirq/linalg/decompositions_test.py
dkafri: Done

cirq/linalg/decompositions.py
Strilanc: Try `pip install -r dev_tools/conf/pip-list-dev-tools.txt`

cirq/linalg/decompositions_test.py
Strilanc: Yeah, we should check that.

I guess it's not too hard to vectorize, but don't worry too much about it. Just make sure to include a `check_preconditions=True` argument so that they can be skipped.

cirq/linalg/decompositions.py
```diff
@@ -854,3 +857,129 @@ def kak_decomposition(unitary_object: Union[np.ndarray, 'cirq.SupportsUnitary'],
...
+        $$ 0 ≤ abs(k_z) ≤ k_y ≤ k_x ≤ π/4 $$
+            if $k_x$ = π/4, $k_z \geq 0$
+
+    Args:
+        unitary: A unitary matrix, or a multi-dimensional array of unitary
+            matrices. Must have shape (..., 4, 4), where the last two axes are
+            for the unitary matrix and other axes are for broadcasting the kak
+            vector computation.
+        atol: How close $k_x$ must be to π/4 to guarantee $k_z$ >= 0. Must be
+            positive.
```
Strilanc: non-negative. The user can set it to zero if they want. It does make it rather unlikely to matter.

cirq/linalg/decompositions.py
```diff
@@ -854,3 +857,129 @@ def kak_decomposition(unitary_object: Union[np.ndarray, 'cirq.SupportsUnitary'],
...
+        >>> unitaries = [cirq.unitary(cirq.CZ),cirq.unitary(cirq.ISWAP)]
+        >>> cirq.kak_vector(unitaries) * 4/np.pi
+        array([[ 1.,  0., -0.],
+               [ 1.,  1.,  0.]])
+    """
+    unitary = np.asarray(unitary)
+
+    if unitary.ndim < 2 or unitary.shape[-2:] != (4, 4):
+        raise ValueError(f'Expected input unitary to have shape (...,4,4), but'
+                         'got {unitary.shape}.')
```
Strilanc: Does this actually format? There's no `f` on the second string.

cirq/linalg/decompositions.py
```diff
@@ -854,3 +857,129 @@ def kak_decomposition(unitary_object: Union[np.ndarray, 'cirq.SupportsUnitary'],
...
+    k_vec = np.mod(k_vec + np.pi / 4, np.pi / 2) - np.pi / 4
+
+    # Sort in descending order with respect to absolute value.
+    order = np.argsort(np.abs(k_vec), axis=-1)
+    k_vec = np.take_along_axis(k_vec, order, axis=-1)[..., ::-1]
+
+    # Multiply x,z and y,z components by -1 to fix x,y sign.
+    x_negative = k_vec[..., 0] < 0
+    if np.any(x_negative):
+        k_vec[x_negative, 0] *= -1
```
Strilanc: I think `k_vec[x_negative, [0, 2]] *= -1` works.

cirq/linalg/decompositions.py
```diff
@@ -854,3 +857,129 @@ def kak_decomposition(unitary_object: Union[np.ndarray, 'cirq.SupportsUnitary'],
...
+    # Get all strengths to (-¼π, ¼π]
+    k_vec = np.mod(k_vec + np.pi / 4, np.pi / 2) - np.pi / 4
+
+    # Sort in descending order with respect to absolute value.
+    order = np.argsort(np.abs(k_vec), axis=-1)
+    k_vec = np.take_along_axis(k_vec, order, axis=-1)[..., ::-1]
+
+    # Multiply x,z and y,z components by -1 to fix x,y sign.
+    x_negative = k_vec[..., 0] < 0
+    if np.any(x_negative):
```
Strilanc: Checking the condition probably takes about the same amount of time as doing the multiplication. If the single-statement version works, just do it unconditionally.

cirq/linalg/decompositions.py
```diff
@@ -854,3 +857,129 @@ def kak_decomposition(unitary_object: Union[np.ndarray, 'cirq.SupportsUnitary'],
...
+    if np.any(y_negative):
+        k_vec[y_negative, 1] *= -1
+        k_vec[y_negative, 2] *= -1
+
+    # If x = π/4, force z to be positive.
+    x_is_pi_over_4 = np.isclose(k_vec[..., 0], np.pi / 4, atol=atol)
+    z_is_negative = k_vec[..., 2] < 0
+    need_diff = np.logical_and(x_is_pi_over_4, z_is_negative)
+    if np.any(need_diff):  # -1 to x and z components, then shift x up by pi/4
+        k_vec[need_diff, 2] *= -1
```
Strilanc: I'd just do this unconditionally.

cirq/linalg/decompositions.py
```diff
@@ -854,3 +857,129 @@ def kak_decomposition(unitary_object: Union[np.ndarray, 'cirq.SupportsUnitary'],
...
+    y_negative = k_vec[..., 1] < 0
+    if np.any(y_negative):
+        k_vec[y_negative, 1] *= -1
+        k_vec[y_negative, 2] *= -1
+
+    # If x = π/4, force z to be positive.
+    x_is_pi_over_4 = np.isclose(k_vec[..., 0], np.pi / 4, atol=atol)
+    z_is_negative = k_vec[..., 2] < 0
+    need_diff = np.logical_and(x_is_pi_over_4, z_is_negative)
+    if np.any(need_diff):  # -1 to x and z components, then shift x up by pi/4
```
Strilanc: The comment differs from what the code does.

cirq/linalg/decompositions.py
dkafri: Nope. I tried that first and it throws an `IndexError`.

cirq/linalg/decompositions.py
dkafri: Good catch.

cirq/linalg/decompositions.py
dkafri: Good point. I think the error I noticed from the previous comment made me think I needed this check.

cirq/linalg/decompositions.py
dkafri: I added a clarification. The comment refers to the operations allowed under local unitaries.

cirq/linalg/decompositions.py
dkafri: Done

cirq/linalg/decompositions.py
dkafri: Done

cirq/linalg/decompositions.py
dkafri: Done.

cirq/linalg/decompositions.py
dkafri: Done.

cirq/linalg/decompositions_test.py
dkafri: Done.

cirq/linalg/decompositions.py
dkafri: That seems to fix it, thanks! The docstring seems to build correctly now.

dkafri:
@googlebot I fixed it.

cirq/linalg/decompositions.py
```diff
@@ -854,3 +857,145 @@ def kak_decomposition(unitary_object: Union[np.ndarray, 'cirq.SupportsUnitary'],
...
+        unitary: A unitary matrix, or a multi-dimensional array of unitary
+            matrices. Must have shape (..., 4, 4), where the last two axes are
+            for the unitary matrix and other axes are for broadcasting the kak
+            vector computation.
+        rtol: Per-matrix-entry relative tolerance on equality. Used in unitarity
+            check of input.
+        atol: Per-matrix-entry absolute tolerance on equality. Used in unitarity
+            check of input. This also determines how close $k_x$ must be to π/4
+            to guarantee $k_z$ ≥ 0. Must be non-negative.
+        check_preconditions: If set, verifies that the input corresponds to a
```
Strilanc: The `if set` is confusing here. It reads like if you don't pass the value it won't verify.

`When set to False, skips verifying that the input is unitary in order to increase performance.`

dkafri: I'll modify the docstring then merge the commit.

Strilanc:
Ah, this is the pagination issue again.

PR #2280, opened 2019-10-07
Move MeasurementGate, measure, and measure_each out of common_gates.py

Strilanc:
- The file has become too large, and we are splitting it up.


mpharrigan:
xref https://github.com/quantumlib/Cirq/pull/2205#issuecomment-539260440

Can we keep compatibility imports in `common_gates` with a deprecation warning?

Strilanc:
I'm not sure if it's possible to do that. We'd need new tooling at the very least. The issue is that these things are attributes at the module level, and we don't have a way to intercept the access. So we would effectively need to intercept usage after access, which would involve e.g. wrapping every element on the class.

I did originally want to prefix all the files with `_` to avoid this sort of thing.

cirq/ops/measure_util.py
```diff
@@ -0,0 +1,82 @@
+# Copyright 2018 The Cirq Developers
```
dabacon: 2019

cirq/ops/measure_util_test.py
```diff
@@ -0,0 +1,65 @@
+# Copyright 2018 The Cirq Developers
```
dabacon: 2019

Strilanc:
This file is likely to grow beyond just `MeasurementGate`, e.g. measurement of pauli strings.

Issue #2279, opened 2019-10-07
Add symbol support to `cirq.Duration`


Strilanc:
Eventually we will add a wait gate to cirq. When we do, we will want to parameterize the amount of waiting and sweep over it (e.g. to do a decay experiment). The natural way to do that is by parameterizing `cirq.Duration`.

Basically:

- Instead of only taking integers, also allow `Duration` to accept `sympy.Basic` for `picos` and `nanos`.
- Add some tests making sure that addition and equality and reprs and etc still work as expected. (Use `cirq._compat.proper_repr` on the underlying fields to make sure the symbolic values get repr-ized properly.)

It's okay to return NotImplemented from the comparison operators when a symbolic value is involved.

PR #2281, opened 2019-10-08
Add Symbol and timedelta interop features to cirq.Duration

Strilanc:
- Introduce `DURATION_LIKE` type to replace many instances of `Union[Duration, timedelta]`
- Add micros and millis as supported units
- Improve the `__str__` and `__repr__` methods to use better scaled units
- Allow `_picos` to be symbolic, add `_parameterized_` and `_resolve_parameters_` methods
- Allow comparisons with timedelta and with 0
- Deprecate `Duration.create` in favor of Duration's constructor
- Add `__bool__` method for easy non-zero checking

```
 >>> print(cirq.Duration(micros=1.5 * sympy.Symbol('t')))
(1500.0*t) ns
```

Fixes https://github.com/quantumlib/Cirq/issues/2279


cirq/value/duration.py
```diff
@@ -13,120 +13,234 @@
...
-            return duration
-        if isinstance(duration, timedelta):
-            duration_in_picos = duration.total_seconds() * 10**12
-            return cls(picos=duration_in_picos)
+        return Duration(duration)
 
-        raise TypeError(
-            'Only datetime.timedelta and cirq.Duration are supported.')
+    def _is_parameterized_(self):
+        from cirq import protocols
```
MichaelBroughton: Why is this import happening here ?

cirq/value/duration.py
```diff
@@ -13,120 +13,234 @@
...
+        return protocols.is_parameterized(self._picos)
 
-    def total_picos(self) -> float:
+    def _resolve_parameters_(self, param_resolver):
+        from cirq import protocols
```
MichaelBroughton: same as above

cirq/value/duration.py
```diff
@@ -13,120 +13,234 @@
...
         return self.__mul__(other)
 
     def __truediv__(self, other) -> Union['Duration', float]:
-        if isinstance(other, (int, float)):
+        if isinstance(other, (int, float, sympy.Basic)):
             return Duration(picos=self._picos / other)
-        if isinstance(other, type(self)):
-            return self._picos / other._picos
+
+        other_duration = _attempt_duration_like_to_duration(other)
```
MichaelBroughton: If div can produce a Duration or a Float should mul do the same ?

cirq/value/duration.py
Strilanc: I thought there'd be a circular reference, but upon testing there isn't. Removed.

cirq/value/duration.py
Strilanc: I thought there'd be a circular reference, but upon testing there isn't. Removed.

cirq/value/duration.py
Strilanc: Nope. second/second is a scalar, but second*second is neither a scalar nor a duration.

Strilanc:
I added the DURATION_LIKE to _GateOperation.

Issue #2283, opened 2019-10-09
Windows builds are all failing


Strilanc:
Example: https://travis-ci.com/quantumlib/Cirq/jobs/243569608

```
$ powershell -command 'Set-MpPreference -DisableArchiveScanning $true'
The command "powershell -command 'Set-MpPreference -DisableArchiveScanning $true'" failed and exited with 1 during .
```

c-poole:
These problems seem to be happening because of a flaky interaction with the "Disabling Windows Defender" stage that appears to be present on the Windows Travis builds. Whether or not that stage succeeds seems to be a function of the powershell commands in the before_install section.

dabacon:
I guess we should be able to write this so that if it fails that is OK (since that would just make the test slower)?

If it is flaky we should probably report it https://travis-ci.community/c/environments/windows

dabacon:
Alternatively we could not put the disabling in "before install"?  
https://travis-ci.community/t/windows-builds-time-out-during-cache-restore/4642/5

c-poole:
From the looks of it we don't have to because by default there is a disabling step that attempts to do it for us. The problem is that that disabling step doesn't always work. If it doesn't work, the build seems to pass, but the installs take longer.

c-poole:
Have the matplotlib and powershell changes fixed this issue? I haven't had an errant failure in any of my commits but I'm only one data point.

vtomole:
There is no Window's-based failure on master for past 2 weeks. @Strilanc PR to fix this hasn't been reviewed yet.

PR #2284, opened 2019-10-09
Workaround windows builds suddenly failing

Strilanc:
I'm going to temporarily comment the line in the travis.yml file causing the problem. We can later re-enable it when whatever is causing the breakage is actually fixed. The line relates to turning off antivirus slowdown.

Part of https://github.com/quantumlib/Cirq/issues/2283

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,6 @@ matrix:
     language: sh
     python: '3.7.3'
     before_install:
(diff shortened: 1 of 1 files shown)


vtomole:
Comment out all the power-shell commands?

Strilanc:
I think it just flaked from the thing @c-poole is looking into. It made it past the commands.

PR #2287, opened 2019-10-09
Add cirq.experiments.t1_decay

### LINKED ISSUE #2150: Add WaitOperation, T1 decay experiment, and an example
Resolving this issue requires doing the following parts (not all in the same PR):

1. Introduce a 0-qubit `WaitOperation` with a `cirq.value.Duration` attribute. Modify circuit drawing code to draw it (see `_draw_moment_in_diagram` in `cirq/circuits/circuit.py`). It's not immediately clear to me the best way to define the concept of waiting, e.g. maybe it should be applied to specific qubits; there could be debate around it (e.g. should it allow a symbolic duration?). But a 0-qubit operation seems like a reasonable way.

2. Create a `cirq.t1_decay` method (crib from `cirq.rabi_oscillations`) which runs the circuit `X(a), wait(t), measure(a)` with increasing values of `t`.

3. Add an example to `examples/` that defines a noise model that looks for wait gates and adds proportional decay, and then uses the `t1_decay` to infer the parameter passed into that noise model.

Strilanc:
- Add `cirq.experiments.t1_decay` method
- Add `cirq.experiments.T1DecayResult` type
- Include 'import pandas as pd' in standard repr test setup code

Fixes #2150

Example:

```
import cirq 
class _TimeDependentDecay(cirq.NoiseModel): 
 
    def noisy_moment(self, moment, system_qubits): 
        duration = max((op.gate.duration 
                        for op in moment.operations 
                        if isinstance(op.gate, cirq.WaitGate)), 
                       default=cirq.Duration(nanos=1)) 
        yield cirq.amplitude_damp(1 - 0.99**duration.total_nanos()).on_each( 
            system_qubits) 
        yield moment 
results = cirq.experiments.t1_decay( 
    sampler=cirq.DensityMatrixSimulator(noise=_TimeDependentDecay()), 
    qubit=cirq.GridQubit(0, 0), 
    num_points=50, 
    repetitions=1000, 
    max_delay=cirq.Duration(nanos=1000)) 
results.plot() 
```

image


mpharrigan:
Random comments for WIP PR:

1. I'm immediately confused about `TrialResult` vs `SampleResult`
2. `_TimeDependentDecay` seems like a bad name since t1 decay is invariant to time translations. I get that it depends on *duration* of gates. Seems like we should have a bona fide t1 noise model (called T1NoiseModel) that does what you'd expect

Strilanc:
@mpharrigan `SampleResult` is gone, `_TimeDependentDecay` is an internal testing class so the name is not important.

cirq/experiments/t1_decay_experiment.py
```diff
@@ -0,0 +1,151 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any, TYPE_CHECKING, Optional
+
+import pandas as pd
+import sympy
+from matplotlib import pyplot as plt
+
+from cirq import circuits, devices, ops, study, work, value
```
MichaelBroughton: This and above aren't in alpha order, not a big deal.

cirq/experiments/t1_decay_experiment.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def t1_decay(sampler: work.Sampler,
+             *,
+             qubit: devices.GridQubit,
+             num_points: int,
+             max_delay: 'cirq.DURATION_LIKE',
+             min_delay: 'cirq.DURATION_LIKE' = None,
+             repetitions: int = 1000) -> 'cirq.experiments.T1DecayResult':
+    """Runs a t1 decay experiment.
```
MichaelBroughton: Maybe put a bit more description here

Strilanc: Done.

cirq/experiments/t1_decay_experiment.py
Strilanc: Done.

PR #2290, opened 2019-10-09
Add default parameters for add_subplot

dstrain115:
- cirq only requires matplotlib 3.0
- add_subplot() with no positional arguments was not added until 3.1

See:

https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.figure.Figure.html#matplotlib.figure.Figure.add_subplot
https://matplotlib.org/3.0.3/api/_as_gen/matplotlib.figure.Figure.html#matplotlib.figure.Figure.add_subplot

### DIFF
--- a/cirq/linalg/decompositions.py
+++ b/cirq/linalg/decompositions.py
@@ -610,7 +610,7 @@ def scatter_plot_normalized_kak_interaction_coefficients(
     """
     if ax is None:
         fig = plt.figure()
-        ax = fig.add_subplot(projection='3d')
+        ax = fig.add_subplot(1, 1, 1, projection='3d')
 
     def coord_transform(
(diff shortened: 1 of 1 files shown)


vtomole:
We don't `==` matplotlib though.

https://github.com/quantumlib/Cirq/blob/289307c37b93684ea7d9e2caaec24c543a07e953/requirements.txt#L4

dstrain115:
> We don't `==` matplotlib though.

Yeah, but ' matplotlib~=3.0 ' means that 3.0.3 is a valid version to have, and, anyone with that version will fail this function.

Unless I am not understanding requirements files correctly?

vtomole:
Ah sure.

PR #2291, opened 2019-10-09
Remove usage of deprecated Circuit.from_ops method

dstrain115:
- Circuit.from_ops() is deprecated, so we shouldn't use it in examples
anymore.
- In addition, it can cause test failures if the deprecation warning
gets triggered in another test before being tested for in
circuit_test.py.


dstrain115:
> If i recall, some `from_ops` were kept to make sure the deprecation "still worked" in the sense that `from_ops` would still run?

Shouldn't the circuit_test.py do that?  I don't think we should leave stuff around in random files.  I left the tests in circuit_test that use it.

I think several of these were do to commits that happened after the migration (phased_iswap for instance).

PR #2296, opened 2019-10-09
Require n_rows > 1 in generate_boixo_2018_supremacy_circuits_v2_bristlecone

### LINKED ISSUE #2295: bug in supremacy test circuit
I ran the function
```
generate_boixo_2018_supremacy_circuits_v2_bristlecone(
        n_rows: int, cz_depth: int, seed: int) -> circuits.Circuit
```
in `cirq.experiments.google_v2_supremacy_circuit` with `n_rows = 1` and got trapped in infinite loops.

I think this is because when we have `n_rows = 1`,  `_make_cz_layer` would never return any cz gate, since there's only one qubit, thus in ` _add_cz_layer`, the loop
```
while not cz_layer:
        qubits = cast(Iterable[devices.GridQubit], circuit.all_qubits())
        cz_layer = list(_make_cz_layer(qubits, layer_index))
        layer_index += 1
```
would never end.

My suggestion would be change `assert 1 <= n_rows <= 11` in `generate_boixo_2018_supremacy_circuits_v2_bristlecone` to `assert 2 <= n_rows <= 11`, since it does not make anysense to have one-qubit cz layer in the first place.

tbcdebug:
Fixes #2295

### DIFF
--- a/cirq/experiments/google_v2_supremacy_circuit.py
+++ b/cirq/experiments/google_v2_supremacy_circuit.py
@@ -132,7 +132,7 @@ def generate_boixo_2018_supremacy_circuits_v2_bristlecone(
                 ]
             return len(list(e for e in possibles if e in qubits))
 
-        assert 1 <= n_rows <= 11
+        assert 2 <= n_rows <= 11
         max_row = n_rows - 1
         dev = google.Bristlecone
         # we need a consistent order of qubits


tbcdebug:
@googlebot I signed it!

tbcdebug:
hmm, I wonder why Travis CI build failed. This change seems to be perfectly harmless

vtomole:
@tbcdebug We've flaking on Windows checks lately. Rerunning.

vtomole:
@Strilanc It seems like the other powershell command broke today.

Strilanc:
Normally I'd add a test checking that you caught the failure, but this is fine as is. Thanks for the fix.

PR #2300, opened 2019-10-09
Add cirq.Sampler.sample method

Strilanc:
- Takes combinations of sweeps and returns a pandas.DataFrame for the results


Strilanc:
It would make a lot more sense to just include the 't' as a column in the index, rather than a hash of it.

I did originally put the parameter values into the index. The problem is that when you have multiple index columns I'm not sure how to properly filter by them. When the parameters are data columns you can say this:

`t1s = result[result['t'] == 1]`

or

`t1s = result[result.t == 1]`

And in the multi variable case you do this:

`df[(df.t == 1) & (df.s == 1.5)]`

But with a multi index it doesn't seem to work in a nice way.

I'm not very experienced with pandas, but in my experience it seems to be much easier to work with data when it's in the bulk of the table instead of in the index. So I put the variable values into the bulk.

What sort of iteration task did you have in mind?

MichaelBroughton:
The kind of use case I had in mind that would exist if we added on a multiindex was something like this:

```python
my_params = # some complicated sweep of stuff
my_circuit = # some circuit with lots of params

my_results = sampler.sample(my_circuit, reps=100, my_params)

for certain_values in my_params:
    certain_value_results = my_results[certain_values] 
   # now this contains a dataframe that is indexed by shots and has the same
   # parameter columns and measurement columns without having to worry about
   # other parameter value being in there by accident
```
I see this kind of use as being relatively common, this makes it very intuitive to extract the results for a particular configuration and doesn't really step on the functionality you are talking about, but saves the user having to know to write a long chain of `&&` statements involving the parameter names themselves.

Strilanc:
If there's one known parameter:

```
for p in param_assignments: 
    print(dx[df['x'] == p['x']]) 
```

If there's two known parameters:

```
for p in param_assignments: 
    print(dx[(df['x'] == p['x']) & (df['y'] == p['y'])]) 
```

If there's an unknown number of unknown parameters:

```
for p in param_assignments: 
    print(df[np.all(df[p] == tuple(p.values()), axis=1)]) 
```

After some in person discussion we found problems with the index idea. We'll think about it for a bit.

Strilanc:
Found a variant that works when the independent variables are in the index:

```python
for p in params:
    print(df.loc[tuple(p[k] for k in df.index.names)])                                                                                                                                     
```

MichaelBroughton:
Maybe a good alternative would be just to just index by "experiment number" which is really just the integer number associated with whatever parameter configuration from the sweep you are on

```python
my_params = # some complicated sweep of stuff
my_circuit = # some circuit with lots of params

my_results = sampler.sample(my_circuit, reps=100, my_params)

for i, certain_values in enumerate(my_params):
    certain_value_results = my_results.iloc[i]
    # now this holds the result data for the experiments run wiht params my_params[i]
```
What do you think ?

Strilanc:
I think that concept makes sense when running a simple sweep that you entirely defined yourself, but it quickly loses usefulness for plausible more complicated code. For example, suppose that you decide to randomize the order of an angle sweep in order to avoid drift causing spurious time correlations. The index mapping suddenly becomes very complicated, and potentially inconsistent across software versions. Similar issues come up if you dynamically generate the next parameterization based on previous samples.

If the user is doing a simple sweep, they can already slice out the data they want by index:

```results[reps*i: reps*(i+1)]```

Strilanc:
After offline discussion I think we're in agreement to keep the table minimal.

PR #2301, opened 2019-10-09
Additional workaround for Windows builds failing

Strilanc:
Not sure why these kept working for one day longer than the other line...

Part of #2283


dabacon:
Do we know how to fix this?  This is going to cause much slowness on our windows builds.

Strilanc:
@dabacon I don't know why it's happening in the first place. This was a quick fix to keep things moving. I'm not really sure where to start investigating in the first place.

PR #2303, opened 2019-10-09
Add cirq.WaitGate

Strilanc:
- This is a single-qubit gate that represents the concept of waiting
- Although currently nothing would respect this gate, in the future things like noise models can watch for it


cirq/ops/wait_gate.py
```diff
@@ -0,0 +1,83 @@
...
+    def _trace_distance_bound_(self):
+        return 0
+
+    def __pow__(self, power):
+        if power == 1 or power == -1:
+            # The inverse of a wait is still a wait.
+            return self
+        # Other scalar exponents could scale the wait... but ultimately it is
+        # ambiguous whether the user wanted to scale the duration or just wanted
+        # to affect the unitary. Play it safe and fail.
```
kevinsung: This ambiguity also applies to the case `power == -1`.

Strilanc: True, but I think that's reasonable because abs(-1) = abs(1).

More problematically, we're kinda forced into this because otherwise `cirq.inverse` will fallback to the decompose method and claim the inverse is no operations at all.

cirq/ops/wait_gate.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def _decompose_(self, qubits):
+        return []
+
+    def _trace_distance_bound_(self):
+        return 0
+
+    def __pow__(self, power):
+        if power == 1 or power == -1:
+            # The inverse of a wait is still a wait.
```
kjsatz: Time machine please?

PR #2304, opened 2019-10-09
Write up the recommendation on plotting.

pingyeh:
After talking to @maffoo, it seems we don't really use wikis for docs, but rather write markdown files in the docs/ directory. Here is the recommendation doc for plotting that captures the requirements and recommended practices for the `plot` methods.

I have tested the recommended `plot` method in interactive sessions in colab, ipython and python interpreter. They all work as expected. @Strilanc, can you test it in PyCharm?

### DIFF
--- a/docs/plotting.md
+++ b/docs/plotting.md
@@ -1,0 +1,95 @@
+# Recommended interface and behavior of a `plot` method
+
+Here we recommend the input arguments, return value, and behavior of the
+`plot` method of a class.
+
+## Requirements
+
+1. **Convenience to interactive users.** This is the highest priority.
+    Compared to being called in a batch script as a library (for composing
+    more complicated plots or other purposes), the `plot` method is mainly
+    used in interactive sessions like ipython, jupyther, colab, PyCharm,
+    and python interpreter.
+1. **No unnecessary messages in interactive sessions.** It should not produce
+    any warning/error messages in normal usages in an interactive sessions.
+    See [#1890](https://github.com/quantumlib/Cirq/issues/1890#issue-473510953)
+    for an example of such message.
+1. **No popups during tests.** It should not produce any pop-up windows during
+    tests.
+
+## Recommendation
+
+The `plot` method must produce a plot when there is no arguments in an
+interactive session. The recommended way to achieve that is illustrated in the
+example below.
+
+```python
+class Foo:
+    ...
+    def plot(self, ax: plt.Axes=None, **plot_kwargs: Any) -> plt.Axes:
+        show_plot = not ax
+        if show_plot:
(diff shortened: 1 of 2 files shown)


docs/plotting.md
```diff
@@ -0,0 +1,97 @@
+# Recommended interface and behavior of a `plot` method
```
Strilanc: You should create a `docs/dev` directory and put this inside of it, since this documentation is intended for developers and not users. We can also move `development.md` into it (but be careful to update any references in sphinx configuration).

docs/plotting.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+This `plot` method works in 2 modes: *memory mode* and *interactive mode*,
+signalled by the presence of the `ax` argument. When present, the method is
+instructed to plot on the provided `ax` instance in memory. No plot is shown
+on the screen. When absent, the code is in *interactive* mode, and it creates
+a figure and shows it.
+
+The returned `ax` instance can be used to further customize the plot if the
+user wants to. Note that if we were to call `plt.show` instead of `fig.show`,
```
Strilanc: What happens if we return `None` when `ax` is not set? Pycharm seem to require `plt.show()` instead of `fig.show()`.

pingyeh: Once `plt.show()` is called, the figure and ax can no longer be customized (e.g., add grid, change an axis label, etc).

Per offline conversation, the problem is that it doesn't work in the PyCharm environment. The non-blocking `fig.show` causes the pop-up to not be shown. Is it actually shown but disappears immediately? Is there a setting that can change PyCharm's behavior?

Customizability of plots is not in the current requirements, but I think it is very important. I'd like to add it to the list of requirements.

Strilanc: It's fine if we don't support pycharm and instead nag their devs to make it work.

docs/dev/plotting.md
```diff
@@ -0,0 +1,108 @@
...
+## Recommendation
+
+The `plot` method must produce a plot when there is no arguments in an
+interactive session. The recommended way to achieve that is illustrated in the
+example below.
+
+```python
+class Foo:
+    ...
+    def plot(self, ax: plt.Axes=None, **plot_kwargs: Any) -> plt.Axes:
```
Strilanc: Add an `import matplotlib.pyplot as plt` line. The example testing test is failing because `plt` is not imported at this point.

pingyeh: Done, though the tests are green before the change.

Strilanc: Good catch. That was a bug in the tests. Not it did fail before the move to a subdir of docs. Fixed in https://github.com/quantumlib/Cirq/pull/2307

Issue #2014, opened 2019-08-22
Implement more complete JSON serialization for sympy


mpharrigan:
Follow up to https://github.com/quantumlib/Cirq/pull/1880#pullrequestreview-278642567

It would be ideal if sympy itself had a serialization format. However, the most promising candidate `srepr` looks to be horribly unsafe to parse on untrusted input. 

https://github.com/quantumlib/Cirq/pull/1880 adds support for `sympy.Symbol`, but we'll probably need support for more constructs; at least arithmetic

dabacon:
More support has been added for simple math.  Do we have more we'd like to add?  I'd recommend we close until we see if there is more complicated math that starts showing up that we need to serialize.

mpharrigan:
Sounds good to me!

PR #2305, opened 2019-10-10
Improve pandas and sympy json support

Strilanc:
- Add support for the types needed to make linear combinations in sympy
- Use a stricter pandas equality check in the tests
- Handle pd.Index and pd.MultiIndex recursively
- Fix index type not being preserved and multi-index just not working
- Fix column index name being lost
- Separate testing of external/builtin json round tripping from testing of cirq stuff

Part of #2014


cirq/protocols/json.py
```diff
@@ -204,14 +212,44 @@ def default(self, o):
...
             return obj_to_dict_helper(o, ['name'], namespace='sympy')
 
+        if isinstance(o, (sympy.Add, sympy.Mul, sympy.Pow)):
+            return obj_to_dict_helper(o, ['args'], namespace='sympy')
+
+        if isinstance(o, sympy.Integer):
+            return {'cirq_type': 'sympy.Integer', 'i': o.p}
+
+        if isinstance(o, sympy.Float):
+            return {'cirq_type': 'sympy.Float', 'approx': float(o)}
```
Strilanc: The attributes of sympy float were private, so I didn't dare depend on them.

cirq/protocols/json.py
```diff
@@ -204,14 +212,44 @@ def default(self, o):
...
+            }
+
+        if isinstance(o, pd.MultiIndex):
+            return {
+                'cirq_type': 'pandas.MultiIndex',
+                'tuples': list(o),
+                'names': list(o.names),
+            }
+
+        if isinstance(o, pd.Index):
```
mpharrigan: pandas has pretty robust support for json serialization of data-frames. You might want to investigate if you can just use that

Strilanc: I looked into it. The `to_json` method on `DataFrame` does not include metadata such as the names of the columns in the index.

```
x = pd.DataFrame(index=pd.MultiIndex.from_tuples([(0, 0)], 
                                                      names=['x', 'y']), 
                      data=[[0, 0]], 
                      columns=pd.Index(['a', 'b'], name='c'))
```

```
print(x)
```

```
c    a  b
x y      
0 0  0  0
```

```
print(x.to_json())
```

```
'{"a":{"[0,0]":0},"b":{"[0,0]":0}}'
```

mpharrigan: oh wild

PR #2307, opened 2019-10-10
Fix `snippets_test` not recursing into sub-directorries of docs/

Strilanc:
Also switch from yielding absolute paths to yielding relative paths, to make the generated test names more compact.


docs/snippets_test.py
```diff
@@ -55,9 +55,10 @@
 """
 import inspect
 import sys
-from typing import Any, Dict, List, Pattern, Tuple
+from typing import Any, Dict, List, Pattern, Tuple, Iterator
 
 import os
+from pathlib import Path
```
dabacon: import pathlib?

Strilanc:
And it immediately caught an oversight in the plotting documentation. Success!

docs/snippets_test.py
Strilanc: Done.

PR #2308, opened 2019-10-10
Example of direct fidelity estimation

tonybruguier-google:
Adding an example that illustrates the fidelity estimation, as per these two papers

Direct Fidelity Estimation from Few Pauli Measurements
https://arxiv.org/abs/1104.4695

Practical characterization of quantum devices without tomography
https://arxiv.org/abs/1104.3835


tonybruguier-google:
Happy to withdraw if this is premature.

I've been working with mpharrigan@ but will listen to advice from others, of course.

examples/fidelity.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def build_circuit():
+  qubits = [cirq.LineQubit(i) for i in range(3)]
+
+  rot0 = cirq.ZPowGate(exponent=0.250)  # T-Gate, non Clifford.
+  rot1 = cirq.XPowGate(exponent=0.123)
+  rot2 = cirq.XPowGate(exponent=0.456)
+
+  circuit = cirq.Circuit()
```
Strilanc:
```
qubits = cirq.LineQubit.range(3)
circuit = cirq.Circuit(
    cirq.Z(qubits[0])**0.25,
    cirq.X(qubits[1])**0.123,
    cirq.X(qubits[2])**0.456)
return circuit, qubits
```

examples/fidelity.py
```diff
@@ -0,0 +1,109 @@
...
+  circuit.append(rot0(cirq.LineQubit(0)))
+  circuit.append(rot1(cirq.LineQubit(1)))
+  circuit.append(rot2(cirq.LineQubit(2)))
+
+  return circuit, qubits
+
+
+def build_noisy_circuit(circuit, qubits):
+  noisy_circuit = circuit.copy()
```
Strilanc: `return circuit.with_noise(cirq.amplitude_damp(0.01))`

(You will have to merge master into your branch for this to work.)

examples/fidelity.py
```diff
@@ -0,0 +1,109 @@
...
+  noisy_circuit = build_noisy_circuit(circuit, qubits)
+
+  circuit.append(cirq.measure(*qubits, key='y'))
+  noisy_circuit.append(cirq.measure(*qubits, key='y'))
+
+  n = len(qubits)
+  d = 2**n
+
+  highest_probs = []
+  for i, Pi in enumerate(itertools.product({'I', 'X', 'Y', 'Z'}, repeat=n)):
```
Strilanc: Instead of using the strings IXYZ, you can use `cirq.X`, `cirq.Y`, `cirq.Z`, and `cirq.I`. Up to you.

examples/fidelity.py
```diff
@@ -0,0 +1,109 @@
...
+  noisy_circuit = circuit.copy()
+
+  noisy_gate = cirq.amplitude_damp(0.01)
+  for qubit in qubits:
+    noisy_circuit.append(noisy_gate(qubit))
+
+  return noisy_circuit
+
+
+def simulate_trace(circuit, Pi):
```
Strilanc: Rename `Pi` to `pauli_gates`, and have it be a cirq.X/Y/Z/I list instead of a string.

examples/fidelity.py
```diff
@@ -0,0 +1,109 @@
...
+      rot = numpy.kron(rot, numpy.asarray([[1, 0], [0, -1]], numpy.complex64))
+
+  trace = 0
+  for x in range(d):
+    xbin = numpy.binary_repr(x, width=n)
+
+    xvec = numpy.zeros([d], numpy.complex64)
+    xvec[x] = 1
+
+    initial_state = numpy.matmul(rot, xvec)
```
Strilanc:
```
n = len(pauli_gates)
trace = 0
for x in range(2**n):
    initial_state = cirq.final_wavefunction(
        cirq.DensePauliString(pauli_gates),
        initial_state=x)
    ...
```

examples/fidelity.py
tonybruguier-google: Done

examples/fidelity.py
tonybruguier-google: Done (but directly called in the main function). Also merged and verified that the code still runs.

examples/fidelity.py
tonybruguier-google: Done.

examples/fidelity.py
tonybruguier-google: Done

examples/fidelity.py
tonybruguier-google: This is a great function to have. Thanks for the tip. I did try the code as you suggested, but it failed.

I left the print debug statements and I am quoting the _sorted_ log below. There are two weird things:

- Having an cirq.I gate reduces the size of 'rotated_initial_state'
- It seems the output is order-independent, for example IIZ seems to be the same as IZI.

`
TONYBOOM rotated_initial_state=[0.-0.j 0.+0.j 0.-0.j 0.+0.j 0.-0.j 0.+0.j 0.-0.j 0.+1.j] pauli_gates=[cirq.X, cirq.X, cirq.Y]
TONYBOOM rotated_initial_state=[0.-0.j 0.-0.j 0.+0.j 0.+0.j 0.-0.j 0.-0.j 0.+0.j 0.+1.j] pauli_gates=[cirq.X, cirq.Y, cirq.X]
TONYBOOM rotated_initial_state=[0.-0.j 0.-0.j 0.-0.j 0.-0.j 0.+0.j 0.+0.j 0.+0.j 0.+1.j] pauli_gates=[cirq.Y, cirq.X, cirq.X]
TONYBOOM rotated_initial_state=[ 0.+0.j -0.+0.j  0.-0.j  0.+0.j -0.+0.j  0.+0.j  0.-0.j -0.-1.j] pauli_gates=[cirq.Y, cirq.Y, cirq.Y]
TONYBOOM rotated_initial_state=[0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 1.+0.j] pauli_gates=[cirq.X, cirq.X, cirq.X]
TONYBOOM rotated_initial_state=[-0.+0.j  0.+0.j  0.-0.j  0.+0.j -0.+0.j  0.+0.j  0.-0.j -1.+0.j] pauli_gates=[cirq.X, cirq.Y, cirq.Y]
TONYBOOM rotated_initial_state=[-0.+0.j  0.+0.j -0.+0.j  0.+0.j  0.-0.j  0.+0.j  0.-0.j -1.+0.j] pauli_gates=[cirq.Y, cirq.X, cirq.Y]
TONYBOOM rotated_initial_state=[-0.+0.j -0.+0.j  0.+0.j  0.+0.j  0.-0.j  0.-0.j  0.+0.j -1.+0.j] pauli_gates=[cirq.Y, cirq.Y, cirq.X]
TONYBOOM rotated_initial_state=[ 0.-0.j  0.+0.j  0.+0.j -0.+0.j  0.-0.j  0.+0.j  0.+1.j -0.+0.j] pauli_gates=[cirq.X, cirq.Y, cirq.Z]
TONYBOOM rotated_initial_state=[ 0.-0.j  0.+0.j  0.-0.j  0.+0.j  0.+0.j -0.+0.j  0.+1.j -0.+0.j] pauli_gates=[cirq.Y, cirq.X, cirq.Z]
TONYBOOM rotated_initial_state=[ 0.+0.j -0.+0.j  0.+0.j -0.+0.j  0.+0.j -0.+0.j  1.+0.j -0.+0.j] pauli_gates=[cirq.X, cirq.X, cirq.Z]
TONYBOOM rotated_initial_state=[-0.+0.j  0.-0.j  0.+0.j -0.+0.j  0.-0.j  0.+0.j -1.+0.j -0.+0.j] pauli_gates=[cirq.Y, cirq.Y, cirq.Z]
TONYBOOM rotated_initial_state=[ 0.-0.j  0.+0.j  0.+0.j -0.+0.j  0.-0.j  0.+1.j  0.+0.j -0.+0.j] pauli_gates=[cirq.X, cirq.Z, cirq.Y]
TONYBOOM rotated_initial_state=[ 0.-0.j  0.-0.j  0.+0.j  0.+0.j  0.+0.j  0.+1.j -0.+0.j -0.+0.j] pauli_gates=[cirq.Y, cirq.Z, cirq.X]
TONYBOOM rotated_initial_state=[ 0.+0.j  0.+0.j -0.+0.j -0.+0.j  0.+0.j  1.+0.j -0.+0.j -0.+0.j] pauli_gates=[cirq.X, cirq.Z, cirq.X]
TONYBOOM rotated_initial_state=[-0.+0.j  0.+0.j  0.-0.j  0.+0.j  0.-0.j -1.+0.j  0.+0.j -0.+0.j] pauli_gates=[cirq.Y, cirq.Z, cirq.Y]
TONYBOOM rotated_initial_state=[ 0.-0.j  0.+0.j  0.+0.j -0.+0.j  0.+1.j -0.+0.j -0.+0.j  0.-0.j] pauli_gates=[cirq.Y, cirq.Z, cirq.Z]
TONYBOOM rotated_initial_state=[ 0.+0.j -0.+0.j -0.+0.j  0.-0.j  1.+0.j -0.+0.j -0.+0.j  0.-0.j] pauli_gates=[cirq.X, cirq.Z, cirq.Z]
TONYBOOM rotated_initial_state=[ 0.-0.j  0.+0.j  0.-0.j  0.+1.j  0.+0.j -0.+0.j  0.+0.j -0.+0.j] pauli_gates=[cirq.Z, cirq.X, cirq.Y]
TONYBOOM rotated_initial_state=[ 0.-0.j  0.-0.j  0.+0.j  0.+1.j  0.+0.j  0.+0.j -0.+0.j -0.+0.j] pauli_gates=[cirq.Z, cirq.Y, cirq.X]
TONYBOOM rotated_initial_state=[0.-0.j 0.+0.j 0.-0.j 0.+1.j] pauli_gates=[cirq.I, cirq.X, cirq.Y]
TONYBOOM rotated_initial_state=[0.-0.j 0.-0.j 0.+0.j 0.+1.j] pauli_gates=[cirq.I, cirq.Y, cirq.X]
TONYBOOM rotated_initial_state=[0.-0.j 0.+0.j 0.-0.j 0.+1.j] pauli_gates=[cirq.X, cirq.I, cirq.Y]
TONYBOOM rotated_initial_state=[0.-0.j 0.+0.j 0.-0.j 0.+1.j] pauli_gates=[cirq.X, cirq.Y, cirq.I]
TONYBOOM rotated_initial_state=[0.-0.j 0.-0.j 0.+0.j 0.+1.j] pauli_gates=[cirq.Y, cirq.I, cirq.X]
TONYBOOM rotated_initial_state=[0.-0.j 0.-0.j 0.+0.j 0.+1.j] pauli_gates=[cirq.Y, cirq.X, cirq.I]
TONYBOOM rotated_initial_state=[ 0.+0.j  0.+0.j  0.+0.j  1.+0.j -0.+0.j -0.+0.j -0.+0.j -0.+0.j] pauli_gates=[cirq.Z, cirq.X, cirq.X]
TONYBOOM rotated_initial_state=[-0.+0.j  0.+0.j  0.-0.j -1.+0.j  0.-0.j  0.+0.j  0.+0.j -0.+0.j] pauli_gates=[cirq.Z, cirq.Y, cirq.Y]
TONYBOOM rotated_initial_state=[0.+0.j 0.+0.j 0.+0.j 1.+0.j] pauli_gates=[cirq.I, cirq.X, cirq.X]
TONYBOOM rotated_initial_state=[-0.+0.j  0.+0.j  0.-0.j -1.+0.j] pauli_gates=[cirq.I, cirq.Y, cirq.Y]
TONYBOOM rotated_initial_state=[0.+0.j 0.+0.j 0.+0.j 1.+0.j] pauli_gates=[cirq.X, cirq.I, cirq.X]
TONYBOOM rotated_initial_state=[0.+0.j 0.+0.j 0.+0.j 1.+0.j] pauli_gates=[cirq.X, cirq.X, cirq.I]
TONYBOOM rotated_initial_state=[-0.+0.j  0.+0.j  0.-0.j -1.+0.j] pauli_gates=[cirq.Y, cirq.I, cirq.Y]
TONYBOOM rotated_initial_state=[-0.+0.j  0.+0.j  0.-0.j -1.+0.j] pauli_gates=[cirq.Y, cirq.Y, cirq.I]
TONYBOOM rotated_initial_state=[ 0.-0.j  0.+0.j  0.+1.j -0.+0.j  0.+0.j -0.+0.j -0.+0.j  0.-0.j] pauli_gates=[cirq.Z, cirq.Y, cirq.Z]
TONYBOOM rotated_initial_state=[ 0.-0.j  0.+0.j  0.+1.j -0.+0.j] pauli_gates=[cirq.I, cirq.Y, cirq.Z]
TONYBOOM rotated_initial_state=[ 0.-0.j  0.+0.j  0.+1.j -0.+0.j] pauli_gates=[cirq.Y, cirq.I, cirq.Z]
TONYBOOM rotated_initial_state=[ 0.-0.j  0.+0.j  0.+1.j -0.+0.j] pauli_gates=[cirq.Y, cirq.Z, cirq.I]
TONYBOOM rotated_initial_state=[ 0.+0.j -0.+0.j  1.+0.j -0.+0.j -0.+0.j  0.-0.j -0.+0.j  0.-0.j] pauli_gates=[cirq.Z, cirq.X, cirq.Z]
TONYBOOM rotated_initial_state=[ 0.+0.j -0.+0.j  1.+0.j -0.+0.j] pauli_gates=[cirq.I, cirq.X, cirq.Z]
TONYBOOM rotated_initial_state=[ 0.+0.j -0.+0.j  1.+0.j -0.+0.j] pauli_gates=[cirq.X, cirq.I, cirq.Z]
TONYBOOM rotated_initial_state=[ 0.+0.j -0.+0.j  1.+0.j -0.+0.j] pauli_gates=[cirq.X, cirq.Z, cirq.I]
TONYBOOM rotated_initial_state=[ 0.-0.j  0.+1.j  0.+0.j -0.+0.j  0.+0.j -0.+0.j -0.+0.j  0.+0.j] pauli_gates=[cirq.Z, cirq.Z, cirq.Y]
TONYBOOM rotated_initial_state=[ 0.-0.j  0.+1.j  0.+0.j -0.+0.j] pauli_gates=[cirq.I, cirq.Z, cirq.Y]
TONYBOOM rotated_initial_state=[ 0.-0.j  0.+1.j  0.+0.j -0.+0.j] pauli_gates=[cirq.Z, cirq.I, cirq.Y]
TONYBOOM rotated_initial_state=[ 0.-0.j  0.+1.j  0.+0.j -0.+0.j] pauli_gates=[cirq.Z, cirq.Y, cirq.I]
TONYBOOM rotated_initial_state=[0.-0.j 0.+1.j] pauli_gates=[cirq.I, cirq.I, cirq.Y]
TONYBOOM rotated_initial_state=[0.-0.j 0.+1.j] pauli_gates=[cirq.I, cirq.Y, cirq.I]
TONYBOOM rotated_initial_state=[0.-0.j 0.+1.j] pauli_gates=[cirq.Y, cirq.I, cirq.I]
TONYBOOM rotated_initial_state=[ 0.+0.j  1.+0.j -0.+0.j -0.+0.j -0.+0.j -0.+0.j  0.-0.j  0.-0.j] pauli_gates=[cirq.Z, cirq.Z, cirq.X]
TONYBOOM rotated_initial_state=[ 0.+0.j  1.+0.j -0.+0.j -0.+0.j] pauli_gates=[cirq.I, cirq.Z, cirq.X]
TONYBOOM rotated_initial_state=[ 0.+0.j  1.+0.j -0.+0.j -0.+0.j] pauli_gates=[cirq.Z, cirq.I, cirq.X]
TONYBOOM rotated_initial_state=[ 0.+0.j  1.+0.j -0.+0.j -0.+0.j] pauli_gates=[cirq.Z, cirq.X, cirq.I]
TONYBOOM rotated_initial_state=[0.+0.j 1.+0.j] pauli_gates=[cirq.I, cirq.I, cirq.X]
TONYBOOM rotated_initial_state=[0.+0.j 1.+0.j] pauli_gates=[cirq.I, cirq.X, cirq.I]
TONYBOOM rotated_initial_state=[0.+0.j 1.+0.j] pauli_gates=[cirq.X, cirq.I, cirq.I]
TONYBOOM rotated_initial_state=[ 1.+0.j -0.+0.j -0.+0.j  0.-0.j -0.+0.j  0.-0.j  0.-0.j  0.+0.j] pauli_gates=[cirq.Z, cirq.Z, cirq.Z]
TONYBOOM rotated_initial_state=[ 1.+0.j -0.+0.j -0.+0.j  0.-0.j] pauli_gates=[cirq.I, cirq.Z, cirq.Z]
TONYBOOM rotated_initial_state=[ 1.+0.j -0.+0.j -0.+0.j  0.-0.j] pauli_gates=[cirq.Z, cirq.I, cirq.Z]
TONYBOOM rotated_initial_state=[ 1.+0.j -0.+0.j -0.+0.j  0.-0.j] pauli_gates=[cirq.Z, cirq.Z, cirq.I]
TONYBOOM rotated_initial_state=[ 1.+0.j -0.+0.j] pauli_gates=[cirq.I, cirq.I, cirq.Z]
TONYBOOM rotated_initial_state=[ 1.+0.j -0.+0.j] pauli_gates=[cirq.I, cirq.Z, cirq.I]
TONYBOOM rotated_initial_state=[ 1.+0.j -0.+0.j] pauli_gates=[cirq.Z, cirq.I, cirq.I]
TONYBOOM rotated_initial_state=[1.+0.j] pauli_gates=[cirq.I, cirq.I, cirq.I]
`

tonybruguier-google: I did look at the code for parameters, or a bu maybe, but I can't find anything obvious.

Leaving as not resolved.

tonybruguier-google:
I took another look this morning, and I think I can resolve the last comment now.

examples/fidelity.py
```diff
@@ -0,0 +1,82 @@
...
+  trace = 0
+  for x in range(d):
+    xbin = numpy.binary_repr(x, width=n)
+
+    rotated_initial_state = cirq.final_wavefunction(
+        cirq.DensePauliString(pauli_gates),
+        qubit_order=qubits,
+        initial_state=x)
+
+    y = simulator.simulate(circuit, initial_state=rotated_initial_state).measurements['y']
```
Strilanc: If you only need the measurement results, i.e. samples, you can use `cirq.sample` instead of asking for the full final wavefunction.

examples/fidelity.py
```diff
@@ -0,0 +1,82 @@
...
+    Pr_i = rho_i * rho_i / d
+
+    if Pr_i > 0:
+      heapq.heappush(highest_probs, (Pr_i, rho_i, i, pauli_gates))
+    if len(highest_probs) > n:
+      heapq.heappop(highest_probs)
+
+  fidelity = 0.0
+  for prob_tuple in highest_probs:
+    Pr_i = prob_tuple[0]
```
Strilanc: Use `x, y, _, z = ` notation.

examples/fidelity.py
```diff
@@ -0,0 +1,82 @@
...
+  d = 2**n
+
+  highest_probs = []
+  for i, pauli_gates in enumerate(itertools.product({cirq.I, cirq.X, cirq.Y, cirq.Z}, repeat=n)):
+    rho_i = simulate_trace(circuit, pauli_gates, qubits)
+
+    Pr_i = rho_i * rho_i / d
+
+    if Pr_i > 0:
+      heapq.heappush(highest_probs, (Pr_i, rho_i, i, pauli_gates))
```
Strilanc: Why are you storing `i` if it's not used later?

examples/fidelity.py
```diff
@@ -0,0 +1,82 @@
...
+    if len(highest_probs) > n:
+      heapq.heappop(highest_probs)
```
Strilanc: This is writing out the details of `heapq.nlargest` or `heapq.nsmallest`, right? We could use those methods directly.

```
pauli_traces = [
     pauli_gates, simulate_trace(circuit, pauli_gates, qubits)
     for pauli_gates in itertools.product([cirq.I, cirq.X, cirq.Y, cirq.Z], repeat=n)
]
highest_probs = heapq.nlargest(n, pauli_traces, key=lambda e: e[1])
```

examples/fidelity.py
```diff
@@ -0,0 +1,82 @@
...
+
+    y = simulator.simulate(circuit, initial_state=rotated_initial_state).measurements['y']
+    trace += sum([int(xbin[i]) == y[i] for i in range(n)])
+
+  return trace
+
+
+def main():
+  circuit, qubits = build_circuit()
+  noisy_circuit = circuit.with_noise(cirq.amplitude_damp(0.01))
```
Strilanc: Move this down a line so that the measurement appending is not done twice.

examples/fidelity.py
```diff
@@ -0,0 +1,82 @@
...
+  qubits = cirq.LineQubit.range(3)
+  circuit = cirq.Circuit(
+      cirq.Z(qubits[0])**0.25,  # T-Gate, non Clifford.
+      cirq.X(qubits[1])**0.123,
+      cirq.X(qubits[2])**0.456)
+  return circuit, qubits
+
+
+def simulate_trace(circuit, pauli_gates, qubits):
+  simulator = cirq.DensityMatrixSimulator()
```
Strilanc: The simulator can take a noise model, if you don't want to make two different circuits.

examples/fidelity.py
```diff
@@ -0,0 +1,82 @@
...
+import itertools
+import math
+import numpy
+
+
+def build_circuit():
+  qubits = cirq.LineQubit.range(3)
+  circuit = cirq.Circuit(
+      cirq.Z(qubits[0])**0.25,  # T-Gate, non Clifford.
+      cirq.X(qubits[1])**0.123,
```
Strilanc: Could you discuss where these constants come from?

examples/fidelity.py
tonybruguier-google: Done.

examples/fidelity.py
tonybruguier-google: Yes, I should have put a comment since it's not obvious. Without i, in the case when we have the same probs, then the comparison in the heap resorts to trying to compare pauli_gates. But since the objects are not comparable, it was throwing an error message.

I added the comment.

examples/fidelity.py
tonybruguier-google: Oh, nice. Done.

examples/fidelity.py
tonybruguier-google: I think it's moot now, since I followed your advice (I think) below to set the noise in the simulator.

examples/fidelity.py
tonybruguier-google: It is arbitrary. I just needed some circuit not too big. Is this OK with the new comment?

examples/fidelity.py
tonybruguier-google: Nice. Done.

examples/fidelity.py
tonybruguier-google: Actually, given your other suggestion, this is moot now. Done.

examples/fidelity.py
tonybruguier-google: Like this?

It seems that the function doesn't have a way to specify the initial state, so instead I pre-pended the rotations to the circuit.

Also left the old way commented out, just in case. I can remove I indeed did what you had in mind.

Leaving unresolved for now.

tonybruguier-google:
Thanks. PTAL.

examples/fidelity.py
```diff
@@ -0,0 +1,86 @@
...
+
+  pauli_traces = [
+      simulate_trace(circuit, pauli_gates, qubits, noise=None) + (pauli_gates,)
+      for pauli_gates in itertools.product([cirq.I, cirq.X, cirq.Y, cirq.Z],
+                                           repeat=n)
+  ]
+  highest_probs = heapq.nlargest(n, pauli_traces, key=lambda e: e[1])
+
+  fidelity = 0.0
+  for prob_tuple in highest_probs:
```
mpharrigan: nit: you can expand the tuple here

```python
for rho_i, Pr_i, pauli_gates in highest_probs:
```

examples/fidelity.py
```diff
@@ -0,0 +1,86 @@
...
+
+  n = len(pauli_gates)
+  d = 2**n
+
+  trace = 0
+  for x in range(d):
+    xbin = numpy.binary_repr(x, width=n)
+
+    pauli_string = cirq.PauliString(dict(zip(qubits, pauli_gates)))
+    display = cirq.approx_pauli_string_expectation(pauli_string, num_samples=n)
```
mpharrigan: why is num_samples set to the number of qubits?

examples/fidelity.py
```diff
@@ -0,0 +1,86 @@
+"""Simulates the fidelity estimation.
```
mpharrigan: "fidelity estimation" is a very generic thing one might do. Conventionally it's done by tomography

Specify "direct fidelity estimation", which is a specific thing given by the referenced papers. The filename should also be more specific

examples/fidelity.py
```diff
@@ -0,0 +1,86 @@
...
+  for x in range(d):
+    xbin = numpy.binary_repr(x, width=n)
+
+    pauli_string = cirq.PauliString(dict(zip(qubits, pauli_gates)))
+    display = cirq.approx_pauli_string_expectation(pauli_string, num_samples=n)
+    y = cirq.sample(cirq.Circuit(
+        pauli_string,
+        circuit)).measurements['y'][0]
+
+    # rotated_initial_state = cirq.final_wavefunction(
```
mpharrigan: remove commented out code

examples/fidelity.py
```diff
@@ -0,0 +1,86 @@
...
+  noise = cirq.ConstantQubitNoiseModel(cirq.depolarize(0.1))
+
+  n = len(qubits)
+
+  pauli_traces = [
+      simulate_trace(circuit, pauli_gates, qubits, noise=None) + (pauli_gates,)
+      for pauli_gates in itertools.product([cirq.I, cirq.X, cirq.Y, cirq.Z],
+                                           repeat=n)
+  ]
+  highest_probs = heapq.nlargest(n, pauli_traces, key=lambda e: e[1])
```
mpharrigan: why are you taking only the n highest probability ones?

examples/fidelity.py
```diff
@@ -0,0 +1,86 @@
...
+  # use of simulators.
+  qubits = cirq.LineQubit.range(3)
+  circuit = cirq.Circuit(
+      cirq.Z(qubits[0])**0.25,  # T-Gate, non Clifford.
+      cirq.X(qubits[1])**0.123,
+      cirq.X(qubits[2])**0.456)
+  return circuit, qubits
+
+
+def simulate_trace(circuit, pauli_gates, qubits, noise):
```
mpharrigan: The flow of this example could be greatly assisted by following the nomenclature of one of the papers and including line comments describing which equation number each quantity corresponds to. For example: I think this function is calculating the "characteristic function" chi from https://arxiv.org/abs/1104.4695 where `circuit` specifies the construction of rho or sigma and `pauli_gates` corresponds to W_k

examples/fidelity.py
```diff
@@ -0,0 +1,86 @@
...
+  d = 2**n
+
+  trace = 0
+  for x in range(d):
+    xbin = numpy.binary_repr(x, width=n)
+
+    pauli_string = cirq.PauliString(dict(zip(qubits, pauli_gates)))
+    display = cirq.approx_pauli_string_expectation(pauli_string, num_samples=n)
+    y = cirq.sample(cirq.Circuit(
+        pauli_string,
```
mpharrigan: why are you adding pauli_string to a circuit?

examples/fidelity.py
tonybruguier-google: Done.

examples/fidelity.py
tonybruguier-google: Done

examples/fidelity.py
tonybruguier-google: I had misread first paragraph of second column of second page of the paper: https://arxiv.org/pdf/1104.3835.pdf

The repetition is *not* within a rotation, so set to 1.

examples/fidelity.py
tonybruguier-google: Renamed some variables and added some comments. It may be that the two papers propose slightly different algo, so I am not sure which to follow... or I may just be plain confused.

examples/fidelity.py
tonybruguier-google: I couldn't find any way to input the initial state in cirq.sample() and thus did the pre-rotation. I remembered that you had told me not to do it.

In some previous iteration, reviewer Craig Gidney (Strilanc) asked me not to use the DenseMatrixSimulator which allowed to input a different initial state, and thus not do the pre-rotation.

I had left that comment unresolved and I will leave this one unresolved too, since I suspect you are both in agreement and I just don't understand what both of you are telling me.

examples/direct_fidelity_estimation.py
```diff
@@ -0,0 +1,82 @@
...
+      cirq.X(qubits[1])**0.123,
+      cirq.X(qubits[2])**0.456)
+  return circuit, qubits
+
+
+def compute_characteristic_function(circuit, P_i, qubits, noise):
+  n = len(P_i)
+  d = 2**n
+
+  trace = 0  # rho_i / sigma_i in https://arxiv.org/pdf/1104.3835.pdf
```
mpharrigan: does this comment mean "rho_i *or* sigma_i" or "rho_i divided by sigma_i". I think the former(?) but both expressions are present in the linked paper!

examples/direct_fidelity_estimation.py
```diff
@@ -0,0 +1,82 @@
...
+  for x in range(d):
+    xbin = numpy.binary_repr(x, width=n)
+
+    pauli_string = cirq.PauliString(dict(zip(qubits, P_i)))
+    display = cirq.approx_pauli_string_expectation(pauli_string, num_samples=1)
```
mpharrigan: I think this function has been removed from cirq `master`. See my comment on this function about how you might consider doing it instead

examples/direct_fidelity_estimation.py
```diff
@@ -0,0 +1,82 @@
...
+  # use of simulators.
+  qubits = cirq.LineQubit.range(3)
+  circuit = cirq.Circuit(
+      cirq.Z(qubits[0])**0.25,  # T-Gate, non Clifford.
+      cirq.X(qubits[1])**0.123,
+      cirq.X(qubits[2])**0.456)
+  return circuit, qubits
+
+
+def compute_characteristic_function(circuit, P_i, qubits, noise):
```
mpharrigan: Thanks for adding the comments so I know what you're trying to compute :)

In the linked paper, rho_i and sigma_i refer to Tr[rho P_i] or Tr[sigma P_i], resp. where rho and/or sigma are density matrices and the P_i are Hermitian operators (in this case, `PauliString`s) 

The secret is to realize that Tr[rho P_i] is the expectation value of P_i given the state rho. There is a function that does this for you in cirq! `PauliString.expectation_value_from_density_matrix`. 

Of course, to certify a state from the actual quantum computer we don't have access to the density matrix so we can talk about how to slot in a sampling-based estimate for operator estimation values to be implemented in a follow-up PR

examples/direct_fidelity_estimation.py
```diff
@@ -0,0 +1,82 @@
...
+  circuit.append(cirq.measure(*qubits, key='y'))
+
+  noise = cirq.ConstantQubitNoiseModel(cirq.depolarize(0.1))
+
+  n = len(qubits)
+
+  # Computes for every \hat{P_i} of https://arxiv.org/pdf/1104.3835.pdf,
+  # estimate rho_i and Pr(i). We then collect tuples (rho_i, Pr(i), \hat{Pi})
+  # inside the variable 'pauli_traces'.
+  pauli_traces = [
```
mpharrigan: I'd use a dictionary, named tuple, or dataclass to document the components of each 3-tuple in this list

examples/direct_fidelity_estimation.py
```diff
@@ -0,0 +1,82 @@
...
+      compute_characteristic_function(circuit, P_i, qubits, noise=None) + (P_i,)
+      for P_i in itertools.product([cirq.I, cirq.X, cirq.Y, cirq.Z],
+                                           repeat=n)
+  ]
+  highest_probs = heapq.nlargest(n, pauli_traces, key=lambda e: e[1])
```
mpharrigan: still confused why you're doing this

examples/direct_fidelity_estimation.py
tonybruguier-google: Sorry, it was indeed a confusing way to write things. It is 'or'.

examples/direct_fidelity_estimation.py

examples/direct_fidelity_estimation.py
tonybruguier-google: Actually, I should have sampled according to prob. Done now.

tonybruguier-google:
Thanks. PTAL.

examples/direct_fidelity_estimation.py
```diff
@@ -29,17 +29,16 @@ def compute_characteristic_function(circuit, P_i, qubits, noise):
...
-    display = cirq.approx_pauli_string_expectation(pauli_string, num_samples=1)
-    y = cirq.sample(cirq.Circuit(
-        pauli_string,
-        circuit)).measurements['y'][0]
-
-    trace += sum([int(xbin[i]) == y[i] for i in range(n)])
+  pauli_string = cirq.PauliString(dict(zip(qubits, P_i)))
+  qubit_map = dict(zip(qubits, range(n)))
+  # rho_i or sigma_i in https://arxiv.org/pdf/1104.3835.pdf
+  trace = pauli_string.expectation_from_density_matrix(density_matrix,
```
mpharrigan: nit: put in a check that the imaginary component is below some tolerance

examples/direct_fidelity_estimation.py
```diff
@@ -29,17 +29,16 @@ def compute_characteristic_function(circuit, P_i, qubits, noise):
   n = len(P_i)
   d = 2**n
 
-  trace = 0  # rho_i / sigma_i in https://arxiv.org/pdf/1104.3835.pdf
-  for x in range(d):
-    xbin = numpy.binary_repr(x, width=n)
+  simulator = cirq.DensityMatrixSimulator()
+  density_matrix = 0  #  rho or sigma in https://arxiv.org/pdf/1104.3835.pdf
+  for step in simulator.simulate_moment_steps(circuit):
+    density_matrix = step.density_matrix()
```
mpharrigan: You can do this with `simulator.simulate` without the for loop

examples/direct_fidelity_estimation.py
```diff
@@ -29,17 +29,16 @@ def compute_characteristic_function(circuit, P_i, qubits, noise):
...
 
-    pauli_string = cirq.PauliString(dict(zip(qubits, P_i)))
-    display = cirq.approx_pauli_string_expectation(pauli_string, num_samples=1)
-    y = cirq.sample(cirq.Circuit(
-        pauli_string,
-        circuit)).measurements['y'][0]
-
-    trace += sum([int(xbin[i]) == y[i] for i in range(n)])
+  pauli_string = cirq.PauliString(dict(zip(qubits, P_i)))
```
mpharrigan: add type annotations to this function's signature. They help a lot. In this case, when reading the code I'm wondering "what is `P_i` and why isn't it a `PauliString` already"

examples/direct_fidelity_estimation.py
```diff
@@ -57,25 +56,31 @@ def main():
...
+  pauli_traces = []
+  for P_i in itertools.product([cirq.I, cirq.X, cirq.Y, cirq.Z], repeat=n):
+    rho_i, Pr_i = compute_characteristic_function(
+        circuit, P_i, qubits, noise=None)
+    pauli_traces.append({'P_i': P_i, 'rho_i': rho_i, 'Pr_i': Pr_i})
+
+  assert len(pauli_traces) == 4**n
+
+  p = [x['Pr_i'] for x in pauli_traces]
+  assert numpy.isclose(sum(p), 1.0, 1e-6)
```
mpharrigan: nit: use a kwarg to document what 1e-6 is. Namely: `atol` or `rtol`. I legitimately don't know which one it actually is since I don't remember the function signature order :)

examples/direct_fidelity_estimation.py
```diff
@@ -57,25 +56,31 @@ def main():
...
 
   fidelity = 0.0
-  N = 0
-  for rho_i, Pr_i, P_i in highest_probs:
-    if Pr_i == 0:
-      break
+  for iter in range(n):
```
mpharrigan: don't use the variable name `iter`. It shadows a python builtin

tonybruguier-google: Good point. Changed to _.

examples/direct_fidelity_estimation.py
tonybruguier-google: It was 'rtol' but I think actually that 'atol' makes more sense, and so I used this.

examples/direct_fidelity_estimation.py

examples/direct_fidelity_estimation.py

examples/direct_fidelity_estimation.py

examples/fidelity.py
tonybruguier-google: Ack

tonybruguier-google:
Thanks. PTAL.

examples/direct_fidelity_estimation.py
```diff
@@ -0,0 +1,93 @@
...
+        circuit, P_i, qubits, noise=None)
+    pauli_traces.append({'P_i': P_i, 'rho_i': rho_i, 'Pr_i': Pr_i})
+
+  assert len(pauli_traces) == 4**n
+
+  p = [x['Pr_i'] for x in pauli_traces]
+  assert numpy.isclose(sum(p), 1.0, atol=1e-6)
+
+  fidelity = 0.0
+  for _ in range(n):
```
mpharrigan: Why are you doing this `n` times, where `n` is the number of qubits? I think this should be a separate parameter. Consider renaming `n` to `n_qubits` to avoid ambiguity between all the numbers

examples/direct_fidelity_estimation.py
```diff
@@ -0,0 +1,93 @@
...
+
+Direct Fidelity Estimation from Few Pauli Measurements
+https://arxiv.org/abs/1104.4695
+
+Practical characterization of quantum devices without tomography
+https://arxiv.org/abs/1104.3835
+"""
+
+import cirq
+import heapq
```
mpharrigan: redundant

examples/direct_fidelity_estimation.py
```diff
@@ -0,0 +1,93 @@
...
+import itertools
+import math
+import numpy
```
mpharrigan: optional: `import numpy as np`

examples/direct_fidelity_estimation.py
```diff
@@ -0,0 +1,93 @@
...
+                                                       qubit_map)
+  assert numpy.isclose(trace.imag, 0.0, atol=1e-6)
+  trace = trace.real
+
+  prob = trace * trace / d  # Pr(i) in https://arxiv.org/pdf/1104.3835.pdf
+
+  return trace, prob
+
+
+def main():
```
mpharrigan: We use four spaces for indent

examples/direct_fidelity_estimation.py

examples/direct_fidelity_estimation.py

tonybruguier-google:
Thanks. PTAL.

mpharrigan:
I think you need to push your changes

examples/direct_fidelity_estimation.py
```diff
@@ -0,0 +1,96 @@
...
+            circuit, P_i, qubits, noise=None)
+        pauli_traces.append({'P_i': P_i, 'rho_i': rho_i, 'Pr_i': Pr_i})
+
+    assert len(pauli_traces) == 4**n
+
+    p = [x['Pr_i'] for x in pauli_traces]
+    assert np.isclose(sum(p), 1.0, atol=1e-6)
+
+    fidelity = 0.0
+    for _ in range(n):
```
mpharrigan: Need to change this from `n` to `N`. I see from your helpful comments that you're following the paper variable names. This is one case where I'd advocate for more programmer-friendly names like `n_qubits` and `n_trials`

examples/direct_fidelity_estimation.py
```diff
@@ -0,0 +1,96 @@
...
+        Pr_i = pauli_traces[i]['Pr_i']
+        P_i = pauli_traces[i]['P_i']
+        rho_i = pauli_traces[i]['rho_i']
+
+        sigma_i, _ = compute_characteristic_function(circuit, P_i, qubits,
+                                                     noise)
+
+        fidelity += Pr_i * sigma_i / rho_i
+
+    print(fidelity / n)
```
mpharrigan: wrong n

mpharrigan:
Cool. now it will just be a matter of getting the CI checks to go green

tonybruguier-google:
Hi Matt,
I'm having trouble with this test:
https://travis-ci.com/quantumlib/Cirq/jobs/249375170

it does indeed seem to be missing from here:
https://github.com/quantumlib/Cirq/blob/1000c9a1995b068ea1890bf5556dff4f9acf4e93/cirq/sim/simulator.py#L503

But it is defined here:
https://github.com/quantumlib/Cirq/blob/1000c9a1995b068ea1890bf5556dff4f9acf4e93/cirq/sim/density_matrix_simulator.py#L500

Do you have advice on how to fix this?

mpharrigan:
You can correct the type with
```python
    density_matrix: cirq.DensityMatrixTrialResult = simulator.simulate(circuit).final_density_matrix
```

You can run the mypy check locally by running `check/mypy`

mpharrigan:
since `sim.simulate` is an abstract method it can't tell that you're running it from `cirq.DensityMatrixSimulator()` so you have to tell the type system explicitly

tonybruguier-google:
Hi Matt,

I tried your suggestion:
`density_matrix: cirq.DensityMatrixTrialResult = simulator.simulate(circuit).final_density_matrix`

But I got a Travis error:
https://travis-ci.com/quantumlib/Cirq/jobs/249708895

I also tried this:
`    trial_results: cirq.DensityMatrixTrialResult = simulator.simulate(circuit)
    density_matrix = trial_results.final_density_matrix`

But I got this error:
`examples/direct_fidelity_estimation.py:38: error: Incompatible types in assignment (expression has type "SimulationTrialResult", variable has type "DensityMatrixTrialResult")
examples/direct_fidelity_estimation.py:39: error: "SimulationTrialResult" has no attribute "final_density_matrix`

So I have the new proposed change. However, it touches another file, so I am not sure it's wise.

mpharrigan:
Opened https://github.com/quantumlib/Cirq/issues/2413

I think the new change is not the right way to go. We shouldn't compromise the actual executable code to kowtow to mypy

Sorry for the previous suggestion not working. I just remembered that the way to force a type is not with an annotation but rather by using `cast` from the `typing` module

```python
cast(DensityMatrixTrialResult, simulator.simulate(circuit)).final_density_matrix
```

*should* do the trick

tonybruguier-google:
All right. I think all the checks are passing now. If I understand correctly, you are the one with write access who can merge changes (if that PR is indeed ready, otherwise happy to work more on it).

Thanks for your patience and all the guidance.

vtomole:
@CirqBot Bad bot.

vtomole:
```python
    if not any(review['state'] == 'APPROVED' for review in review_status):
        return CannotAutomergeError('No approved review.')
```
How!?

Strilanc:
It's because of a pagination issue.

Strilanc:
(testing)

PR #2316, opened 2019-10-11
Offline simulation support in XEB fidelity estimator

viathor:


cirq/experiments/fidelity_estimation.py
```diff
@@ -81,6 +85,11 @@ def linear_xeb_fidelity(
...
-    output_state = final_wavefunction(circuit, qubit_order=qubit_order)
-    output_probabilities = np.abs(output_state)**2
-    return dim * np.mean(output_probabilities[bitstrings]) - 1
+    if amplitudes is None:
+        output_state = final_wavefunction(circuit, qubit_order=qubit_order)
+        output_probabilities = np.abs(output_state)**2
+        bitstring_probabilities = output_probabilities[bitstrings]
+    else:
+        bitstring_probabilities = np.array(
+            [np.abs(amplitudes[bitstring])**2 for bitstring in bitstrings])
```
kevinsung:
```
bitstring_probabilities = np.abs([amplitudes[bitstring] for bitstring in bitstrings])**2
```

cirq/experiments/fidelity_estimation.py
```diff
@@ -26,6 +26,7 @@ def linear_xeb_fidelity(
         circuit: Circuit,
         bitstrings: Sequence[int],
         qubit_order: QubitOrderOrList = QubitOrder.DEFAULT,
+        amplitudes: Optional[Dict[int, complex]] = None,
```
kevinsung: Mapping instead of Dict?

cirq/experiments/fidelity_estimation.py
viathor: Good catch! Thanks!

cirq/experiments/fidelity_estimation.py

PR #2318, opened 2019-10-11
Fix simulator seeding to ensure runs are reproducible

### LINKED ISSUE #2020: Use np.random.RandomState for simulator seeds.

### LINKED ISSUE #2317: Simulators change global random state, resulting in non-reproducible runs
The following gives AssertionError:
```
import cirq
import numpy as np

a = cirq.NamedQubit('a')
circuit = cirq.Circuit(cirq.X(a)**0.5, cirq.measure(a))

sim = cirq.Simulator(seed=1234)
result1 = sim.run(circuit, repetitions=10)

sim = cirq.Simulator(seed=1234)
_ = np.random.randn()
result2 = sim.run(circuit, repetitions=10)

assert result1 == result2
```
Same thing with DensityMatrixSimulator.

kevinsung:
Fixes #2317 
Fixes #2020 

Also fixes a bug in which `cirq.sample` did not pass its seed to the density matrix simulator.

### DIFF
--- a/cirq/sim/density_matrix_simulator.py
+++ b/cirq/sim/density_matrix_simulator.py
@@ -122,16 +122,14 @@ class DensityMatrixSimulator(simulator.SimulatesSamples,
                  *,
                  dtype: Type[np.number] = np.complex64,
                  noise: 'cirq.NOISE_MODEL_LIKE' = None,
-                 seed: Optional[int] = None):
+                 seed: Optional[Union[int, np.random.RandomState]] = None):
         """Density matrix simulator.
 
          Args:
             dtype: The `numpy.dtype` used by the simulation. One of
                 `numpy.complex64` or `numpy.complex128`
             noise: A noise model to apply while simulating.
-            seed: The random seed to use for this simulator. Sets numpy's
-                random seed. Setting numpy's seed different in between
-                use of this class will lead to non-seeded behavior.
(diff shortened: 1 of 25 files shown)


cirq/sim/density_matrix_simulator.py
```diff
@@ -122,25 +122,28 @@ def __init__(self,
...
         if dtype not in {np.complex64, np.complex128}:
             raise ValueError(
                 'dtype must be complex64 or complex128, was {}'.format(dtype))
 
         self._dtype = dtype
         self.noise = devices.NoiseModel.from_noise_model_like(noise)
-        if seed:
-            np.random.seed(seed)
+
+        if seed is None:
```
dabacon: This block of code is repeated in a bunch of places, with various different cases for None.  For None I'd always set it to np.random.  I'd then either stick it into a util class somewhere, or alternatively only make it as Optional in the main entry points where it is used and otherwise set it as a default to seed=np.random.

mpharrigan: For further color, networkx has a decorator https://github.com/networkx/networkx/blob/master/networkx/utils/decorators.py#L354 that they apply to all their functions that need randomness

They also have nice documentation https://networkx.github.io/documentation/stable/reference/randomness.html

PR #2321, opened 2019-10-11
Add `value: Any = 1` parameter to `cirq.one_hot`

jitendrs:


jitendrs:
@vtomole I am a stack license issue. I already submitted the license agreement with the require user but I don't understand why it's failing again and again. do you have any clue?

dabacon:
Strangely this shows two commits from you, one with a user that github seems to understand the other where it doesn't.  My guess is that this is the problem.

dabacon:
@googlebot I fixed it.

dabacon:
I verified this user has signed the CLA.

cirq/linalg/states.py
```diff
@@ -21,6 +21,7 @@
 def one_hot(*,
             index: Union[None, int, Sequence[int]] = None,
             shape: Union[int, Sequence[int]],
+            value: int = 1,
             dtype: Type[np.number]) -> np.ndarray:
     """Returns a numpy array with a single 1 entry, and 0 everywhere else.
```
dabacon: update this top level doc string as it isn't correct now.

jitendrs: Let me check

jitendrs: @dabacon can you please elaborate more about this. I didn't understand. Which docstring you are talking about.

dabacon: Instead of `Returns a numpy array with a single 1 entry, and 0 everywhere else.` this should be something like
`Returns a numpy array with a single non-zero entry (default 1), and 0 everywhere else.`

jitendrs: Thanks @dabacon please review it again.

dabacon: Ah sorry the string I gave you was too long so our pylint check is failing, maybe this is less than 80 chars.
`Returns a numpy array with all 0s and a single non-zero entry (default 1).`

jitendrs: Thanks @dabacon I corrected doc string now.

cirq/linalg/states.py
```diff
@@ -21,13 +21,15 @@
 def one_hot(*,
             index: Union[None, int, Sequence[int]] = None,
             shape: Union[int, Sequence[int]],
+            value: int = 1,
             dtype: Type[np.number]) -> np.ndarray:
-    """Returns a numpy array with a single 1 entry, and 0 everywhere else.
+    """Returns a numpy array with all 0s and a single non-zero entry(default 1).
 
     Args:
         index: The index that should store 1 instead of 0.
```
vtomole:
```suggestion
        index: The index that should store the 'value' argument instead of 0.
```

Strilanc: Use backticks \`value\` instead of single-quotes 'value', so it is understood as a code reference `value`.

cirq/linalg/states.py
```diff
@@ -21,13 +21,15 @@
...
             If not specified, defaults to the start of the array.
         shape: The shape of the array.
+        value: The value of the integer.
```
Strilanc: The array may not be an integer array, it could be an object array. Just

```
The hot value to place at `index` in the result.
```

cirq/linalg/states.py
```diff
@@ -21,13 +21,15 @@
 def one_hot(*,
             index: Union[None, int, Sequence[int]] = None,
             shape: Union[int, Sequence[int]],
+            value: int = 1,
```
Strilanc: `: Any` instead of `: int`.

jitendrs: Default value should be "1" right?

Strilanc: yes

PR #2333, opened 2019-10-11
Compatibility shim for common gate

mpharrigan:
fixes https://github.com/quantumlib/Cirq/pull/2205#issuecomment-539260440

I need this. Feel free to take an add the rest of the moved classes

### DIFF
--- a/cirq/ops/common_gates.py
+++ b/cirq/ops/common_gates.py
@@ -24,6 +24,7 @@ This module creates Gate instances for the following gates:
 Each of these are implemented as EigenGates, which means that they can be
 raised to a power (i.e. cirq.H**0.5). See the definition in EigenGate.
 """
+import warnings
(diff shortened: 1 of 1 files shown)


cirq/ops/common_gates.py
```diff
@@ -37,6 +38,33 @@
 from cirq.type_workarounds import NotImplementedType
 
 
+def SWAP(*args, **kwargs):
```
Strilanc: This is only a half-working shim. For example, `type(cirq.ops.common_gates.SWAP)` will return something different.

I recommend just setting `from cirq.ops.swap_gate import SWAP` with a compatibility note. We will figure out how to deprecate it later.

mpharrigan: done

mpharrigan:
windows doesn't print the failures

Strilanc:
It does. This is the multithreading flake, not a real error.

PR #2337, opened 2019-10-11
TrialResult histogram fold_func should take nparray

smitsanghavi:
fold_func in TrialResult histogram methods doesn't use pd.Series anymore.
Also, fix some TrialResult documentation that referenced pd.Series.


cirq/study/trial_result.py
```diff
@@ -144,7 +145,7 @@ def multi_measurement_histogram(  # type: ignore
             self,
             *,  # Forces keyword args.
             keys: Iterable[TMeasurementKey],
-            fold_func: Callable[[pd.Series], T] = _tuple_of_big_endian_int
+            fold_func: Callable[[np.array], T] = _tuple_of_big_endian_int
```
dabacon: I guess this is np.array since this is always one dimensional (and not np.ndarray, otherwise)?

smitsanghavi: Right, the fold_fn is supposed to operate on the measurement outcomes for a particular measurement key in a particular repetition, so it's guaranteed to be 1-dimensional (the set of qubits in this key).
Actually changing it to use Tuple instead of np.array since that reflects the implementation of the method more accurately.

PR #2338, opened 2019-10-11
Split AsyncSampler out of Sampler

Strilanc:
- Drop auto-threaded run_async (I suspect this will fix the windows build flake)

### DIFF
--- a/cirq/work/collector.py
+++ b/cirq/work/collector.py
@@ -103,7 +103,7 @@ class Collector(metaclass=abc.ABCMeta):
         """
 
     def collect(self,
-                sampler: 'cirq.Sampler',
+                async_sampler: 'cirq.AsyncSampler',
                 *,
                 concurrency: int = 2,
                 max_total_samples: Optional[int] = None) -> None:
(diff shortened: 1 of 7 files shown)


cirq/work/collector.py
```diff
@@ -103,7 +103,7 @@ def on_job_result(self, job: CircuitSampleJob,
         """
 
     def collect(self,
-                sampler: 'cirq.Sampler',
+                async_sampler: 'cirq.AsyncSampler',
```
dabacon: but now collector doesn't work for non-async samplers?  (I think this is why your build is not green?)

Strilanc: Correct. But I think that's a necessary part of how Collector interacts with the world.

dabacon: Hm not being able use a Collector on Simulator seems bad (or I'm missing something)?

mpharrigan:
-1 on this change. Having a stub implementation of async sampling on built-in `Simulator` and friends is helpful for testing async code. Especially if you've hacked your own version of `QuantumEngineSampler` that uses async methods for polling for results. 

It also seems odd to change an aspect of one of our primary APIs for the sake of the particular windows CI service we've chose

Strilanc:
Okay, I will add an implementation of AsyncSampler to the simulator.

The basic problem is that it is not safe to just take unknown code and run it on a separate thread. Implementors should have to opt into multi-threaded behavior like that, indicating it's safe to do. The windows tests failing and the network library sometimes blowing up are both examples of this general problem.

Strilanc:
Obsoleted by https://github.com/quantumlib/Cirq/pull/2562

PR #2342, opened 2019-10-12
Make greedy routing initialization and routing deterministic given seed

bryano:


cirq/contrib/routing/initialization.py
```diff
@@ -50,7 +50,8 @@ def get_initial_mapping(
     """
     if random_state is None:
         prng = np.random
-    elif isinstance(random_state, np.random.RandomState):
+    elif (isinstance(random_state, np.random.RandomState) or
+          random_state == np.random):
```
kevinsung: From the type declaration of `random_state` it can't be `np.random`, so you can remove that check.

mpharrigan: type declarations aren't enforced by anything

bryano: That's a good point.

The check is actually needed given the way that the `random_state` argument is currently parsed in `_GreedyRouter.__init__`. If it's the default `None`, then `_GreedyRouter.prng == np.random`, which is passed to `get_initial_mapping`. So there are two fixes:

1. Update the type declaration of `random_state` to include `np.random`.
2. Modify `_GreedyRouter` so that it only passes `get_initial_mapping` a `random_state` argument consistent with the current type declaration thereof.

(1) Seems better, but it's not clear to me what the right type to add is.

bryano: Independently of the above, it seems like the `random_state` argument parsing boilerplate is getting ripe for refactoring. Any thoughts on where that should go? `cirq.value.random_state`?

kevinsung: Sounds good to me.

PR #2344, opened 2019-10-14
Support both linear and logarithmic XEB fidelity estimators

viathor:
There are other estimators we can also define later, e.g. one based on the HOG test.


cirq/experiments/fidelity_estimation.py
```diff
@@ -22,10 +22,67 @@
...
+
+def xeb_fidelity(
         circuit: Circuit,
         bitstrings: Sequence[int],
         qubit_order: QubitOrderOrList = QubitOrder.DEFAULT,
+        estimator: Callable[[int, Sequence[float]],
+                            float] = linear_xeb_fidelity_estimator,
         amplitudes: Optional[Mapping[int, complex]] = None,
 ) -> float:
     """Computes fidelity estimate from one circuit using linear XEB estimator.
```
vtomole: Needs updating now that it strictly doesn't take `linear_estimator`?

viathor: Thanks for catching! Done.

cirq/experiments/fidelity_estimation.py
```diff
@@ -22,13 +22,70 @@
...
+        probabilities: Ideal probabilities of bitstrings observed in
+            experiment.
+    Returns:
+        Estimate of fidelity associated with an experimental realization
+        of a quantum circuit.
+    """
+    return hilbert_space_dimension * np.mean(probabilities) - 1
+
+
+def log_xeb_fidelity_estimator(hilbert_space_dimension: int,
```
Strilanc: Based on the name of this function, I expected it to return another function or object (an 'estimator'). Rename to something like `log_xeb_fidelity_from_probabilities`.

cirq/__init__.py
```diff
@@ -86,10 +86,12 @@
 )
 
 from cirq.experiments import (
-    linear_xeb_fidelity,
+    linear_xeb_fidelity_estimator,
+    log_xeb_fidelity_estimator,
     generate_boixo_2018_supremacy_circuits_v2,
     generate_boixo_2018_supremacy_circuits_v2_bristlecone,
     generate_boixo_2018_supremacy_circuits_v2_grid,
+    xeb_fidelity,
```
Strilanc: Just have `linear_xeb_fidelity` and `log_xeb_fidelity`. It's not necessary to parameterize this method so flexibly.

cirq/experiments/fidelity_estimation.py
```diff
@@ -22,13 +22,70 @@
...
+
+    Estimates fidelity from ideal probabilities of observed bitstrings.
+    The variance of this estimator is
+
+        (pi^2/6 - f^2) / M
+
+    where f is the fidelity and M the number of observations, equal to
+    len(probabilities). This is better than linear XEB (see above) when
+    fidelity is f > 0.32.
+
```
Strilanc: Describe in more detail what this value represents, e.g. the average Bayes prediction loss in bits per sample.

cirq/experiments/fidelity_estimation.py
```diff
@@ -22,13 +22,70 @@
...
+    return (np.log(hilbert_space_dimension) + np.euler_gamma +
+            np.mean(np.log(probabilities)))
+
+
+def xeb_fidelity(
         circuit: Circuit,
         bitstrings: Sequence[int],
         qubit_order: QubitOrderOrList = QubitOrder.DEFAULT,
+        estimator: Callable[[int, Sequence[float]],
+                            float] = linear_xeb_fidelity_estimator,
```
kevinsung: Unless you really think the flexibility of having a callable is necessary and would be used by people, I think we should use constants named `LINEAR` and `LOG` instead.

cirq/experiments/fidelity_estimation.py

cirq/experiments/fidelity_estimation.py
viathor: Done (described relationship of estimator's MSE to its variance which is particularly simple in this case since the estimators are unbiased).

cirq/__init__.py
viathor: XEB allows you to use any of a whole range of fidelity estimators of which log and linear are just two examples. The estimators differ in important ways such as in the assumptions they make about the output distribution(*) or in the conditions under which their estimate has good quality(**). Also, in some cases, it is interesting to investigate/plot results obtained from more than one estimator. The benefits of being able to replace the estimator while re-using the rest of the code is one of the reasons that motivated #2113.

BTW: Two additional estimators I plan to add:
1. HOG score based fidelity estimator (user requested ability to compare log/linear XEB vs HOG).
2. Cross-entropy ratio since it does not require the assumption that output distribution is PT (and hence may be used on shallow circuits).

----
(*) A common assumption is that the output distribution has Porter-Thomas shape which allows analytical derivation of the simple formulas in the code.
(**) For example, linear estimator is good in low-fidelity regime (e.g. large, deep circuits), but log estimator is better in high-fidelity regime (e.g. benchmarking of a 2-qubit gate).

cirq/experiments/fidelity_estimation.py
viathor: See my reply to a comment above.

cirq/__init__.py
Strilanc: Do you believe that user code will also parameterize over the method in this way? I suspect users will always just be calling a specific one.

Basically, this is setting off my boilerplate alarms. The user wants the linear xeb fidelity, and in order to get that they need to provide the linear estimator to the xeb fidelity method. Why not just ask directly for what they want?

It's okay if we also have an abstract method behind the scenes, or exposed through some more obviously-for-when-generalizing name like "generalized_xeb_fidelity", but I do think we should directly expose `linear_xeb` and `log_xeb` methods.

viathor: It makes sense to me to have functions dedicated to the common case and a general one to address the more nuanced cases requiring extra flexibility.

I kept the name `xeb_fidelity` for the general method and updated the docstring to point out the generality of this function (i.e. that it takes a user-supplied estimator). I think the name `generalized_xeb_fidelity` would suggest the overall methodology of XEB is somehow generalized here which is not the case. Also, in literature all these variants are referred to simply as "XEB" and when a specific case is meant we use terms like "linear XEB" or "log XEB". This agrees well with our present function naming.

viathor:
PTAL

Issue #2326, opened 2019-10-11
cirq.merge_single_qubit_gates_into_phased_x_z(circuit) is exponential


dabacon:
```
    qubits = cirq.LineQubit.range(1)
    gate_domain = {
        cirq.X**0.2: 1,
        cirq.X**0.7: 1,
        cirq.X: 1,
        cirq.Y**0.2: 1,
        cirq.Y**0.7: 1,
        cirq.Y: 1,
        cirq.Z**0.2: 1,
        cirq.Z**0.7: 1,
        cirq.Z: 1,
    }

    for n in range(0, 1000, 100):
        circuit = cirq.testing.random_circuit(qubits=qubits, n_moments=n,
                                              op_density=0.9999, gate_domain=gate_domain)
        start = time.time()
        cirq.merge_single_qubit_gates_into_phased_x_z(circuit)
        delta = time.time() - start
        print('n: {} time: {}'.format(n, delta))
```
scales poorly:
```
n: 0 time: 0.00012683868408203125
n: 100 time: 0.16895222663879395
n: 200 time: 1.22210693359375
n: 300 time: 4.136816501617432
n: 400 time: 9.669109582901001
n: 500 time: 19.23963689804077
n: 600 time: 34.95179057121277
```

dstrain115:
This issue is resolved and now runs 100x faster.

PR #2345, opened 2019-10-14
Speed up linalg.dot

dstrain115:
The current linalg.combinators.dot uses np.linalg_multi_dot which does some fancy calculations to figure out the correct order to apply the dot product.  Since, most of the time, we are just multiplying matrices of the same size (2x2 or 4x4), there is no optimal order and we should just dot them in order.  This new function just does this and gets rid of the multi_dot call.  

I have kept the old functionality in case anyone needs it.

Here is the new output from https://github.com/quantumlib/Cirq/issues/2326 (shows over 100x improvement):
n: 0 time: 0.00013518333435058594
n: 100 time: 0.004878044128417969
n: 200 time: 0.00897073745727539
n: 300 time: 0.013186454772949219
n: 400 time: 0.018898725509643555
n: 500 time: 0.022220134735107422
n: 600 time: 0.02526688575744629
n: 700 time: 0.029251575469970703
n: 800 time: 0.034186601638793945
n: 900 time: 0.03860616683959961


cirq/linalg/combinators.py
```diff
@@ -112,6 +114,31 @@ def dot(*values: Union[float, complex, np.ndarray]
...
+    sizes.  If you are using this for matrices of large and differing sizes,
+    consider using multi_dot for better performance
+
+    Args:
+        *values: The values to combine with the dot/matrix product.
+
+    Returns:
+        The resulting value or matrix.
+    """
+
```
dabacon: does this work for the len 0 case?

cirq/linalg/combinators.py
```diff
@@ -112,6 +114,31 @@ def dot(*values: Union[float, complex, np.ndarray]
...
+    if len(values) == 1:
+        if isinstance(values[0], np.ndarray):
+            return np.array(values[0])
+        return values[0]
```
cduck: I think you meant `if not isintance...`.

You can juse use `return np.asarray(values[0])` here.

Strilanc: I wouldn't do that. The caller may assume that the output will not alias with the input.

cirq/linalg/combinators.py
```diff
@@ -112,6 +114,31 @@ def dot(*values: Union[float, complex, np.ndarray]
     return np.linalg.multi_dot(values)
 
 
+def dot(*values: Union[float, complex, np.ndarray]
```
Strilanc: While we're tweaking this method we should add a `dtype=None` parameter for forcing the output dtype.

cirq/linalg/combinators.py
Strilanc: I think we should purposefully fail in the 0 length case, to be safe. There's no way to know the user's expected shape. We could return the scalar 1, but that will almost definitely cause code to break in corner cases anyways.

dstrain115: It's not clear to me what you would want to return with no arguments, so I will just add a raise ValueError.

cirq/linalg/combinators.py
dstrain115: I don't think this is correct.  I think the original intent was to return a copy of the array.

cduck: Then why `return values[0]`?

cirq/linalg/combinators.py
viathor: The value of any inner product (and thus also dot product as a special case) is well-defined in the zero-dimensional space. This space consists of a single vector - the zero vector. And one of the inner product axioms explicitly demands that <0, 0> = 0.

This is also consistent with another widely used convention that calls for the neutral element to be the value of a zero-length sum/product/etc (so zero-length sum is zero, zero-length product is one etc). Dot product is defined as a sum, so empty dot is zero.

viathor: Numpy agrees:

```
>>> z = np.empty(shape=(0,))
>>> z
array([], dtype=float64)
>>> np.dot(z, z)
0.0
>>> 
```

dstrain115: I am not sure that is the correct analogy.  This is not a dot product in a zero-dimensional space.  Your example is a dot product of 2 elements (len=2) of zero-dimensions.  This is a dot product of no elements.  That doesn't make any sense and is likely a programming error.

As a counter point, numpy fails with the analogous case:

```
In [3]: np.linalg.multi_dot()                                                                                                                           
TypeError                                 Traceback (most recent call last)
<ipython-input-3-aaa16d36e400> in <module>
----> 1 np.linalg.multi_dot()

TypeError: multi_dot() missing 1 required positional argument: 'arrays'
```

viathor: You're right. Sorry for not having read the full context.

Not even the rule about the neutral element can be applied here because without knowing the shape we wouldn't be able to pick the correct identity matrix.

cirq/linalg/combinators.py
dstrain115: According to the mypy type annotation, if it's not an numpy array, it's a float or complex which are immutable.

cirq/linalg/__init__.py
```diff
@@ -18,13 +18,8 @@
...
 
-from cirq.linalg.combinators import (
-    block_diag,
-    CONTROL_TAG,
-    dot,
-    kron,
-    kron_with_controls,
-)
+from cirq.linalg.combinators import (block_diag, CONTROL_TAG, dot, kron,
+                                     kron_with_controls, multi_dot)
```
dabacon: I think format made you do this because you don't have a trailing `,` after `multi_dot`?

cirq/linalg/combinators.py
```diff
@@ -93,25 +93,56 @@ def kron_with_controls(*factors: Union[np.ndarray, complex, float]
...
+       ) -> Union[float, complex, np.ndarray]:
+    """Computes the dot/matrix product of a sequence of values.
+
+    Performs the computation in serial order without regard to the matrix
+    sizes.  If you are using this for matrices of large and differing sizes,
+    consider using multi_dot for better performance
```
dabacon: nit: missing period.

cirq/linalg/combinators.py
dabacon: Do we have a use case for this?  `astype()` is an easy way to do this, I'd say wait until you see that this is commonly needed?

Strilanc: k

cirq/linalg/combinators.py
```diff
@@ -93,25 +93,56 @@ def kron_with_controls(*factors: Union[np.ndarray, complex, float]
     return product
 
 
-def dot(*values: Union[float, complex, np.ndarray]
-        ) -> Union[float, complex, np.ndarray]:
+def multi_dot(*values: Union[float, complex, np.ndarray]
```
Strilanc: Why did you rename the method? This is a breaking change.

dstrain115: I am replacing this with another method.  I change the old one to multi_dot() and then creating a new dot() to replace it.  

Do you think I should just delete the old one instead?

Strilanc: I think we should just have `dot` and have it just skip all the optimization stuff. If people want that they can call numpy directly.

dstrain115: Ok, I deleted the old version.  Now it's just the serial dot product.

PR #2346, opened 2019-10-14
Add simplified `extra_cell_makers` argument option to `quirk_url_to_circuit`

Strilanc:
- If the user specifies a dictionary from strings to gates, wrap it in the obvious way

### DIFF
--- a/cirq/contrib/quirk/url_to_circuit.py
+++ b/cirq/contrib/quirk/url_to_circuit.py
@@ -127,7 +127,7 @@ def quirk_url_to_circuit(
         raise ValueError('Circuit JSON cols must be a list.\n'
                          f'URL={quirk_url}')
 
-    # Parse column json into cells.
+    # Collect registry of quirk cell types.
     if isinstance(extra_cell_makers, Mapping):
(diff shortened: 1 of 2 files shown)


cirq/contrib/quirk/url_to_circuit.py
```diff
@@ -46,10 +46,12 @@ def quirk_url_to_circuit(
             qubits). The maximum number of qubits in a Quirk circuit is 16.
             This argument defaults to `cirq.LineQubit.range(16)` when not
             specified.
-        extra_cell_makers: A list of non-standard Quirk cell makers. This can be
+        extra_cell_makers: Non-standard Quirk cells to accept. This can be
             used to parse URLs that come from a modified version of Quirk that
-            includes gates that Quirk doesn't define. See
-            `cirq.contrib.quirk.cells.CellMaker`.
+            includes gates that Quirk doesn't define. This can be specified
+             as either a list of `cirq.contrib.quirk.cells.CellMaker` instances,
```
viathor: nit: this and two following lines are indented by one more space than they should be

PR #2348, opened 2019-10-15
adds parse_random_state and associated types

bryano:


cirq/value/random_state.py
```diff
@@ -0,0 +1,49 @@
...
+from typing_extensions import Protocol
+
+import numpy as np
+
+RANDOM_STATE_LIKE = Optional[Union[np.random.RandomState, int]]
+
+
+class PseudoRandNumGen(Protocol):
+
+    def choice(self, *args, **kwargs):
```
mpharrigan: this is going to be annoying for IDE autocomplete and doc hints. Can't you just make `parse_random_state` return a `RandomState`. I know `np.random` technically is a *module* and not a `RandomState`, but it isn't anything a well placed `cast` can't fix. 

You can also do the other thing of actually extracting the numpy global `RandomState`, but it involves accessing private members

cirq/value/random_state.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def parse_random_state(random_state: RANDOM_STATE_LIKE) -> PseudoRandNumGen:
+    if random_state is None:
+        return np.random
+    elif (isinstance(random_state, np.random.RandomState) or
+          random_state == np.random):
+        return cast(PseudoRandNumGen, random_state)
+    elif isinstance(random_state, int):
+        return np.random.RandomState(random_state)
```
mpharrigan: mypy can't be happy about any of these return values (?)

cirq/value/random_state.py
```diff
@@ -0,0 +1,49 @@
...
+    raise TypeError(f'Argument must be of type cirq.value.SEED.')
```
mpharrigan: "SEED" -> "RANDOM_STATE_LIKE"

cirq/value/random_state.py
kevinsung: I agree what we don't need to make this new class. We can use `types.ModuleType` for the type of `np.random`.

mpharrigan: well I think one of the advantages of the below-function would be to return an object with a consistent type. I really wish numpy would just expose their global `RandomState` with public methods!

kevinsung: Not that this really solves the issue, but we can define a type alias `PRNG_LIKE = Union[np.random.RandomState, types.ModuleType]`

mpharrigan:
this would be good to get in

cirq/value/random_state.py
bryano: Fixed. Note that this is technically incorrect (because the argument can be `np.random`), but I think it's fine because I imagine that only being used internally, e.g., when a user passes `None` to a function that then passes `np.random` to another one.

cirq/value/random_state.py
bryano: Why not?

cirq/value/random_state.py
bryano: I went with casting. While `PRNG_LIKE` would be more "honest" with mypy, the premise is that `np.random` quacks like a `RandomState`.

mpharrigan: EFleC2FXkAEphM1

cirq/value/random_state.py
mpharrigan: I think before there were different return types for different code paths(?) anyways looks good now

mpharrigan:
test failure is relevant. I think the max seed is 2**32-1

cirq/value/random_state.py
kevinsung: I don't think we should cast it to a RandomState when it isn't one. This could lead, for instance, to users accidentally changing the global random state by calling `prng.seed` because they think `prng` is a RandomState when really it is `np.random`.

mpharrigan: I don't think the type annotation is the appropriate place to "document" this caveat. For one thing, both types have a `seed` method that does the same thing so it doesn't disambiguate it at all.

kevinsung: I'm not sure what you mean by the `seed` method doing the same thing; in one case it changes the global seed and in another it only changes the seed of the particular RandomState instance. In any case that was just one example of what could go wrong in lying about the type; I can't think of another example for this particular case right now but it still seems like a bad idea to me.

cirq/value/random_state_test.py
```diff
@@ -0,0 +1,49 @@
...
+    prngs = [
+        np.random,
+        cirq.value.parse_random_state(np.random),
+        cirq.value.parse_random_state(None)
+    ]
+    vals = [rand(prng) for prng in prngs]
+    eq = cirq.testing.EqualsTester()
+    eq.add_equality_group(*vals)
+
+    seed = np.random.randint(-2**31, 2**31)
```
kevinsung: Negative numbers aren't valid inputs to np.random.RandomState`:

```
>>> np.random.RandomState(-1)
  File "<stdin>", line 1, in <module>
  File "mtrand.pyx", line 644, in mtrand.RandomState.__init__
  File "mtrand.pyx", line 680, in mtrand.RandomState.seed
ValueError: Seed must be between 0 and 2**32 - 1
```

cirq/value/random_state_test.py
```diff
@@ -0,0 +1,49 @@
...
+    np.random.RandomState(seed)
```
kevinsung: This statement doesn't do anything (its return value is not being assigned to a variable).

bryano: Removed.

cirq/value/random_state.py
mpharrigan: In one case it seeds the global `RandomState` and the other seeds a local `RandomState`. The only caveat is that *technically* the global `RandomState` is a module composed entirely of functions that defer to the global `RandomState` object which is hidden with a preceding underscore. 

It's the difference between `x = cirq.CNOT` and `x = cirq.CNotPowGate()`. They're both of type `CNotPowGate`

mpharrigan: and we're using this helper function to paper over the technicality and assert that global and local random states share a type, which they basically do

cirq/value/random_state_test.py
bryano: Fixed

cirq/value/random_state.py
bryano: Maybe @Strilanc can weigh in.

Strilanc: Technically it's already the case that the local RandomState instances you get may be used in lots of other places, and calling `seed` on it can cause problems. If someone calls `seed` on an argument they were given then they are probably doing something wrong *anyways*. You don't mutate your arguments without clear naming conventions and documentation saying that you are.

The only reason it's a little iffy here is that we're the ones silently passing the global object instead of a new local instance. The main worry would be that the user infers we are creating independent instances when no seed or random state instance is specified, thinks they can seed it, and then shoots themselves in the foot. E.g. they might have some utility method that seeds random data, that they call all the time, effectively making their computation re-use the same random numbers again and again.

We could use a wrapper object that intercepted the call to seed and raised an error.

We could use a "copy on write" wrapper object that switched to a real RandomState if you called seed.

Probably the simplest thing is to just let them shoot themselves in the foot and then tell them not to use the seed method :l

PR #2349, opened 2019-10-15
Add cirq.PauliSum.__pow__ method, generalize cirq.PauliSum.__mul__ methods

### LINKED ISSUE #2265: PauliSum should support multiplication with PauliSum and raising to integer power

peterse:
Fixes #2265


cirq/ops/linear_combinations.py
```diff
@@ -487,17 +487,50 @@ def __neg__(self):
...
-    def __rmul__(self, a: value.Scalar):
-        return self.__mul__(a)
+    def __mul__(self, other: PauliSumLike):
+        if isinstance(other, (float, int, complex)):
+            result = self.copy()
+            result *= other
+            return result
+        elif isinstance(other, PauliString):
+            return PauliSum.from_pauli_strings(
+                [term * other for term in iter(self)])
```
kevinsung: The `iter` shouldn't be necessary.

cirq/ops/linear_combinations.py
```diff
@@ -487,17 +487,50 @@ def __neg__(self):
...
-    def __mul__(self, a: value.Scalar):
-        result = self.copy()
-        result *= a
-        return result
+    def __imul__(self, other: PauliSumLike):
+        if isinstance(other, (float, int, complex)):
+            self._linear_dict *= other
+            return self
+        # No convenient way to mutate paulistring members.
+        return self.__mul__(other)
```
Strilanc: We need to be consistent about whether or not `*=` modifies the origin or makes a copy. This is a very surprising change in aliasing behavior based on the type of the input. The method has to stick with the current behavior of modifying inline.

Note that you shouldn't be mutating the pauli strings themselves, but you should be changing the keys and values that are in the dictionary.

I think the correct behavior here would be equivalent to `self._linear_dict = self.__mul__(other)._linear_dict`. But you should implement mul in terms of imul instead of vice versa.

cirq/ops/linear_combinations_test.py
```diff
@@ -967,6 +967,72 @@ def test_bad_arithmetic():
...
+    psum3 = cirq.X(q0) * cirq.Z(q1) + 1.3 * cirq.Z(q0)
+    sqd = cirq.PauliSum.from_pauli_strings(
+        [2.69 * cirq.PauliString(cirq.I(q0))])
+    assert cirq.approx_eq(psum3 ** 2, sqd, atol=1e-8)
+
+    psum4 = cirq.X(q0) * cirq.Z(q1) + 1.3 * cirq.Z(q1)
+    sqd2 = cirq.PauliSum.from_pauli_strings(
+        [2.69 * cirq.PauliString(cirq.I(q0)), 2.6 * cirq.X(q0)])
+    assert cirq.approx_eq(psum4 ** 2, sqd2, atol=1e-8)
+
```
Strilanc: Add this test:

```
test_imul_aliasing():
    psum1 = ...(something)...
    psum2 = psum1
    psum2 *= ...(pauli string)...
    assert psum1 == psum2
```

cirq/ops/linear_combinations.py
mpharrigan: I think by convention the `*=` *must* modify the original (?). Otherwise you shouldn't support `*=`

cirq/ops/linear_combinations_test.py
peterse: done

cirq/ops/linear_combinations.py
peterse: okay good catch; fixed it.

peterse:
So the paulisum * paulistring and paulisum * paulisum multiplication methods are not optimized, for example
 - psum1 * psum2 both on single qubit would be faster using `operator_spaces.pow_pauli_combination` 
 - psum * cirq.I should short-circuit return psum
 - both of the above applied to any product over a subset of terms between two paulisums

I chose not to implement these because parsing linear dicts quickly becomes unreadable for such a small performance boost, but ya'll may not feel the same way.

cirq/ops/linear_combinations.py
```diff
@@ -500,17 +500,50 @@ def __neg__(self):
         factory = type(self)
         return factory(-self._linear_dict)
 
-    def __imul__(self, a: value.Scalar):
-        self._linear_dict *= a
+    def __imul__(self, other: PauliSumLike):
+        if not isinstance(other, (float, int, complex, PauliString, PauliSum)):
```
kevinsung: Use `numbers.Complex` instead of ` float, int, complex`.

cirq/ops/linear_combinations.py
```diff
@@ -500,17 +500,50 @@ def __neg__(self):
...
+            return NotImplemented
+        if isinstance(other, (float, int, complex)):
```
kevinsung: Use `numbers.Complex` instead of ` float, int, complex`.

cirq/ops/linear_combinations.py
```diff
@@ -500,17 +500,50 @@ def __neg__(self):
...
+            return NotImplemented
         result = self.copy()
-        result *= a
+        result *= other
         return result
 
-    def __rmul__(self, a: value.Scalar):
-        return self.__mul__(a)
+    def __rmul__(self, other: PauliSumLike):
+        if isinstance(other, (float, int, complex)):
```
kevinsung: Use `numbers.Complex` instead of ` float, int, complex`.

cirq/ops/linear_combinations.py
```diff
@@ -500,17 +500,50 @@ def __neg__(self):
...
+        elif isinstance(other, PauliSum):
+            temp = PauliSum.from_pauli_strings(
+                [term * other_term for term in self for other_term in other])
+            self._linear_dict = temp._linear_dict
+
         return self
 
-    def __mul__(self, a: value.Scalar):
+    def __mul__(self, other: PauliSumLike):
+        if not isinstance(other, (float, int, complex, PauliString, PauliSum)):
```
kevinsung: Use `numbers.Complex` instead of ` float, int, complex`.

cirq/ops/linear_combinations.py
```diff
@@ -500,17 +500,50 @@ def __neg__(self):
...
+            result = self.copy()
+            result *= other
+            return result
+        elif isinstance(other, PauliString):
+            result = self.copy()
+            return PauliSum.from_pauli_strings([other]) * result
+        return NotImplemented
+
+    def __pow__(self, exponent: int):
+        if not isinstance(exponent, int):
```
kevinsung: Use `numbers.Integral` instead of `int`.

cirq/ops/linear_combinations.py
peterse: done

Issue #2179, opened 2019-09-25
PauliString.pass_operations_over doesn't accept Pauli


JulianSKelly:
`PauliString.pass_overations_over` only accepts `SingleQubitCliffordGate`. For example:

```python
q = cirq.NamedQubit('q0')
pauli_string = cirq.PauliString({q: cirq.X})
op = cirq.X.on(q)
pauli_string.pass_operations_over([op])
```
raises `TypeError: Unsupported operation: cirq.X.on(cirq.NamedQubit('q0'))`

The expected use-case is AFAICT:
```python
q = cirq.NamedQubit('q0')
pauli_string = cirq.PauliString({q: cirq.X})
op = cirq.SingleQubitCliffordGate.from_pauli(cirq.X).on(q)
pauli_string.pass_operations_over([op])
```

This is pretty awkward given the constructor takes a `Pauli`.

JulianSKelly:
There is a test explicitly for this behavior: https://github.com/quantumlib/Cirq/blob/116fbf5480c40a04af64aad4cb006db3adb5ffac/cirq/ops/pauli_string_test.py#L613-L617

Strilanc:
This is a usability bug. The code and the test should be fixed to work when given `cirq.X`, `cirq.Y`, `cirq.Z`, `cirq.CNOT`, etc.

If we're feeling really ambitious, we should make stuff like `cirq.SingleQubitMatrix` work when it happens to approximately encode a Clifford. That would require writing a matrix-to-clifford method.

I suspect there are several other places where the "Clifford" gates interop poorly with the "normal" gates.

PR #2351, opened 2019-10-15
Fix `PauliString.pass_operations_over` not supporting common gates

Strilanc:
- Add `_decompose_into_clifford_` method to `GateOperation`
- Add `_decompose_into_clifford_with_qubits_` method to several common gates
- Add `PauliString.conjugated_by`, which accepts an `OP_TREE` defining a composite Clifford operation in Circuit order. This differs from `pass_ops_over`, because `pass_ops_over` takes a list of operations in reverse circuit order (which is confusing).
- Fix op_gate_isinstance type annotation not allowing type tuples
- Add `PauliString.dense` method

Code such as the following now works:

```
print(cirq.X(a).pass_operations_over([cirq.H(a)]))
print(cirq.X(a).conjugated_by(cirq.Circuit(cirq.H(a), cirq.CNOT(a, b))))
```

```
Z(0)
Z(0)*X(1)
```

Fixes https://github.com/quantumlib/Cirq/issues/2179


Strilanc:
Reviewers: please check that `conjugated_by` does the thing that you would expect based on the name. Ideally, predict it before reading the code or the tests.

Strilanc:
Yes, I kept the decompose methods private without a publicly exposed method because I didn't want to commit to the interface yet.

mpharrigan:
is anything blocking this?

Strilanc:
@mpharrigan Nothing except actually getting a review of it.

PR #2352, opened 2019-10-15
Speed up findall_operations_until_blocked

dstrain115:
- slicing a Circuit actually creates a new Circuit (see circuits.py:214) which we do not need to iterate on the moments.  Instead, use self._moments
- Also return early if we run out of active_qubits.

For an example circuit (depth 1270 circuit of 16 qubits, mixed 1q/2q gates), this speeds the merge_rots optimization (which depends on findall_operations_until_blocked) from 140 seconds to 1.5 seconds.


dabacon:
Nice.

PR #2353, opened 2019-10-15
TESTING: Python 3.8 on Travis

vtomole:


dabacon:
We are starting to slam travis pretty hard, I wonder if we gain much by doing this?  Something to discuss at Cirq Cync :)

vtomole:
@dabacon Noted.

PR #2354, opened 2019-10-15
Speeds up circuit copying

dstrain115:
- Circuit copying currently appends moments one by one to the new
circuit.
- This causes a lot of qid shape verification and device verification.
- This is not needed for direct copying (copy()) and for slices
(circuit[2:).
- For a depth 1000 circuit with 16 qubits, this speeds circuit copy()
from 0.09 seconds to 0.00004 seconds. (About a 1000x improvement)


cirq/circuits/circuit.py
```diff
@@ -211,7 +213,9 @@ def __getitem__(self, key: int) -> 'cirq.Moment':
 
     def __getitem__(self, key):
         if isinstance(key, slice):
-            return Circuit(self._moments[key], device=self.device)
+            sliced_circuit = Circuit(device=self.device)
```
dabacon: in this case does it actually cause a speed difference, since this only appends a single moment?

cirq/circuits/circuit.py
```diff
@@ -211,7 +213,9 @@ def __getitem__(self, key: int) -> 'cirq.Moment':
...
+            sliced_circuit._moments = self._moments[key]
```
dabacon: I'm confused because this looks like List = Moment?

cirq/circuits/circuit.py
dstrain115: The key is a slice.  It's not just one element.  So, if you do large_circuit[2:], it copies the whole circuit to do the slice.  If the circuit is big, doing this slice can be a surprisingly inefficient waste of time.

cirq/circuits/circuit.py
dstrain115: See above.  It's a slice, not an element.

cirq/circuits/circuit.py
dabacon: ah got it, a signature on the key would be helpful

PR #2358, opened 2019-10-15
Add support for serializing linear combinations of symbols

Strilanc:


cirq/google/arg_func_langs.py
```diff
@@ -17,4 +17,5 @@
 import sympy
 
 # Argument types for gates. ArgFunction's are not currently supported.
```
dabacon: This comment is no longer true.

cirq/google/op_serializer.py
```diff
@@ -200,6 +201,14 @@ def _arg_value_to_proto(self, value: arg_func_langs.ArgValue,
             msg.arg_value.bool_values.values.extend(value)
         elif isinstance(value, sympy.Symbol):
             msg.symbol = str(value.free_symbols.pop())
+        elif isinstance(value, sympy.Add):
```
dabacon: You need to also set the arg func language in program.proto (don't forget there are current two copies during migration of paths)
`arg_function_language = 2;`

cirq/google/op_deserializer.py
```diff
@@ -120,30 +120,59 @@ def _args_from_proto(self, proto: v2.program_pb2.Operation
...
+                return float(arg_value.float_value)
+            if which_val == 'bool_values':
+                return list(arg_value.bool_values.values)
+            if which_val == 'string_value':
+                return str(arg_value.string_value)
+            raise ValueError(f'Unrecognized value type: {which_val!r}')
+
+        if which == 'symbol':
+            return sympy.Symbol(arg_proto.symbol)
+
```
dabacon: validate that the `arg_func_language` is correct.

cirq/google/arg_func_langs.py
```diff
@@ -17,4 +17,5 @@
 import sympy
```
dabacon: I'd define `arg_func_language` from program.proto here as a constant, and possibly describe it (or move functionality that does serialization and deserialization here).

cirq/google/op_serializer.py
Strilanc: Urgh it's so much worse than that. Nothing is passing around the language at the moment, or even setting it to something. I have to plumb it the whole way through.

Strilanc:
@dabacon PTAL. This has massively changed.

cirq/google/arg_func_langs.py
```diff
@@ -12,9 +12,196 @@
...
 
+import numpy as np
 import sympy
+from cirq.api.google import v2
 
-# Argument types for gates. ArgFunction's are not currently supported.
-ArgValue = Union[int, float, List[bool], str, sympy.Symbol]
+SUPPORTED_FUNCTIONS_FOR_LANGUAGE = {
+    '': set(),
+    'linear': {'add', 'mul'},
```
dstrain115: A general question:  you've made things very general here.  However, I am trying to think through why we might need things so complicated.  This is all in cirq/google, so it can be specific to our use case.  Can you explain a case why we might need multiple different arg languages besides nothing and everything?

Is this to provide a smooth upgrade path for when we add new functions to servers/clients at different times?  I suppose the simulator could support less options?

cirq/google/serializable_gate_set.py
```diff
@@ -98,15 +111,32 @@ def serialize(self,
             msg = v2.program_pb2.Program()
         msg.language.gate_set = self.gate_set_name
         if isinstance(program, circuits.Circuit):
-            self._serialize_circuit(program, msg.circuit)
+            self._serialize_circuit(program,
+                                    msg.circuit,
+                                    arg_function_language=arg_function_language)
+            if arg_function_language is None:
+                arg_function_language = (
+                    arg_func_langs._infer_function_language_from_circuit(
```
dstrain115: Iterating through a large circuit could be a decent sized overhead.  Why can't we just assume the most permissive in this case?  Isn't that why you added None to the dictionary above?

cirq/google/arg_func_langs.py
```diff
@@ -12,9 +12,196 @@
...
+
+    supported = SUPPORTED_FUNCTIONS_FOR_LANGUAGE.get(arg_function_language)
+    if supported is None:
+        raise ValueError(f'Unrecognized arg_function_language: '
+                         f'{arg_function_language!r}')
+
+    msg = v2.program_pb2.Arg() if out is None else out
+
+    def check_support(func_type: str) -> str:
+        if func_type not in cast(Set[str], supported):
```
dstrain115: Would it be cleaner to put a type annotation on SUPPORTED_FUNCTIONS_FOR_LANGUAGE rather than need to cast here?  Will that work?

cirq/google/arg_func_langs.py
```diff
@@ -17,4 +17,5 @@
 import sympy
 
 # Argument types for gates. ArgFunction's are not currently supported.
-ArgValue = Union[int, float, List[bool], str, sympy.Symbol]
+ArgValue = Union[int, float, List[bool], str, sympy.Symbol, sympy.Add, sympy.
```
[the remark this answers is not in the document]
Strilanc: There is now a `SUPPORTED_FUNCTIONS_FOR_LANGUAGE` constant at the top of the file.

cirq/google/arg_func_langs.py
```diff
@@ -12,9 +12,196 @@
...
+    'linear': {'add', 'mul'},
+    # None means any. Is used when inferring the language during serialization.
+    None: {'add', 'mul'},
+}
+
+# Argument types for gates.
+ARG_LIKE = Union[int, float, List[bool], str, sympy.Symbol, sympy.Add, sympy.
+                 Mul]
+
+LANGUAGE_ORDER = [
```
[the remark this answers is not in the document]
Strilanc: Added a comment.

cirq/google/arg_func_langs.py
Strilanc: It's to provide a smooth upgrade path. For example, suppose we add support for trig functions. Clients that support trig functions but run parameterized circuits that don't use trig but do use linear combinations will still be able to talk to a server that only supports linear combinations.

cirq/google/arg_func_langs.py
Strilanc: The issue is actually that mypy things `supported` could be `None`. I rewrote the above lines to convince it otherwise. Also added the type annotation.

cirq/google/op_serializer.py
```diff
@@ -164,7 +164,7 @@ def _value_from_gate(self, gate: 'cirq.Gate',
                 'Argument {} is required, but could not get from gate {!r}'.
                 format(arg.serialized_name, gate))
 
-        if isinstance(value, sympy.Symbol):
+        if isinstance(value, (sympy.Symbol, sympy.Add, sympy.Mul)):
```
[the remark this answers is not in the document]
Strilanc: Done.

cirq/google/serializable_gate_set.py
Strilanc: We want to use the least permissive, in order to run against older servers.

cirq/google/arg_func_langs.py
```diff
@@ -12,9 +12,211 @@
...
+                          arg_function_language=arg_function_language,
+                          out=msg.func.args.add())
+    elif isinstance(value, sympy.Mul):
+        msg.func.type = check_support('mul')
+        for arg in value.args:
+            _arg_to_proto(arg,
+                          arg_function_language=arg_function_language,
+                          out=msg.func.args.add())
+    else:
+        raise ValueError(f'Unrecognized of arg type: {type(value)}')
```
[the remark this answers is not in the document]
Strilanc: fixed.

cirq/google/arg_func_langs_test.py
```diff
@@ -0,0 +1,155 @@
...
+from cirq.google.arg_func_langs import (
+    _arg_from_proto,
+    _arg_to_proto,
+    ARG_LIKE,
+    LANGUAGE_ORDER,
+)
+from cirq.api.google import v2
+
+
+@pytest.mark.parametrize('min_lang,value,proto', [
```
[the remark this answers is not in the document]
Strilanc: I've seen where combining test data into a common place goes, and it's not good. The common data grows and grows in its complications as more tests use it, everyone adds on their own little parameter to make their one test work, and ultimately individual tests become totally incomprehensible because you don't know which properties matter and which don't for that test.

Tests details should be kept very local, even if that results in code duplication. There are many exceptions to this, but the underlying idea of "tolerate more duplication in tests" is a good rule of thumb.

cirq/google/arg_func_langs.py
```diff
@@ -12,9 +12,211 @@
...
+
+    def check_support(func_type: str) -> str:
+        if func_type not in supported:
+            lang = (repr(arg_function_language)
+                    if arg_function_language is not None else '[any]')
+            raise ValueError(f'Function type {func_type!r} not supported by '
+                             f'arg_function_language {lang}')
+        return func_type
+
+    if isinstance(value,
```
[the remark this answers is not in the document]
Strilanc: > Would this miss things like `sympy.pi` ?

I added `sympy.NumberSymbol` which will handle pi by turning it into a float. It's okay if we miss some of these cases initially because when we serialize we don't have to remember if it was the symbol or the raw number.

> putting all of these if's elif's in a seperate function

This function is already the extracted function for holding all the if's and elif's :stuck_out_tongue: ! We'll refactor as needed as it grows.

cirq/google/arg_func_langs.py
```diff
@@ -12,9 +12,211 @@
...
+        arg_proto: v2.program_pb2.Arg,
+        *,
+        arg_function_language: str,
+        required_arg_name: Optional[str] = None,
+) -> Optional[ARG_LIKE]:
+    """Extracts a python value from an argument value proto.
+
+    Args:
+        arg_proto: The proto containing a serialized value.
+        required_arg_name: If set to `None`, the method will return `None` when
```
dabacon: nit: order does not match arg order

cirq/google/op_deserializer.py
```diff
@@ -12,10 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Optional,
+    Sequence,
+    TYPE_CHECKING,
```
dabacon: nit: remove comma to get formatting that doesn't fill up my screen when I open the file :)

cirq/google/arg_func_langs.py
```diff
@@ -12,9 +12,211 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import List, Union
+from typing import (
```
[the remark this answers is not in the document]
Strilanc: Acknowledged. I prefer the vertical list style to the block style, as it makes git diffs cleaner.

cirq/google/arg_func_langs.py
Strilanc: Fixed.

cirq/google/op_deserializer.py
Strilanc: Acknowledged. I prefer the vertical list style to the block style, as it makes git diffs cleaner.

dabacon: I think the white space is a big waste.  Plus no one else is doing this in the code base.

PR #2361, opened 2019-10-16
Turn on flag to parallelize build docs

dabacon:
-j auto uses the number of cpus.  

On my box this sped up the test by a factor of 2.


dabacon:
sphinx-markdown-table is marked as parallelizable in version on github, but not in one that is pip installed.

dabacon:
OK, hacked in the fix for tables to see if we got a speed improvement, and...no.  So closing.

PR #2363, opened 2019-10-16
Add apply method for mixtures.

MichaelBroughton:
At @Strilanc 's reqeust. Adds the `apply_mixture` protocol to let people define fast `_apply_mixture_` functions, with fallbacks to`_apply_unitary_`, `_unitary_` and `_mixture_`.


cirq/protocols/apply_mixture_protocol.py
```diff
@@ -0,0 +1,309 @@
...
+            as the target tensor.
+        auxiliary_buffer1: Pre-allocated workspace with the same shape
+            and dtype as the target tensor.
+        left_axes: Which axes to multiply the left action of the mixture upon.
+        right_axes: Which axes to multiply the right action of the mixture upon.
+    """
+
+    def __init__(self, target_tensor: np.ndarray, out_buffer: np.ndarray,
+                 auxiliary_buffer0: np.ndarray, auxiliary_buffer1: np.ndarray,
+                 left_axes: Iterable[int], right_axes: Iterable[int]):
```
Strilanc: We should make `right_axes` optional, in order to support the use case of applying to a state vector. For consistency we could also make `left_axes` optional.

cirq/protocols/apply_mixture_protocol.py
```diff
@@ -0,0 +1,309 @@
...
+                         'are not equal. Got {!r} and {!r}.'.format(
+                             left_shape, right_shape))
+    if val_qid_shape != left_shape:
+        raise ValueError('Invalid mixture qid shape is not equal to the '
+                         'selected left and right shape of target_tensor. '
+                         'Got {!r} but expected {!r}.'.format(
+                             val_qid_shape, left_shape))
+
+    # Check if the specialized method is present.
+    func = getattr(val, '_apply_mixture_', None)
```
Strilanc: Rewrite this method in the style of `cirq.apply_unitary`, i.e. an explicit list of strategies which is iterated over until one succeeds with each strategy being a method that is self-contained. So the general progression would be:

1. Validate input.
2. Make strategies list.
3. Iterate over strategies trying to get a result.
4. Fail or return default value.

cirq/protocols/apply_mixture_protocol.py
```diff
@@ -0,0 +1,309 @@
...
+                        "auxiliary_buffer{}. This type violates the contract "
+                        "that appears in apply_mixture's documentation.".format(
+                            type(val), buf_num_str))
+
+            assert result is not args.auxiliary_buffer0, err_str('0')
+            assert result is not args.auxiliary_buffer1, err_str('1')
+            return result
+
+    # Possibly use `apply_unitary`.
+    result = _apply_unitary(val, args)
```
Strilanc: You also need a decompose based strategy.

Because the decompose here will be redundant with the decompose inside apply unitary, we should consider inlining some of the apply unitary logic. Specifically, two of the strategies will be "look for `_apply_unitary_` and try to use it" as well as "look for `_unitary_` and try to use it".

cirq/protocols/apply_mixture_protocol.py
```diff
@@ -0,0 +1,309 @@
...
+        return None
+    right_args = ApplyUnitaryArgs(target_tensor=np.conjugate(left_result),
+                                  available_buffer=args.auxiliary_buffer0,
+                                  axes=args.right_axes)
+    right_result = apply_unitary(val, right_args)
+    np.conjugate(right_result, out=right_result)
+    return right_result
+
+
+def _apply_unitary_fallback(
```
Strilanc: The naming here is confusing. What's the actual conceptual distinction between apply unitary and apply unitary fallback?

cirq/protocols/apply_mixture_protocol_test.py
```diff
@@ -0,0 +1,303 @@
...
+
+import cirq
+
+
+def make_buffers(shape, dtype):
+    return (np.empty(shape, dtype=dtype), np.empty(shape, dtype=dtype),
+            np.empty(shape, dtype=dtype))
+
+
+def apply_mixture(val,
```
Strilanc: This needs a better name, and type annotations. Maybe it should be refactored into a "applying X to Y produces result Z" testing method?

cirq/protocols/apply_mixture_protocol.py
```diff
@@ -0,0 +1,309 @@
...
+            call, and vice versa).
+
+            The receiving object is also permitted to allocate a new
+            numpy.ndarray and return that as its result.
+        """
+
+
+def apply_mixture(val: Any,
+                  args: ApplyMixtureArgs,
+                  default: TDefault = RaiseTypeErrorIfNotProvided
```
Strilanc: We haven't been property consistent about this everywhere, but force keyword args on the default value.

cirq/protocols/apply_mixture_protocol_test.py
```diff
@@ -0,0 +1,303 @@
...
+        def _apply_unitary_(self, args: cirq.ApplyMixtureArgs):
+            return NotImplemented
+
+    for val in (HasUnitary(), HasUnitaryButReturnsNotImplemented()):
+        result = apply_mixture(val,
+                               rho,
+                               left_axes=[1],
+                               right_axes=[3],
+                               assert_result_is_out_buf=False)
+        print(result)
```
Strilanc: debug code

cirq/protocols/apply_mixture_protocol.py
MichaelBroughton: fixed

cirq/protocols/apply_mixture_protocol.py
MichaelBroughton: Outlined the procedure, let me know what you think

cirq/protocols/apply_mixture_protocol.py
MichaelBroughton: the protocols.apply_unitary attempt here tries to use the decompose funcitonality. Am I missing something here ?

cirq/protocols/apply_mixture_protocol.py
MichaelBroughton: changed

cirq/protocols/apply_mixture_protocol.py
MichaelBroughton: done

cirq/protocols/apply_mixture_protocol_test.py
MichaelBroughton: removed

cirq/protocols/apply_mixture_protocol_test.py
MichaelBroughton: This also happens in the density matrix channel apply tests as well I'm totally up for doing it but I think it should be in a separate PR.

cirq/protocols/apply_mixture_protocol.py
```diff
@@ -0,0 +1,368 @@
...
+    If `val` does not support `apply_unitary` returns None.
+    """
+    left_args = ApplyUnitaryArgs(target_tensor=args.target_tensor,
+                                 available_buffer=args.auxiliary_buffer0,
+                                 axes=args.left_axes)
+    left_result = apply_unitary(val, left_args, None)
+    if left_result is None:
+        return None
+
+    if not is_dm:
```
Strilanc: Expand to `is_density_matrix` everywhere

cirq/protocols/apply_mixture_protocol_test.py
```diff
@@ -0,0 +1,330 @@
...
+    class HasMixture():
+
+        def _mixture_(self):
+            return ((0.5, np.eye(2, dtype=np.complex128)), (0.5, cirq.X))
+
+    rho = np.copy(x)
+    result = apply_mixture(HasMixture(),
+                           rho, [0], [1],
+                           assert_result_is_out_buf=True)
+    np.testing.assert_almost_equal(result, x)
```
Strilanc: Every call to `apply_mixture` is followed by an assertion about what the output is equal to. Fold the output almost equal check into the method and rename it to `assert_apply_mixture_returns`.

cirq/protocols/apply_mixture_protocol_test.py
```diff
@@ -0,0 +1,330 @@
...
+
+    class ApplyUnitaryNotImplemeneted:
+
+        def _apply_unitary_(self, args: cirq.ApplyUnitaryArgs):
+            return NotImplemented
+
+    rho = np.ones((2, 2, 2, 2), dtype=np.complex128)
+    out_buf, aux_buf0, aux_buf1 = make_buffers((2, 2, 2, 2), dtype=rho.dtype)
+
+    with pytest.raises(TypeError):
```
Strilanc: Include a `match="..."` argument for all `pytest.raises`, to avoid false negatives.

cirq/protocols/apply_mixture_protocol.py
MichaelBroughton: fixed

cirq/protocols/apply_mixture_protocol_test.py
MichaelBroughton: changed

cirq/protocols/apply_mixture_protocol_test.py
MichaelBroughton: added

PR #2366, opened 2019-10-17
XEB fidelity estimator based on normalized HOG score

viathor:


viathor:
Tests failed due to #2368. The relevant test in fidelity_estimation_test.py does pass.

viathor:
@yzniu This is the estimator we discussed.

cirq/experiments/fidelity_estimation.py
```diff
@@ -99,6 +99,41 @@ def log_xeb_fidelity_from_probabilities(
...
+
+    See `linear_xeb_fidelity_from_probabilities` for the assumptions made
+    by this estimator.
+
+    The mean of this estimator is the true fidelity f and the variance is
+
+        (1/log(2)^2 - f^2) / M
+
+    where f is the fidelity and M the number of observations, equal to
+    len(probabilities). This is always worse than log XEB (see above).
```
kevinsung: If it's always worse, why would someone want to use it?

viathor: One use-case I encountered recently: plotting various estimators for comparison. A reason to include HOG-based estimates on that plot was that HOG test is fairly well-known (and variance was negligible for all of them anyway).

cirq/experiments/fidelity_estimation.py
```diff
@@ -99,6 +99,41 @@ def log_xeb_fidelity_from_probabilities(
             np.mean(np.log(probabilities)))
 
 
+def hog_score_xeb_fidelity_from_probabilities(
+        hilbert_space_dimension: int,
+        probabilities: Sequence[float],
+) -> float:
+    """XEB fidelity estimator based on normalized HOG score.
```
Strilanc: Add a reference that defines the HOG score.

PR #2370, opened 2019-10-17
Escape unicode characters when producing latex

### LINKED ISSUE #338: Escape unicode characters when producing latex
Currently, the qcircuit to latex code passes through unicode characters like ↺ ↻. LaTeX may not be happy about this. We should replace the unicode characters with some that LaTeX is happy with. Or perhaps throw an error?

bt3gl:
Checked with `pylint` and tested with `pytest`.

Fixes #338


cirq/contrib/qcircuit/qcircuit_pdf.py
```diff
@@ -50,6 +50,10 @@ def circuit_to_pdf_using_qcircuit_via_tex(circuit: circuits.Circuit,
     doc = Document(documentclass=documentclass, document_options='dvips')
     doc.packages.append(Package('amsmath'))
     doc.packages.append(Package('qcircuit'))
+    try:
+        tex.encode('ascii')
+    except UnicodeEncodeError:
+        doc.preamble.append(Package('inputenc', options=['utf8']))
```
Strilanc: Let's just add this unconditionally. (Is there a downside?) It will make the behavior of the code more consistent, and make it easier to cover the code in tests.

docs/install.md
```diff
@@ -83,6 +83,14 @@ That way you have control over when a breaking change affects you.
     python -m pip install cirq[contrib]
     ```
 
+    Install system dependencies that pip can't handle.
+
+    ```bash
+    brew cask install mactex
```
Strilanc: nice

cirq/contrib/qcircuit/qcircuit_pdf.py
bt3gl: Done! I agree this is cleaner, the only reason why I added that was because you folks mentioned it in the ticket.

docs/development.md
```diff
@@ -174,7 +174,7 @@ Here is an example docstring:
 
 ```
 def some_method(a: int, b: str) -> float:
-    r"""One line summary of method.
+    """One line summary of method.
```
Strilanc: That `r` was needed! There are backslashes in the doc string. At the moment the docs are going lose the `\\` because it will turn into a `\`, so the matrix will draw wrong.

bt3gl: Gotcha! Sorry about that, fixed!

PR #2374, opened 2019-10-17
Add py.typed for type checkers

### LINKED ISSUE #2373: Cirq should ship a py.typed file
[PEP 561](https://www.python.org/dev/peps/pep-0561/) says that any packages that ship with type information should have a py.typed file in their package. Otherwise, type checkers like mypy can't find Cirq. (FWIW I just did `touch ~/.virtualenvs/.../cirq/py.typed`, and then mypy type-checks the file correctly.)

Other than that, Cirq seems pretty awesome so far :ok_hand:.

refi64:
Works from local testing.

Fixes #2373

### DIFF
--- a/cirq/py.typed
+++ b/cirq/py.typed
@@ -1,0 +1,1 @@
+# Marker file for PEP 561. The cirq package uses inline types.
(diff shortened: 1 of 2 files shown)


Strilanc:
Thanks for the contribution! A nice small change with a big quality of life impact. After you add the comment that Victory mentioned, I can merge it in.

refi64:
Whoops, now I added the description and rebased on master.

Issue #2375, opened 2019-10-17
Replace (int, float, complex) and permutations with numbers.Number


kevinsung:
Instances of np.int64 are not instances of int, leading to unhandled cases. For instance,
```
import cirq
import numpy as np

a = np.arange(1)
print(isinstance(a[0], int))

q = cirq.LineQubit(0)
p = cirq.X(q)

print(0 * p)
print(a[0] * p)
```
gives
```
False
0j*X(0)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-17-72d730548027> in <module>
      9 
     10 print(0 * p)
---> 11 print(a[0] * p)

TypeError: operand type(s) all returned NotImplemented from __array_ufunc__(<ufunc 'multiply'>, '__call__', 0, cirq.X.on(cirq.LineQubit(0))): 'int64', 'SingleQubitPauliStringGateOperation'
```

Strilanc:
Urgh, this is probably all over the code base.

kevinsung:
Indeed, it is a `numbers.Number`. That is the way to go then.

c-poole:
In many instances it may be more appropriate to check against `numbers.Integral`, `numbers.Real`, and `numbers.Complex` to verify that the code will behave as expected w.r.t. integer vs. floating point arithmetic or other similar concerns.

c-poole:
For the specific problem given by the example test, replacing our number checks with these other checks won't fix the problem. We also need to add cases for multiplication and for each other operation we want to support to the referenced method. Otherwise, numpy will have no idea what to do with our Cirq objects and the same error message will result. For these other operations, we will need to cast the numpy numbers into python numbers to avoid an infinite recursive loop. Adding support for numpy data types in dunder methods will have similar complexity anywhere else in the code base we want to include them. Similar complications also exists for sympy symbols that are numbers. 

https://github.com/quantumlib/Cirq/blob/c572eb7fd98be8390858c0edf9a865a89e33940b/cirq/ops/pauli_string.py#L527-L531

mpharrigan:
Unfortunately numpy 1.20 type hints doesn't seem to like `numbers.Complex` as coercible to numpy array.

daxfohl:
Should it be `numbers.Complex` instead of `numbers.Number`? The latter could ostensibly include quaternions and whatever else.

PR #2376, opened 2019-10-18
Define cirq.STATE_VECTOR_LIKE

Strilanc:
- Add list-of-qid-values to the supported STATE_VECTOR_LIKE types
- Add tensor-of-amplitudes to the supported STATE_VECTOR_LIKE types
- Add non-numpy sequences to supported STATE_VECTOR_LIKE types
- Infer values-vs-tensor-vs-vector using shape, fail when ambiguous (can only occur for non-standard shapes involving qudits of dimension 1)
- Make `num_qubits` argument of `to_valid_state_vector` optional (if `qid_shape` specified)


cirq/sim/wave_function.py
```diff
@@ -15,14 +15,26 @@
...
 from cirq.sim import simulator
 
+if TYPE_CHECKING:
+    import cirq
+
+STATE_VECTOR_LIKE = Union[
+    # Full big-endian computational basis state index.
+    int,
+    # Per-qudit computational basis values.
+    Sequence[Union[int, bool]],
```
kevinsung: Since bools are ints, we can just say int instead of the union.

cirq/sim/wave_function.py
```diff
@@ -15,14 +15,26 @@
...
+    # Explicit state vector or state tensor.
+    np.ndarray, Sequence[Union[int, float, complex]],]
```
cduck: Nit:
```suggestion
    np.ndarray, Sequence[Union[int, float, complex]]]
```

cirq/sim/wave_function.py
Strilanc:
```
In [1]: isinstance(bool, int)                                                                                                   
Out[1]: False
```

cirq/sim/wave_function.py
```diff
@@ -345,36 +357,137 @@ def to_valid_state_vector(
...
+                                         dtype=dtype,
+                                         atol=atol)
+    return tensor.reshape(tensor.size)  # Flatten.
+
+
+def _state_like_to_state_tensor(*, state_like: 'cirq.STATE_VECTOR_LIKE',
+                                qid_shape: Tuple[int, ...],
+                                dtype: Type[np.number], atol) -> np.ndarray:
+
+    if isinstance(state_like, int):
```
cduck: See #2375.

cirq/sim/wave_function.py
```diff
@@ -345,36 +357,137 @@ def to_valid_state_vector(
...
+        if len(qid_shape) == prod:
+            if (not isinstance(state_like, np.ndarray) or
+                    state_like.dtype.kind != 'c'):
+                raise ValueError(
+                    'Because len(qid_shape) == product(qid_shape), it is '
+                    'ambiguous whether or not the given `state_like` is a '
+                    'state vector or a list of computational basis values for '
+                    'the qudits. In this situation you are required to pass '
+                    'in a state vector that is a numpy array with a complex '
+                    'dtype.')
```
cduck: This is worrying.  As a user, I would it expect it to work for all input sizes if it works for most input sizes.

cirq/sim/wave_function.py
cduck:
```
>>> isinstance(True, int)
True
```

cirq/sim/wave_function.py
```diff
@@ -345,36 +357,137 @@ def to_valid_state_vector(
...
+        q = qid_shape[i]
+        if not 0 <= s < q:
+            raise ValueError(
+                f'Qudit value {s} at index {i} is out of bounds for '
+                f'qudit dimension {q}.\n'
+                f'\n'
+                f'qid_shape={qid_shape!r}\n'
+                f'state={state!r}\n')
+
+    if state.dtype.kind[0] not in '?bBiu':
```
cduck: Is there a better way to check this?

cirq/sim/wave_function.py
Strilanc: Hahaha, I made a dumb mistake and @cduck showed me the error in my ways. `isinstance(False, int)` returns true.

Strilanc: I do think that there's some user benefit to listing bool explicitly, but done.

cirq/sim/wave_function.py
Strilanc: Done.

cirq/sim/wave_function.py
Strilanc: In this case I really do want it to be an integer and not a generic number such as a float. I suppose numpy ints would also be acceptable, though.

cirq/sim/wave_function.py
Strilanc: Because `2**n > n` for all `n >= 0`, you never see this unless you have qudits with dimension 1. I think this is a strange enough corner case that we don't have to ensure it works perfectly.

We could reduce the number of relevant cases by checking if the array has non-integer entries in it (in which case it must be a state vector) or if it has integers summing to more than one (in which case it must be a qudit value list). The only truly ambiguous case is something like `shape=(2, 2, 1, 1), value=[1, 0, 0, 0]` where the value is one-hot and has an ambiguous length.

cirq/sim/wave_function.py
Strilanc: I'm honestly not sure. Multiply it by 0, add 0.1, and see if it's still 0?

cduck:
```
>>> issubclass(np.int8, np.integer)
True
>>> issubclass(np.float64, np.integer)
False
```
https://stackoverflow.com/questions/934616/how-do-i-find-out-if-a-numpy-array-contains-integers

c-poole: Checking against `numbers.Integral` should verify that the variable obeys integer arithmetic.
```
>>> issubclass(int, numbers.Integral)
True
>>> issubclass(bool, numbers.Integral)
True
>>> issubclass(np.int8, numbers.Integral)
True
>>> issubclass(np.int16, numbers.Integral)
True
>>> issubclass(np.bool, numbers.Integral)
True
>>> issubclass(float, numbers.Integral)
False
>>> issubclass(np.float64, numbers.Integral)
False
```

Strilanc: There seems to be something tricky going on. Although `np.bool` is integral, the dtype type of an array initialized to have dtype `np.bool` is not:

```
>>> issubclass(np.bool, numbers.Integral)                                                                                  
True
```

```
>>> issubclass(np.array([1, 2, 3], dtype=np.bool).dtype.type, numbers.Integral)                                            
False
```

```
>>> issubclass(np.array([1, 2, 3], dtype=np.int32).dtype.type, numbers.Integral)                                            
True
```

c-poole: The trickery appears to be that when you ask for an array of `np.bool` you don't get an array of `np.bool`. Instead you get an array of `np.bool_` which inherits from `np.generic` instead of `np.int`. 
```
>>> test_array = np.array([1, 2, 3], dtype=np.bool)
>>> test_array.dtype.type == np.bool
False
>>> second_test.dtype.type
<class 'numpy.bool_'>
>>> test_array.dtype.type.__bases__
(<class 'numpy.generic'>,)
>>> np.bool.__bases__
(<class 'int'>,)
>>> test_array.dtype.type == np.bool_
True
```

c-poole: `np.bool_`s do not behave like numbers with respect to arithmetic since they do not inherit from `np.number`. I don't know of a way for a numpy array of boolean values to have a dtype other than `np.bool_`. Whenever you index such an array, you will get a `np.bool_` object. In order to make a numpy array give you python `bool`s it seems that you need to use the `item` function. 

You will probably need to add a check for `np.bool_` in addition to the `numbers.Integral` check.
`if not (isinstance(state.dtype, np.bool_) or issubclass(state.dtype, numbers.Integral)):`

If the array has at least one element, an equivalent check would be given by the below code since the `item` function would return a python `bool` which is a `numbers.Integral`.
`if not isinstance(state.item(0), numbers.Integral):` 

 As written, if someone passed in an empty array for `state` and an empty tuple for `qid_shape`, the output of this function will be a `Scalar` of type `dtype`. Is this desired? 

```
_qudit_values_to_state_tensor(state=np.empty(shape=(0),dtype=np.int64),
                                                 qid_shape=tuple(), dtype=np.int64)
array(1)
_qudit_values_to_state_tensor(state=np.empty(shape=(0),dtype=np.int64),
                                                 qid_shape=tuple(), dtype=np.bool)
array(True)
```

cirq/sim/wave_function.py
mpharrigan: yeah these type annotations are for the user so there's no harm in being explicit

cirq/circuits/circuit.py
```diff
@@ -1410,7 +1410,7 @@ def unitary(self,
 
     def final_wavefunction(
             self,
-            initial_state: Union[int, np.ndarray] = 0,
+            initial_state: 'cirq.STATE_VECTOR_LIKE' = 0,
```
mpharrigan: I mean ... this really exacerbates the "wavefunction vs. state vector" terminology question

cirq/sim/wave_function.py
mpharrigan: tfw your qudits have dimension 1

cirq/circuits/circuit.py
Strilanc: You have a point.

Issue #2259, opened 2019-10-03
Replace sympy-based parser in circuit_from_quirk_url with ported parser from Quirk


Strilanc:
Basically this would involve copy-pasting the code from `src/math/FormulaParser.js` in Quirk and translating it into python. The goal is functional equivalence (i.e. bug parity).

This is a good idea because sympy's parser has a lot of exploitable properties, e.g. you can easily cause it to take hours to parse a string.

PR #2377, opened 2019-10-19
Replace sympy-based parser in circuit_from_quirk_url with ported parser from Quirk

Strilanc:
Fixes https://github.com/quantumlib/Cirq/issues/2259


cirq/contrib/quirk/cells/parse.py
```diff
@@ -12,121 +12,376 @@
...
+        self.func = func
+        self.weight = weight
+
+
+_HangingToken = Union[_ResolvedToken, str, _CustomQuirkOperationToken]
+
+
+def _translate_token(token_id: str,
+                     token_map: Mapping[str, _HangingToken]) -> _HangingToken:
+    if re.match(r'[0-9]+(\.[0-9]+)?', token_id):
```
viathor: Note that `re.match` matches the regex *at the beginning* of text, i.e. `re.match('a', 'ab')` is considered a match (see also [documentation](https://docs.python.org/3/library/re.html)). Therefore, for defense-in-depth I'd append `$` to the regex to force exact match:

```python
if re.match(r'[0-9]+(\.[0-9]+)?$', token_id):
```

This way someone reading the code doesn't have to wonder whether `float(token_id)` raises `ValueError` or not.

cirq/contrib/quirk/cells/parse.py
```diff
@@ -12,121 +12,376 @@
...
+    token_map = {**PARSE_COMPLEX_TOKEN_MAP_RAD, 't': sympy.Symbol('t')}
+    result = _parse_formula_using_token_map(formula, token_map)
+
+    if isinstance(result, sympy.Basic):
+        if result.free_symbols:
+            return result
+        result = complex(result)
+
+    if isinstance(result, complex):
+        if abs(np.imag(result)) > 1e-8:
```
viathor: nit: pass in tolerance for consistency with rest of cirq

Strilanc: Acknowledged, but that would just be a "be inconsistent with Quirk" parameter.

cirq/contrib/quirk/cells/parse.py
Strilanc: Acknowledged, but I'd prefer to keep it closer to the Quirk implementation.

cirq/contrib/quirk/cells/parse.py
```diff
@@ -12,121 +12,375 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import re
-from typing import Any, Union
+from typing import (List, TypeVar, Callable, Any, Iterator, Iterable, Union,
+                    Dict, Optional, cast, SupportsFloat, Mapping)
```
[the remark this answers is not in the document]
Strilanc: Done.

I think I'm going to start pushing back on these unless we write a tool to do it for us. I don't ever read the typing import lists in a way where it matters if they are alphabetized.

PR #2379, opened 2019-10-19
Add quirk parsing support for initial states

Strilanc:


cirq/contrib/quirk/url_to_circuit.py
```diff
@@ -223,6 +225,33 @@ def map_qubit(qubit: 'cirq.Qid') -> 'cirq.Qid':
...
 
+def _init_ops(data) -> 'cirq.OP_TREE':
+    if 'init' not in data:
+        return []
+    init = data['init']
+    if not isinstance(init, List):
+        raise ValueError(f'Circuit JSON init must be a list but was {init!r}.')
+    init_ops = []
+    for i in range(len(init)):
+        state = init[i]
```
viathor: More idiomatic alternatives:

```python
for i, state in enumerate(init):
    qubit = devices.LineQubit(i)
```

or

```python
for state, qubit in zip(init, devices.LinQubit.range(len(init))):
```

Issue #2356, opened 2019-10-15
Qubits shouldn't be mutable


mpharrigan:
```python
import cirq
q1 = cirq.LineQubit(1)
circuit = cirq.Circuit(cirq.X(q1))
print(circuit)
q1.x = 2
print(circuit)
```

```
1: ───X───
2: ───X───
```

Strilanc:
Agreed.

LineQubit and GridQubit should have their values hidden behind property methods.

dabacon:
A new way to remap qubits!  (j/k).  Yes to immutable.

mpharrigan:
We can also consider using dataclasses with `frozen=True`, which offers a stronger form of immutability

bt3gl:
Could you please assign this issue to me? (#2385)

fedimser:
Can you please assign this to me?
@vtomole @Strilanc

PR #2385, opened 2019-10-20
Qubits shouldn't be mutable

bt3gl:
Resolves # #2356 

Uses `dataclasses` with `frozen=True` to make LineQubit, GridQubit and NamedQubit immutable.

### NamedQubit

```
import cirq
q1 = cirq.NamedQubit('awesome_qubit')
circuit = cirq.Circuit(cirq.X(q1))
print(circuit)
q1.name = 'not_awesome'
print(circuit)
```

returns

```
dataclasses.FrozenInstanceError: cannot assign to field 'name'
```

Pylint & pytest pass.

### GridQubit 
```
import cirq
q1 = cirq.GridQubit(1, 1)
circuit = cirq.Circuit(cirq.X(q1))
print(circuit)
q1.row = 2
print(circuit)
```

returns

```
dataclasses.FrozenInstanceError: cannot assign to field 'row'
```

Pylint & pytest pass.

### LineQubit

```
import cirq
q1 = cirq.LineQubit(1)
circuit = cirq.Circuit(cirq.X(q1))
print(circuit)
q1.x = 2
print(circuit)
```

returns

```
dataclasses.FrozenInstanceError: cannot assign to field 'x'
```

Pylint pass.
Pytest fails 2 tests due to the different hashes comparing with  `LineQid`, since this class is still mutable (they are different objects):

```
AssertionError: Items in the same group produced different hashes. Example: hash(cirq.LineQubit(0)) is 3430018387555 but hash(cirq.LineQid(0, dimension=2)) is 8601707190144160275.
```

If we try to remediate this, trying to add  `@dataclass(frozen=True)` to `LineQid`, we have a problem with `dimension` and the base class `_BaseLineQid`. 

So my question is: **are the eq hash tests important (do they need to be the same object?) so should we refactor this class to handle `dataclasses`?**

Another question: **should we handle `dataclasses.FrozenInstanceError`, and if yes, what's the best way?**


cirq/devices/line_qubit.py
```diff
@@ -184,6 +186,8 @@ class LineQubit(_BaseLineQid):
         cirq.LineQubit(1)
     """
 
+    x: int
```
mpharrigan: I don't think this is right considering the inheritance here.

cirq/ops/named_qubit.py
```diff
@@ -23,13 +26,10 @@ class NamedQubit(raw_types.Qid):
...
     """
-
-    def __init__(self, name: str) -> None:
-        self._name = name
-        self._comp_key = _pad_digits(name)
+    _name: str
 
     def _comparison_key(self):
-        return self._comp_key
+        return _pad_digits(self._name)
```
mpharrigan: caching this value was likely done for performance reasons. Do you have a sense for the slowdown one would expect from this change?

cirq/devices/line_qubit.py
cduck: I think all three classes here should be dataclasses with `_BaseLineQid` defining the `x` field and `LineQid` defining the `dimension` field.

cirq/ops/named_qubit.py
Strilanc: Yes, this caching should not be removed.

vtomole:
Ping @bt3gl

bt3gl:
> Ping @bt3gl

Hi @vtomole I am sorry this PR got carried way. I will be working on the feedbacks and cleaning up everything this Saturday

bt3gl:
Sorry, folks, I didn't have the chance to work on this in a timely manner so I will be just closing the PR.

PR #2388, opened 2019-10-21
Quantum volume: num_circuits vs repetitions

mpharrigan:
"repetitions" refer to running the same circuit a bunch of times.
"num_circuits" is how many circuits you use to estimate quantum
volume / HOG.

### DIFF
--- a/cirq/contrib/quantum_volume/quantum_volume.py
+++ b/cirq/contrib/quantum_volume/quantum_volume.py
@@ -104,7 +104,7 @@ def sample_heavy_set(circuit: cirq.Circuit,
     Args:
         circuit: The circuit to sample.
         heavy_set: The previously-computed heavy set for the given circuit.
-        repetitions: The number of runs to sample the circuit.
+        repetitions: The number of times to sample the circuit.
         sampler: The sampler to run on the given circuit.
         mapping: An optional mapping from compiled qubits to original qubits,
             to maintain the ordering between the model and compiled circuits.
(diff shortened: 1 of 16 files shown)


cirq/contrib/quantum_volume/quantum_volume.py
```diff
@@ -95,7 +95,7 @@ def compute_heavy_set(circuit: cirq.Circuit) -> List[int]:
 def sample_heavy_set(circuit: cirq.Circuit,
                      heavy_set: List[int],
                      *,
-                     repetitions=10000,
+                     repetitions=10_000,
```
Strilanc: I like it.

cirq/contrib/quantum_volume/quantum_volume.py
```diff
@@ -207,23 +207,23 @@ def prepare_circuits(
...
+        num_circuits: The number of circuits to create.
         random_state: A way to seed the RandomState.
 
     Returns:
         A list of tuples where the first element is a generated model
         circuit and the second element is the heavy set for that circuit.
     """
     circuits = []
-    for repetition in range(num_repetitions):
+    for repetition in range(num_circuits):
```
Strilanc: Rename the iteration variable too.

cirq/contrib/quantum_volume/quantum_volume.py
mpharrigan: hashtag feature I floated in python-dev irc room a couple years ago and everyone shouted at me for saying it's dumb before implementing it a couple releases later

cirq/contrib/quantum_volume/quantum_volume.py
mpharrigan: good catch. updated docstrings as well

mpharrigan:
cc @KevinVillela

KevinVillela:
Thanks for doing this. I believe you may also need to update https://github.com/quantumlib/Cirq/blob/master/examples/advanced/quantum_volume.py.

mpharrigan:
updated that file too

examples/advanced/quantum_volume.py
```diff
@@ -68,7 +68,7 @@ def parse_arguments(args):
                         type=int,
                         help='Seed for the Random Number Generator.')
     parser.add_argument(
-        '--num_repetitions',
+        '--num_circuits',
         default=100,
         type=int,
         help='The number of times to run the circuit on the quantum computer.'
```
KevinVillela: This doc should now say something like "The number of circuits to generate". Or, to not reproduce the message, "Pass-through to the Quantum Volume module".

PR #2389, opened 2019-10-21
Move files from cirq.google into subpackages

dstrain115:
- Create cirq.google.devices for Google devices
- Create cirq.google.optimizers for gate compilers


cirq/devices/grid_qubit_test.py
```diff
@@ -79,7 +79,7 @@ def test_diagram():
 -----KL-----
 """
     assert (cirq.GridQubit.from_diagram(s) ==
-            cirq.google.known_devices._parse_device(s)[0])
+            cirq.google.devices.known_devices._parse_device(s)[0])
```
maffoo: Consider exporting this `_parse_device` function on `cirq.google.devices`. We could rename it to `parse_device_ascii` or something like that.

cirq/google/devices/known_devices_test.py
```diff
@@ -19,11 +19,12 @@ def test_foxtail_qubits():
     for i in range(0, 2):
         for j in range(0, 11):
             expected_qubits.append(cirq.GridQubit(i, j))
-    assert set(expected_qubits) == cirq.google.known_devices.Foxtail.qubits
+    assert set(
+        expected_qubits) == cirq.google.devices.known_devices.Foxtail.qubits
```
maffoo: Can just use `cirq.google.Foxtail` here.

cirq/protocols/json_test.py
```diff
@@ -259,10 +259,6 @@ def test_fail_to_resolve():
     cirq.Z,
     'S':
     cirq.S,
-    'SerializableDevice':
-    cirq.google.SerializableDevice.from_proto(
-        proto=cirq.google.known_devices.FOXTAIL_PROTO,
-        gate_set=cirq.google.XMON),
```
maffoo: Why was this removed?

cirq/devices/grid_qubit_test.py
dstrain115: I just decided to change this test instead.  I am not sure why it was depending on this function.  It seemed like a bad test.

cirq/google/devices/known_devices_test.py
dstrain115: Done.

cirq/protocols/json_test.py
dstrain115: Ironically, SerializableDevice cannot yet be serialized into JSON.  I'll add that functionality later.

PR #2390, opened 2019-10-22
Change int, float, complex isinstance to use numbers module

c-poole:
Makes the following substitutions across the repository as a first step in adding operational support for numpy and sympy number types. See #2375 

`isinstance(x, int)` -> `isinstance(x, numbers.Integral)`
`isinstance(x, (int, float))` -> `isinstance(x, numbers.Real)`
`isinstance(x, (int, float, complex))` -> `isinstance(x, numbers.Complex)`
`isinstance(x, numbers.Number)` -> `isinstance(x, numbers.Complex)`

No tests failed with this change but mypy is complaining on a number of the changes.


vtomole:
Ping @c-poole

c-poole:
I will be revisiting this as well as my other PRs soon. Hopefully joining the meetings again starting in February.

c-poole:
I've made another run through of the repo to update this. I haven't looked at pylint, mypy and formatting so I'm expecting failures there. Some notes.

Now that the `Decimal` library is in use, I had to make some exemptions in the replacement rule
`isinstance(x, numbers.Number) -> isinstance(x, numbers.Complex)` because a `Decimal` is a `numbers.Number` but not a `numbers.Complex`. This is because certain operations between `Decimal` and other number types aren't supported and thus `Decimal` does not even meet the requirements of `numbers.Real`. 

The following two rules were also implemented in addition to those in the opening post.
`isinstance(x, float)` -> `isinstance(x, numbers.Real) and not isinstance(x, numbers.Integral)`
This change allows for numpy floats but excludes integer types 

`isinstance(x, complex)` -> `isinstance(x, numbers.Complex)`

c-poole:
I'm electing to close this since I'm no longer convinced that this repo wide substitution is beneficial. We should instead focus on supporting the non-python-native types we want wherever we find we want to use them.

PR #2396, opened 2019-10-23
Add Sycamore gate and gate_set to cirq

dstrain115:
- This a gate similar to the one used in the Google Nature paper:
https://www.nature.com/articles/s41586-019-1666-5


cirq/google/__init__.py
```diff
@@ -45,7 +45,9 @@
 )
 
 from cirq.google.gate_sets import (
-    XMON,)
+    XMON,
+    SYCAMORE_GATESET,
```
dabacon: Since Sycamore is the chip, and the gate is SYC, maybe we should call this SYC_GATESET?

cirq/google/gate_sets.py
```diff
@@ -12,116 +12,163 @@
...
-                op_serializer.SerializingArg(
-                    serialized_name='key',
-                    serialized_type=str,
-                    gate_getter=protocols.measurement_key),
-                op_serializer.SerializingArg(serialized_name='invert_mask',
-                                             serialized_type=List[bool],
-                                             gate_getter='invert_mask')
-            ])
+        _SYCAMORE_SERIALIZER, *SINGLE_QUBIT_SERIALIZERS,
+        *SINGLE_QUBIT_HALF_PI_SERIALIZERS, MEASUREMENT_SERIALIZER
```
dabacon: nit: personally I'd add a comma on the end to make it format one per line.  But I have a tall screen :)

cirq/ops/sycamore_gate.py
```diff
@@ -0,0 +1,40 @@
+"""Gates for Google's Sycamore architecture."""
```
dabacon: missing copyright

dabacon: Suggest: A gate that is naturally easy on Google's Sycamore chip

cirq/ops/sycamore_gate.py
```diff
@@ -0,0 +1,40 @@
...
+import numpy as np
+
+from cirq import ops, protocols
+
+if TYPE_CHECKING:
+    import cirq
+
+
+class SycamoreGate(ops.FSimGate):
+    """Sycamore two-qubit gate is a FSimGate(π/2, π/6) gate.
```
dabacon: Sycamore -> The Sycamore Gate

cirq/ops/sycamore_gate.py
```diff
@@ -0,0 +1,40 @@
...
+
+    The unitary of this matrix is
+
+        [[1, 0, 0, 0],
+         [0, 0, -1j, 0],
+         [0, -1j, 0, 0],
+         [0, 0, 0, exp(- 1j * π/6)]]
+    """
```
dabacon: Comment that this gate is a gate that is "easy" to get on the Sycamore chip and add a link to the nature publication

cirq/ops/sycamore_gate_test.py
```diff
@@ -0,0 +1,49 @@
+import numpy as np
```
dabacon: missing copyright

cirq/ops/sycamore_gate.py
```diff
@@ -0,0 +1,40 @@
...
+from cirq import ops, protocols
+
+if TYPE_CHECKING:
+    import cirq
+
+
+class SycamoreGate(ops.FSimGate):
+    """Sycamore two-qubit gate is a FSimGate(π/2, π/6) gate.
+
+    The unitary of this matrix is
```
kevinsung: "The unitary of this gate"

cirq/ops/sycamore_gate.py
```diff
@@ -0,0 +1,40 @@
...
+         [0, 0, -1j, 0],
+         [0, -1j, 0, 0],
+         [0, 0, 0, exp(- 1j * π/6)]]
+    """
+
+    def __init__(self):
+        super().__init__(theta=np.pi / 2, phi=np.pi / 6)
+
+    def __repr__(self) -> str:
+        return 'cirq.SycamoreGate()'
```
kevinsung: `cirq.SYC` also works

cirq/google/gate_sets.py
dstrain115: Done.

cirq/ops/sycamore_gate_test.py
dstrain115: Done.

cirq/google/gate_sets.py
```diff
@@ -12,116 +12,167 @@
...
-    ])
+        _SYC_DESERIALIZER,
+        *SINGLE_QUBIT_DESERIALIZERS,
+        *SINGLE_QUBIT_HALF_PI_DESERIALIZERS,
+        MEASUREMENT_DESERIALIZER,
+    ],
+)
+
+# The xmon gate set.
+XMON: serializable_gate_set.SerializableGateSet = (
```
mpharrigan: why is this type annotation needed? why isn't it needed for syc gate set

cirq/google/gate_sets.py
```diff
@@ -12,116 +12,167 @@
...
+                ]),
+            op_serializer.GateOpSerializer(gate_type=ops.ZPowGate,
+                                           serialized_gate_id='exp_z',
+                                           args=[
+                                               op_serializer.SerializingArg(
+                                                   serialized_name='half_turns',
+                                                   serialized_type=float,
+                                                   gate_getter='exponent')
+                                           ]),
+            op_serializer.GateOpSerializer(
```
mpharrigan: why doesn't this use the `common_serializers`. Otherwise they don't seem so common

cirq/google/gate_sets.py
dstrain115: I am not totally sure why it is needed.  I think something in the import order changed and mypy started complaining about not knowing the type of XMON, which is used as the default in cirq/google/engine.py.  This is the only way I could get it to stop.

cirq/google/gate_sets.py
dstrain115: Sigh.  XMON uses a different naming convention than those serializers.  "xy" vs "exp_w", etc, and those strings are used all over the place in tests.

How about this?  I will change it over in a follow-up PR, but I think doing so in this PR would conflate the change, since XMON hasn't really changed.

mpharrigan: for sure

mpharrigan: common_gates.py:

```python
def get_single_qubit_gates(what_do_you_call_a_phasedxpow: str):
  if what_do_you_call_a_phasedxpow not in ['xy', 'exp_w']:
    raise ValueError("https://xkcd.com/927/")
  ...
```

cirq/ops/sycamore_gate.py
```diff
@@ -0,0 +1,58 @@
...
+
+    The unitary of this gate is
+
+        [[1, 0, 0, 0],
+         [0, 0, -1j, 0],
+         [0, -1j, 0, 0],
+         [0, 0, 0, exp(- 1j * π/6)]]
+
+    This gate can be performed on the Sycamore chip by Google and
+    was used to demonstrate quantum supremacy as can be seen in
```
dabacon: This should probably not say that this gate was used, since we used variations of this gate (differ across the device).  So possibly could be

"This gate can be performed on the Google's Sycamore chip and
is close to the gates that were used to demonstrate quantum 
supremacy used in this paper: 
https://www.nature.com/articles/s41586-019-1666-5."

PR #2397, opened 2019-10-23
Add optional gate attribute on Operation

maffoo:
This allows us to remove the `op_gate_isinstance` and `op_gate_of_type` helper methods, in favor of just doing regular `isinstance` checks. Unfortunately, this change touched a lot of files which also seems to be triggering a lot of lint issues. I'm still working through those. Note: I'd like to get this in before the 0.6.0 release so the `op_gate_*` functions don't go out with a release.


mpharrigan:
I know `op_gate_of_type` hasn't made it to a "release" but they've been there for quite a while and since it's been such a long time since a release I suspect there is code which uses these utilities. I strongly suspect I have a bunch of code that uses these utilities

cirq/ops/pauli_string.py
```diff
@@ -856,8 +857,8 @@ def inline_times_pauli_string_like(self,
         if isinstance(contents, PauliString):
             # Note: cirq.X/Y/Z(qubit) are PauliString instances.
             self.inline_times_pauli_string(contents)
-        elif gate_operation.op_gate_isinstance(contents,
-                                               common_gates.IdentityGate):
+        elif (isinstance(contents, raw_types.Operation) and
+              isinstance(contents.gate, common_gates.IdentityGate)):
```
cduck: The point of `op_gate_isinstance` was to remove the need for a double `isinstance` check.  With the new Operation.gate property, this check should just be `isinstance(op.gate, ...)`.

cduck:
@maffoo `op_gate_of_type` actually was in the 0.5 release: `https://github.com/quantumlib/Cirq/blob/v0.5.0/cirq/ops/gate_operation.py#L188`

cirq/ops/pauli_string.py
maffoo: `contents` is a `PAULI_STRING_LIKE`, not an `Operation`, which is why we need the double check.

maffoo:
Thanks for the heads up, @mpharrigan and @cduck. I've added the helper methods back and annotated them as being deprecated.

maffoo:
Ugh, need to add back imports and tests as well...

maffoo:
@Strilanc, @cduck, this should be ready to have another look.

mpharrigan:
In practice, does anyone use operations that aren't `GateOperation`s?

cduck:
> In practice, does anyone use operations that aren't `GateOperation`s?

`class PauliString(raw_types.Operation):`

cirq/ion/ion_device.py
```diff
@@ -186,7 +185,7 @@ def _value_equality_values_(self):
 def _verify_unique_measurement_keys(operations: Iterable[ops.Operation]):
     seen: Set[str] = set()
     for op in operations:
-        meas = ops.op_gate_of_type(op, ops.MeasurementGate)
+        meas = op.gate if isinstance(op.gate, ops.MeasurementGate) else None
         if meas:
```
cduck: The inline if can be merged with the if statement.

cirq/ops/dense_pauli_string.py
```diff
@@ -511,7 +511,10 @@ def _str_to_pauli_mask(text: str) -> np.ndarray:
 
 
 def _attempt_value_to_pauli_index(v: Any) -> Optional[Tuple[int, int]]:
-    pauli_gate = gate_operation.op_gate_of_type(v, pauli_gates.Pauli)
+    if not isinstance(v, raw_types.Operation):
+        return None
+
+    pauli_gate = v.gate if isinstance(v.gate, pauli_gates.Pauli) else None
     if pauli_gate is None:
```
cduck: Can be simplified to only use one if.

cirq/ops/parallel_gate_operation.py
```diff
@@ -87,7 +87,7 @@ def _apply_unitary_(self, args: 'protocols.ApplyUnitaryArgs'
         """
         if not protocols.has_unitary(self.gate):
             return NotImplemented
-        return protocols.apply_unitaries((self.gate(q) for q in self.qubits),
+        return protocols.apply_unitaries((self._gate(q) for q in self.qubits),
```
cduck: Why change this?  The rest of the class still uses `self.gate`?

cirq/sim/density_matrix_simulator.py
```diff
@@ -276,7 +275,8 @@ def keep(potential_op: ops.Operation) -> bool:
             for op in channel_ops_and_measurements:
                 indices = [qubit_map[qubit] for qubit in op.qubits]
                 # TODO: support more general measurements.
-                meas = ops.op_gate_of_type(op, ops.MeasurementGate)
+                meas = op.gate if isinstance(op.gate,
+                                             ops.MeasurementGate) else None
                 if meas:
```
cduck:
```
    meas = op.gate
```

If we could use python 3.8:
```
if isinstance(meas := op.gate, ops.MeasurementGate):
```

cirq/sim/sparse_simulator.py
```diff
@@ -313,7 +313,7 @@ def _simulate_unitary(self, op: ops.Operation, data: _StateAndBuffer,
     def _simulate_reset(self, op: ops.Operation, data: _StateAndBuffer,
                         indices: List[int]) -> None:
         """Simulate an op that is a reset to the |0> state."""
-        reset = ops.op_gate_of_type(op, ops.ResetChannel)
+        reset = op.gate if isinstance(op.gate, ops.ResetChannel) else None
         if reset:
```
cduck: Nit: Another double if.

cirq/sim/sparse_simulator.py
```diff
@@ -330,7 +330,7 @@ def _simulate_measurement(self, op: ops.Operation, data: _StateAndBuffer,
                               measurements: Dict[str, List[int]],
                               num_qubits: int) -> None:
         """Simulate an op that is a measurement in the computational basis."""
-        meas = ops.op_gate_of_type(op, ops.MeasurementGate)
+        meas = op.gate if isinstance(op.gate, ops.MeasurementGate) else None
         # TODO: support measurement outside computational basis.
         if meas:
```
cduck: Nit: Another double if.

docs/api.rst
```diff
@@ -195,7 +195,6 @@ Circuits, Operations, and Moments.
     InsertStrategy
     Moment
     moment_by_moment_schedule
-    op_gate_of_type
```
cduck: Do we normally remove deprecated functions from the API?

cirq/ops/pauli_string.py
cduck: I see.  I liked `op_gate_isinstance`.

cirq/circuits/circuit.py
```diff
@@ -1961,9 +1961,9 @@ def on_stuck(bad_op):
 
 
 def _decompose_measurement_inversions(op: 'cirq.Operation') -> 'cirq.OP_TREE':
-    gate = ops.op_gate_of_type(op, ops.MeasurementGate)
-    if gate:
-        return [ops.X(q) for q, b in zip(op.qubits, gate.invert_mask) if b]
+    if (isinstance(op, ops.Operation) and
+            isinstance(op.gate, ops.MeasurementGate)):
```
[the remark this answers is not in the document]
maffoo: Turns out the type annotation on `op` is a lie and we actually do need the first check. The decompose protocol calls this with values of other types, including generator instances which then cause errors here if we just access `op.gate`. I haven't dug into the decompose code to figure out exactly what's going on.

cirq/circuits/qasm_output.py
```diff
@@ -148,12 +148,11 @@ def __init__(self,
                  header: str = '',
                  precision: int = 10,
                  version: str = '2.0') -> None:
-        self.operations = tuple(ops.flatten_op_tree(operations))
+        self.operations = tuple(ops.flatten_to_ops(operations))
```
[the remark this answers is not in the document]
maffoo: Not deprecated yet, but `flatten_op_tree` has a return type of `Iterator[Union[Operation, Moment]]` because it can optionally preserve moments; `flatten_to_ops` doesn't have this option so it returns `Iterator[Operation]` which is what we want here anyway and keeps mypy happy.

cirq/ion/ion_device.py
maffoo: Agreed, looks much better to put isinstance in the if condition. Changed throughout. (This sort of thing will be even nicer with the walrus operator in 3.8.)

cirq/ops/dense_pauli_string.py
maffoo: fixed.

cirq/ops/parallel_gate_operation.py
maffoo: I was getting mypy errors here, but can't reproduce them now. Reverted.

cirq/sim/density_matrix_simulator.py
maffoo: yup; I'm very excited for the walrus operator :-)

cirq/sim/sparse_simulator.py
maffoo: fixed.

cirq/sim/sparse_simulator.py
maffoo: fixed.

docs/api.rst
maffoo: I don't know the policy here. @Strilanc?

maffoo: I don't know the policy here. @Strilanc?

cirq/ops/parallel_gate_operation.py
maffoo: Oh right, it's pylint that is not happy because it sees that `self.gate` is a property and doesn't like that we're calling it. Fixed by using `self.gate.on(q)`.

Issue #2322, opened 2019-10-11
Abbreviate PhasedXPowGate circuit representation as PhX


mpharrigan:
`PhasedX(0.123)^0.456` makes the circuit diagrams really long. I propose changing it to `PhX(0.123)^0.456`

dabacon:
I wonder how confused people are by "phased"?  The one reason I don't like it is that it is a basis dependent concept, i.e. it means phased in Z basis.  RotXY, for example, would generalize to other two pauli rotations nicely, RotYZ, RotXZ.  On the other hand for qubits that are clocks it really is a natural gate.

dabacon:
But I do like `PhX`.

mpharrigan:
PhasedXPowGate certainly was not obvious to me at first but yeah I'm mainly concerned about circuit diagram width :)

Strilanc:
Sounds good to me.

jitendrs:
can you please assign it to me. I will have a look.

jitendrs:
Hi,
I can see reference to PhasedX has been spread across below listed file. which includes docstring as well.
let me know if it's ok changed all of them.


cirq/circuits/circuit_test.py
cirq/google/convert_to_xmon_gates_test.py
cirq/neutral_atoms/convert_to_neutral_atom_gates_test.py
cirq/ops/moment.py
cirq/ops/phased_x_gate.py
cirq/ops/phased_x_gate_test.py
cirq/optimizers/decompositions.py
cirq/optimizers/eject_phased_paulis.py
cirq/optimizers/merge_single_qubit_gates.py
cirq/optimizers/two_qubit_decompositions_test.py
cirq/protocols/decompose_protocol.py

Strilanc:
Where do you see PhasedX in `moment.py`?

You should only change `PhasedX` to `PhX` in strings that are circuit diagrams or circuit diagram info.

jitendrs:
ahh apologies,  moment.py wasn't part of it. I will follow your instructions.

PR #2404, opened 2019-10-25
 PhasedXPowGate circuit representation as PhX #2322

jitendrs:
Changes has been made to rename PhasedX to PhX in strings that are circuit diagrams or circuit diagram info.

Fixes: https://github.com/quantumlib/Cirq/issues/2322


mpharrigan:
sweeeet

Issue #1940, opened 2019-08-13
Proposal: operator tags


dabacon:
When building a circuit is often useful to have some extra information about certain parts of the circuit.  For example you may want to preserve a certain part of the circuit while performing a transformation (optimization pass).  There isn't a great way to do this right now, but we were able to tag our operations, then we could use the tag to do this.

Tagging could also be used for diagrams.

Tagging might also be important when differentiating different "types" of gates for the same unitary.  For example sometimes exp(iZt) can be compiled out of a circuit, but sometimes you want it to stick around.

dabacon:
Thoughts?

dabacon:
Amusingly got a request for this today from someone else.

dstrain115:
It seems like it would be low effort and generally useful for a variety of tasks.

It seems like there are a fair amount of different tasks where we have a generally accepted behavior or optimization pass, except we want to ignore one specific use case, so this would be a good way to accomplish all of those kinds of tasks.

dabacon:
Another question is whether Gates should also have tags.  Any gate with a tag transfers it to its operator when applied using `on`

dabacon:
Another question is about equality.  I guess gates and ops with different tags need to not be different.  I was going to lazily add the tags, but if we want that to work then there is a bit more work.

mpharrigan:
How much structure do you want tags to have? A string field? a list of strings? something more like an enum?

dabacon:
list of strings was my thought

dstrain115:
This has been done in #2407

95-martin-orion:
> This has been done in #2407

Correction: #2407 was dropped in favor of #2670, which has been submitted.

PR #2407, opened 2019-10-25
Add operator tags

dstrain115:
- Lite Implementation of the proposal in #1940 to introduce gate tags.
- Adds Tags as a class that you can add a string value annotation to or subclass.  The Gate class will support adding these tags and listing them.

I wanted to send a preliminary PR out first to see if you agree on the implementation approach.  

If so, I will add some of the finer details (making sure gates serialize their tags to JSON, possibly adding tags to GateOperation, etc).


cirq/ops/raw_types.py
```diff
@@ -153,6 +153,38 @@ def _json_dict_(self):
...
 
 
+class Tag:
+    """A wrapper around string that can be user to differentiate gates.
+
+    This can be instantiated using a string value or sub-classed to
+    differentiate or tag specific instances of gates.
+
+    For instance,
+    cirq.X().add_tag(Tag('do_not_touch')
```
mpharrigan: missing close paren

cirq/ops/raw_types.py
```diff
@@ -206,6 +238,13 @@ def wrap_in_linear_combination(
         from cirq.ops import linear_combinations
         return linear_combinations.LinearCombinationOfGates({self: coefficient})
 
+    def add_tag(self, tag: Tag):
+        self._tags = getattr(self, '_tags', [])
```
mpharrigan: using `getattr` seems really odd and will likely confuse my dear pycharm. Why not initialize to `[]` in `__init__`?

cirq/ops/raw_types.py
```diff
@@ -206,6 +238,13 @@ def wrap_in_linear_combination(
...
+        self._tags.append(tag)
```
mpharrigan: `return self` for chaining:

```python
cirq.Circuit(
  cirq.X(q).add_tag('woo'),
  cirq.Z(q)
)

File: cirq/ops/raw_types.py [hunk 4]
```diff
@@ -206,6 +238,13 @@ def wrap_in_linear_combination(
         from cirq.ops import linear_combinations
         return linear_combinations.LinearCombinationOfGates({self: coefficient})
 
+    def add_tag(self, tag: Tag):
```
mpharrigan: consider accepting bare strings and taking care of wrapping in `Tag`

cirq/ops/raw_types.py
dstrain115: Done.  Also changed this so it renders correctly as example code.  There were a bunch of other issues in this code as well.

cirq/ops/raw_types.py
dstrain115: Because Gate is an abstract type and doesn't define __init__.  We could do this, but then every other Gate implementation has to call super.__init__() which it doesn't right now.

cirq/ops/raw_types.py

cirq/ops/raw_types.py

dstrain115:
> Do we need `Tag`? why not just strings?
> Did you consider something more structured like having a metadata dict?

Good ideas.  I put forward this PR partially motivate these ideas.  There are several ways to implement this.  I chose this approach for the following reasons:

1) Having a container around strings allows both untyped tagging (by specifying the string) as well as typed tagging (by subclassing the class).
2) Creating a class allows people to extend the tagging functionality as they fit.

My intuition is that having a dictionary is over-kill and also enables too many "magic strings" in a format that does not allow checking.  (i.e. was the string "should_compile" or "shouldCompile"?)

However, I am open to other ideas.

cirq/ops/raw_types.py
```diff
@@ -153,6 +153,48 @@ def _json_dict_(self):
...
+
+    def __init__(self, value: Optional[str] = None):
+        self.value = value
+
+    def __eq__(self, other):
+        return (isinstance(other, self.__class__) and
+                isinstance(self, other.__class__) and self.value == other.value)
+
+    def __str__(self):
+        return f'Tag(\'{self.value}\')'
```
mpharrigan: It's generally more pythonic to use `"` for the outer quotes so you don't have to escape the `'`s

cirq/ops/raw_types.py
mpharrigan: oooo

mpharrigan: this might actually be worth putting in as a comment to prevent someone from wasting time in the future

mpharrigan:
> I put forward this PR partially motivate these ideas
Excellent. I'm a big fan of centering these design discussions around concrete proposals 

> My intuition is that having a dictionary is over-kill

It's possible someone could subclass `Tag` and make a very rich data structure
```python
@dataclass
class CompilationOptions(Tag):
  should_compile: bool
  original_gate: cirq.Gate
  echo_sequence: cirq.Circuit
```

Do you see this as abuse/antipattern or working as intended (I don't know myself whether this would be good or bad)

cirq/ops/raw_types.py

dstrain115:
> Do you see this as abuse/antipattern or working as intended (I don't know myself whether this would be good or bad)

Seems like it is working as intended to me, though it would certainly be one of the more advanced use-cases of this proposal.  Then, the compiler in question could see if the gate has any tags and if they match type "CompilerOptions".  If so, those options would override behavior in the intended way.  Note that this could not be easily accomplished by having strings or a dict of strings.

dabacon:
My worry about this approach is that it breaks equality.  So operators with different tags are equal.  Since a tag is an attribute of an operator this worries me.  

Also the Tag class feels heavy handed: why not just use strings?  Maybe I just missed it but I don't see the added benefit of this type.

dstrain115:
> My worry about this approach is that it breaks equality. So operators with different tags are equal. Since a tag is an attribute of an operator this worries me.

Let me see if I can fix equality of gates.  I wasn't actually sure if gates with different tags should be the same or different.  I'll see what the best way to implement this.

> Also the Tag class feels heavy handed: why not just use strings? Maybe I just missed it but I don't see the added benefit of this type.

See my conversation with Matt above.  I think using the class allows for greater flexibility and sub-classing.  I think it would generally better to sub-class rather than to use a magic string value.  For instance, having a class named PhysicalZTag makes it impossible to screw up since the interpreter would immediately blow up if you misspell it.  Otherwise, imagine tagging your gate with "physical_Z" but it doesn't work right and silently ignores your tag.  How long would it take to realize that you are supposed to tag it with "physical_z" instead?  

Matt also has a fairly compelling example for a tag that allows a compiler to override default behavior for a specific gate in various ways that would require more complicated information than a string.  Allowing sub-classing would allow for the later addition of meta-data that is not apparent to me right now.

cirq/ops/raw_types.py
```diff
@@ -153,6 +153,48 @@ def _json_dict_(self):
         return protocols.obj_to_dict_helper(self, ['qubit', 'dimension'])
 
 
+class Tag:
+    """A wrapper around string that can be user to differentiate gates.
```
Strilanc: used

cirq/ops/raw_types.py
```diff
@@ -153,6 +153,48 @@ def _json_dict_(self):
         return protocols.obj_to_dict_helper(self, ['qubit', 'dimension'])
 
 
+class Tag:
```
Strilanc: What value does this class add over simply having the tags be an `Any`?

dstrain115:
> 2\. What are the user stories that this helps with? This is a rather far reaching change; we should have concrete problems that it solves and an understanding of why it solves it well.

I am not sure about the other questions, but there are at least two distinct user cases that would benefit from this:
1) Denoting physical Z from virtual Z and other hardware-specific gates that are not relevant to the circuit operation, but very relevant to the hardware operation.
2) Denoting exceptions for optimizers (for instance, do not combine single qubit rotations if they are spin echoes, random benchmarking, or noise modeling)

cirq/ops/raw_types.py
dstrain115: I guess you could technically do it that way, but that would be untyped and very error-prone.  It would not help the user figure out how to use the feature and you could do all sorts of non-sensical stuff, like add a circuit as a tag.

cirq/ops/raw_types.py

dstrain115:
This branch is closed.  We have discussed a new approach and I will send a new PR out for that one instead.

PR #2410, opened 2019-10-25
Add property to get at measurement keys in TrialResult

zchen088:


zchen088:
Hmmm, actually I'm wondering what the ordering should be. Is there any way to sort the measurement keys "top left to bottom right" as seen in the circuit diagram?

cirq/study/trial_result.py
```diff
@@ -115,6 +115,10 @@ def data(self) -> pd.DataFrame:
             self._data = pd.DataFrame(converted_dict, dtype=np.int64)
         return self._data
 
+    @property
+    def keys(self) -> List[str]:
```
Strilanc: Rename to `measurement_keys`

vtomole:
@zchen088 Ping

PR #2420, opened 2019-10-28
Add 2-qubit gate decomp based on randomly filling the Weyl chamber to contrib

dkafri:
Current (experimental) version of 2 qubit gate compiler. Given a single base gate A, the compiler tabulates many products of the form A k A  and A k A k A, where k are randomly generated 1-local unitaries. After many samples (and assuming an appropriate gate A), this fills up the Weyl chamber with a set of unitaries such that any target unitary U_t is epsilon close to a tabulated unitary.


cirq/contrib/two_qubit_gates/example.py
```diff
@@ -0,0 +1,61 @@
+from time import time
+
+import numpy as np
+from matplotlib import pyplot as plt
+
+# pylint: disable=line-too-long
```
KevinVillela: Instead of disabling the linter, I feel like we should probably just let yapf format this. It will probably do something like:
from cirq.contrib.two_qubit_gates.math_utils import (
  random_two_qubit_unitaries_and_kak_vecs,
  unitary_entanglement_fidelity,
)

cirq/contrib/two_qubit_gates/example.py
```diff
@@ -0,0 +1,61 @@
+from time import time
```
KevinVillela: This file could use a top-level comment describing what it does and how to run it.

cirq/contrib/two_qubit_gates/example.py
```diff
@@ -0,0 +1,61 @@
...
+# pylint: disable=line-too-long
+from cirq import FSimGate, unitary
+from cirq.contrib.two_qubit_gates.gate_compilation import gate_product_tabulation
+from cirq.contrib.two_qubit_gates.math_utils import random_two_qubit_unitaries_and_kak_vecs, unitary_entanglement_fidelity
+
+# pylint: enable=line-too-long
+
+
+def main():
+    # coverage: ignore
```
KevinVillela: Optional: MIght be a good idea to have a test that just runs through this and makes sure it doesn't err. Otherwise, this could stop working and we wouldn't know. I think if you do that, you can probably remove these coverage: ignore pragmas.

cirq/contrib/two_qubit_gates/gate_compilation_test.py
```diff
@@ -0,0 +1,59 @@
...
+import pytest
+
+# pylint: disable=line-too-long
+from cirq import unitary, FSimGate
+from cirq.contrib.two_qubit_gates.gate_compilation import gate_product_tabulation
+from cirq.contrib.two_qubit_gates.math_utils import random_two_qubit_unitaries_and_kak_vecs, unitary_entanglement_fidelity
+
+# pylint: enable-line-too-long
+
+# coverage: ignore
```
KevinVillela: I don't think this should be required for tests, they should already be ignored for coverage.

cirq/contrib/two_qubit_gates/math_utils.py
```diff
@@ -0,0 +1,292 @@
...
+    out = _single_qubit_unitary(theta, phi_d, phi_o)
+
+    if sample_phase:
+        # coverage: ignore
+        global_phase = np.exp(1j * TWO_PI * np.random.rand(number))
+        np.einsum('t,tab->tab', global_phase, out, out=out)
+    return out
+
+
+def vector_kron(first: np.ndarray, second: np.ndarray) -> np.ndarray:
```
KevinVillela: Can we have linalg/decompositions_test.py reference this so it isn't duplicated?

cirq/contrib/two_qubit_gates/math_utils.py
```diff
@@ -0,0 +1,292 @@
...
+    """Vectorized implementation of kron for square matrices."""
+    s_0, s_1 = first.shape[-2:], second.shape[-2:]
+    assert s_0[0] == s_0[1]
+    assert s_1[0] == s_1[1]
+    out = np.einsum('...ab,...cd->...acbd', first, second)
+    s_v = out.shape[:-4]
+    return out.reshape(s_v + (s_0[0] * s_1[0],) * 2)
+
+
+_MAGIC = np.array([[1, 0, 0, 1j], [0, 1j, 1, 0], [0, 1j, -1, 0], [1, 0, 0, -1j]
```
KevinVillela: Are these used anywhere?

cirq/contrib/two_qubit_gates/math_utils.py
```diff
@@ -0,0 +1,292 @@
...
+    pi_4 = np.pi / 4
+
+    x_inside = np.logical_and(0 <= xp, xp <= pi_4)
+
+    y_inside = np.logical_and(0 <= yp, yp <= pi_4)
+    y_inside = np.logical_and(y_inside, xp >= yp)
+
+    z_inside = np.abs(zp) <= yp
+
+    out = np.logical_and(x_inside, y_inside)
```
KevinVillela: return np.logical_and.reduce((x_inside, y_inside, z_inside))

cirq/contrib/two_qubit_gates/math_utils.py
```diff
@@ -0,0 +1,292 @@
...
+
+    Args:
+        spacing: Euclidean distance between neighboring KAK vectors.
+
+    Returns:
+        np.ndarray of shape (N,3) corresponding to the points in the Weyl
+        chamber.
+    """
+    if spacing < 1e-3:  # memory required ~ 1 GB
+        # coverage: ignore
```
KevinVillela: This actually seems like a good thing to test :)

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,328 @@
...
+
+    Returns:
+        A GateTabulation object used to compile new two-qubit gates from
+        products of the base gate with 1-local unitaries.
+    """
+    assert 1 / 2 > max_infidelity > 0
+    spacing = np.sqrt(max_infidelity / 3)
+    mesh_points = weyl_chamber_mesh(spacing)
+
+    num_samples = mesh_points.shape[0] * 50
```
KevinVillela: Can we add a comment about where "50" comes from here? Adjusting this would seem to have an effect on how dense the Weyl chamber is, do you think it makes sense to allow it to be passed in as an argument, or is `max_infidelity` enough?

cirq/contrib/two_qubit_gates/math_utils.py
```diff
@@ -0,0 +1,292 @@
...
+    (...,M,M) (as long as both shapes can be broadcast together).
+
+    Returns:
+        The entanglement fidelity between the two unitaries. For inputs with
+        shape (...,M,M), the output has shape (...).
+
+    """
+    U_actual = np.asarray(U_actual)
+    U_ideal = np.asarray(U_ideal)
+    if U_actual.shape[-1] != U_actual.shape[-2]:
```
KevinVillela: assert U_actual.shape[-1] != U_actual.shape[-2], "Inputs' trailing dimensions must be equal (square).")

cirq/contrib/two_qubit_gates/example.py
dkafri: yapf doesn't seem to fix this...

Strilanc: Put parens around the import list and then yapf will fix it. Don't disable the line too long check.

cirq/contrib/two_qubit_gates/example.py
Strilanc: Example code should still be covered. It's okay if the test just runs the method and confirms it doesn't throw an error, without checking what it actually did.

cirq/contrib/two_qubit_gates/example.py
```diff
@@ -0,0 +1,61 @@
...
+             '--',
+             label='Maximum tabulation infidelity')
+    plt.xlabel('Compiled gate infidelity vs target')
+    plt.ylabel('Counts')
+    plt.legend()
+    plt.title(f'Base FSim(theta={theta:.4f}, phi={phi:.4f})')
+    plt.show()
+
+
+# coverage: ignore
```
Strilanc: This will become unnecessary after you have a test file executing this file's main method.

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,328 @@
...
+with a product A k A."""
+from functools import reduce
+from typing import Tuple, Dict, Sequence, List
+from warnings import warn
+
+import numpy as np
+import attr
+from cirq import kak_decomposition, kak_vector
+# pylint: disable=line-too-long
+from cirq.contrib.two_qubit_gates.math_utils import KAK_vector_infidelity, vector_kron, weyl_chamber_mesh, random_qubit_unitary, KAK_vector_to_unitary
```
Strilanc: Parens around impor tlist, remove pylint disable.

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,328 @@
...
+
+# pylint: enable=line-too-long
+
+_SingleQubitGatePair = Tuple[np.ndarray, np.ndarray]
+
+
+@attr.s(auto_attribs=True)
+class GateTabulation(object):
```
Strilanc: Remove redundant `(object)`.

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,328 @@
...
+    """A 2-qubit gate compiler based on precomputing/tabulating gate products.
+
+    """
+    base_gate: np.ndarray  # Base two qubit gate. (4x4 unitary)
+    # Sequence of KAK vectors, ideally "dense" in the Weyl chamber. Shape (N,3).
+    KAK_vecs: np.ndarray
```
Strilanc: Nit: lower case

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,328 @@
...
+    target_decomp = kak_decomposition(target)
+    base_decomp = kak_decomposition(base)
+
+    # From the KAK decomposition, we have
+    # kLt At kRt = kL kLb Ab KRb kR
+    # If At=Ab, we can solve for the right and left local unitaries as
+    # kLt = kL kLb --> kL = kLt kLb^\dagger
+    # kRt = kRb kR --> kR = kRb\dagger kRt
+
+    kLt0, kLt1 = target_decomp.single_qubit_operations_after
```
Strilanc: Nit: these names are not very clear.

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,328 @@
...
+
+
+def _outer_locals_for_unitary(
+        target: np.ndarray, base: np.ndarray
+) -> Tuple[Tuple[_SingleQubitGatePair, _SingleQubitGatePair], np.ndarray]:
+    """Local unitaries mapping between locally equivalent 2-local unitaries.
+
+    Finds the left and right 1-local unitaries k_L, k_R such that
+
+    U_target = k_L @ U_base @ k_R
```
Strilanc: Do we need the `k_`? Why not just `L` and `R`?

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,328 @@
...
+        out = [outer_locals[0]]
+        out.extend(self.single_qubit_gates[nearest_ind])
+        out.append(outer_locals[1])
+
+        return out, actual, success
+
+
+def _outer_locals_for_unitary(
+        target: np.ndarray, base: np.ndarray
+) -> Tuple[Tuple[_SingleQubitGatePair, _SingleQubitGatePair], np.ndarray]:
```
Strilanc: Flatten into a triplet instead of a nested pair. Consider using a `NamedTuple`.

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,328 @@
...
+    return (k_R, k_L), actual
+
+
+def _tabulate_KAK_vectors(
+        tabulation: Dict[int, Tuple[_SingleQubitGatePair, ...]],
+        base_gate: np.ndarray,
+        max_dist: float,
+        KAK_mesh: np.ndarray,
+        *local_unitary_pairs: _SingleQubitGatePair,
+) -> Tuple[List[np.ndarray], List[Tuple[_SingleQubitGatePair, ...]]]:
```
Strilanc: I notice you're using names to refer to the output values in the return doc. Change this return type into a `NamedTuple`.

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,328 @@
...
+    kak_vecs.extend(out[0])
+    sq_cycles.extend(out[1])
+
+    # Will be used later for getting missing KAK vectors.
+    kak_vecs_single = np.array(kak_vecs[1:])
+    sq_cycles_single = list(sq_cycles[1:])
+
+    if verbose:
+        # coverage: ignore
+        print(f'fraction satisfied with 2 gates'
```
Strilanc: This kind of output is not really appropriate within a method inside of a large framework. E.g. if you called the quantum service and saw output about 'fraction satisfied with 2 gates' you'd be pretty confused by that. This information should be e.g. a method on the type returned by the method. The fact that you had to ignore it for coverage purposes is indicative of this.

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,328 @@
...
+    # Compare the KAK vector of any of those products to the already tabulated
+    # KAK vectors from single products of the form
+    # base_gate k0 base_gate.
+    # If they are close, then base_gate^\dagger k A  ~ base_gate k0 base_gate
+    # So we may compute the outer local unitaries kL, kR such that
+    #    base_gate^\dagger k A = kL base_gate k0 base_gate kR
+    #    A = k^\dagger base_gate kL base_gate k0 base_gate kR
+    #    the single-qubit unitary kL is the one we need to get the desired
+    #    KAK vector.
+    base_gate_dag = base_gate.conj().T
```
Strilanc: This method is very long. Can it be split into two logical pieces? Maybe the loop body can be extracted into a subroutine?

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,328 @@
...
+            # Add to the enumeration
+            sq_cycles.append((old_sq_cycle, kRkL[1]))
+            kak_vecs.append(
+                kak_vector(base_gate @ actual, check_preconditions=False))
+        elif include_warnings:
+            # coverage: ignore
+            warn(f'Failed to tabulate a KAK vector near {missing_vec}')
+
+    kak_vecs = np.array(kak_vecs)
+    if verbose:
```
Strilanc: Delete this block.

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,328 @@
...
+    # repeat for double products
+    # Multiply by the same local unitary!
+    out = _tabulate_KAK_vectors(u_locals_for_gate, base_gate, tabulation_cutoff,
+                                mesh_points, (u_locals_0, u_locals_1),
+                                (u_locals_0, u_locals_1))
+
+    kak_vecs.extend(out[0])
+    sq_cycles.extend(out[1])
+
+    if verbose:
```
Strilanc: Delete this block.

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,328 @@
...
+                            include_warnings: bool = True) -> GateTabulation:
+    r"""Generate a GateTabulation for a base two qubit unitary.
+
+    Args:
+        base_gate: The base gate of the tabulation.
+        max_infidelity: Sets the desired density of tabulated product unitaries.
+            The typical nearest neighbor Euclidean spacing (of the KAK vectors)
+            will be on the order of \sqrt(max_infidelity). Thus the number of
+            tabulated points will scale as max_infidelity^{-3/2}.
+        verbose: Whether to print gate tabulation statistics.
```
Strilanc: Remove the verbose parameter. Replace it with methods on `GateTabulation` that provide access to the information that would have been printed.

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,328 @@
...
+            new_product = products[new_ind]
+
+            kRkL, actual = _outer_locals_for_unitary(new_product, base_product)
+            # Add to the enumeration
+            sq_cycles.append((old_sq_cycle, kRkL[1]))
+            kak_vecs.append(
+                kak_vector(base_gate @ actual, check_preconditions=False))
+        elif include_warnings:
+            # coverage: ignore
+            warn(f'Failed to tabulate a KAK vector near {missing_vec}')
```
Strilanc: If the user wants all points to be covered, this should be an error instead of a warning.

Add a method to `GateTabulation` to print any missed vectors. Change `include_warnings` to `allow_missed_points`. If a point is missed and `allow_missed_points` is false, raise an error.

cirq/contrib/two_qubit_gates/gate_compilation_test.py
```diff
@@ -0,0 +1,59 @@
...
+numpy.random.seed(11)  # for determinism
```
Strilanc: Do not affect global state during tests. Especially not at module import time. You have no guarantee that you are not affecting other tests by doing this, nor that other tests will not generate random numbers and break the tests here.

Instead, you should modify the tabulation method to take a `random_state: Optional[Union[np.random.RandomState, int]] = None`. Copy what `cirq.testing.random_circuit` does.

cirq/contrib/two_qubit_gates/math_utils.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+def random_qubit_unitary(number: int = 1,
+                         sample_phase: bool = False) -> np.ndarray:
+    """Random qubit unitary distributed over the Haar measure.
+
+    The implementation is vectorized for speed.
+
+    Args:
+        number: If not 1, an ndarray of shape (number,2,2) is returned.
```
Strilanc: Refactor this into a `shape` parameter. The output shape is always `tuple(shape) + (2, 2)`.

cirq/contrib/two_qubit_gates/math_utils.py
```diff
@@ -0,0 +1,292 @@
...
+            Otherwise a single 2x2 matrix is returned.
+        sample_phase: (Default False) If True, a global phase is also sampled
```
Strilanc: Rename this to `randomize_global_phase`.

cirq/contrib/two_qubit_gates/math_utils.py
Strilanc: This could potentially be merged into `cirq.kron` in the form of an `axes` parameter.

cirq/contrib/two_qubit_gates/math_utils.py
```diff
@@ -0,0 +1,292 @@
+from itertools import permutations
```
Strilanc: Nit: just `import itertools`

cirq/contrib/two_qubit_gates/math_utils.py
```diff
@@ -0,0 +1,292 @@
...
+from typing import Tuple, Union
+
+import numpy as np
+
+TWO_PI = np.pi * 2
```
Strilanc: Nit: not worth a constant for this.

cirq/contrib/two_qubit_gates/math_utils.py
```diff
@@ -0,0 +1,292 @@
...
+_negations[2, (0, 1, 2), (0, 1, 2)] = (-1, 1, -1)
+_negations[3, (0, 1, 2), (0, 1, 2)] = (-1, -1, 1)
+
+_offsets = np.zeros((4, 3))
+_offsets[1, (1, 2)] = np.pi / 2
+_offsets[1, (0, 2)] = np.pi / 2
+_offsets[1, (0, 1)] = np.pi / 2
+
+
+def _kak_equivalent_vectors(kak_vec) -> np.ndarray:
```
Strilanc: Instead of broadcasting single kak vectors into many possible vectors, you should canonicalize incoming vectors into their canonical form. This will reduce space usage.

cirq/contrib/two_qubit_gates/math_utils.py
```diff
@@ -0,0 +1,292 @@
...
+
+    k_diff = k_vec_a - k_vec_b
+
+    out = 1 - np.product(np.cos(k_diff), axis=-1)**2
+    out -= np.product(np.sin(k_diff), axis=-1)**2  # (...,24)
+
+    return out.min(axis=-1)
+
+
+def in_weyl_chamber(xp: np.ndarray, yp: np.ndarray,
```
Strilanc: Why not take the same type that is returned by `kak_vector`, i.e. have the last axis be the length-3 x,y,z axis?

cirq/contrib/two_qubit_gates/math_utils.py
dkafri: I think we do need to consider equivalent vectors, specifically if we are near the k_x = pi/4 surface of the Weyl chamber.

```python
x, y, z = np.pi / 4, 1, 0.5
kak_0 = kak_canonicalize_vector(x, y, z).interaction_coefficients
kak_1 = kak_canonicalize_vector(x - 1e-3, y, z).interaction_coefficients

inf_check_equivalent = KAK_vector_infidelity(kak_0, kak_1, False)
inf_ignore_equivalent = KAK_vector_infidelity(kak_0, kak_1, True)

print(f'infidelity without checking equivalent: {inf_ignore_equivalent}')
print(f'infidelity with checking equivalent: {inf_check_equivalent}')
```
returns:
```
infidelity without checking equivalent: 0.7080737102000556
infidelity with checking equivalent: 9.999996666287814e-07
```

Checking this actually revealed me an error in the code. I should probably write some tests to confirm that `KAK_vector_infidelity` is working properly.  In practice I am only using it with the `ignore_equivalent_vectors=True` option, in which case it only gives an upper bound to the infidelity.

cirq/contrib/two_qubit_gates/gate_compilation_test.py
dkafri: It's also done in fidelity_estimation_test.py and clifford_simulator_test.py

cirq/contrib/two_qubit_gates/example.py
dkafri: Done

cirq/contrib/two_qubit_gates/example.py
dkafri: Done

cirq/contrib/two_qubit_gates/example.py
dkafri: Done

cirq/contrib/two_qubit_gates/gate_compilation_test.py
dkafri: Done

cirq/contrib/two_qubit_gates/math_utils.py
dkafri: Removed.

cirq/contrib/two_qubit_gates/math_utils.py
dkafri: Done

cirq/contrib/two_qubit_gates/math_utils.py
dkafri: Done

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: Done. Added optional argument.

cirq/contrib/two_qubit_gates/math_utils.py
dkafri: Done

cirq/contrib/two_qubit_gates/example.py
dkafri: Done

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: Done

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: Done

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: Done

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: Add a comment, but for brevity I kept variable names unchanged. The comment above also explains the notation.

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: Done

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: Done

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: Done

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: I intend to do this in the future. Currently the algorithm is still under development, as I would like to allow for more than 3 usages of the base gate. I think this refactoring will be more useful once I  think about how to implement this.

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: Done

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: Done

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: Done.

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: Done.

cirq/contrib/two_qubit_gates/gate_compilation_test.py
dkafri: Done.

cirq/contrib/two_qubit_gates/math_utils.py
dkafri: Done.

cirq/contrib/two_qubit_gates/math_utils.py
dkafri: Done.

cirq/contrib/two_qubit_gates/math_utils.py
dkafri: Done.

cirq/contrib/two_qubit_gates/math_utils.py
dkafri: Done.

cirq/contrib/two_qubit_gates/math_utils.py
dkafri: Done.

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: Done.

cirq/contrib/two_qubit_gates/math_utils.py
```diff
@@ -0,0 +1,310 @@
...
+                          ignore_equivalent_vectors: bool = False
+                         ) -> np.ndarray:
+    """Minimum entanglement infidelity between two KAK vectors. """
+    k_vec_a, k_vec_b = np.asarray(k_vec_a), np.asarray(k_vec_b)
+    if ignore_equivalent_vectors:
+        k_diff = k_vec_a - k_vec_b
+        out = 1 - np.product(np.cos(k_diff), axis=-1)**2
+        out -= np.product(np.sin(k_diff), axis=-1)**2
+        return out
+    # coverage: ignore
```
KevinVillela: Per your other comment, +1 to adding tests here :)

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,374 @@
...
+                kept = True
+
+        if kept:
+            kept_kaks.append(vec)
+            kept_cycles.append(cycles_for_gate)
+
+    return _TabulationStepResult(kept_kaks, kept_cycles)
+
+
+def gate_product_tabulation(base_gate: np.ndarray,
```
KevinVillela: nit: I'd consider making the optional arguments named parameters.

cirq/contrib/two_qubit_gates/gate_compilation_test.py
```diff
@@ -0,0 +1,95 @@
...
+
+def test_gate_compilation_missing_points_raises_error():
+    with pytest.raises(ValueError, match='Failed to tabulate a'):
+        gate_product_tabulation(np.eye(4),
+                                0.4,
+                                allow_missed_points=False,
+                                random_state=_rng)
+
+
+def test_random_qubit_unitary_shape():
```
KevinVillela: Similarly, these should go in math_utils_test.py.

dkafri: Done.

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: Done.

cirq/contrib/two_qubit_gates/math_utils.py
dkafri: Done. Also added some documentation.

cirq/contrib/two_qubit_gates/gate_compilation_test.py
```diff
@@ -0,0 +1,95 @@
...
+    result = tabulation.compile_two_qubit_gate(base_gate)
+
+    assert len(result.local_unitaries) == 2
+    assert result.success
+    fidelity = unitary_entanglement_fidelity(result.actual_gate, base_gate)
+    assert fidelity > 0.99999
+
+
+def test_gate_compilation_example():
+    example.main(samples=10, max_infidelity=0.3, verbose=False)
```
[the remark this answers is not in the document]
dkafri: Done.

cirq/contrib/two_qubit_gates/gate_compilation_test.py
```diff
@@ -0,0 +1,95 @@
...
+_random_2Q_unitaries, _ = random_two_qubit_unitaries_and_kak_vecs(100, _rng)
+
+
+@pytest.mark.parametrize('tabulation',
+                         [sycamore_tabulation, sqrt_iswap_tabulation])
+@pytest.mark.parametrize('target', _random_2Q_unitaries)
+def test_gate_compilation_matches_expected_max_infidelity(tabulation, target):
+    result = tabulation.compile_two_qubit_gate(target)
+
+    if result.success:
```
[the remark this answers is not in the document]
dkafri: Good point. For the very loose fidelity bound it looks like it's always successful. In other cases (i.e., a poor chose of base gate) it may only be successful some of the time.

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,374 @@
...
+    for local_cycle in local_cycles[1:]:
+        np.einsum('ab,...bc,...cd', base_gate, local_cycle, prods, out=prods)
+
+    kak_vectors = kak_vector(prods, check_preconditions=False)
+
+    kept_kaks = []
+    kept_cycles = []
+
+    for ind, vec in enumerate(kak_vectors):
+        # dists = KAK_vector_infidelity(vec, KAK_mesh)
```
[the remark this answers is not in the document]
dkafri: Done

cirq/contrib/two_qubit_gates/example.py
```diff
@@ -0,0 +1,114 @@
...
+             ylim,
+             '--',
+             label='Maximum tabulation infidelity')
+    plt.xlabel('Compiled gate infidelity vs target')
+    plt.ylabel('Counts')
+    plt.legend()
+    plt.title(f'Base FSim(theta={theta:.4f}, phi={phi:.4f})')
+    if verbose:
+        # coverage: ignore
+        plt.show()
```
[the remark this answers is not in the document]
dkafri: Noted, thanks!

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,374 @@
...
+
+class _TabulationStepResult(NamedTuple):
+    # Generated KAK vectors that are uniquely close to at least one mesh point.
+    kept_kaks: List[np.ndarray]
+    # The corresponding single qubit unitaries required to obtain the desired
+    # KAK vectors.
+    kept_cycles: List[Tuple[_SingleQubitGatePair, ...]]
+
+
+def _tabulate_kak_vectors(
```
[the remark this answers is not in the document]
dkafri: Done. I also realized the input can be simplified a bit.

cirq/contrib/two_qubit_gates/math_utils.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+_XX = np.zeros((4, 4))
+_XX[(0, 1, 2, 3), (3, 2, 1, 0)] = 1
+_ZZ = np.diag([1, -1, -1, 1])
+_YY = -_XX @ _ZZ
+_kak_gens = np.array([_XX, _YY, _ZZ])
+
+
+def random_two_qubit_unitaries_and_kak_vecs(
```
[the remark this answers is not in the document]
dkafri: You're right. I replaced it with random_special_unitary.

cirq/contrib/two_qubit_gates/math_utils.py
dkafri: Done.

cirq/contrib/two_qubit_gates/gate_compilation.py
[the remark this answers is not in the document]
dkafri: Done.

cirq/contrib/two_qubit_gates/math_utils.py
```diff
@@ -128,27 +130,56 @@ def _kak_equivalent_vectors(kak_vec) -> np.ndarray:
     return np.reshape(out, out.shape[:-4] + (192, 3))
 
 
-def KAK_vector_infidelity(k_vec_a: np.ndarray,
+def kak_vector_infidelity(k_vec_a: np.ndarray,
                           k_vec_b: np.ndarray,
                           ignore_equivalent_vectors: bool = False
                          ) -> np.ndarray:
-    """Minimum entanglement infidelity between two KAK vectors. """
+    r"""The locally invariant infidelity between two KAK vectors.
```
KevinVillela: r?

cirq/contrib/two_qubit_gates/math_utils.py
```diff
@@ -128,27 +130,56 @@ def _kak_equivalent_vectors(kak_vec) -> np.ndarray:
...
+
+    This is the quantity
```
KevinVillela: Missing the end of this sentence :)

cirq/contrib/two_qubit_gates/math_utils.py
dkafri: There is a backslash in the docstring.

cirq/contrib/two_qubit_gates/math_utils.py
dkafri: The sentence continues in the equation below. I added a line-break for clarity.

dkafri:
@Strilanc : Any changes outstanding before the PR can be approved?

cirq/contrib/two_qubit_gates/example.py
Strilanc: It's still there.

Strilanc: It's still there.

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,379 @@
+"""Attempt to tabulate single qubit gates required to generate a target 2Q gate
+with a product A k A."""
+from functools import reduce
+from typing import Tuple, Sequence, List, NamedTuple
+
+import numpy as np
+import attr
```
Strilanc: I don't think we have `attr` listed as a requirement anywhere. Maybe it's an indirect dependency?

I recommend using `@dataclass` instead, as that has been standardized in python 3.7.

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,379 @@
...
+    actual_gate: np.ndarray
+    success: bool
+
+
+@attr.s(auto_attribs=True)
+class GateTabulation:
+    """A 2-qubit gate compiler based on precomputing/tabulating gate products.
+
+    """
+    base_gate: np.ndarray  # Base two qubit gate. (4x4 unitary)
```
Strilanc: Rename to `base_gate_unitary` to disambiguate w.r.t. `cirq.Gate` instances

cirq/contrib/two_qubit_gates/gate_compilation.py
```diff
@@ -0,0 +1,379 @@
...
+        Args:
+            unitary: Unitary (U above) to compile.
+
+        Returns:
+            A TwoQubitGateCompilation object encoding the required local
+            unitaries and resulting product above.
+        """
+        unitary = np.asarray(unitary)
+        kak_vec = kak_vector(unitary, check_preconditions=False)
+        infidelities = kak_vector_infidelity(kak_vec,
```
Strilanc: These seems like a really inefficient way to find the closest point, though it's quite simple which is nice. How fast is it in practice for various max infidelities? Is it usable at max infidelity `0.001`? `0.0001`? The infidelity of the decomposition should be negligible compared to the physical gate error, so 10^{-4} is presumably a good target.

cirq/contrib/two_qubit_gates/gate_compilation_test.py
```diff
@@ -0,0 +1,70 @@
...
+from cirq import unitary, FSimGate, value
+from cirq.contrib.two_qubit_gates.gate_compilation import (
+    gate_product_tabulation)
+from cirq.contrib.two_qubit_gates.math_utils import (
+    unitary_entanglement_fidelity)
+from cirq.testing import random_special_unitary
+
+_rng = value.parse_random_state(11)  # for determinism
+
+sycamore_tabulation = gate_product_tabulation(unitary(
```
Strilanc: How long does this test file take to run? It should be less than ten seconds, ideally less than a second.

cirq/contrib/two_qubit_gates/example_test.py
```diff
@@ -0,0 +1,5 @@
+from cirq.contrib.two_qubit_gates import example
+
+
+def test_gate_compilation_example():
+    example.main(samples=10, max_infidelity=0.3, verbose=False)
```
Strilanc: Nice job reducing the cost of running the example test.

dkafri: Thanks!

cirq/contrib/two_qubit_gates/gate_compilation_test.py
dkafri: It takes about 0.6 seconds on my machine. I can lower the specs further if you like.

dkafri: 0.6 seconds on my local machine. I can lower the specs further if you like.

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: In practice this should scale with the number of mesh points, which is scales as (1/d)^3. Here d is the euclidean lattice spacing, which is \sqrt(max_infidelity)/3.  So the total number of mesh points scales as max_infidelity^{-3/2}. 

In practice the gate tabulation takes significantly longer than the gate compilation, so I can't test this for very small max_infidelity in a reasonable amount of time. In fact, I am seeing that the tabulation scales as max_infidelity^{-2}, which is surprising and I should look into. 

At 0.01, the tabulation time is about 7 seconds and the total compilation time (per gate) is 0.0022 seconds (Sorry I haven't set up a profiler on this branch so I can't get exact timing for this line). If I push max_infidelity down to 0.005, the tabulation time is about 29.5 seconds and the total compilation time is 0.0023 s. Likewise, if I set max_infidelity to 0.004, the tabulation time is about 50 seconds and the time per gate is 0.0024 s. So, worst case, the line you mention takes about 1e-4 for max_infidelity=0.01, and I would guess it takes 1e-4 * (100^(3/2) = 0.1 s for max_infidelity=0.0001. 

From my previous profiling on my own version of the code, I recall that (at least for max_infidelity not too small) the line that takes up a majority of the time is the call to `kak_decomposition`.

Update: I was able to run the example with max_infidelity=0.001. The tabulation took 1900 seconds (~31 minutes) while the compilation took 0.0043 seconds per gate. So I suspect my previous estimate is actually rather pessimistic.

cirq/contrib/two_qubit_gates/example_test.py
dkafri: Thanks!

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: Done

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: Done

cirq/contrib/two_qubit_gates/gate_compilation.py
dkafri: The current implementation will not be very practical for max_infidelity much smaller then 0.01, unless we precompute the tabulation once and save it. That said, in the future we can make the compiler more sophisticated by allowing for local optimization using the tabulated point as an initial guess.

KevinVillela:
Awesome, thanks a ton for doing this! As soon as we get #2445 resolved, I'll try running the QV stuff with this decomposition algorithm. I think it's going to give us some good improvements :)

PR #2426, opened 2019-10-28
Change XMON to use common serializers

dstrain115:
This changes XMON gate set to use the same serializers as everything else.

This changes the serialized ids from exp_w to xy and exp_z to z, etc.

It allows us to compress a lot of the code in gate_sets down to only point to the common serializers.


cirq/google/gate_sets_test.py
```diff
@@ -33,7 +33,7 @@ def test_serialize_exp_w(gate, axis_half_turns, half_turns):
     q = cirq.GridQubit(1, 2)
     expected = {
         'gate': {
-            'id': 'exp_w'
+            'id': 'xy'
```
dabacon: I guess this change is ok because no service is consuming these new v2 protos?

cirq/google/common_serializers.py
```diff
@@ -218,3 +218,20 @@ def _near_mod_2(e, t, atol=1e-8):
                                              value_func=lambda _: 0.5),
         ]),
 ]
+"""CZ Serializer"""
+CZ_SERIALIZER = op_serializer.GateOpSerializer(
```
dabacon: I'm always torn calling this CZ.  CPHASE_SERIALIZER or CZ_POW_SERIALIZER since this allows arbitrary phase?  Otherwise this doesn't follow pattern of SYC_SERIALIZER where SYC is a fixed gate.

cirq/google/common_serializers.py
```diff
@@ -218,3 +218,20 @@ def _near_mod_2(e, t, atol=1e-8):
...
+    gate_type=ops.CZPowGate,
+    serialized_gate_id='cz',
+    args=[
+        op_serializer.SerializingArg(serialized_name='half_turns',
+                                     serialized_type=float,
+                                     gate_getter='exponent')
+    ])
+"""CZ Deserializer"""
```
maffoo: nit: I would suggest using a comment here instead of a string literal, since this isn't actually a docstring for anything. Could make that change in a separate PR since this would affect existing code in this module. Would also be nice to add whitespace between the serializer instances, though maybe the formatter will do that if we convert to comments.

cirq/google/gate_sets_test.py
dstrain115: Correct, nothing should be using XMON that I know of.  It looks like there are some devices using this, but none are hooked up to cirq_server, so this should be fine to make backwards incompatible changes.

cirq/google/gate_sets.py
```diff
@@ -67,112 +71,14 @@ def _near_mod_2pi(e, t, atol=1e-8):
...
+                                              serializers=[
+                                                  *SINGLE_QUBIT_SERIALIZERS,
+                                                  CZ_SERIALIZER,
+                                                  MEASUREMENT_SERIALIZER,
+                                              ],
+                                              deserializers=[
+                                                  *SINGLE_QUBIT_DESERIALIZERS,
+                                                  CZ_DESERIALIZER,
+                                                  MEASUREMENT_DESERIALIZER,
+                                              ]))
```
maffoo: nit: add trailing comma after the last argument to make this format better. Also, is the type annotation needed? If not, can put the constructor on the first line and remove the outer parens.

dstrain115: Done.  

Without the type annotation, I get a bunch of mypy warnings that I couldn't figure out how to get rid of otherwise.

```
cirq/google/engine/engine.py:227: error: Cannot determine type of 'XMON'
cirq/google/engine/engine.py:284: error: Cannot determine type of 'XMON'
cirq/google/engine/engine.py:328: error: Cannot determine type of 'XMON'
cirq/google/engine/engine.py:466: error: Cannot determine type of 'XMON'
cirq/google/engine/engine.py:499: error: Cannot determine type of 'XMON'
```

cirq/google/common_serializers.py

cirq/google/common_serializers.py
dstrain115: Changed to CZ_POW

cirq/google/gate_sets.py
mpharrigan: gate_set should be an attribute on the `Circuit` imo

PR #2428, opened 2019-10-28
Use helper to simplify seeding

dabacon:
This adds a method `prng_from_seed` which is only used on the sim module.  This is only used when the prng is actually needed, making it easier to understand what this is when you read the code where it is used (by doing these conversion higher in the call chain one doesn't quite see why one is doing this).

### DIFF
--- a/cirq/sim/density_matrix_simulator.py
+++ b/cirq/sim/density_matrix_simulator.py
@@ -135,13 +135,7 @@ class DensityMatrixSimulator(simulator.SimulatesSamples,
 
         self._dtype = dtype
         self.noise = devices.NoiseModel.from_noise_model_like(noise)
-
-        if seed is None:
-            self.prng = None
-        elif isinstance(seed, np.random.RandomState):
-            self.prng = seed
-        else:
-            self.prng = np.random.RandomState(seed)
+        self._seed = seed
 
     def _run(self, circuit: circuits.Circuit,
              param_resolver: study.ParamResolver,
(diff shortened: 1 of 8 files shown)


cirq/sim/density_matrix_simulator.py
```diff
@@ -285,7 +279,7 @@ def keep(potential_op: ops.Operation) -> bool:
                             indices,
                             qid_shape=qid_shape,
                             out=state.tensor,
-                            seed=self.prng)
+                            seed=self._seed)
```
kevinsung: If `self._seed` is an int, won't this line cause the same RandomState to be initialized each time it is executed? That's why I believe the correct behavior is to initialize a RandomState during initialization of the simulator and pass that object around; we should never have more than one RandomState be initialized during the course of a single simulation.

cirq/sim/sparse_simulator.py
```diff
@@ -355,7 +349,7 @@ def _simulate_mixture(self, op: ops.Operation, data: _StateAndBuffer,
         # We work around numpy barfing on choosing from a list of
         # numpy arrays (which is not `one-dimensional`) by selecting
         # the index of the unitary.
-        prng = self.prng or np.random
+        prng = random.prng_from_seed(self._seed)
```
kevinsung: I think this line also behaves badly if self._seed is an int.

cirq/sim/sparse_simulator.py
```diff
@@ -340,7 +334,7 @@ def _simulate_measurement(self, op: ops.Operation, data: _StateAndBuffer,
                 indices,
                 out=data.state,
                 qid_shape=data.state.shape,
-                seed=self.prng)
+                seed=self._seed)
```
kevinsung: This line too.

kevinsung:
I think you've introduced some incorrect behavior here and I've added some tests in #2430 which catch these errors. Also, I think we should merge #2348 and use that instead.

PR #2430, opened 2019-10-29
Add more tests of simulator seeding

kevinsung:
This PR should be merged before any further changes to simulator seeding behavior, such as #2428 .


cirq/sim/density_matrix_simulator_test.py
```diff
@@ -936,6 +936,19 @@ def test_random_seed_does_not_modify_global_state_non_terminal_measurements():
     assert result1 == result2
 
 
+def test_random_seed_non_terminal_measurements_deterministic():
```
Strilanc: Increase the number of repetitions to 30. We don't want false negatives.

kevinsung: Done.

cirq/sim/density_matrix_simulator_test.py
```diff
@@ -936,6 +936,23 @@ def test_random_seed_does_not_modify_global_state_non_terminal_measurements():
...
+    result = sim.run(circuit, repetitions=30)
+    assert np.all(result.measurements['a'] ==
+                  [[0], [0], [1], [0], [1], [0], [1], [0], [1], [1], [0], [0],
+                   [1], [0], [0], [1], [1], [1], [0], [0], [0], [0], [1], [0],
+                   [0], [0], [1], [1], [1], [1]])
+    assert np.all(result.measurements['b'] ==
+                  [[1], [1], [0], [1], [1], [1], [1], [1], [0], [1], [1], [0],
+                   [1], [1], [1], [0], [0], [1], [1], [1], [0], [1], [1], [1],
+                   [1], [1], [0], [1], [1], [1]])
+
```
dabacon: To catch the kind of stupid mistakes I made, I think we also want one of these for terminal measurements so that it will catch the sample cases right?

PR #2434, opened 2019-10-29
Raise auto_merge script's ability to see an approving review

Strilanc:
- Specify a `per_page` argument of 100 (the maximum) (default was 30)

E.g. this made the difference in https://github.com/quantumlib/Cirq/pull/2308


Strilanc:
It is

PR #2435, opened 2019-10-29
Add atol parameter to PauliSum expectation methods

kevinsung:
Makes the signatures identical to the corresponding methods in PauliString.


cirq/ops/linear_combinations.py
```diff
@@ -347,6 +347,7 @@ def expectation_from_wavefunction(self,
                                       state: np.ndarray,
                                       qubit_map: Mapping[raw_types.Qid, int],
                                       *,
+                                      atol: float = 1e-7,
```
viathor: nit: need to add this to args section of docstring below; applies to the other function, too

Issue #2134, opened 2019-09-20
Implement the quantum part of Shor's algorithm.


vtomole:
Classical part has already been implemented: https://github.com/quantumlib/Cirq/blob/master/examples/shor.py

hanusaj:
Assign to me.

PR #2436, opened 2019-10-29
Shor's algorithm, the quantum parts

viathor:
This implements the quantum order finding algorithm.

Example use:

```
$ time python examples/shor.py 21 --order_finder=quantum
7 is a non-trivial factor of 21

real	0m14.425s
user	0m14.758s
sys	0m1.175s
```

Follow-up to #2038.
Fixes https://github.com/quantumlib/Cirq/issues/2134


examples/shor.py
```diff
@@ -83,9 +95,179 @@ def naive_order_finder(x: int, n: int) -> int:
...
+                a += 1
+            if qubit in self.exponent_register:
+                wire_symbols.append('e' + str(e))
+                e += 1
+        return cirq.CircuitDiagramInfo(wire_symbols=tuple(wire_symbols))
+
+
+def bitlength(n: int) -> int:
+    """Returns minimum number of bits needed to specify a positive integer."""
+    return int(math.ceil(math.log(n) / math.log(2)))
```
Strilanc: This is a python built-in

```
>>> (4).bit_length()
2
```

examples/shor.py
```diff
@@ -83,9 +95,179 @@ def naive_order_finder(x: int, n: int) -> int:
...
+    Returns:
+        s/r where r is the order of x modulo n and s is in [0..r-1].
+    """
+    exponent_as_integer = result.data['exponent'][0]
+    exponent_num_bits = result.measurements['exponent'].shape[1]
+    return float(exponent_as_integer / 2**exponent_num_bits)
+
+
+def quantum_order_finder(x: int, n: int) -> Optional[int]:
+    """Computes smallest positive r such that x^r mod n == 1.
```
Strilanc: Use python notation for equations. ** instead of ^

examples/shor.py
```diff
@@ -83,9 +95,179 @@ def naive_order_finder(x: int, n: int) -> int:
...
+        U|y⟩ = |y⟩                n <= y
+
+    in the header of this file.
+
+    Quantum order finding algorithm (which is the quantum part of the Shor's
+    algorithm) uses quantum modular exponentiation together with the Quantum
+    Phase Estimation to compute the order of x modulo n.
+    """
+
+    def __init__(self, ancilla_register: Sequence[cirq.Qid],
```
Strilanc: Rename to `target`, `exponent`, `base`, `modulus`.

examples/shor.py
```diff
@@ -83,9 +95,179 @@ def naive_order_finder(x: int, n: int) -> int:
...
+        self.n = n
+
+    def registers(self) -> Sequence[Sequence[cirq.Qid]]:
+        return self.ancilla_register, self.exponent_register
+
+    def with_registers(
+            self,
+            *new_registers: Union[int, Sequence['cirq.Qid']],
+    ) -> cirq.ArithmeticOperation:
+        # coverage: ignore
```
Strilanc: Test this.

examples/shor.py
```diff
@@ -83,9 +95,179 @@ def naive_order_finder(x: int, n: int) -> int:
...
+        assert len(new_registers) == 2
+        assert isinstance(new_registers[0], Sequence)
+        assert isinstance(new_registers[1], Sequence)
```
Strilanc: The exponent does not technically have to be quantum. It can be classical.

examples/shor.py
viathor: Done throughout.

examples/shor.py
viathor: Much nicer, thanks!

examples/shor.py
viathor: Done (originally left out since it is essentially dead code, unused in this example).

examples/shor.py
viathor: Done.

Initially, I gave this gate the ~minimum functionality necessary for the Shor's algorithm example to work. Now with extra test coverage and additional functionality I wonder whether we should move this gate to cirq.ops. WDYT?

Strilanc: I don't think it makes sense as an exposed gate, because it is very very *very* non-NISQ.

PR #2437, opened 2019-10-29
Add WaitGate proto serialization

Strilanc:


cirq/google/common_serializers.py
```diff
@@ -280,3 +280,24 @@ def _near_mod_2(e, t, atol=1e-8):
         gate_constructor=lambda: ops.FSimGate(theta=-np.pi / 4, phi=0),
         args=[]),
 ]
+"""Measurement serializer."""
```
dstrain115: Fix comments.  Also, Matthew suggested in a different PR that we use comments instead of string literals to improve formatting (I'll fix the other ones)

cirq/google/gate_sets.py
```diff
@@ -17,17 +17,11 @@
...
-    SYC_SERIALIZER,
-    SYC_DESERIALIZER,
-    SQRT_ISWAP_SERIALIZERS,
-    SQRT_ISWAP_DESERIALIZERS,
-)
+    SINGLE_QUBIT_SERIALIZERS, SINGLE_QUBIT_DESERIALIZERS,
+    SINGLE_QUBIT_HALF_PI_SERIALIZERS, SINGLE_QUBIT_HALF_PI_DESERIALIZERS,
+    MEASUREMENT_SERIALIZER, MEASUREMENT_DESERIALIZER, SYC_SERIALIZER,
+    SYC_DESERIALIZER, SQRT_ISWAP_SERIALIZERS, SQRT_ISWAP_DESERIALIZERS,
+    WAIT_GATE_SERIALIZER, WAIT_GATE_DESERIALIZER)
```
dstrain115: Add trailing comma so that the formatting remains the same?

cirq/google/gate_sets.py
```diff
@@ -36,12 +30,14 @@
         *SINGLE_QUBIT_SERIALIZERS,
         *SINGLE_QUBIT_HALF_PI_SERIALIZERS,
         MEASUREMENT_SERIALIZER,
+        WAIT_GATE_SERIALIZER,
```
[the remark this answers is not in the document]
Strilanc: Isn't the server using this same gate set object, though? How can I add it to one before the other if they're the same thing?

I suppose I could add it to the deserializers first.

cirq/google/common_serializers.py
Strilanc: Done.

cirq/google/gate_sets.py
Strilanc: Done.

cirq/google/gate_sets.py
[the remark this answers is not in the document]
Strilanc: Done.

Strilanc:
Refactored this to only define the serializer and deserializer, without adding it to the gate sets.

cirq/google/common_serializers.py
```diff
@@ -336,3 +336,26 @@ def _near_mod_2(e, t, atol=1e-8):
...
+# WaitGate serializer and deserializer
+#
+WAIT_GATE_SERIALIZER = op_serializer.GateOpSerializer(
+    gate_type=ops.WaitGate,
+    serialized_gate_id='wait',
+    args=[
+        op_serializer.SerializingArg(
+            serialized_name='nanos',
+            serialized_type=float,
+            gate_getter=lambda e: e.duration.total_nanos()),
```
dstrain115: I wonder if we should put a limit here, like don't allow waits of more than 10 milliseconds (or should this be on the pyle side)?

Strilanc: Yeah, that'd be a good idea.

dabacon: At least maybe yell if this is negative for now.

PR #2438, opened 2019-10-29
Lazy load param_dict and use symbol name

dstrain115:
- Only load a param_hash if we actually need it.
- Also, use a symbol's name rather than it's str(), which invokes
a bunch of custom sympy Printer nonsense.
- Speeds a sample param resolution by almost 100x.
    (see comment below for details)

### DIFF
--- a/cirq/study/resolver.py
+++ b/cirq/study/resolver.py
@@ -46,13 +46,12 @@ class ParamResolver(object):
         return super().__new__(cls)
 
     def __init__(self, param_dict: ParamResolverOrSimilarType = None) -> None:
-        if hasattr(self, '_param_hash'):
+        if hasattr(self, 'param_dict'):
             return  # Already initialized. Got wrapped as part of the __new__.
 
         self.param_dict = cast(
             Dict[Union[str, sympy.Symbol], Union[float, str, sympy.Symbol]],
             {} if param_dict is None else param_dict)
-        self._param_hash = hash(frozenset(self.param_dict.items()))
 
     def value_of(self,
                  value: Union[sympy.Basic, float, str]) -> value.TParamVal:
@@ -79,7 +78,6 @@ class ParamResolver(object):
         Returns:
             The value of the parameter as resolved by this resolver.
         """
-
         # Input is a float, no resolution needed: return early
         if isinstance(value, float):
             return value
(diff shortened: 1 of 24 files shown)


dstrain115:
Simple example:

```
import cirq
import sympy
import time

for n in [20, 40, 60, 80, 100]:
  qubits = cirq.GridQubit.square(n)
  ops = [cirq.XPowGate(exponent = sympy.Symbol(str(q)))(q) for q in qubits]
  params = {str(q):0.0 for q in qubits}
  t0 = time.time()
  for op in ops:
    cirq.resolve_parameters(op, params)
  t1 = time.time()
  print(f"n={n}: {t1-t0} secs")
```

Results before:
```
n=20: 0.021507740020751953 secs
n=40: 0.23191571235656738 secs
n=60: 1.7617378234863281 secs
n=80: 5.837962865829468 secs
n=100: 13.246250867843628 secs
```

Results after:
```
n=20: 0.004717588424682617 secs
n=40: 0.018138647079467773 secs
n=60: 0.042388200759887695 secs
n=80: 0.0766453742980957 secs
n=100: 0.12091302871704102 secs
```

cirq/study/resolver.py
```diff
@@ -133,6 +131,8 @@ def __getitem__(self, key):
         return self.value_of(key)
 
     def __hash__(self):
+        if not hasattr(self, '_param_hash'):
```
maffoo: Might be better to set `self._param_hash = None` in `__init__` and then check for the `None` value here.

cirq/study/resolver.py
```diff
@@ -102,8 +100,8 @@ def value_of(self,
 
         # Input is a symbol (sympy.Symbol('a')) and its string maps to a number
         # in the dictionary ({'a': 1.0}).  Return it.
-        if isinstance(value, sympy.Symbol) and str(value) in self.param_dict:
-            param_value = self.param_dict[str(value)]
+        if (isinstance(value, sympy.Symbol) and value.name in self.param_dict):
```
maffoo: Is it possible for `symbol.name` and `str(symbol)` to ever disagree?

dstrain115: I don't think so.  The string printing code in sympy is worthy of a scary programmer-themed Halloween movie, but, as far as I can, this basically just outputs symbol.name, at least for sympy.Symbol types.

https://github.com/sympy/sympy/blob/master/sympy/printing/str.py#L747

cirq/study/resolver.py
dstrain115: Good idea.  Done.

Issue #2432, opened 2019-10-29
Wavefunction simulator should accept NoiseModel too


kevinsung:
It can simulate mixtures and coherent noise models.

Strilanc:
I think we need to have noise models declare themselves as "monte carlo unitary safe" or something like that in order for this to work properly.

kevinsung:
I think we should add `is_mixture_of_unitaries` and `is_coherent` methods/properties to NoiseModel.

kevinsung:
We decided at #2440 to make the simulator do all the work instead of letting noise models declare whether they can be simulated.

Strilanc:
In the specific case of the wave functio nsimulator being given a noise model, it should be understood that if you give a noise model it can't handle then you will get an error.

daxfohl:
Is this just a matter of copying the noise preprocessing code from density matrix simulator to the sparse simulator?

daxfohl:
Oh, I see, the way noise currently works by creating an `OP_TREE`, and the way sparse simulator currently works via the `act_on` protocol, are incompatible. You end up with the following if you try:

```
>       raise TypeError(
            "Failed to act action on state argument.\n"
            "Tried both action._act_on_ and args._act_on_fallback_.\n"
            "\n"
            f"State argument type: {type(args)}\n"
            f"Action type: {type(action)}\n"
            f"Action repr: {action!r}\n"
        )
E       TypeError: Failed to act action on state argument.
E       Tried both action._act_on_ and args._act_on_fallback_.
E
E       State argument type: <class 'cirq.sim.act_on_state_vector_args.ActOnStateVectorArgs'>
E       Action type: <class 'cirq.ops.moment.Moment'>
E       Action repr: cirq.Moment(
E           cirq.measure(cirq.LineQubit(0)),
E       )

cirq/protocols/act_on_protocol.py:118: TypeError
```
Would it work to define noise in terms of subcircuits instead of sub-moments?

daxfohl:
It does work! The only problem is in `test_noise_composition` it no longer merges the noise because they're in subcircuits.

The first one looks like:
```

[cirq.CircuitOperation(
    circuit=cirq.FrozenCircuit([
        cirq.Moment(
            cirq.CircuitOperation(
                circuit=cirq.FrozenCircuit([
                    cirq.Moment(
                        cirq.X(cirq.LineQubit(0)),
                    ),
                    cirq.Moment(
[...]

PR #2440, opened 2019-10-29
Add is_mixture_of_unitaries and is_coherent methods to NoiseModel

kevinsung:
Part of #2432

### DIFF
--- a/cirq/devices/noise_model.py
+++ b/cirq/devices/noise_model.py
@@ -14,7 +14,9 @@
 
 from typing import TYPE_CHECKING, Sequence, Union
 
-from cirq import ops, protocols, value
+import abc
+
+from cirq import ops, linalg, protocols, value
 
 if TYPE_CHECKING:
     from typing import Iterable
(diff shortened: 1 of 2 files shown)


cirq/devices/noise_model.py
```diff
@@ -151,6 +153,14 @@ def noisy_operation(self, operation: 'cirq.Operation') -> 'cirq.OP_TREE':
             noisy version of the given operation.
         """
 
+    @abc.abstractmethod
+    def is_mixture_of_unitaries(self) -> bool:
```
Strilanc: We can't introduce new abstract methods like this. It will break existing implementations.

We could refactor into

`definitely_only_adds_mixture_operations`

and

`definitely_only_adds_unitary_operations`

with default implementations returning `False`.

kevinsung:
Just to clarify: with the solution you're proposing in which the simulator applies the noise model right before simulating a moment, it would simulate a circuit until running into something it can't simulate and then raise a runtime error, right? I'm fine with that; that would remove the need for the methods introduced here.

kevinsung:
Or, are you proposing to expand the circuit completely and check if it can be simulated before starting the simulation? That might be a better idea.

Strilanc:
Expand first, then analyze, then simulate.

kevinsung:
Ok let's go with that.

PR #2441, opened 2019-10-29
Speed up circuit creation

dstrain115:
- Reorganize qubit check so that containment is checked on a set.
- Moment.with_operation() cheats and uses private variables so that it
changes from O(n) to O(1).
- Overload validate_moment in unconstrained device so we don't iterate
over all moments.
- Perf improvement is like 100x for example code listed in comments
below.

### DIFF
--- a/cirq/contrib/quirk/cells/arithmetic_cells.py
+++ b/cirq/contrib/quirk/cells/arithmetic_cells.py
@@ -49,10 +49,10 @@ class QuirkArithmeticOperation(ops.ArithmeticOperation):
         self.target = target
         self.inputs = inputs
 
-        for input in self.inputs:
-            if isinstance(input, int):
+        for input_register in self.inputs:
+            if isinstance(input_register, int):
                 continue
-            if set(self.target) & set(input):
+            if set(self.target) & set(input_register):
                 raise ValueError(f'Overlapping registers: '
                                  f'{self.target} {self.inputs}')
 
--- a/cirq/contrib/quirk/cells/arithmetic_cells.py
+++ b/cirq/contrib/quirk/cells/arithmetic_cells.py
@@ -49,6 +49,13 @@ class QuirkArithmeticOperation(ops.ArithmeticOperation):
         self.target = target
         self.inputs = inputs
 
+        for input in self.inputs:
+            if isinstance(input, int):
+                continue
+            if set(self.target) & set(input):
+                raise ValueError(f'Overlapping registers: '
+                                 f'{self.target} {self.inputs}')
+
         if self.operation.is_modular:
             r = inputs[-1]
             if isinstance(r, int):
--- a/cirq/ops/moment.py
+++ b/cirq/ops/moment.py
@@ -81,7 +81,7 @@ class Moment:
         Returns:
             Whether this moment has operations involving the qubits.
         """
-        return any(q in self.qubits for q in qubits)
+        return bool(set(qubits) & self.qubits)
 
     def with_operation(self, operation: raw_types.Operation):
         """Returns an equal moment, but with the given op added.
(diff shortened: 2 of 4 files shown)


dstrain115:
Sample code:
```
import cirq
import sympy
import time

for n in [20, 40, 60, 80, 100]::
  global ops
  qubits = cirq.GridQubit.square(n)
  ops = [cirq.XPowGate(exponent = sympy.Symbol(str(q)))(q) for q in qubits]
  t0 = time.time()
  _ = cirq.Circuit(ops)
  t1 = time.time()
  print(f"n={n}: {t1-t0} secs")
```

Output before:
```
n=20: 0.2778351306915283 secs
n=40: 4.419932126998901 secs
n=60: 22.919171571731567 secs
n=80: 72.89886021614075 secs
n=100: 177.47597742080688 secs
```

Output after:
```
n=20: 0.0062236785888671875 secs
n=40: 0.04632425308227539 secs
n=60: 0.20130109786987305 secs
n=80: 0.6193749904632568 secs
n=100: 1.452638864517212 secs
```

cirq/ops/moment.py
```diff
@@ -81,7 +81,7 @@ def operates_on(self, qubits: Iterable[raw_types.Qid]) -> bool:
         Returns:
             Whether this moment has operations involving the qubits.
         """
-        return any(q in qubits for q in self.qubits)
+        return any(q in self.qubits for q in qubits)
```
maffoo: oh boy, how did this even work before?

cirq/ops/moment.py
```diff
@@ -92,7 +92,15 @@ def with_operation(self, operation: raw_types.Operation):
         Returns:
             The new moment.
         """
-        return Moment(self.operations + (operation,))
+        if any(q in self._qubits for q in operation.qubits):
+            raise ValueError('Overlapping operation: {}'.format(operation))
+
+        # Use private variables to facilitate a quick copy
```
dabacon: This is because you are trying to avoid the overlap check being evaluated?

dstrain115: I am trying to skip the whole __init__ function, so I have to replicate the overlap check as well.  I am basically re-implementing the __init__ function for this special case (which is used heavily in circuit initialization)

cirq/ops/moment.py
dabacon: It may be faster to say
`return set(qubits) & self.qubits`
See https://stackoverflow.com/questions/2197482/efficiently-knowing-if-intersection-of-two-list-is-empty-or-not-in-python

dstrain115: Done.  I don't see much difference, but it might be significant for larger numbers of qubits.

cirq/contrib/quirk/cells/arithmetic_cells.py
```diff
@@ -49,6 +49,13 @@ def __init__(self, identifier: str, target: Sequence['cirq.Qid'],
         self.target = target
         self.inputs = inputs
 
+        for input_register in self.inputs:
```
Strilanc: Quirk input registers are technically allowed to overlap ( https://algassert.com/quirk#circuit={%22cols%22:[[%22X%22,1,1,%22X%22],[%22inputA3%22,1,%22inputB3%22,1,1,1,1,%22+=AB3%22]]} ), it just doesn't really come up because the program fights you on it.

I would split this out into a separate PR.

Strilanc: In general an operation with overlapping inputs is well-defined. Though I can see why this would confuse the moment overlap check down the line.

dstrain115:
> On further thought, it's fine to block the overlapping registers for now.

Sure, feel free to fix it correctly in the future.  I was just trying to keep the previous behavior.  I didn't really understand what was happening there.

Issue #2442, opened 2019-10-29
Remove cirq.contrib import inside protocols/json.py


Strilanc:
```
grep -Rl "\bcirq\.contrib\b" cirq | grep -v "cirq/contrib" | grep -v "__"
```

```
cirq/protocols/json.py
cirq/protocols/json_test.py
```

Contrib is explicitly for things that are unstable and changing. We should not be referencing things from inside the more stable parts of the library. *Particularly* serialization, which must remain compatible indefinitely.

I caught this is because it triggered an actual test failure when I was attempting to verify the next release.

```
_______________________ ERROR collecting python3/lib/python3.7/site-packages/cirq/protocols/json_test.py _______________________
python3/lib/python3.7/site-packages/cirq/protocols/json_test.py:30: in <module>
    from cirq.contrib.quantum_volume import QuantumVolumeResult
python3/lib/python3.7/site-packages/cirq/contrib/quantum_volume/__init__.py:16: in <module>
    from cirq.contrib.quantum_volume.quantum_volume import (
python3/lib/python3.7/site-packages/cirq/contrib/quantum_volume/quantum_volume.py:195: in <module>
    @dataclass
python3/lib/python3.7/site-packages/dataclasses.py:958: in dataclass
    return wrap(_cls)
python3/lib/python3.7/site-packages/dataclasses.py:950: in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash, frozen)
python3/lib/python3.7/site-packages/dataclasses.py:801: in _process_class
    for name, type in cls_annotations.items()]
python3/lib/python3.7/site-packages/dataclasses.py:801: in <listcomp>
    for name, type in cls_annotations.items()]
python3/lib/python3.7/site-packages/dataclasses.py:659: in _get_field
    if (_is_classvar(a_type, typing)
python3/lib/python3.7/site-packages/dataclasses.py:550: in _is_classvar
    return type(a_type) is typing._ClassVar
E   AttributeError: module 'typing' has no attribute '_ClassVar'
```

Something to do with dataclasses and 3.7 vs 3.6.

PR #2443, opened 2019-10-29
Add a build check for contrib references within non-contrib code

Strilanc:
- Remove the contrib reference inside json.py and json_test.py

Fixes https://github.com/quantumlib/Cirq/issues/2442

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,11 @@ matrix:
     script: check/pytest --benchmark-skip --actually-quiet
   - stage: test
     os: linux
+    env: NAME=misc
+    python: '3.6'
+    script: check/misc
+  - stage: test
+    os: linux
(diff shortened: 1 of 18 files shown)


mpharrigan:
Do you have any ideas for how to register deserializers? We could force users to pass a custom resolver to `cirq.read_json` or have a contrib version of `read_json` or have a "registration" mechanism that appends contrib classes to the default resolvers if you import the contrib module

Strilanc:
Ah, yes, a pressure relief mechanism would be good.

The to_json and read_json methods could take a "custom_types" argument that was a dictionary from string to type.

We could define `cirq.contrib.to_json`... meh.

mpharrigan:
it already takes a `resolvers` argument; it's just pretty clunky without wrapping it in something more convenient

Issue #2431, opened 2019-10-29
Switch from cirq-dev pypi package to cirq-unstable pypi package


Strilanc:
I think this is less confusing in important ways. Opinions?

@mpharrigan 
@dabacon 
@maffoo 
@cduck

maffoo:
@Strilanc, can you describe the confusion you're hoping to prevent? I'm not tied to the name `cirq-dev`, just want to make sure I understand the thinking here and that we pick a different name that avoids the problems you see.

Another thing that I'd like to at least explore before we do this is whether it's possible to prevent pip from installing both cirq and cirq-whatever in the same virtualenv, since currently this is possible and leads to weird failures (see #2347). For people who now have cirq-dev installed, the same issue will come up if we switch to cirq-unstable, so if there's a way to fix this that would ease the transition.

Strilanc:
> can you describe the confusion you're hoping to prevent?

We have the package variant cirq[dev-env], which installs a developer environment.
We have the alternate package cirq-dev, which installs the latest version of cirq from master.

They both use the word "dev" in a packaging context as a suffix to cirq, but it means drastically different things in each case.

Strilanc:
> Another thing that I'd like to at least explore before we do this is whether it's possible to prevent pip from installing both cirq and cirq-whatever in the same virtualenv, since currently this is possible and leads to weird failures

The same thing happens with `tensorflow` and `tf-nightly`

```
mkvirtualenv breakage --python=/usr/bin/python3
pip install tf-nightly
python -c "import tensorflow"  # works
pip install tensorflow
pip uninstall tensorflow
python -c "import tensorflow"  # blows up
pip install tf-nightly
python -c "import tensorflow"  # blows up
```

PR #2444, opened 2019-10-29
Rename cirq-dev to cirq-unstable

Strilanc:
- Refactor verify-published-package to have an --unstable flag.
- Refactor verify-published-package to require --prod/--test/--unstable be specified
- Rename various CIRQ_DEV type things to CIRQ_UNSTABLE type things

Fixes https://github.com/quantumlib/Cirq/issues/2431

### DIFF
--- a/dev_tools/packaging/publish-dev-package.sh
+++ b/dev_tools/packaging/publish-dev-package.sh
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 ################################################################################
-# Produces and uploads dev-version wheels to a pypi package repository cirq-dev
+# Produces and uploads dev-version wheels to a pypi package 'cirq-unstable'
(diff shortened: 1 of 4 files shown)


Strilanc:
At least until next time.

dev_tools/packaging/publish-dev-package.sh
```diff
@@ -15,7 +15,7 @@
 # limitations under the License.
```
dabacon: rename script since it now pushes to unstable?

dabacon:
> this is the last time we change the name of this!

Said by every software engineer every time.

balopat:
> > this is the last time we change the name of this!
> 
> Said by every software engineer every time.

Maybe https://github.com/quantumlib/Cirq/pull/3527 is the last time we rename this...?

Issue #1452, opened 2019-03-20
Define a destructive measurement channel


Strilanc:
Related to https://github.com/quantumlib/Cirq/issues/882 but I see that issue as an optimization of the simulator as opposed to defining a channel.

dabacon:
Is this for measurement and reset?  Or for actual destructive measurement?

Channel for measurement and reset to 0 has
A_0 = [[1, 0], [0, 0]]
A_1 = [[0, 1], [0, 0]]
(you can also reset to different basis, so it might be useful to build the correct primitive version of these that are basis dependent).

If it is a real destructive measurement (qubit disappears, like a photon) then probably then the channel goes to a one dimensional space
A_0 = [[1, 0]]
A_1 = [[0, 1]]

dabacon:
Related #1357

tanujkhattar:
We support measurement and reset channel now via `ResetChannel`. 

Is the destructive measurement channel useful in practice and something that should be added? If not, we can close this issue.

mhucka:
Closing, as this appears to be either sufficiently handled or no longer a priority. (If this is wrong, someone please feel free to comment and reopen this issue.)

PR #2448, opened 2019-10-30
Validate Kraus operators

### LINKED ISSUE #2271: Add validation for Kraus operators
Many of the methods involving noise/channels mention that the matrices making up a mixture or channel must satisfy \sum_{k=0}^{r-1} A_i^\dagger A_i = I. This is never enforced, which allows the user to simulate unphysical systems and get invalid density matrices (see example below).

Validation is easy enough (`assert np.sum([m.T @ m for m in object._channel_()`) but I'm not sure where to put this @Strilanc ? Ideally this will be called exactly once, before a call to `simulate`. 

Code Example:
```
from typing import Iterable

import cirq
import numpy as np

class UnnormalizedChannel(cirq.GeneralizedAmplitudeDampingChannel):
    """Overwrite the correct amplitude damping channel with a bad one."""
    def _channel_(self) -> Iterable[np.ndarray]:
        p0 = 1
        sqrt_g = np.sqrt(self._gamma)
        return (
            p0 * np.array([[1., 0.], [0., 1.]]),
            p0 * np.array([[0., sqrt_g], [0., 0.]]),
        )


def unnormalized_channel(gamma):
    return UnnormalizedChannel(1.0, gamma)


q0 = cirq.LineQubit(0)
bad_noise_inst = unnormalized_channel(0.3)
c = cirq.Circuit.from_ops(cirq.X(q0), bad_noise_inst.on(q0))
dm = cirq.DensityMatrixSimulator().simulate(c).final_density_matrix
kraus_sum = np.sum(m.T @ m for m in bad_noise_inst._channel_())
print(np.allclose(kraus_sum, np.eye(2)))
print(np.isclose(np.trace(dm), 1.0))
```
which prints
```
>>> False
>>> False
```

peterse:
fixes #2271 

The channel/mixture validations in `common_channels` is purely cosmetic, but it feels like a good way to nudge the user towards validating their own custom channel (and hopefully understand the tolerance on trace preservation this introduces to density matrices passed through the simulator).

also adds tolerances for `validate_mixture`


cirq/ops/common_channels.py
```diff
@@ -59,6 +59,7 @@ def __init__(self, p_x: float, p_y: float, p_z: float) -> None:
         self._p_z = value.validate_probability(p_z, 'p_z')
         self._p_i = 1 - value.validate_probability(p_x + p_y + p_z,
                                                    'p_x + p_y + p_z')
+        protocols.validate_mixture(self._mixture_())
```
Strilanc: I don't think we should be validating the mixture every time a user constructs a class. The class is supposed to guarantee this property based on validating the probabilities. Verifying that implication is the purpose of unit tests.

cirq/ops/common_channels.py
```diff
@@ -289,6 +290,7 @@ def __init__(self, p: float, gamma: float) -> None:
         """
         self._gamma = value.validate_probability(gamma, 'gamma')
         self._p = value.validate_probability(p, 'p')
+        protocols.validate_channel(self._channel_())
```
Strilanc: Don't validate on construction.

cirq/ops/common_channels.py
```diff
@@ -630,6 +632,7 @@ def __init__(self, gamma: float) -> None:
             ValueError: if gamma is not a valid probability.
         """
         self._gamma = value.validate_probability(gamma, 'gamma')
+        protocols.validate_channel(self._channel_())
```
Strilanc: Don't validate on construction.

cirq/protocols/channel.py
```diff
@@ -175,3 +178,19 @@ def has_channel(val: Any) -> bool:
 
     # No has methods, use `_channel_` or delegates instead.
     return channel(val, None) is not None
+
+
+def validate_channel(supports_channel: SupportsChannel, rtol: int = 1e-5, atol: int = 1e-8):
+    """Validates that the channel's unitaries are a valid representation."""
```
dabacon, line 184: Channel's don't have unitaries :)  Doc should be something like "Validates that an object's channel Krauss operators are normalized."  

Also should also explicitly mention that this is
\sum_{k=0}^{r-1} A_i^\dagger A_i = I

cirq/protocols/channel.py
```diff
@@ -175,3 +178,19 @@ def has_channel(val: Any) -> bool:
 
     # No has methods, use `_channel_` or delegates instead.
     return channel(val, None) is not None
+
+
+def validate_channel(supports_channel: SupportsChannel, rtol: int = 1e-5, atol: int = 1e-8):
```
dabacon, line 183: The type for `supports_channel` isn't quite correct because this also works for channel mixtures.  Suggest changing this to a union type that includes mixture and unitary cases.  Then I'd just call this parameter `channel`.

cirq/protocols/channel.py
```diff
@@ -175,3 +178,19 @@ def has_channel(val: Any) -> bool:
...
+    """Validates that the channel's unitaries are a valid representation."""
+    channel_result = channel(supports_channel, None)
+    if channel_result is None:
+        raise TypeError("object of type '{}' has no _channel_ or _mixture_ or "
+                        "_unitary_ method.".format(type(supports_channel)))
+    shapes = list(set([m.shape for m in channel_result]))
+    if len(shapes) != 1 or len(shapes[0]) != 2 or shapes[0][0] != shapes[0][1]:
```
dabacon, line 190: I know that currently we only support channels whose input space is equal to its output space, but technically this not required.  I would suggest not enforcing this.  There are plans for operators that would need to change space like #1452.

cirq/protocols/channel_test.py
```diff
@@ -154,3 +154,57 @@ def _has_unitary_(self) -> bool:
...
+                       match='Operator sum is not trace preserving'):
+        cirq.validate_channel(InvalidOperatorSum1(), rtol=1e-5, atol=0.01)
+
+    class InvalidOperatorSum2:
+        def _channel_(self) -> Iterable[np.ndarray]:
+            return (
+                np.array([[1., 0.], [0., 1.]]),
+                np.array([[0., np.sqrt(0.05)], [0., 0.]]))
+
+    # TODO: make this work
```
dabacon, line 206: why doesn't this work?

balopat:
Hi @peterse - can you confirm whether you're planning to work on this? Otherwise, we can just close it and you can reopen when you have the cycles to finish it.

peterse:
okay I'll reopen or start a new PR when I get the chance to look at it again

PR #2450, opened 2019-10-30
Fix dataclasses package being installed for python 3.7+

Strilanc:
Python 3.7 has a built-in dataclasses module. This was causing a failure when verifying v0.6.0's package, because the dataclasses shim package was accessing private properties from typing that no longer existed in 3.7.


Strilanc:
*cough*

PR #2451, opened 2019-10-30
Fix __radd__ losing device

Strilanc:


cirq/circuits/circuit.py
```diff
@@ -276,8 +276,9 @@ def __radd__(self, other):
         if not isinstance(other, (ops.Operation, Iterable)):
             return NotImplemented
         # Auto wrap OP_TREE inputs into a circuit.
-        result = Circuit(other)
-        return result.__iadd__(self)
+        result = self.copy()
+        result._moments[:0] = Circuit(other)._moments
```
cduck: This doesn't appear to validate the new moment against the device.

cirq/circuits/circuit_test.py
```diff
@@ -277,6 +277,11 @@ def test_radd_op_tree():
         [cirq.Moment([cirq.X(a)]),
          cirq.Moment([cirq.Y(b)])])
 
+    # Preserves device.
+    c = cirq.Circuit(device=cirq.google.Bristlecone)
+    c2 = [] + c
+    assert c2.device is cirq.google.Bristlecone
+
```
cduck: Also consider adding a test that attempts to prepend a moment containing a disallowed gate.

Strilanc: Done.

cirq/circuits/circuit.py
Strilanc: Fixed.

cduck: Is validating the entire new circuit needed?  I would just validate the new moment:

```
result._moments[:0] = Circuit(other, device=self.device)._moments
```

c-poole: There are constraints on operations that can be applied in sequence on the same qubits for certain devices, so validating the whole circuit is necessary to catch that issue.

cduck: OK, but Circuit.insert only validates the moment.  This is a separate issue from this PR.

PR #2458, opened 2019-10-30
Allow creating new gatesets with added gates

maffoo:
This will let us create "pre-release" gatesets to deploy server-side support for new gates that are not yet supported by the API.


cirq/google/serializable_gate_set.py
```diff
@@ -51,12 +49,35 @@ def __init__(self, gate_set_name: str,
...
-            list)  # type: Dict[Type, List[op_serializer.GateOpSerializer]]
+        self.serializers: Dict[Type, List[op_serializer.GateOpSerializer]] = {}
         for s in serializers:
-            self.serializers[s.gate_type].append(s)
+            self.serializers.setdefault(s.gate_type, []).append(s)
         self.deserializers = {d.serialized_gate_id: d for d in deserializers}
 
+    def with_gates(
+            self,
+            *,
```
dstrain115: Why do we need this asterisk?  Is there a reason to have an empty args?

cirq/google/serializable_gate_set.py
```diff
@@ -51,12 +49,35 @@ def __init__(self, gate_set_name: str,
...
+        Args:
+            name: Optional new name of the gateset. If not given, use the same
+                name as this gateset.
+            serializers: Serializers to add to those in this gateset.
+            deserializers: Deserializers to add to those in this gateset.
+        """
+        return SerializableGateSet(
+            name or self.gate_set_name,
+            serializers=[*self._all_serializers(), *serializers],
+            deserializers=[*self.deserializers.values(), *deserializers])
```
dstrain115: Why do we need _all_serializers() but not deserializers?

cirq/google/serializable_gate_set.py
```diff
@@ -51,12 +49,35 @@ def __init__(self, gate_set_name: str,
...
         self.gate_set_name = gate_set_name
-        self.serializers = defaultdict(
-            list)  # type: Dict[Type, List[op_serializer.GateOpSerializer]]
+        self.serializers: Dict[Type, List[op_serializer.GateOpSerializer]] = {}
         for s in serializers:
-            self.serializers[s.gate_type].append(s)
+            self.serializers.setdefault(s.gate_type, []).append(s)
         self.deserializers = {d.serialized_gate_id: d for d in deserializers}
 
+    def with_gates(
```
dstrain115: This function needs tests.

cirq/google/serializable_gate_set.py
maffoo: Serializers are stored as a map from gate_id to list of serializers, so we have to flatten those into one list. Deserializers are stored as a map from gate id to deserializer, so we can just use `.values()`. I will refactor to just inline the expression, since it's probably not worth having a whole method for the serializer flattening.

cirq/google/serializable_gate_set.py
maffoo: This forces using keyword args, which I think helps clarity. You're correct that caling this with no arguments is not very useful because it just returns an equivalent gateset. I could also make serializers and deserializers be positional args and put the optional name last. Do you have a preference?

cirq/google/serializable_gate_set.py
maffoo: Added.

Strilanc: Rename this to `with_more_gates` or something similar, to make it clearer that the original gates are not being discarded.

maffoo: Changed to `with_added_gates`.

Issue #2406, opened 2019-10-25
PointOptimizer that respects moment structure


mpharrigan:
Right now ops get flattened during initialization of PointOptimizationSummary and during circuit.insert_at_frontier. Would be nice to be able to transform circuits in a more structured way

tanujkhattar:
I'll marked this as fixed given the new moment preserving primitives, which are now intended to replace PointOptimizer completely. Please feel free to file another issue if you run into a scenario where the newly added moment preserving primitives are not sufficient and you need a moment preserving version of PointOptimizer.

tanujkhattar:
https://github.com/quantumlib/Cirq/pull/5154 adds a tutorial to Cirq on new transformer infrastructure, with details of newly added moment preserving primitives.

PR #2461, opened 2019-10-30
PointOptimizerSummary option to avoid flattening

mpharrigan:
This isn't actually enough as the point optimizer will flatten it (again). But this *does* make it so I can write my own PointOptimizer replacement that can respect it

xref #2406

### DIFF
--- a/cirq/circuits/optimization_pass.py
+++ b/cirq/circuits/optimization_pass.py
@@ -31,7 +31,7 @@ class PointOptimizationSummary:
     """A description of a local optimization to perform."""
 
     def __init__(self, clear_span: int, clear_qubits: Iterable['cirq.Qid'],
-                 new_operations: 'cirq.OP_TREE') -> None:
+                 new_operations: 'cirq.OP_TREE', preserve_moments: bool = False) -> None:
         """
         Args:
             clear_span: Defines the range of moments to affect. Specifically,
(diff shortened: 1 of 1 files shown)


cirq/circuits/optimization_pass.py
```diff
@@ -41,8 +41,11 @@ def __init__(self, clear_span: int, clear_qubits: Iterable['cirq.Qid'],
                 with each affected moment.
             new_operations: The operations to replace the cleared out
                 operations with.
+            preserve_moments: Whether to keep Moments intact instead of
+                flattening them
```
kevinsung: Since, as you said, moments will still be flattened anyway, is this description a bit misleading?

mpharrigan: updated description

cirq/circuits/optimization_pass.py
```diff
@@ -41,8 +44,14 @@ def __init__(self, clear_span: int, clear_qubits: Iterable['cirq.Qid'],
                 with each affected moment.
             new_operations: The operations to replace the cleared out
                 operations with.
+            preserve_moments: Whether to keep Moments intact instead of
+                flattening them in this summary. Please be advised that
+                a PointOptimizer consuming this summary will flatten
+                operations no matter what, see gh-2406.
```
Strilanc: Use a full link instead of the abbreviation.

cirq/circuits/optimization_pass.py
```diff
@@ -41,8 +44,14 @@ def __init__(self, clear_span: int, clear_qubits: Iterable['cirq.Qid'],
                 with each affected moment.
             new_operations: The operations to replace the cleared out
                 operations with.
+            preserve_moments: Whether to keep Moments intact instead of
```
Strilanc: To a user looking over this for the first time, this description is not very clear. I would rephrase into something like

```
If set, `cirq.Moment` instances within `new_operations` will be preserved exactly.
Normally the operations would be repacked to fit better into the target space, which
may move them between moments.
```

mpharrigan: you should update the description on `flatten_op_tree`, as I have replicated it here :) Using your wording plus the additional warning per kevin's concerns

cirq/circuits/optimization_pass.py
mpharrigan: done

PR #2464, opened 2019-10-31
[WIP] Generate API reference using `documented` decorator

Strilanc:
- Puts documented values into a private global index
- Test that public values are marked as documented
- Test that private values are not marked as documented


dabacon:
Couple of questions (I don't know the answers)
- Why not try to do this by blacklisting and not whitelisting and use module name?
- Do we care about module path not matching category?

cirq/_compat.py
```diff
@@ -15,13 +15,83 @@
...
 
+DocProperties = NamedTuple(
+    'DocProperties',
+    [
+        ('api_reference_category', str),
+        ('manual_doc_string', Optional[str]),
+    ],
+)
+
+seen_documentation: Dict[Union[int, str], DocProperties] = {}
```
mpharrigan: global vars should be all caps

mpharrigan:
It makes me uncomfortable to see so many files changed just to categorize objects. I think this is a fairly unusual approach for documenting python packages. If you want to do automatic reference doc generation with whitelisted class/object names you can *definitely* do that using only sphinx features. 

Have you profiled if there's any impact on the time it takes to do `import cirq` with this change?

I think this is what @dabacon was suggesting, but you can probably just comb through all the public classes and use the module to indicate the category. You could even have an "exceptions" dictionary to re-categorize things (like e.g. for functions in utility modules)

Strilanc:
> I think this is a fairly unusual approach for documenting python packages.

This approach solves a few problems at the same time.

1. We need a way to associate docstrings with global constants. It really *really* sucks that `PAULI_STRING_LIKE`'s documentation is the docstring for `typing.Union`.

2. We need to ensure the API list is kept up to date as new objects are added. This change includes a test asserting that the set of all the things passed into `documented` equals the set of all things found by scraping.

3. We need to categorize objects in a way that's not directly tied to the implementation details. For example, we may split gates into "basic" and "advanced" or merge the study package and the simulation package into just "simulation". The user-exposed categories are not necessarily the same as the internal-developer categories, which makes sense since they have different goals.

(3) could be solved reasonably well with a special annotation within the doc string.

(2) could be solved reasonably well by auto-generating the API.rst file by scraping. I do like that this approach makes it so explicit which values are top-level values.

For (1) I have no idea how to do it other than with the approach here. I guess we could try to find the source code where the constant is defined and parse out the comment right before it... *ugh*. We could limit this idea to just the constants instead of also decorating the functions.

Regardless of whether we do it with a decorator or a docstring section, we will need to touch all these files. This is a project-spanning issue.

mpharrigan:
For (1), there's some discussion on #2471 

For (2), you could check that the set of all things listed in `api.rst` equals the set of all things found by scraping

(3) is already solved by headings / `autosummary`s in api.rst

----

> Regardless of whether we do it with a decorator or a docstring section, we will need to touch all these files. This is a project-spanning issue.

I think this speaks to a philosophical design choice that maybe someone can come up with a name for. Should objects know how they're documented (ie their doc category) or should the documentation system keep a registry of how things are documented. I naturally gravitate towards the latter in these decisions because I find that it keeps things more modular. For example: It would be easier to rip out the `linalg` module and vendor it into an otherwise unrelated project because you wouldn't have to shim, remove, or port the documentation annotations.

A perhaps-controversial decision of a similar flavor happened in my pyquil days. Since the simulator and compiler were separate projects it was impossible to get the unitary matrices for named gates. Instead of teaching each gate its matrix, I added a separate module that included a mapping from a gate to its matrix

-------

anyways, autodoc is kinda garbage so if we need to put something in like this then fine; but I think it is rather unusual and we should consider all options

Strilanc:
Obsoleted.

PR #2470, opened 2019-10-31
Speed up QAOA example

kevinsung:
Computes the objective function values using vectorized numpy operations. As written the example only uses 6 qubits and already the improvement is from ~7 seconds to ~2 seconds. This makes the code actually practical at larger qubit numbers where the speed up is astronomical.

### DIFF
--- a/examples/qaoa.py
+++ b/examples/qaoa.py
@@ -93,16 +93,15 @@ def main(repetitions=1000, maxiter=50):
         result = simulator.run(circuit, repetitions=repetitions)
         bitstrings = result.measurements['m']
         # Process bitstrings
-        sum_of_cut_values = 0
         nonlocal largest_cut_found
         nonlocal largest_cut_value_found
-        for bitstring in bitstrings:
-            value = cut_value(bitstring, graph)
-            sum_of_cut_values += value
-            if value > largest_cut_value_found:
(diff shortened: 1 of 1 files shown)


examples/qaoa.py
```diff
@@ -151,8 +150,12 @@ def qaoa_max_cut_circuit(qubits, betas, gammas,
         cirq.measure(*qubits, key='m'))
 
 
-def cut_value(bitstring, graph):
-    return sum(bitstring[i] != bitstring[j] for i, j in graph.edges)
+def cut_values(bitstrings, graph):
+    mat = networkx.adjacency_matrix(graph, nodelist=sorted(graph.nodes))
+    vecs = (-1)**bitstrings
+    vals = 0.5 * np.einsum('...i,...i', vecs, (mat * vecs.T).T)
```
mpharrigan: do you need einsum here? I think `'...i,...i'` is the default numpy broadcasting behavior (?) Could you achieve the same effect with

```python
vecs * (mat * vecs.T).T)
```

**or**

can you write a better einsum to avoid one or both of the `.T`'s

mpharrigan: or something like 
```python
vecs.T @ mat @ vecs
```

PR #2471, opened 2019-11-01
Associate docstrings with public constants

Strilanc:
- Add `cirq/_doc.py` with a `document` function
- Pass public constants into `document` when they are initialized
- Fixes the fact that e.g. [`cirq.OP_TREE` was being documented using `typing.Union`'s docstring](https://cirq.readthedocs.io/en/stable/generated/cirq.OP_TREE.html)

### DIFF
--- a/cirq/__init__.py
+++ b/cirq/__init__.py
@@ -22,6 +22,7 @@ from cirq import (
     # Low level
     _version,
     _compat,
+    _doc,
     type_workarounds,
 )
 with _import.delay_import('cirq.protocols'):
--- a/cirq/_doc.py
+++ b/cirq/_doc.py
@@ -1,0 +1,40 @@
+# Copyright 2018 The Cirq Developers
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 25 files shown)


mpharrigan:
So there's a way to do this in sphinx -- sortof.

If you read the documentation (particularly for the `autodata` directive) it says you can just add a docstring under it

```python
OP_TREE = Union[Operation, OpTree]
"""loo loo loo"""
```

At first blush this doesn't work. That's because `api.rst` lists `cirq.OP_TREE`. After some digging, these data docstrings don't make it through import forwarding (https://github.com/sphinx-doc/sphinx/issues/6495). So if you change the listing to `~cirq.ops.op_tree.OP_TREE` it shows the docstring. The `~` means "hide the module qualifiers" so it still shows up in the API table as just `OP_TREE`; but the title of the page and its toc entry has the fully qualified name (which is bad). One could probably fix this with custom autodoc templates. It gets dicey for the places where we *do* want some sort of name qualification, like in `google.XXX` and `testing.XXX`

I think it's important to clearly understand the limitations of autodoc before we re-implement (parts of) it

Strilanc:
Where did you find documentation on the tilde modifier?

mpharrigan:
A very good question. I learned it at some point and was actually trying to find documentation on it to see if there were additional options to control the *degree* of name qualification but alas I couldn't find anything about `~`. Empirically it works though

Strilanc:
After group discussion, we decided to go with this `document` function.

PR #2479, opened 2019-11-02
Add JSON resolver decorator

### LINKED ISSUE #2445: (re-)register deserializer for QuantumVolumeResult
following #2443

cc @KevinVillela

KevinVillela:
Will help to resolve #2445.

The idea here is to move away from the model for the json.py file has to reference all of the other classes, and instead have a Plug-In model where all of the classes can register themselves with the json resolver. That way, code in contrib/ can also use this decorator, and json.py will not need to reference it directly.

This PR is just a proof of concept - if it makes sense, I will add the decorator to all the other classes in Cirq. Any input is appreciated :) @mpharrigan @Strilanc

### DIFF
--- a/cirq/protocols/json.py
+++ b/cirq/protocols/json.py
@@ -48,15 +48,26 @@ class _ResolverCache:
     def __init__(self):
         self._crd = None
 
+    def add_resolver(self, key: str, cls: Type):
+        if self._crd is None:
+            self._crd = {}
+        if key not in self._crd:
+            self._crd[key] = cls
+
     @property
     def cirq_class_resolver_dictionary(self) -> Dict[str, Type]:
         if self._crd is None:
+            self._crd = {}
+        # Lazily initialize third-party resolvers (which we don't have
+        # control over and therefore can't add with the @AddJson decorater)
+        # by checking if one of these resolvers has already been added to
+        # the map or not.
+        if 'pandas.DataFrame' not in self._crd:
             import cirq
             from cirq.devices.noise_model import _NoNoiseModel
             from cirq.google.devices.known_devices import (
                 _NamedConstantXmonDevice)
-            self._crd = {
-                'AmplitudeDampingChannel': cirq.AmplitudeDampingChannel,
+            third_party = {
                 'AsymmetricDepolarizingChannel':
                 cirq.AsymmetricDepolarizingChannel,
                 'BitFlipChannel': cirq.BitFlipChannel,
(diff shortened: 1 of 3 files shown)


cirq/protocols/json.py
```diff
@@ -132,12 +143,33 @@ def cirq_class_resolver_dictionary(self) -> Dict[str, Type]:
...
                 'complex': complex,
             }
+            self._crd.update(third_party)
         return self._crd
 
 
 RESOLVER_CACHE = _ResolverCache()
 
 
+def AddJson(Cls=None, *, key=""):
```
mpharrigan: Technically I think this should be in snake case, ie `add_json(cls=...)`. Or maybe "`with_json`" so that it reads nicer as a decorator.

It might be nice to have `key` default to the class name

cirq/protocols/json.py
```diff
@@ -48,15 +48,26 @@ class _ResolverCache:
...
+        # by checking if one of these resolvers has already been added to
+        # the map or not.
+        if 'pandas.DataFrame' not in self._crd:
             import cirq
             from cirq.devices.noise_model import _NoNoiseModel
             from cirq.google.devices.known_devices import (
                 _NamedConstantXmonDevice)
-            self._crd = {
-                'AmplitudeDampingChannel': cirq.AmplitudeDampingChannel,
+            third_party = {
```
mpharrigan: first_party?

KevinVillela:
Thanks for the response! 

A. Having every library make their own `read_json` seems a bit annoying. 
C, D. I agree that imports having side effects is not ideal, either.

Manually constructing a list of `resolvers` doesn't seem too bad. I'm OK with having a `register_json_deserializers` function as well. Maybe the real solution here lies in enforcing these patterns and helping the user to find them. I'd imagine we can have similar tests in json_test.py in other directories, and craft better error messages for when the user doesn't provide the correct resolvers or doesn't call the resolver function.

Strilanc:
I'm very worried by the fact that the parsing behavior can change at runtime due to importing a module.

I don't think we should have things in contrib adding runtime objects into non-contrib modules. That's just turning the disallowed static dependency into an even more troubling dynamic dependency. I would not consider this PR to be a solution for #2445, because I would not want the quantum volume result class to have this annotation added while it was in contrib. This would be a banned annotation within contrib.

For ensuring api.rst was consistent with the library I initially tried this kind of annotation approach. We ultimately decided to instead go with api.rst being the single source of truth, because that put all the information in one spot with the needed context for working with it. I feel similarly about serialization: we can disperse the details throughout the library, but it's important that we have a central area where we can definitively state whether or not something is included or not.

KevinVillela:
Agreed - sorry, I had meant to close this out and send out a different solution, maybe adding a `register_json_deserializers` function in contrib.

Thanks for the response!

PR #2484, opened 2019-11-04
Add cirq.ThermalRelaxationChannel

MichaelBroughton:
Added a new noise channel to allow for simultaneous phase and amplitude noise. I opted to use the choi matrix so that the math would be a little nicer on the eyes (also I kept fiddling with the GAD kraus operators and ultimately concluded I couldn't figure a way to just substitute in the beta values on the off diagonals).

Also wouldn't mind a better name than `thermal_relaxation`


cirq/ops/thermal_relaxation_channel.py
```diff
@@ -0,0 +1,229 @@
...
+        Construct a channel to model energy dissipation into the environment
+        as well as the environment depositing energy into the system. The
+        probabilities with which the energy exchange occur are given by `gamma`
+        and `beta` with the probability of the environment being not excited is
+        given by `p`.
+
+        The stationary state of this channel is the diagonal density matrix
+        with probability `p` of being |0⟩ and probability `1-p` of being |1⟩.
+
+        Recall the kraus opeartors for the `GeneralizedAmplitudeDampingChannel`:
```
viathor: It'd be good to list here the Kraus operators for `ThermalRelaxationChannel`, not for `GeneralizedAmplitudeDampingChannel`.

cirq/ops/thermal_relaxation_channel.py
```diff
@@ -0,0 +1,229 @@
...
+        """
+
+        self._gamma = value.validate_probability(gamma, 'gamma')
+        self._beta = value.validate_probability(beta, 'beta')
+        self._p = value.validate_probability(p, 'p')
+
+        choi = np.array([[1. - (1. - p) * gamma, 0, 0,
+                          np.sqrt(1. - beta)], [0, (1. - p) * gamma, 0, 0],
+                         [0, 0, p * gamma, 0],
+                         [np.sqrt(1. - beta), 0, 0, 1. - p * gamma]])
```
viathor: We define other channels in terms of Kraus operators, not the Choi matrix. It'd be nice to follow this approach for consistency and to enable users to compare the specs of the channels.

----
It may also be a good idea to add a protocol to obtain the Choi matrix for a channel and then you could use that in a unit test to verify that the Kraus operators for `ThermalRelaxationChannel` really do yield the Choi matrix you calculated. One minor caveat in defining such a protocol is that Choi matrix is dependent on operator basis.

cirq/ops/thermal_relaxation_channel.py
MichaelBroughton: I can't really write down the kraus ops for the `ThermalRelaxationChannel` since the expression for their entries would be pretty ugly. I try and walk the user through the logic in the docstring and by doing this I think it's more clear than if I just wrote down some massive expressions etc, but I'm not super sure.

cirq/ops/thermal_relaxation_channel.py
MichaelBroughton: Following kraus operators is pretty hard here since they aren't easy to write down. I'm also a little warry of if we need choi matrix utilities right away (since this is the first time using choi so far in the repo). Maybe if a few more cases pop up where we need it, it'd be worth making a protocol ? :)

cirq/ops/thermal_relaxation_channel.py
```diff
@@ -0,0 +1,229 @@
...
+        basis element in the space of density matrices.
+
+        For the `ThermalRelaxationChannel` channel we simply substitute `beta`
+        in for `gamma` on the off diagonal elements of the above choi matrix
+        and then evolve the channing use the kraus operators found by sqrt
+        factorization of the choi matrix.
+
+        Args:
+            gamma: the probability of the interaction being dissipative.
+            beta: the probability of the interaction being dissipative.
```
Strilanc: beta and gamma are exactly the same according to this doc

cirq/ops/thermal_relaxation_channel.py
```diff
@@ -0,0 +1,229 @@
...
+@value.value_equality
+class ThermalRelaxationChannel(gate_features.SingleQubitGate):
+    """Dampen qubits at varying rates through non ideal disipation.
+
+    This channel models the effect of energy dissipation into the environment
+    as well as the environment depositing energy into the system at varying
+    rates.
+    """
+
+    def __init__(self, p: float, gamma: float, beta: float) -> None:
```
Strilanc: Rename this to describe what they actually are, instead of referring to the mathematical symbols. E.g. rename to `gamma` to `dissipate_probability`.

cirq/ops/thermal_relaxation_channel.py
MichaelBroughton: fixed.

cirq/ops/thermal_relaxation_channel.py
MichaelBroughton: all other channels follow the construction I have here and in the docstring I show some matrices involving `GeneralizedAmplitudeDampingChannel` and having the variables have the same name makes things pretty clear. Are you sure you want me to change them ?

Strilanc: We shouldn't have done that there. It's more serious here because the parameters are more conceptually similar. Rename 'em.

MichaelBroughton: Will do.

cirq/ops/thermal_relaxation_channel.py
```diff
@@ -0,0 +1,267 @@
...
+    """
+
+    def __init__(self, p_exchange: float, p_relaxation: float,
+                 p_dephasing: float) -> None:
+        r"""The thermal relaxation channel.
+
+        Construct a channel to model energy dissipation into the environment
+        as well as the environment depositing energy into the system. The
+        probabilities with which the energy exchange occur are given by `gamma`
+        and `beta` with the probability of the environment being not excited is
```
viathor: nit: gamma, beta and p need updating

cirq/ops/thermal_relaxation_channel.py
```diff
@@ -0,0 +1,267 @@
...
+        `p_dephasing` in for `p_relaxation` on the off diagonal elements of the
+        above choi matrix and then evolve the channel using the kraus operators
+        found by sqrt factorization of the choi matrix.
+
+        Args:
+            p_relaxation: the probability of amplitude damping/relaxation
+                occuring.
+            p_dephasing: the probability of dephasing occuring.
+            p_exchange: the probability of the qubit and environment exchanging
+                energy.
```
viathor: This is unclear. Relaxation is also an example of energy exchange between qubit and environment. Do you mean to imply that `p_exchange` subsumes `p_relaxation`? In this case we must require that `p_exchange > p_relaxation`. I think it'd be cleaner to have API without such dependency on parameters. For example, you could take `p_relaxation` and `p_excitation`.

cirq/ops/thermal_relaxation_channel.py
viathor: There is unitary freedom in the Kraus operators (see theorem 8.2 on p.372 in Mike & Ike). Can we exploit it to make the formulas more succinct?

cirq/ops/thermal_relaxation_channel.py
```diff
@@ -0,0 +1,267 @@
...
+        r"""The thermal relaxation channel.
+
+        Construct a channel to model energy dissipation into the environment
+        as well as the environment depositing energy into the system. The
+        probabilities with which the energy exchange occur are given by `gamma`
+        and `beta` with the probability of the environment being not excited is
+        given by `p`.
+
+        The stationary state of this channel is the diagonal density matrix
+        with probability `p` of being |0⟩ and probability `1-p` of being |1⟩.
```
viathor: This is also true of the generalized amplitude damping channel. I'd at least point this out to avoid misleading the reader. Better yet, remove the comment and turn it into a unit test :-)

cirq/ops/thermal_relaxation_channel_test.py
```diff
@@ -0,0 +1,139 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Iterable
+import numpy as np
+import pytest
+
+import cirq
+
+
+def apply_channel(rho: np.ndarray, chan: Iterable[np.ndarray]) -> np.ndarray:
```
viathor: We should probably use the [apply channel protocol](https://github.com/quantumlib/Cirq/blob/master/cirq/protocols/apply_channel_protocol.py).

cirq/ops/thermal_relaxation_channel_test.py
```diff
@@ -0,0 +1,139 @@
...
+    res = np.zeros_like(rho)
+    for kraus in chan:
+        res += kraus @ rho @ np.conj(kraus).T
+    return res
+
+
+def assert_same_effect(rho: np.ndarray, a: np.ndarray, b: np.ndarray) -> None:
```
viathor: Type annotations for `a` and `b` contradict type annotations on `apply_channel` above. I'm surprised mypy keeps missing these things.

cirq/ops/thermal_relaxation_channel.py
```diff
@@ -0,0 +1,267 @@
...
+             \end{aligned}
+            $$
+
+        Where each 2x2 block can be thought of as how the channel acts on each
+        basis element in the space of density matrices.
+
+        For the `ThermalRelaxationChannel` channel we simply substitute
+        `p_dephasing` in for `p_relaxation` on the off diagonal elements of the
+        above choi matrix and then evolve the channel using the kraus operators
+        found by sqrt factorization of the choi matrix.
```
viathor: This comment makes the generalized amplitude damping channel very prominent and buries the definition of the thermal relaxation channel at the very end.

We should first define the thermal relaxation channel and then point out how it relates to the generalized amplitude damping channel. If the Kraus operators for the thermal relaxation channel are indeed to complicated to write down (see another comment for a hint of what can be exploited to help simplify them), then let's define the thermal relaxation channel in terms of its Choi matrix.

I don't think we should be displaying Kraus operators for the generalized amplitude damping channel here at all.

MichaelBroughton: I've removed the kraus ops for the GAD channel and updated the docstring, let me know what you think.

cirq/ops/thermal_relaxation_channel_test.py
MichaelBroughton: fixed.

cirq/ops/thermal_relaxation_channel_test.py
MichaelBroughton: changed.

cirq/ops/thermal_relaxation_channel.py
MichaelBroughton: Changed the docs.

cirq/ops/thermal_relaxation_channel.py
MichaelBroughton: `p_exchange` is still needed because it will eventually give you the population of the excited state or how often you get a 1 when you want a 0 as time grows. I don't think we can do without `p_exchange`, but maybe I can update it to something clearer

cirq/ops/thermal_relaxation_channel.py
MichaelBroughton: fixed.

cirq/ops/thermal_relaxation_channel.py
```diff
@@ -0,0 +1,219 @@
...
+        probabilities with which the energy exchange occur are given by
+        `p_relaxation` and `p_dephasing` with the probability of the environment
+        being not excited is given by `p_exchange`.
+
+        Like the `GeneralizedAmplitudeDampingChannel`, the stationary state of
+        this channel is the diagonal density matrix with probability
+        `p_exchange` of being |0⟩ and probability `1-p_exchange` of being |1⟩.
+
+        We begin by examining the `GeneralizedAmplitudeDampingChannel`'s choi
+        matrix:
```
viathor: This should be the Choi matrix of the thermal relaxation channel. The relationship to GAD shouldn't be the focus of the docstring. We should point it out *after* we define the thermal relaxation channel.

cirq/ops/thermal_relaxation_channel.py
viathor: Isn't it true that `p_exchange = p_relaxation + p_excitation`? If it is, then we can compute p_exchange from the other two. So if we add p_excitation as a parameter then we end up with sufficient information to instantiate the channel, but without the requirement that renders some parameter combinations invalid (such as p_relaxation exceeding p_exchange).

Or am I misunderstanding the meaning of the parameters?

cirq/ops/thermal_relaxation_channel.py
```diff
@@ -0,0 +1,219 @@
...
+
+            $$
+             \begin{aligned}
+                CHO =&  \begin{bmatrix}
+                            1 - (1-p_exchange) p_relaxation & 0 & 0 &
+                                \sqrt{1 - p_relaxation} \\
+                            0 & (1-p_exchange) p_relaxation 0 & 0 \\
+                            0 & 0 & p_exchange p_relaxation & 0 \\
+                            \sqrt{1 - p_relaxation} & 0 & 0 &
+                                1 - p_exchange p_relaxation
```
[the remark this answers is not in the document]
MichaelBroughton: fixed.

cirq/ops/thermal_relaxation_channel.py
```diff
@@ -0,0 +1,219 @@
...
+                        \end{bmatrix}
+             \end{aligned}
+            $$
+
+        Where each 2x2 block can be thought of as how the channel acts on each
+        basis element in the space of density matrices.
```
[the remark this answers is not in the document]
MichaelBroughton: I removed the commend to avoid the confusion

cirq/ops/thermal_relaxation_channel.py
MichaelBroughton: Fixed the matrix. Also I can't seem to reply to your below question, but there are valid channels produced when `p_exchange != p_relaxation + p_excitation`

cirq/ops/thermal_relaxation_channel.py
```diff
@@ -0,0 +1,203 @@
...
+
+import numpy as np
+
+from cirq import protocols, value
+from cirq.ops import gate_features
+
+
+@value.value_equality
+class ThermalRelaxationChannel(gate_features.SingleQubitGate):
+    """Dampen qubits at varying rates through non ideal disipation.
```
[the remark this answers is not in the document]
MichaelBroughton: fixed

cirq/ops/thermal_relaxation_channel.py
```diff
@@ -0,0 +1,203 @@
...
+
+        # Convert to kraus using choi's theorem on CP maps.
+        self._kraus: Iterable[np.ndarray] = ()
+        for i in range(len(vals)):
+            vals += 1e-9  # fix small roundoffs.
+            if vals[i] < 0:
+                raise ValueError('Thermal relaxation with '
+                                 'p_exchange={}, p_relaxation={}, '
+                                 'p_dephasing={} breaks CP'
+                                 ' requirement.'.format(p_exchange,
```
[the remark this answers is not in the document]
MichaelBroughton: Fixed all, but one which relied on precision and would have become a lot uglier

cirq/ops/thermal_relaxation_channel.py
dkafri: Sorry to barge in and argue against this PR. I believe there is a cleaner implementation that provides an easier to understand interface as well. It's based on the fact that the amplitude damping (AD) and phase damping (PD) channels commute. So really we only need two parameters to define this channel _p_AD_ and _p_PD_. The Kraus operators for the AD and PD channels can be found in equations 3 and 5 of this [paper](https://arxiv.org/pdf/1404.3747.pdf). The Kraus operators for the combined channel are just the products of the AD and PD Kraus operators; it turns out that one of the products vanishes, so you only need 3 Kraus operators.

Although you only need two parameters to define this channel, sometimes it is very convenient (from a physics perspective) to define the channel with three: evolution time _t_ , T_1, and T_2. See equation 7 for a nice interpretation of this channel with respect to these parameters. The map from these three quantities to (_p_AD_, _p_PD_) can be computed in a straightforward way (see the sentence after equation 7). So we could also make a separate factory method for this channel based on these arguments. This is nice because it ties directly to observed noise models. It also allows for straightforward merging of consecutive channels: Use the fact that the mapping is overdetermined to get the same T_1 and T_2 parameters for both channels, but different times _t_ and _t_'. Decay for time _t_ followed by decay for time _t_' is then decay for time _t_ + _t_'.

I actually already made this alternate implementation for my own use but I'm happy to share it.

viathor: Note that currently this PR implements the composition of PD with GAD(*), not AD. These still do commute(**), but working out the minimum representation in terms of Kraus operators is a little more involved than taking the products.

That said, I do like the idea of breaking this up into simpler channels that commute.

---
(*) Generalized Amplitude Damping channel. In AD, the reservoir is at the absolute zero and consequently the Kraus operators do not include a term that excites the qubit. GAD models a reservoir with a non-zero temperature and has four Kraus operators.
(**) This is most easily seen by their action on the Bloch vector: PD shrinks rx and ry by a fixed factor and preserves rz while GAD shrinks rx and ry by a fixed factor while applying an affine function to rz.

viathor: Regarding the question of initialization parameters, the three suggested by @dkafri make sense to me. One possible concern is the potential need for infinities to specify certain probability values, but AFAICT, these should never arise in practice.

dkafri: My mistake!

dkafri: At least in the case I considered, I believe we can avoid infinities as long as T2 <= 2 * T1, which is what is expected physically.

MichaelBroughton:
@95-martin-orion

cirq/ops/thermal_relaxation_channel.py
viathor: I just noticed that we already have both PD and GAD implemented in cirq, so I wonder whether we really need the new channel introduced by this PR at all? @MichaelBroughton, do the existing channel work for you in combination? Perhaps all we need here is a factory function as suggested by @dkafri above?

balopat:
Hi All, just checking in on this one - what are the next steps with this PR, what decisions need to be made? @MichaelBroughton @viathor @95-martin-orion

MichaelBroughton:
I believe @95-martin-orion took this over and decided we didn't need anything this complicated in order to make some noise that closely resembled the device. Since then this also kind of fell through the cracks. Think we can close this.

balopat:
Thank you! Closing!

95-martin-orion:
> I believe @95-martin-orion took this over and decided we didn't need anything this complicated in order to make some noise that closely resembled the device. Since then this also kind of fell through the cracks. Think we can close this.

This is pretty much correct - the "less-complicated" resulting model was added in #2715.

PR #2487, opened 2019-11-05
Add cirq.final_density_matrix utility method

### LINKED ISSUE #2421: Add cirq.final_density_matrix method
This method would work similar to `cirq.final_wavefunction`, but it would return a density matrix instead of a wavefunction (state vector).

The method would take a circuit, possibly do some analysis to decide which simulator to use (e.g. it's faster to simulate using the wavefunction simulator and only create the density matrix at the end), run the simulation, extract a density matrix, and return it.

One notable difference vs `final_wavefunction` is that `final_density_matrix` would by default not ignore terminal measurements.

`final_density_matrix` should take an optional `noise: NOISE_MODEL_LIKE=None` parameter.

Ashalynd:
Fixes #2421


Ashalynd:
Not ready yet, the way to access the simulated final state was ok for pytest but it was actually hacky, so that part has to change. Will have another look tomorrow.

cirq/sim/mux.py
```diff
@@ -183,3 +200,74 @@ def sample_sweep(program: Union[circuits.Circuit, schedules.Schedule],
                               seed=prng)
         trial_results.append(measurements)
     return trial_results
+
+
+def final_density_matrix(
+        program: Union[circuits.Circuit, ops.Gate, ops.OP_TREE, schedules.
```
Strilanc: `program: 'cirq.CIRCUIT_LIKE'`

If we don't already have that exposed as a constant, define it equal to the given expression and expose it.

cirq/sim/mux.py
```diff
@@ -183,3 +200,74 @@ def sample_sweep(program: Union[circuits.Circuit, schedules.Schedule],
...
+                       Schedule],
+        *,
+        noise: 'cirq.NOISE_MODEL_LIKE' = None,
+        initial_state: Union[int, Sequence[Union[int, float, complex]], np.
```
Strilanc: I have a PR defining `cirq.STATE_VECTOR_LIKE` ( https://github.com/quantumlib/Cirq/pull/2376 ). We should define an analogous `cirq.DENSITY_MATRIX_LIKE` for this case.

cirq/sim/mux.py
```diff
@@ -183,3 +200,74 @@ def sample_sweep(program: Union[circuits.Circuit, schedules.Schedule],
...
+                             ndarray] = 0,
+        param_resolver: study.ParamResolverOrSimilarType = None,
+        qubit_order: ops.QubitOrderOrList = ops.QubitOrder.DEFAULT,
+        dtype: Type[np.number] = np.complex64,
+        seed: Optional[Union[int, np.random.RandomState]] = None
+) -> 'np.ndarray':
+    """Returns the density matrix resulting from simulating the circuit.
+    The terminal measurements would affect the returned result.
```
Strilanc: This statement is unclear on its own. Add some context... like

```
"""Returns the density matrix resulting from simulating the circuit.

Note that, unlike `cirq.final_wavefunction`, terminal measurements
are not omitted. Instead, all measurements are treated as sources
of decoherence (i.e. measurements do not collapse, they dephase).
"""
```

cirq/sim/mux_test.py
```diff
@@ -238,3 +238,125 @@ def test_repetitions(repetitions):
...
+
+    np.testing.assert_allclose(
+        cirq.final_density_matrix([cirq.X(a), cirq.X(b)**0.5],
+                                  qubit_order=[b, a],
+                                  noise=cirq.ConstantQubitNoiseModel(
+                                      cirq.amplitude_damp(1.0))),
+        [[1, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0]])
+
+
+def test_final_density_matrix_seed():
```
Strilanc: Hmmmm. This is a bit of a curve ball, but I think that `final_density_matrix` should treat measurement as dephasing instead of as a collapsing process. That is to say, I think the final density matrix of the circuit "H then measure" should be the maximally mixed state instead of randomly |0> or |1>.

There should perhaps be an argument to the method to decide between the two behaviors. The basic issue is that "final density matrix" should be a single thing, instead of a distribution of things. We should probably discuss this at the sync tomorrow, since it's a rather big semantic difference.

cirq/sim/mux.py
Ashalynd: Sure, that makes sense. By the way, there seems to be no state_vector() method defined for DensityMatrixTrialResult - is that by design?

Strilanc: A density matrix is more general than a state vector, so it's not possible to define such a method. (Or rather, it would need to have different semantics such as sampling a state vector from the density matrix, or failing if the density matrix did not correspond to a pure state.)

Strilanc:
Here is one way to perform a measurement without conditioning on the result:

                    meas = op.gate
                    for i, q in enumerate(op.qubits):
                        self._apply_op_channel(cirq.phase_damp(1).on(q),
                                               state,
                                               [indices[i]])

As a stop-gap, we could add a `replace_measurement_with_dephasing` option to `DensityMatrixSimulator`, and set that option to `True` when computing `final_density_matrix`.

cirq/sim/mux.py
```diff
@@ -183,3 +200,74 @@ def sample_sweep(program: Union[circuits.Circuit, schedules.Schedule],
...
+    """
+    initial_state_like = None
+    if not isinstance(initial_state, int):
+        initial_state_like = np.asarray(initial_state, dtype=dtype)
+    else:
+        initial_state_like = initial_state
+
+    noise_model = devices.NoiseModel.from_noise_model_like(noise)
+    circuit_like = _to_circuit_like(program)
+
```
Strilanc:
```
    can_do_unitary_simulation = True
    if not protocols.has_unitary(circuit_like):
        can_do_unitary_simulation = False
    if circuit_like.has_measurements():  # Including terminal measurements.
        can_do_unitary_simulation = False
    if noise_model != devices.NO_NOISE:
        can_do_unitary_simulation = False
```

cirq/sim/mux.py
```diff
@@ -183,3 +200,74 @@ def sample_sweep(program: Union[circuits.Circuit, schedules.Schedule],
...
+    if not noise_model == devices.NO_NOISE and not protocols.has_unitary(
```
Strilanc: Circuits may contain non-unitary operations and then have noise added on top. That is allowed.

cirq/sim/mux.py
Ashalynd: Thanks, that makes sense. Common suffix TrialResult makes it look like all the result classes should have a common interface (or a common ancestor), but they don't, which can be a bit confusing.

Ashalynd:
> Here is one way to perform a measurement without conditioning on the result:
> 
> ```
>             if isinstance(op.gate, ops.MeasurementGate):
>                 meas = op.gate
>                 for i, q in enumerate(op.qubits):
>                     self._apply_op_channel(cirq.phase_damp(1).on(q),
>                                            state,
>                                            [indices[i]])
> ```
> 
> As a stop-gap, we could add a `replace_measurement_with_dephasing` option to `DensityMatrixSimulator`, and set that option to `True` when computing `final_density_matrix`.

So this should apply to all measurements in the circuit, not only to the final ones? 
Why would it be desirable?

Strilanc:
> So this should apply to all measurements in the circuit, not only to the final ones?
> Why would it be desirable?

Because otherwise the output will not be a single density matrix but a probability distribution of density matrices.

cirq/sim/mux.py
Ashalynd: OK, removed the check.

Ashalynd:
> Here is one way to perform a measurement without conditioning on the result:
> 
> ```
>             if isinstance(op.gate, ops.MeasurementGate):
>                 meas = op.gate
>                 for i, q in enumerate(op.qubits):
>                     self._apply_op_channel(cirq.phase_damp(1).on(q),
>                                            state,
>                                            [indices[i]])
> ```
> 
> As a stop-gap, we could add a `replace_measurement_with_dephasing` option to `DensityMatrixSimulator`, and set that option to `True` when computing `final_density_matrix`.

Hm, there are currently two states there: `perform_measurements` True|False. It could be replaced with some enum, e.g. MeasurementStrategy, with values like: NO_MEASUREMENT, COLLAPSING_MEASUREMENT, DEPHASING_MEASUREMENT (or some better wording).

cirq/sim/mux.py
Ashalynd: Done.

cirq/sim/mux.py
Ashalynd: This might be better to do as a separate PR.

cirq/sim/mux.py
Ashalynd: Done.

cirq/sim/mux_test.py
Ashalynd: I made the changes you have suggested and added a few tests (probably not enough).

Ashalynd:
I did minimal changes for now to make it work as suggested (with dephasing option instead of randomly choosing one of the possible values during measurement). LMK what else is missing.

cirq/sim/mux_test.py
Ashalynd: Done.

cirq/sim/mux.py
Ashalynd: Done

Ashalynd:
PTAL ?

cirq/sim/density_matrix_simulator.py
```diff
@@ -120,14 +119,36 @@ def __init__(self,
...
             dtype: The `numpy.dtype` used by the simulation. One of
                 `numpy.complex64` or `numpy.complex128`
             noise: A noise model to apply while simulating.
             seed: The random seed to use for this simulator.
+            replace_measurement_with_dephasing: if True, then the simulation
+                will treat measurement as dephasing instead of collapsing
+                process.
+
+                Example:
+                q0 = cirq.LineQubit.range(1)
```
Strilanc: Examples should be written in doctest format, so that they are tested.

```
Examples:
    >>> circuit = ...
    >>> print(...)
    expected output

    >>> more stuff
```

cirq/sim/density_matrix_simulator.py
```diff
@@ -120,14 +119,36 @@ def __init__(self,
                  *,
                  dtype: Type[np.number] = np.complex64,
                  noise: 'cirq.NOISE_MODEL_LIKE' = None,
-                 seed: value.RANDOM_STATE_LIKE = None):
+                 seed: value.RANDOM_STATE_LIKE = None,
+                 replace_measurement_with_dephasing: bool = False):
```
Strilanc: Rename this to `ignore_measurement_results`, as more general kinds of measurement will not be equivalent to dephasing.

cirq/sim/density_matrix_simulator_test.py
```diff
@@ -807,22 +823,45 @@ def test_density_matrix_trial_result_repr():
                 "qubit_map={cirq.LineQubit(0): 0}))""")
 
 
+class XAsOp(cirq.Operation):
+
+    def __init__(self, q):
+        # coverage: ignore
```
Strilanc: If the test class method is not covered, just `raise NotImplementedError()` instead of bothering to implement it.

If the test class method is covered, no need for the coverage ignore.

cirq/sim/density_matrix_simulator_test.py
```diff
@@ -832,6 +871,31 @@ def _channel_(self):
...
+
+        def __init__(self, q):
+            self.q = q
+
+        @property
+        def qubits(self):
+            return self.q,
+
+        def with_qubits(self, *new_qubits):
+            # coverage: ignore
```
Strilanc: `raise NotImplementedError()`

cirq/sim/mux.py
```diff
@@ -22,11 +22,21 @@
...
+                      wave_function_simulator)
 
 if TYPE_CHECKING:
     import cirq
 
+CIRCUIT_LIKE = Union[circuits.Circuit, ops.Gate, ops.OP_TREE, schedules.
+                     Schedule]
+document(
+    CIRCUIT_LIKE,  # type: ignore
+    """A `cirq.Circuit`, a `schedules.Schedule`,  or a value that can be
```
Strilanc: `cirq.Schedule`

cirq/sim/mux.py
```diff
@@ -22,11 +22,21 @@
...
+    trivially converted into one of these.
```
Strilanc: Explain the values that can be trivially converted: a gate, an operation, and a list or tree of operations.

cirq/sim/mux.py
Strilanc: That works.

cirq/sim/mux.py
```diff
@@ -176,3 +192,88 @@ def sample_sweep(program: Union[circuits.Circuit, schedules.Schedule],
...
+        program: 'cirq.CIRCUIT_LIKE',
+        *,
+        noise: 'cirq.NOISE_MODEL_LIKE' = None,
+        initial_state: Union[int, Sequence[Union[int, float, complex]], np.
+                             ndarray] = 0,
+        param_resolver: study.ParamResolverOrSimilarType = None,
+        qubit_order: ops.QubitOrderOrList = ops.QubitOrder.DEFAULT,
+        dtype: Type[np.number] = np.complex64,
+        seed: Optional[Union[int, np.random.RandomState]] = None,
+        replace_measurement_with_dephasing: bool = True) -> 'np.ndarray':
```
Strilanc: Rename to `ignoring_measurement_results`.

cirq/sim/mux.py
```diff
@@ -176,3 +192,88 @@ def sample_sweep(program: Union[circuits.Circuit, schedules.Schedule],
...
+    """Returns the density matrix resulting from simulating the circuit.
+    Note that, unlike `cirq.final_wavefunction`, terminal measurements
+    are not omitted. Instead, all measurements are treated as sources
+    of decoherence (i.e. measurements do not collapse, they dephase).
```
Strilanc: This paragraph should mention the parameter that affects this behavior.

cirq/sim/mux.py
```diff
@@ -176,3 +192,88 @@ def sample_sweep(program: Union[circuits.Circuit, schedules.Schedule],
...
+        noise: 'cirq.NOISE_MODEL_LIKE' = None,
+        initial_state: Union[int, Sequence[Union[int, float, complex]], np.
+                             ndarray] = 0,
+        param_resolver: study.ParamResolverOrSimilarType = None,
+        qubit_order: ops.QubitOrderOrList = ops.QubitOrder.DEFAULT,
+        dtype: Type[np.number] = np.complex64,
+        seed: Optional[Union[int, np.random.RandomState]] = None,
+        replace_measurement_with_dephasing: bool = True) -> 'np.ndarray':
+    """Returns the density matrix resulting from simulating the circuit.
+    Note that, unlike `cirq.final_wavefunction`, terminal measurements
```
Strilanc: Nit: add empty line after summary line.

cirq/sim/mux.py
```diff
@@ -176,3 +192,88 @@ def sample_sweep(program: Union[circuits.Circuit, schedules.Schedule],
...
+            ordering of the computational basis states.
+        initial_state: If an int, the state is set to the computational
+            basis state corresponding to this state. Otherwise  if this
+            is a np.ndarray it is the full initial state. In this case it
+            must be the correct size, be normalized (an L2 norm of 1), and
+            be safely castable to an appropriate dtype for the simulator.
+        dtype: The `numpy.dtype` used by the simulation. Typically one of
+            `numpy.complex64` or `numpy.complex128`.
+        seed: The random seed to use for this simulator.
+        replace_measurement_with_dephasing: if True, then the simulation
```
Strilanc:
```
ignoring_measurement_results: Defaults to True. When True, the returned
    density matrix is not conditioned on any measurement results. For example,
    this effectively replaces computational basis measurement with dephasing
    noise. The result density matrix in this case should be unique. When False,
    the result will be conditioned on sampled (but unreported) measurement
    results. In this case the result may vary from call to call.

cirq/sim/density_matrix_simulator.py
Ashalynd: Done

cirq/sim/density_matrix_simulator.py
Ashalynd: Done

cirq/sim/density_matrix_simulator_test.py
Ashalynd: Done

cirq/sim/density_matrix_simulator_test.py
Ashalynd: Done

cirq/sim/mux.py
Ashalynd: Removed because cirq.Schedule is now gone.

cirq/sim/mux.py
Ashalynd: Done

cirq/sim/mux.py
Ashalynd: Added reference to ignore_measurement_results.

cirq/sim/mux.py
Ashalynd: Done

cirq/sim/mux.py
Ashalynd: Comment updated.

Ashalynd:
Ping?

vtomole:
@Ashalynd Please resolve the conflicts.

PR #2490, opened 2019-11-05
Increase prominence of doc links in readme

### LINKED ISSUE #2475: Add a clearer link to readthedocs

mpharrigan:
Fixes #2475


mpharrigan:
yeah that would be better

mpharrigan:
@vtomole I did as suggested but also included another link to the documentation with the url (worth saying!) The final count is four links to pretty much the same place :p PTAL and let me know if this looks good

mpharrigan:
I think the bullet will appeal to people skimming it

PR #2494, opened 2019-11-05
Use @dataclass for cirq.google.[De]SerializingArg

Strilanc:


cirq/google/op_deserializer.py
```diff
@@ -16,26 +16,23 @@
...
-            ('constructor_arg_name', str),
-            ('value_func', Optional[Callable[[arg_func_langs.ArgValue], Any]]),
-            ('required', bool),
-        ])):
+@dataclass
+class DeserializingArg:
     """Specification of the arguments to deserialize an argument to a gate.
 
-    Attributes:
+    Args:
```
Strilanc: I realize this looks silly. But in my experimentation sphynx only seemed to understand the Args section, not the Attributes section.

mpharrigan: you could try digging into the napoleon documentation to see if there's some flag you need to set. IIRC napoleon (used for turning google-style docstrings into sphinx/rst) has a lot of flags you can set

cirq/google/op_serializer.py
```diff
@@ -31,17 +33,11 @@
...
 
-class SerializingArg(
-        NamedTuple(
-            'SerializingArg',
-            [('serialized_name', str),
-             ('serialized_type', Type[arg_func_langs.ArgValue]),
-             ('gate_getter',
-              Union[str, Callable[['cirq.Gate'], arg_func_langs.ArgValue]]),
-             ('required', bool)])):
+@dataclass
```
mpharrigan: consider `@dataclass(frozen=True)` for immutability and hashability

cirq/google/op_deserializer.py
Strilanc: Reverted attributes->args change (will address in later PR if needed)

cirq/google/op_serializer.py
Strilanc: Done.

PR #2497, opened 2019-11-05
Various doc fixes

Strilanc:
- Add a few missing doc strings
- Fix some example sections coming before args/results
- Fix a few missing Args
- Change some type annotations to refer to exposed 'cirq.XXX' instead of local cirq.YYY.ZZZ.XXXX

### DIFF
--- a/cirq/devices/noise_model.py
+++ b/cirq/devices/noise_model.py
@@ -183,7 +183,11 @@ class _NoNoiseModel(NoiseModel):
 
 @value.value_equality
 class ConstantQubitNoiseModel(NoiseModel):
-    """Applies noise to each qubit individually at the start of every moment."""
+    """Applies noise to each qubit individually at the start of every moment.
+
+    This is the noise model that is wrapped around an operation when that
+    operation is given as "the noise to use" for a `NOISE_MODEL_LIKE` parameter.
+    """
 
     def __init__(self, qubit_noise_gate: 'cirq.Gate'):
         if qubit_noise_gate.num_qubits() != 1:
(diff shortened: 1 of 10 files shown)


cirq/linalg/transformations.py
```diff
@@ -388,10 +388,15 @@ def subwavefunction(wavefunction: np.ndarray,
...
-    \theta} |y\rangle) \otimes (|x\rangle) = (|y\rangle) \otimes (e^{i \theta}
-    |x\rangle)$ . This method randomizes the global phase of $|x\rangle$ in
-    order to avoid accidental reliance on it.
+    return $|x\rangle$. Note that $|x\rangle$ is not unique, because
+
+    $$
+        (e^{i \theta} |y\rangle) \otimes (|x\rangle)
+        = (|y\rangle) \otimes (e^{i \theta} |x\rangle)
+    $$
+
```
viathor: This is certainly an improvement, but parsing LaTeX is still a pain when looking at code (e.g. in github). In this case, the pain may be completely avoided by spelling this out in English:

> ..., because scalar multiplication may be absorbed by any factor of a tensor product.

Strilanc: done.

cirq/devices/noise_model.py
```diff
@@ -183,7 +183,11 @@ def _json_dict_(self):
 
 @value.value_equality
 class ConstantQubitNoiseModel(NoiseModel):
-    """Applies noise to each qubit individually at the start of every moment."""
+    """Applies noise to each qubit individually at the start of every moment.
+
+    This is the noise model that is wrapped around an operation when that
+    operation is given as "the noise to use" for a `NOISE_MODEL_LIKE` parameter.
```
dstrain115: What is "the noise to use"?  Is this a specific parameter?  Can we use something more precise if possible.

cirq/linalg/transformations.py
dstrain115: Agreed

Issue #2285, opened 2019-10-09
Automate API reference documentation


Strilanc:
Currently, it is very easy to accidentally forget to add a cirq method or class to the API reference. There's probably dozens missing. Here is my proposal for fixing this:

1. Introduce decorators like `@cirq._internal.doc(category='circuits')` and helper methods like `cirq._internal.documented_constant(val, category='ops')`. These decorates and methods build up a global list of documented things.
2. Add a test that fails if any top-level class or method has not been doc'd, similar to the json serialization test.
3. Have a script that basically imports cirq and dumps the list of documented things into a `.rst` file that is the api reference.

cduck:
Suggestion: Make the functions no-ops during normal execution (just like typing.cast) to reduce import-time overhead.

Strilanc:
I'm not sure how much that will save, since I'm imagining the bodies consisting of basically `some_list.append(the_thing)`. The string construction would all happen basically unconditionally.

mpharrigan:
You can probably maneuver sphinx autodoc/autosummary to do what you want without putting annotations everywhere

PR #2498, opened 2019-11-05
Add a documentation coverage test

Strilanc:
- Verify that names exposed by api.rst equals names exposed by (specific parts of) cirq
- Verify that each section in api.rst is sorted
- Re-arrange api.rst
- Fix documentation warnings in objects that now appear in api.rst

Fixes https://github.com/quantumlib/Cirq/issues/2285


cirq/devices/noise_model.py
```diff
@@ -44,14 +44,16 @@ def from_noise_model_like(cls, noise: 'cirq.NOISE_MODEL_LIKE'
...
         Returns:
-            When given `None`: returns `cirq.NO_NOISE`.
-            When given a single qubit operation: returns
-                `cirq.ConstantQubitNoiseModel(gate)`, which applies the gate to
-                every qubit before every moment.
-            When given a `cirq.NoiseModel`: just returns the input.
+            ``cirq.NO_NOISE`` when given ``None``,
+            ``cirq.ConstantQubitNoiseModel(gate)`` when given a single qubit
+            gate,
+            or the given value if it is already a ``cirq.NoiseModel``.
```
viathor: nit: this line break is weird

cirq/study/sweepable.py
```diff
@@ -68,11 +68,12 @@ def to_sweeps(sweepable: Sweepable) -> List[Sweep]:
 def to_sweep(sweep_or_resolver_list: Union['Sweep', ParamResolverOrSimilarType,
                                            Iterable[ParamResolverOrSimilarType]]
             ) -> 'Sweep':
-    """Converts the argument into a `Sweep`.
+    """Converts the argument into a ``cirq.Sweep``.
 
     Args:
-        sweep_or_resolver_list: The object to try to turn into a `Sweep`.  A
-            `Sweep`, a single `ParamResolver`, or a list of `ParamResolver`s.
+        sweep_or_resolver_list: The object to try to turn into a ``cirq.Sweep``
```
viathor: nit: need dot at end of this line (following text begins with capital letter)

docs/docs_coverage_test.py
```diff
@@ -0,0 +1,105 @@
...
+def test_public_values_equals_documented_values():
+    in_actual_api = _all_public()
+    in_api_reference = {
+        fullname for section in _api_rst_fullnames_per_section()
+        for fullname in section
+    }
+    unlisted = in_actual_api - in_api_reference
+    hidden = {
+        fullname for fullname in in_api_reference - in_actual_api
+        if not fullname.startswith('cirq.contrib.')
```
viathor: Make me wonder whether we should also complain when a `cirq.contrib` name *is* described in api.rst. This might be interpreted that we endorse contrib as part of cirq's public API, i.e. stronger than we'd want to.

Strilanc:
> It would actually be really useful to figure out how to hack sphinx to let you use a single backtick for monospace. Using two of them makes the (raw) docstrings harder to read, imo

Agree so hard. We have a ton of single-tick sequences.

Also I was surprised to find out that it doesn't link to other places in the docs.

docs/docs_coverage_test.py
Strilanc: We actually do coarsely list modules in the api reference.

PR #2499, opened 2019-11-05
Networkx requirement: bump and loosen

mpharrigan:


mpharrigan:
oh didn't know that was why. qiskit is just a contrib requirement though (?). It's kindof weird that a qiskit issue is constraining our dependencies especially on a popular, well-behaved (w.r.t. versioning) scientific python package

mpharrigan:
When we pin things, we should add a note to the requirements file explaining why

Strilanc:
We should also bump the version of qiskit; the one we're using is quite old

Agree on the documentation of why pinning

PR #2500, opened 2019-11-06
use python -m install everywhere in travis.yml

Ashalynd:


Strilanc:
Just let the merge bot do its work; it will handle merging master when needed.

PR #2507, opened 2019-11-06
Use sum instead of einsum in QAOA example

kevinsung:
@mpharrigan pointed out this possibility at https://github.com/quantumlib/Cirq/pull/2470#discussion_r342795041


mpharrigan:
I don't know; do people find this easier to understand or is the previous one better

Strilanc:
Up to you. I don't really have a preference.

kevinsung:
I like this one.

PR #2510, opened 2019-11-07
Add check/pytest-changed-files-and-incremental-coverage

Strilanc:
- Checks if changed lines are covered by tests associated with changed files (e.g. modifying x.py results in x_test.py being included)
- This check is slightly stricter than the usual incremental coverage check (because it runs a subset of all tests) but MUCH faster
- Fixed pytest-changed-files deduping by piping into uniq without first piping into sort
- Refactored run_pytest_and_incremental_coverage.py into check_incremental_coverage_annotations.py


mpharrigan:
is there a way we can make this more discoverable?

Strilanc:
@mpharrigan Added to `development.md`

check/pytest-changed-files-and-incremental-coverage
```diff
@@ -0,0 +1,101 @@
...
+################################################################################
+# Finds changed lines not covered by tests related to changed files.
+#
+# Usage:
+#     check/pytest-changed-files-and-incremental-coverage [BASE_REVISION]
+#
+# This tool is not clever; it does not understand dependencies between all
+# files. If "d/x.py" or "d/x_test.py" have been modified, it will include
+# "d/x_test.py" in the arguments given to pytest. It will furthermore include
+# "--cov=d" as an argument to pytest-cov, to restrict coverage results to
```
dstrain115: Will this run all the tests in that directory as well?

If y is in the same directory as x and you've modified x.py, will it run y_test.py?  If not, will it complain that y.py is completely untested?  That would be annoying.

Strilanc: It will not run `y_test.py`. Yes, it is slightly annoying. But running all tests in ops is much more expensive than running a single file of tests. This will encourage the test file to cover the associated source file, which I think is a good idea, but that is not a requirement.

dstrain115: But what about the second part of the question?  The doc says "restrict coverage results to directories containing changed files."  Does that mean that it will run coverage on y.py and complain that it is completely untested (even though it is covered by the unexecuted y_test.py?

PR #2512, opened 2019-11-07
Move quirk url importing out of contrib into interop

Strilanc:
- Leave exporting behind as it is not yet ready
- Add to api.rst, fix a few documentation snags.


Strilanc:
Oops. Fixed.

PR #2515, opened 2019-11-08
identity operation -> gate property

peterse:
Sometime before the v06 release the `gate` property of `IdentityOperation` was removed. This leads to a sort of confusing interface with operations
```
cirq.X(qubit).gate 
>>> cirq.X
cirq.I(qubit).gate
>>> None
```
was there a reason for this change?


Strilanc:
The gate property was never removed. The gate property was moved from GateOperation to the base class Operation, transforming this call from an error to a default implementation returning None.

PR #2516, opened 2019-11-08
Initial draft of ConvertToSycamore

dstrain115:
- Port of ConvertToSycamore from cirq_internal
- Cleaned up format, mypy, deprecated commands
- Refactored various functions to try to clean them up.
- Got rid of half-pi only rotations


cirq/google/optimizers/textbook_gates_from_sycamore.py
```diff
@@ -0,0 +1,291 @@
...
+
+    # Apply interaction part
+    yield program
+
+    # Apply final corrections
+    yield (gate(q0) for gate in optimizers.single_qubit_matrix_to_gates(a_0))
+    yield (gate(q1) for gate in optimizers.single_qubit_matrix_to_gates(a_1))
+
+
+def zztheta(theta: float, q0: ops.Qid, q1: ops.Qid) -> ops.OP_TREE:
```
kevinsung: I prefer the name `rzz`.

cirq/google/optimizers/textbook_gates_from_sycamore.py
```diff
@@ -0,0 +1,291 @@
...
+        a cirq program implementating cphase
+    rtype:
+        cirq.OP_TREE
+    """
+    yield zztheta(-theta / 4, q0, q1)
+    yield ops.Rz(theta / 2).on(q0)
+    yield ops.Rz(theta / 2).on(q1)
+
+
+def swap_zztheta(theta: float, q0: ops.Qid, q1: ops.Qid) -> ops.OP_TREE:
```
kevinsung: I prefer the name `swap_rzz`.

cirq/google/optimizers/textbook_gates_from_sycamore.py
dstrain115: Done.

cirq/google/optimizers/textbook_gates_from_sycamore.py
dstrain115: Done.

dstrain115:
Ping.  Could I get a review of this?  I would like to get this committed before the next import of google3 and to move away from the need for cirq_internal.  Thanks.

dstrain115:
> I don't like the name `textbook_gates_from_sycamore` and think it should just be merged into `convert_to_sycamore_gates`. Maybe open a low-prio follow-up ticket?

Done.  Also cleaned up a bunch of tests to make them parameterized and got rid of a bunch of unnecessary cruft.

Want to take another look before I submit?

mpharrigan:
Sorry to be a diva, but is there any way we could keep some (most) of the functions as free functions (i.e. not methods on a class)? They're really helpful atoms for crafting one's own optimizer if one is not happy with how `ConvertToSycamore` messes up all the moments xref #2553 #2406 #2461 

At present, the only usage of `self` is to call other methods which otherwise don't use `self`

dstrain115:
> Sorry to be a diva, but is there any way we could keep some (most) of the functions as free functions (i.e. not methods on a class)? They're really helpful atoms for crafting one's own optimizer if one is not happy with how `ConvertToSycamore` messes up all the moments xref #2553 #2406 #2461

I have changed them all to class methods or staticmethods.  That way you can call them without instantiating the class.  That to me seems better than having a bunch of loose functions hanging around.  What do you think?

PR #2521, opened 2019-11-10
Fix timings for Google devices

dstrain115:
- 1000ns is just for read-out.  If we want to use this timing for
non-terminal measurements, there is also a 3000ns "ring down" for
signals to depopulate.

### DIFF
--- a/cirq/google/devices/known_devices.py
+++ b/cirq/google/devices/known_devices.py
@@ -201,7 +201,7 @@ _DURATIONS_FOR_XMON = {
     'cz': 50_000,
     'xy': 20_000,
     'z': 0,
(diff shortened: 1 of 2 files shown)


cirq/google/devices/known_devices.py
```diff
@@ -201,7 +201,7 @@ def _json_dict_(self):
     'cz': 50_000,
     'xy': 20_000,
     'z': 0,
-    'meas': 1_000_000,
+    'meas': 4_000_000,  # 1000 ms for readout, 3000ns for "ring down"
```
Strilanc: I think you're mixing up microseconds and milliseconds here. Did you mean `us`?

dstrain115: I actually meant nano-seconds.  I got it correct later in the same sentence at least...

PR #2522, opened 2019-11-11
Add optional readout error correction to Quantum Volume

KevinVillela:
This adds a parity qubit for each regular qubit in the compiled circuit. Each result where these qubits don't match up will be thrown out.


cirq/contrib/quantum_volume/quantum_volume.py
```diff
@@ -91,45 +92,116 @@ def compute_heavy_set(circuit: cirq.Circuit) -> List[int]:
...
+        measurements removed.
+
+    """
+    bad_measurements = set()
+    # The circuit's mapping from physical qubit to logical qubit.
+    inverse_mapping: Dict[cirq.Qid, cirq.Qid] = {
+        v: k for k, v in mapping.items()
+    }
+
+    for final_qubit, original_qubit in mapping.items():
```
Strilanc: Does this work?

```
data = trial_result.data
bad_measurements = set()
for q, pf in parity_mapping.items():
    p = inverse_mapping[pf]
    mismatches = np.nonzero(data[str(q)] != data[str(p)])
    bad_measurements.update(mismatches)
```

Or more involved:

```
data = trial_result.data
kept = np.ones(data.shape[0], dtype=np.bool)
for q, pf in parity_mapping.items():
    p = inverse_mapping[pf]
    kept &= data[str(q)] == data[str(p)]
return data.where(kept)
```

cirq/contrib/quantum_volume/quantum_volume.py
```diff
@@ -91,45 +92,116 @@ def compute_heavy_set(circuit: cirq.Circuit) -> List[int]:
...
     """
+    mapping = compilation_result.swap_network.final_mapping()
+    circuit = compilation_result.swap_network.circuit
+
     # Add measure gates to the end of (a copy of) the circuit. Ensure that those
     # gates measure those in the given mapping, preserving this order.
     qubits = circuit.all_qubits()
     key = None
     if mapping:
+        # Add any qubits that were not explicitly mapped, so they aren't lost in
```
mpharrigan: you could do `key = lambda q: mapping.get(q, q)` if you don't actually need these in `mapping`

cirq/contrib/quantum_volume/quantum_volume.py
```diff
@@ -91,45 +92,116 @@ def compute_heavy_set(circuit: cirq.Circuit) -> List[int]:
...
+    }
+
+    for final_qubit, original_qubit in mapping.items():
+        if original_qubit in parity_mapping:
+            parity_qubit = parity_mapping[original_qubit]
+            qubit_meas = trial_result.measurements[str(final_qubit)]
+            final_parity_qubit = inverse_mapping[parity_qubit]
+            parity_meas = trial_result.measurements[str(final_parity_qubit)]
+            # Check each bit's parity qubit results to see if they are correct.
+            for idx, qubit_val in enumerate(qubit_meas):
```
mpharrigan: I bet this could be faster with vectorized numpy operations

```python
correct_strings = np.ones(dtype=bool)
for dataq, parityq in whatever:
  parity_check = (measurements[dataq] == measurements[parityq])
  correct_strings = np.logical_and(correct_strings, parity_check)
bitstrings = bitstrings[correct_strings]
```

(roughly)

cirq/contrib/quantum_volume/quantum_volume.py
```diff
@@ -338,6 +437,12 @@ def calculate_quantum_volume(
         routing_attempts: The number of times to route each model circuit onto
             the device. Each attempt will be graded using an ideal simulator
             and the best one will be used.
+        add_readout_error_correction: If true, add some parity bits that willx
```
mpharrigan: willx -> will

cirq/contrib/quantum_volume/quantum_volume.py
```diff
@@ -338,6 +437,12 @@ def calculate_quantum_volume(
...
+            later be used to detect readout error. WARNING: This makes the
+            simulator run extremely slowly for any width/depth of 4 or more,
+            probably because it doubles the circuit size. In reality, the
```
mpharrigan: You can probably strike "probably" :) This is indeed the reason

cirq/contrib/quantum_volume/quantum_volume.py
```diff
@@ -264,6 +360,8 @@ def execute_circuits(
         compiler: An optional function to compiler the model circuit's
             gates down to the target devices gate set and the optimize it.
         repetitions: The number of bitstrings to sample per circuit.
+        add_readout_error_correction: If true, add some parity bits that willx
```
mpharrigan: willx -> will

cirq/contrib/quantum_volume/quantum_volume.py
```diff
@@ -162,34 +237,55 @@ def compile_circuit(
...
     compiled_circuit = circuit.copy()
+
+    # Optionally add some the parity check bits.
+    parity_map: Dict[cirq.Qid, cirq.Qid] = {}  # original -> parity
+    if add_readout_error_correction:
+        parity_ops = cirq.Moment()
+        # Sort just to make it deterministic.
+        for idx, qubit in enumerate(sorted(compiled_circuit.all_qubits())):
+            # For each qubit, create a new qubit that will serve as its parity
+            # check. This parity bit is initialized to 0 and then CNOTed with
```
mpharrigan: readout error is asymmetric: p(0|0) > p(1|1). You can account for that by encoding into the single-excitation subspace; i.e. use an inverse-controlled not: `X CNOT X` to encode 0 -> 01 and 1 -> 10 so you (should) always readout one 1

cirq/contrib/quantum_volume/quantum_volume.py
KevinVillela: Nice, done.

cirq/contrib/quantum_volume/quantum_volume.py
KevinVillela: Done.

cirq/contrib/quantum_volume/quantum_volume.py
KevinVillela: Haha good to know, thanks.

cirq/contrib/quantum_volume/quantum_volume.py
KevinVillela: Done.

cirq/contrib/quantum_volume/quantum_volume.py
KevinVillela: Neat, I did not know that - Fixed. Out of curiosity, why is readout error asymmetric? Is that just for our device, or for all devices?

Also, probably a dumb question, but could I instead do `X(ancilla) CNOT(physical, ancilla)` instead of `X(physical) CNOT(physical, ancilla) X(physical)` ?

cirq/contrib/quantum_volume/quantum_volume.py
KevinVillela: Yes, something similar to that does. Thanks, much cleaner now!

cirq/contrib/quantum_volume/quantum_volume.py
KevinVillela: Agreed that this should be improved - @Strilanc has a similar idea that uses the Pandas DataFrame directly, so I went with that.

cirq/contrib/quantum_volume/quantum_volume.py
mpharrigan: Readout takes longer than gates (~1000ns vs 10) so the qubit can undergo T1 decay during readout (see cirq.contrib.noise_models). I think it's fairly common in superconducting architectures like ours. 

You could do the suggested alternate, but please make sure that the X(ancilla) is done as late as possible. If cirq is left to its own devices, the X will happen at the beginning, the qubit will be hanging out in |1> and could more readily undergo T1 decay

PR #2523, opened 2019-11-11
DFE with physical simulations

tonybruguier-google:
Attempt to have direct fidelity estimation without using the density matrix. Instead, use simulator.run().

Also, while writing the code, I found two mistakes. First is that the second run of compute_characteristic_function() did not use the 'noise' parameter. Second, the density matrix was re-computed 4**n_qbit times, which I think is wasteful. Sorry about these two mistakes, I think they are now fixed.

Also, sorry for the delay. I had decided to re-read my textbook, because my first PR required a lot of help from the reviewers.

I have been working with Matthew Harrigan, but happy to receive feedback from anyone.


examples/direct_fidelity_estimation.py
```diff
@@ -52,54 +58,108 @@ def compute_characteristic_function(circuit: cirq.Circuit,
...
 
-    assert len(pauli_traces) == 4**n_qubits
+    if physical:
+        assert len(pauli_traces) == n_trials
+    else:
+        assert len(pauli_traces) == 4**n_qubits
 
     p = [x['Pr_i'] for x in pauli_traces]
-    assert np.isclose(sum(p), 1.0, atol=1e-6)
+    if physical:
```
Strilanc: If it's physical it's allowed to equal 0 but otherwise it has to be close to 1? This seems very strange.

examples/direct_fidelity_estimation.py
```diff
@@ -52,54 +58,108 @@ def compute_characteristic_function(circuit: cirq.Circuit,
...
+        assert np.isclose(sum(p), 1.0, atol=1e-6)
 
     # The package np.random.choice() is quite sensitive to probabilities not
     # summing up to 1.0. Even an absolute difference below 1e-6 (as checked just
     # above) does bother it, so we re-normalize the probs.
     inv_sum_p = 1 / sum(p)
     norm_p = [x * inv_sum_p for x in p]
 
+    simulator = cirq.DensityMatrixSimulator(noise=noise)
+    if physical:
```
Strilanc: You say `if physical` three times in this method. This suggests you should have two methods. The large common parts of this method can be extracted into helper methods. You also say `if physical` in other places... if this were not an example, I would recommend extracting a class with methods corresponding to each of these `if` blocks.

examples/direct_fidelity_estimation.py
```diff
@@ -52,54 +58,108 @@ def compute_characteristic_function(circuit: cirq.Circuit,
     return trace, prob
 
 
+def sample_pauli_states(n_qubits: int, n_trials: int, physical: bool):
+    pauli_gates = [cirq.I, cirq.X, cirq.Y, cirq.Z]
+    if physical:
```
Strilanc: `physical` is not a very descriptive name. For example, it seems to control whether or not you are sampling pauli strings to try instead of trying all of them. Maybe `sampled`? `sparse`?

examples/direct_fidelity_estimation.py
```diff
@@ -29,15 +31,19 @@ def build_circuit():
...
-    # rho or sigma in https://arxiv.org/pdf/1104.3835.pdf
-    density_matrix = cast(cirq.DensityMatrixTrialResult,
-                          simulator.simulate(circuit)).final_density_matrix
+    if physical:
+        assert density_matrix is None
+        density_matrix = np.zeros([d, d], dtype=np.complex64)
+        res = simulator.run(circuit, repetitions=1).measurements['y']
+        # Convert binary to index.
+        idx = sum([res[0][i] * 2**i for i in range(len(res))])
+        density_matrix[idx][idx] = 1
```
mpharrigan: this isn't quite right

One of the points of DFE in contrast to regular tomography is you don't have to ever figure out the full density matrix. Using the expressions presented in the papers, you only need to know quantities of the form Tr(P_i rho) which you can estimate based on sampling using a scheme like that found in `PauliSumCollector`. 

The only complication is figuring out the value of `prob` (aka Pr(i)). The papers suggest that if you constrain your circuit to be only clifford operations, you can generate all (input) pauli strings and efficiently simulate how the circuit will transform them. Then you have a collection of output pauli strings uniformly distributed (i.e. Pr(i) = constant). For this PR, maybe it will be easier to still calculate `prob` via simulation; but the trace can be found by estimating operator expectation values

examples/direct_fidelity_estimation.py
tonybruguier-google: No longer present, given Matthew's comments. Sorry and thanks for catching this.

examples/direct_fidelity_estimation.py
tonybruguier-google: Changed to sampled.

examples/direct_fidelity_estimation.py
tonybruguier-google: With Matthew's comment, I think it's only in one place now. Would you like to have the section of the code under the single 'if sampled:' in its separate function. I could have a shared 'for' loop.

Not feeling strongly about it.

examples/direct_fidelity_estimation.py
tonybruguier-google: Thanks. Doing Pr(i) estimation in simulation sounds good for now.

Added the call to PauliSumCollector. I wasn't sure whether I should use energy of its square root. I think it's without the square root, but not sure.

I also looked at the code just to understand what's happening, and it seems that another collector could be an actual, physical measurement, right?

tonybruguier-google:
Hello,

Do you have any comments on this PR? No hurry at all, but I am still interested in working on it.

examples/direct_fidelity_estimation.py
```diff
@@ -52,54 +58,108 @@ def compute_characteristic_function(circuit: cirq.Circuit,
...
 
 
+def sample_pauli_states(n_qubits: int, n_trials: int, physical: bool):
+    pauli_gates = [cirq.I, cirq.X, cirq.Y, cirq.Z]
+    if physical:
+        return [[pauli_gates[j]
+                 for j in np.random.choice(4, n_qubits)]
+                for i in range(n_trials)]
+    else:
+        return itertools.product(pauli_gates, repeat=n_qubits)
```
mpharrigan: "physical" is a bad name. You can do exhaustive or random sampling with or without a simulator.

examples/direct_fidelity_estimation.py
```diff
@@ -52,54 +58,108 @@ def compute_characteristic_function(circuit: cirq.Circuit,
...
+
+    for P_i in sample_pauli_states(n_qubits, n_trials, physical):
+        rho_i, Pr_i = compute_characteristic_function(circuit, P_i, qubits,
+                                                      physical, simulator,
+                                                      density_matrix)
         pauli_traces.append({'P_i': P_i, 'rho_i': rho_i, 'Pr_i': Pr_i})
 
-    assert len(pauli_traces) == 4**n_qubits
+    if physical:
+        assert len(pauli_traces) == n_trials
```
mpharrigan: unnecessary assertion, given that it needs an if statement

examples/direct_fidelity_estimation.py
```diff
@@ -52,54 +58,108 @@ def compute_characteristic_function(circuit: cirq.Circuit,
...
+    else:
+        assert len(pauli_traces) == 4**n_qubits
 
     p = [x['Pr_i'] for x in pauli_traces]
-    assert np.isclose(sum(p), 1.0, atol=1e-6)
+    if physical:
+        # Sometimes, the probabilities add up to 0.0 just because of randomness
```
mpharrigan: again, this is because of the way we sample the paulis and has nothing to do with whether a real device is used

examples/direct_fidelity_estimation.py
```diff
@@ -52,54 +58,108 @@ def compute_characteristic_function(circuit: cirq.Circuit,
...
     # above) does bother it, so we re-normalize the probs.
     inv_sum_p = 1 / sum(p)
     norm_p = [x * inv_sum_p for x in p]
 
+    simulator = cirq.DensityMatrixSimulator(noise=noise)
+    if physical:
+        density_matrix = None
+    else:
+        # sigma in https://arxiv.org/pdf/1104.3835.pdf
+        density_matrix = cast(cirq.DensityMatrixTrialResult,
```
mpharrigan: duplicated code? why is this here twice

examples/direct_fidelity_estimation.py
```diff
@@ -29,15 +31,19 @@ def build_circuit():
...
+                                    density_matrix: Optional[np.ndarray]):
     n_qubits = len(P_i)
     d = 2**n_qubits
 
-    simulator = cirq.DensityMatrixSimulator()
-    # rho or sigma in https://arxiv.org/pdf/1104.3835.pdf
-    density_matrix = cast(cirq.DensityMatrixTrialResult,
-                          simulator.simulate(circuit)).final_density_matrix
+    if physical:
+        assert density_matrix is None
```
mpharrigan: this function is mostly in an if statement and one of the arguments is unused. consider splitting into two functions

examples/direct_fidelity_estimation.py
mpharrigan: I don't see any changes (i.e. using pauli sum collector)

tonybruguier-google: Sorry about this. Let me try to upload again. I had added the Pauli sum collector.

examples/direct_fidelity_estimation.py
tonybruguier-google: Merge issue here too. I had gotten rid of all 'physical' mentions.

examples/direct_fidelity_estimation.py
tonybruguier-google: Gone.

examples/direct_fidelity_estimation.py
tonybruguier-google: Gone.

examples/direct_fidelity_estimation.py
tonybruguier-google: I have a compute_characteristic_function() function that only does one thing. Would you like me to put the code that uses PauliSumCollector in its own function too? Right now, it's not because it is only called once.

examples/direct_fidelity_estimation.py
tonybruguier-google: Merge issue here I think.

tonybruguier-google:
Sorry for the noise in this review. I think what happened is that I replied to your comments a few minutes *before* I had pushed the changes. I did the reverse here, and I do see the call to PauliSumCollector on line 105. Sorry again.

examples/direct_fidelity_estimation.py
```diff
@@ -82,35 +88,93 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
...
+            sigma_i = p.estimated_energy()
+            assert np.isclose(sigma_i.imag, 0.0, atol=1e-6)
+            sigma_i = sigma_i.real
+
+            fidelity += Pr_i * sigma_i / rho_i
+
+        return fidelity
+    else:
+        # sigma in https://arxiv.org/pdf/1104.3835.pdf
+        density_matrix = cast(cirq.DensityMatrixTrialResult,
```
mpharrigan: nit: re-use of the variable name `density_matrix` from above

examples/direct_fidelity_estimation.py
```diff
@@ -82,35 +88,93 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
     inv_sum_p = 1 / sum(p)
     norm_p = [x * inv_sum_p for x in p]
```
mpharrigan: these should probably be done with numpy operations for speed and readability

```python
p = np.asarray([...])
p /= np.sum(p)
```

examples/direct_fidelity_estimation.py
```diff
@@ -53,22 +50,31 @@ def compute_characteristic_function(circuit: cirq.Circuit,
 
 
 def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
-                               noise: cirq.NoiseModel, n_trials: int):
+                               noise: cirq.NoiseModel, sampled: bool,
+                               n_trials: int, samples_per_term: int):
```
mpharrigan: What should I pass for `samples_per_term` if `sampled=False`?

examples/direct_fidelity_estimation.py
```diff
@@ -82,35 +88,93 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
...
+            fidelity += Pr_i * sigma_i / rho_i
+
+        return fidelity
+    else:
+        # sigma in https://arxiv.org/pdf/1104.3835.pdf
+        density_matrix = cast(cirq.DensityMatrixTrialResult,
+                              simulator.simulate(circuit)).final_density_matrix
+
+        fidelity = 0.0
+        for _ in range(n_trials):
```
mpharrigan: there's a decent amount of shared code in these two branches. Is there any way to factor it out?

examples/direct_fidelity_estimation.py
tonybruguier-google: Good point. Changed this variable name and also changed the other variable name, so that it's more explicit.

examples/direct_fidelity_estimation.py
tonybruguier-google: Good point. Removed the flag --sampled and now use only --samples_per_term to decide whether to sample (if > 0).

examples/direct_fidelity_estimation.py
tonybruguier-google: Refactored the code to have a separate function.

tonybruguier-google:
Thanks. PTAL.

tonybruguier-google:
Friendly ping. No hurry at all.

In the meantime, would it make sense for me to study this?
https://arxiv.org/abs/0907.2833

That would be to address the TODO that this PR adds.

mpharrigan:
For a reference, I'd check out `cirq.sim.clifford` both for the code (which you will likely use) and the references it contains

examples/direct_fidelity_estimation.py
```diff
@@ -49,9 +50,32 @@ def compute_characteristic_function(circuit: cirq.Circuit,
...
+
+    pauli_string = cirq.PauliString(dict(zip(qubits, P_i)))
+    qubit_map = dict(zip(qubits, range(n_qubits)))
+
+    p = cirq.PauliSumCollector(circuit=circuit,
+                               observable=pauli_string,
+                               samples_per_term=samples_per_term)
+
+    completion = p.collect_async(sampler=simulator)
+    assert await completion is None
```
mpharrigan: Is there a reason you do this? Why not just `await p.collect_async(sampler=simulator)`

examples/direct_fidelity_estimation.py
```diff
@@ -52,65 +50,129 @@ def compute_characteristic_function(circuit: cirq.Circuit,
...
+    assert np.isclose(sigma_i.imag, 0.0, atol=1e-6)
+    sigma_i = sigma_i.real
+
+    return sigma_i
+
+
 def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
-                               noise: cirq.NoiseModel, n_trials: int):
+                               noise: cirq.NoiseModel, n_trials: int,
+                               samples_per_term: int):
```
mpharrigan: Add a docstring. Especially note that setting `samples_per_term` to `0` changes the behavior of this function

examples/direct_fidelity_estimation.py
```diff
@@ -52,65 +50,129 @@ def compute_characteristic_function(circuit: cirq.Circuit,
...
     # inside the variable 'pauli_traces'.
     pauli_traces = []
+
+    simulator = cirq.DensityMatrixSimulator()
+    # rho in https://arxiv.org/pdf/1104.3835.pdf
+    clean_density_matrix = cast(
+        cirq.DensityMatrixTrialResult,
+        simulator.simulate(circuit)).final_density_matrix
+
+    # TODO(tonybruguier): Sample the Pauli states more efficenitly when the
```
mpharrigan: prefer opening a github issue and linking that here instead of a single person's name

examples/direct_fidelity_estimation.py

examples/direct_fidelity_estimation.py
tonybruguier-google: I had seen it done this way in the unit test. Changed.

tonybruguier-google:
Thanks for the review. I tried to address your comments.

Do you have advice on how to resolve Travis errors? They seem to both come from asyncio.run() which apparently doesn't exist for it (but it runs fine for me)?

PTAL.

mpharrigan:
`asyncio.run` was introduced in python 3.7 as syntactic sugar. To support python 3.6, change it to

```python
loop = asyncio.get_event_loop()
# Blocking call which returns when the hello_world() coroutine is done
loop.run_until_complete(hello_world())
loop.close()
```

see also https://docs.python.org/3.6/library/asyncio-task.html#example-hello-world-coroutine vs https://docs.python.org/3.7/library/asyncio-task.html#coroutines

tonybruguier-google:
Thanks. I also made the Travis test pass.

PTAL.

mpharrigan:
any ideas @Strilanc ?

tonybruguier-google:
Hello,
Not urgent, but is there an update on the merging? It seems that the CLA was missing, which is strange but appears to have been resolved.
Thanks in advance.

tonybruguier-google:
Thanks. Changed the comments.

tonybruguier-google:
> Link to abstracts instead of pdfs. (eg, https://arxiv.org/abs/1104.3835 instead of https://arxiv.org/pdf/1104.3835.pdf)

examples/direct_fidelity_estimation.py
```diff
@@ -30,87 +33,160 @@ def build_circuit():
...
-    prob = trace * trace / d  # Pr(i) in https://arxiv.org/pdf/1104.3835.pdf
+    prob = trace * trace / d  # Pr(i) in https://arxiv.org/abs/1104.3835
 
     return trace, prob
 
 
-def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
-                               noise: cirq.NoiseModel, n_trials: int):
-    # n_trials is upper-case N in https://arxiv.org/pdf/1104.3835.pdf
+async def estimate_characteristic_function(
```
vtomole: Add docstring

tonybruguier-google:
I added some comments and the docstring. I am nowhere near an expert, so please let me know if anything is inaccurate. Thanks for helping me.

examples/direct_fidelity_estimation.py

tonybruguier-google:
Fixed the

tonybruguier-google:
Fixed the check errors

examples/direct_fidelity_estimation.py
```diff
@@ -1,17 +1,30 @@
 """Implements direct fidelity estimation.
 
+Fidelity between the desired pure state rho and the actual state sigma is
+defined as:
+F(rho, sigma) = Tr [rho sigma]
```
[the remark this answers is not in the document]
tonybruguier-google: That's right, no difference intended. Changed to () for clarity.

tonybruguier-google:
Not a problem at all. I am not very good with git yet, so I was mostly confused. Thank you.

tonybruguier-google:
Travis error appears unrelated.

Issue #2425, opened 2019-10-28
Add backward-compatibility tests for json serialization


Strilanc:
Currently we do some round trip testing, but there's nothing that checks that we aren't changing the json serialization from version to version in a way that breaks the ability to parse previously serialized things.

There are two main criteria I have in mind:

1. It should somehow be clear to developers that it is not okay to change the test data json, since that would defeat the purpose of a compatibility test.

2. It should re-use the mechanism in `json_test` for ensuring we are testing every exposed class.

PR #2527, opened 2019-11-12
Refactor json_test's TEST_OBJECTS value into test data files

Strilanc:
This spreads out the gigantic list a bit, but mainly it allows us a format for adding backwards compatibility tests (where the repr file contains the intended result, instead of the object that initially produced the JSON).

The test files were generated by iterating over the test object and using cirq.to_json and repr.

Fixes https://github.com/quantumlib/Cirq/issues/2425

This is maybe the third time I find myself refactoring this code. I wonder what the next one will be?


cirq/protocols/json_test.py
```diff
@@ -667,6 +518,88 @@ def test_to_from_strings():
         cirq.read_json(io.StringIO(), json_text=x_json_text)
 
 
+def _eval_repr_data_file(path: pathlib.Path):
+    return eval(path.read_text(), {
```
mpharrigan: this seems mildly dangerous

Strilanc: I agree in principle... but to be pragmatic we're _already_ parsing files and running them as python when we invoke pytest. I don't see any additional risk from this... if there was, it would be of the style "a tool was setup to detect people replacing code with delete everything commands, but the tool ignored things that didn't end with .py".

mpharrigan: yup, logically this isn't much different than the gaping security hole of python / pytest / everything already :)

PR #2528, opened 2019-11-12
Initial draft for convert_to_sqrt_iswap

dstrain115:
- Optimizer to decompile to ISWAP ** -0.5 gates.
- Ported from cirq_internal
- Modified to accept symbol input as well.


cirq/google/optimizers/convert_to_sqrt_iswap.py
```diff
@@ -0,0 +1,290 @@
...
+
+    def _convert_one(self, op: cirq.Operation) -> cirq.OP_TREE:
+        """
+        Decomposer intercept:  Let cirq decompose one-qubit gates,
+        intercept on 2-qubit gates if they are known gates.
+        """
+        if isinstance(op, cirq.GlobalPhaseOperation):
+            return []
+
+        gate = cast(cirq.GateOperation, op).gate
```
maffoo: The cast is not needed since `gate` is always defined on `Operation`.

cirq/google/optimizers/convert_to_sqrt_iswap.py
```diff
@@ -0,0 +1,290 @@
...
+
+        if len(op.qubits) != 2:
+            return NotImplemented
+
+        if isinstance(gate, cirq.CZPowGate):
+            turns = cast(cirq.CZPowGate, gate).exponent
```
maffoo: I think mypy understands after an isinstance check that `gate` is an instance of `cirq.CZPowGate` so this cast should not be needed. Same thing for the other cases below. Also, consider calling the variable `exponent` instead of `turns`.

cirq/google/optimizers/convert_to_sqrt_iswap.py
```diff
@@ -0,0 +1,290 @@
...
+    """Check if the given operation is compatible with the SQRT_ISWAP_FISHFOOD
+    gate set.
+
+    Args:
+        op: Input operation.
+
+    Returns:
+        True if the operation is native to the gate set, false otherwise.
+    """
+    return (isinstance(op, cirq.GateOperation) and
```
maffoo: This check is not needed since `Operation` always has a `gate` attribute.

cirq/google/optimizers/convert_to_sqrt_iswap.py
```diff
@@ -0,0 +1,290 @@
...
+            (is_basic_gate(op.gate) or is_sqrt_iswap(op.gate)))
+
+
+def is_sqrt_iswap(gate: cirq.Gate) -> bool:
+    """Checks if this is a ± sqrt(iSWAP) gate specified using either
+    ISwapPowGate or with the equivalent FSimGate.
+    """
+    if (isinstance(gate, cirq.FSimGate) and
+            not isinstance(gate.theta, sympy.Symbol) and
```
maffoo: `gate.theta` could be a more complicated expression than just a symbol. I think this should check `sympy.Basic` instead, though it's probably safer to use `cirq.is_parametrized`. Same thing for the `ISwapPowGate` case below.

cirq/google/optimizers/convert_to_sqrt_iswap.py
```diff
@@ -0,0 +1,290 @@
...
+            return fsim_gate(op.qubits[0], op.qubits[1],
+                             cast(cirq.FSimGate, gate).theta,
+                             cast(cirq.FSimGate, gate).phi)
+
+        return NotImplemented
+
+    def convert(self, op: cirq.Operation) -> List[cirq.Operation]:
+
+        def on_stuck_raise(bad):
+            return TypeError("Don't know how to work with {!r}. "
```
maffoo: nit: could use f-string instead of `.format`.

cirq/google/optimizers/convert_to_sqrt_iswap.py
```diff
@@ -0,0 +1,290 @@
...
+
+    Returns:
+        True if the gate is native to the gate set, false otherwise.
+    """
+    return isinstance(gate, (cirq.MeasurementGate, cirq.PhasedXPowGate,
+                             cirq.XPowGate, cirq.YPowGate, cirq.ZPowGate))
+
+
+def cphase_to_sqrt_iswap(a, b, turns):
+    """Implement a C-Phase gate using two CZ gates and single-qubit
```
maffoo: "using two SqrtISwap gates"

cirq/google/optimizers/convert_to_sqrt_iswap.py
```diff
@@ -0,0 +1,290 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import List, cast
+
+import numpy as np
+
+import cirq
+import sympy
+
+SQRT_ISWAP = cirq.ISWAP**-0.5
```
maffoo: Note that the device can handle both `cirq.ISWAP**0.5` and `cirq.ISWAP**-0.5` natively, since these are related by `ISWAP(a, b)**0.5 = Z(a) ISWAP(a, b)**-0.5 Z(b)`. This latter sequence of an iswap conjugated by Zs on one qubit appears in various places in the decompositions below, so we might be able to simplify things by defining:
```python
SQRT_ISWAP = cirq.ISWAP**0.5
SQRT_ISWAP_INV = cirq.ISWAP**-0.5
```
and then using both of these in the decompositions. This should avoid confusion about why we're choosing the negative exponent for "SQRT_ISWAP", and it also might make some of the decompositions more symmetrical.

cirq/google/optimizers/convert_to_sqrt_iswap.py
```diff
@@ -0,0 +1,290 @@
...
+    yield SQRT_ISWAP(a, b)
+    yield cirq.Z(a)**(-turns / 2 + 1)
+    yield cirq.Z(b)**(turns / 2)
+    yield SQRT_ISWAP(a, b)
+    yield cirq.Z(a)**0.25
+    yield cirq.Z(b)**-0.25
+
+
+def swap_to_sqrt_iswap(a, b, turns):
+    r"""Implement the evolution of the hopping term using two sqrt_iswap gates
```
maffoo: nit: does this need to be a raw string? I would think that's only necessary if there are literal backslashes, which isn't the case here.

cirq/google/optimizers/convert_to_sqrt_iswap_test.py
```diff
@@ -0,0 +1,121 @@
...
+        (cirq.CZ, 8),
+        (cirq.SWAP, 13),
+        (cirq.CNOT, 9),
+        (cirq.ISWAP**0.5, 1),
+        (cirq.ISWAP**-0.5, 1),
+        (cirq.ISwapPowGate(exponent=0.5), 1),
+        (cirq.ISwapPowGate(exponent=-0.5), 1),
+        (cirq.FSimGate(theta=np.pi / 4, phi=0), 1),
+    ] +
+    [(cirq.SwapPowGate(exponent=a), 13) for a in np.linspace(0, 2 * np.pi, 20)]
```
maffoo: nit: instead of concatenating lists, might be cleaner to splat iterables into a single list:
```python
@pytest.mark.parametrize(
    'gate, expected_length',
    [
        (cast(cirq.Gate, cirq.ISWAP), 7),
        ...
        (cirq.FSimGate(theta=np.pi / 4, phi=0), 1),
        *[(cirq.CZPowGate(exponent=a), 8) for a in np.linspace(0, 2 * np.pi, 20)],
        *[(cirq.ISwapPowGate(exponent=a), 5) for a in np.linspace(0, 2 * np.pi, 20)],
        ...
    ])
```
I have no idea which of these the formatter will handle better, so feel free to ignore if this doesn't help.

cirq/google/optimizers/convert_to_sqrt_iswap_test.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # Check if unitaries are the same
+    for val in np.linspace(0, 2 * np.pi, 20):
+        assert _unitaries_allclose(
+            cirq.resolve_parameters(original_circuit, {'t': val}),
+            cirq.resolve_parameters(converted_circuit, {'t': val}))
+
+
+def test_cphase():
+    """Test if the sqrt_iswap synthesis for a givens rotation is correct"""
```
maffoo: docstring is wrong.

cirq/google/optimizers/convert_to_sqrt_iswap.py
dstrain115: Done.

cirq/google/optimizers/convert_to_sqrt_iswap.py
dstrain115: Weird.  I totally thought mypy complained about this, but I guess not.

cirq/google/optimizers/convert_to_sqrt_iswap.py
dstrain115: Done.

cirq/google/optimizers/convert_to_sqrt_iswap.py
dstrain115: Done.

cirq/google/optimizers/convert_to_sqrt_iswap.py
dstrain115: Done.

cirq/google/optimizers/convert_to_sqrt_iswap.py
dstrain115: Done.

cirq/google/optimizers/convert_to_sqrt_iswap_test.py
dstrain115: Done.

cirq/google/optimizers/convert_to_sqrt_iswap_test.py
dstrain115: Good idea.  It looks a lot better.

cirq/google/optimizers/convert_to_sqrt_iswap.py
dstrain115: Done.

cirq/google/optimizers/convert_to_sqrt_iswap.py
```diff
@@ -0,0 +1,283 @@
...
+        Decomposer intercept:  Let cirq decompose one-qubit gates,
+        intercept on 2-qubit gates if they are known gates.
+        """
+        if isinstance(op, cirq.GlobalPhaseOperation):
+            return []
+
+        gate = op.gate
+
+        if len(op.qubits) != 2:
+            return NotImplemented
```
maffoo: nit: after this check you could do something like `q0, q1 = op.qubits` and then use `q0` and `q1` to simplify the expressions below.

cirq/google/optimizers/convert_to_sqrt_iswap.py
```diff
@@ -0,0 +1,283 @@
...
+            return iswap_to_sqrt_iswap(op.qubits[0], op.qubits[1],
+                                       gate.exponent)
+        if isinstance(gate, cirq.FSimGate):
+            return fsim_gate(op.qubits[0], op.qubits[1], gate.theta, gate.phi)
+
+        return NotImplemented
+
+    def convert(self, op: cirq.Operation) -> List[cirq.Operation]:
+
+        def on_stuck_raise(bad):
```
maffoo: I'm not sure the performance implication of defining a local function like this, but it might be faster to make this an internal method on the class instead so it doesn't get redefined on each call. Might be worth benchmarking to see if that makes any difference when optimizing typical circuits.

cirq/google/optimizers/convert_to_sqrt_iswap.py
dstrain115: Good suggestion, done.

cirq/google/optimizers/convert_to_sqrt_iswap.py
dstrain115: Not sure that it matters, but we don't really need an inline function anyway, so I just moved it out.

PR #2532, opened 2019-11-13
Add single qubit readout calibration

kevinsung:


cirq/experiments/single_qubit_readout_calibration.py
```diff
@@ -0,0 +1,112 @@
...
+    """Result of estimating single qubit readout error.
+
+    Attributes:
+        zero_state_errors: A dictionary from qubit to probability of measuring
+            a 1 when the qubit is initialized to |0⟩.
+        one_state_errors: A dictionary from qubit to probability of measuring
+            a 0 when the qubit is initialized to |1⟩.
+        repetitions: The number of repetitions that were used to estimate the
+            probabilities.
+        timestamp: A timestamp for the time the data was taken.
```
mpharrigan: describe units

cirq/experiments/single_qubit_readout_calibration.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+    zero_state_errors: Dict[ops.Qid, float]
+    one_state_errors: Dict[ops.Qid, float]
+    repetitions: int
+    timestamp: float
```
mpharrigan: float??

cirq/experiments/single_qubit_readout_calibration.py
```diff
@@ -0,0 +1,112 @@
...
+    def __repr__(self):
+        return ('cirq.experiments.SingleQubitReadoutCalibrationResult('
+                f'zero_state_errors={self.zero_state_errors!r}, '
+                f'one_state_errors={self.one_state_errors!r}, '
+                f'repetitions={self.repetitions!r}, '
+                f'timestamp={self.timestamp!r})')
+
+
+def estimate_single_qubit_readout_errors(
+        sampler: work.Sampler,
```
mpharrigan: I think the new convention is to do

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
  import cirq

def thing(sampler: 'cirq.Sampler'):
  pass
```

cirq/experiments/single_qubit_readout_calibration.py
```diff
@@ -0,0 +1,112 @@
...
+        repetitions: The number of measurement repetitions to perform.
+
+    Returns:
+        A SingleQubitReadoutCalibrationResult storing the readout error
+        probabilities as well as the number of repetitions used to estimate
+        the probabilties.
+    """
+    qubits = list(qubits)
+
+    zeros_circuit = circuits.Circuit(ops.measure_each(*qubits, key_func=repr))
```
mpharrigan: why `repr` and not `str`?

cirq/experiments/single_qubit_readout_calibration.py
```diff
@@ -0,0 +1,112 @@
...
+    ones_circuit = circuits.Circuit(ops.X.on_each(*qubits),
+                                    ops.measure_each(*qubits, key_func=repr))
+
+    zeros_result = sampler.run(zeros_circuit, repetitions=repetitions)
+    ones_result = sampler.run(ones_circuit, repetitions=repetitions)
+    timestamp = time.time()
```
mpharrigan: 1. you could put more timestamps (before, middle, after)
2. At least document which time is stamped in the result object

cirq/experiments/single_qubit_readout_calibration.py
kevinsung: The repr is required to be unique for different qubits.

cirq/experiments/single_qubit_readout_calibration.py
kevinsung: I think having 3 timestamps is excessive; people will want to just look at one. The exact time down to the second shouldn't matter that much. But sure, I'll document it.

cirq/experiments/single_qubit_readout_calibration.py
```diff
@@ -0,0 +1,112 @@
...
+
+import dataclasses
+import time
+
+import numpy as np
+
+from cirq import circuits, ops, work
+
+
+@dataclasses.dataclass
```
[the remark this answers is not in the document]
kevinsung: Done.

cirq/experiments/single_qubit_readout_calibration.py
kevinsung: Done.

cirq/experiments/single_qubit_readout_calibration.py
kevinsung: Yes, seconds since the epoch.

cirq/experiments/single_qubit_readout_calibration.py
kevinsung: Done.

cirq/experiments/single_qubit_readout_calibration.py
mpharrigan: oh yeah I guess python uses floats

cirq/experiments/single_qubit_readout_calibration.py
mpharrigan: I think this is also the case for `str`, no?

PR #2535, opened 2019-11-13
Make WaitGate a SingleQubitGate

dstrain115:
- This helps serializers understand how many qubits the gate has
without instantiating an instance of the class.


dstrain115:
> What do you mean "helps serializers understand"? Where is this relevant?

The device specification proto uses number of qubits for validation:
https://github.com/quantumlib/Cirq/blob/master/cirq/google/api/v2/device.proto#L46

The SerializableDevice uses this to validate that the gate is used correctly.  Currently this is done by examining whether it has SingleQubitGate as a feature.  I couldn't think of a better way to do it since we only have a type and not an instance of the type:
https://github.com/quantumlib/Cirq/blob/master/cirq/google/devices/known_devices.py#L133

We could also change this logic to exclude it like we do for measurement gate which would allow you to have any number of qubits for serialization purposes, but would be a special case for this class.

> Are we absolutely sure we'll never want a concept of multi-qubit waiting?

I'm not sure.  However, it seems like changing _num_qubits_ behavior is approximately as backwards incompatible as changing the feature, since they do the same thing.

dstrain115:
We discussed off-line, and I will instead change the logic of known_devices to special case Wait Gate.

I will close this PR and create a new one.

Issue #2310, opened 2019-10-10
Convert documentation to jupyter notebooks. 


dabacon:
These display nicely in github, and we can directly link to opening them in colab and other places where you can run the notebooks.

mpharrigan:
nbsphinx also works well

bt3gl:
Hi, kindly assign this to me, as I am aiming to resolve it in #2542 .

vtomole:
Closed by https://github.com/quantumlib/Cirq/pull/2713

PR #2542, opened 2019-11-13
Convert documentation in docs/ to jupyter notebooks

bt3gl:
Fixes [#2310](https://github.com/quantumlib/Cirq/issues/2310), adding:

- A notebook for each of the [markdown pages inside docs/](https://github.com/quantumlib/Cirq/blob/master/docs/) (checked against `latest`). I also checked everything with Grammarly, fixing grammar typos when necessary.

You can look at each notebook in GH with "View file":

Screen Shot 2019-11-14 at 1 33 45 PM

### Testing Locally

Install [anaconda](https://www.anaconda.com/) and [miniconda](https://docs.conda.io/en/latest/miniconda.html), and create a virtual environment with:

```
conda create -n venv python=3
conda activate venv
conda install --yes --file requirements.txt
conda install -c conda-forge google-api-python-client
conda install -c conda-forge cirq
```

### Questions to be answered by the reviewers

1. Is the structure `docs/notebooks/` the best place to keep these notebooks?
2. Should I include a `README.md` file in the root of this directory, with instructions such as above's "Testing locally"?
3. Should I add a notebook for `development.md` and `install.md` as well?
4. In the original issue, it was mentioned the possibility of using [nbsphinx](https://nbsphinx.readthedocs.io/en/0.4.3/) as well. Kindly advise whether this is still an option.

Please advise.


bt3gl:
I just realized that the document page in `stable` is different from `latest` (https://cirq.readthedocs.io/en/latest/tutorial.html) and that Qudits is in latest.

I am going to check all the notebooks against that page and update whatever is needed in the next hours.

vtomole:
> Should I include a README.md file in the root of this directory, with instructions such as above's "Testing locally"?

Those instructuion are too much. We can tell users to run `jupyter notebook` in their virtual env and open the notebook they want from the UI. What do you think @mpharrigan ?

>Should I add a notebook for development.md and install.md as well?

No, there is not a lot of Python code in those files.

bt3gl:
@Strilanc np. I removed a `{` from one of the notebook's (json source) files. Would you have another test (or breaking) in mind?

bt3gl:
All right, so [one test failed](https://travis-ci.com/quantumlib/Cirq/builds/137087939) for `build protos`. But I am not sure whether this suffices.

Screen Shot 2019-11-18 at 2 15 02 PM


I just reverted the doc. Let me know what's next.

bt3gl:
Hi Folks, this PR became a mess so I was considering to simply close it and open another one. Let me know if this is a good idea, or whether we can squash the commits when merging.  

That's how it became so messy:
1.  Strilanc suggested whether I could break one of the notebooks (to check whether tests would trigger it) so I removed a `{`
2. Travis failed with `build protos` (which seems to be unrelated?). I added the `{` back and Travis was still failing.
3. Being not sure about the problem, I decided to rebase the commit to the state when things were passing. Rebases can be messy in big projects, and I should know better...

If you folks don't mind about the silly commits, we can just keep the way it is (Travis is happy now). Otherwise, let me know and I can create a clean PR.

Also, what should we do regarding tests? What could be a better test for jupyter notebooks? Should this be another ticket?

So many questions...

bt3gl:
@googlebot I consent

Strilanc:
Looks like you did a weird merge that confused CLA bot. Don't worry about it, I can override that.

The test of the breaking curly brace failed for a bunch of reasons that aren't your fault. Please try it again now. Unfortunately, the previous run errored instead of actually failing a test. build-protos should be fixed.

Strilanc:
Don't worry about opening a new PR. If worse comes to worse, you can always just squash your whole PR down into one commit and force push the branch. But then no one would get to see the cool explosions from the car crash.

docs/notebooks/noise.ipynb
```diff
@@ -0,0 +1,172 @@
+
```
bt3gl: @Strilanc This is the notebook that is "broken" by removing a `{` in the beginning. If you try to open view the file, you get:
Screen Shot 2019-11-21 at 9 29 04 AM

Travis passed, so this type of error is not being caught. What should we do?

Strilanc: Open an issue to write tooling that tests the notebooks.

bt3gl:
@googlebot I consent.

bt3gl:
Removed `schedules`, cleaned up (because of CLA problem), and moved to #2713

PR #2546, opened 2019-11-14
Deprecate IdentityOperation obsoleted by IdentityGate

Strilanc:
- Convert identity operation json tests to backwards compatibility mode
- Remove existing usage and type checks
- Add _mul delegation methods to Gate/GateOperation to allow I(a)*OTHER to still return OTHER
- Update instance checks in other classes
- Fix some non-absolute public type annotations in raw_types.py
- Fix cirq.I(q)*1j not returning a pauli string


cirq/ops/gate_operation.py
```diff
@@ -189,6 +189,18 @@ def __pow__(self, exponent: Any) -> 'raw_types.Operation':
             return NotImplemented
         return self.with_gate(new_gate)
 
+    def __mul__(self, other):
```
viathor: Missing type annotations here and on `__rmul__` below.

cirq/ops/identity.py
```diff
@@ -103,6 +99,11 @@ def on_each(self, *targets: Union[raw_types.Qid, Iterable[Any]]
                     format(type(target)))
         return operations
 
+    def __pow__(self, power):
```
viathor: Missing type annotations.

cirq/ops/identity.py
```diff
@@ -147,98 +147,45 @@ def _json_dict_(self):
             **other,
         }
 
+    def _mul_with_qubits(self, qubits: Tuple['cirq.Qid', ...], other):
+        if isinstance(other, raw_types.Operation):
+            return other
+        if isinstance(other, (complex, float, int)):
+            from cirq.ops.pauli_string import PauliString
+            return PauliString(coefficient=other)
```
viathor: This is surprising here. I think most users would expect to get a `GlobalPhaseOperation`.

cirq/ops/gate_operation.py
Strilanc: Added `Any`s.

cirq/ops/identity.py
Strilanc: Added `Any`s.

cirq/ops/identity.py
Strilanc: Acknowledged.

The use case I have in mind is users writing things like `P(a) * -1 * Q(b)` where `P` and `Q` are pauli operations. In that context a pauli string is expected to result for P=I,X,Y,Z. I think it would be reasonable to return a global phase operation, as long as it was capable of pretending to be a pauli string sort of thing...

viathor: Did you find that `GlobalPhaseOperation` fails to interoperate with the `PauliString`? I think it should, so if it doesn't maybe add a TODO and file an issue. If it does, let's return `GlobalPhaseOperation` from here.

PR #2547, opened 2019-11-14
Add Sycamore23 device

dstrain115:
- 23 qubit subset of the Sycamore 54 qubit layout


cirq/google/devices/known_devices.py
```diff
@@ -279,3 +279,28 @@ def _json_dict_(self):
...
+-ABC------
+ABCDE-----
+-CDEFG----
+--EFGHI---
+---GHI----
+----I-----
+"""
+
+SYCAMORE23_PROTO = create_device_proto_from_diagram(
+    _SYCAMORE_GRID,
```
cduck: Pylint didn't catch that `_SYCAMORE23_GRID` wasn't used?
```suggestion
    _SYCAMORE23_GRID,
```

cirq/google/devices/known_devices.py
```diff
@@ -279,3 +279,28 @@ def _json_dict_(self):
...
+    [gate_sets.SQRT_ISWAP_GATESET, gate_sets.SYC_GATESET],
+    _SYCAMORE_DURATIONS_PICOS,
+)
+
+Sycamore23 = SerializableDevice.from_proto(
+    proto=SYCAMORE_PROTO,
```
cduck:
```suggestion
    proto=SYCAMORE23_PROTO,
```

cirq/google/devices/known_devices_test.py
```diff
@@ -385,17 +386,17 @@ def test_json_dict():
...
+@pytest.mark.parametrize('device', [cg.Sycamore, cgdk.Sycamore23])
+def test_sycamore_device(device):
     q0 = cirq.GridQubit(5, 4)
     q1 = cirq.GridQubit(5, 5)
     syc = cirq.FSimGate(theta=np.pi / 2, phi=np.pi / 6)(q0, q1)
     sqrt_iswap = cirq.FSimGate(theta=np.pi / 4, phi=0)(q0, q1)
-    cg.Sycamore.validate_operation(syc)
-    cg.Sycamore.validate_operation(sqrt_iswap)
+    device.validate_operation(syc)
+    device.validate_operation(sqrt_iswap)
```
cduck: I think you need a test for invalid operations to catch the typos mentioned earlier.

dstrain115: Done.  Thanks for the good catch!  That will teach me for hurrying out PRs as I am heading out for the evening!

cduck:
I think your change is correct but I'm not sure about the value of `ParamDictType = Dict[Union[str, sympy.Basic], Union[float, str, sympy.Symbol]]`.  I would expect it to be `Dict[Union[str, sympy.Symbol], Union[float, str, sympy.Basic]]` (map symbols or symbol names to values, symbol names, or arbitrary expressions).

dstrain115:
I think that would be a more precise typing for a ParamDictType.  sympy
substitutions are weird.  It was unexpected that you could resolve (or
substitute) formulas for different formulas in sympy.  I think changing the
type would clarify that this is not the intention in cirq.  It's probably
just someone (maybe me even?) getting confused between sympy.Symbol and
sympy.Basic.

It seems irrelevant to the current PR though.  I'll create a different PR
for it?

cduck:
@dstrain115 I meant to comment on 2550.

cirq/google/devices/known_devices_test.py
```diff
@@ -17,6 +17,7 @@
 import cirq
 import cirq.google as cg
 import cirq.google.common_serializers as cgc
+import cirq.google.devices.known_devices as cgdk
```
mpharrigan: is this how we're supposed to use it?

dstrain115: Good question.  I wasn't sure if we wanted to expose it in the package for cirq.google or not.

Any opinions?

mpharrigan: I personally find the "use the first letter of each submodule" convention very obfuscating and would support any other convention that avoids it :) 

The rest of cirq operates under the "put everything in the top-level namespace" convention

Strilanc: We should expose Sycamore23 from cirq.google, but not from top-level cirq, as it is a google-specific device.

`cirq.google.Sycamore23` same as `cirq.google.Bristlecone`.

dstrain115: Done.

dstrain115:
Ping.  Could I get approval on this in the next few days?  I would like to commit this before my next import to google3, which I would like to do by the end of the week.  Thanks.

PR #2548, opened 2019-11-14
Add developer doc explaining serialization

Strilanc:


docs/dev/serialization.md
```diff
@@ -0,0 +1,110 @@
...
+
+```python
+obj = cirq.read_json(json_text=string)
+```
+
+## Mechanism
+
+When writing JSON, Cirq checks if the given object has a `_json_dict_` method.
+If it does, the object is replaced by the output of that method.
+Otherwise there are a series of several hardcoded cases for complex numbers, numpy arrays, sympy expressions, and a few others.
```
dstrain115: nit: I think it should be "Otherwise, there are..." with a comma.

docs/dev/serialization.md
```diff
@@ -0,0 +1,110 @@
+This developer document explains how Cirq serializes objects into (and out of) JSON.
+It also explains how to add a new serializable object,
+how to remove a serializable object from cirq while maintaing backwards compatibility with old serialized files,
+and various related guidelines.
+
+## Exposed API
```
dstrain115: Optional: should there be a short background doc to explain why we want/need this?

Strilanc:
> I'm confused by the line wrapping in this file. It seems to be by sentence, except when it isn't

It line wraps at the end of sentences and also when the length gets too long. I wrote it in github's editor so I didn't have the 80 column guide.

docs/dev/serialization.md
Strilanc: Added.

docs/dev/serialization.md
Strilanc: Done.

PR #2550, opened 2019-11-14
Fix ParamResolver.param_dict type

kevinsung:
Is this correct? The new type allows keys that are sympy.Basic, not just sympy.Symbol.

### DIFF
--- a/cirq/study/resolver.py
+++ b/cirq/study/resolver.py
@@ -22,7 +22,7 @@ if TYPE_CHECKING:
     import cirq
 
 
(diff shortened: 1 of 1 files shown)


cduck:
I think your change is correct but I'm not sure about the value of `ParamDictType = Dict[Union[str, sympy.Basic], Union[float, str, sympy.Symbol]]`.  I would expect it to be `Dict[Union[str, sympy.Symbol], Union[float, str, sympy.Basic]]` (map symbols or symbol names to values, symbol names, or arbitrary expressions).

kevinsung:
Yes, I think you are correct. I've changed it.

PR #2551, opened 2019-11-14
Store ParamResolver keys as strings

kevinsung:
The reason I want to do this is to make JSON serialization simple; in JSON, dictionary keys need to be strings or something simple like that. Note that I didn't actually need to change the test; it still passed but was incorrect because keys are not supposed to be Sympy formulas.

We could avoid modifying the input dictionary by adjusting `ParamDictType` to only allow string keys, but that wouldn't be backwards compatible.

### DIFF
--- a/cirq/study/resolver.py
+++ b/cirq/study/resolver.py
@@ -93,20 +93,16 @@ class ParamResolver:
         if isinstance(value, float):
             return value
 
-        # Handles 2 cases:
-        # Input is a string and maps to a number in the dictionary
-        # Input is a symbol and maps to a number in the dictionary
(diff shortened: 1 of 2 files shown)


kevinsung:
I'm actually still not sure this is good due to the potential performance hit. Perhaps we just need to store the key types during JSON serialization?

Strilanc:
We are capable of serializing sympy symbols, so there shouldn't be a serialization obstacle.

kevinsung:
>you could turn the dictionary into a list of key,val tuples

I don't know why I didn't think of that!

PR #2558, opened 2019-11-15
Add Gate and Operation Guidelines dev docs

Strilanc:
- Add a `docs/dev/index.rst` to fix dev doc sections being inlined into the left side index of user documentation

Before:

image

After:

image


Strilanc:
> Why the difference?

Because readthedocs is primarily for user documentation. The developer documentation about coding conventions would be distracting. So we hide it behind a single link.

Strilanc:
> We have Development: https://github.com/quantumlib/Cirq/blob/master/docs/development.md and this (part of Internal Documentation for Cirq Developers). I think they should be merged. Thoughts?

Yes, we should probably move the development doc into this folder and rename it to "Getting Started" or something like that.

PR #2559, opened 2019-11-15
Add explicit JSON resolver function in contrib

### LINKED ISSUE #2445: (re-)register deserializer for QuantumVolumeResult

KevinVillela:
Fixes #2445 

This PR adds an explicit function to return a JSON serializer containing Contrib classes. Right now, only QuantumVolumeResult is JSONable, but in the future other classes can be added here as well. Users who want to use the JSON family of functions will need to explicitly pass the contrib_class_resolver, as is done in the test.

Additionally, this pulls out a useful json_test function into the testing directory, for use elsewhere.


cirq/testing/json.py
```diff
@@ -0,0 +1,38 @@
+import io
+import numpy as np
+import pandas as pd
+
+import cirq
+
+
+def assert_roundtrip(obj, text_should_be=None, resolvers=None):
```
Strilanc: This name is too vague; there is not enough context to know what `cirq.testing.assert_roundtrip` refers to.

Rename this to something more explicit like `assert_json_roundtrip_works`.

We may want to include this into the standard test battery for gates (`assert_implements_consistent_protocols`).

cirq/contrib/json.py
```diff
@@ -0,0 +1,15 @@
...
+"""
+
+
+def contrib_class_resolver(cirq_type: str):
+    """Extend cirq's JSON API with resolvers for cirq contrib classes."""
+    from cirq.contrib.quantum_volume import QuantumVolumeResult
+    classes = [
+        QuantumVolumeResult,
+    ]
+    print(QuantumVolumeResult.__name__)
```
Strilanc: Leftover debug line

KevinVillela: D'oh I fixed that and forgot to commit it :P Done.

cirq/testing/json.py
KevinVillela: Fixed the name, thanks!

I tried adding this to assert_implements_consistent_protocols, but that function is called with a ton of test gates (e.g. `GateUsingWorkspaceForApplyUnitary`) :\

KevinVillela:
Friendly ping :)

cirq/contrib/json.py
```diff
@@ -0,0 +1,14 @@
+"""Contains functions for adding JSON serialization and de-serialization for
+classes in Contrib.
+
+"""
+
+
```
mpharrigan: you could add
```python
DEFAULT_RESOLVERS = [contrib_class_resolver] + protocols.json.DEFAULT_RESOLVERS]
```

then

```python
import cirq
import cirq.contrib
data = cirq.read_json(fn, resolvers=cirq.contrib.DEFAULT_RESOLVERS)
```

The above relies on python namespacing for disambiguation. You could make the name more specific: `DEFAULT_CONTRIB_RESOLVERS` if you want

KevinVillela: Good idea, done and used in the test.

mpharrigan:
@Strilanc you good?

PR #2560, opened 2019-11-15
Test gate being added multiple times.

dstrain115:
I wasn't sure what would happen if you added a gate to a gate set multiple times, so I decided to add tests to make sure it worked.  Also, made a fix to a test for with_added_gates() as there was a typo and it didn't cause the tests to fail.

This PR does not change functionality.  It only adds tests.

### DIFF
--- a/cirq/google/devices/known_devices_test.py
+++ b/cirq/google/devices/known_devices_test.py
@@ -458,3 +458,54 @@ valid_targets {
   }
 }
 """
+
+
+def test_adding_gates_multiple_times():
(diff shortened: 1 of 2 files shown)


cirq/google/devices/known_devices_test.py
```diff
@@ -458,3 +458,54 @@ def test_proto_with_waitgate():
...
+}
+valid_targets {
+  name: "2_qubit_targets"
+  target_ordering: SYMMETRIC
+  targets {
+    ids: "0_0"
+    ids: "0_1"
+  }
+}
+"""
```
maffoo: nit: add trailing newline

PR #2561, opened 2019-11-15
Implement JSON serialization for TrialResult

kevinsung:
- Packs bits into hex strings to save a factor of 50 or so in space

This scheme is not compatible with higher-dimensional qudit-valued measurements, but TrialResult currently does not support that anyway.

### DIFF
--- a/cirq/protocols/json_test_data/TrialResult.repr
+++ b/cirq/protocols/json_test_data/TrialResult.repr
@@ -1,0 +1,1 @@
+cirq.TrialResult(params=cirq.ParamResolver({'a': 0.5}), measurements={'m': np.array([[1, 1], [1, 1], [0, 0], [1, 1], [1, 1]], dtype=np.uint8), 'n': np.array([[0, 1, 0], [1, 1, 0], [1, 1, 1], [1, 0, 1], [1, 1, 0]], dtype=np.uint8)})
--- a/cirq/protocols/json.py
+++ b/cirq/protocols/json.py
@@ -112,6 +112,7 @@ class _ResolverCache:
                 cirq.SingleQubitPauliStringGateOperation,
                 'SwapPowGate': cirq.SwapPowGate,
                 'SycamoreGate': cirq.google.SycamoreGate,
+                'TrialResult': cirq.TrialResult,
                 'TwoQubitMatrixGate': cirq.TwoQubitMatrixGate,
                 '_UnconstrainedDevice':
                 cirq.devices.unconstrained_device._UnconstrainedDevice,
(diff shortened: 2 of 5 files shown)


cirq/protocols/json_test_data/TrialResult.repr
```diff
@@ -0,0 +1 @@
+cirq.TrialResult(params=cirq.ParamResolver({'a': 0.5}), measurements={'m': np.array([[1, 1], [1, 1], [0, 0], [1, 1], [1, 1]], dtype=np.uint8), 'n': np.array([[0, 1, 0], [1, 1, 0], [1, 1, 1], [1, 0, 1], [1, 1, 0]], dtype=np.uint8)})
```
Strilanc: Change this into a list, and include a test case for larger measurements using `np.uint16`, `np.uint32`, `np.uint64`, and `np.object` (used for measurements that don't fit in a word).

kevinsung: Do you mean to test measurements that are not {0, 1}-valued? The current code doesn't work for that, though I could probably put something together. If we do that we should also support qudits in the other aspects of TrialResult. For instance, the conversion to Pandas DataFrame assumes bit-valued measurements.

kevinsung: We can use Numpy's .npy format to save the array, which would be completely general. However, in the specific case of {0, 1}-valued measurements, it is 8x less space efficient than here.

Strilanc: > Do you mean to test measurements that are not {0, 1}-valued?

No, that's not what I mean.

Currently, the code that generates the pandas dataframe is careful to pick a minimum sized uint type to fit the bits. For example, if there are 13 measurements, it will pick a uint16. Serialization code should be able to preserve this type information. The most likely way it would break is in using a larger type, or some sort of error when the np array type becomes `object` for measurements with more than 64 bits.

kevinsung: This code doesn't serialize the pandas dataframe. It serializes the dictionary of measurements. Each value in the dictionary of measurements is a numpy array of zeros and ones (assuming qubits). The rows correspond to the repetition and the columns correspond to the qubit being measured. So the dtype is always uint8.

kevinsung:
I'm having second thoughts about implementing this in a way that is incompatible with qudits. But as I described above, the most straightforward way to accommodate qudits is 8x less space-efficient for qubits. @Strilanc do you have thoughts about how to handle this? Should we handle qudits but first check to see if all measurements are binary-valued, in which case we use the more efficient method? Or should we use the general method in all cases for the sake of consistency and simplicity?

kevinsung:
Closing in favor of #2576

PR #2562, opened 2019-11-18
Don't run async methods on other threads by default

Strilanc:
- Was causing bugs due to code not being thread safe by default
- Instead default to running synchronously on await


cirq/work/sampler.py
```diff
@@ -172,9 +170,7 @@ def run_sweep(
         Returns:
             An awaitable TrialResult.
         """
-        results = await self.run_sweep_async(program, study.UnitSweep,
-                                             repetitions)
-        return results[0]
+        return self.run(program, repetitions=repetitions)
```
viathor: The comment

```
By default, this method calls `run` on another thread...
```

above needs to be updated.

(Applies below, too.)

Strilanc: Done.

PR #2568, opened 2019-11-19
Fix several dozen relative type annotations

Strilanc:
- Replace stuff like raw_types.Qid with 'cirq.Qid'

This improves the documentation and IDE autocompletion, by making it refer to the user-visible type instead of internal variants.


mpharrigan:
looks fine; fingers crossed this doesn't cause too many merge conflicts with open prs

PR #2569, opened 2019-11-19
Add CircuitDiagramInfoArgs.format_real

Strilanc:
- Also fix the qasm UGate not having a decomposition or a correct repr

### DIFF
--- a/cirq/circuits/qasm_output.py
+++ b/cirq/circuits/qasm_output.py
@@ -54,24 +54,25 @@ class QasmUGate(ops.SingleQubitGate):
             pre_phase / np.pi,
         )
 
-    def _qasm_(self, qubits: Tuple[ops.Qid, ...], args: 'cirq.QasmArgs') -> str:
+    def _qasm_(self, qubits: Tuple['cirq.Qid', ...], args: 'cirq.QasmArgs') -> str:
         args.validate_version('2.0')
         return args.format(
             'u3({0:half_turns},{1:half_turns},{2:half_turns}) {3};\n',
             self.theta, self.phi, self.lmda, qubits[0])
 
     def __repr__(self) -> str:
-        return 'cirq.QasmUGate({}, {}, {})'.format(self.theta, self.phi,
-                                                   self.lmda)
+        return (f'cirq.circuits.qasm_output.QasmUGate('
+                f'theta={self.theta!r}, '
+                f'phi={self.phi!r}, '
+                f'lmda={self.lmda})')
 
-    def _unitary_(self) -> np.ndarray:
-        # Source: https://arxiv.org/abs/1707.03429 (equation 2)
-        operations = [
-            ops.Rz(self.phi * np.pi),
-            ops.Ry(self.theta * np.pi),
-            ops.Rz(self.lmda * np.pi),
+    def _decompose_(self, qubits):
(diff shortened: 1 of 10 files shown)


cirq/protocols/circuit_diagram_info_protocol.py
```diff
@@ -133,6 +134,15 @@ def __repr__(self):
                                          self.use_unicode_characters,
                                          self.precision, self.qubit_map))
 
+    def format_real(self, val: Union[sympy.Basic, int, float]):
+        if isinstance(val, sympy.Basic):
+            return str(val)
```
viathor: Optional: check that `val` is real.

cirq/protocols/circuit_diagram_info_protocol.py
```diff
@@ -133,6 +134,15 @@ def __repr__(self):
                                          self.use_unicode_characters,
                                          self.precision, self.qubit_map))
 
+    def format_real(self, val: Union[sympy.Basic, int, float]):
```
viathor: `-> str`

cirq/protocols/circuit_diagram_info_protocol.py
```diff
@@ -133,6 +134,15 @@ def __repr__(self):
...
+        if isinstance(val, sympy.Basic):
+            return str(val)
+        if val == int(val):
+            return str(int(val))
+        if self.precision is None:
+            return str(val)
+        return '{{:.{}}}'.format(self.precision).format(float(val))
```
viathor: No need to resort to calling `format`:

```python
return f'{float(val):.{self.precision}f}'
```

should work.

cirq/circuits/qasm_output.py
```diff
@@ -54,24 +54,26 @@ def from_matrix(mat: np.array) -> 'QasmUGate':
...
-        return 'cirq.QasmUGate({}, {}, {})'.format(self.theta, self.phi,
-                                                   self.lmda)
-
-    def _unitary_(self) -> np.ndarray:
-        # Source: https://arxiv.org/abs/1707.03429 (equation 2)
-        operations = [
-            ops.Rz(self.phi * np.pi),
-            ops.Ry(self.theta * np.pi),
-            ops.Rz(self.lmda * np.pi),
+        return (f'cirq.circuits.qasm_output.QasmUGate('
```
viathor: Maybe import `QasmUGate`  into top level namespace instead? The name is unlikely to collide and the repr would look a lot better.

cirq/protocols/circuit_diagram_info_protocol.py
Strilanc: Done.

cirq/protocols/circuit_diagram_info_protocol.py
Strilanc: Acknowledged.

cirq/protocols/circuit_diagram_info_protocol.py
Strilanc: Much nicer. Done.

cirq/circuits/qasm_output.py
Strilanc: Will leave it here for now. Can be done in separate change if needed.

Issue #2142, opened 2019-09-20
On PartialGate and cirq.Z(a).controlled_by(b) returning a CZ


Strilanc:
I've been experimenting with whether it's feasible to have special cases for `controlled_by`. By having specific classes (`GateOperation`, `ControlledGate`, `ZPowGate`) override their `controlled_by` method, and preferentially calling this method instead of `ControlledOperation`'s constructor in a few places, we can do things like have Z.controlled_by(x) turn into a CZ.

The main decide is that, in order for it to really work, it is necessary to introduce a `PartialGate` class that has some of the qubits specified. Otherwise it is not possible for `cirq.Z.controlled_by(q)` to return a CZ that has the control filled in but not the target. This situation arises in the code path within `GateOperation`.

An alternative could be to have `Gate.controlled_by` take an integer for the number of controls, and return a gate without them specified.

I'm a bit worried that this partial gate concept will start appearing in more places as the code base grows. It's particularly tricky at the moment because it's ambiguous whether `gate.num_qubits()` is the number of qubits *to pass in order to get an operation* (relevant to `on(...)`) or the number of qubits *that will be present in the final operation* (relevant to `_unitary_`).

cduck:
I don’t see a need for PartialGate here.  It seems reasonable for Z.on(a).controlled_by(b) to return CZ(b, a) and Z.control(1) to return CZ (once #2090 is merged).

Strilanc:
The `controlled_by` method of `Gate` currently doesn't take qubits to apply to; `ZPowGate.controlled_by` has to return a `Gate` that takes a single qubit. Even if you say `Z.on(a).controlled_by(b)`, the `ZPowGate.controlled_by` method is being invoked behind the scenes when `GateOperation` delegates to it.

mpharrigan:
I think I made some sort of argument like this in #1973 , but if you do decide that `PartialGate` is needed, you might as well go all the way an collapse `Gate`, `PartialGate` and `Operation` into one thing, corresponding to none, some, and all of the qubits filled out.

I still question the utility of having partial gates. In your example, they're very quickly promoted to full operations by chaining methods. Can you give an example where you'd want to have `Z.controlled_by(q)` be assigned to a variable and used later?

Strilanc:
Yes, this does relate pretty closely to #1973 .

> Can you give an example where you'd want to have `Z.controlled_by(q)` be assigned to a variable and used later?

[...]

PR #2572, opened 2019-11-19
Update ZPow.controlled specialization

smitsanghavi:
Respect Z.controlled(a).controlled(b) === Z.controlled(b+a)

### DIFF
--- a/cirq/ops/common_gates.py
+++ b/cirq/ops/common_gates.py
@@ -33,7 +33,7 @@ import cirq
 from cirq import protocols, value
 from cirq._compat import proper_repr
 from cirq._doc import document
-from cirq.ops import gate_features, eigen_gate, raw_types
+from cirq.ops import controlled_gate, gate_features, eigen_gate, raw_types
 
 from cirq.type_workarounds import NotImplementedType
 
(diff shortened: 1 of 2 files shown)


smitsanghavi:
#2142

cirq/ops/common_gates.py
```diff
@@ -387,15 +387,20 @@ def controlled(self,
...
+        Specialize controlled for ZPow to return corresponding controlled CZPow
+        when the last control (which acts first semantically) is a default-type
+        control qubit.
         """
         result = super().controlled(num_controls, control_values,
                                     control_qid_shape)
-        if (result.control_values == ((1,),) and  # type: ignore
-                result.control_qid_shape == (2,)):  # type: ignore
+        if (isinstance(result, controlled_gate.ControlledGate) and
+                result.control_values[-1] == (1,) and
```
Strilanc: Does this blow up when calling `controlled()` with an empty list of qubits?

smitsanghavi: No, that's why I still kept the super().controlled above. It populates sane defaults for all unspecified args.
Added additional tests to cover this case.

Issue #2506, opened 2019-11-06
Include "global" tests in pytest-changed-files


Strilanc:
A "global" test is a fast test that should fire if any .py file is modified. I think there are only two at the moment: `cirq/protocols/json_test.py` and (soon to be) `docs/docs_coverage_test.py`.

PR #2575, opened 2019-11-20
Make pytest-changed-files include api tests when an __init__ file is changed

Strilanc:
Fixes https://github.com/quantumlib/Cirq/issues/2506

### DIFF
--- a/check/pytest-changed-files
+++ b/check/pytest-changed-files
@@ -61,11 +61,16 @@ changed=$(git diff --name-only ${rev} -- \
     | sort \
     | uniq \
 )
-num_changed=$(echo -e "${changed}" | wc -w)
+if git diff --name-only "${rev}" -- | grep "__init__\.py$" > /dev/null; then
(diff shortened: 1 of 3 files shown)


check/pytest-changed-files
```diff
@@ -61,11 +61,16 @@ changed=$(git diff --name-only ${rev} -- \
...
 
 # Run it.
-echo "Found ${num_changed} differing files with associated tests." >&2
-if [ -z "${changed}" ]; then
+echo "Found ${num_changed} test files associated with changes." >&2
+if [ "${num_changed}" -eq 0 ]; then
     exit 0
 fi
-pytest ${rest} ${changed}
+pytest ${rest} ${changed[@]}
```
viathor: I don't think `[@]` has any effect unless you place the variable in double quotes: `"${changed[@]}"`.

Strilanc: reverted

Strilanc: Restored. Broke tests.

Issue #2508, opened 2019-11-06
Naming rotations


viathor:
This issue seeks to settle the naming convention for factory functions that produce gate instances given rotation angles. I found the following factories:

```
Rx, Ry, Rz, GivensRotation, ISwapRotation, MS
```

These are clearly inconsistent. Proposals:

1. `Rx, Ry, Rz, Rgivens, Riswap, Rms`
2. `rx, ry, rz, rgivens, riswap, rms`
3. `XRotation, YRotation, ZRotation, GivensRotation, ISwapRotation, MSRotation`
4. `x_rotation, y_rotation, z_rotation, givens_rotation, iswap_rotation, ms_rotation`

I'm uneasy about conventions that lead to names like `cirq.rms` since this may be interpreted as "root mean square".

To broaden the set of examples, here are some names we may reasonably anticipate in future: 

```
Rzz, pauli_string_rotation
```

Context: [comment](https://github.com/quantumlib/Cirq/commit/5c13dea6ead7dc3721c1a84c523dd1e5d805bef0#commitcomment-35820146)

Thoughts?

dabacon:
related #1681

dabacon:
I think method names should follow style conventions, so `snake_case`.  When I see `cirq.Rx` I think it is a class and start looking around for the class definition.  Maybe,
5. `rx, ry, rz, givens, iswap, ms`.
I think the factory methods main benefit is being short-cuts.

viathor:
I agree about `snake_case` (hence conventions 2 and 4).

Convention 5 is inconsistent with its use of the `r` prefix.

dabacon:
I think calling a unitary gate a rotation is kind of redundant.  "Rotation about givens" or "Rotation about Molmer-Sorenson" seems odd to me.

viathor:
Good point. I'm starting to like 5. One more concern with it though: `cirq.iswap` and `cirq.ISWAP` meaning slightly different things is potentially confusing and error-prone.

viathor:
Thought about it and I think I'm willing to accept `cirq.iswap` and `cirq.ISWAP` given that they would actually enter python expressions quite differently: one as a function call like `cirq.iswap(np.pi/5)` and the other standalone or exponentiated, e.g. `cirq.ISWAP**0.5`. This use is also consistent with the naming convention which implies the former is a function and the latter is a constant.

Thoughts?

Strilanc:
[...]

PR #2580, opened 2019-11-20
Rename Rx, Ry and Rz factories

### LINKED ISSUE #1681: Follow method style guide for gates
We currently have `cirq.Rx` which is a method, not a gate.  Method names should be snake case.  This is confusing since it makes it seem that `cirq.Rx` is a class name.  Especially considering that users expect there to be a `cirq.Rx` gate and not `cirq.XPowGate`

viathor:
Fixes #1681 and #2508.


viathor:
This bothered me initially too but after trying it out I decided against it for a few reasons:
 * "rx", "ry" and "rz" look different from the gates in formulas and diagrams in textbooks/papers/etc,
 * they are less readable,
 * they would give these gates a look that is inconsistent with other gates which are rendered with at least some capital letters.

Also, note that gate symbols for diagrams are generally not the same as gate names or their reprs, e.g. PhX for PhasedXPowGate.

docs/api.rst
```diff
@@ -52,6 +52,9 @@ Unitary effects that can be applied to one or more qubits.
     cirq.XX
     cirq.YY
     cirq.ZZ
+    cirq.rx
+    cirq.ry
+    cirq.rz
```
Strilanc: You forgot to move `cirq.Rx` and friends into the deprecated section

viathor: Thanks for catching! #2585.

PR #2584, opened 2019-11-21
Fix pytest

dstrain115:
- check/pytest-changed-files is broken
- The list of fils to test is put in quotes so it is considered to
be one file by pytest rather than a list of files.


vtomole:
Hey @dstrain115 how does this change affect us with respect to https://github.com/koalaman/shellcheck/wiki/SC2068?

dstrain115:
I am not sure if this is the correct way to fix the problem, but it seems that all tests were broken after this PR: 
https://github.com/quantumlib/Cirq/pull/2575

Let's call this the quick hot-fix, and I'll talk to Craig when he gets in to see if there's a better way

Issue #2587, opened 2019-11-21
Add DiagonalTwoQubitGate


vtomole:
```python
def CZPowGate00(g):
      return np.array([[g, 0, 0, 0],
         [0, 1, 0, 0],
         [0, 0, 1, 0],
         [0, 0, 0, 1]])
``` 
e.t.c

These pyquil built-in gates: https://github.com/rigetti/pyquil/blob/2f7e4b4dd5edcb99da1f5e7d8f4459ba11e81af3/pyquil/gates.py#L298 are needed to be compatible with QUIL. See:  https://github.com/quantumlib/Cirq/issues/2386

balopat:
Resolved in #3211

PR #2588, opened 2019-11-21
Implement CZPowGate00 and CZ00

vtomole:
This is the first part of https://github.com/quantumlib/Cirq/issues/2587


vtomole:
@Strilanc I agree.

Issue #2478, opened 2019-11-01
Delete cirq.Schedule


Strilanc:
I was discussing this idea with @dstrain and @maffoo and @MichaelBroughton and got a surprisingly positive response. So I want to bring it up as a thing we should consider doing.

Originally, we added Schedule as a way to have a finer degree of control over the hardware. But over time the following has happened:

1. Even when calling actual hardware, users have stuck with `Circuit`. We haven't had anyone complain about lack of flexibility with `Circuit` as far as I know or can remember. It seems like `Circuit` is sufficient.

2. `Schedule` has received very little work over time, and so it is in janky bit-rotted state. For example, it didn't even have a `_unitary_` method until very recently. We would have to invest significant effort to make `Schedule` good.

3. Using `Schedule` is very easy to get wrong in the context of real hardware. There are often soft constraints like "after X you can immediately do Y, but you should wait T nanoseconds before doing Z". Coming up with a format to represent all of that and achieve optimal performance is a hard ever-deepening task.

Overall, it seems like we are being pulled towards just Circuit as the right level of abstraction.

I'm particularly interested in opinions from @dabacon and @mpharrigan .

dstrain115:
For the record, I vote yes in dropping this functionality.  The schedule is too detailed of a level of abstraction and is nearly impossible to schedule correctly or efficiently.  I do not think we should support this level of control.  

Having the WaitGate should mitigate many use cases that would have previously needed a schedule.

dabacon:
So one thing people haven't done, which I sort of expected, is to pack things a bit tighter than the Circuit allows (if you have gates with differing times).  WaitGate doesn't cover this case.

But this reminds me of an idea that could mitigate a bunch of these ideas: what if we added to operations instructions about when to schedule them, i.e. if you have "open time" before your gate, you can slide into this window.  Some sort of fine gained control like this might satisfy everything we can do with schedules and would cover, with WaitGate, a lot of cases.  If we were able to cover these cases with something like this, I'd say getting rid of scheduled operations sounds good.

c-poole:
I think it will be simpler to modify `Circuit` to have any currently missing functionality than to attempt to repair the neglect we've had for `Schedule` up until now. I vote in favor of dropping it.

Strilanc:
> what if we added to operations instructions about when to schedule them, i.e. if you have "open time" before your gate

Are you referring to relative placement within a moment, or are you referring to tight overlapped packing between different moments?

[...]

PR #2589, opened 2019-11-22
Delete schedules

Strilanc:
- Delete cirq/schedules directory
- Remove Device.duration_of/validate_schedule/validate_scheduled_operation
- Remove all validate_schedule methods (but leave duration_of on specific devices for now)
- Remove all references to cirq.Schedule
- Maintain engine backwards compatibility by auto-converting to circuits in schedule contexts

Skipping deprecation because there's no reasonable way to half support schedules, and
this would tie us up for months bending over backwards for a feature we think
nobody uses.

Fixes https://github.com/quantumlib/Cirq/issues/2478


cirq/aqt/aqt_sampler.py
```diff
@@ -187,8 +187,7 @@ def run_sweep(self,
             resolver.
         """
         meas_name = 'm'  # TODO: Get measurement name from circuit. Issue #2195
-        circuit = (program.to_circuit()
-                   if isinstance(program, schedules.Schedule) else program)
+        circuit = program
```
dabacon: guess we don't want to change interface.  Could just use program throughout instead of circuit?

cirq/google/engine/engine.py
```diff
@@ -65,8 +65,8 @@ def __init__(self, message):
         super().__init__(message)
 
 
-# Quantum programs to run can be specified as circuits or schedules.
-TProgram = Union[circuits.Circuit, schedules.Schedule]
+# Quantum programs to run can be specified as circuits.
+TProgram = 'cirq.Circuit'
```
dabacon: think we could just use cirq.Circuit for these types?

dabacon:
Long live schedules.

cirq/google/engine/engine.py
Strilanc: Done.

cirq/aqt/aqt_sampler.py
Strilanc: Done.

PR #2594, opened 2019-11-22
Delete things marked as deprecated-until-v0.7

Strilanc:
- Since we're now v0.7 dev, we can delete these things


mpharrigan:
Can we extend `op_gate_of_type` for another cycle?

Strilanc:
> Can we extend op_gate_of_type for another cycle?

Done.

PR #2597, opened 2019-11-24
Add more specialized controlled implementations

smitsanghavi:
#2142


smitsanghavi:
Ping!

cirq/ops/common_gates.py
```diff
@@ -115,6 +115,29 @@ def _trace_distance_bound_(self) -> Optional[float]:
...
         return abs(np.sin(self._exponent * 0.5 * np.pi))
 
+    def controlled(self,
+                   num_controls: int = None,
+                   control_values: Optional[Sequence[
+                       Union[int, Collection[int]]]] = None,
+                   control_qid_shape: Optional[Tuple[int, ...]] = None
+                  ) -> raw_types.Gate:
+        """
+        Specialize controlled for XPow to return corresponding controlled
```
dstrain115: Specialized controlled for XPowGate

cirq/ops/common_gates.py
```diff
@@ -115,6 +115,29 @@ def _trace_distance_bound_(self) -> Optional[float]:
...
+        CNotPow when the last control (which acts first semantically) is a
+        default-type control qubit.
```
dstrain115: I would actually revise this whole pydoc to make it more clear.  It's fairly confusing.

cirq/ops/common_gates.py
```diff
@@ -689,6 +712,29 @@ def _pauli_expansion_(self) -> value.LinearDict[str]:
...
+    def controlled(self,
+                   num_controls: int = None,
+                   control_values: Optional[Sequence[
+                       Union[int, Collection[int]]]] = None,
+                   control_qid_shape: Optional[Tuple[int, ...]] = None
+                  ) -> raw_types.Gate:
+        """
+        Specialize controlled for CZPow to return corresponding controlled
+        CCZPow when the last control (which acts first semantically) is a
+        default-type control qubit.
```
dstrain115: I think I would also revise this pydoc.  Again, I think that we should use the actual class names (CZPowGate, etc) instead of abbreviations in our docs.

Same for the other pydocs below.

cirq/ops/common_gates.py
```diff
@@ -820,6 +866,29 @@ def _pauli_expansion_(self) -> value.LinearDict[str]:
...
 
+    def controlled(self,
+                   num_controls: int = None,
+                   control_values: Optional[Sequence[
+                       Union[int, Collection[int]]]] = None,
+                   control_qid_shape: Optional[Tuple[int, ...]] = None
+                  ) -> raw_types.Gate:
+        """
+        Specialize controlled for CNotPow to return corresponding controlled
+        CCXPow when the last control (which acts first semantically) is a
```
dstrain115: I think it might be worth elaborating why a controlled-CNOT is a CCX gate here.  I'm not sure that's immediately obvious to everyone.  I had to think about it for a minute anyway.  The other examples that you added are more straightforward (though it might not hurt to add extra explanation), but this one seems less obvious.

cirq/ops/common_gates_test.py
```diff
@@ -123,57 +123,62 @@ def test_z_init():
...
-    assert cirq.ZPowGate(exponent=5).controlled() == cirq.CZPowGate(exponent=5)
+@pytest.mark.parametrize(
+    'input_gate, specialized_output',
+    [(cirq.Z, cirq.CZ), (cirq.CZ, cirq.CCZ), (cirq.X, cirq.CX),
+     (cirq.CX, cirq.CCX),
+     (cirq.ZPowGate(exponent=0.5), cirq.CZPowGate(exponent=0.5)),
+     (cirq.CZPowGate(exponent=0.5), cirq.CCZPowGate(exponent=0.5)),
+     (cirq.XPowGate(exponent=0.5), cirq.CNotPowGate(exponent=0.5)),
+     (cirq.CNotPowGate(exponent=0.5), cirq.CCXPowGate(exponent=0.5))])
+def test_specialized_control(input_gate, specialized_output):
```
dstrain115: Is it worth checking that the unitaries are the same just to double check that the specialized gate actually is the same as the controlled input gate?

cirq/ops/common_gates.py

cirq/ops/common_gates.py

cirq/ops/common_gates.py

cirq/ops/common_gates.py
smitsanghavi: My understanding is that by definition CCX means controlled-controlled-X and CNOT means controlled-X so I am not sure how to elaborate on this in the pydoc.

cirq/ops/common_gates_test.py
smitsanghavi: Done. Thanks for the suggestion!

cirq/ops/common_gates.py
dstrain115: Specialize to Specialized

cirq/ops/common_gates.py
```diff
@@ -116,6 +116,30 @@ def _trace_distance_bound_(self) -> Optional[float]:
...
+        first semantically) is a default-type control qubit.
+        """
+        result = super().controlled(num_controls, control_values,
+                                    control_qid_shape)
+        if (isinstance(result, controlled_gate.ControlledGate) and
+                result.control_values[-1] == (1,) and
+                result.control_qid_shape[-1] == 2):
+            return cirq.CNotPowGate(exponent=self._exponent,
+                                    global_shift=self._global_shift).controlled(
+                                        result.num_controls() - 1,
```
dstrain115: What happens if num_controls() == 1?  Will this still work correctly?
Is this handled by the super() method?

cirq/ops/common_gates.py
```diff
@@ -116,6 +116,30 @@ def _trace_distance_bound_(self) -> Optional[float]:
...
 
+    def controlled(self,
+                   num_controls: int = None,
+                   control_values: Optional[Sequence[
+                       Union[int, Collection[int]]]] = None,
+                   control_qid_shape: Optional[Tuple[int, ...]] = None
+                  ) -> raw_types.Gate:
+        """
+        Specialize controlled method for XPowGate to construct a corresponding
+        controlled CNotPowGate, if applicable.
```
dstrain115: I still think this pydoc could use some improvement.  Typically pydoc is a one-line summary with a more in-depth paragraph form to follow.

Here is an example of what I am thinking:

```
Constructs CNotPowGate from controlled XPowGate when applicable.

This method is a specialized controlled method for XPowGate.  It overrides
the default behavior of returning a ControlledGate by transforming the
underlying controlled gate to a CNotGatePow and removing the last
specified control qubit (which acts first semantically).  If this is a gate with
multiple control qubits, it will now be a ControlledGate with one less control.

This behavior only occurs when the last control qubit is a default-type control qubit.

(Note that a CNotPowGate is, by definition, a controlled-X gate.)
```

cirq/ops/common_gates.py
```diff
@@ -116,6 +116,30 @@ def _trace_distance_bound_(self) -> Optional[float]:
...
+        Specialization only happens when the last specified control (which acts
+        first semantically) is a default-type control qubit.
```
dstrain115: Actually, now that I am writing an example pydoc, what does "default-type" control qubit mean?  Is this just a qubit and not a qudit?

Strilanc: It means a control that is satisfied by the qubit being ON, as opposed to OFF. Also the qubit vs qutrit thing.

cirq/ops/common_gates.py
Strilanc: Yeah, the purpose of calling the super method is to delegate all the special case handling to it.

cirq/ops/common_gates.py

cirq/ops/common_gates.py
smitsanghavi: Now addressed in the pydoc

cirq/ops/common_gates.py
smitsanghavi: This definitely looks much better. Using this with some changes like adding the explanation for default-qubit below.

cirq/ops/common_gates.py
```diff
@@ -115,6 +115,41 @@ def _trace_distance_bound_(self) -> Optional[float]:
...
+                   control_values: Optional[Sequence[
+                       Union[int, Collection[int]]]] = None,
+                   control_qid_shape: Optional[Tuple[int, ...]] = None
+                  ) -> raw_types.Gate:
+        """
+        Constructs CNotPowGate from controlled XPowGate when applicable.
+
+        This method is a specialized controlled method for XPowGate. It
+        overrides the default behavior of returning a ControlledGate by
+        transforming the underlying controlled gate to a CNotGatePow and
```
dstrain115: CnotPowGate

cirq/ops/common_gates.py
```diff
@@ -115,6 +115,41 @@ def _trace_distance_bound_(self) -> Optional[float]:
...
+        removing the last specified control qubit (which acts first
+        semantically).  If this is a gate with multiple control qubits, it will
+        now be a ControlledGate with one less control.
+
+        This behavior only occurs when the last control qubit is a default-type
+        control qubit. A default-type control qubit is one with shape of 2 (not
+        a generic qudit) and where the control is satisfied by the qubit being
+        ON, as opposed to OFF.
+
+        (Note that a CNotPowGate is, by definition, a controlled-XPow gate.)
```
dstrain115: Optional:  Would controlled-XPowGate be better?  I'll leave it up to you.  Same comments below.

cirq/ops/common_gates.py
```diff
@@ -387,9 +422,21 @@ def controlled(self,
...
                   ) -> raw_types.Gate:
         """
-        Specialize controlled for ZPow to return corresponding controlled CZPow
-        when the last control (which acts first semantically) is a default-type
-        control qubit.
+        Constructs CZPowGate from controlled ZPowGate when applicable.
+
+        This method is a specialized controlled method for ZPowGate. It
+        overrides the default behavior of returning a ControlledGate by
+        transforming the underlying controlled gate to a CZGatePow and
```
dstrain115: CZPowGate

cirq/ops/common_gates.py
```diff
@@ -699,6 +746,41 @@ def _pauli_expansion_(self) -> value.LinearDict[str]:
...
+                   control_values: Optional[Sequence[
+                       Union[int, Collection[int]]]] = None,
+                   control_qid_shape: Optional[Tuple[int, ...]] = None
+                  ) -> raw_types.Gate:
+        """
+        Constructs CCZPowGate from controlled CZPowGate when applicable.
+
+        This method is a specialized controlled method for CZPowGate. It
+        overrides the default behavior of returning a ControlledGate by
+        transforming the underlying controlled gate to a CCZGatePow and
```
dstrain115: CCZPowGate

cirq/ops/common_gates.py
```diff
@@ -830,6 +912,41 @@ def _pauli_expansion_(self) -> value.LinearDict[str]:
...
+                   control_values: Optional[Sequence[
+                       Union[int, Collection[int]]]] = None,
+                   control_qid_shape: Optional[Tuple[int, ...]] = None
+                  ) -> raw_types.Gate:
+        """
+        Constructs CCXPowGate from controlled CNotPowGate when applicable.
+
+        This method is a specialized controlled method for CNotPowGate. It
+        overrides the default behavior of returning a ControlledGate by
+        transforming the underlying controlled gate to a CCXGatePow and
```
dstrain115: CCXPowGate

cirq/ops/common_gates.py
smitsanghavi: Looks better. Made the changes.

PR #2598, opened 2019-11-25
Add (de)serializer for PhasedXZGate

maffoo:


cirq/google/common_serializers_test.py
```diff
@@ -231,6 +231,63 @@ def test_half_pi_does_not_serialize_arbitrary_xy():
...
+        'qubits': [{
+            'id': '1_2'
+        }]
+    }
+    assert gateset.serialize_op_dict(op) == expected
+    deserialized_op = gateset.deserialize_op_dict(expected)
+    cirq.testing.assert_allclose_up_to_global_phase(
+        cirq.unitary(deserialized_op),
+        cirq.unitary(op),
+        atol=1e-7,
```
viathor: All the input values you've chosen can be represented as floats exactly, so I wonder whether we could test this for exact equality?

I'm concerned that we may be missing an issue where a float is serialized as string with insufficient number of characters.

cirq/google/common_serializers.py
```diff
@@ -127,6 +127,54 @@ def _near_mod_2(e, t, atol=1e-8):
...
+        ),
+        op_deserializer.DeserializingArg(
+            serialized_name='axis_phase_exponent',
+            constructor_arg_name='axis_phase_exponent',
+        ),
+    ],
+)
+
+
+
```
viathor: nit: let's make it two empty lines for consistency

cirq/google/common_serializers.py
```diff
@@ -127,6 +127,54 @@ def _near_mod_2(e, t, atol=1e-8):
 ]
 
 
+#
+# PhasedXZ gate (de)serializer
+# TODO(maffoo): Move to SINGLE_QUBIT_(DE)SERIALIZERS when server support is in.
```
viathor: Better to file an issue and reference it in place of your username.

cirq/google/common_serializers.py
maffoo: Done.

cirq/google/common_serializers_test.py
maffoo: Many tests in this module have the same issue, in that we are comparing serialized representations but we have to pick certain values to get exact equality of floats. I agree this is something we should solve, but perhaps we can do that in a separate PR for the entire module, rather than a one-off for this new case. WDYT?

cirq/google/common_serializers.py
maffoo: Done: #2602.

cirq/google/common_serializers_test.py
viathor: Yeah, this is definitely out of the scope of this PR.

FTR from our offline discussion: serialization and deserialization of values such as 0.5 or 0.125 that are represented exactly in both single and double precision floats should yield result exactly equal to the original value. We should not be throwing our hands up just because floats ;-)

Issue #2423, opened 2019-10-28
Improve Clifford circuit simulation support


Strilanc:
- **Make the state more inspectable**. StabilizerStateChForm is a canonical circuit form, so you should be able to output the canonical circuit. `CliffordTableau` should have methods for returning the stabilizers and destabilizers as `cirq.DensePauliString` instances.

    - Add `to_circuit` method to `StabilizerStateChForm`
    - Add `stabilizers` method to `CliffordTableau`.
    - Add `destabilizers` method to `CliffordTableau`.

- **Simplify the internal state**. CliffordSimulator has both a tableau state and a ch state internally. We should only use one of them. My guess is it should be the CH form because that generalizes better to non-Clifford operations (which we could include support for later; not in this issue). Inside the simulator we only need one.

    - Remove `CliffordTableau` type attribute from `CliffordSimulator`

- **Use within the rest of the cirq.** Currently, cirq.sample does some basic analysis of the circuit in order to determine if it can be run by the unitary simulator or requires density matrix stuff. We should extend that analysis to include the Clifford simulator, so if all operations are Clifford, use the more efficient simulator. Methods that should be supported include `cirq.sample`, `cirq.final_wavefunction`, and possibly some others (e.g. we want to add `final_density_matrix`, and it could be useful there as well).

    - Add support for using clifford simulation when possible in `cirq.sample`
    - Add support for using clifford simulation when possible in `cirq.final_wavefunction`

- **Support more gates, including gates that may not be named now**. Need to add the concept of a gate or operation being "Clifford" to cirq more globally. For example, the iswap is clifford but it isn't listed in the explicit list of supported gates inside of CliffordSimulator currently. We need a method to check if a gate is clifford, and alternatively to decompose into Cliffords. This implies we need to add `_is_clifford_` and `_decompose_into_cliffords_` protocols, and perhaps also a `_clifford_tableau_` method akin to `_unitary_` but for the clifford stabilizer table matrix.

    - Add an `_is_clifford_` protocol
    - Add a `_decompose_into_cliffords_` protocol (note: some of these methods are already added in https://github.com/quantumlib/Cirq/pull/2351/)
    - Add a `_clifford_tableau_` protocol (akin to `_unitary_` but for clifford operations)

vtomole:
> Simplify the internal state
The current implementation relies on the tableaux for performing measurements.

vtomole:
[...]

PR #2600, opened 2019-11-26
Add CliffordSimulator to cirq.Sample

smitsanghavi:
1. cirq.sample calls CliffordSimulator when appropriate. 
2. All simulators called by cirq.sample now support 0 repetitions even when measurements are not terminal
3. ClifforSimulator exposes supported gates and supports Param resolution


smitsanghavi:
#2423

cirq/sim/clifford/clifford_simulator_test.py
```diff
@@ -14,6 +15,14 @@ def test_simulate_no_circuit():
     assert len(result.measurements) == 0
 
 
+def test_run_no_repititions():
```
vtomole: Nit: typo for `repetitions`

cirq/sim/mux.py
```diff
@@ -23,11 +23,19 @@
...
 from cirq import circuits, protocols, study, devices, ops, value
 from cirq.sim import sparse_simulator, density_matrix_simulator
+from cirq.sim.clifford import clifford_simulator
 
 if TYPE_CHECKING:
     import cirq
 
 
+def _is_clifford_circuit(program: 'cirq.Circuit') -> bool:
+    return all(
```
Strilanc: Don't treat measurement differently from the other operations.

cirq/sim/mux.py
```diff
@@ -23,11 +23,19 @@
...
+        op.gate in clifford_simulator.CliffordSimulator.get_supported_gates()
```
Strilanc: Don't recompute the supported gates list for every single operation.

Strilanc: This check will fail on gates that decompose into Clifford operations, such as cirq.ISWAP.

I think what we need to do here is to define an is clifford protocol as well as formalize the `decompose_into_clifford` protocol. This line would then reduce to `has_clifford(circuit)`.

cirq/sim/mux.py
smitsanghavi: Done

cirq/sim/mux.py
smitsanghavi: Done.

smitsanghavi: Added a TODO for now

cirq/sim/mux.py
smitsanghavi: Had to treat them differently since I can't directly add them to the List of gates. Should I also limit to terminal measurements here for now?

smitsanghavi:
I agree with the need to implement all the changes you suggested to make this fully work as intended in #2423 . This is just the initial version which aims to get this to start working for the simplest eligible circuits. Added a TODO to document that this approach is not permanent.

PR #2605, opened 2019-11-27
Add optional qubit_set property to cirq.Device

Strilanc:
- Make 'validate_operation' optional to implement
- qubit_set falls back to looking for qubit/_qubit fields/methods, otherwise returns None

This makes devices easier to consume (because a common task is to want to view the qubits, knowing that the device is not the unconstrained device) and easier to make (because when prototyping you don't care about validating operations yet).

The shortest interesting device class now looks like this:

```
class CustomDevice:
    def qubits(self):
        return cirq.LineQubit.range(6)
```

or this:

```
class CustomDevice:
    def __init__(self, qubits):
        self.qubits = qubits
```


cirq/devices/device.py
```diff
@@ -12,20 +12,44 @@
...
-
 
 class Device(metaclass=abc.ABCMeta):
     """Hardware constraints for validating circuits."""
 
+    def qubit_set(self) -> Optional[AbstractSet['cirq.Qid']]:
+        """Returns a set or frozenset of qubits on the device, if possible.
+
+        This method returns None when the set of qubits is unavailable. For
+        example, `cirq.UnconstrainedDevice` returns `None` because allows any
```
dstrain115: This is redundant with the Returns statement.  I would probably keep one or the other rather than repeating it.

cirq/devices/device.py
```diff
@@ -12,20 +12,44 @@
...
+            If the device has a finite set of qubits, then a set or frozen set
+            of all qubits on the device is returned.
+
+            If the device has no well defined finite set of qubits (e.g.
+            `cirq.UnconstrainedDevice` has this property), then `None` is
+            returned.
+        """
+
+        # Compatibility hack to work with devices that were written before this
+        # method was defined.
```
maffoo: As @dstrain115 mentioned, it'd be nice to have a plan for where we are going with `Device` so that we can converge on an interface in the future and get rid of this hack. Maybe create an issue to track this and reference the issue number here.

cirq/devices/device_test.py
```diff
@@ -0,0 +1,41 @@
...
+    class QubitMethodDevice(cirq.Device):
+
+        def qubits(self):
+            return cirq.LineQubit.range(5)
+
+    assert QubitMethodDevice().qubit_set() == frozenset(cirq.LineQubit.range(5))
+
+    class PrivateQubitMethodDevice(cirq.Device):
+
+        def qubits(self):
```
maffoo: Should be `_qubits`

mpharrigan:
echoing others' comments about long-term-plan. "Qubits" are inexorably linked to gate sets.

Is `set` the right data structure here?

 1. introduces asymmetry between the constructor and the property. You give a list but duplicates and order is "erased"
 2. Confusion around the order of the returned thing and/or you can't index in

I might want to write:

```python
qubits = Sycamore.qubits
cirq.Circuit(cirq.CNOT(qubits[0], qubits[1]))
```

but that's not possible, so I might write

```python
qubits = list(Sycamore.qubits)
....
```

and then wonder why my program is different after python changes set ordering / we change how qubits are fed into the device constructor / who knows what

Strilanc:
> and then wonder why my program is different after python changes set ordering / we change how qubits are fed into the device constructor / who knows what

Sets preserve order in python 3, so you shouldn't see the order change like that.

cirq/devices/device.py
Strilanc: Done.

cirq/devices/device.py
Strilanc: Acknowledged.

cirq/devices/device_test.py
Strilanc: Done.

Strilanc:
I have discovered that sets do not preserve order in python 3, only dictionaries.

mpharrigan:
> I have discovered that sets do not preserve order in python 3, only dictionaries.

See my concerns above. I think it will be more helpful to have ordered qubits in a device

Issue #2606, opened 2019-11-30
Remove/reword "We will break your code when we make new releases"


vtomole:
Since we deprecate features now, this sentence is not accurate.

PR #2611, opened 2019-12-02
Update Alpha Disclaimer

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/2606


vtomole:
Ah that's better. PTAL @Strilanc

PR #2615, opened 2019-12-03
Add optional tabulation to ConvertToSycamoreGates

KevinVillela:
- Moves @dkafri's tabulation code from common to google/optimizers (not sure if that's OK or not - alternatively, I can make another optimizer that is called before ConvertToSycamoreGates).
- Add SwapPermutationGate to contrib JSON resolvers
- Quick fix to Quantum Volume:
  - After routing, save the qubit mapping and remove the SwapPermutationGates. This makes it so that optimizers don't need to know about or preserve SwapPermutationGates, which is a routing implementation detail.

@mpharrigan This should give us much better Quantum Volume numbers.


cirq/google/optimizers/convert_to_sycamore_gates.py
```diff
@@ -176,24 +187,44 @@ def known_two_q_operations_to_sycamore_operations(self, qubit_a: ops.Qid,
...
-                op.qubits[0], op.qubits[1], op, allow_partial_czs=True)
-            gate_ops = []
-            for new_op in new_ops:
-                num_qubits = len(new_op.qubits)
-                if num_qubits == 1:
-                    gate_ops.extend([
-                        term.on(new_op.qubits[0])
-                        for term in optimizers.single_qubit_matrix_to_gates(
-                            protocols.unitary(new_op))
+            if tabulation:
```
mpharrigan: factor out into functions

```python
if tabulation:
  decompose_arbitrary_into_syc_tabulation(...)
else:
  decompose_arbitrary_into_syc_analytic(...)
```

cirq/google/optimizers/two_qubit_gates/gate_compilation.py
```diff
@@ -80,7 +78,7 @@ def compile_two_qubit_gate(self,
             unitaries and resulting product above.
         """
         unitary = np.asarray(unitary)
-        kak_vec = kak_vector(unitary, check_preconditions=False)
+        kak_vec = cirq.kak_vector(unitary, check_preconditions=False)
```
mpharrigan: The cirq convention for internal library code is to import the submodule that contains the function

```python
from cirq import linalg

linalg.kak_vector
```

cirq/google/optimizers/convert_to_sycamore_gates.py
KevinVillela: Done

cirq/google/optimizers/two_qubit_gates/gate_compilation.py
KevinVillela: Done

cirq/google/optimizers/convert_to_sycamore_gates_test.py
```diff
@@ -223,3 +225,18 @@ def test_convert_to_sycamore_equivalent_unitaries(gate):
     u1 = cirq.unitary(cirq.Circuit(converted))
     u2 = cirq.unitary(operation)
     cirq.testing.assert_allclose_up_to_global_phase(u1, u2, atol=1e-8)
+
+
+def test_convert_to_sycamore_tabulation():
+    # A tabulation for the sycamore gate with an infidelity of .1.
+    sycamore_tabulation = gate_product_tabulation(cirq.unitary(cirq.google.SYC),
+                                                  .1)
```
dkafri: I would specify a random_state argument to make sure the test is deterministic.

cirq/google/optimizers/convert_to_sycamore_gates.py
```diff
@@ -43,14 +45,19 @@ class ConvertToSycamoreGates(circuits.PointOptimizer):
...
     """
 
-    def __init__(self, ignore_failures=False) -> None:
+    def __init__(self,
+                 tabulation: Optional[GateTabulation] = None,
+                 ignore_failures=False) -> None:
         """
         Args:
+            tabulation: If set, a tabulation for the Sycamore gate to use for
+                decomposing gates.
```
mpharrigan: re: @dkafri 's suggestion. I'd expand on this docstring more

 1. tabulation only used for matrix gate
 2. otherwise you use another scheme

cirq/google/optimizers/convert_to_sycamore_gates.py
```diff
@@ -134,68 +141,124 @@ def optimization_at(self, circuit, index, op):
...
+        else:
+            return decompose_arbitrary_into_syc_analytic_(qubit_a, qubit_b, op)
+    else:
+        raise ValueError("Unrecognized gate: {!r}".format(op))
 
-        Args:
-            qubit_a: first qubit of GateOperation
-            qubit_b: second qubit of GateOperation
+
+def decompose_arbitrary_into_syc_tabulation_(qubit_a: ops.Qid, qubit_b: ops.Qid,
```
mpharrigan: why do these methods have a trailing `_`? afaik this isn't a convention in python

cirq/google/optimizers/convert_to_sycamore_gates.py
mpharrigan: 3. how do I get a tabulation

cirq/google/optimizers/convert_to_sycamore_gates.py
```diff
@@ -43,14 +45,19 @@ class ConvertToSycamoreGates(circuits.PointOptimizer):
...
         """
         Args:
+            tabulation: If set, a tabulation for the Sycamore gate to use for
+                decomposing gates.
             ignore_failures: If set, gates that fail to convert are forwarded
                 unchanged. If not set, conversion failures raise a TypeError.
         """
         super().__init__()
         self.ignore_failures = ignore_failures
+        self.tabulation = tabulation
```
mpharrigan: I would do some validation here, like checking that tabulation is None or actually a valid `GateTabulation`. Otherwise the user gets a more mysterious "tabulation object has no compile_two_qubit_gate" method

cirq/google/optimizers/convert_to_sycamore_gates.py
KevinVillela: I plead the fifth.

cirq/google/optimizers/convert_to_sycamore_gates.py
KevinVillela: Done - but humor me here, as I'm not great at Python. Isn't that was mypy is supposed to do?

cirq/contrib/quantum_volume/quantum_volume.py
```diff
@@ -197,6 +198,22 @@ def process_results(mapping: Dict[cirq.Qid, cirq.Qid],
...
 
 
+class SwapPermutationReplacer(cirq.PointOptimizer):
+    """Replaces SwapPermutationGates with their underlying implementation
+    gate."""
+
+    def __init__(self):
+        super().__init__()
+
+    def optimization_at(self, circuit, index, op):
```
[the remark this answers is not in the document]
KevinVillela: Done.

cirq/google/optimizers/convert_to_sycamore_gates.py
KevinVillela: Done.

cirq/google/optimizers/convert_to_sycamore_gates.py
```diff
@@ -147,6 +156,8 @@ def known_two_q_operations_to_sycamore_operations(self, qubit_a: ops.Qid,
             qubit_a: first qubit of GateOperation
             qubit_b: second qubit of GateOperation
             op: operation to decompose
+            tabulation: A tabulation for the Sycamore gate to use for
+                decomposing gates.
```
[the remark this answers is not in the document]
KevinVillela: Done.

cirq/google/optimizers/convert_to_sycamore_gates_test.py
KevinVillela: Good call, done.

KevinVillela:
Thanks for the comments! The test failure appears to be unrelated AFAICT.

cirq/google/optimizers/convert_to_sycamore_gates.py
mpharrigan: mypy is basically a linter. The type annotations do nothing at runtime

cirq/google/optimizers/convert_to_sycamore_gates.py
```diff
@@ -134,68 +147,124 @@ def optimization_at(self, circuit, index, op):
...
+
+    Args:
+        qubit_a: first qubit of the operation
+        qubit_b: second qubit of the operation
+        op: operation to decompose
+        tabulation: A tabulation for the Sycamore gate.
+    Returns:
+        New operations iterable object
+
+        """
```
mpharrigan: nit: extra newline and extra indent on `"""`

cirq/google/optimizers/convert_to_sycamore_gates.py
```diff
@@ -51,12 +51,18 @@ def __init__(self,
...
+                constructor a base gate (in this case, usually cirq.google.SYC)
+                and a maximmum infidelity.
             ignore_failures: If set, gates that fail to convert are forwarded
                 unchanged. If not set, conversion failures raise a TypeError.
         """
         super().__init__()
         self.ignore_failures = ignore_failures
+        assert tabulation is None or isinstance(
+            tabulation, GateTabulation
+        ), "provided tabulation must be of type GateTabulation"
```
mpharrigan: one additional python thing: it's more idiomatic to do

```python
if [whatever check is violated]:
  raise ValueError("message")
```

instead of using asserts. A violated assertion returns an `AssertionError` whereas a `ValueError` is more appropriate

cirq/google/optimizers/convert_to_sycamore_gates.py
```diff
@@ -51,12 +51,18 @@ def __init__(self,
         """
         Args:
             tabulation: If set, a tabulation for the Sycamore gate to use for
-                decomposing gates.
+                decomposing Matrix gates. If unset, an analytic calculation is
+                used for Matrix gates. To get a GateTabulation, provide its
+                constructor a base gate (in this case, usually cirq.google.SYC)
```
mpharrigan: oh interesting. So the tabulation is done in the constructor of `GateTabulation`? It might be slightly nicer to have a function that does the tabulation (i.e. the hard work) and returns a filled-in `GateTabulation` object. I personally don't expect a constructor to do work. This is probably related to my dislike for over-oop (in python; I know it's different in other languages). 

However, this is fine at present

dkafri: Actually, the code behaves as you suggest. The `GateTabulation` object just stores the result of the tabulation and uses it to compile gates. The actual construction is done with `gate_product_tabulation`, which Kevin moved [here](https://github.com/quantumlib/Cirq/blob/04e4fbcce767ac342b10f56cf185ac9dba4a181b/cirq/google/optimizers/two_qubit_gates/gate_compilation.py).

cirq/google/optimizers/convert_to_sycamore_gates.py
KevinVillela: Done

cirq/google/optimizers/convert_to_sycamore_gates.py
KevinVillela: Ah good point, done.

cirq/google/optimizers/convert_to_sycamore_gates.py
KevinVillela: Woops, updated the comment. Thanks for catching!

cirq/google/optimizers/convert_to_sycamore_gates.py
KevinVillela: Sure, but in theory, it would prevent this situation from ever happening, because we wouldn't be able to submit any code that violates the types. Coming from a TypeScript background, the type system doesn't do anything at runtime but can be relied on to prevent these sort of situations because it runs on every compile. I guess the difference here is that mypy is not run as often, e.g. during development.

cirq/google/optimizers/convert_to_sycamore_gates.py
mpharrigan: ah good. yeah that function name is what I wanted documented :)

cirq/google/optimizers/convert_to_sycamore_gates.py
mpharrigan: Bear in mind that cirq is basically an SDK and users will be calling this function from their hacky scripts. This type checking and my feedback about documenting "where you get a tabulation object" are two sides of the same coin: we've got to help the users use this new feature!

KevinVillela: Sounds reasonable to me, thanks for the guidance!

PR #2616, opened 2019-12-03
Add pretty __str__ for devices with grid qubits

maffoo:
I also cleaned up a few little things in the SerializableDevice code while I was here, but I can back those out and submit separately if you'd prefer.


cirq/google/devices/serializable_device.py
```diff
@@ -185,35 +185,66 @@ def from_proto(
...
+
+            qubits = cast(List['cirq.GridQubit'], self.qubits)
+            for q in qubits:
+                print()
+                diagram.write(q.col, q.row, str(q))
+
+            # Find pairs that are connected by two-qubit gates.
+            Pair = Tuple['cirq.GridQubit', 'cirq.GridQubit']
+            pairs: Set[Pair] = set()
+            for gate_defs in self.gate_definitions.values():
```
Strilanc: Works well as a comprehension:

```
pairs = {
    cast(Pair, pair)
    for gate_defs in self.gate_definitions.values()
    for gate_def in gate_defs
    if gate_def.number_of_qubits == 2
    for pair in gate_def.target_set
    if len(pair) == 2
}
```

maffoo: Done.

Issue #2282, opened 2019-10-08
contrib.routing algorithm hangs


kevinsung:
The repro code below constructs a QAOA circuit and then tries to route it on a device 100 times, printing some output before each attempt. It prints up to `Attempt 1` and then hangs.

```
import random
import networkx as nx
import numpy as np
import cirq
from cirq.contrib import routing


def random_3_regular_model(n, seed):
    connectivity_graph = nx.random_regular_graph(3, n, seed=seed)
    graph = nx.Graph()
    for i, j in connectivity_graph.edges:
        graph.add_edge(cirq.LineQubit(i), cirq.LineQubit(j),
                       weight=1)
    return graph


def Rzz(rads):
    return cirq.ZZPowGate(exponent=2 * rads / np.pi, global_shift=-0.5)


def qaoa_rzz_layer(graph, gamma):
    return (Rzz(gamma * graph[q][r]['weight']).on(q, r)
            for q, r in graph.edges)


def qaoa_rx_layer(graph, beta):
    return cirq.Rx(2 * beta).on_each(*graph.nodes)


def qaoa_unitary(graph, gammas, betas):
    for gamma, beta in zip(gammas, betas):
        yield qaoa_rzz_layer(graph, gamma)
        yield qaoa_rx_layer(graph, beta)


def qaoa_circuit(graph, gammas, betas):
    return cirq.Circuit.from_ops(
        cirq.H.on_each(*graph.nodes),
        qaoa_unitary(graph, gammas, betas))


def _parse_device(s: str):
    lines = s.strip().split('\n')
    qubits = []  # type: List[GridQubit]
    for row, line in enumerate(lines):
        for col, c in enumerate(line.strip()):
            if c != '-':
                qubit = cirq.GridQubit(row, col)
                qubits.append(qubit)
    return qubits


device_grid = """
--A----
-ABC---
ABCDE--
-CDEFG-
--EFGHI
---GHI-
----I--
"""
device_qubits = _parse_device(device_grid)
device_graph = routing.gridqubits_to_graph_device(device_qubits)

n = 20
p = 2

prng = np.random.RandomState(36587)

gammas = prng.randn(p)
betas = prng.randn(p)

graph = random_3_regular_model(n, seed=prng)
circuit = qaoa_circuit(graph, gammas, betas)

for i in range(100):
    print(f'Attempt {i}')
    _ = routing.route_circuit(circuit, device_graph, algo_name='greedy', random_state=prng)
```

[...]

PR #2619, opened 2019-12-04
fix of router hanging #2282

askolik:
When generating candidate swaps to perform for the next set of operations to apply, the routing algorithm only checked if there is a candidate swap. It did not check if performing this swap will lead to a new adjacent connection in the graph, considering the max_search_radius. This leads to the algorithm being stuck in an infinite loop for some graph instances, where it keeps performing swaps that don't lead to adjacent connections for the remaining operations.

Fixed this by adding a check for adjacent connections in the new graph, and if there are no new adjacent connections, perform the swap that brings together the farthest nodes.

Fixes: https://github.com/quantumlib/Cirq/issues/2282


askolik:
@googlebot I fixed it.

cirq/contrib/routing/greedy.py
```diff
@@ -284,7 +284,9 @@ def apply_next_swaps(self):
                 ]
                 if len(candidate_swap_sets) == 1:
                     self.apply_swap(*candidate_swap_sets[0])
-                    return
+
+                    if len(list(self.remaining_dag.findall_nodes_until_blocked(self.acts_on_nonadjacent_qubits))) > 0:
```
bryano:
```suggestion
                    if list(self.remaining_dag.findall_nodes_until_blocked(self.acts_on_nonadjacent_qubits))):
                        return
                    else:
                        break               
```

bryano: No reason to keep going through the time slices.

askolik: You're right, fixed.

PR #2621, opened 2019-12-05
Add random quantum circuit generation

kevinsung:
Adapted from pyle.

Let me know if the name `random_quantum_circuit` is too generic. What would be a better name?

@XiaoMiQC


kevinsung:
mypy complains
```
cirq/experiments/random_quantum_circuit.py:224: error: List comprehension has incompatible type List[Operation]; expected List[GateOperation]
```
but #2626

kevinsung:
Actually I modified the code so mypy no longer complains.

cirq/experiments/__init__.py
```diff
@@ -30,6 +30,13 @@
     xeb_fidelity,
 )
 
+from cirq.experiments.random_quantum_circuit import (
```
Strilanc: The name of this code needs to be more descriptive. "Random quantum circuit" can mean a lot of things. Make it like "random_rotations_between_interaction_layers" or something. Be as descriptive as possible... although the name should fit on a line.

cirq/experiments/random_quantum_circuit_test.py
```diff
@@ -0,0 +1,233 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import cirq
+
+
+def test_random_quantum_circuit():
```
viathor: This test verifies too little too strictly. Consequently, when someone changes the RQC generation algorithm in future, they'll be forced to re-run it and copy-paste the new circuit for comparison. If in the process an important property of the circuits which we may want to keep invariant changes, we would not be alerted to such unintended consequence.

Instead, we should test for specific properties we want these circuits to have, e.g.:
- each layer contains either only single-qubit gates or only two-qubit gates,
- single- and two-qubit layers alternate,
- first and last layer are single-qubit layers,
- single-qubit gates do not repeat between layers k and k+1,
- two-qubit gate patterns cycle as appropriate.

cirq/experiments/random_quantum_circuit_test.py
```diff
@@ -0,0 +1,233 @@
...
+    # pylint: disable=line-too-long
+    qubits = cirq.GridQubit.rect(3, 2)
+    depth = 9
+    circuit = cirq.experiments.random_quantum_circuit(qubits, depth, seed=1234)
+    cirq.testing.assert_has_diagram(circuit,
+                                    """
```
viathor: Here we're unintentionally testing diagram drawing code. Consequently, when someone changes that code they will have to modify the tests for RQC generation.

The tests should examine the generated circuit instead. Fortunately, the Circuit and related classes make it easy since they expose pythonic API for iteration and item retrieval.

cirq/experiments/random_quantum_circuit.py
```diff
@@ -0,0 +1,229 @@
...
+if TYPE_CHECKING:
+    import numpy as np
+    import cirq
+
+# A set of 2-qubit interactions that can be performed simultaneously on a grid.
+TwoQubitInteractionLayer = Container[
+    Tuple[devices.GridQubit, devices.GridQubit]]
+
+
+class GmonLayer(TwoQubitInteractionLayer):
```
viathor: I'd remove "Gmon" from the name since this could conceivably be used with other devices as long as they provide a full grid of qubits. Perhaps `GridLayer`?

cirq/experiments/random_quantum_circuit.py
```diff
@@ -0,0 +1,229 @@
...
+        # qubits should be 1 column apart.
+        if (a.row != b.row) or (b.col != a.col + 1):
+            return False
+
+        # mod to get the position in the 2 x 2 unit cell with column offset.
+        pos = a.row % 2, (a.col - self.col_offset) % 2
+        return pos == (0, 0) or pos == (1, self.stagger)
+
+
+GMON_HARD_PATTERN = (
```
viathor: nit: Perhaps `GRID_STAGGERED_PATTERN` as it's more descriptive?

cirq/experiments/random_quantum_circuit.py
```diff
@@ -0,0 +1,229 @@
...
+)
+document(
+    GMON_HARD_PATTERN, """A pattern of two-qubit gates that is hard to simulate.
+
+    This pattern of gates was used in the paper
+    https://www.nature.com/articles/s41586-019-1666-5
+    to demonstrate quantum supremacy.
+    """)
+
+GMON_EASY_PATTERN = (
```
viathor: nit: `GRID_NON_STAGGERED_PATTERN` or `GRID_PARALLLEL_PATTERN?

cirq/experiments/__init__.py
viathor: I think we could keep the module name as is or perhaps just add "generation" to it (as in `random_quantum_circuit_generation` or `random_circuit_generation`) with the intention for any future RQC-generation code to be added here as well.

cirq/experiments/__init__.py
```diff
@@ -30,6 +30,13 @@
...
+    GMON_EASY_PATTERN,
+    GMON_HARD_PATTERN,
+    GmonLayer,
+    random_quantum_circuit,
```
viathor: As suggested by @Strilanc above, we should name this more specific. This will also make the module future-proof by preventing name collisions. Example name for the function: `make_layered_random_circuit`.

cirq/experiments/__init__.py
kevinsung: I renamed the module to `random_quantum_circuit_generation`

cirq/experiments/random_quantum_circuit_test.py
kevinsung: Done.

cirq/experiments/random_quantum_circuit_test.py
kevinsung: Done.

cirq/experiments/random_quantum_circuit.py
kevinsung: I renamed it to `GridInteractionLayer` and deleted `TwoQubitInteractionLayer`.

cirq/experiments/random_quantum_circuit.py
kevinsung: Done.

cirq/experiments/random_quantum_circuit.py
kevinsung: I chose `GRID_PARALLEL_PATTERN`.

cirq/experiments/__init__.py
kevinsung: I renamed it `random_rotations_between_grid_interaction_layers_circuit`.

kevinsung:
Windows build is failing.

cirq/experiments/__init__.py
```diff
@@ -30,6 +30,13 @@
     xeb_fidelity,
 )
 
+from cirq.experiments.random_quantum_circuit_generation import (
+    GRID_PARALLEL_PATTERN,
```
viathor: nit: OK, this is my fault for giving you only bad suggestions to choose from (GRID_PARALLEL_PATERN and GRID_NON_STAGGERED_PATTERN), but here is a good name for this pattern: GRID_ALIGNED_PATTERN.

The issue is that "parallel" is a poor term for the pattern's shape: both patterns (staggered and aligned) are in fact parallel since the term only speaks to the direction of the active couplers, not to the endpoint locations. "Aligned" is the term that means the endpoints' locations match up.

I apologize for nitpicking on the names - you've certainly chosen the better of the two options I had suggested earlier.

cirq/experiments/random_quantum_circuit_generation.py
```diff
@@ -0,0 +1,226 @@
...
+    import numpy as np
+    import cirq
+
+
+class GridInteractionLayer(
+        Container[Tuple[devices.GridQubit, devices.GridQubit]]):
+    """A layer of parallel or staggered two-qubit interactions on a grid.
+
+    Layers of this type have two different basic structures,
+    parallel:
```
viathor: nit: s/parallel/aligned/

cirq/experiments/random_quantum_circuit_generation.py
```diff
@@ -0,0 +1,226 @@
...
+    *-* *-* *-*
+    * *-* *-* *
+    *-* *-* *-*
+    * *-* *-* *
+
+    Other variants are obtained by offsetting these lattices to the right by
+    some number of columns, and/or transposing into the vertical orientation.
+    There are a total of 4 parallel and 4 staggered variants.
+
+    The 2x2 unit cells for the parallel and staggered versions of this layer
```
viathor: ditto

cirq/experiments/random_quantum_circuit_generation.py
```diff
@@ -0,0 +1,226 @@
...
+
+    return pairs
+
+
+def _single_qubit_layer(
+        qubits: List['cirq.GridQubit'],
+        single_qubit_gates: Sequence['cirq.Gate'],
+        previous_single_qubit_layer: Dict['cirq.GridQubit', 'cirq.Gate'],
+        prng: 'np.random.RandomState',
+) -> Dict['cirq.GridQubit', 'cirq.Gate']:
```
viathor: For defense-in-depth, you could check consistency of qubits and previous layer:

```python
if not previous_single_qubit_layer and set(qubits) != set(previous_single_qubit_layer.keys()):
    raise ValueError(...)
```

kevinsung: I think you put the `not` there by mistake. I think this is quite unnecessary; I'd rather omit it.

cirq/experiments/__init__.py
kevinsung: I agree. Done.

cirq/experiments/random_quantum_circuit_generation.py
kevinsung: Done.

cirq/experiments/random_quantum_circuit_generation.py
kevinsung: Done.

cirq/experiments/__init__.py
kevinsung: @Strilanc you okay with this?

cirq/experiments/random_quantum_circuit.py
```diff
@@ -0,0 +1,229 @@
+from typing import (Callable, Container, Dict, Iterable, List, Sequence,
```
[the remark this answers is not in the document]
kevinsung: done.

PR #2622, opened 2019-12-05
Return unrecognized random state unmodified

kevinsung:
This allows people to use their own custom pseudorandom number generator implementations.


bryano:
@kevinsung Do you envision the PRNGs people use being written specifically for Cirq or from a general package? I think a PRNG abstract class would be better than allowing anything, but it would require a Cirq-specific implementation (or at least a wrapper).

kevinsung:
I have an application where the PRNG is not related to Cirq but can be used as a substitute for a RandomState in appropriate situations.

bryano:
Gotchya. In that case, we might as well just set `RANDOM_STATE_LIKE` to `Any`. We could get rid of it entirely, but I think it's worth keeping as documentation.

cirq/value/random_state.py
```diff
@@ -12,20 +12,25 @@
...
 
 import numpy as np
 
-RANDOM_STATE_LIKE = Optional[Union[np.random.RandomState, int]]
+RANDOM_STATE_LIKE = Optional[Union[np.random.RandomState, int, Any]]
 
 
 def parse_random_state(random_state: RANDOM_STATE_LIKE
                       ) -> np.random.RandomState:
+    """Interpret an object as the state of a pseudorandom number generator.
```
vtomole: Add `Args` and `Returns` to this docstring

cirq/value/random_state.py
```diff
@@ -12,20 +12,25 @@
...
     if random_state is None:
         return cast(np.random.RandomState, np.random)
-    elif (isinstance(random_state, np.random.RandomState) or
-          random_state == np.random):
-        return cast(np.random.RandomState, random_state)
     elif isinstance(random_state, int):
         return np.random.RandomState(random_state)
-    raise TypeError(f'Argument must be of type cirq.value.RANDOM_STATE_LIKE.')
+    else:
+        return cast(np.random.RandomState, random_state)
```
Strilanc: Add a check for one of the expected methods as a safety. Otherwise very pythonic.

kevinsung: On further thought, I'd prefer not to make any such check. The user should not be forced to implement any particular method; she should only have to implement the ones that she knows are needed for her use case. For instance, the Cirq simulator only uses `prng.choice`, while the supremacy circuit generation code only uses `prng.random`.

cirq/value/random_state.py
kevinsung: Done.

cirq/value/random_state.py
kevinsung: Per our offline conversation, I have added documentation to `RANDOM_STATE_LIKE` and left this alone.

Issue #2468, opened 2019-10-31
test_single_qubit_op_to_framed_phase_form_output_on_example_case fails on my machine and sometimes on Travis Windows build


kevinsung:
```
_____________________________________ test_single_qubit_op_to_framed_phase_form_output_on_example_case _____________________________________

    def test_single_qubit_op_to_framed_phase_form_output_on_example_case():
        u, t, g = cirq.single_qubit_op_to_framed_phase_form(
            cirq.unitary(cirq.Y**0.25))
>       cirq.testing.assert_allclose_up_to_global_phase(u, cirq.unitary(cirq.X**0.5), atol=1e-7)

optimizers/decompositions_test.py:180: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

actual = array([[ 7.07106781e-01-2.23668991e-17j, -2.70720689e-16-7.07106781e-01j],
       [-7.07106781e-01+3.73578175e-16j, -2.70720689e-16-7.07106781e-01j]])
desired = array([[7.07106781e-01+5.55111512e-17j, 5.55111512e-17-7.07106781e-01j],
       [5.55111512e-17-7.07106781e-01j, 7.07106781e-01+5.55111512e-17j]])

    def assert_allclose_up_to_global_phase(
            actual: np.ndarray,
            desired: np.ndarray,
            *,  # Forces keyword args.
            rtol: float = 1e-7,
            atol: float,  # Require atol to be specified
            equal_nan: bool = True,
            err_msg: Optional[str] = '',
            verbose: bool = True) -> None:
        """Checks if a ~= b * exp(i t) for some t.
    
        Args:
            actual: A numpy array.
            desired: Another numpy array.
            rtol: Relative error tolerance.
            atol: Absolute error tolerance.
            equal_nan: Whether or not NaN entries should be considered equal to
                other NaN entries.
            err_msg: The error message to be printed in case of failure.
            verbose: If True, the conflicting values are appended to the error
                message.
    
        Raises:
            AssertionError: The matrices aren't nearly equal up to global phase.
        """
        actual, desired = linalg.match_global_phase(actual, desired)
        np.testing.assert_allclose(
            actual=actual,
            desired=desired,
[...]

PR #2630, opened 2019-12-06
Mark test that sometimes fails on windows as xfail.

maffoo:
See #2468. This doesn't fix the bug, obviously, but the flaky test causes failures in unrelated PRs and I think we should try to avoid that.

### DIFF
--- a/cirq/optimizers/decompositions_test.py
+++ b/cirq/optimizers/decompositions_test.py
@@ -14,6 +14,7 @@
 
 import math
 import random
+import sys
 from typing import Sequence
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


maffoo:
@vtomole, I agree that could be a risk. The problem is that a flaky test like this slows down _everyone_ working on the code, not just people who might be working on the bug in question or be qualified to fix it. We should definitely assign that bug so someone is responsible for it, not just wait for someone to get sufficiently annoyed :-)

vtomole:
LGTM. Going to merge.

vtomole:
Note: Have to revert once we find a fix.

CC: @kevinsung  @maffoo @Strilanc

PR #2632, opened 2019-12-10
Update dominated_indices in greedy.py

tbcdebug:
Previously, the condition is `if all(v>=w)`. Suppose that the vectors are just scalars and there are two minimal value v1 and v2. Now, because v1 >= v2 and v2 >= v1, they are both added to `dominated_indices`. This means that the corresponding swap candidates both won't be considered. Additionally for any w >= v1 = v2, the corresponding swap candidate for w won't be considered. In the end, this means that none the swap candidates will be considered. In stead, the algorithm falls back everytime on `bring_farthest_pair_together`.


bryano:
It's not clear to me that this is always better, though it might be. Let's make it an option: introduce a `strict_cmp: bool` argument to `_get_dominated_indices` and attribute to `GreedyRouter`.

dabacon:
@mrwojtek any opinion here?

mrwojtek:
I didn't have a chance to use this code yet but having another option to tune the heuristic is something that I would definitely appreciate very much.

balopat:
Hi @tbcdebug - are you planning to work on this still? In case yes, would you mind adding a parameter to tune the heuristic as suggested by @bryano and @mrwojtek? Also a test case covering it would be great.

dabacon:
It seems like this has stalled out.  I'm going to close, but if there is a desire to revive this, please reopen.

PR #2633, opened 2019-12-10
Fix documentation and key error in calibration

dstrain115:
- Tried to retrieve historical calibration and it was broken.
- Fixed docs to my understanding and fixed key error.
- Please verify that this is correct now per the API.

### DIFF
--- a/cirq/google/engine/engine.py
+++ b/cirq/google/engine/engine.py
@@ -737,7 +737,8 @@ class Engine:
 
         Params:
             calibration_name: A string of the form
-                `<processor name>/calibrations/<ms since epoch>`
+                `projects/<project_id>/processors/<processor name>`
+                `/calibrations/<ms since epoch>`
 
         Returns:
             A dictionary containing the metadata.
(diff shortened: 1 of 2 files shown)


cirq/google/engine/engine.py
```diff
@@ -737,15 +737,16 @@ def get_calibration(self, calibration_name: str) -> calibration.Calibration:
 
         Params:
             calibration_name: A string of the form
-                `<processor name>/calibrations/<ms since epoch>`
+                `projects/<project_id>/processors/<processor name>`
+                `/calibrations/<ms since epoch>`
```
lindmarkm: The original doc using <processor name> is correct (except for the timestamp units) but I'm not certain what convention is being used here.
The full resource name for a calibration object is:
`projects/<project id>/processors/<processor id>/calibrations/<timestamp in seconds since epoch>`

dstrain115: Yeah, it looks like you need the whole resource name or else it fails.  Good catch on the units.

PR #2645, opened 2019-12-15
Quantum Volume: Add the ability to specify the device graph to compile the model circuit onto

KevinVillela:
This should give us the ability to try different parts of the device to find the optimal configuration. Note that the qubits that were used are already returned in the QuantumVolumeResult.compiled_circuit property.

@mpharrigan


KevinVillela:
Filed https://github.com/quantumlib/Cirq/issues/2646 for the Travis failure.

cirq/contrib/quantum_volume/quantum_volume.py
```diff
@@ -473,9 +474,12 @@ def calculate_quantum_volume(
                                 depth=depth,
                                 num_circuits=num_circuits,
                                 random_state=random_state)
+    device_graph = ccr.gridqubits_to_graph_device(device_or_qubits) if type(
```
mpharrigan: python nit:

```python
if isintance(device_graph, list):
  device_graph = blah
else:
  device_graph = blah2
```

KevinVillela: Done

KevinVillela:
Windows pytest failed...
Wtf is going on today haha

cirq/contrib/quantum_volume/quantum_volume.py
```diff
@@ -475,9 +476,14 @@ def calculate_quantum_volume(
                                 depth=depth,
                                 num_circuits=num_circuits,
                                 random_state=random_state)
-    device_graph = ccr.gridqubits_to_graph_device(device_or_qubits) if type(
-        device_or_qubits) is list else ccr.xmon_device_to_graph(
-            device_or_qubits)
+
+    # Get the device graph from the given qubits or device.
+    device_graph = None
```
mpharrigan: thing of interest: usually you don't have to include this. most linters are smart enough to notice that `device_graph` is initialized no matter what and all the scoping in python is so horribly leaky you don't need to declare it outside

KevinVillela: mypy was complaining that the variable device_graph was referenced before assignment. I wasn't able to find an existing issue for this online, maybe its WAI?

mpharrigan: ah interesting

PR #2647, opened 2019-12-16
Pin sympy version to 1.4

KevinVillela:
Cirq appears to be incompatible with the latest SymPy release of 1.5: https://github.com/sympy/sympy/releases/tag/sympy-1.5
Fixes #2646

PR #2648, opened 2019-12-16
Pin all versions in requirements.txt to prevent against breakages

### LINKED ISSUE #2646: Cirq tests appear to be failing
I have two PRs that are completely unrelated: https://github.com/quantumlib/Cirq/pull/2645 and https://github.com/quantumlib/Cirq/pull/2629. They both cause the same (unrelated to the PRs) tests to fail: https://travis-ci.com/quantumlib/Cirq/jobs/267385668. Running the following on my machine passes in either of these PRs passes:
$ pip install --upgrade pip setuptools wheel
$ python -m venv venv
$ source venv/bin/activate
$ python -m pip install -r requirements.txt
$ python -m pip install -r cirq/contrib/contrib-requirements.txt
$ python -m pip install -r dev_tools/conf/pip-list-dev-tools.txt
$ ./check/all

KevinVillela:
Cirq appears to be incompatible with the latest SymPy release of 1.5: https://github.com/sympy/sympy/releases/tag/sympy-1.5. While we're here, might as well pin em all.
Fixes #2646
@vtomole This is a PR to replace https://github.com/quantumlib/Cirq/pull/2647 because I messed up that PRs commit history something fierce.


mpharrigan:
If we do this, we should remove the part of `setup.py` that reads in the `requirements.txt` file and uses it directly, xref https://github.com/quantumlib/OpenFermion-Cirq/pull/352 xref https://packaging.python.org/discussions/install-requires-vs-requirements/

We should also open an issue to fix Cirq to work with the latest sympy

KevinVillela:
Done. Also opened https://github.com/quantumlib/Cirq/issues/2650.

vtomole:
@mpharrigan  Why don't we want to use it directly?

mpharrigan:
Please see the packaging.python.org link above, but in summary:

 1. requirements.txt is for setting up a reproducible environment
 2. install_requires is for listing the bare minimum reqs to get something working
 3. (me editorializing:) it's crucial especially for a developer-oriented library like cirq (i.e. not an end-user application) to support some degree of freedom in the developer environment. I don't appreciate it when libraries keep over-writing / re-installing my version of numpy. You can get into even trickier situations (see linked openfermion-cirq issue) where too many pins can make it impossible to create a consistent environment when *technically* it would work fine

KevinVillela:
Done. Is this change still desirable now that the breakage is fixed btw? Or do we want to just take breakages as they come and fix them on the fly?

mpharrigan:
for reference: fixed by #2655

setup.py
```diff
@@ -42,6 +42,10 @@
 # Read in requirements
 requirements = open('requirements.txt').readlines()
 requirements = [r.strip() for r in requirements]
+# install_requires should not be overly specific.
+# Change any == pinned deps to >=.
+# https://packaging.python.org/discussions/install-requires-vs-requirements/
+requirements = [r.replace('==', '>=') for r in requirements]
```
Strilanc: Remove this change. Our requirements.txt file should be the literal source of truth. This also causes the code installed on user machines to be broken, e.g. it uses the bad version of sympy.

mpharrigan: So I suggested this change. I agree that as written it has sympy issues. In general, I'd advocate for never being overly specific on which versions are required for a developer-oriented library. Usually the way python packages can handle this is by pinning things in `requirements.txt` so e.g. CI runs under a consistent environment and users who don't care can `pip install -r requirements.txt` and get a good environment and leaving `install_requires` with a looser set of restrictions so `pip install cirq` has a low probability of upgrading or downgrading an existing install of a package (particularly heavy-weight ones like matplotlib or numpy). 

If you insist on having one set of version specs for these two use cases, each restriction in `requirements.txt` should have a corresponding reason associated with it. If the reason is "there's a bug in the most recent release" then that's good and a future developer knows the condition under which it is safe to change the version spec. If the reason is "well I seem to have pandas 0.25.1 and I can't be bothered to check older or newer versions" ... that's less compelling

vtomole:
Forgot to revert my approval after @mpharrigan's comment.

setup.py
KevinVillela: If we are going to have reasons required for restrictions in `requirements.txt`, seems like this PR isn't useful anymore, because the SymPy stuff was fixed. I can revert if that's the direction we want to go in.

KevinVillela:
Seems like right now we will want to only pin requirements when necessary, and since SymPy is fixed there are nothing we need to actually pin. I'm going to revert this for now, but feel free to let me know if I should bring it back!

mpharrigan:
Thanks @KevinVillela ! I think that's the right call for now

PR #2651, opened 2019-12-16
Add with_qubits_mapped to SingleQubitReadoutCalibrationResult

kevinsung:
This is used when there is a nontrivial mapping between qubits abstractly defined for a problem and qubits on the device.

### DIFF
--- a/cirq/experiments/single_qubit_readout_calibration.py
+++ b/cirq/experiments/single_qubit_readout_calibration.py
@@ -43,6 +43,29 @@ class SingleQubitReadoutCalibrationResult:
     repetitions: int
     timestamp: float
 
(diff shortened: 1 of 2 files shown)


mpharrigan:
This is a little bizarre --  When you take data, aren't the results very-much tied to particular qubits?

From your PR description: wouldn't it make more sense to do the mapping during execution of the experiment and have the resultant objects reflect actual qubit values?

Strilanc:
Should be called `transform_qubits` to agree with other places in the code.

kevinsung:
After thinking about it more, I do think this is a weird way to do what I wanted. Qubits should be mapped separately from the calibration result.

Issue #2650, opened 2019-12-16
Cirq does not work with latest version of Sympy (1.5)


KevinVillela:
See https://github.com/quantumlib/Cirq/issues/2646 for details.

Strilanc:
Fixing in https://github.com/quantumlib/Cirq/pull/2655

PR #2655, opened 2019-12-16
Fix serialization breakage due to sympy.Rational now being a numbers.Number

Strilanc:
- Reorder the sympy checks to come before generic number checks
- Rename json.py to json_serialization.py to avoid collisions with the built-in json library
- Detect integral values when deserializing symbolic protos, so that x - y does not become x - 1.0*y.
- Pin to 1.4 until https://github.com/sympy/sympy/issues/18056 is fixed

Fixes https://github.com/quantumlib/Cirq/issues/2650
Fixes https://github.com/quantumlib/Cirq/issues/2646

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,18 +75,18 @@ matrix:
       - python -m pip install -r dev_tools/conf/pip-list-dev-tools.txt
       - sudo apt-get install pandoc
     script: dev_tools/build-docs.sh
-  - stage: test
-    os: osx
-    env: NAME=pytest (macOS)
-    python: '3.6'
-    language: generic
-    install:
-      - python3.6 -m venv venv
-      - source venv/bin/activate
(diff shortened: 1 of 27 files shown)


KevinVillela:
Yayyyy thanks!

Strilanc:
> I imagine sympy changed something so isinstance(obj, numbers.Integral) now returns True for sympy.Integer?

Yeah something like that.

Strilanc:
The OSX build is failing for some unknown reason

image

I have disabled it in this PR and will open an issue to re-enable it.

Strilanc:
You are killing me here, cirqbot.

Strilanc:
Urgh, just found https://github.com/sympy/sympy/issues/18056 which puts a big blocker on using the new version of sympy. I am going to pin to the old one.

PR #2663, opened 2019-12-20
Add documentation for Engine

dstrain115:
- Adds a stub page for Engine documentation.
- Includes information about device specification and calibration metrics.


dstrain115:
Ping.  Could I get a review for this documentation?

docs/google/devices.md
```diff
@@ -84,6 +84,44 @@ using a error rate of 0.5% per gate, a circuit of depth 20 and width 20 could be
...
 
+### Use sweeps when possible
+
+Round trip network time to and from the engine typically adds latency on the order of a second
+to the overall computation time.  Reducing the number of trips and allowing the engine to
+properly batch circuits can improve the throughput of your calculations.  One way to do this
+is to use parameter sweeps to send multiple variations of a circuit at once.
+
+One example is to turn single-qubit gates on or off by using parameter sweeps.  
+For instance, the following code demonstrates combining mesauring the |0> and the |1> state
```
wcourtney: "combining measuring" reads strangely. How about "the following code creates a single circuit that parameterizes measurement of the |0> and the |1> state"?

docs/google/devices.md
```diff
@@ -84,6 +84,44 @@ using a error rate of 0.5% per gate, a circuit of depth 20 and width 20 could be
...
+m0 = cirq.Circuit(cirq.measure(q))
+# Measure |1>
+m1 = cirq.Circuit(cirq.X(q), cirq.measure(q))
+
+# Executing the circuits requires two calls to the simulator (or engine)
+sim.run(m0)
+sim.run(m1)
+
+# Parameterized circuit
+m_parameterized=cirq.Circuit(cirq.X(q) ** sympy.Symbol('e'), cirq.measure(q))
```
wcourtney: Nit: "e" is a loaded term. What do you think of "i" to evoke e^pi*i?

docs/google/engine.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+More information on creating application default credentials can be found on the
+[Google cloud](https://cloud.google.com/docs/authentication/production) website.
+
+## Engine class
+
+The `Engine` class is the entry point to communicate with the API.
+
+It can be initialized using your project id (found within your cloud console).
```
wcourtney: Add a link to https://console.cloud.google.com?

docs/google/engine.md
```diff
@@ -0,0 +1,95 @@
...
+# Upload the program and submit jobs to run in one call.
+# Replace PROCESSOR_ID with the processor that you are allowed to run on.
+job = engine.run_sweep(
+    program=circuit,
+    program_id=name,
+    repetitions=10000,
+    processor_ids=['PROCESSOR_ID'],
+    gate_set=cirq.google.SYC_GATESET)
+
+# During this time, the job will be scheduled and pending execution.
```
wcourtney: s/During/At

docs/google/engine.md
```diff
@@ -0,0 +1,95 @@
...
+
+This proto can be queried directly to get information about the device or can be transformed
+into a `cirq.Device` by using `cirq.google.SerializableDevice.from_proto()` that will
+enforce constraints imposed by the hardware.
+
+
+## Calibration Metrics
+
+Metrics from the current status of the device can be retrieved using the `get_latest_calibration` 
+method of the `Engine` object.  This will return a python dictionary where each key is the
```
wcourtney: Capitalize "Python".

docs/google/engine.md
wcourtney: UX guidance is to use Google Cloud Platform to refer to the console UI on first use and "Cloud Platform" on subsequent use.

docs/google/devices.md
dstrain115: Used your suggestion.  Thanks.

docs/google/engine.md
dstrain115: Done.

docs/google/devices.md
dstrain115: Changed to use 't'.  I think that 'i' is also a loaded term in the same way that e is.

docs/google/engine.md
dstrain115: Done.

docs/google/engine.md
dstrain115: Changed to "Google Cloud Platform Console",  Is that what you mean, or should it just be "Google Cloud Platform"?

dstrain115:
Added documentation on device specification protos and gate durations.  Please take another look.

dstrain115:
Could I get another look at this?  I would love to get this documentation in so that I can continue improving our documentation.

docs/google/devices.md
```diff
@@ -84,6 +84,45 @@ using a error rate of 0.5% per gate, a circuit of depth 20 and width 20 could be
...
+### Use sweeps when possible
+
+Round trip network time to and from the engine typically adds latency on the order of a second
+to the overall computation time.  Reducing the number of trips and allowing the engine to
+properly batch circuits can improve the throughput of your calculations.  One way to do this
+is to use parameter sweeps to send multiple variations of a circuit at once.
+
+One example is to turn single-qubit gates on or off by using parameter sweeps.  
+For instance, the following code creates a single circuit that parameterizes
+the decision of measuring the |0> or the |1> state.  This code uses a simulator
```
Strilanc: A more relevant example would be Z axis vs X axis basis measurement. Flipping the bit is easy to do classically in the post-processing.

Here's my take:

```python
import cirq
import sympy
q = cirq.GridQubit(1, 1)
sampler = cirq.Simulator()

# STRATEGY #1: Have a separate circuit and sample call for each basis.
circuit_z = cirq.Circuit(
    cirq.measure(q, key='out'))
circuit_x = cirq.Circuit(
    cirq.H(q),
    cirq.measure(q, key='out'))
samples_z = sampler.sample(circuit_z, repetitions=5)
samples_x = sampler.sample(circuit_x, repetitions=5)

print(samples_z)
# prints something like:
#    out
# 0    0
# 1    0
# 2    0
# 3    0
# 4    0

print(samples_x)
# prints something like:
#    out
# 0    0
# 1    1
# 2    1
# 3    0
# 4    0

# STRATEGY #2: Have a parameterized circuit.
circuit_sweep = cirq.Circuit(
    cirq.H(q)**sympy.Symbol('t'),
    cirq.measure(q, key='out'))

samples_sweep = sampler.sample(circuit_sweep,
                               repetitions=5,
                               params=[{'t': 0}, {'t': 1}])
print(samples_sweep)
# prints something like:
#    t  out
# 0  0    0
# 1  0    0
# 2  0    0
# 3  0    0
# 4  0    0
# 0  1    0
# 1  1    1
# 2  1    1
# 3  1    0
# 4  1    1
```

docs/google/devices.md
```diff
@@ -105,6 +144,16 @@ of the circuit by pushing them back to the next non-commuting operator. If the r
 circuit still contains Z operations, they should be aggregated into their own moment,
 if possible.
 
+### Use caution with symbols
+
+Symbols are extremely useful for constructing parameterized circuits (see above).  However,
+only some sympy formulas can be serialized for network transport to the engine.
+Currently, only linear combinations of symbols are supported.  See `cirq.google.arg_func_langs`
```
Strilanc: We support sums and products, not just linear combinations. For example, `x*x*y` is allowed.

docs/google/engine.md
```diff
@@ -0,0 +1,103 @@
...
+## Engine class
+
+The `Engine` class is the entry point to communicate with the API.
+
+It can be initialized using your project id (found within your
+[Google Cloud Platform Console](https://console.cloud.google.com)).
+You can use this instance to run quantum circuits or sweeps (parameterized
+variants of a general circuit).
+
+```
```
Strilanc: Tag as python. You may also want to tag this block as one that shouldn't execute when testing the documentation.

I'm surprised that this doesn't cause the doc tests to fail, actually, since it should attempt to connect and fail.

docs/google/engine.md
```diff
@@ -0,0 +1,103 @@
...
+# A simple sample circuit
+qubit = cirq.GridQubit(5, 2)
+circuit = cirq.Circuit(
+    cirq.X(qubit)**0.5,                 # Square root of NOT.
+    cirq.measure(qubit, key='result')   # Measurement.
+)
+
+# Create an Engine object to use.
+# Replace the project id with the id from your cloud project.
+engine = cirq.google.Engine(project_id='your_project_id',
```
Strilanc: instead of a string, use the variable YOUR_PROJECT_ID. This will cause python to barf if they forget to replace it, instead of trying to make a connection.

docs/google/engine.md
```diff
@@ -0,0 +1,103 @@
...
+                            proto_version=cirq.google.ProtoVersion.V2)
+
+# Create a unique name for the program.
+name = 'example-%s' % ''.join(random.choice(
+    string.ascii_uppercase + string.digits) for _ in range(10))
+
+# Upload the program and submit jobs to run in one call.
+# Replace PROCESSOR_ID with the processor that you are allowed to run on.
+job = engine.run_sweep(
```
Strilanc: I'd prefer `sampler = engine.sampler(processor_id=...)` and then just work with it like any other sampler. Ask @mpharrigan and friends what they find most useful.

docs/google/devices.md
dstrain115: LGTM, I like this better.  I took your code as verbatim.

docs/google/devices.md
dstrain115: Changed.

docs/google/engine.md
dstrain115: Done.  It seems that the doctest only tests cirq/**.py right now.

docs/google/engine.md
dstrain115: Done.

docs/google/engine.md
dstrain115: Changed to use sampler.

docs/google/engine.md
dstrain115: Nm, was looking in the wrong place.  docs/snippets_test.py is the thing that tests this.
Fixed the test by adding appropriate substitution to use a simulator.

docs/google/engine.md
```diff
@@ -0,0 +1,111 @@
...
+    cirq.X(qubit)**0.5,                 # Square root of NOT.
+    cirq.measure(qubit, key='result')   # Measurement.
+)
+
+# Create an Engine object to use.
+# Replace YOUR_PROJECT_ID with the id from your cloud project.
+engine = cg.Engine(project_id=YOUR_PROJECT_ID, proto_version=cg.ProtoVersion.V2)
+
+# Create a unique name for the program.
+name = 'example-%s' % ''.join(random.choice(
```
mpharrigan: this isn't used anywhere

dstrain115: Thanks, fixed.

docs/google/index.rst
```diff
@@ -10,3 +10,5 @@ of devices that have become public information.
     :maxdepth: 1
 
     devices
+    engine
```
mpharrigan: sorry; I know this is already merged but:

Now that there's headings in the TOC this seperate index page is a bit redundant and actually makes the TOC and TOC headings less useful. I propose removing this `docs/google.index.rst` and linking directly to `google/devices`, `google/engine`, and `google/specification` directly from `docs/index.rst`

PR #2670, opened 2020-01-05
Add Tags and TaggedOperation class

dstrain115:
Draft of basic Operator Tag functionality.

- Tags are a wrapper around strings that allow you to
mark a specific instance of an operator for special processing.
- This can be used, for example, to exclude certain operations
from optimizer routines or give operations information specific
to the underlying hardware.

This PR adds a Tag class which is a wrapper around a string and
can also be sub-classed.

It also adds a TaggedOperation class which wraps an Operation and
includes Tags on that Operation.


dstrain115:
Adding an additional reviewer.  I'd love to get this reviewed this week.  Thanks!

cirq/ops/raw_types.py
```diff
@@ -460,6 +468,104 @@ def validate_args(self, qubits: Sequence['cirq.Qid']):
         _validate_qid_shape(self, qubits)
 
 
+class Tag:
+    """A wrapper around string that can be used to differentiate operations.
```
Strilanc: What's the benefit of wrapping string instead of just using a raw string? Seems like easily removed boilerplate.

cirq/ops/raw_types.py
```diff
@@ -460,6 +468,104 @@ def validate_args(self, qubits: Sequence['cirq.Qid']):
...
+
+
+@value.value_equality
+class TaggedOperation(Operation):
+    """A specific operation instance that has been identified with a set
+    of Tags for special processing.  This can be initialized with
+    Using Operation.with_tags(tag) or by TaggedOperation(op, tag).
+    """
+
+    def __init__(self, sub_operation: 'cirq.Operation', *tags: 'cirq.Tag'):
```
Strilanc: I'd change the type of `tags` to just `Any`.

cirq/ops/raw_types.py
```diff
@@ -460,6 +468,104 @@ def validate_args(self, qubits: Sequence['cirq.Qid']):
...
+                      *control_qubits: 'cirq.Qid',
+                      control_values: Optional[Sequence[
+                          Union[int, Collection[int]]]] = None
+                     ) -> 'cirq.Operation':
+        return TaggedOperation(
+            self.sub_operation.controlled_by(*control_qubits,
+                                             control_values=control_values),
+            *self._tags)
+
+    @property
```
Strilanc: I think you need like twenty protocol methods here forwarding to the sub operation. For example, is this operation considered unitary when it wraps a unitary operation?

cirq/ops/raw_types.py
dstrain115: Removed.

cirq/ops/raw_types.py

cirq/ops/raw_types.py

cirq/ops/raw_types.py
```diff
@@ -445,6 +464,136 @@ def validate_args(self, qubits: Sequence['cirq.Qid']):
...
+    def _circuit_diagram_info_(self, args: 'cirq.CircuitDiagramInfoArgs'
+                              ) -> 'cirq.CircuitDiagramInfo':
+        return protocols.circuit_diagram_info(self.sub_operation, args,
+                                              NotImplemented)
+
+    def _trace_distance_bound_(self) -> float:
+        return protocols.trace_distance_bound(self.sub_operation)
+
+    def _phase_by_(self, phase_turns: float,
+                   qubit_index: int) -> 'TaggedOperation':
```
95-martin-orion: The return type here is TaggedOperation, but the method doesn't appear to preserve the tags.

dstrain115: Done.  The intended behavior is that tagging only applies to _that_ specific operation.  Any modifications should drop the tags.   This is probably the only way to have a consistent behavior.

For that reason, I have changed the return type to Operation.  I also modified controlled_by, which was not following this convention.

cirq/ops/raw_types.py
```diff
@@ -392,6 +393,24 @@ def with_qubits(self, *new_qubits: 'cirq.Qid') -> 'cirq.Operation':
                 `qubits` property.
         """
 
+    def with_tags(self, *new_tags: Any) -> 'cirq.TaggedOperation':
```
95-martin-orion: Should `TaggedOperation` override this to add `new_tags` to itself (or replace its tags with `new_tags`), instead of nesting it in another `TaggedOperation`?

dstrain115: Good idea.  I added a with_tags() on TaggedOperation that avoids nesting.

cirq/ops/raw_types.py
```diff
@@ -445,6 +464,143 @@ def validate_args(self, qubits: Sequence['cirq.Qid']):
...
+
+    def __str__(self):
+        tag_repr = ','.join(repr(t) for t in self._tags)
+        return f"cirq.TaggedOperation({repr(self.sub_operation)}, {tag_repr})"
+
+    def __repr__(self):
+        return str(self)
+
+    def _value_equality_values_(self):
+        return (self.sub_operation, *self._tags)
```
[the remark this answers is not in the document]
dabacon: Why would it be less efficient than just the sub_operation plus tags?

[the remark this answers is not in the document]
cduck: I was thinking to avoid creating a new tuple of length `1+len(tags)` by nesting tuples instead of flattening.  Now that I think about it, there won't usually be a large number of tags so it really doesn't matter.
```
# in __init__
self._tags = tuple(...)

    return (self.sub_operation, self._tags)
```

dstrain115:
> We could consider switching the convention to "basic operations preserve tags" but I don't consider it blocking.

I felt that was tricky because it raises the question of what a basic operation is, and when does or doesn't this apply.  Making it so that any new creation deletes tags is sometimes annoying, but never ambiguous.

cirq/ops/raw_types.py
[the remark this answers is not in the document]
dstrain115: Ah, I was also confused about what you were concerned about.  I have changed it to a tuple, though I doubt the performance will matter too much.

cirq/ops/raw_types.py
```diff
@@ -445,6 +464,143 @@ def validate_args(self, qubits: Sequence['cirq.Qid']):
...
+@value.value_equality
+class TaggedOperation(Operation):
+    """A specific operation instance that has been identified with a set
+    of Tags for special processing.  This can be initialized with
+    Using Operation.with_tags(tag) or by TaggedOperation(op, tag).
+
+    See Operation.with_tags() for more information on intended usage.
+    """
+
+    def __init__(self, sub_operation: 'cirq.Operation', *tags: Any):
```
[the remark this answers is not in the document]
dstrain115: Changed to Hashable and put a warning in the pydoc to make tags JSON serialiable if you want to serialize the circuit.  I don't think JSON serializability should be a strict requirement, but it's nice to remember.

cirq/ops/raw_types.py
```diff
@@ -445,6 +464,143 @@ def validate_args(self, qubits: Sequence['cirq.Qid']):
...
+        self.sub_operation = sub_operation
+        self._tags = list(tags)
```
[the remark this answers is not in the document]

cirq/ops/raw_types.py
```diff
@@ -445,6 +466,149 @@ def validate_args(self, qubits: Sequence['cirq.Qid']):
...
+    def controlled_by(self,
+                      *control_qubits: 'cirq.Qid',
+                      control_values: Optional[Sequence[
+                          Union[int, Collection[int]]]] = None
+                     ) -> 'cirq.Operation':
+        return self.sub_operation.controlled_by(*control_qubits,
+                                                control_values=control_values)
+
+    @property
+    def tags(self):
```
cduck: Return type?

cirq/ops/raw_types.py
```diff
@@ -445,6 +466,149 @@ def validate_args(self, qubits: Sequence['cirq.Qid']):
...
+        """Returns a tuple of the operation's tags."""
+        return self._tags
+
+    def with_tags(self, *new_tags: Any) -> 'cirq.TaggedOperation':
```
cduck: Change to `Hashable` here as well.

cirq/ops/raw_types.py
dstrain115: Added.

cirq/ops/raw_types.py

PR #2672, opened 2020-01-07
Change symbols in phased_iswap_test

dstrain115:
- In sympy 1.2, print(1 - sympy.Symbol('t')) outputs "-t + 1"
- While the requirements do require sympy 1.4, the actual symbols used
in this test are not important to the testing of phased iswap, and this
change will simplify people who are forced to use older versions of
sympy (i.e. me).


dstrain115:
> Change `sympy.Symbol('t') + 1)` to `1 + sympy.Symbol('t'))`? The second change is more "minor"

I'd have to go back and check, but I don't think that has a consistent printing in sympy either.  I think that sympy 1.2 prints it as "1+t" but sympy 1.4 prints it as "t+1"

vtomole:
```python
# 1.2
>>> print(1 + sympy.Symbol('t'))
t + 1
>>> print(sympy.Symbol('t') + 1)
t + 1
```

```python
# 1 .4
>>> import sympy
>>> print(1 + sympy.Symbol('t'))
t + 1
>>> print(sympy.Symbol('t') + 1)
t + 1

```

PR #2674, opened 2020-01-10
Copied over BUILD from cirq.api.google.

MichaelBroughton:
Since `cirq.api.google` is on its way out, I copied over the BUILD file from there to `cirq.google.api`. This is a needed change for other projects.


Strilanc:
Obsoleted by @dstrain115

PR #2684, opened 2020-01-17
Add a bare-bones utility for generating noise models from calibration data.

95-martin-orion:
Given an Engine instance and a target processor, `simple_noise_from_calibration_metrics` will generate a depolarizing noise model with per-qubit depolarization probabilities pulled directly from the most recent calibration data.

Known issues/possible fixes:
- Assumes the most recent calibration data is valid and relevant to users.
    - Allow users to specify a calibration run? (i.e. `get_calibration`)
- Requires connection to the engine.
    - Provide static sample calibration data as an alternative?
- Does not incorporate other error channels (dephasing, amplitude damping)
    - Update as relevant metrics become available.


cirq/contrib/noise_models/noise_models.py
```diff
@@ -142,3 +144,51 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
+            return [
+                moment,
+                ops.Moment(
+                    ops.DepolarizingChannel(self.depol_prob_map[q])(q)
+                    for q in system_qubits)
+            ]
+
+
+def simple_noise_from_calibration_metrics(eng: engine.Engine, processor_id: str
+                                         ) -> devices.NoiseModel:
```
dstrain115: I think this should take a calibration object instead of an engine object.  Note that a few of the problems you specified in the description would be solved by specifying a calibration object instead.  I think doing this online only saves one line of code and is not necessarily what you want.  Like, I don't know that you want your noise model to suddenly shift if we upload new calibrations.  Also, you wouldn't be able to easily go back and run a consistent noise model (i.e. I want to run the same thing as yesterday).

cirq/contrib/noise_models/noise_models.py
```diff
@@ -142,3 +144,51 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
+
+    def noisy_moment(self, moment: 'cirq.Moment',
+                     system_qubits: Sequence['cirq.Qid']):
+        if _homogeneous_moment_is_measurements(moment):
+            return moment
+        else:
+            return [
+                moment,
+                ops.Moment(
+                    ops.DepolarizingChannel(self.depol_prob_map[q])(q)
```
dstrain115: Since you are modeling the error based on 1-q error rates, I think you would want a depolarizing channel only for qubits in the moment, (exclude idle qubits?)

95-martin-orion: Done. Left idle qubits unchanged for now - should they have an AmplitudeDamping channel based on the calibration T1 values?

cirq/contrib/noise_models/noise_models.py
95-martin-orion: Makes sense - done. The test still uses the 'retrieve Calibration from (mock) engine' step, partly because it required the least change and partly because it's a decent example for users.

95-martin-orion:
Windows build is failing, doesn't seem related to this PR:
```
The command "python -m pip install -r dev_tools/conf/pip-list-dev-tools.txt" failed and exited with 1 during .
Your build has been stopped.
ERROR: No matching distribution found for mkl (from cvxopt->qiskit-aqua==0.6.1->qiskit~=0.13.0->-r dev_tools/conf/pip-list-dev-tools.txt (line 18))
```

cirq/contrib/noise_models/noise_models.py
```diff
@@ -142,3 +144,56 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
+
+    def noisy_moment(self, moment: 'cirq.Moment',
+                     system_qubits: Sequence['cirq.Qid']):
+        if _homogeneous_moment_is_measurements(moment):
+            return moment
+        else:
+            gated_qubits = [
+                q for q in system_qubits if moment.operates_on_single_qubit(q)
+            ]
+            print(gated_qubits)
```
mpharrigan: remove this before merging

cirq/contrib/noise_models/noise_models.py
mpharrigan: In practice, we generally treat the "idle gate" has having roughly the same fidelity as other single-qubit operations

95-martin-orion: Got it. I'll leave more complex noise models for a follow-up PR.

cirq/contrib/noise_models/noise_models_test.py
```diff
@@ -157,3 +160,105 @@ def test_decay_noise_after_moment():
...
+def test_per_qubit_noise_from_data(build):
+    service = mock.Mock()
+    build.return_value = service
+    calibrations = service.projects().processors().calibrations()
+    calibrations.list().execute.return_value = ({
+        'calibrations': [_CALIBRATION]
+    })
+    eng = cirq.google.Engine(project_id='myproject')
+    calibration = eng.get_latest_calibration('fake_processor_name')
+    noise_model = simple_noise_from_calibration_metrics(calibration)
```
Strilanc: Why are you testing the calibration-getting path as part of testing your method? They seem like entirely separate concerns to me.

If this is the simplest way to create an engine.Calibration instance, we should fix that.

cirq/contrib/noise_models/noise_models_test.py
```diff
@@ -157,3 +160,105 @@ def test_decay_noise_after_moment():
...
+    program.append([
+        cirq.measure(qubits[0], key='q0'),
+        cirq.measure(qubits[1], key='q1'),
+        cirq.measure(qubits[2], key='q2')
+    ],
+                   strategy=cirq.InsertStrategy.NEW_THEN_INLINE)
+    noisy_circuit = cirq.Circuit(noise_model.noisy_moments(program, qubits))
+
+    # Insert channels explicitly to construct expected output.
+    true_noisy_program = cirq.Circuit()
```
Strilanc: Rename to `expected`

Note the expression

```
cirq.Circuit(
    cirq.Moment(ops1),
    cirq.Moment(ops2),
    cirq.Moment(ops3))
```

would be a much more compact way to generate the circuit.

cirq/contrib/noise_models/noise_models_test.py
95-martin-orion: Good point. I considered leaving this as an example of how to get calibration data from the engine, but the docstring in the `simple_noise_from_calibration_metrics` function should be sufficient. (Plus, it ended up being 90% mock code and 10% useful examples).

Switched to direct initialization.

cirq/contrib/noise_models/noise_models_test.py
95-martin-orion: Done and done.

95-martin-orion:
This change is included as part of #2734.

Issue #2692, opened 2020-01-21
Windows check failing


vtomole:
`ERROR: Could not find a version that satisfies the requirement mkl (from cvxopt->qiskit-aqua==0.6.1->qiskit~=0.13.0->-r dev_tools/conf/pip-list-dev-tools.txt (line 18)) (from versions: none)`

PR #2693, opened 2020-01-21
Remove pip upgrade on Windows check in Travis

vtomole:
The latest release of pip (https://pypi.org/project/pip/20.0.1/) breaks the installation of our qiskit dependency. This commit pins our Windows pip version until https://github.com/pypa/pip/issues/7626 is fixed.

Fixes: https://github.com/quantumlib/Cirq/issues/2692

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ matrix:
     before_install:
       - choco install python --version 3.6.3
       - export PATH="/c/Python36:/c/Python36/Scripts:$PATH"
-      - python -m pip install --upgrade pip wheel
+      - python -m pip install --upgrade wheel
     install:
       - python -m pip --version
       - python -m pip install -v mkl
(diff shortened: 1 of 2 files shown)


dstrain115:
Can you explain why this fixes the issue (and add it to the PR description)?

It's not really obvious why not upgrading pip fixes the issue.

vtomole:
@dstrain115 Added.

vtomole:
@dstrain115 Yes we should. I keep forgetting this stuff :(.  Added https://github.com/quantumlib/Cirq/issues/2695

.travis.yml
```diff
@@ -51,7 +51,7 @@ matrix:
     before_install:
       - choco install python --version 3.7.3
       - export PATH="/c/Python37:/c/Python37/Scripts:$PATH"
-      - python -m pip install --upgrade pip wheel
+      - python -m pip install --upgrade wheel
```
Strilanc: Should we add a TODO to restore?

PR #2697, opened 2020-01-22
Plumb gate tabulation through optimized_for_sycamore

mpharrigan:
- Convert list of optimizers to functions that return configured
   optimizers
 - plumb through `tolerance` argument
 - plumb through `tabulation` for SYC gate decomposition
   (this should also work for sqrt_iswap but I haven't tested/enabled it)
 - use lru_cache to avoid re-tabulation


cirq/google/optimizers/optimize_for_sycamore.py
```diff
@@ -12,73 +12,113 @@
...
-    'sycamore': _SYCAMORE_OPTIMIZERS,
+    'xmon': _get_xmon_optimizers,
+    'xmon_partial_cz': _get_xmon_optimizers_part_cz,
+    'sqrt_iswap': _get_sqrt_iswap_optimizers,
+    'sycamore': _get_sycamore_optimizers,
 }
 
 
+@lru_cache()
+def _gate_product_tabulation_cached(optimizer_type, tabulation_resolution):
```
Strilanc: type annotations

cirq/google/optimizers/optimize_for_sycamore.py
```diff
@@ -12,73 +12,113 @@
...
+
 def optimized_for_sycamore(
         circuit: 'cirq.Circuit',
         new_device: Optional['cirq.google.XmonDevice'] = None,
         qubit_map: Callable[['cirq.Qid'], devices.GridQubit] = lambda e: cast(
             devices.GridQubit, e),
-        optimizer_type: str = 'sqrt_iswap') -> 'cirq.Circuit':
+        optimizer_type: str = 'sqrt_iswap',
+        tolerance: float = 1e-5,
+        tabulation_resolution: Optional[float] = None,
```
Strilanc: We should force keywords starting from `new_device`.

cirq/google/optimizers/optimize_for_sycamore.py
```diff
@@ -12,73 +12,113 @@
...
+        -> List[Callable[['cirq.Circuit'], None]]:
+    if tabulation is not None:
+        raise ValueError("Gate tabulation not supported for sqrt_iswap")
+    return [
+        ConvertToSqrtIswapGates().optimize_circuit,
+        lambda c: optimizers.merge_single_qubit_gates_into_phased_x_z(
+            c, tolerance),
+        optimizers.EjectPhasedPaulis(tolerance=tolerance).optimize_circuit,
+        optimizers.EjectZ(tolerance=tolerance).optimize_circuit,
+        optimizers.DropNegligible(tolerance=tolerance).optimize_circuit,
```
dstrain115: I wonder if we should combine these last 3 optimizers into a common function since all 4 classes use them.

mpharrigan: done

cirq/google/optimizers/optimize_for_sycamore.py
mpharrigan: done

cirq/google/optimizers/optimize_for_sycamore.py
mpharrigan: done

PR #2699, opened 2020-01-23
Make n-qubit tomography code available under cirq.experiments

viathor:
Also updates tests to access the symbols under cirq instead of cirq.experiments.n_qubit_tomography. This forces n_qubit_tomography to import individual modules (which is also consistent with how we do this elsewhere).

### DIFF
--- a/cirq/experiments/n_qubit_tomography.py
+++ b/cirq/experiments/n_qubit_tomography.py
@@ -17,13 +17,16 @@ different pre-measurement rotations.
 The code is designed to be modular with regards to data collection
 so that occurs outside of the StateTomographyExperiment class.
 """
-from typing import List, Optional, Sequence, Tuple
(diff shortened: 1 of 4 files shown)


cirq/experiments/n_qubit_tomography.py
```diff
@@ -45,7 +45,7 @@ class StateTomographyExperiment:
     """
 
     def __init__(self,
-                 qubits: Sequence[cirq.Qid],
+                 qubits: Sequence[ops.Qid],
```
cduck: I think the current consensus for annotations is to add
```python
if TYPE_CHECKING:
    import cirq
```
and use `'cirq.Qid'` so the documentation is consistent.

viathor: Good point. Done.

PR #2706, opened 2020-01-23
Add json_serializable_dataclass decorator

mpharrigan:
The structured nature of `@dataclass`es plays very nicely with JSON serialization. This decorator automatically adds the requisite `_json_dict_` method. You still have to register with the "resolver" for loading.


mrwojtek:
Do you plan on auto-generating resolver? This is also a natural for this functionality and will most likely be very similar to all the usages.

Did you encounter a lot of cases where you wanted to have a specialized serialization for different dataclass fields within the class, without using wrappers? I guess in that case the correct way of doing things is to not use @json_serializable_dataclass but implement the serialization protocol directly? (I'm thinking to also extend the dataclasses.field with support for serialization.)

cirq/__init__.py
```diff
@@ -396,6 +396,7 @@
     decompose,
     decompose_once,
     decompose_once_with_qubits,
+    DEFAULT_RESOLVERS,
```
mrwojtek: This change also exposes resolvers to the public which was not done before?

cirq/protocols/json_serialization.py
```diff
@@ -232,6 +233,59 @@ class name via a dot (.)
...
+
+    This function defers to the ordinary ``dataclass`` decorator but appends
+    the ``_json_dict_`` protocol method which automatically determines
+    the appropriate fields from the dataclass.
+
+    Args:
+        namespace: An optional prefix to the value associated with the
+            key "cirq_type". The namespace name will be joined with the
+            class name via a dot (.)
+        init, repr, eq, order, unsafe_hash: Forwarded to the ``dataclass``
```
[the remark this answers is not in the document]
mpharrigan: done!

cirq/__init__.py
mpharrigan: Yes, I did this since I needed this when writing a test and in general people can inject custom resolvers by doing

```python
cirq.read_json(fn, resolvers=[my_resolver] + cirq.DEFAULT_RESOLVERS)
```

I can put this back if others don't like it at the top level

mpharrigan:
> Do you plan on auto-generating resolver?

No, we consciously have decided to not have auto-registration of deserialization routines. It's too much magic and can be a threat vector

PR #2708, opened 2020-01-25
Fix special new line characters in a file

### LINKED ISSUE #2036: [Style] cirq/protocols/equal_up_to_global_phase.py has carriage returns
This command finds all files with carriage returns:
```
find . -name '*.py' | xargs grep -l --binary -P '\x0d'
```

tonybruguier-google:
It seems that this file had weird unicode characters for new lines. Maybe it would make sense to fix them?

Fixes #2036.


tonybruguier-google:
Actually, I looked more and the character is ASCII 0x0A a.k.a. LN a.k.a. line-feed.

mpharrigan:
My IDE does indeed report that the file uses CRLF (windows-style) line endings. The convention is LF (unix-style) line endings

vtomole:
@cduck Should we add that command as part of `format-incremental`?

cduck:
@vtomole That might be a good idea although this issue doesn't occur often.

Issue #2639, opened 2019-12-12
Direct fidelity estimation should not use O(4^n) estimations when the circuit is Clifford


tonybruguier-google:
During the PR review:
https://github.com/quantumlib/Cirq/pull/2523#discussion_r356816645

A TODO was added to sample the states efficiently when the circuit is Clifford. Address the TODO.

tonybruguier-google:
Sent PR #2709

PR #2709, opened 2020-01-26
Speed up direct fidelity estimation for Clifford circuits (Issue #2639)

tonybruguier-google:
During the review of PR #2523 I promised to implement the speedup for Clifford circuits: Issue #2639 

This PR is an attempt at doing just that


examples/direct_fidelity_estimation.py
```diff
@@ -30,16 +30,29 @@
 
 
 def build_circuit():
```
vtomole: Add return type annotation to `build_circuit()`

examples/direct_fidelity_estimation.py
```diff
@@ -30,16 +30,29 @@
...
+        cirq.X(qubits[2]),
+    ]
+    for gate in gates:
+        circuit.append(gate)
+        if state is not None:
+            try:
+                state.apply_unitary(gate)
+            except ValueError:
+                print('Circuit is non Clifford.')
+                state = None
```
vtomole: `state` being `None` to me seems like a `null` state? Doesn't make sense. Let's change the variable name of `state` to something more fitting. `clifford_state` maybe?

examples/direct_fidelity_estimation.py

examples/direct_fidelity_estimation.py

tonybruguier-google:
Thanks. Sorry about the pollution of the history chain. I think it looks cleaner now.

examples/direct_fidelity_estimation.py
```diff
@@ -29,27 +30,44 @@
...
-    circuit = cirq.Circuit(
-        cirq.Z(qubits[0])**0.25,  # T-Gate, non Clifford.
-        cirq.X(qubits[1])**0.123,
-        cirq.X(qubits[2])**0.456)
+def build_circuit(
+) -> Tuple[cirq.Circuit, List[cirq.Qid], Optional[cirq.CliffordState]]:
+    # Builds an arbitrary circuit to test. If the circuit is Clifford, then it
+    # also returns a state from which we can get the stabilitzers. If the
+    # circuit is non Clifford, the state is None.
+    qubits: List[cirq.Qid] = [qubit for qubit in cirq.LineQubit.range(3)]
```
mpharrigan: `qubits = cirq.LineQubit.range(3)`

examples/direct_fidelity_estimation.py
```diff
@@ -29,27 +30,44 @@
...
+        qubit_map={qubits[i]: i for i in range(len(qubits))})
+    is_clifford = True
+    gates = [
+        cirq.Z(qubits[0]),
+        cirq.X(qubits[1]),
+        cirq.X(qubits[2]),
+    ]
+    for gate in gates:
+        circuit.append(gate)
+        if is_clifford:
```
mpharrigan: I'm confused by the logic here. You're explicitly constructing a clifford circuit so this will always be `True`

examples/direct_fidelity_estimation.py
```diff
@@ -29,27 +30,44 @@
...
+            except ValueError:
+                print('Circuit is non Clifford.')
+                is_clifford = False
     print('Circuit used:')
     print(circuit)
-    return circuit, qubits
+    if is_clifford:
+        return circuit, qubits, clifford_state
+    else:
+        return circuit, qubits, None
```
mpharrigan: likewise, different return types is a code smell. I would suggest always returning a clifford state. Maybe you can have another function that returns a non-clifford circuit and a flag somewhere else to switch between them. This would probably require factoring out the "check if circuit is clifford" logic elsewhere

examples/direct_fidelity_estimation.py
```diff
@@ -133,19 +151,24 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
...
         simulator.simulate(circuit)).final_density_matrix
 
-    # TODO(#2639): Sample the Pauli states more efficiently when the circuit
-    # consists of Clifford gates only, as described on page 4 of:
+    # When the circuit consists of Clifford gates only, we can sample the Pauli
+    # states more efficiently as described on page 4 of:
     # https://arxiv.org/abs/1104.4695
-    for P_i in itertools.product([cirq.I, cirq.X, cirq.Y, cirq.Z],
-                                 repeat=n_qubits):
+    if clifford_state is not None:
```
mpharrigan: instead of taking in `clifford_state`, consider trying to evaluate `circuit` as a clifford circuit and include fall-back logic to the exhaustive method

examples/direct_fidelity_estimation.py
tonybruguier-google: That was stemming from my buggy code that added a measurement gate, which in turn made iteration through the circuit operation not translatable to a Clifford state because the last gate isn't Clifford.

Actually, this was a bug from previous PRs (sorry), which made the fidelity values not correct, I think. As a check, I reran the job without any Clifford acceleration and without noise. I think under these settings the fidelity should be exactly 1.0. It was, however, exactly 0.125, so I added fixed this by adding the missing coefficient. 

Thanks for the carefully review and catching this.

examples/direct_fidelity_estimation.py

examples/direct_fidelity_estimation.py
tonybruguier-google: No longer present. Done.

tonybruguier-google:
Thanks for the careful review and your help by email.

I think this makes more sense to me now. Thank you.

PTAL, no hurry.

tonybruguier-google:
Hello,

It's not super urgent, of course, but do you have any more comments? It seems that the Travis error above are due to a flaky test, which I am attempting to fix in another PR:
https://github.com/quantumlib/Cirq/pull/2758

No hurry and thanks again for having helped me along the way.

examples/direct_fidelity_estimation.py
```diff
@@ -121,65 +121,122 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
...
+            # compute_characteristic_function() on the results (albeit with a
+            # wave function instead of a density matrix). It is, however,
+            # unncessary to do so.
+            rho_i = dense_pauli_string.coefficient
+
+            assert np.isclose(rho_i.imag, 0.0, atol=1e-6)
+            rho_i = rho_i.real
+
+            dense_pauli_string *= rho_i
+            Pr_i = rho_i * rho_i / d
```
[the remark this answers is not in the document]
tonybruguier-google: Indeed. Added an assert that rho_i = +/- 1 and then just have Pr_i = 1/d.

examples/direct_fidelity_estimation.py
```diff
@@ -121,65 +121,122 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
...
+    clifford_circuit = True
+    try:
+        clifford_state = cirq.CliffordState(
+            qubit_map={qubits[i]: i for i in range(len(qubits))})
+        for gate in circuit.all_operations():
+            clifford_state.apply_unitary(gate)
+    except ValueError:
+        clifford_circuit = False
+
+    if clifford_circuit:
```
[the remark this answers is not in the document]

examples/direct_fidelity_estimation.py
```diff
@@ -121,65 +121,122 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
...
+        for _ in range(n_clifford_trials):
+            # Build the Pauli string as a random sample of the basis elements.
+            dense_pauli_string = cirq.DensePauliString.eye(n_qubits)
+            for stabilizer in stabilizer_basis:
+                if np.random.randint(2) == 1:
+                    dense_pauli_string *= stabilizer
+
+            # The code below is equivalent to calling
+            # clifford_state.wave_function() and then calling
+            # compute_characteristic_function() on the results (albeit with a
```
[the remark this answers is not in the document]
tonybruguier-google: Added test_same_pauli_traces_clifford()

Also, as per your other comment, I hard code 1/d and so I added an assert rho_i = +/- 1

examples/direct_fidelity_estimation.py
```diff
@@ -121,65 +121,122 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
...
+        # https://arxiv.org/abs/1104.4695
+
+        # The stabilizers_basis variable only contains basis vectors. For
+        # example, if we have n=3 qubits, then we should have 2**n=8 Pauli
+        # states that we can sample, but the basis will still have 3 entries. We
+        # must flip a coin for each, whether or not to include them.
+        stabilizer_basis = clifford_state.stabilizers()
+
+        # TODO(tonybruguier): Find a better way to decide how many samples.
+        n_clifford_trials = n_qubits
```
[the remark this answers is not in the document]

examples/direct_fidelity_estimation.py
```diff
@@ -121,65 +121,122 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
...
-    for P_i in itertools.product([cirq.I, cirq.X, cirq.Y, cirq.Z],
-                                 repeat=n_qubits):
-        rho_i, Pr_i = compute_characteristic_function(circuit, P_i, qubits,
-                                                      clean_density_matrix)
-        pauli_traces.append({'P_i': P_i, 'rho_i': rho_i, 'Pr_i': Pr_i})
-
-    assert len(pauli_traces) == 4**n_qubits
+    clifford_circuit = True
+    try:
+        clifford_state = cirq.CliffordState(
```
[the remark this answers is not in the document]
tonybruguier-google: Not quite sure how to test. Set it to None before this block. Is this what you had in mind?

examples/direct_fidelity_estimation.py
```diff
@@ -121,65 +121,122 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
...
+            for stabilizer in stabilizer_basis:
+                if np.random.randint(2) == 1:
+                    dense_pauli_string *= stabilizer
+
+            # The code below is equivalent to calling
+            # clifford_state.wave_function() and then calling
+            # compute_characteristic_function() on the results (albeit with a
+            # wave function instead of a density matrix). It is, however,
+            # unncessary to do so.
+            rho_i = dense_pauli_string.coefficient
```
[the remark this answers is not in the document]
tonybruguier-google: Added a comment that says it's a scalar.

tonybruguier-google:
Thanks. PTAL.

examples/direct_fidelity_estimation.py
```diff
@@ -95,9 +95,74 @@ def compute_characteristic_function(circuit: cirq.Circuit,
     return sigma_i
 
 
+def _estimate_pauli_traces_clifford(n_qubits, clifford_state,
```
mpharrigan: type annotations?

examples/direct_fidelity_estimation.py
```diff
@@ -248,6 +263,11 @@ def parse_arguments(args):
                         type=int,
                         help='Number of trials to run.')
 
+    parser.add_argument('--n_clifford_trials',
+                        default=3,
+                        type=int,
+                        help='Number of trials for Clifford circuits')
```
mpharrigan: Can you provide a more descriptive help message for this? Mention tradeoffs between setting this too large / too small

examples/direct_fidelity_estimation.py
mpharrigan: I know this is a private function, but this is an example so a little docstring would be helpful.

examples/direct_fidelity_estimation.py
```diff
@@ -109,6 +174,8 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
         qubits: The list of qubits.
         noise: The noise model when doing a simulation.
         n_trial: The total number of Pauli measurements.
+        n_clifford_trials: In case the circuit is Clifford, we specify the
+            number of trials to estimate the noise-free pauli traces.
```
mpharrigan: offer a little more guidance on how to pick this? 

Also open up a follow-up issue for the following enhancement: DFE shows that you can do monte-carlo (random) sampling of the pauli stabilizers for a clifford circuit allowing you to estimate the fidelity of an n-qubit clifford state in O(1) time instead of doing it "exhaustively" which is O(2^n) [IIRC]. However, these asymptotic arguments really only come into play for "large" n because you still have to monte-carlo sample a lot of terms [for some definition of "a lot"]. If you need to do 500 trials, then log2(500) ~ 9 so for n < 9, you might as well do the exhaustive approach. So as a follow up, there are a couple of avenues to pursue

 - add an option for "exhaustive" sampling of the terms, which can be faster at low `n`.
 - do a numerical study to figure out what a reasonable value for `n_clifford_trials` is (which gives good convergence but isn't too large, which would mean a long runtime on the quantum processor)
 - [after those two] tell the user when they would be better off switching from exhaustive to monte-carlo

examples/direct_fidelity_estimation.py
tonybruguier-google: Done and done.

I also took the liberty of doing the same for the other private function _estimate_pauli_traces_general().

examples/direct_fidelity_estimation.py
tonybruguier-google: Added some comments.

examples/direct_fidelity_estimation.py
tonybruguier-google: Opened issue #2802 and added a TODO (but not right at this line of the code, but just above the definition of the flag).

tonybruguier-google:
Thanks. PTAL.

PR #2712, opened 2020-01-27
JSON wrappers for numpy arrays

mpharrigan:
When saving data, you can optionally wrap your big arrays in these simple wrapper classes which will cause the data to be serialized to a nicer json representation; namely a hex-encoded string of the binary representation. This also includes a specialization for arrays of bits (1's and 0's) which first packs them into uint8's


kevinsung:
I'd like to defer to @Strilanc here because the idea of a JSON "wrapper" is a new thing to Cirq; in all other uses of JSON serialization so far, we have been defining the protocol directly for the object itself.

mpharrigan:
@Strilanc thoughts?

kevinsung:
Now that I think of it, Numpy arrays are the only JSON serializable type (as far as I know) for which
```
cirq.read_json(json_text=cirq.to_json(x))
```
does not return something equivalent to the object `x`. I think this fact is somewhat undesirable and we should consider enforcing that the line above should always return something equivalent to `x`.

mpharrigan:
xref #2742; it's not the only type like that

kevinsung:
I meant, the only type that we explicitly define a serialization protocol for in Cirq.

mpharrigan:
This affords the user some control over which format is best for their use case

 - A smallish list of numbers that happens to be in a numpy array probably makes sense to serialize as a list for interop and/or readability
- A beefy array can use the efficient but general numpy packing
- An array of bits can save a factor of 8 by using the one optimized for bits

Also look forward to more of these from me, e.g. making `nx.Graph` serializable :) 

I don't know if this is the right way of going about it but it certainly works for my use cases :)

kevinsung:
I'm with @Strilanc. It's nicer to treat all Numpy arrays the same way and to have the protocol be reversible. Also, we can automatically detect whether a Numpy array contains bits and apply the more efficient bit-packing method in that case. This is what the JSON serialization for TrialResult does.

mpharrigan:
I'm going to close for now

PR #2715, opened 2020-01-28
Support readout errors modeling in noise-from-metrics utility.

95-martin-orion:
Depends on #2684. This extends the "PerQubitDepolarizationNoiseModel" from that PR to also support amplitude damping during readout and other readout error.


95-martin-orion:
+ @dkafri to review, as this is partially based off of his T1T2Decay channel. Due to nontrivial differences in implementation requirements (including metrics available in Cirq), I opted not to import that channel directly.

cirq/contrib/noise_models/noise_models.py
```diff
@@ -142,3 +145,121 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
                 ops.Moment(self.qubit_noise_gate(q) for q in system_qubits)
             ]
+
+
+class PerQubitDepolarizingWithDampedReadoutNoiseModel(devices.NoiseModel):
+    """DepolarizingWithDampedReadoutNoiseModel which allows probabilities to be
+    specified separately for each qubit.
+
+    This simulates asymmetric readout error. The noise is structured
+    so the T1 decay is applied, then the readout bitflip, then measurement.
```
dkafri: Looking at the code, it looks like the actual order of operations is bitflip -> T1 decay -> measurement

cirq/contrib/noise_models/noise_models.py
```diff
@@ -142,3 +145,121 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
+    Note that T1 decay is only applied to measurement, not other gates.
+
```
dkafri: How does the noise model act on moments without measurement?

cirq/contrib/noise_models/noise_models.py
```diff
@@ -142,3 +145,121 @@ def noisy_moment(self, moment: 'cirq.Moment',
                 moment,
                 ops.Moment(self.qubit_noise_gate(q) for q in system_qubits)
             ]
+
+
+class PerQubitDepolarizingWithDampedReadoutNoiseModel(devices.NoiseModel):
```
dkafri: One comment on the direction of this approach. I think it would be better to make different noise models composable, instead of making a different noise model for each combination of error A, B and C. In other words, it would probably be better to use DepolarisingNoiseModel on your circuit, then DampedReadoutNoiseModel on the resulting moments.

cirq/contrib/noise_models/noise_models.py
```diff
@@ -142,3 +145,121 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
+                                          depolNoise: bool = False,
+                                          dampingNoise: bool = False,
+                                          readoutDecayNoise: bool = False,
+                                          readoutErrorNoise: bool = False
+                                         ) -> devices.NoiseModel:
+    """Creates a reasonable PerQubitDepolarizingNoiseModel using the provided
+    calibration data. This object can be retrived from the engine by calling
+    'get_latest_calibration()' or 'get_calibration()' using the ID of the
+    target processor.
+    """
```
dkafri: How about adding Args and Returns sections to this docstring? I don't know what an engine.Calibration is. Looking at the code, it seems like you're making some implicit assumptions about the structure of `calibration` which should be made clear.

cirq/contrib/noise_models/noise_models.py
95-martin-orion: My concern with this is (AFAICT) composing noise models can result in duplicated noise channels. This is illustrated in the example below:
```
create circuit C with gates {X, H, M}
create a noise model K which adds k_err before M gates, apply K to C
-> C is now {X, H, k_err, M}
create a noise model D which adds d_err before non-M gates, apply D to C
-> C is now {X, d_err, H, d_err, k_err, d_err, M}
```
where the final `d_err` is not desired, but will be added since we have no clear way to differentiate between "gates representing noise" and "gates representing real operations." (We might have this behavior and I've simply missed it - please point it out if this is the case.)

cirq/contrib/noise_models/noise_models.py
95-martin-orion: Non-measurement gates will have depolarizing noise applied; idle qubits will remain unaffected. Updated docstring.

cirq/contrib/noise_models/noise_models.py
95-martin-orion: Whoops. Fixed, and switched to `append` to avoid confusing myself.

cirq/contrib/noise_models/noise_models.py
95-martin-orion: It looks like this is relevant to issue #1835 - @Strilanc @mpharrigan, do we expect any issues with the method above? If not, let's apply that fix and I'll come back to this PR afterward.

mpharrigan: Yes, that's the relevant issue. If someone wants to tackle it; that would be great. One of the reasons why these noise models are in `contrib` instead of cirq proper is because we'd ideally like to solve the problem of composable noise models and just use that

95-martin-orion: From group discussions it sounds like this is a good use case for TaggedOperations (see #1940, #2670). Opened #2719 to track this change - I'll revisit this PR once that issue is resolved.

95-martin-orion: On further inspection, I've determined that noise models will need significant work before supporting composition in a consistent and intuitive way. In order to avoid delaying the noise-from-metrics feature any further, I'm moving forward with this PR.

After the noise-model work is done, we can split this monolithic model up into its component pieces and compose them as necessary.

cirq/contrib/noise_models/noise_models.py
```diff
@@ -173,16 +174,136 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
+            moments.append(moment)
+            return moments
+        else:
+            moments.append(moment)
+            if self.depol_prob_map:
+                gated_qubits = [
+                    q for q in system_qubits
+                    if moment.operates_on_single_qubit(q)
+                ]
+                moments.append(
```
dstrain115: Should this have an` if gated_qubits:` to prevent adding empty moments?  (not sure if that case is possible?)

cirq/contrib/noise_models/noise_models.py
```diff
@@ -173,16 +174,136 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
+                               (decay_prob_map, "readout decay prob")]:
+            if prob_map:
+                for qubit, prob in prob_map.items():
+                    value.validate_probability(prob, f'{desc} of {qubit}')
+        self.depol_prob_map = depol_prob_map
+        self.bitflip_prob_map = bitflip_prob_map
+        self.decay_prob_map = decay_prob_map
+
+    def noisy_moment(self, moment: 'cirq.Moment',
+                     system_qubits: Sequence['cirq.Qid']):
```
dstrain115: Optional: add a short docstring?

cirq/contrib/noise_models/noise_models.py
```diff
@@ -173,16 +174,136 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
+    readout_decay_map: Dict['cirq.Qid', float] = {}
+    readout_error_map: Dict['cirq.Qid', float] = {}
+
+    if depolNoise:
+        depol_prob_map = {
+            qubit[0]: depol_prob[0] for qubit, depol_prob in
+            calibration['single_qubit_rb_total_error'].items()
+        }
+    if dampingNoise:
+        # TODO: implement per-gate amplitude damping noise.
```
dstrain115: Can this be linked to a GH issue?

cirq/contrib/noise_models/noise_models.py
```diff
@@ -173,16 +174,136 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
+    if readoutDecayNoise:
+        # Copied from Sycamore readout duration in known_devices.py
+        # TODO: replace with polling from DeviceSpecification.
+        readout_micros = 1
+        readout_decay_map = {
+            qubit[0]: exp(1 - readout_micros / t1[0])
+            for qubit, t1 in calibration['single_qubit_idle_t1_micros'].items()
+        }
+    if readoutErrorNoise:
+        # This assumes that p(<1|0>) is negligible for readout timescales.
```
dstrain115: Is this correct?  Isn't the probability that a state decays from |1> to |0> generally higher than measuring |1> if it was supposed to be |0>?

Shouldn't this use p1 error somehow?  Is the intention to use p1 in the amplitude damping condition?

95-martin-orion: The intent is to model decay from `|1>` to `|0>` - if I've got my notation correct, `p(<1|0>)` is the probability of going from `|0>` to `|1>`.

Is `single_qubit_readout_p0_error` the right metric for `|1>` to `|0>` decay? If we have a metric for the `|0>` to `|1>` error, I could add that as well - it would require replacing the AmplitudeDamping channel with a GAD channel in the model.

cirq/contrib/noise_models/noise_models.py
95-martin-orion: Done. Also updated the other PerQubit noise model with a similar fix.

cirq/contrib/noise_models/noise_models.py
95-martin-orion: Done.

cirq/contrib/noise_models/noise_models.py
95-martin-orion: There's a docstring for the base NoiseModel class - is there anything here that you think needs additional explanation beyond that?

cirq/contrib/noise_models/noise_models.py
dstrain115: The p0 (or p00) experiment is to prepare the state in |0> and try to measure |0>.  The error is the chance that this doesn't happen.
The p1 (or p11) experiment is to prepare the state in |1> and try to measure |1>.  The error is the chance that this doesn't happen.

So, I think you want to p1 error if you are trying to capture the probability that the state will decay from |1> to |0> either by incorrect measurement, incorrect state preparation, decay, or what-have-you.

cirq/contrib/noise_models/noise_models.py
dstrain115: That's why I left it as optional.  It's a method without a docstring but might be simple enough to leave off the docstring.  I'll leave it to your discretion whether to add any implementation details here.

95-martin-orion: SGTM. I think the docstring for the class gives pretty good coverage of this, so I'll leave as-is.

cirq/contrib/noise_models/noise_models.py
95-martin-orion: Whoops, I think I misread my own code. `readout_error_map` should capture error in the readout, not from qubit decay. Reading the doc in #2761, it seems like `single_qubit_readout_separation_error` is actually what we want here.

(T1 decay during readout is captured by the aptly-named `readout_decay_map`.)

Fix in progress.

95-martin-orion: Done. Also switched to `single_qubit_rb_pauli_error` for depolarization, since that appears to be more precise than using total error.

cirq/contrib/noise_models/noise_models.py
```diff
@@ -173,16 +176,138 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
-        calibration['single_qubit_rb_total_error'].items()
-    }
-    return PerQubitDepolarizingNoiseModel(rb_data)
+    depol_prob_map: Dict['cirq.Qid', float] = {}
+    readout_decay_map: Dict['cirq.Qid', float] = {}
+    readout_error_map: Dict['cirq.Qid', float] = {}
+
+    if depolNoise:
+        depol_prob_map = {
+            qubit[0]: depol_prob[0] for qubit, depol_prob in
```
95-martin-orion: I'm noticing now that this basically treats every non-measurement, non-virtual gate as equivalent to an RB gate.

@dstrain115, do you have a notion of how terrible this assumption is?

cirq/contrib/noise_models/noise_models.py
```diff
@@ -211,54 +212,140 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
-    specified separately for each qubit.
+class PerQubitDepolarizingWithDampedReadoutNoiseModel(devices.NoiseModel):
+    """DepolarizingWithDampedReadoutNoiseModel which allows depolarization
+    probabilities to be specified separately for each qubit.
 
-    Similar to depol_prob in DepolarizingNoiseModel, depol_prob_map should map
-    Qids in the device to their depolarization probability.
+    This simulates asymmetric readout error. The noise is structured
+    so the T1 decay is applied, then the readout bitflip, then measurement.
+    Note that T1 decay is only applied to measurement, not other gates.
```
dstrain115: I think I agree with this, since the error metrics should already include T1 for gate errors, but it might be worth putting in a sentence explaining why.

cirq/contrib/noise_models/noise_models.py
```diff
@@ -211,54 +212,140 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
+        self.decay_prob_map = decay_prob_map
 
     def noisy_moment(self, moment: 'cirq.Moment',
                      system_qubits: Sequence['cirq.Qid']):
-        if (_homogeneous_moment_is_measurements(moment) or
-                self.is_virtual_moment(moment)):
+        if self.is_virtual_moment(moment):
             return moment
+        moments = []
+        if _homogeneous_moment_is_measurements(moment):
```
dstrain115: Will this handle the case that the measurements are (purposefully or accidentally) mixed with other types of gates?

It seems like this needs to be handled by either putting amplitude damping in, or for throwing an exception if it is a case we do not support.

cirq/contrib/noise_models/noise_models.py
```diff
@@ -211,54 +212,140 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
-        qubit[0]: depol_prob[0] for qubit, depol_prob in
-        calibration['single_qubit_rb_total_error'].items()
-    }
-    return PerQubitDepolarizingNoiseModel(rb_data)
+    depol_prob_map: Dict['cirq.Qid', float] = {}
+    readout_decay_map: Dict['cirq.Qid', float] = {}
+    readout_error_map: Dict['cirq.Qid', float] = {}
+
+    if depolNoise:
+        # TODO: replace with Pauli error once it's available.
```
dstrain115: I would recommend connecting this to a GH issue #

cirq/contrib/noise_models/noise_models.py
```diff
@@ -211,54 +212,140 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
+            calibration['single_qubit_rb_total_error'].items()
+        }
+    if dampingNoise:
+        # TODO: implement per-gate amplitude damping noise.
+        # Github issue: https://github.com/quantumlib/Cirq/issues/2807
+        raise NotImplementedError('Gate damping is not yet supported.')
+
+    if readoutDecayNoise:
+        # Copied from Sycamore readout duration in known_devices.py
+        # TODO: replace with polling from DeviceSpecification.
```
dstrain115: Again, should we connect this to a GH issue for tracking?

cirq/contrib/noise_models/noise_models.py
95-martin-orion: The actual reasoning behind this decision is that applying accurate T1 decay for gates requires knowing their duration, which requires the DeviceSpecification (highlighted in the damping TODO).

cirq/contrib/noise_models/noise_models.py
95-martin-orion: Currently the `_homogeneous_moment_is_measurements` function raises an exception for moments containing both measurement and non-measurement gates. Once damping is supported for general gates, we can optionally apply it to measurement gates and remove this check.

cirq/contrib/noise_models/noise_models.py
```diff
@@ -211,54 +212,142 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
-    """DepolarizingNoiseModel which allows depolarization probabilities to be
-    specified separately for each qubit.
+class PerQubitDepolarizingWithDampedReadoutNoiseModel(devices.NoiseModel):
+    """DepolarizingWithDampedReadoutNoiseModel which allows depolarization
+    probabilities to be specified separately for each qubit.
 
-    Similar to depol_prob in DepolarizingNoiseModel, depol_prob_map should map
-    Qids in the device to their depolarization probability.
+    This simulates asymmetric readout error. The noise is structured
+    so the T1 decay is applied, then the readout bitflip, then measurement.
```
dabacon: Wording is slightly confusing.  Should be something more like
> T1 decay is applied after all gates that are not measurements, then for moments that are all measurements amplitude damping followed by a bit flip channel is applied before the measurement.  

I would also mention that this noise channel only works when all measurements are alone in moments.

cirq/contrib/noise_models/noise_models.py
```diff
@@ -211,54 +212,142 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
+    that remain idle will be unaffected by this model.
+
+    As with the DepolarizingWithDampedReadoutNoiseModel, if a circuit contains
+    measurements, they must be in moments that don't also contain gates.
     """
 
     def __init__(
             self,
-            depol_prob_map: Dict['cirq.Qid', float],
+            depol_prob_map: Dict['cirq.Qid', float] = None,
```
dabacon: nit: would call this `depol_probs` since type here really tells you that it is a map (also python people would say `_dict`).  similar for other variables

cirq/contrib/noise_models/noise_models.py
```diff
@@ -211,54 +212,142 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
 
         Args:
             depol_prob_map: Map of depolarizing probabilities for each qubit.
+            bitflip_prob: Probability of a bit-flip during measurement.
+            decay_prob: Probability of T1 decay during measurement.
+                Bitflip noise is applied first, then amplitude decay.
         """
-        for qubit, depol_prob in depol_prob_map.items():
-            value.validate_probability(depol_prob, f'depol prob of {qubit}')
+        for prob_map, desc in [(depol_prob_map, "depolarization prob"),
```
dabacon: `map` --> `dict` in names to be more in line with python nomenclature

cirq/contrib/noise_models/noise_models.py
```diff
@@ -211,54 +212,142 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
+                               (decay_prob_map, "readout decay prob")]:
+            if prob_map:
+                for qubit, prob in prob_map.items():
+                    value.validate_probability(prob, f'{desc} of {qubit}')
         self.depol_prob_map = depol_prob_map
+        self.bitflip_prob_map = bitflip_prob_map
+        self.decay_prob_map = decay_prob_map
 
     def noisy_moment(self, moment: 'cirq.Moment',
                      system_qubits: Sequence['cirq.Qid']):
```
dabacon: nit: return type

cirq/contrib/noise_models/noise_models.py
```diff
@@ -211,54 +212,142 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
+                ]
+                if gated_qubits:
+                    moments.append(
+                        ops.Moment(
+                            ops.DepolarizingChannel(self.depol_prob_map[q])(q)
+                            for q in gated_qubits))
+            return moments
+
+
+def simple_noise_from_calibration_metrics(calibration: engine.Calibration,
```
dabacon: can we put this in cirq.google?  Does this cause a problem making a dependency from engine to contrib?  If so we should probably create contrib.google

cirq/contrib/noise_models/noise_models.py
```diff
@@ -211,54 +212,142 @@ def noisy_moment(self, moment: 'cirq.Moment',
...
+                                          depolNoise: bool = False,
```
dabacon: style: `depol_noise` here and below

cirq/contrib/noise_models/noise_models_test.py
```diff
@@ -315,14 +316,109 @@ def test_aggregate_decay_noise_after_moment():
...
     }]
 """, v2.metrics_pb2.MetricsSnapshot())
 
 
-def test_per_qubit_noise_from_data():
-    # Generate the noise model from calibration data.
+def test_noise_from_metrics_requires_type():
+    # Attempt to generate a noise model without specifying a noise type.
+    calibration = cirq.google.Calibration(_CALIBRATION_DATA)
+    with pytest.raises(ValueError) as e:
```
dabacon: there is a helper for matching error messages.  
``
with pytest.raises(ValueError, match='regext matching string'):

cirq/contrib/noise_models/noise_models.py
```diff
@@ -211,54 +212,142 @@ def noisy_moment(self, moment: 'cirq.Moment',
             ]
 
 
-class PerQubitDepolarizingNoiseModel(devices.NoiseModel):
-    """DepolarizingNoiseModel which allows depolarization probabilities to be
-    specified separately for each qubit.
+class PerQubitDepolarizingWithDampedReadoutNoiseModel(devices.NoiseModel):
+    """DepolarizingWithDampedReadoutNoiseModel which allows depolarization
```
dabacon: style nit: First line of pydoc should be a single line. No need to repeat class name.  
``Noise model with T1 added to gates and damping and bitflips applied to measurement.``

cirq/contrib/noise_models/noise_models.py
95-martin-orion: Done.

cirq/contrib/noise_models/noise_models.py
95-martin-orion: Done (opted for `_probs` suffix in all cases)

cirq/contrib/noise_models/noise_models.py
95-martin-orion: Done.

cirq/contrib/noise_models/noise_models.py
95-martin-orion: Done.

cirq/contrib/noise_models/noise_models_test.py
95-martin-orion: Done.

cirq/contrib/noise_models/noise_models.py
95-martin-orion: Done. The previous comment was referencing the treats-all-qubits-equally version of DepolarizingWithDampedReadoutNoiseModel, but this should be clearer.

cirq/contrib/noise_models/noise_models.py
95-martin-orion: The top-level `__init__.py` prohibits depending on `contrib`:
https://github.com/quantumlib/Cirq/blob/626e16e41d12098e201ce48a4bb25c047cc86bcd/cirq/__init__.py#L53-L55

We _could_ move both this method and the `PerQubit...` type to `cirq.google`, though - that might be cleaner than adding a new directory just for this type. What do you think?

95-martin-orion: Discussed offline - I added a TODO to move this later on.

PR #2716, opened 2020-01-29
Add algorithm for decomposing multi-controlled rotation

### LINKED ISSUE #2625: multiqubit controlled gate decompositions
It would be a nice feature if Cirq would support the multiqubit controlled gate decompositions presented in https://arxiv.org/abs/quant-ph/9503016

fedimser:
This is an implementation of algorithm described in https://arxiv.org/abs/quant-ph/9503016
Fixes #2625.


cirq/optimizers/controlled_gate_decomposer.py
```diff
@@ -0,0 +1,284 @@
...
+            ops.ZPowGate(exponent=-0.25).on(c1),
+            ops.CNOT.on(c0, c1),
+        ]
+
+    def _ccnot_congruent(self, c0: 'cirq.Qid', c1: 'cirq.Qid',
+                         target: 'cirq.Qid'):
+        """Implements 'congruent' CCNOT.
+
+        This is almost like CCNOT, but multiplies 4th qubit by -1.
+        See lemma 6.2 in [1]."""
```
Strilanc: What is meant by "it multiplies the 4th qubit by -1"? There's only three qubits.

This set of operations doesn't perform a Toffoli, they perform a Toffoli with an inverted control and also there is a CCZ present.

A series of operations closer to a Toffoli is this one:

image

Which is equivalent to this:

image

It's better because the Toffoli has no inverted controls. There is of course still some phase error in the form of the CCZ and the Z.

cirq/optimizers/controlled_gate_decomposer.py
```diff
@@ -0,0 +1,284 @@
...
+            ops.ry(np.pi / 4).on(target),
+            ops.CNOT(c1, target),
+            ops.ry(np.pi / 4).on(target),
+        ]
+
+    def _ccnot(self,
+               c0: 'cirq.Qid',
+               c1: 'cirq.Qid',
+               target: 'cirq.Qid',
+               congruent: bool = False):
```
Strilanc: I would rename `congruent` to `allow_incorrect_phase`.

cirq/optimizers/controlled_gate_decomposer.py
```diff
@@ -0,0 +1,284 @@
...
+            # only for completeness.
+            X = np.array([[0, 1], [1, 0]])
+            return self.decompose(X, controls, target)
+
+    def decompose(self,
+                  matrix: np.ndarray,
+                  controls: List['cirq.Qid'],
+                  target: 'cirq.Qid',
+                  free_qubits: List['cirq.Qid'] = None,
+                  power: float = 1.0):
```
Strilanc: Drop the `power` argument. If the user wants a different operation (such as a smaller power), they should just provide the matrix for that operation.

cirq/optimizers/controlled_gate_decomposer.py
```diff
@@ -0,0 +1,284 @@
...
+    Q = np.array(eig_vectors)
+    return Q @ np.diag(np.exp(p * 1j * np.angle(eig_vals))) @ Q.conj().T
+
+
+def _is_identity(M):
+    """Checks whether M is identity."""
+    return np.allclose(M, np.eye(M.shape[0]))
+
+
+def _ry(x):
```
Strilanc: `protocols.unitary(cirq.Ry(x))`

Don't duplicate the definition of these matrices.

cirq/optimizers/controlled_gate_decomposer.py
```diff
@@ -0,0 +1,284 @@
...
+                self._ccnot(controls[m - 2 - i],
+                            free_qubits[m - 4 - i],
+                            free_qubits[m - 3 - i],
+                            congruent=True) for i in range(m - 3)
+            ]
+            gates2 = self._ccnot(controls[0],
+                                 controls[1],
+                                 free_qubits[0],
+                                 congruent=True)
+            gates3 = _flatten(gates1) + gates2 + _flatten(gates1[::-1])
```
Strilanc: `gates3 = [*gates1, gates2, *gates1[::-1]]`

cirq/optimizers/controlled_gate_decomposer.py
```diff
@@ -0,0 +1,284 @@
...
+                                            free_qubits=[ctrl[-1]])
+                result = []
+                result += self.decompose(C, [ctrl[-1]], target)
+                result += cnot_seq
+                result += self.decompose(B, [ctrl[-1]], target)
+                result += cnot_seq
+                result += self.decompose(A, [ctrl[-1]], target)
+                return result
+            else:
+                # O(n^2) decomposition - [1], lemma 7.5.
```
Strilanc: Extract these large blocks into subroutines.

Strilanc: Note that it's never really necessary to use O(n^2) gates. See https://algassert.com/circuits/2015/06/22/Using-Quantum-Gates-instead-of-Ancilla-Bits.html . But I admit implementing that would be unnecessarily more work...

cirq/optimizers/controlled_gate_decomposer.py
```diff
@@ -0,0 +1,284 @@
...
+        else:
+            gate_is_special_unitary = np.allclose(delta, 0)
+
+            if gate_is_special_unitary:
+                # O(n) decomposition of SU matrix - [1], lemma 7.9.
+                cnot_seq = self.decompose_x(ctrl[:-1],
+                                            target,
+                                            free_qubits=[ctrl[-1]])
+                result = []
+                result += self.decompose(C, [ctrl[-1]], target)
```
Strilanc: I would just use the operation `C(target).controlled_by(ctrl[-1])` and rely on the caller to further decompose if they need to.

cirq/optimizers/controlled_gate_decomposer.py
```diff
@@ -0,0 +1,284 @@
...
+        """
+        Args:
+            allow_toffoli - whether Toffoli (CCNOT) gates are allowed in the
+               output. If `False`, output will consist only of single-qubit
+               gates and CNOT gates. Defaults to `True`.
+        """
+        self.allow_toffoli = allow_toffoli
+
+    def _ccnot_true(self, c0: 'cirq.Qid', c1: 'cirq.Qid', target: 'cirq.Qid'):
+        """Implements action of Toffoli gate."""
```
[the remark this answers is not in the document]
fedimser: According to the paper, some Toffoli gates can be replaced with "congruent" gates, while others have to be implemented exactly.
If I return just Toffoli gates in my result, the user won't know which ones can be replaced with exact decomposition and which can be replaced with "congruent" decomposition.
That's why I have to do this decomposition inside my algorithm. 
Am I wrong?

cirq/optimizers/controlled_gate_decomposer.py
fedimser: Sorry, this should say "4th bit", i.e. bit 100.

About your suggestion - I'm declaring that my algorithm returns only single-qubit gates and CNOT gates. If I use cirquit your suggested - that will no longer be true.

cirq/optimizers/controlled_gate_decomposer.py
fedimser: I added this to prevent precision loss.
I agree it doesn't add value to user, but I call this method recursively, and every time I call it I need to calculate square root of a matrix. I found that it's more accurate to pass the same matrix but device power by 2.

If I make this method private, and provide public method without "power" argument - will it solve this problem?

cirq/optimizers/controlled_gate_decomposer.py
```diff
@@ -0,0 +1,284 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import List
+
+from cirq import linalg, ops
+import numpy as np
+
+
+def _unitary_power(U, p):
```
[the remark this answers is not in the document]
fedimser: Done.

cirq/optimizers/controlled_gate_decomposer.py
```diff
@@ -0,0 +1,284 @@
...
+
+def _rz(x):
+    return np.diag(np.exp([0.5j * x, -0.5j * x]))
+
+
+def _flatten(x):
+    return sum(x, [])
+
+
+class ControlledGateDecomposer:
```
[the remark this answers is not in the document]
fedimser: Removed class.

cirq/optimizers/controlled_gate_decomposer.py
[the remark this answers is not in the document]
fedimser: As discussed in meeting, will return Toffolis without decomposing them. But if Toffoli can be replaced with "congruent" Toffoli (namely, 3 CNOTs and 4 single-qubt gates) - will return it.

cirq/optimizers/controlled_gate_decomposer.py
fedimser: I removed option to provide correct decomposition of Toffoli - instead I will just return cir.CCNOT and let user decompose it if they want.
So now we only have version with "congruent=True", so no need for this parameter.

cirq/optimizers/controlled_gate_decomposer.py
fedimser: Removed `power` from public function, but I still left it in the `_decompose_recursive` function, because it improves precision.

cirq/optimizers/controlled_gate_decomposer.py
fedimser: Done.

cirq/optimizers/controlled_gate_decomposer.py
fedimser: I am implementing an algorithm which results in arbitrary single-qubit gates, CNOT and CCNOT gates - arbitrary single-controlled gates are not allowed. That's my understanding of the original paper.

Maybe it makes more sense to not decompose single-controlled gates and return them "as is", but it will be a different algorithm.

cirq/optimizers/controlled_gate_decomposer.py
fedimser: Done.
Btw, there is difference in sign between definition of Ry, Rz in paper and implementation in Cirq.

cirq/optimizers/controlled_gate_decomposer.py
fedimser: No, this doesn't work. This would work if gates1 would be List of gates. But it's list of lists of gates.

cirq/optimizers/controlled_gate_decomposer.py
fedimser: Thanks you. I can try to implement that if necessary, but maybe in another PR?

cirq/optimizers/controlled_gate_decomposer.py
fedimser: I kept this function but made comment clearer.

fedimser:
Craig, thank you for your comments.
I addressed all of them, PTAL.

I thnk it's important to specify what API we need to provide. In particular, which gates are allowed in output. 
Maybe it's actually better to leave all Toffoli gates instead of trying to decompose them in clever way.
This depends on in what applications this algorithm will be used, which I don't know.
Alexandru - can you please describe in what aplications this algorithm is supposed to be used?

@Strilanc 
@alexandrupaler

cirq/optimizers/controlled_gate_decomposer.py
Strilanc: The convention we use in Cirq is the one from Mike and Ike, so we'll stick with it.

cirq/optimizers/controlled_gate_decomposition.py
```diff
@@ -0,0 +1,254 @@
...
+        *_decompose_single_ctrl(_unitary_power(matrix, 0.5 * power),
+                                controls[-1], target), *cnots,
+        *_decompose_single_ctrl(_unitary_power(matrix, -0.5 * power),
+                                controls[-1], target), *cnots,
+        *_decompose_recursive(matrix, 0.5 * power, controls[:-1], target,
+                              [controls[-1]] + free_qubits)
+    ]
+
+
+def decompose_multi_controlled_unitary(matrix: np.ndarray,
```
Strilanc: Rename to `decompose_multi_controlled_rotation` to avoid ambiguity between single-qubit and multi-qubit unitaries.

cirq/optimizers/controlled_gate_decomposition.py
```diff
@@ -0,0 +1,254 @@
...
+    single-qubit gate with matrix `matrix` on `target`, controlled by
+    `controls`.
+
+    Result is guaranteed to consist exclusively of 1-qubit, CNOT and CCNOT
+    gates.
+
+    If matrix is special unitary, result has length `O(len(controls))`.
+    Otherwise result has length `O(len(controls)**2)`.
+
+    === REFERENCE ===
```
Strilanc: Make this a `References:` section similar to `Args:`.

cirq/optimizers/controlled_gate_decomposition.py
```diff
@@ -0,0 +1,254 @@
...
+        applying `MatrixGate(matrix).on(target).controlled_by(*controls)`.
+    """
+    assert is_unitary(matrix)
+    assert matrix.shape == (2, 2)
+
+    if len(controls) == 0:
+        return [ops.MatrixGate(matrix).on(target)]
+    elif len(controls) == 1:
+        return _decompose_single_ctrl(matrix, controls[0], target)
+    else:
```
Strilanc: `elif is_special_unitary`

cirq/optimizers/controlled_gate_decomposition.py
fedimser: Done.

cirq/optimizers/controlled_gate_decomposition.py
fedimser: Done.

cirq/optimizers/controlled_gate_decomposition.py
fedimser: Done.

fedimser:
I addressed the comments. CI is passing.
@Strilanc

fedimser:
Can someone please take another look at this PR and approve it if no further changes are needed?
@Strilanc @alexandrupaler @viathor

PR #2717, opened 2020-01-29
Remove deprecated methods and improve reprs

Strilanc:
- Delete all v0.8 deprecated methods
- Allow cirq.Moment to take OP_TREE arguments, deprecate `operations=` keyword.
- GateOperation now prefers reprs like "{gate}({qubits})" when it can prove "{gate}" is safe to append () onto
- ControlledOperation now prefers reprs like "{op}.controlled_by({qubits})"
- GateOperation can now delegate to _op_repr_ of its gate
- MeasurementGate now prefers op reprs like "cirq.measure({})"
- Fix a few type annotations not using top-level types


cirq/ops/gate_operation.py
```diff
@@ -59,15 +58,23 @@ def with_gate(self, new_gate: 'cirq.Gate') -> 'cirq.Operation':
...
+        if hasattr(self.gate, '_op_repr_'):
+            result = self.gate._op_repr_(self.qubits)
+            if result is not None and result is not NotImplemented:
+                return result
+        gate_repr = repr(self.gate)
+        qubit_args_repr = ', '.join(repr(q) for q in self.qubits)
+        assert type(self.gate).__call__ == raw_types.Gate.__call__
+
         # Abbreviate when possible.
+        dont_need_on = re.match(r'^[a-zA-Z0-9.()]+$', gate_repr)
```
viathor: Regex is too big a hammer for this since we're not checking any structural properties of the string, but merely that its characters belong to a given set (and that the string is non-empty). I don't think we really need to check whether the string is non-empty, so we can probably just do this:

```python
dont_need_on = all(c.isalnum() or c in '(.)' for c in gate_repr)
```

And if we do need to know that `gate_repr` is non-empty then you can just say

```python
dont_need_on = gate_repr and all(c.isalnum() or c in '(.)' for c in gate_repr)
```

dstrain115:
So, I think the PR is fine, but I am more worried about these deprecated operations.  We did say 0.8.0, but it seems like this happened fairly quickly due to our faster release schedule.

Particularly the Circuit.from_ops() and the Gate.on() seem to be very basic changes that will break a lot of code.  

Should we email cirq-announce@ or do any other preemptive action to make sure we don't break all the cirq code out there?  (For instance, I bet there's a lot of broken cirq tutorial colabs, for instance).

mpharrigan:
I agree with @dstrain115 that `Circuit.from_ops` seems to be going away rather quickly relative to its prominence (i.e. nearly every script ever used it until 6 months ago). I'd add `op_gate_of_type` into the "maybe preserve" bucket as well because I know we use it a lot writing custom compilation routines for applications and it doesn't seem to be hurting anyone.

I don't think `Gate.on()` is changing? Just the repr looks nicer

PR #2725, opened 2020-01-31
Initial T2 experiment

dstrain115:
- Creates initial draft experiment to measure T2 and T2*

- This experiment sets up a qubit on the Bloch sphere equator,
evolves the system (optionally performing a spin echo half way),
then measures the system in two bases to determine an approximate
length of the Bloch vector.


cirq/experiments/t2_decay_experiment.py
```diff
@@ -0,0 +1,336 @@
...
+        A T2DecayResult object that stores and can plot the data.
+    """
+    min_delay_dur = value.Duration(min_delay)
+    max_delay_dur = value.Duration(max_delay)
+
+    # Input validation
+    if repetitions <= 0:
+        raise ValueError('repetitions <= 0')
+    if repetitions % 2 != 0:
+        raise ValueError('Total repetitions must be an even number.')
```
peterse: out of curiousity, how come?

cirq/experiments/t2_decay_experiment.py
```diff
@@ -0,0 +1,336 @@
...
+    after the second half of the delay period.
+
+    CPMG, or the Carr-Purcell-Meiboom-Gill sequence, is currently not
+    implemented.
+
+    Args:
+        sampler: The quantum engine or simulator to run the circuits.
+        qubit: The qubit under test.
+        experiment_type: The type of T2 test to run.
+        num_points: The number of evenly spaced delays to test.
```
peterse: It might be worth allowing logarithmic sampling scheme; doing a T2 fit (for example eq. 2.3 in [O'Malley, 2016](https://web.physics.ucsb.edu/~martinisgroup/theses/OMalley2016.pdf)) you can get higher certainty fit parameters with a larger range of times that includes the full range of decay. 

I don't know what a clean implementation of this would look like. Maybe a `Sequence[cirq.DURATION_LIKE]` input type instead of minimum, maximum, steps

cirq/experiments/t2_decay_experiment.py
```diff
@@ -0,0 +1,336 @@
...
+    for tab in [x_basis_tabulation, y_basis_tabulation]:
+        for col_index, name in [(1, 0), (2, 1)]:
+            if name not in tab:
+                tab.insert(col_index, name, [0] * tab.shape[0])
+
+    # Return the results in a container object
+    return T2DecayResult(x_basis_tabulation, y_basis_tabulation)
+
+
+class T2DecayResult:
```
peterse: This is cool! Are there plans for some kind of abstract class that provides expectations with respect to specific runs/sweeps? This seems like a special case of that

dstrain115: @mpharrigan is working on a framework to do something like this.  Matt, are you presenting your proposed observation / expectation idea during any of the upcoming cirq cyncs?  (you should).

cirq/experiments/t2_decay_experiment.py
dstrain115: Sounds like a good idea.  How about I add this in a future PR?  I am not sure I want to tackle it right now.  I'd rather get the basics in, then I can add different sampling techniques.  It shouldn't be too difficult to add in.

cirq/experiments/t2_decay_experiment.py
dstrain115: Whoops.  It was initially because I split the repetitions between x-basis and y-basis measurements.  It got too confusing whether things should be reps or reps/2 all over the place.

Removed this check.

cirq/experiments/t2_decay_experiment.py
mpharrigan: It's been pushed to the back burner because of some other projects, but I should have some code to share soon!

dstrain115: Going over the design doc with the public cirq community might be good to get feedback before you do too much coding though!

peterse: cool, looking forward to it

cirq/experiments/t2_decay_experiment.py
peterse: Yes, fair enough

cirq/experiments/__init__.py
```diff
@@ -39,3 +39,7 @@
     t1_decay,
     T1DecayResult,
 )
+from cirq.experiments.t2_decay_experiment import (
+    t2_decay,
+    T2DecayResult,
+)
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/experiments/t2_decay_experiment.py
```diff
@@ -0,0 +1,336 @@
+# Copyright 2019 The Cirq Developers
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/experiments/t2_decay_experiment.py
```diff
@@ -0,0 +1,336 @@
...
+        sweep = study.Product(delay_sweep, tomography_sweep)
+    elif experiment_type == ExperimentType.HAHN_ECHO:
+        # Hahn / Spin Echo T2 experiment
+        # Use sqrt(Y) to flip to the equator.
+        # Evolve the state for half the given amount of delay time
+        # Flip the state using an X gate
+        # Evolve the state for half the given amount of delay time
+        # Then measure the state in both X and Y bases.
+
+        # TODO: Should be 0.5 * delay_var
```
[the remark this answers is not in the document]
dstrain115: Just changed it now instead.

cirq/experiments/t2_decay_experiment_test.py
```diff
@@ -0,0 +1,276 @@
+# Copyright 2019 The Cirq Developers
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/experiments/__init__.py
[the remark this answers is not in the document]
vtomole: Maybe something wrong with my browser. I don't see the newline added.

cirq/experiments/t2_decay_experiment.py
```diff
@@ -0,0 +1,333 @@
...
+class ExperimentType(enum.Enum):
+    RAMSEY = 1  # Often denoted as t2*
+    HAHN_ECHO = 2  # Spin echo or t2
+    CPMG = 3  # Carr-Purcell-Meiboom-Gill sequence
+
+
+_T2_COLUMNS = ['delay_ns', 0, 1]
+
+
+def t2_decay(sampler: work.Sampler,
```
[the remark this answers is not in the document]
dstrain115: Based on offline comments:

1) We have agreed that estimating Bloch vector size should be sufficient.  Adding the detuning as specified in this comment seems to give confusing results, and I am not sure if it is doing what is intended.  For now, we will not include this functionality.

2) This comment will be included for CPMG in a follow-up PR and not for Spin Echo.

dstrain115:
@XiaoMiQC mentioned offline to go ahead and commit this, since he has no time to review.  I can go ahead and make further revisions as necessary after committing if anyone has additional comments.

dstrain115:
> LGTM, maybe we need an issue to track CPMG?

Added #3001

PR #2734, opened 2020-02-04
Apply 'virtual' tag in noise models.

95-martin-orion:
Depends on #2670.

In order to prevent noise models from applying noise to pre-existing noise, we append the 'virtual' tag to all gates added by the noise model. This makes composition of noise models much simpler; to demonstrate this, the multiple-noise-type models in `contrib` have been split up into single types.


dstrain115:
Two questions:
1) Should we make Virtual a class rather than a string?
2) How does this interact with the 'virtual' nature of Virtual Z gates?  Is this the same tag, or different tags?

95-martin-orion:
> Two questions:
> 
> 1. Should we make Virtual a class rather than a string?
> 2. How does this interact with the 'virtual' nature of Virtual Z gates?  Is this the same tag, or different tags?

1. Would this mean having a VirtualOperation class, or somehow supporting tags of different types? If it's the latter, I don't think this PR will be affected (we'd still need to check for `virtual in op.tags`). For the former, I think there's been discussion about this, but there's some disagreement as far as how the "Virtual-ness" of an operation would propagate e.g. through compilation.
2. As I understand it, `virtual` here means "this operation has no duration, or takes place as part of another gate which *does* have a duration". That seems to match with the idea of virtual-Z gates - I don't think we need to apply noise to a zero-duration gate.

dstrain115:
Regarding (1), I mean the second option, creating a class VirtualTag rather than use the magic string 'virtual'.  I always hesitate to use magic strings that are copied all over the place.  This shouldn't change the PR much beyond checking for a class rather than a string, but then it could be shared more easily between noise models and gates.

95-martin-orion:
> Regarding (1), I mean the second option, creating a class VirtualTag rather than use the magic string 'virtual'.

That seems reasonable.

95-martin-orion:
> > Regarding (1), I mean the second option, creating a class VirtualTag rather than use the magic string 'virtual'.
> 
> That seems reasonable.

VirtualTag has now been converted to a class - ready for review.

cirq/contrib/noise_models/noise_models.py
```diff
@@ -52,93 +54,134 @@ def __init__(self, depol_prob: float):
...
+        Args:
+            depol_prob_map: Map of depolarizing probabilities for each qubit.
+        """
+        for qubit, depol_prob in depol_prob_map.items():
+            value.validate_probability(depol_prob, f'depol prob of {qubit}')
+        self.depol_prob_map = depol_prob_map
+
+    def noisy_moment(self, moment: 'cirq.Moment',
+                     system_qubits: Sequence['cirq.Qid']):
+        if _homogeneous_moment_is_measurements(moment):
```
dstrain115: Does this need a is_virtual_moment check?

cirq/ops/raw_types.py
```diff
@@ -609,6 +609,24 @@ def _equal_up_to_global_phase_(self,
                                                   atol=atol)
 
 
+class VirtualTag():
```
dstrain115: I am not sure that raw_types is the best place to accumulate tags.  

We should either make a virtual_tag.py or a tags.py to put this in.

cirq/contrib/noise_models/noise_models.py
95-martin-orion: Good point - done.

cirq/ops/raw_types.py
95-martin-orion: Opted for `tags.py` in lieu of other tag types.

cirq/ops/tags.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2018 The Cirq Developers
```
dstrain115: 2020

cirq/ops/tags_test.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2018 The Cirq Developers
```
dstrain115: 2020

95-martin-orion:
Thanks Doug!

PR #2739, opened 2020-02-05
Fix n-qubit state tomography

### LINKED ISSUE #2737: Too many implementations of state tomography
We have three implementations of state tomography:
- for 1-qubit states,
- for 2-qubit states,
- for n-qubit states for arbitrary n.

The last one isn't well tested (yet). Once it is, we should compare performance and if there isn't a clear advantage to the fixed-size implementations we should remove them.

viathor:
Fix #2737 by taking qubit order into account in formulating the system of linear equations.

### DIFF
--- a/cirq/experiments/n_qubit_tomography.py
+++ b/cirq/experiments/n_qubit_tomography.py
@@ -80,10 +80,16 @@ class StateTomographyExperiment:
 
         self.rot_circuit = circuits.Circuit(operations)
         self.rot_sweep = study.Product(*sweeps)
-        self.mat = self._make_state_tomography_matrix()
+        self.mat = self._make_state_tomography_matrix(qubits)
 
-    def _make_state_tomography_matrix(self) -> np.ndarray:
+    def _make_state_tomography_matrix(
+            self,
+            qubits: Sequence['cirq.Qid'],
+    ) -> np.ndarray:
         """Gets the matrix used for solving the linear system of the tomography.
+
+        Args:
+            qubits: Qubits to do the tomography on.
 
         Returns:
             A matrix of dimension ((number of rotations)**n * 2**n, 4**n)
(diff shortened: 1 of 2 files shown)


cirq/experiments/n_qubit_tomography.py
```diff
@@ -96,8 +102,8 @@ def _make_state_tomography_matrix(self) -> np.ndarray:
 
         # Unitary matrices of each rotation circuit.
         unitaries = np.array([
-            protocols.unitary(
-                protocols.resolve_parameters(self.rot_circuit, rots))
+            protocols.resolve_parameters(self.rot_circuit,
+                                         rots).unitary(qubit_order=qubits)
```
mpharrigan: aren't you supposed to do `qubit_order = cirq.QubitOrder.explicit(qubits)` or did we simplify this?

viathor: I've checked and think the present code is OK for two reasons:
1. Type annotation is `qubit_order: 'cirq.QubitOrderOrList'` which is defined as `Union[qubit_order.QubitOrder, Iterable[raw_types.Qid]]` see ([here](https://github.com/quantumlib/Cirq/blob/9c0eb4c0ea81f9b197cd2bfa87110ce531f3b6f9/cirq/ops/qubit_order_or_list.py#L26)).
2. The first thing `Circuit.unitary` does with the argument is to call `ops.QubitOrder.as_qubit_order(qubit_order)` (see [here](https://github.com/quantumlib/Cirq/blob/8612933e2dc8f4db1392187a72a2ecd132f6cee8/cirq/circuits/circuit.py#L1409)) which if the argument is an iterable calls `QubitOrder.explicit` (see [here](https://github.com/quantumlib/Cirq/blob/9c0eb4c0ea81f9b197cd2bfa87110ce531f3b6f9/cirq/ops/qubit_order.py#L125)).

PR #2744, opened 2020-02-06
Fix sign error in PhasedXZGate._canonical

Strilanc:


cirq/ops/phased_x_z_gate_test.py
```diff
@@ -198,10 +203,18 @@ def test_from_matrix():
...
                           g,
                           atol=1e-8)
-    u = cirq.testing.random_unitary(2)
+
+
+@pytest.mark.parametrize('unitary', [
+    cirq.testing.random_unitary(2),
+    cirq.testing.random_unitary(2),
+    cirq.testing.random_unitary(2),
+    np.array([[0, 1], [1j, 0]]),
```
Strilanc: This case used to fail.

cirq/ops/phased_x_z_gate_test.py
```diff
@@ -120,7 +121,11 @@ def f(x, z, a):
     t = f(1, 0.25, 0.5)._canonical()
     assert t.x_exponent == 1
     assert t.z_exponent == 0
-    assert t.axis_phase_exponent == 0.375
+    assert t.axis_phase_exponent == -0.375
```
Strilanc: The sign in this test was wrong.

mpharrigan:
Could this have caused any incorrect results? Or is this just a failure to adhere to the canonicalization convention?

Issue #2746, opened 2020-02-06
aqt module import


alexandrupaler:
without explicitly doing `from cirq.aqt import AQTSampler` although` import cirq` was done, the code line below raises `AttributeError: module 'cirq' has no attribute 'aqt'`

```
 aqt_sampler = cirq.aqt.AQTSampler(
        remote_host="rmeote",
        access_token="mykey"
    )
```

mpharrigan:
FYI this is pretty standard python behavior

```
In[10]: import scipy
In[11]: scipy.linalg.expm()
Traceback (most recent call last):
  File "/usr/local/google/home/mpharrigan/cirq/py37/lib/python3.7/site-packages/IPython/core/interactiveshell.py", line 3326, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-11-d534fa3c30db>", line 1, in <module>
    scipy.linalg.expm()
AttributeError: module 'scipy' has no attribute 'linalg'
In[12]: import scipy.linalg
In[13]: scipy.linalg.expm()
Traceback (most recent call last):
  File "/usr/local/google/home/mpharrigan/cirq/py37/lib/python3.7/site-packages/IPython/core/interactiveshell.py", line 3326, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-13-d534fa3c30db>", line 1, in <module>
    scipy.linalg.expm()
TypeError: expm() missing 1 required positional argument: 'A'
```

alexandrupaler:
Thank you. Did not know. Nevertheless, I find it annoying.

mpharrigan:
Yes, in Cirq we import everything, so I think #2748 is the correct change! 

I think scipy doesn't do it to keep import times low

vtomole:
This is not an issue anymore due to the recently adopted packaging strategy.
```python
>>> import cirq_aqt
>>> cirq_aqt.AQTSampler(remote_host="remote", access_token="mykey")
<cirq_aqt.aqt_sampler.AQTSampler object at 0x7fb613b97280>
```

PR #2748, opened 2020-02-06
simplified imports in aqt to work with user scripts

alexandrupaler:
this should solve #2746


cirq/__init__.py
```diff
@@ -485,3 +485,7 @@
     google,
     testing,
 )
+
+from cirq import (
+    aqt,
```
mpharrigan: include in the list above titled "unflattened sub-modules"

cirq/aqt/aqt_sampler.py
```diff
@@ -29,19 +29,19 @@
 
 import numpy as np
 from requests import put
-from cirq import circuits, Sampler, resolve_parameters, LineQubit
-from cirq.study.sweeps import Sweep
-from cirq.aqt.aqt_device import AQTSimulator, get_op_string
-from cirq import study, ops, IonDevice
+
+import cirq
```
mpharrigan: @Strilanc what's the convention for submodules importing cirq directly? I think you're still "supposed" to 

```python
from cirq import circuits, ops, protocols
```

etc.

mpharrigan: ping @Strilanc

mpharrigan: xref https://github.com/quantumlib/Cirq/issues/2787 

Unless someone tells me otherwise, I don't think you're supposed to `import cirq` at the top level for any "internal" code (that's part of cirq)

dstrain115: We should not "import cirq" from within non-test python files within cirq.

In tests, you can "import cirq"

cirq/aqt/__init__.py
```diff
@@ -12,6 +12,12 @@
...
-from cirq.aqt.aqt_sampler import AQTSampler, AQTSamplerLocalSimulator
-from cirq.aqt.aqt_device import AQTSimulator
+from cirq.aqt.aqt_sampler import (
+    AQTSampler,
+    AQTSamplerLocalSimulator,
+)
+
+from cirq.aqt.aqt_device import (AQTSimulator, AQTNoiseModel, get_aqt_device,
+                                 get_aqt_default_noise_dict,
+                                 get_aqt_device_op_string)
```
dstrain115: This formatting looks a little goofy.  Maybe try adding a trailing comma at the end of get_aqt_device_op_string to see if it fixes things?

dstrain115: This formatting looks a little goofy.  Maybe try adding a trailing comma at the end of get_aqt_device_op_string to see if it fixes things?

cirq/aqt/aqt_device.py
```diff
@@ -26,50 +26,32 @@
...
-from cirq import ops, devices, study
-from cirq import Circuit, LineQubit, IonDevice, Duration
-from cirq import DensityMatrixSimulator
 
-gate_dict = {'X': ops.X, 'Y': ops.Y, 'Z': ops.Z, 'MS': ops.XX}
+from cirq.ops import MeasurementGate, \
+    X, Y, Z, XX,\
+    ZPowGate, XPowGate, YPowGate, XXPowGate,\
+    Moment, Operation, Qid,\
+    depolarize, measure
```
dstrain115: I think you can add parentheses around the imports to avoid the "\"s

I think the convention is to say "from cirq import ops" and then use ops.X below.  That way you don't need to list every single operation in ops.

cirq/aqt/aqt_device.py
```diff
@@ -26,50 +26,32 @@
...
-        op_str = 'Z'
-    elif isinstance(op_obj, ops.MeasurementGate) or isinstance(
-            op_obj.gate, ops.MeasurementGate):
-        op_str = 'Meas'
-    else:
-        raise ValueError('Got unknown gate:', op_obj)
-    return op_str
+from cirq.study import TrialResult
+
+gate_dict = {'X': X, 'Y': Y, 'Z': Z, 'MS': XX}
```
dstrain115: Global variables should be all capitals.  Also, I would put a comment explaining why this is needed.

cirq/aqt/aqt_device.py
```diff
@@ -26,50 +26,32 @@
...
 
 
-class AQTNoiseModel(devices.NoiseModel):
+class AQTNoiseModel(NoiseModel):
     """A noise model for the AQT ion trap """
```
dstrain115: What noise does this model?  Add more description here please.

cirq/aqt/aqt_device.py
```diff
@@ -221,13 +203,38 @@ def get_aqt_device(num_qubits: int) -> Tuple[IonDevice, List[LineQubit]]:
...
+        'X': depolarize(1e-3),
+        'Y': depolarize(1e-3),
+        'Z': depolarize(1e-3),
+        'MS': depolarize(1e-2),
         'crosstalk': 0.03
     }
     return default_noise_dict
+
+
+def get_aqt_device_op_string(op_obj: Operation) -> str:
```
dstrain115: Is this needed for users?  If so, why?  

If not, I would change it to _get_aqt_device_op_string to denote that it is private.

cirq/aqt/aqt_sampler.py
```diff
@@ -25,20 +25,35 @@
...
+
+from cirq.ops import GateOperation, EigenGate
+
+from cirq.protocols import resolve_parameters
+
+from cirq.ion import IonDevice
+
+from cirq.devices import LineQubit
+
+from cirq.aqt.aqt_device import AQTSimulator, get_aqt_device_op_string
```
dstrain115: Unneeded whitespace.  Also, I think these imports were fine before.

The problem is importing all of cirq with `import cirq`  
Importing ops should be fine.  Casey's test should prevent more accidental circular imports

cirq/aqt/aqt_device.py
alexandrupaler: I am not sure what noise model this is. I am not the original developer of the module.

cirq/aqt/aqt_sampler.py
alexandrupaler: If it is fine, I will keep it like this. I thought that importing ops is also too big. I would prefer not to change everywhere in the code the occurences.

cirq/aqt/__init__.py
```diff
@@ -12,6 +12,16 @@
...
+    AQTSampler,
+    AQTSamplerLocalSimulator,
+)
+
+from cirq.aqt.aqt_device import (
+    AQTSimulator,
+    AQTNoiseModel,
+    get_aqt_device,
+    get_aqt_default_noise_dict,
+    _get_aqt_device_op_string,
```
dabacon, line 25: Do you want to expose _get_aqt_device_op_string to users?  If so I would not make it private (remove '_')

cirq/aqt/aqt_device.py
```diff
@@ -26,50 +26,28 @@
 import json
 from typing import Union, Tuple, List, Dict, Sequence, Any, cast
 import numpy as np
-from cirq import ops, devices, study
-from cirq import Circuit, LineQubit, IonDevice, Duration
-from cirq import DensityMatrixSimulator
 
-gate_dict = {'X': ops.X, 'Y': ops.Y, 'Z': ops.Z, 'MS': ops.XX}
+from cirq.ops import (MeasurementGate, X, Y, Z, XX, ZPowGate, XPowGate,
```
dabacon, line 30: is there a reason you moved away from `from cirq import ops, ...` pattern.  Google style guide is that you should try to only import packages and modules (https://engdoc.corp.google.com/eng/doc/devguide/py/style/index.md?cl=head#s2.2-imports)

cirq/aqt/aqt_device.py
```diff
@@ -221,13 +199,39 @@ def get_aqt_device(num_qubits: int) -> Tuple[IonDevice, List[LineQubit]]:
     return ion_device, qubit_list
 
 
-def get_default_noise_dict() -> Dict[str, Any]:
+def get_aqt_default_noise_dict() -> Dict[str, Any]:
```
dabacon, line 202: Orthogonal (no action necessary): How stable is this api?  I think one should avoid `get` in method names as they do not add much.  `aqt_default_noise_dict()` should be just as good.

balopat:
Hi @alexandrupaler - friendly ping - are you planning to continue working on this? It looks like it's not too far away. However, given that we'll redesign the packages in the coming month, maybe it's better to just close it now. Let me know how strongly you feel about it!

alexandrupaler:
it is ok to be closed. i will not continue it.

PR #2752, opened 2020-02-10
Remove Google index page and link directly to Google pages

dstrain115:


dstrain115:
> LGTM; please confirm that it builds locally and looks good

Looks good when I build it locally.  Thanks!

PR #2754, opened 2020-02-10
Make Best Practices its own page

dstrain115:
- This looks like it may be better on its own page
- This will also facilitate ease of modification as early uses
will be able to add to this page more easily than if combined
with other stuff.


docs/google/best_practices.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Use caution with symbols
+
+Symbols are extremely useful for constructing parameterized circuits (see above).  However,
+only some sympy formulas can be serialized for network transport to the engine.
+Currently, sums and products of symbols, including linear combinations, are supported.
+See `cirq.google.arg_func_langs` for details.
+
+The sympy library is also infamous for being slow, so avoid using complicated formulas if you
+care about performance.  Avoid using parameter resolvers that have formulas in them.
```
dabacon: Aside: does using the flatten capabilities tend to help this?  Mentioning it here would probably be good if it does.

PR #2756, opened 2020-02-10
Make LineQubit and GridQubit immutable

### LINKED ISSUE #2356: Qubits shouldn't be mutable

fedimser:
Coordinates of qubits on line or grid now are properties and can't be changed.

Fixes #2356.


fedimser:
Hi. I am not sure whether this solution is good enough.
It doesn't make qubits "really" immutable, because I still can access private properies (e.g. qubit._x), and I can add and modfy any other fields.

If this is acceptable solution, then NamedQubit is already good, because `name` is a property.

@mpharrigan @Strilanc

cirq/devices/grid_qubit_test.py
```diff
@@ -185,3 +185,9 @@ def test_to_json():
         'row': 5,
         'col': 6,
     }
+
+
+def test_immutable():
+    with pytest.raises(AttributeError, match="can't set attribute"):
+        q = cirq.GridQubit(1, 2)
+        q.col = 3
```
cduck: Also test `.row`.

mpharrigan:
Yes, nothing is truly immutable in python. This is good

mpharrigan:
Or `object.__setattr__()` (which is how frozen dataclasses set their members to begin with :)

cirq/devices/grid_qubit_test.py
fedimser: Added a test.

PR #2759, opened 2020-02-11
Add tags to circuit diagrams

dstrain115:
- Print out tags in circuit diagrams
- Also includes a parameter 'include_tags' that can disable this feature.


cirq/ops/raw_types.py
```diff
@@ -578,8 +578,13 @@ def _resolve_parameters_(self, resolver):
 
     def _circuit_diagram_info_(self, args: 'cirq.CircuitDiagramInfoArgs'
                               ) -> 'cirq.CircuitDiagramInfo':
-        return protocols.circuit_diagram_info(self.sub_operation, args,
-                                              NotImplemented)
+        sub_op_info = protocols.circuit_diagram_info(self.sub_operation, args,
+                                                     NotImplemented)
+        if args and args.include_tags:
+            sub_op_info.wire_symbols = (
+                (sub_op_info.wire_symbols[0] + str(list(self._tags)),) +
```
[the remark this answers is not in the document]
dstrain115: Yup.  Fixed this and added a test, though it seems that returning NotImplemented returns a TypeError from the protocol

cirq/ops/raw_types.py
```diff
@@ -578,8 +578,13 @@ def _resolve_parameters_(self, resolver):
 
     def _circuit_diagram_info_(self, args: 'cirq.CircuitDiagramInfoArgs'
                               ) -> 'cirq.CircuitDiagramInfo':
-        return protocols.circuit_diagram_info(self.sub_operation, args,
-                                              NotImplemented)
+        sub_op_info = protocols.circuit_diagram_info(self.sub_operation, args,
+                                                     NotImplemented)
+        if args and args.include_tags:
```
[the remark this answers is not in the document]

cirq/ops/raw_types.py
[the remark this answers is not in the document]
cduck: `args` will never be None.  The protocol handles the default value for `args`.

If `_circuit_diagram_info_` returns `NotImplemented`, the diagram drawer will fallback to the gate repr.  With this code, tags won't be printed for gates that don't implement `_circuit_diagram_info_`.  I've had a similar issue dealing with `NotImplemented` in the past when wrapping gates.  I think the long-term solution is to make `cirq.circuit_diagram_info` handle the default case instead of the diagram drawer and always return a valid `CircuitDiagramInfo` object.

cirq/protocols/circuit_diagram_info_protocol.py
```diff
@@ -102,37 +102,45 @@ class CircuitDiagramInfoArgs:
...
-    def __init__(self, known_qubits: Optional[Iterable['cirq.Qid']],
-                 known_qubit_count: Optional[int], use_unicode_characters: bool,
+    def __init__(self,
+                 known_qubits: Optional[Iterable['cirq.Qid']],
+                 known_qubit_count: Optional[int],
+                 use_unicode_characters: bool,
                  precision: Optional[int],
-                 qubit_map: Optional[Dict['cirq.Qid', int]]) -> None:
+                 qubit_map: Optional[Dict['cirq.Qid', int]],
+                 include_tags: bool = True) -> None:
```
cduck: Should probably add some tests for the new attribute (equality group test, etc.).

cirq/ops/raw_types_test.py
```diff
@@ -459,6 +459,49 @@ def test_tagged_operation():
...
+    diagram_with_tags = "(1, 1): ───H['tag1']───"
+    diagram_without_tags = "(1, 1): ───H───"
+    assert str(cirq.Circuit(tagged_h)) == diagram_with_tags
+    assert c.to_text_diagram() == diagram_with_tags
+    assert c.to_text_diagram(include_tags=False) == diagram_without_tags
+
+
+def test_circuit_diagram_corner_cases():
+    global_phase = cirq.GlobalPhaseOperation(coefficient=1.0).with_tags('tag0')
+    assert global_phase._circuit_diagram_info_(args=None) == NotImplemented
```
cduck: This should test `cirq.circuit_diagram_info(globa_phase) == ?`

cirq/ops/raw_types_test.py
```diff
@@ -459,6 +459,49 @@ def test_tagged_operation():
...
+    assert cirq.Circuit(global_phase).to_text_diagram() == ''
```
cduck: This appears to be two bugs in circuit diagrams.  It should be `'\n\nglobal phase: 1.0'`.  It is `'\n\nglobal phase:'` with an untagged `GlobalPhaseOperation`.  But it is not recognized as a global phase when it is wrapped in `TaggedOperation` and just returns `''`.

Do we want to show tagged global phase operations in a circuit?  How should this work when the circuit combines all global phase ops into a single op?

cirq/ops/raw_types_test.py
```diff
@@ -459,6 +459,49 @@ def test_tagged_operation():
...
+
+    # Operation with no qubits and returns diagram info with no wire symbols
+    class NoWireSymbols(cirq.GlobalPhaseOperation):
+
+        def _circuit_diagram_info_(self, args: 'cirq.CircuitDiagramInfoArgs'
+                                  ) -> 'cirq.CircuitDiagramInfo':
+            return expected
+
+    no_wire_symbol_op = NoWireSymbols(coefficient=1.0).with_tags('tag0')
+    assert no_wire_symbol_op._circuit_diagram_info_(args=None) == expected
```
cduck: Same comment as above.

cirq/ops/raw_types_test.py
```diff
@@ -459,6 +459,49 @@ def test_tagged_operation():
...
+    assert cirq.Circuit(no_wire_symbol_op).to_text_diagram() == ''
```
cduck: I don't know what this output should be but it probably shouldn't be empty.

cirq/ops/raw_types.py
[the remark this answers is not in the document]
dstrain115: I added some code to handle TaggedOperations within the drawer classes in circuit.  I was a little hesitant to directly modify the protocol code.

I also added a test for a gate without a _circuit_diagram_info_

What do you think?

cirq/protocols/circuit_diagram_info_protocol.py
dstrain115: Added some lines for equality testing, but couldn't think of much else to test here.

cirq/ops/raw_types_test.py
dstrain115: This is tricky, since that just raises a TypeError.  Even though this accesses a private method, I felt like it was a better test, since it tests the forwarding.  We could test that the cirq.circuit_diagram_info(global_phase) raises TypeError, but that doesn't feel right to me, since there's a bunch of ways this can throw Type Error.

cirq/ops/raw_types_test.py
dstrain115: Fixed up the global_phase print outs.  Let me know what you think.

cirq/ops/raw_types_test.py
dstrain115: Fixed up the global_phase print outs.  Let me know what you think.

cirq/ops/raw_types_test.py
dstrain115: Fixed up the global_phase print outs.  Let me know what you think.

dstrain115:
Thanks for thinking of corner cases.  Let me know if you find others!

cirq/ops/raw_types_test.py
dstrain115: Wait, I don't think I did this quite right.  Let me try again....

cirq/circuits/circuit.py
```diff
@@ -1570,22 +1574,26 @@ def to_text_diagram_drawer(
...
-                isinstance(op, cirq.GlobalPhaseOperation)
-                for op in self.all_operations()):
-            diagram.write(0,
-                          max(qubit_map.values(), default=0) + 1,
-                          'global phase:')
+
+        # Print out global phase information
+        for op in self.all_operations():
+            if isinstance(op, cirq.GlobalPhaseOperation):
+                diagram.write(0,
```
cduck: This appears to overwrite the entry of the diagram for each global phase in the circuit so if there are more than one, the diagram will only show the last.  How do you want to handle coalescing tags from multiple global phase ops?

I think what the original code did (but I haven't tested) is just add 'global phase:' here and then print the value of each global phase op under each moment.  You can just add tags there.

cirq/circuits/circuit.py
```diff
@@ -1570,22 +1574,26 @@ def to_text_diagram_drawer(
...
+                              max(qubit_map.values(), default=0) + 1,
```
cduck: Aside: Changing this to -1 would remove the extra '\n\n' for diagrams with only global phase.

cirq/circuits/circuit.py
```diff
@@ -1741,12 +1749,20 @@ def _get_operation_circuit_diagram_info_with_fallback(
...
+
     # Representation usually looks like 'gate(qubit1, qubit2, etc)'.
     # Try to cut off the qubit part, since that would be redundant information.
     redundant_tail = '({})'.format(', '.join(str(e) for e in op.qubits))
     if name.endswith(redundant_tail):
         name = name[:-len(redundant_tail)]
 
+    # Add tags onto the representation, if they exist
+    if isinstance(op, cirq.TaggedOperation):
+        name += f' {list(op.tags)}'
```
cduck: Is the extra space intentional?  The other version just appended it without a space.

cirq/ops/raw_types_test.py
cduck: Then test that `cirq.circuit_diagram_info(..., default='default') == 'default'`.

cirq/ops/raw_types_test.py
```diff
@@ -459,6 +459,69 @@ def test_tagged_operation():
...
+    assert c.to_text_diagram(include_tags=False) == diagram_without_tags
+
+
+def test_circuit_diagram_corner_cases():
+    # Tests global phase operation
+    global_phase = cirq.GlobalPhaseOperation(coefficient=1.0).with_tags('tag0')
+    expected_diagram = "\n\nglobal phase ['tag0']: 1.0"
+    assert global_phase._circuit_diagram_info_(args=None) == NotImplemented
+    assert cirq.Circuit(global_phase).to_text_diagram() == expected_diagram
+    expected = cirq.CircuitDiagramInfo(wire_symbols=tuple(),
```
cduck: Nit: `()` is equivalent.

cirq/ops/raw_types_test.py
```diff
@@ -459,6 +459,69 @@ def test_tagged_operation():
...
+                                       exponent=1.0,
+                                       connected=True,
+                                       exponent_qubit_index=None,
+                                       auto_exponent_parens=True)
+
+    # Operation with no qubits and returns diagram info with no wire symbols
+    class NoWireSymbols(cirq.GlobalPhaseOperation):
```
cduck: I don't know if this needs to be tested since non-global phase zero-qubit ops are not really supported anyways.

cirq/ops/raw_types_test.py
dstrain115: Ok, committed a new patch to print out global phases correctly with and without tags.  I added some more tests and examples, so I think this looks better.  PTAL.

cirq/circuits/circuit.py
dstrain115: Yeah, I didn't quite understand how this worked when I made the first commit.  Can you take another look?

cirq/ops/raw_types_test.py
dstrain115: Maybe not, but I wanted to make sure that it didn't raise an exception, since parts of the code assume that a wire symbol exists.  I'll keep it in for now.

cirq/circuits/circuit.py
dstrain115: Changed.

cirq/ops/raw_types_test.py
dstrain115: Changed.

cirq/ops/raw_types_test.py
```diff
@@ -459,6 +459,100 @@ def test_tagged_operation():
...
+                                    use_unicode_characters=True)
+
+    tag1 = cirq.GlobalPhaseOperation(coefficient=1j).with_tags('tag1')
+    tag2 = cirq.GlobalPhaseOperation(coefficient=1j).with_tags('tag2')
+    c = cirq.Circuit([cirq.X(q), tag1, tag2])
+    cirq.testing.assert_has_diagram(c,
+                                    """\
+a: ─────────────X───────────────────
+
+global phase:   π['tag1', 'tag2']""",
```
cduck: Do you think combining tags is the right way to go?  What if they are both tagged tag1?  Should this be `['tag1', 'tag1']`?

dstrain115: That's how I have it implemented currently.  I don't know.  I think it's the most sensible option.  At least this gives you an indication that there are two tagged operations hiding in the moment and distinguishes it from one being tagged and the other not being tagged.  

Having two (redundant) GlobalPhaseOperations in the same moment both tagged with the same tags and displaying it in a circuit diagram is a goofy case, so I am not sure it's worth worrying about too much.

cirq/circuits/circuit.py
```diff
@@ -1861,15 +1874,24 @@ def _draw_moment_in_diagram(
...
-        for e in moment
-        if isinstance(e, ops.GlobalPhaseOperation)
-    ])
-    if global_phase != 1:
+    global_phase = None
+    tags: List[Any] = []
+    for op in moment:
+        if isinstance(op, ops.TaggedOperation):
+            tags.extend(op.tags)
+            op = op.sub_operation
```
cduck: Does this collect tags from all ops in the moment?  Add a test where the non-global phase gate has a tag.

cirq/ops/raw_types.py
```diff
@@ -578,8 +578,15 @@ def _resolve_parameters_(self, resolver):
 
     def _circuit_diagram_info_(self, args: 'cirq.CircuitDiagramInfoArgs'
                               ) -> 'cirq.CircuitDiagramInfo':
-        return protocols.circuit_diagram_info(self.sub_operation, args,
-                                              NotImplemented)
+        sub_op_info = protocols.circuit_diagram_info(self.sub_operation, args,
+                                                     NotImplemented)
+        # Add tag to wire symbol if it exists.
+        if (sub_op_info is not NotImplemented and args and args.include_tags and
```
cduck: I still don't think you need the check for args.  The type annotation guarantees this is not None.

cirq/ops/raw_types_test.py
cduck: I agree.  I forgot that many global phases could go into one moment.  So what does the output look like with a global phase in two different moments?

cirq/ops/raw_types.py
dstrain115: Removed.

cirq/circuits/circuit.py
dstrain115: Oops, yes.  Good catch.

Also, added test that has 2 moments, with non-global gates on each moment.

cduck:
@dstrain115 Ready to merge?

PR #2760, opened 2020-02-12
Add has_stabilizer_effect protocol and support native Clifford simulator gates

smitsanghavi:
Follow up to #2600 and part of #2423


smitsanghavi:
Only adds support for gates that can be run by CliffordState. Does not, yet, try to decompose.

cirq/ops/common_gates.py
```diff
@@ -191,6 +191,9 @@ def _phase_by_(self, phase_turns, qubit_index):
             exponent=self._exponent,
             phase_exponent=phase_turns * 2)
 
+    def _has_stabilizer_effect_(self) -> bool:
+        return self.exponent % 1 == 0
```
Strilanc: The exponent may be a symbol. If it's a symbol, return None. `if self._is_parameterized_(): ...`

cirq/ops/common_gates_test.py
```diff
@@ -530,6 +530,66 @@ def test_rz_unitary():
...
+    assert cirq.has_stabilizer_effect(gate)
+    assert not cirq.has_stabilizer_effect(gate**0.5)
+    assert cirq.has_stabilizer_effect(gate**0)
+    assert not cirq.has_stabilizer_effect(gate**-0.5)
+    assert cirq.has_stabilizer_effect(gate**4)
+    assert not cirq.has_stabilizer_effect(gate**1.2)
+
+
+def test_cnot_stabilizer():
+    gate = cirq.CNOT
```
Strilanc: Add symbol test cases.

cirq/protocols/has_stabilizer_effect_protocol.py
```diff
@@ -0,0 +1,61 @@
...
+
+def _strat_has_stabilizer_effect_from_has_stabilizer_effect(val: Any
+                                                           ) -> Optional[bool]:
+    """
+    Attempts to infer whether val has stabilzer effect via its
+    _has_stabilizer_effect_ method.
+    """
+    if hasattr(val, '_has_stabilizer_effect_'):
+        result = val._has_stabilizer_effect_()
+        if result is not NotImplemented:
```
Strilanc: `and result is not None`

cirq/protocols/has_stabilizer_effect_protocol_test.py
```diff
@@ -0,0 +1,112 @@
...
+class YesOp(EmptyOp):
+
+    @property
+    def gate(self):
+        return Yes()
+
+
+def test_inconclusive():
+    assert not cirq.has_stabilizer_effect(object())
+    assert not cirq.has_stabilizer_effect('boo')
```
Strilanc: The call should also work on raw gates.

cirq/ops/common_gates.py

cirq/ops/common_gates_test.py
smitsanghavi: Done

cirq/protocols/has_stabilizer_effect_protocol.py
smitsanghavi: Done

cirq/protocols/has_stabilizer_effect_protocol_test.py
smitsanghavi: I assume you didn't mean raw_type.Gate. Added a test case for a simple gate.

smitsanghavi:
Ping!

cirq/ops/common_gates_test.py
```diff
@@ -530,6 +530,78 @@ def test_rz_unitary():
...
+    assert cirq.has_stabilizer_effect(gate**4)
+    assert not cirq.has_stabilizer_effect(gate**1.2)
+    foo = sympy.Symbol('foo')
+    assert not cirq.has_stabilizer_effect(gate**foo)
+
+
+def test_y_stabilizer():
+    gate = cirq.Y
+    assert cirq.has_stabilizer_effect(gate)
+    assert not cirq.has_stabilizer_effect(gate**0.5)
```
Strilanc: Square root of Y is a stabilizer gate.

cirq/ops/common_gates_test.py
```diff
@@ -530,6 +530,78 @@ def test_rz_unitary():
...
+    assert cirq.has_stabilizer_effect(gate**0)
+    assert not cirq.has_stabilizer_effect(gate**-0.5)
```
Strilanc: Should be true not false.

cirq/ops/common_gates_test.py
```diff
@@ -530,6 +530,78 @@ def test_rz_unitary():
                                np.array([[1j, 0], [0, -1j]]))
 
 
+def test_x_stabilizer():
+    gate = cirq.X
+    assert cirq.has_stabilizer_effect(gate)
+    assert not cirq.has_stabilizer_effect(gate**0.5)
```
Strilanc: Should be true not false.

cirq/ops/common_gates_test.py
```diff
@@ -530,6 +530,78 @@ def test_rz_unitary():
...
+    assert cirq.has_stabilizer_effect(gate**0)
+    assert not cirq.has_stabilizer_effect(gate**-0.5)
```
Strilanc: Should be true not false.

cirq/ops/common_gates_test.py
```diff
@@ -530,6 +530,78 @@ def test_rz_unitary():
...
+    assert cirq.has_stabilizer_effect(gate**4)
+    assert not cirq.has_stabilizer_effect(gate**1.2)
+    foo = sympy.Symbol('foo')
+    assert not cirq.has_stabilizer_effect(gate**foo)
+
+
+def test_z_stabilizer():
+    gate = cirq.Z
+    assert cirq.has_stabilizer_effect(gate)
+    assert cirq.has_stabilizer_effect(gate**0.5)
```
Strilanc: Should be true not false.

cirq/ops/common_gates_test.py
```diff
@@ -530,6 +530,78 @@ def test_rz_unitary():
...
+    assert cirq.has_stabilizer_effect(gate**0)
+    assert cirq.has_stabilizer_effect(gate**-0.5)
```
Strilanc: Should be true not false.

cirq/ops/common_gates_test.py
```diff
@@ -530,6 +530,78 @@ def test_rz_unitary():
...
+    assert cirq.has_stabilizer_effect(gate)
+    assert not cirq.has_stabilizer_effect(gate**0.5)
+    assert cirq.has_stabilizer_effect(gate**0)
+    assert not cirq.has_stabilizer_effect(gate**-0.5)
+    assert cirq.has_stabilizer_effect(gate**4)
+    assert not cirq.has_stabilizer_effect(gate**1.2)
+    foo = sympy.Symbol('foo')
+    assert not cirq.has_stabilizer_effect(gate**foo)
+
+
```
Strilanc: You should also test that measurement is a stabilizer effect, and that gates like CCZ and the depolarizing channel are not.

cirq/ops/common_gates_test.py
smitsanghavi: As I mentioned in https://github.com/quantumlib/Cirq/pull/2760#issuecomment-585366509, I'm just adding support for the operations that CliffordState supports right now (just like it was before this) and does not try to decompose. Y**0.5 is a stabilizer gate and so is iSwap but just restricting myself to the original 7 gates before #2803 adds all of them to CliffordState since right now the simulator would just balk at Y**0.5.

If you feel we should add all these gates to the protocol at once, I can wait for the #2803 to go in. I can then have the protocol just do what is_supported_gate doing now as described in https://github.com/quantumlib/Cirq/pull/2803/files#r390152850. It would save on double work and definitely seems like a better option than having to define _has_stabilizer_effect_ on all gates.

PR #2764, opened 2020-02-12
Speed up single qubit gate layer generation for random quantum circuit

kevinsung:
This method is much faster when the set of single qubit gates is large since it avoids generating a new list of gates to choose from. For instance, I was using a set of 64 gates and got around a 15x speedup.

I also modified the code to handle the case of when only one choice of single-qubit gate is given. In that case, the condition of no repeating single-qubit gates is simply not enforced.

### DIFF
--- a/cirq/experiments/random_quantum_circuit_generation.py
+++ b/cirq/experiments/random_quantum_circuit_generation.py
@@ -13,8 +13,8 @@
 # limitations under the License.
 """Code for generating random quantum circuits."""
 
-from typing import (Callable, Container, Dict, Iterable, List, Sequence,
-                    TYPE_CHECKING, Tuple)
+from typing import (Callable, Container, Dict, Iterable, List, Optional,
+                    Sequence, TYPE_CHECKING, Tuple)
 
 from cirq import circuits, devices, google, ops, value
 from cirq._doc import document
@@ -198,10 +198,16 @@ def random_rotations_between_grid_interaction_layers_circuit(
 
     circuit = circuits.Circuit()
     previous_single_qubit_layer = {}  # type: Dict[cirq.GridQubit, cirq.Gate]
+    if len(set(single_qubit_gates)) == 1:
+        fixed_single_qubit_layer = {q: single_qubit_gates[0] for q in qubits}
+        single_qubit_layer_factory = _fixed_single_qubit_layer
+    else:
+        fixed_single_qubit_layer = None
+        single_qubit_layer_factory = _random_single_qubit_layer
     for i in range(depth):
-        single_qubit_layer = _single_qubit_layer(qubits, single_qubit_gates,
-                                                 previous_single_qubit_layer,
-                                                 prng)
+        single_qubit_layer = single_qubit_layer_factory(
+            qubits, single_qubit_gates, previous_single_qubit_layer, prng,
+            fixed_single_qubit_layer)
         two_qubit_layer = _two_qubit_layer(coupled_qubit_pairs,
                                            two_qubit_op_factory,
                                            pattern[i % len(pattern)], prng)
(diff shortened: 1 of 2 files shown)


cirq/experiments/random_quantum_circuit_generation.py
```diff
@@ -237,12 +241,16 @@ def _single_qubit_layer(
         previous_single_qubit_layer: Dict['cirq.GridQubit', 'cirq.Gate'],
         prng: 'np.random.RandomState',
 ) -> Dict['cirq.GridQubit', 'cirq.Gate']:
+    if len(single_qubit_gates) == 1:
```
kevinsung: So `len(set(single_qubit_gates))` would be safer here, but it diminishes the speedup down from 15x to around 3x.

kevinsung: I added to the documentation that the given gates should not repeat.

kevinsung:
The yapf on Travis is asking for something very strange, which is to have two consecutive newlines inside a function. My local yapf is not asking for that...

cirq/experiments/random_quantum_circuit_generation.py
viathor: Another way would be two have two versions of `_single_qubit_layer`: one for the case with only one type of allowed single-qubit gate and one for the case with more than one type. You can then perform the check to detect which case is taking place before the main for loop of `random_rotations_between_grid_interaction_layers_circuit`. This might make it a small bit faster, too.

kevinsung: Do you mean to have the two versions of `single_qubit_layer` have the same function call signature, even though the one with one type of allowed gate would not use any of its arguments? I think that would be too ugly. Any other way of doing it would require checking for which case we are in within the main loop anyway.

kevinsung:
I'm pretty reluctant to give in to the format check here. The change it's asking for doesn't make sense, and it conflicts with my local yapf, which is the same version as the one specified in `dev_tools/conf/pip-list-dev-tools.txt`.

cirq/experiments/random_quantum_circuit_generation.py
viathor: I don't think that's too ugly at all. You need an internal API to accommodate the two cases, so there is nothing wrong with ignoring parameters not relevant in one of the cases. If you place the functions near each other it should be fairly clear to the reader why the unused arguments exist. Also, the practice is in fact sufficiently common that pylint has a specific mechanism to accommodate it (by `del`eting the argument) and google python style guide explicitly discusses this, see [1].

On the other hand, the current code has a severe failure mode that's going to be tricky to debug. Specifically, when one passes `[cirq.X, cirq.X]` for `single_qubit_gates` the code enters an infinite loop!

---
[1] http://go/pystyle#lint

kevinsung: Very well, I have made the suggested change.

cirq/experiments/random_quantum_circuit_generation_test.py
```diff
@@ -77,9 +77,23 @@ def two_qubit_op_factory(a, b, prng):
...
+    depth = 20
+    pattern = cirq.experiments.GRID_STAGGERED_PATTERN
+    circuit = random_rotations_between_grid_interaction_layers_circuit(
+        qubits, depth, single_qubit_gates=[cirq.X**0.5, cirq.X**0.5], seed=1234)
+
+    assert len(circuit) == 2 * depth + 1
+    _validate_single_qubit_layers(qubits,
+                                  circuit[::2],
+                                  non_repeating_layers=False)
+    _validate_two_qubit_layers(qubits, circuit[1::2], pattern)
```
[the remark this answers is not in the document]
kevinsung: Done.

cirq/experiments/random_quantum_circuit_generation.py
```diff
@@ -231,22 +244,35 @@ def add_pair(neighbor: 'cirq.GridQubit'):
...
+def _fixed_single_qubit_layer(
+        qubits: List['cirq.GridQubit'],
+        single_qubit_gates: Sequence['cirq.Gate'],
+        previous_single_qubit_layer: Dict['cirq.GridQubit', 'cirq.Gate'],
+        prng: 'np.random.RandomState',
+        fixed_single_qubit_layer: Optional[Dict['cirq.GridQubit', 'cirq.Gate']]
+) -> Dict['cirq.GridQubit', 'cirq.Gate']:
+    return cast(Dict['cirq.GridQubit', 'cirq.Gate'], fixed_single_qubit_layer)
+
+
```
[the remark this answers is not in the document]
kevinsung: Done.

PR #2765, opened 2020-02-13
Save repetitions in CrossEntropyResult and make it into a JSON serializable dataclass

kevinsung:
@XiaoMiQC

### DIFF
--- a/cirq/experiments/cross_entropy_benchmarking.py
+++ b/cirq/experiments/cross_entropy_benchmarking.py
@@ -14,7 +14,6 @@
 
 from typing import (Any, Dict, Iterable, List, NamedTuple, Optional, Sequence,
                     Set, Tuple, Union)
-import dataclasses
 import numpy as np
 from matplotlib import pyplot as plt
 from cirq import circuits, devices, ops, protocols, sim, work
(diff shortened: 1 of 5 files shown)


kevinsung:
Mypy error same as #2769

kevinsung:
I just made mypy ignore the error.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -16,33 +16,25 @@
...
-        Each CrossEntropyPair is a NamedTuple that contains a cycle number and
-        the corresponding XEB fidelity.
-        """
-        return self._data
+    """Results from a cross-entropy benchmarking (XEB) experiment.
+
+    Attributes:
+        data: A sequence of NamedTuples, each of which
+            contains two fields: num_cycle which returns the circuit
+            depth as the number of cycles and xeb_fidelity which returns
```
viathor: nit: "as the number of cycles" explains nothing unless "cycle" itself is defined. I've seen the term being used for both a single layer of gates and for two layers: single-qubit gates following by two-qubit gates. Probably worth commenting on this explicitly.

kevinsung: Done.

PR #2766, opened 2020-02-13
Enable optimized_for_sycamore with single qubit MatrixGates

mpharrigan:
this is slightly inconsistent with the sycamore optimizer (which already handles this case), but only because the handling of single qubit operations is already inconsistent between the two


cirq/google/optimizers/convert_to_sqrt_iswap.py
```diff
@@ -101,6 +101,9 @@ def convert(self, op: 'cirq.Operation') -> List['cirq.Operation']:
         return a
 
     def optimization_at(self, circuit, index, op):
+        if isinstance(op.gate, ops.MatrixGate) and len(op.qubits) == 1:
+            return None
```
dstrain115: Won't this leave Matrix gates alone?  Don't we need to transform them to PhasedXZ's or something to serialize them correctly?

mpharrigan: see the test, but single qubit gates are compiled explicitly as the second step in `optimized_for_sycamore`.

mpharrigan: I've verified that this change does indeed let you run single qubit matrix gates through quantum engine

dstrain115: The test just tests that the number of moments is 2.  Can we add an extra line to verify that it serializes correctly too?

mpharrigan: added

Strilanc:
I'm testing the new API call to hit the update branch button.

mpharrigan:
Thwarted by #2770 !

PR #2767, opened 2020-02-13
get_engine_sampler

mpharrigan:
Remove tons of boilerplate


cirq/google/engine/engine_sampler_test.py
```diff
@@ -53,3 +55,14 @@ def test_engine_sampler_engine_property():
                                       processor_id='tmp',
                                       gate_set=cg.XMON)
     assert sampler.engine is engine
+
+
+def test_get_engine_sampler(monkeypatch):
+    monkeypatch.setenv('GOOGLE_CLOUD_PROJECT', 'myproj')
```
vtomole: Today i learned about monkeypatch

cirq/google/gate_sets.py
```diff
@@ -83,5 +83,9 @@
         MEASUREMENT_DESERIALIZER,
     ],
 )
-
 document(XMON, """Gate set for XMON devices.""")
+
+NAMED_GATESETS = {
```
vtomole: Why this name? Are other gates sets not named?

mpharrigan: named as in string names. The others are just python objects

vtomole: Nit: maybe we want `STRING_NAMED_GATESETS'

mpharrigan:
hmm I'll have to dig into the mac ci

cirq/google/engine/engine_sampler.py
```diff
@@ -64,3 +64,26 @@ def run_sweep(
...
+        gate_set_name: One of ['sqrt-iswap', 'syc'].
+            See `cirq.google.NAMED_GATESETS`.
+    """
+    try:
+        gate_set = gate_sets.NAMED_GATESETS[gate_set_name]
+    except KeyError:
+        raise ValueError(f"Please use one of the following gateset names: "
+                         f"{sorted(gate_sets.NAMED_GATESETS.keys())}")
+
+    return engine.Engine(project_id=os.environ['GOOGLE_CLOUD_PROJECT'],
```
maffoo: FWIW, in situations I like to add an arg so the environment dict can be injected in tests, rather than monkey-patching:
```python
def get_engine_sampler(processor_id: str, gate_set_name: str, env: Dict[str, str] = os.environ):
    ...
    return engine.Engine(project_id=env['GOOGLE_CLOUD_PROJECT'], ...)

# in tests:
get_engine_sampler('foo', 'gates', {'GOOGLE_CLOUD_PROJECT': 'my-project'})
```
Some people are not too opposed to monkey-patching, but it always scares me :-)

mpharrigan: 1. Actually yeah, I should add an argument that defaults to the environment variable
2. I'm going to leave the monkey patching test in as well. I think the primary utility of this method is dispatching to the environment variable for project id so people can share code that doesn't have their personal project id hardcoded

mpharrigan: added!

mpharrigan:
Anyone else have an opinion on this?

cirq/google/engine/engine_sampler.py
```diff
@@ -64,3 +64,34 @@ def run_sweep(
     @property
     def engine(self) -> 'cirq.google.Engine':
         return self._engine
+
+
+def get_engine_sampler(processor_id: str, gate_set_name: str,
```
kevinsung: I think that instead of taking a gate set name and having the NAMED_GATESETS dictionary, we should just take the gate set directly. There's no reason to add another layer of translation.

mpharrigan: at the risk of angering @dstrain115 by invoking the following argument: it's more pythonic to expose enum's using strings, especially in a user-facing api. See e.g. many numpy functions. In this case there are several advantages

 - the gateset can be specified by command line arg and/or environment variable
 - you don't have to find the correct import to find the gateset objects
 - gate_set_name can live comfortably in a pandas dataframe or database table

mpharrigan: I thought we were discouraging (if not deprecating) `engine.run` as it does not follow the sampler api

dabacon: The sampler api isn't flexible enough IMO to replace `run`.  It requires that you stick any extra information about your run into the sampler constructor, so if you really do care about this information you will end up creating samplers all over the place.

kevinsung: Matt gave some good reasons to make it a string, so I'm fine with that. But we should be consistent.

cirq/google/gate_sets.py
```diff
@@ -83,5 +83,9 @@
...
+    'sqrt-iswap': SQRT_ISWAP_GATESET,
```
dabacon, line 89: Strange that these names don't match the gate set names (which are used when they are serialized).

cirq/google/engine/engine_sampler.py
dabacon: the only good reason I can think of not to use a string is that it makes it less extensible, but since there are only a set number of blessed gate sets, this seems fine to me.

cirq/google/engine/engine_sampler.py
```diff
@@ -64,3 +64,34 @@ def run_sweep(
...
+                       project_id: Optional[str] = None) \
+        -> 'cirq.google.QuantumEngineSampler':
+    """Get an EngineSampler assuming some sensible defaults.
+
+    This uses the environment variable GOOGLE_GLOUD_PROJECT for the Engine
+    project_id, unless set explicitly. We assume you are using ProtoVersion.V2.
```
dabacon, line 75: I think we can remove comment about protoversion since no one should be using v1.

cirq/google/engine/engine_sampler.py
```diff
@@ -64,3 +64,34 @@ def run_sweep(
...
+
+    Args:
+        processor_id: Engine processor ID (from cloud console)
```
dabacon, line 78: or from `Engine.list_processors`.  Also nit: period to be consist with other args.

cirq/google/gate_sets.py
mpharrigan: changed

cirq/google/engine/engine_sampler.py
mpharrigan: done

cirq/google/engine/engine_sampler.py
mpharrigan: done

mpharrigan:
I merged master and noted there's a new global variable `GOOGLE_GATESETS` that was missing the new "fsim" gateset; which I also added to this function

mpharrigan:
We good?

PR #2768, opened 2020-02-13
Added QAOA and Rabi Demo notebooks.

MichaelBroughton:
Added introductory notebooks for Rabi and QAOA.


mpharrigan:
compare and contrast with `examples/qaoa.py`? and to some extent `examples/qubit_characteriations_example.py`?

MichaelBroughton:
These are for upcoming workshop demos at APS. They are really just simplified versions of both of these examples that have more explicit walkthroughs of the code and are in notebook form.

PR #2770, opened 2020-02-14
Serialization for PhysicalZ gates

dstrain115:
- ZPowGates with the PhysicalZTag now get the type "physical_z"
- Since a bunch of functions used gates instead operation:
  -- modified can_serialize_predicate to take operations
  -- modified op_deserialize with op_wrapper that can add tags back on

### DIFF
--- a/cirq/google/common_serializers.py
+++ b/cirq/google/common_serializers.py
@@ -393,28 +393,28 @@ SQRT_ISWAP_SERIALIZERS = [
         gate_type=ops.FSimGate,
         serialized_gate_id='fsim_pi_4',
         args=[],
-        can_serialize_predicate=(
-            lambda e: _near_mod_2pi(cast(ops.FSimGate, e).theta, np.pi / 4) and
-            _near_mod_2pi(cast(ops.FSimGate, e).phi, 0))),
+        can_serialize_predicate=(lambda e: _near_mod_2pi(
+            cast(ops.FSimGate, e.gate).theta, np.pi / 4) and _near_mod_2pi(
+                cast(ops.FSimGate, e.gate).phi, 0))),
     op_serializer.GateOpSerializer(
         gate_type=ops.ISwapPowGate,
         serialized_gate_id='fsim_pi_4',
         args=[],
         can_serialize_predicate=(lambda e: _near_mod_n(
-            cast(ops.ISwapPowGate, e).exponent, -0.5, 4))),
+            cast(ops.ISwapPowGate, e.gate).exponent, -0.5, 4))),
     op_serializer.GateOpSerializer(
         gate_type=ops.FSimGate,
         serialized_gate_id='inv_fsim_pi_4',
         args=[],
-        can_serialize_predicate=(
-            lambda e: _near_mod_2pi(cast(ops.FSimGate, e).theta, -np.pi / 4) and
-            _near_mod_2pi(cast(ops.FSimGate, e).phi, 0))),
+        can_serialize_predicate=(lambda e: _near_mod_2pi(
+            cast(ops.FSimGate, e.gate).theta, -np.pi / 4) and _near_mod_2pi(
+                cast(ops.FSimGate, e.gate).phi, 0))),
     op_serializer.GateOpSerializer(
         gate_type=ops.ISwapPowGate,
         serialized_gate_id='inv_fsim_pi_4',
         args=[],
         can_serialize_predicate=(lambda e: _near_mod_n(
-            cast(ops.ISwapPowGate, e).exponent, +0.5, 4))),
+            cast(ops.ISwapPowGate, e.gate).exponent, +0.5, 4))),
 ]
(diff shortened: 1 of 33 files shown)


cirq/google/common_serializers.py
```diff
@@ -118,7 +126,24 @@ def _near_mod_2(e, t, atol=1e-8):
...
+        args=[
+            op_serializer.SerializingArg(
+                serialized_name='half_turns',
+                serialized_type=float,
+                gate_getter='exponent',
+            ),
+            op_serializer.SerializingArg(
+                serialized_name='type',
+                serialized_type=str,
+                gate_getter=lambda x: 'physical_z',
```
maffoo: I feel like it'd be conceptually simpler to have just one serializer for `ZPowGate` that creates correct type here, e.g.:
```python
op_serializer.SerializingArg(
    serialized_name='type',
    serialized_type=str,
    op_getter=lambda op: 'physical' if PhysicalZTag() in op.tags else 'virtual_propagates_forward'
)
```
Obviously this would require changing the arg serializers to allow adding "getters" on the op instead of just the gate, but you've already done this here for `GateOpSerializer` and it seems like we might want it for the args as well.

It also looks like `Operation` does not currently define a `tags` property, but I think that'd be useful so users do not have to always check for TaggedOperation before looking at tags (we did a similar thing with the `gate` property, which is defined on `Operation` even though it could be `None`).

cirq/google/common_serializers.py
```diff
@@ -233,7 +257,7 @@ def _near_mod_2(e, t, atol=1e-8):
                                          gate_getter='phase_exponent'),
         ],
         can_serialize_predicate=lambda x: _near_mod_2(
-            cast(ops.PhasedXPowGate, x).exponent, 1)),
+            cast(ops.PhasedXPowGate, x.gate).exponent, 1)),
```
maffoo: Maybe we can change the arg name to `op` in these lambdas (same thing applies throughout this PR). Could also do this in a future change.

cirq/google/common_serializers_test.py
```diff
@@ -438,11 +438,33 @@ def test_serialize_z(gate, half_turns):
...
+        },
+        'args': {
+            'half_turns': {
+                'arg_value': {
+                    'float_value': half_turns
+                }
+            },
+            'type': {
+                'arg_value': {
+                    'string_value': 'physical_z'
```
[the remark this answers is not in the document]
maffoo: sorry, typo, I meant to suggest calling this `physical`.

cirq/google/op_serializer.py
```diff
@@ -65,17 +65,17 @@ def __init__(self,
...
+                 lambda x: True):
         """Construct the serializer.
 
         Args:
             gate_type: The type of the gate that is being serialized.
             serialized_gate_id: The string id of the gate when serialized.
-            can_serialize_predicate: Sometimes a gate can only be serialized for
-                particular gate parameters. If this is not None, then
+            can_serialize_predicate: Sometimes an Operation can only be
+                serialized for particular parameters. If this is not None, then
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/common_serializers_test.py
```diff
@@ -438,11 +438,33 @@ def test_serialize_z(gate, half_turns):
...
+        'qubits': [{
+            'id': '1_2'
+        }]
+    }
 
 
 @pytest.mark.parametrize(('axis_half_turns', 'half_turns'), [
     (0.25, 0.25),
     (0, 0.25),
+    (0, 0.25),
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/common_serializers_test.py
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/common_serializers.py
dstrain115: Done.  Changed gate_getter to op_getter.

I'll consider adding tags to Operation in a separate PR since it is distinct from this change.

cirq/google/common_serializers.py

cirq/google/op_serializer.py
```diff
@@ -136,31 +136,31 @@ def to_proto(
...
-    def _value_from_gate(self, gate: 'cirq.Gate', arg: SerializingArg
+    def _value_from_gate(self, op: 'cirq.Operation', arg: SerializingArg
                         ) -> Optional[arg_func_langs.ARG_LIKE]:
         value = None
-        gate_getter = arg.gate_getter
-
-        if isinstance(gate_getter, str):
-            value = getattr(gate, gate_getter, None)
+        op_getter = arg.op_getter
+        gate = op.gate
```
maffoo: nit: could move this line inside the `if isinstance(op_getter, str):` block.

PR #2772, opened 2020-02-15
Add possible implementation of aligning Z, 1-qubit, and 2-qubit gates optimizer

KevinVillela:
This is a bit of a strawman implementation of the optimizer that separates moments into Z ops, other 1-qubit ops, and 2-qubit ops. I tried for a while to find a good straightforward algorithm to do this, but I almost immediately ran into a situation where said algorithm would break down (see test files for examples).

On the other hand, I found that using Simulated Annealing resulted in very good results that only took a few seconds to compute (and this could be sped up drastically). I know that it seems like overkill for what we want, so I thought I'd send a PR so we can discuss it here. Thanks!

### DIFF
--- a/cirq/google/optimizers/align_interactions.py
+++ b/cirq/google/optimizers/align_interactions.py
@@ -1,0 +1,282 @@
+import random
+from enum import Enum
+from typing import List, Set
+
+import cirq
+from cirq import circuits, ops, InsertStrategy
+
+from cirq.google.line.placement import optimization
+
+
+class AlignInteractions():
+    """Aligns two-qubit gates, single-qubit gates, and Z-gates into their own
+    moments. Note that this optimizer does not guarantee that the resulting
+    circuit will be faster - in fact, it may result in a circuit with more
+    moments than the original. Instead, it attempts to split Z operations, non-Z
+    operations, and 2-qubit operations into their own homogeneous moments, with
+    the restriction that single qubit gates cannot move past 2 qubit gates.
+
+    This function's algorithm is non-trivial because it boils down to the
+    following Constraint Satisfiability Problem: given a set of operations which
+    each have ranges where they can be applied, how can they be reorganized in
+    such a way as to minimize the number of heterogeneous and total moments? See
+    the tests for some cases that quickly become tricky without using a full CSP
+    solver.
+
+    Because of the intractable complexity of brute-forcing CSPs, we instead use
+    Simulated Annealing to search for an optimized circuit. An adjacent circuit
+    is defined as one where a single operation is moved to a brand new moment or
+    swapped with an existing operation in another moment.
+
+    Args:
+        num_repetitions: The number of repetitions per temperature in the
+        Simulated Annealing algorithm. The lower this number is, the quicker the
+        algorithm will run, but the greater chance that it ends up with a
+        non-optimal or partially heterogeneous circuit. Usually, the default is
+        fine.
+        random_state: A seed for random number generation.
+       """
+
+    num_repetitions: int
+    random_state: cirq.value.RANDOM_STATE_LIKE
+    max_num_moments: int  # The largest the circuit is allowed to grow to.
+
+    def __init__(self,
+                 *,
+                 num_repetitions=10,
+                 random_state: cirq.value.RANDOM_STATE_LIKE = None) -> None:
+        self.random_state = cirq.value.parse_random_state(random_state)
+        self.num_repetitions = num_repetitions
+
+    def optimize_circuit(self, circuit: circuits.Circuit) -> None:
+        """Optimize the circuit inline according to the algorithm specified in
+        the class comment.
+
+        This algorithm is currently slow - it takes about 5 seconds to run on an
+        8x8 circuit. It could easily be optimized in the following ways:
+
+          - Currently, circuits are used as the model during the simulated
+            annealing, which causes a lot of wasted effort around copying
+            circuits and moving operations around. Converting the initial
+            circuit to a simpler Data Structure before optimizing and then back
+            after would save a lot of time.
+          - We can detect when we are at a local maximum by checking if any
+            operations can be moved to a location that decreases our score. This
+            could be implemented into our search and could potentially be used
+            to end the search early.
+
+        Args:
+            circuit: The circuit to optimize, which will be mutated.
+        """
+        self.max_num_moments = len(circuit) * 3
+        solution = optimization.anneal_minimize(
+            initial=circuit.copy(),
+            cost_func=self.cost_func,
+            move_func=self.move_func,
+            random_sample=self.random_state.random_sample,
+            repeat=self.num_repetitions)
+
+        del circuit[:]
+        for moment in solution:
+            circuit.append(moment)
+
+    def trace_func(self, current: cirq.Circuit, temp: float, cost: float,
+                   accept: float, decision: bool):
+        """A trace function that can be supplied to the annealing function that
+        prints debug information."""
+        print(f"current: {len(current)}, temp: {temp}, cost: {cost}, "
+              f"accept: {accept}, decision: {decision}")
+        if decision:
+            print(current)
+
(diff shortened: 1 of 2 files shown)


cirq/google/optimizers/align_interactions_test.py
```diff
@@ -0,0 +1,284 @@
+from typing import Callable, List, Dict
```
vtomole: Add copyright

cirq/google/optimizers/align_interactions.py
```diff
@@ -0,0 +1,264 @@
+import random
```
vtomole: Add copyright

mpharrigan:
Why is it so hard? Does the problem become simpler if you're assured that someone has run the single qubit merger optimizer first? Right now that merges all adjacent single qubit rotations into at most one PhX and Z. As a side issue; we should make that optimizer output the new PhXZ gate.

cirq/google/optimizers/align_interactions_test.py
KevinVillela: Done.

cirq/google/optimizers/align_interactions.py
KevinVillela: Done.

KevinVillela:
> Why is it so hard? Does the problem become simpler if you're assured that someone has run the single qubit merger optimizer first? Right now that merges all adjacent single qubit rotations into at most one PhX and Z. As a side issue; we should make that optimizer output the new PhXZ gate.

I don't think so. Assuming there are only two types of gates, X and SWAP, and X's can't go past SWAPs, then I think the problem is still a CSP, and the difficulty lies in finding the minimum number of moments. Take the following example:
```
              ┌───────────┐               ┌──────────┐
0: ────────────iSwap──────────X───────X────iSwap─────────────X───────
               │                           │
1: ───iSwap────iSwap──────────X───────X────┼────iSwap────────iSwap───
      │                                    │    │            │
2: ───iSwap────iSwap──────────X───────X────┼────iSwap────────iSwap───
               │                           │
3: ────────────┼────iSwap─────────────X────iSwap─────────────────────
               │    │
4: ───X────────┼────┼────X────X───────X────iSwap─────────────X───────
               │    │                      │
5: ────────────┼────iSwap─────X────────────┼─────────────────X───────
               │                           │
6: ───iSwap────┼────X─────────────X────────┼─────────────────iSwap───
      │        │                           │                 │
7: ───iSwap────iSwap───────────────────────iSwap─────────────iSwap───
              └───────────┘               └──────────┘
```
A solution is 
```
              ┌──────────┐                           ┌──────────┐
0: ────────────iSwap─────────X───────X─────────────────────iSwap────X───
               │                                           │
1: ───iSwap────iSwap─────────X───X───────────iSwap────iSwap┼────────────
      │                                      │        │    │
2: ───iSwap────iSwap─────────X───────X───────iSwap────iSwap┼────────────
               │                                           │
3: ────────────┼────iSwap────────────X─────────────────────iSwap────────
               │    │
4: ────────────┼────┼────────X───X───X───X───iSwap──────────────────X───
               │    │                        │
5: ────────────┼────iSwap────X───────────X───┼──────────────────────────
               │                             │
6: ───iSwap────┼─────────────────────X───X───┼────────iSwap─────────────
      │        │                             │        │
7: ───iSwap────iSwap─────────────────────────iSwap────iSwap─────────────
              └──────────┘                           └──────────┘
```

In this example, some of the operations move forward (like the Xs on the left), and some of them move backward (like the SWAPs on the right) - but how each gate moves depends on how the gates around it move. In other words, each gate has a range of locations that it can go, and each range depends on the other operations - and oftentimes, how those other operations move matters. With that mental model, I think the only way to know for sure is to try every single combination of ranges, which is way too slow.

mpharrigan:
but in your example, the single qubit gates haven't been merged. I'm pretty sure a strategy of:

 1. merge all single qubit gates
 2. greedily fill in alternating layers of single and two qubit moments
 3. drop any empty moments

works

KevinVillela:
That's what I was hoping as well, but it doesn't work for the following type of circuit:
```
0: ───X───×───────────
          │
1: ───────×───────────

2: ───────×───X───×───
          │       │
3: ───────×───────×───
```
which should optimize to
```
0: ───────X───×───
              │
1: ───────────×───

2: ───×───X───×───
      │       │
3: ───×───────×───
```
I've added a test to catch this explicitly.

maffoo:
@KevinVillela, the example circuit you showed would work with the simple algorithm outlined by @mpharrigan if step 2 runs in reverse, pushing gates toward later moments instead of earlier ones. This is generally how we do things internally because it's good to eliminate dead time after gates when qubits can decay before measurement.

KevinVillela:
Thanks for the response @maffoo! That does work for the above example, but I don't think it will for the following:
```
0: ─X──×──X───
       │
1: ────×─────

2: ────×─────
       │       
3: ─X──×─────
```
which is already optimized, but with the algorithm of greedily grabbing the single-qubit gates from the end, would result in:
```
0: ─X──×──X───────
       │
1: ────×──────────

2: ──────────×────
             │       
3: ───────X──×────
```

maffoo:
@KevinVillela, that's a good point; basically you need a way to decide if the last moment is going to be single qubit gates or two-qubit gates. It still seems to me that there should be a fairly simple way to decide this that keeps the spirit of what @mpharrigan proposed. Or if there's no fool-proof way to decide, the optimizer could just try both and pick the one that results in the shortest circuit.

KevinVillela:
@maffoo ah but the fun is only just beginning! I agree that in the examples above, we just need to decide what the last moment will be, and can even try both and take the better one. But one can imagine a circuit that contains multiple combinations of each of the above circuit patterns serially and in the middle of other operations. Even if we went with an approach of subdividing the circuit somehow and deciding on a strategy, each decision affects the rest of the circuit and will affect the strategies we want to use in turn.

maffoo:
The other assumption I am making is that we want the final circuit to consist of _alternating_ single- and two-qubit moments. I think this avoids the decision cascade you describe, once the "parity" is chosen (whether to end with a single- or two-qubit moment). We use circuits with such an alternating structure to calibrate the gates (as in the supremacy experiment) so I would expect higher fidelity with such a circuit structure. In addition, even if a particular circuit has no single-qubit gates in one of the nominal single-qubit moments, under various randomized compiling protocols we would insert gates in those slots anyway, and "fill in" the circuit while preserving the alternating moment structure. So while you're right that the general alignment problem may be quite complicated, I think having an optimizer that produces this simpler structure would be useful for near-term experiments.

KevinVillela:
Oh okay, thanks for the info, I was not aware of that. Couple of follow-up questions:
  - I agree that the algorithm for this should be pretty straightforward, and I'm happy to write it up. That being said, do we have any evidence of it being useful in practice? I worry that the increase in fidelity is going to be undercut by an increase in circuit size and time to execute.
  - Is the current algorithm (aligning 2-qubit and single-qubit gates and minimizing circuit length instead of alternating) useful for anything? If so, I could make it into a different strategy for this optimizer, or a different optimizer entirely. If not, I'll probably just delete it.

Thanks again for the responses!

mpharrigan:
> I worry that the increase in fidelity is going to be undercut by an increase in circuit size and time to execute

But keeping the moments homogeneous reduces the circuit duration because the two-qubit layers can be roughly half as long if they don't have to "wait" for simultaneous 1q gates that would otherwise be in that moment.

Another useful piece of context is that: after merging single qubit operations these circuits are going to be *pretty close* to having the alternating structure already. Most 2q gate compilations *need* flanking single qubit operations. People rarely want to do exactly ISWAP * CPHASE(pi/6)

KevinVillela:
Sounds good, thanks for the comments - I'll send out another PR in the next couple of days

KevinVillela:
Closing this in favor of https://github.com/quantumlib/Cirq/pull/2869, thanks all for the comments!

PR #2774, opened 2020-02-17
Increase tolerance in n_qubit_tomography_test'

fedimser:
I just got CI failure with this test failing with message 
```
assert np.linalg.norm(error_rho) < 0.05
AssertionError: assert 0.055035732500366044 < 0.05
```


fedimser:
I'm sorry, I accidentally added files from other branch as well.

PR #2776, opened 2020-02-18
Fix density matrix simulation of noise

### LINKED ISSUE #2567: Density matrix simulator simulates noise incorrectly
The code below constructs two circuits, each with one measurement gate that should have the same measurement results. Instead, one measures 0 and the other measures 1.
```
import cirq

q = cirq.LineQubit(0)
circuit1 = cirq.Circuit(cirq.measure(q))
circuit2 = circuit1 + cirq.I(q)

print('Circuit 1:')
print(circuit1)
print()
print('Circuit2:')
print(circuit2)
print()

simulator = cirq.DensityMatrixSimulator(noise=cirq.X)
result1 = simulator.run(circuit1, repetitions=10)
result2 = simulator.run(circuit2, repetitions=10)

print('Result 1:')
print(result1)
print()
print('Result 2:')
print(result2)
print()
```
Output:
```
Circuit 1:
0: ───M───

Circuit2:
0: ───M───I───

Result 1:
0=1111111111

Result 2:
0=0000000000
```

kevinsung:
Fixes #2567 .

The strategy is to add a flag to `_base_iterator` indicating whether all measurements are terminal. If so, then the iterator ignores operations on qubits that have already been measured.

For review, it may be more convenient to look at the first commit, with message "fix noise simulation". The second commit includes unrelated changes to the file that resulted from running the formatter.

This does not fix the second bug raised by @Ashalynd at #2567 . I will open a separate issue for that.

### DIFF
--- a/cirq/sim/density_matrix_simulator.py
+++ b/cirq/sim/density_matrix_simulator.py
@@ -11,19 +11,20 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 """Simulator for density matrices that simulates noisy quantum circuits."""
 
 import collections
 
-from typing import Dict, Iterator, List, Type, Union
+from typing import Dict, Iterator, List, TYPE_CHECKING, Type, Union
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


kevinsung:
Good point. I combined the flags, getting rid of `perform_measurements`.

Issue #942, opened 2018-09-19
Feature Request: Gate or method to measure qubits in Pauli eigenstate basis


babbush:
Often, people will want to run circuits repeatedly to measure the expectation value of a particular Pauli string. For instance, they might want to measure `<X2 Y3 Z4>`. Of course, one can use phase estimation to measure that operator directly and obtain 1 / eps scaling but that's not very NISQ. Typically, people will repeat the circuit 1 / eps^2 times and sample the expectation value. To do this, one needs a sequence of single qubit rotations. For instance, to measure X2 Y3 Z4, one must rotate qubit 2 around Y by pi / 2 and rotate qubit 3 around X by pi / 2, then measure qubits 1, 2 and 3.

I am tempted to make something like a "MeasurementRotationGate" which would take as input some specification of a Pauli string and then output the appropriate gates to perform rotations that allow one to sample in that basis. Thus, it would take X2 Y3 Z4 as input, and output a gate which performs the rotations and measurements mentioned above.

First, do people think this is a good idea? I think it would be very frequently used in the context of VQE. If so, how should the Pauli string input be specified? I would be partial to allowing either of the ways that we initialize QubitOperators in OpenFermion; i.e. we either allow them to provide an actual string like "X2 Y3 Z4" or they can give a tuple like ((2, "X"), (3, "Y"), (4, "Z")).

dabacon:
Craig and I have talked about something like this in the past. #941 will need something like this.  There are two places we need to discuss how this would work
1) on the cirq side where I think we can use something like #932 or related to make this more transparent than a tuple or a string (what is the 1 in X1...people should have access to the GridQubits so we want them to be able to use that)
2) What sort of feature we add to the api to support this. 

Another thing to note is that one can already create such a gate in Cirq that does this, by creating the appropriate composite gate.  @Strilanc comments?

Strilanc:
I think this falls under the purview of "displays", which would be special inspection gates that basically say "tell me an expectation value at this point in the circuit". Naively, every display requires its own set of runs of the circuit. But in a simulator you can fill them all in in one pass, assuming there are no measurements present.

dabacon:
Since we've abandoned displays, we should revisit the main gist of this request, which is an easy way to measure in different Pauli eigenstate basis.  

This could be

1. A composite gate which takes in the PauliString and a key and is a gate and that decompose spits out the correct rotations followed by measurement.
[...]

PR #2781, opened 2020-02-20
Observable Measurement

mpharrigan:
Quantum applications involve measuring PauliStrings after preparing a state according to a Circuit. Some PauliStrings can be measured simultaneously. This PR is for an API that sits atop the sampler API to help with these cases.

We key each run of a sampler interface by MeasurementSpec and the sampled results are contained in a BitstringAccumulator. We implement infrastructure to construct/organize a collection of MeasurementSpecs from a high-level description of a tomographic-like experiment. BitstringAccumulators can be queried for measurement results and hook into a batching and stopping criteria framework to control the execution of jobs.

observables - data

By structuring your experiment in this way, we can automatically implement some pretty good tricks:
 - Measurement basis change rotations done via sweeps for better batching
 - Chunking of submission if you want more than (max_shots) from quantum engine
 - Checkpointing so you don't loose your work halfway through a job
 - Measuring to a variance tolerance rather than a pre-specified number of repetitions
 - Readout error symmetrization and mitigation

## Examples

Please check out the examples! I tried to make them interesting and informative of how to use this feature.

### Readout correction
om-ro

### VQE-ish experiment

om-vqe

### Process tomography
```
[[ 0.537+0.j     0.455-0.036j -0.418-0.j     0.416+0.002j]
 [ 0.455+0.036j  0.463-0.j    -0.421+0.j     0.418+0.j   ]
 [-0.418+0.j    -0.421-0.j     0.539+0.j    -0.381-0.037j]
 [ 0.416-0.002j  0.418-0.j    -0.381+0.037j  0.461-0.j   ]]
------------------------------
[[ 0.5  0.5 -0.5  0.5]
 [ 0.5  0.5 -0.5  0.5]
 [-0.5 -0.5  0.5 -0.5]
 [ 0.5  0.5 -0.5  0.5]]
Frobenius norm: 0.3112813656088466
```

As a special treat, some of the above is real data from a real device.

This PR requires #2766 and #2767. I'll rebase once those are in. I'm still working on adding some more tests and/or documentation, but figured this is in a good enough state for people to start looking at it.

### For follow up work

 - complimentary code for *simulating* a batch of observables
 - chunking over parameters (currently chunk over number of repetitions)

### DIFF
--- a/cirq/work/observable_measurement.py
+++ b/cirq/work/observable_measurement.py
@@ -1,0 +1,708 @@
+# Copyright 2020 The Cirq developers
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import abc
+import dataclasses
+import itertools
+import os
+import tempfile
+import warnings
+from typing import Optional, Union, Iterable, Dict, List, Tuple, \
+    TYPE_CHECKING, Set, Sequence
+
+import numpy as np
+import pandas as pd
+import sympy
+
+from cirq import circuits, study, ops, value, protocols
+from cirq.work.observable_grouping import group_settings_greedy
+from cirq.work.observable_measurement_data import BitstringAccumulator
+from cirq.work.observable_settings import InitObsSetting, observables_to_settings, zeros_state, \
+    MeasurementSpec
+
+if TYPE_CHECKING:
+    import cirq
+
+
+def _with_parameterized_layers(circuit: 'cirq.Circuit',
+                               qubits: Sequence['cirq.Qid'],
+                               no_initialization: bool,
+                               ) -> 'cirq.Circuit':
+    """Return a copy of the input circuit with a parameterized single-qubit
+    rotations.
+
+    These rotations flank the circuit: the initial two layers of X and Y gates
+    are given parameter names "{qubit}-Xi" and "{qubit}-Yi" and are used
+    to set up the initial state. If `no_initialization` is set to True,
+    these two layers of gates are omitted.
+
+    The final two layers of X and Y gates are given parameter names
+    "{qubit}-Xf" and "{qubit}-Yf" and are use to change the frame of the
+    qubit before measurement, effectively measuring in bases other than Z.
+    """
+    x_beg_mom = ops.Moment([
+        ops.X(q) ** sympy.Symbol(f'{q}-Xi')
+        for q in qubits
+    ])
+    y_beg_mom = ops.Moment([
+        ops.Y(q) ** sympy.Symbol(f'{q}-Yi')
+        for q in qubits
+    ])
+    x_end_mom = ops.Moment([
+        ops.X(q) ** sympy.Symbol(f'{q}-Xf')
+        for q in qubits
+    ])
+    y_end_mom = ops.Moment([
+        ops.Y(q) ** sympy.Symbol(f'{q}-Yf')
+        for q in qubits
+    ])
+    meas_mom = ops.Moment([ops.measure(*qubits, key='z')])
+    if no_initialization:
+        total_circuit = circuit.copy()
+    else:
+        total_circuit = circuits.Circuit([x_beg_mom, y_beg_mom])
+        total_circuit += circuit.copy()
+    total_circuit.append([x_end_mom, y_end_mom, meas_mom])
+    return total_circuit
+
+
+class StoppingCriteria(abc.ABC):
+    """An abstract object that queries a BitstringAccumulator to figure out
+    whether that `meas_spec` is complete."""
+
+    @abc.abstractmethod
+    def more_repetitions(self, accumulator) -> int:
+        """Return the number of additional repetitions to take.
+
+        StoppingCriteria should be respectful and have some notion of a
+        maximum number of repetitions per chunk.
+        """
+
+
+@dataclasses.dataclass(frozen=True)
+class VarianceStoppingCriteria(StoppingCriteria):
+    """Stop sampling when average variance per term drops
+    below a variance bound."""
+    variance_bound: float
+    repetitions_per_chunk: int = 10_000
+
+    def more_repetitions(self, accumulator: BitstringAccumulator):
+        if len(accumulator.bitstrings) == 0:
+            return self.repetitions_per_chunk
+
+        cov = accumulator.covariance()
+        o = np.ones(cov.shape[0])
+        sum_variance = o @ cov @ o.T
+        var_of_the_e = sum_variance / len(accumulator.bitstrings)
+        vpt = var_of_the_e / len(o)
+
+        done = vpt <= self.variance_bound
+        if done:
+            return 0
+        return self.repetitions_per_chunk
+
+
+@dataclasses.dataclass(frozen=True)
+class RepetitionsStoppingCriteria(StoppingCriteria):
+    """Stop sampling when the number of repetitions has been reached."""
+    total_repetitions: int
+    repetitions_per_chunk: int = 10_000
+
+    def more_repetitions(self, accumulator: BitstringAccumulator):
+        done = accumulator.n_repetitions
+        todo = self.total_repetitions - done
+        if todo <= 0:
+            return 0
+
+        to_do_next = min(self.repetitions_per_chunk, todo)
+        return to_do_next
+
+
+_PAULI_TO_PARAM_VAL = {
+    (ops.X, False): (0, -1 / 2),
+    (ops.X, True): (0, +1 / 2),
+    (ops.Y, False): (1 / 2, 0),
+    (ops.Y, True): (-1 / 2, 0),
+    (ops.Z, False): (0, 0),
+    (ops.Z, True): (1, 0),
+}
+
+_STATE_TO_PARAM_VAL = {
+    value.KET_PLUS: (0, +1 / 2),
+    value.KET_MINUS: (0, -1 / 2),
+    value.KET_IMAG: (-1 / 2, 0),
+    value.KET_MINUS_IMAG: (+1 / 2, 0),
+    value.KET_ZERO: (0, 0),
+    value.KET_ONE: (1, 0),
+}
+
+
+def _get_params_for_setting(setting: InitObsSetting,
+                            flips: Iterable[bool],
+                            qubits: Sequence['cirq.Qid'],
+                            no_initialization: bool,
+                            ) -> 'cirq.ParamDictType':
+    """Return the parameter dictionary for the given setting.
+
+    This must be used in conjunction with a circuit generated by
+    `_with_parameterized_layers`. `flips` (used for readout symmetrization)
+    should be of the same length as `qubits` and will modify the parameters
+    to also include a bit flip (`X`). Code responsible for running the
+    circuit should make sure to flip bits back prior to analysis.
+    """
+    params = {}
+    for qubit, flip in itertools.zip_longest(qubits, flips):
+        if qubit is None or flip is None:
+            raise ValueError("`qubits` and `flips` must be equal length")
+        # When getting the one-qubit state / observable for this qubit,
+        # you may be wondering what if there's no observable specified
+        # for that qubit. We mandate that by the time you get to this stage,
+        # each _max_setting has
+        # weight(in_state) == weight(out_operator) == len(qubits)
+        # See _pad_setting
+        pauli = setting.observable[qubit]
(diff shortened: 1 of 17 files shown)


vtomole:
> 3k+ lines of code and 45 files changed.

Let's keep this one open but open up multiple PRs that appropriately break this one into bite sized chunks if possible. Makes it easier for me to parse/review.

mpharrigan:
Yes, sorry. A lot of the apparent churn is from the two linked PRs. It should simplify drastically after those are merged. I can still split it up into two PRs

 1. adding TensorProductState and friends
 2. observable measurement

I decided not to do this because

 1. The splittable components are already pretty split: all the states changes are in states.py; all the observable measurement changes are in observable_measurement.py. The only exception is the rote changes to JSON serialization listing
 2. states by themselves are pretty boring and not very well motivated

However, because of (1) I can easily split it up if people think that would be better

cirq/google/engine/engine_sampler.py
```diff
@@ -64,3 +64,26 @@ def run_sweep(
     @property
     def engine(self) -> 'cirq.google.Engine':
         return self._engine
+
+
+def get_engine_sampler(processor_id: str, gate_set_name: str) \
```
Strilanc: It should be possible to override the environment variable by specifying an argument here.

cirq/protocols/json_test_data/BitstringAccumulator.repr
```diff
@@ -0,0 +1,14 @@
+cirq.BitstringAccumulator(
+    meas_spec=cirq.MeasurementSpec(
+        max_setting=cirq.InitObsSetting(
+            init_state=cirq.PLUS_Z(cirq.GridQubit(5, 2)) * cirq.MINUS_X(cirq.GridQubit(7, 8)),
+            observable=cirq.Z(cirq.GridQubit(5, 2)) * 1),
```
Strilanc: The `* 1` here is a bit odd. Just the repr unconditionally including it I guess.

cirq/sim/mux.py
```diff
@@ -251,11 +264,14 @@ def final_density_matrix(
...
 
     """
-    initial_state_like = None
-    if not isinstance(initial_state, int):
-        initial_state_like = np.asarray(initial_state, dtype=dtype)
+    qubit_order = ops.QubitOrder.as_qubit_order(qubit_order)
+
+    if isinstance(initial_state, int):
+        pass
+    elif isinstance(initial_state, value.TensorProductState):
```
Strilanc: This logic is liable to repeated a bunch of times. Consider extracting a method like `initial_state_like_to_state_vector`.

cirq/value/states.py
```diff
@@ -0,0 +1,365 @@
...
+        # Descendants should be singletons determined solely by the class name.
+        # Otherwise, you must override this method.
+        return hash(self.__class__.__name__)
+
+
+@dataclass(frozen=True)
+class TensorProductState:
+    """A quantum state that is a tensor product of one qubit states.
+
+    For example, the |00> state is `cirq.PLUS_Z(q0) * cirq.PLUS_Z(q1)`.
```
Strilanc: I'm a bit worried about the names here. There's an apparent similarly, but large difference, between `+cirq.Z(q0) * +cirq.Z(q1)` and `cirq.PLUS_Z(q0) * cirq.PLUS_Z(q1)`. It's also not the case that cirq.MINUS_Z(q) is equal to -cirq.PLUS_Z(q).

(In somewhat related news I foolishly initially thought that PLUS_Z was referring to the plus state |+> with respect to the Z basis.)

I don't have a better suggestion at the moment, just this worry about these names leading to predictable confusions.

Some ideas:

- Prefix the names with KET_ to more clearly mark them as states, and break the "plus_x equals -minus_x" connotation.
- Switch to a completely different strategy, like a dictionary from qubits to plus or minus cirq.X/Y/Z.
- Have a method like "PlusEigenstate" and do stuff like `PlusEigenstate(cirq.X(q)) * PlusEigenstate(-cirq.Z(q))`
- Give Pauli gates a helper method `.eigen_ket()` and do stuff like `cirq.X(q).eigen_ket(+1) * cirq.Z(q).eigen_ket(-1)`.
- Generalize to StabilizerState and take a list of stabilizers (with the single qubit stabilizers as a special case).
- Use names like `KET_ZERO`, `KET_PLUS`, and `KET_IMAG`.

cirq/value/states.py
```diff
@@ -0,0 +1,365 @@
...
+
+    def __iter__(self):
+        yield from self.states.items()
+
+    def __len__(self):
+        return len(self.states)
+
+    def __eq__(self, other):
+        if not isinstance(other, TensorProductState):
+            return False
```
Strilanc: return NotImplemented instead of False, to allow later classes to override the behavior. If you use `cirq.testing.EqualsTester` it should catch this.

cirq/value/states.py
```diff
@@ -0,0 +1,365 @@
...
+
+    def stabilized_by(self):
+        from cirq import ops
+        return 1, ops.Y
+
+
+class _MinusY(_Named1QState):
+
+    def __str__(self):
+        return '-Y'
```
Strilanc: These methods could be attributes of a concrete base class, instead of implementations of abstract methods of an abstract base class.

cirq/value/states.py
```diff
@@ -0,0 +1,365 @@
...
+    PLUS_Y,
+    MINUS_Y,
+    PLUS_Z,
+    MINUS_Z,
+]
+document(PAULI_STATES,
+         """All one-qubit states stabalized by the pauli operators.""")
+
+
+def zeros_state(qubits: Iterable['cirq.Qid']) -> TensorProductState:
```
Strilanc: Is there a context where you would do this instead of using `initial_state=0`?

cirq/value/states.py
```diff
@@ -0,0 +1,365 @@
...
+    def stabilized_by(self):
+        from cirq import ops
+        return 1, ops.X
+
+
+class _MinusX(_Named1QState):
+    def __str__(self):
+        return '-X'
+
+    def __repr__(self):
```
Strilanc: Make sure `cirq.H * cirq.MINUS_X` returns `cirq.MINUS_Z`. Also all other combinations of gates and states.

... just kidding. 😝

cirq/google/engine/engine_sampler.py
mpharrigan: addressed in #2767 ; I'll rebase once that's merged

cirq/protocols/json_test_data/BitstringAccumulator.repr
mpharrigan: There are several dirty things going on here.

1. Multiplying by 1 turns a `Z` gate into a `PauliString`, which `observable` must be
2. This isn't actually the result of calling `repr` on `BitstringAccumulator`. Rather, this is crafted to create an object that passes the serialization tests. I don't think there's anything that makes sure the saved `.repr` file matches the output from `repr`. 

Of course, expecting the repr to be a complete serialization of such a rich bag-of-data class such as this is a bit too much to ask. How do other classes initialized by numpy arrays work?

cirq/sim/mux.py
mpharrigan: Luckily this now exists (`cirq.to_valid_state_vector`). Will move everything there.

cirq/value/states.py
mpharrigan: This is modeled after how e.g. `cirq.X` and friends work. The alternative (as you suggest) would be a concrete base, e.g.

```python
KET_PLUS = _Named1QState(
    name='KET_PLUS',
    str='+X',
    state_vector = np.array([1,1])/np.sqrt(2),
    stabilized_by=(+1, ops.X)
)
```

big problem: stabilized by needs to import `cirq.ops` which would cause a circular import because of `value_equality`.

Small problem: we'd be initializing the arrays at import time. We don't do this anywhere else

Other consideration: It would be marginally easier with this scheme to input arbitrary or nonsensical values for `name`, `str`, etc. whereas it makes more sense to consider this as "enum-like"

cirq/value/states.py
mpharrigan: > Switch to a completely different strategy, like a dictionary from qubits to plus or minus cirq.X/Y/Z.

This is pretty close to what we're actually doing, but the dictionary is to `_Named1QState` instead. The thought is that it would be easier to extend this to other states of note (like SIC basis for tomography; magic states, ...)

> Prefix the names with KET_ to more clearly mark them as states, and break the "plus_x equals -minus_x" connotation.
> Use names like KET_ZERO, KET_PLUS, and KET_IMAG.

These two suggestions seem similar and I will do the latter

> Have a method like "PlusEigenstate" and do stuff like PlusEigenstate(cirq.X(q)) * PlusEigenstate(-cirq.Z(q))
> Give Pauli gates a helper method .eigen_ket() and do stuff like cirq.X(q).eigen_ket(+1) * cirq.Z(q).eigen_ket(-1).

These seem a little complicated and verbose and stem from a stance that the `Pauli` object is the be-all-end-all and all other objects should kneel before them. I think the concept of a single-qubit state with a special name is distinct enough that it can exist as a 1st class object (and you can get back to pauli's via the `stabilized_by` method).

cirq/value/states.py
mpharrigan: removing this for now

cirq/work/observable_measurement.py
```diff
@@ -0,0 +1,1317 @@
...
+    """Like `_max_weight_state`, but will raise a ValueError instead of
+    returning None for non-compatible states.
+    """
+    max_weight_state = _max_weight_state(states)
+    if max_weight_state is None:
+        raise ValueError("Grouping error for states {}".format(states))
+    return max_weight_state
+
+
+def group_settings_greedy(settings: Iterable[InitObsSetting]) \
```
tanujkhattar, line 168: I think the current design is not generic enough to enable different types of grouping strategies because:

- Different types of grouping strategies can introduce different unitary rotation gates at the end of the circuit for each group. 
- The post-processing for obtaining the results for individual `List[InitObsSetting]` from the single `InitObsSetting` can be different for every grouping strategy. 
- The final measurement strategy can be different for each grouping strategy. For eg: basis change + measuring a single qubit in Z basis (using `PauliMeasurementGate` proposed in   https://github.com/quantumlib/Cirq/issues/942#issuecomment-658933767) instead of measuring all qubits in the Z basis (current `cirq.measure(qubits)`) (I don't have an immediate example for this).

An example for the first two cases would be if we wish to implement the grouping strategies proposed in [arxiv:1907.09386](https://arxiv.org/pdf/1907.09386.pdf) or [arxiv:1908.06942](https://arxiv.org/pdf/1908.06942.pdf)

In order to fix these issues, I propose that we have a separate base class for defining the grouping strategy which performs the grouping, returns a `circuit_to_append` including measurement gates and consumes the results of measurement to perform the post-processing. The expectation corresponding to every original `InitObsSetting` can then be queried from this.

cirq/work/observable_measurement.py
```diff
@@ -0,0 +1,1317 @@
+# This file contains code modified from PyQuil.
```
tanujkhattar, line 1: I think there's a lot happening in this single file. It would be much easier to read/extend/test the different components if we break this file into a module/submodule with each of the different component like `grouping_strategy`, `stopping_criteria` etc. in separate files.

cirq/work/observable_measurement.py
```diff
@@ -0,0 +1,1317 @@
...
+        for qubit, pauli in observable.items():
+            if qubit in qubit_pauli_map:
+                if qubit_pauli_map[qubit] != pauli:
+                    return None
+            else:
+                qubit_pauli_map[qubit] = pauli
+    return ops.PauliString(qubit_pauli_map)
+
+
+def _max_weight_observable_strict(
```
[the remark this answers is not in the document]
mpharrigan: Good point! I'm sure I was using them at some point during the development of this feature :)

cirq/work/observable_measurement.py
mpharrigan: At the risk of making this PR even more untenably large, can we punt on more general term measurement strategies? Your proposal makes sense at a surface level but I think it would work better as a follow-up change with a fully-featured implementation of 1907.09386 instead of providing a set of api hooks we *think* will be sufficient for the follow on feature.

What do you think?

cirq/work/observable_measurement.py
mpharrigan: I can pull some of these things out

cirq/work/observable_measurement.py
tanujkhattar: The primary concern I have with your proposal is that the PR implementing 1907.09386 would also be responsible to refactor the work done here OR leave it in a legacy state (which is bad). The refactoring would also deprecate the current `measure_observables` interface, which will be an added effort. 
Let's continue the discussion in cirq sync.

mpharrigan:
@viathor @tanujkhattar @balopat and all: I've pushed a rebased version with some of the stuff split out into their own files.

For review purposes, I'm likely going to split the files into independent PRs based on their dependencies. I think it will be something like:

 - `observable_settings.py` and tests
 - `observable_grouping.py` and tests
 - `observable_measurement_data.py` and tests
 - `observable_measurement.py` and tests
 - the three examples (can split these up into their own PRs too)

PR #2782, opened 2020-02-20
Make auto_merge script use new update-branch API call

Strilanc:
- No more cookies!

### DIFF
--- a/dev_tools/auto_merge.py
+++ b/dev_tools/auto_merge.py
@@ -1,5 +1,4 @@
 import datetime
-import re
 import traceback
 from typing import Optional, List, Any, Dict, Set, Union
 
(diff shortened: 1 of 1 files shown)


dev_tools/auto_merge.py
```diff
@@ -464,57 +458,38 @@ def delete_comment(repo: GithubRepository, comment_id: int) -> None:
...
-        '.*</form>.*', html, re.DOTALL)
-    if form_guts is None:
-        if '(Logged out)' in html:
-            return CannotAutomergeError('Need a fresh :cookie:.')
-        raise RuntimeError(
-            'Failed to find update branch button. Html: {}.'.format(
-                html))
+    As of Feb 2020 this API feature is still in beta. Note that currently, if
+    you attempt to update branch when already synced to master, a vacuous merge
+    commit will be created.
```
mpharrigan: but this will be squashed out when the PR is merged, yes?

Strilanc: Right. Also the method is not called if it isn't needed. This is just a warning for someone looking at the method on its own.

mpharrigan:
Beyond the scope of this PR, but your error handling in this file is very interesting. It's sortof functional-style (but without an explicit `Maybe` type). Python typically uses try/catch exception handling

PR #2784, opened 2020-02-21
Fit depolarizing model to XEB result

kevinsung:


cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -14,14 +14,39 @@
...
 CrossEntropyPair = NamedTuple('CrossEntropyPair', [('num_cycle', int),
                                                    ('xeb_fidelity', float)])
 
 
+@dataclasses.dataclass
+class DepolarizingModel:
+    """A model of depolarizing noise.
+
+    Models the fidelity of a quantum process as an exponential decay of the form
+    f = S * p^d, where f is the fidelity, S is the coefficient, p is the
```
viathor: Generally speaking, it makes more sense to fit XEB results to A * p^d + B to account for SPAM errors which are independent of d.

viathor: Is "f" the average or entanglement fidelity?

kevinsung: I think it's neither. I rewrote the documentation to better describe S and p, and I'm no longer using the term "fidelity".

kevinsung: I thought the coefficient A already accounts for SPAM errors. @sboixo suggested this model to me so perhaps he can clarify.

sboixo: I guess fitting to A * p^d + B doesn't hurt, but B should be 0 in XEB. B is not 0 in RB, there is a difference there. 
As far as I can tell, we have been using the fit to A * p^d + B so far simply because we fit it as if it was RB for convenience.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -14,14 +14,41 @@
...
                                                    ('xeb_fidelity', float)])
 
 
+@dataclasses.dataclass
+class DepolarizingModel:
+    """A model of depolarizing noise.
+
+    A model of a depolarizing process that maps a density matrix ρ as
+
+        ρ → (S * p^d) ρ + (1 - S * p^d) I / D
```
viathor: The comment and formula mix two independent ideas. First, we use the depolarizing channel to model errors

    \rho -> p_eff \rho + (1 - p_eff) I / D

where p_eff is the effective decay constant of the channel. Then, we separate p_eff into SPAM errors which are depth-independent and circuit execution errors (decoherence and control errors) which exhibit exponential decay with depth. The separation works by fitting p_eff obtained for multiple circuit depths to

    p_eff = S * p^d

Note that we exploit the fact that the composition of depolarizing channel with decay constant a with depolarizing channel with decay constant b yields depolarizing channel with decay constant a*b.

Suggestion: perhaps describe the two ideas (how to model errors and how to separate SPAM and execution errors) separately.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -14,14 +14,41 @@
...
+
+    where I / D is the maximally mixed state, S and p are real numbers, and
+    d is the number of iterations of the process. For instance, in a cross
+    entropy benchmarking experiment, d would be the number of cycles, where a
+    cycle consists of a layer of single-qubit gates followed by a layer of
+    two-qubit gates.
+
+    Attributes:
+        coefficient: The value S in the above formula.
+        decay_constant: The value p in the above formula.
```
viathor: The two parameters (coefficient and decay_constant) can be given better description, e.g.

```
coefficient: depolarization constant for state preparation and measurement, i.e. S in the above formula.
decay_constant: depolarization constant for circuit execution, i.e. p in the above formula.
```

You could also name the two parameters better :-)

cirq/experiments/cross_entropy_benchmarking.py
viathor: Thanks, Sergio! FTR: Found a reference for this: section IV D in [1] and formula (35).

[1]: https://arxiv.org/pdf/1910.11333.pdf

cirq/experiments/cross_entropy_benchmarking.py
kevinsung: I've updated the documentation with your suggestions and correspondingly renamed the class to be specific to cross-entropy benchmarking.

cirq/experiments/cross_entropy_benchmarking.py
kevinsung: I've updated the descriptions and renamed the parameters to `spam_depolarization` and `cycle_depolarization`.

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -61,6 +101,38 @@ def plot(self, ax: Optional[plt.Axes] = None,
...
+        y = [fidelity for _, fidelity in self.data]
+
+        def f(d, S, p):
+            return S * p**d
+
+        params, covariance = scipy.optimize.curve_fit(f, x, y, p0=p0)
+
+        return CrossEntropyDepolarizingModel(spam_depolarization=params[0],
+                                             cycle_depolarization=params[1],
+                                             covariance=covariance)
```
viathor: Out of curiosity: Do you see non-zero off-diagonal entries in the covariance matrix in practice?

Issue #2698, opened 2020-01-22
Expand SWEEP_LIKE to work for {'t': [0, 1, 3]}


Strilanc:
Currently if you make a call like `sampler.sample(circuit, params={'t': range(5)})` it doesn't work, because it expects e.g. a list of dictionaries instead of a dictionary containing lists. For convenience we should just support both cases. Check if the input is a dictionary with list values (and possibly singletons too), and if so take the cartesian product of the various entries and turn it into a sweep.

Test cases:

```
assert cirq.to_sweeps({'t': [0, 2, 3]}) == (
    cirq.to_sweeps([{'t': 0}, {'t': 2}, {'t': 3}]}))

assert cirq.to_sweeps({'t': [0, 1], 's': [2, 3], 'r': 4}) == (
    cirq.to_sweeps([
        {'t': 0, 's': 2, 'r': 4},
        {'t': 0, 's': 3, 'r': 4},
        {'t': 1, 's': 2, 'r': 4},
        {'t': 1, 's': 3, 'r': 4},
    ]))
```

crystalzhaizhai:
Could you please assign it to me?

kevinsung:
The following code still doesn't work:
```python
import cirq
import sympy

a = cirq.LineQubit(0)
circuit = cirq.Circuit(cirq.X(a)**sympy.Symbol('t'),
                        cirq.measure(a, key='m'))
params = {'t': [0.3, 0.5]}
sim = cirq.Simulator()
result = sim.run_sweep(circuit, params=params)
```

maffoo:
I think we decided not to do this so that you can resolve list-valued parameters (we have use cases for this internally). Also, there's an ambiguity if you have multiple parameters; given `{'a': [1, 2], 'b': [3, 4]}` should that do a cartesian product sweep over `a` and `b` or a zip sweep? Instead you need to make the sweep explicit at the top level, either as `[{'t': 0.3}, {'t': 0.5}]` or using actual sweep objects: `cirq.Points('t', [0.3, 0.5])`.

kevinsung:
If that's the case, then we need to roll back https://github.com/quantumlib/Cirq/pull/2786. That PR already partially implemented this feature, using the Cartesian product definition.

balopat:
Cirq Cynq: 

Verdict: we'll keep the dict shorthand but with explicit methods
Two new methods  

```
dict_to_zip_sweep()
dict_to_product_sweep()
```

[...]

PR #2786, opened 2020-02-24
Expand SWEEP_LIKE to work for {'t': [0, 1, 3]}

crystalzhaizhai:
Fixes: https://github.com/quantumlib/Cirq/issues/2698


vtomole:
From (https://stackoverflow.com/a/40623158/5716192). Let me know what you think.
```python
import itertools

def dict_product(dicts):
    return (dict(zip(dicts, x)) for x in itertools.product(*dicts.values()))

print(list(dict_product({'t': [0, 2, 3]}))) # [{'t': 0}, {'t': 2}, {'t': 3}]
print(list(dict_product({'t': [0, 1], 's': [2, 3], 'r': 4}))) # Will not work for this call. Handle this case because '4' is not a list
```

crystalzhaizhai:
> From (https://stackoverflow.com/a/40623158/5716192). Let me know what you think.
> 
> ```python
> import itertools
> 
> def dict_product(dicts):
>     return (dict(zip(dicts, x)) for x in itertools.product(*dicts.values()))
> 
> print(list(dict_product({'t': [0, 2, 3]}))) # [{'t': 0}, {'t': 2}, {'t': 3}]
> print(list(dict_product({'t': [0, 1], 's': [2, 3], 'r': 4}))) # Will not work for this call. Handle this case because '4' is not a list
> ```

We recreated lots of packages by ourselves, I don't know what external packages we are supposed to use.

vtomole:
@crystalzhaizhai   Let's use `itertools` here. I bet it's more efficient than what we are doing now.

crystalzhaizhai:
> @crystalzhaizhai Let's use `itertools` here. I bet it's more efficient than what we are doing now.

It is more efficient. However, it looks like it won't work for our test case. And it also will break our previous cases. Because it only accepts lists as values.

vtomole:
That code snippet was just an example of how we should use `itertools`. As @Strilanc said,
`Check if the input is a dictionary with list values (and possibly singletons too), and if so take the cartesian product of the various entries and turn it into a sweep.`

The `else` will deal with previous cases.

cirq/study/sweepable.py
```diff
@@ -55,7 +55,24 @@ def to_sweeps(sweepable: Sweepable) -> List[Sweep]:
     if isinstance(sweepable, Sweep):
         return [sweepable]
     if isinstance(sweepable, dict):
-        return [_resolver_to_sweep(ParamResolver(cast(Dict, sweepable)))]
+        expandsweepable = [cast(dict, {})]
```
dstrain115: Worth a comment to explain what case this code is handling?

cirq/study/sweepable.py
```diff
@@ -55,7 +55,24 @@ def to_sweeps(sweepable: Sweepable) -> List[Sweep]:
...
+        for key in sweepable.keys():
```
dstrain115: You should be able to use something like:

`for key,value in sweepable.items():`

cirq/study/sweepable_test.py
```diff
@@ -83,6 +83,18 @@ def test_to_sweeps_iterable_sweeps():
     assert cirq.study.to_sweeps(sweeps) == sweeps
 
 
+def test_to_sweeps_dictionary_of_list():
+    assert cirq.study.to_sweeps({'t': [0, 2, 3]}) == \
```
dstrain115: You can use parentheses to avoid the "\"

```
assert (cirq.study.to_sweeps({'t': [0, 2, 3]}) == 
   cirq.study.to_sweeps([{'t': 0}, {'t': 2}, {'t': 3}]))
```

cirq/study/sweepable_test.py
```diff
@@ -83,6 +83,18 @@ def test_to_sweeps_iterable_sweeps():
...
+        cirq.study.to_sweeps([{'t': 0}, {'t': 2}, {'t': 3}])
+    assert cirq.study.to_sweeps({'t': [0, 1], 's': [2, 3], 'r': 4}) == \
```
dstrain115: I think this case is quite ambiguous.  You are assuming that this is a product, but it could be a zip, i.e.

```
cirq.study.to_sweeps([
            {'t': 0, 's': 2, 'r': 4},
            {'t': 1, 's': 3, 'r': 4},
        ])
```

At the very least, we need to document that multiple lists will be interpreted as a Product and not a Zip.

cirq/study/sweepable.py
```diff
@@ -55,7 +55,24 @@ def to_sweeps(sweepable: Sweepable) -> List[Sweep]:
...
+            value = sweepable[key]
+            if isinstance(value, Iterable):
+                tempsweepable = []
+                for item in value:
+                    for shortdict in expandsweepable:
+                        tempdict = shortdict.copy()
+                        tempdict[key] = item
+                        tempsweepable.append(cast(dict, tempdict))
```
dstrain115: The above code is quite tricky to understand.  It took me a while to figure out what this was doing, since it is repeatedly assigning the values.  I would echo Victory's comment for using itertools if possible.  This will also avoid recreating the entire dictionary every time there is a new list found as well.

cirq/study/sweepable.py
```diff
@@ -55,7 +56,21 @@ def to_sweeps(sweepable: Sweepable) -> List[Sweep]:
     if isinstance(sweepable, Sweep):
         return [sweepable]
     if isinstance(sweepable, dict):
-        return [_resolver_to_sweep(ParamResolver(cast(Dict, sweepable)))]
+        """
```
vtomole: Remove this docstring and replace it with `#`.

PR #2789, opened 2020-02-25
Respect virtual moments with noisy_ops.

95-martin-orion:
Previously, noise models defined with `noisy_operations` would ignore virtual moments, potentially leading to some unusual behavior. This is now enforced in the base NoiseModel.

My concerns: is this the correct place to put this? This makes `noisy_operations` default to respecting virtual moments, but neither of the other methods (`noisy_moment[s]`) have such a default.

### DIFF
--- a/cirq/devices/noise_model.py
+++ b/cirq/devices/noise_model.py
@@ -106,7 +106,8 @@ class NoiseModel(metaclass=value.ABCMetaImplementAnyOneOf):
         for moment in moments:
             if self.is_virtual_moment(moment):
                 result.append(moment)
-            result.append([self.noisy_operation(op) for op in moment])
+            else:
+                result.append([self.noisy_operation(op) for op in moment])
         return result
 
     @value.alternative(requires='noisy_moment',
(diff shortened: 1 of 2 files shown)


cirq/devices/noise_model.py
```diff
@@ -133,6 +136,8 @@ def _noisy_moment_impl_moments(self, moment: 'cirq.Moment',
     def _noisy_moment_impl_operation(self, moment: 'cirq.Moment',
                                      system_qubits: Sequence['cirq.Qid']
                                     ) -> 'cirq.OP_TREE':
+        if self.is_virtual_moment(moment):
```
dstrain115: Worth a short comment to explain why we ignore noise in virtual moments?

95-martin-orion: Done.

95-martin-orion:
In order to avoid adding complexity to the already-complex noise model space, I'm closing this PR without merging. We can revisit at a later point if it becomes clear this is the right course of action.

PR #2790, opened 2020-02-25
Upgraded Gate documentation

dstrain115:
- Reorganized gate.md in order of simple to complex:

- Added new section about Gates vs Operations
- Moved up common gates sections and added new gates
- Moved down magic methods in advanced section on creating custom gates
- Resolved "TODOs" hanging around the page


docs/gates.md
```diff
@@ -1,46 +1,224 @@
...
+``cirq.SWAP ** t`` is the same as ``cirq.SwapPowGate(exponent = t)``
 
-We describe some magic methods below.
+**cirq.ISWAP / cirq.ISwapPowGate**  The swap gate swaps the |01⟩ and |10⟩
+states and adds a relative phase of i.  ``cirq.ISWAP ** t`` is the same as
+``cirq.ISwapPowGate(exponent = t)``
 
+**Parity gates**: The gates cirq.XX, cirq.YY, and cirq.ZZ are equivalent to
+performing the equivalent one-qubit Pauli gates on both qubits.  The gates
+cirq.XXPowGate, cirq.YYPowGate, and cirq.ZZPowGate are the exponentiations of
```
kevinsung: I think "power" is more accurate than "exponentiation", which to me sounds ambiguous as to whether the gate is the base or the exponent.

kevinsung: We should have an explanation of what it means to raise a gate to a power, but that doesn't need to be in this PR.

docs/gates.md
```diff
@@ -1,46 +1,224 @@
...
+# Gates
 
 A ``Gate`` is an operation that can be applied to a collection of
 qubits (objects with a ``Qid``).  ``Gates`` can be applied
 to qubits by calling their ``on`` method, or, alternatively
 calling the gate on the qubits.  The object created by such calls
 is an ``Operation``.
+
+This can be compared to circuit diagrams often seen in quantum computing
+papers and textbooks.
```
kevinsung: It's not clear what "This" is referring to. I would get rid of this sentence.

docs/gates.md
```diff
@@ -1,46 +1,224 @@
...
+Most gates also have a unitary matrix representation, which can be accessed
+by ``cirq.unitary(gate)``.  
 
-The most common type of ``Gate`` is one that corresponds to applying
-a unitary evolution on the qubits that the gate acts on.
 ``Gate``s can also correspond to noisy evolution on the qubits. This
 version of a gate is not used when sending the circuit to a
 quantum computer for execution, but it can be used with
-various simulators. See [noise documentation](noise.md).
+various simulators. See [noise documentation](noise.md) for more details.
```
kevinsung: I know you didn't write these sentences, but I think the part about noise is unclear. There aren't noisy "versions" of gates (are there?). In other words, a gate doesn't have a noiseless and noisy version; it just can be noise if we use it in a noise model. The noise model treats the given gate as is, without modifying it.

docs/gates.md
```diff
@@ -1,46 +1,224 @@
...
+**cirq.X / cirq.XPowGate / cirq.rx** Rotations about the Pauli ``X`` axis.
+The gate `cirq.X**t` and the equivalent representation
+`cirq.XPowGate(exponent=t)` are measured in terms of 180 degree turns, while
+the function `cirq.rx(θ)` uses a radian angle of rotation.
+
+**cirq.PhasedXPowGate**
+This gate is a rotation about an axis in the XY plane of the Bloch sphere.
+The ``PhasedXPowGate`` takes two parameters, ``exponent`` and ``phase_exponent``.
+The gate is equivalent to the circuit `───Z^-p───X^t───Z^p───` where `p` is the
+`phase_exponent` and `t` is the `exponent`.
```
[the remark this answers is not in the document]
kevinsung: Z^p is defined as e^{i pi |1⟩⟨1| t} (see above). PhasedXPowGate is the natural single-qubit gate on Google's current devices.

docs/gates.md
```diff
@@ -1,46 +1,224 @@
...
+
+**cirq.H / cirq.HPowGate** The Hadamard gate is a rotation around the X+Z axis.
+``cirq.HPowGate(exponent=t)`` is a variable rotation of t turns around this
+axis. ```cirq.H` is a π rotation and is equivalent to
```
kevinsung: It looks like the backticks around `cirq.H` are incorrect (three on the left and one on the right, should be two on both sides).

docs/gates.md
```diff
@@ -1,46 +1,224 @@
...
 
-Standard magic methods in python are `__add__`, `__eq__`, and `__len__`.
-Cirq defines several additional magic methods, for functionality such as parameterization, diagramming, and simulation.
-For example, if a gate specifies a `_unitary_` method that returns a matrix for the gate, then simulators will be able to simulate applying the gate.
-Or, if a gate specifies a `__pow__` method that works for an exponent of -1, then `cirq.inverse` will start to work on lists including the gate.
+**cirq.SWAP / cirq.SwapPowGate** The swap gate swaps the |01⟩ and |10⟩ states.
+``cirq.SWAP ** t`` is the same as ``cirq.SwapPowGate(exponent = t)``
 
-We describe some magic methods below.
+**cirq.ISWAP / cirq.ISwapPowGate**  The swap gate swaps the |01⟩ and |10⟩
```
kevinsung: "swap gate" should be "i-swap gate"

docs/gates.md
```diff
@@ -1,46 +1,224 @@
...
+### Other Gates
+
+**cirq.MatrixGate**: A gate defined by its unitary matrix in the form of a
+numpy ndarray.
+
+**cirq.WaitGate**:  This gate does nothing for a specified `cirq.Duration`
+amount of time.  This is useful for conducting T1 and T2 decay experiments. 
+
+**cirq.CCNOT, cirq.CCX, cirq.TOFFOLI, cirq.CCXPowGate**: Three qubit gates
+representing the controlled-controlled-X gates and its resulting exponentiation.
```
kevinsung: Again, I think "power" is less ambiguous than "exponentiation". This comment applies below too.

docs/gates.md
dstrain115: Changed all instances of exponentiations to powers.

docs/gates.md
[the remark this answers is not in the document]
dstrain115: Correct.  Let me know if the explanation is still unclear.

docs/gates.md
dstrain115: Done.  I explain it in the next section anyway.

docs/gates.md
```diff
@@ -1,46 +1,224 @@
...
+
+In this example, a Hadamard is an example of a `Gate` object that can be applied
+in many different circumstances and to many different qubits.  Note that the
+above example has two instances of an H gate but applied to different qubits.
+This is an example of one `Gate` type with two `Operation` instances, one
+applied to the qubit '|b⟩' and the other applied to qubit '|a⟩'.
+
+Gates can generally be applied to any type of qubit (``NamedQubit``,
+``LineQubit``, ``GridQubit``, etc) to create an Operation.
+However, depending on the application, you may want to prefer a specific type
```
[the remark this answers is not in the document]
dstrain115: Done.

docs/gates.md
dstrain115: I took another try at rewriting this paragraph so it is more clear.  What do you think?

docs/gates.md
```diff
@@ -1,46 +1,224 @@
...
+### Single qubit gates
+
+Most single-qubit gates can be thought of as rotation around an axis in the
+[Bloch Sphere](https://en.wikipedia.org/wiki/Bloch_sphere) representation and
+are usually referred to by their axis of rotation.
+
+
+**cirq.Z / cirq.ZPowGate / cirq.rz** Rotations about the Pauli ``Z`` axis.
+The matrix of `cirq.Z**t` and the equivalent representation
+`cirq.ZPowGate(exponent=t)` is ``exp(i pi |1><1| t)`` whereas the matrix of
```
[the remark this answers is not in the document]
dstrain115: Ok, I think I converted all the angle brackets now.

docs/gates.md
dstrain115: Fixed.

docs/gates.md
dstrain115: Changed.

docs/gates.md
dstrain115: Changed.

docs/gates.md
```diff
@@ -1,58 +1,235 @@
...
 
-### Magic Methods
+Most gates also have a unitary matrix representation, which can be accessed
+by ``cirq.unitary(gate)``.  
+
+``Gate``s can also correspond to noisy evolution on the qubits.  These `Gate`s
+may not correspond to unitary operations and may instead represent a noise
+`channel` or may produce a probabilistic `mixture` of states. Such gates are
+generally only used in simulations.  See [noise documentation](noise.md)
+for more details.
```
kevinsung: How about
> Not all ``Gate``s correspond to unitary evolution. They may represent a probabilistic mixture of unitaries, or a general quantum channel. The component unitaries and associated probabilities of a gate can be accessed by ``cirq.mixture(gate)``. The Kraus operator representation of a channel can be accessed by ``cirq.channel(gate)``. Non-unitary gates are often used in the simulation of noise. See [noise documentation](noise.md)
for more details.

docs/gates.md
kevinsung: I commented below with a suggested rewrite.

docs/gates.md
dstrain115: Looks good to me.  Thanks for reviewing and for writing this paragraph!

docs/gates.md
```diff
@@ -1,58 +1,236 @@
...
+by ``cirq.unitary(gate)``.  
+
+Not all `Gate`s correspond to unitary evolution. They may represent a
+probabilistic mixture of unitaries, or a general quantum channel. The component
+unitaries and associated probabilities of a mixture can be accessed by
+`cirq.mixture(gate)`. The Kraus operator representation of a channel can be
+accessed by `cirq.channel(gate)`. Non-unitary gates are often used in the
+simulation of noise. See [noise documentation](noise.md) for more details.
+
+Many arithmetic operators will also work in the expected way when applied to
```
kevinsung: Remove the word "also".

docs/gates.md
```diff
@@ -1,58 +1,236 @@
...
+axis. `cirq.H` is a π rotation and is equivalent to
+``cirq.HPowGate(exponent=1)``
+
+**S** The square root of Z gate, equivalent to ``cirq.Z ** 0.5``
+
+**T** The fourth root of Z gate, equivalent to ``cirq.Z ** 0.25``.  Often used
+in error correction.
+
+
+## Two qubit gates
```
kevinsung: This heading should be `###`, right?

docs/gates.md
```diff
@@ -1,58 +1,236 @@
...
+## Common gates
+
+Cirq supports a number of gates natively, with the opportunity to extend these
+gates for more advanced use cases.
+
+### Measurement gate
+
+**cirq.MeasurementGate** This is a measurement in the computational basis.
+This gate can be applied to a variable number of qubits.  The function
+``cirq.measure(q0,q1,...)`` can also be used as a short-hand to create a
```
mpharrigan: nit: spaces around commas

docs/gates.md
dstrain115: Done.

docs/gates.md
dstrain115: Done.

docs/gates.md
dstrain115: Done

docs/gates.md
```diff
@@ -1,58 +1,236 @@
...
+``MeasurementGate`` .
```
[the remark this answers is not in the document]
dstrain115: Removed double backticks here.

docs/gates.md
```diff
@@ -1,58 +1,236 @@
...
+
+**cirq.H / cirq.HPowGate** The Hadamard gate is a rotation around the X+Z axis.
+``cirq.HPowGate(exponent=t)`` is a variable rotation of t turns around this
+axis. `cirq.H` is a π rotation and is equivalent to
+``cirq.HPowGate(exponent=1)``
+
+**S** The square root of Z gate, equivalent to ``cirq.Z ** 0.5``
+
+**T** The fourth root of Z gate, equivalent to ``cirq.Z ** 0.25``.  Often used
+in error correction.
```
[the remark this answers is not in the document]
dstrain115: Done.

docs/gates.md
```diff
@@ -1,58 +1,236 @@
...
+The ``PhasedXPowGate`` takes two parameters, ``exponent`` and ``phase_exponent``.
+The gate is equivalent to the circuit `───Z^-p───X^t───Z^p───` where `p` is the
+`phase_exponent` and `t` is the `exponent`.
+
+**cirq.H / cirq.HPowGate** The Hadamard gate is a rotation around the X+Z axis.
+`cirq.HPowGate(exponent=t)` is a variable rotation of t turns around this
+axis. `cirq.H` is a π rotation and is equivalent to
+`cirq.HPowGate(exponent=1)`
+
+**S** The square root of Z gate, equivalent to `cirq.Z ** 0.5`
```
kevinsung: Above you omit spaces around `**` but here you include them. Be consistent. I suggest omitting the spaces because that's how we write code.

docs/gates.md
```diff
@@ -1,58 +1,236 @@
...
+
+**T** The fourth root of Z gate, equivalent to `cirq.Z ** 0.25`.  Often used
+in fault-tolerant algoithms.
```
viathor: This doesn't sound right. The significance of T in the context of fault-tolerant quantum computation is that for many quantum error correcting codes T is the only gate in the standard set of universal gates (i.e. {H, T, CNOT}) which cannot be done transversally. Instead, one uses gate teleportation and measurements of Clifford operators to perform the T gate, see [1] (the same technique works for Toffoli).

That said, I'm not sure any of this needs to be written here. Perhaps just remark on the universality?

Suggestion: replace "Often used in fault-tolerant algoithms." with "Together with Hadamard and CNOT makes a universal set of gates.".

---
[1] https://arxiv.org/abs/quant-ph/9908010

docs/gates.md
dstrain115: Removed all spaces around ** operations.

docs/gates.md
dstrain115: I just removed this sentence.  I don't think it's worth getting into any of that.

FYI, this is replacing the description "and our error correcting friend the T gate" in the original documentation.

docs/gates.md
```diff
@@ -1,58 +1,235 @@
...
+The matrix of `cirq.Z**t` and the equivalent representation
+`cirq.ZPowGate(exponent=t)` is ``exp(i pi |1⟩⟨1| t)`` whereas the matrix of
+`cirq.rz(θ)` is `exp(-i Z θ/2)`.  Since computation is often done in the
+Z-basis, these are implemented as phase changes on later operations on
+[Google devices](google/devices.md) instead of a physical modification applied
+to the qubits.
+
+**cirq.Y / cirq.YPowGate / cirq.ry** Rotations about the Pauli ``Y`` axis.
+The gate `cirq.Y**t` and the equivalent representation
+`cirq.YPowGate(exponent=t)` are measured in terms of 180 degree turns, while
```
smitsanghavi:
```suggestion
`cirq.YPowGate(exponent=t)` are measured in terms of 180 degree turns (half turns), while
```
since a lot of documentation uses the half-turn terminology.

docs/gates.md
```diff
@@ -1,58 +1,235 @@
...
+
+Many arithmetic operators will work in the expected way when applied to
+gates.  For instance, ``cirq.X**0.5`` represents a square root of X gate.
+These can also be applied to Operators for a more compact representation, such
+as ``cirq.X(q1)**0.5`` will be a square root of X gate applied to the q1
+qubit.  This functionality depends on the "magic methods" of the gate
+being defined (see below for details).
+
+Gates can be converted to a controlled version by using `Gate.controlled()`.
+For instance, `cirq.X.controlled()` returns a `cirq.CNOT` gate.
```
smitsanghavi: This is a special case. I'm concerned that just giving this example might give the false impression. 
In general, `Gate.controlled()` on an arbitrary `Gate` gives a `ControlledGate` except in special circumstances of these well known gates.

docs/gates.md
```diff
@@ -1,58 +1,235 @@
-## Gates
+# Gates
 
 A ``Gate`` is an operation that can be applied to a collection of
 qubits (objects with a ``Qid``).  ``Gates`` can be applied
 to qubits by calling their ``on`` method, or, alternatively
 calling the gate on the qubits.  The object created by such calls
 is an ``Operation``.
```
smitsanghavi: Starting the paragraph with "A Gate is an operation that..." and ending with the definition of `Operation` is awkward and circular. 
If `Operation` is the more fundamental concept, let's start with defining that and then defining `Gate` as a operation factory that, given input qubits, generates its associated operation.
If we want to keep `Gate` as fundamental and `Operation` as derived from Gate+qubits, change the first line to not contain the word "operation".

docs/gates.md
dstrain115: Also changed on Z gate explanation.

docs/gates.md
dstrain115: Done.

docs/gates.md
dstrain115: Changed the first line to use 'effect' instead of operation.  I also added the explanation of a gate as an operation factory to the end of the paragraph.  I figured that it is best to explain the concepts in several ways.

docs/gates.md
```diff
@@ -1,58 +1,241 @@
...
+
+Most single-qubit gates can be thought of as rotation around an axis in the
+[Bloch Sphere](https://en.wikipedia.org/wiki/Bloch_sphere) representation and
+are usually referred to by their axis of rotation.
+
+
+**cirq.Z / cirq.ZPowGate / cirq.rz** Rotations about the Pauli ``Z`` axis.
+The matrix of `cirq.Z**t` and the equivalent representation
+`cirq.ZPowGate(exponent=t)` is ``exp(i pi |1⟩⟨1| t)`` whereas the matrix of
+`cirq.rz(θ)` is `exp(-i Z θ/2)`.  Since computation is often done in the
```
tanujkhattar: Instead of only specifying the matrix for rz, should we add a brief section on rotation operators similar to https://www.quantum-inspire.com/kbase/rotation-operators/ ? Or at least make the cirq.Y and cirq.X definition similar to cirq.Z ? The current representation seems to give an impression that cirq.rz is special and matrix representation is therefore given.

docs/gates.md
```diff
@@ -1,58 +1,241 @@
...
+
+
+## Advanced: Creating your own gates
+
+If the above gates are not sufficient for your use case, it is fairly simple
+to create your own gate.   In order to do so, you can define your class and
+inherit the `cirq.Gate` class and define the functionality in your class.
+
+At minimum, you will need to define either the ``_num_qubits_`` or
+``_qid_shape_`` magic method to define the number of qubits (or qudits) used
```
tanujkhattar: magic method is used before it's defined in the next paragraph. We can either assume it to be a prerequisite or define before using it?

docs/gates.md
```diff
@@ -1,58 +1,241 @@
...
+
+### Magic Methods
+
+
+#### Standard python magic methods
+
+There are many standard magic methods in python.  Here are a few of the most
+important ones used in cirq:
+  * `__str__` for user-friendly string output and  `__repr__` which should be
+  able to be evaluated by python to the object itself.
```
tanujkhattar: "which should be able to be evaluated by python to the object itself"  - this is not very clear.

dstrain115: Changed to state that eval(repr(y))==y should always be true.

docs/gates.md
dstrain115: Rearranged the two paragraphs so that the definition comes first.

docs/gates.md
dstrain115: I have expanded this section and split out the entries for rx,ry,rz so that it is hopefully more clear.

dstrain115:
Thanks.  Submitting!  If you have further comments or notice anything incorrect or awkward, please let me know or file an issue, and I will follow up with a fix.

Thanks everyone for reviewing!

PR #2794, opened 2020-02-26
Add EngineProcessorTimeSlot to cirq.google.engine

dstrain115:
- This class will be a wrapper to hold timeslots for reservations.


cirq/google/engine/engine_processor_timeslot.py
```diff
@@ -0,0 +1,48 @@
...
+@dataclasses.dataclass
+class EngineProcessorTimeSlot:
+    """A python wrapping of a Quantum Engine timeslot.
+  """
+    start_seconds: int
+    end_seconds: int
+    slot_type: EngineProcessorTimeSlotType = _DEFAULT_SLOT_TYPE
+    project_id: Optional[int] = None
+    maintenance_title: Optional[str] = None
+    maintenance_description: Optional[str] = None
```
wcourtney: The refreshed interface now consistently exposes only primary keys as fields and provides accessors that lazily load the underlying object. Consider keeping that pattern for consistency?

e.g.
https://github.com/quantumlib/Cirq/blob/master/cirq/google/engine/engine_program.py
https://github.com/quantumlib/Cirq/blob/master/cirq/google/engine/engine_job.py

dstrain115: Used frozen=True instead of adding accessors to have a short, readable class.

cirq/google/engine/engine_processor_timeslot.py
```diff
@@ -0,0 +1,48 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import dataclasses
+import enum
+
+from typing import Optional
+
+
+class EngineProcessorTimeSlotType(enum.Enum):
```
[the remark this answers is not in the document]
dstrain115: Moved this enum into client/..../enum.py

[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/engine/client/quantum_v1alpha1/gapic/enums.py
```diff
@@ -94,6 +94,23 @@ class Health(enum.IntEnum):
         DOWN = 2
         UNAVAILABLE = 3
 
+    class TimeSlotType(enum.IntEnum):
```
wcourtney: In the API, the TimeSlotType nested inside of QuantumTimeSlot. Does this need to match to work appropriately with the gRPC client, or is the plan to translate these? I think that this code was originally generated from the API protos, so I'm guessing that they should be in sync.

cirq/google/engine/engine_timeslot.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@dataclasses.dataclass(frozen=True)
+class EngineTimeSlot:
+    """A python wrapping of a Quantum Engine timeslot.
+  """
+    start_seconds: int
+    end_seconds: int
+    slot_type: enums.QuantumProcessor.TimeSlotType = _DEFAULT_TYPE
+    project_id: Optional[int] = None
```
wcourtney: project_id should be a string..

cirq/google/engine/engine_timeslot.py
```diff
@@ -0,0 +1,31 @@
...
+from cirq.google.engine.client.quantum_v1alpha1.gapic import enums
+
+_DEFAULT_TYPE = enums.QuantumProcessor.TimeSlotType.TIME_SLOT_TYPE_UNSPECIFIED
+
+
+@dataclasses.dataclass(frozen=True)
+class EngineTimeSlot:
+    """A python wrapping of a Quantum Engine timeslot.
+  """
+    start_seconds: int
```
wcourtney: WDYT of datetime for start and end times? Is there a reason to prefer seconds?

dstrain115: Well, this would require some additional conventions.  The proto defines this as a Timestamp which is seconds since epoch:

https://cs.corp.google.com/piper///depot/google3/google/protobuf/timestamp.proto?l=125

If we define this in datetime, we have to figure out how to translate this to datetime with timezone information and watch out that we don't screw up when serializing / deserializing.  (For instance, if one researcher serializes to JSON and another in the project but in a different time zone gets the timeslot and uses it.)

It's totally doable, but it adds some complexity here.  Are you sure we need this information?

cirq/google/engine/engine_timeslot.py
dstrain115: Done.

cirq/google/engine/client/quantum_v1alpha1/gapic/enums.py
dstrain115: Done.

cirq/google/engine/engine_timeslot.py
wcourtney: We don't need it, but I'm concerned about the usability if this is the interface for viewing a processor's schedule. Time slots are read-only, and IIUC `datetime.fromtimestamp()` will convert to local time, so it sounds like we're just pushing that onto the users. Not a big deal yet though and I can wait to see where this goes.

dstrain115: Ok, changed to datetime.  I figure its easier to change now than later, and we can probably create a from_proto() function once the protos are in cirq.

PR #2795, opened 2020-02-26
In unitary_eig, check preconditions before performing computing eigenvalues

kevinsung:
They are _pre_-conditions after all.

Also defines `cirq.is_normal`.

### DIFF
--- a/cirq/__init__.py
+++ b/cirq/__init__.py
@@ -122,6 +122,7 @@ from cirq.linalg import (
     eye_tensor,
     is_diagonal,
     is_hermitian,
+    is_normal,
     is_orthogonal,
     is_special_orthogonal,
     is_special_unitary,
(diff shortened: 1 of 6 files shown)


kevinsung:
The diagonal check is not cheaper because it requires first computing the factorization, which is much more expensive than checking whether the matrix is normal. The purpose of performing the check should be to prevent the expensive computation from happening in the event of a failed check.

viathor:
I disagree. You're only taking into account the cost when the matrix fails to be normal. In this case before the PR we indeed pay the full cost of the Schur decomposition which is greater than the cost of commutativity check. However, the commutativity check costs more when the matrix is in fact normal. Consequently, as long as we think that non-normal inputs are *rare* this PR increases the overall cost.

(That said, I don't think the increase is going to be a problem in practice.)

kevinsung:
Ah, I see what you're saying. I still prefer the semantics of the change that I made (and again, users can always disable the check).

viathor:
Yes, I agree with your PR: it is weird that `check_precondition` checks a post-condition. I don't think we were discussing a significant cost and in any case it can be disabled, so LGTM.

PR #2796, opened 2020-02-27
Convert single-qubit gates to single-gate PhasedXZ for sycamore

maffoo:
I've added a separate decomposition and merge optimizer for the `PhasedXZ` case. I think ideally this would just replace the existing code which emits separate `PhasedXPow` and `Z` gates, but that is currently used in, for example, the ion trap decompositions where it is not clear that these devices and/or APIs support `PhasedXZ`. Once this is in I have a separate PR to add support for `PhasedXZ` in the `EjectZ` optimizer.


cirq/optimizers/merge_single_qubit_gates.py
```diff
@@ -121,3 +121,25 @@ def synth(qubit: 'cirq.Qid', matrix: np.ndarray) -> List[ops.Operation]:
...
     MergeSingleQubitGates(synthesizer=synth).optimize_circuit(circuit)
+
+
+def merge_single_qubit_gates_into_phxz(
+        circuit: circuits.Circuit,
+        atol: float = 1e-8,
+) -> None:
+    """Canonicalizes runs of single-qubit rotations in a circuit.
+
+    Specifically, any run of non-parameterized circuits will be replaced by an
```
dstrain115: Would "any run of non-parameterized single-qubit operations" be more precise?

cirq/optimizers/merge_single_qubit_gates.py
```diff
@@ -121,3 +121,25 @@ def synth(qubit: 'cirq.Qid', matrix: np.ndarray) -> List[ops.Operation]:
...
+        circuit: The circuit to rewrite. This value is mutated in-place.
+        atol: Absolute tolerance to angle error. Larger values allow more
+            negligible gates to be dropped, smaller values increase accuracy.
+    """
+
+    def synth(qubit: 'cirq.Qid', matrix: np.ndarray) -> List[ops.Operation]:
+        gate = decompositions.single_qubit_matrix_to_phxz(matrix, atol)
+        return [gate(qubit)] if gate else []
+
+    MergeSingleQubitGates(synthesizer=synth).optimize_circuit(circuit)
```
dstrain115: Is this tested anywhere?  If so, I missed it.

cirq/optimizers/merge_single_qubit_gates.py
maffoo: That's definitely better; fixed.

cirq/optimizers/merge_single_qubit_gates.py
maffoo: It was being tested indirectly since it gets used in `optimized_for_sycamore`; will add an explicit test for just this optimizer.

cirq/optimizers/decompositions.py
```diff
@@ -211,3 +211,43 @@ def single_qubit_matrix_to_phased_x_z(
...
+        x_exponent=2 * xy_turn,
+        axis_phase_exponent=2 * xy_phase_turn,
+        z_exponent=2 * total_z_turn,
+    )
+
+    if protocols.trace_distance_bound(g) <= atol:
+        return None
+
+    # Special case: XY half-turns can absorb Z rotations.
+    if abs(xy_turn) >= 0.5 - atol:
```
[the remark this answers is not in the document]
maffoo: `xy_turn` is in the range `[-0.5, 0.5]` thanks to the `_signed_mod_1` function, so this is equivalent to checking if `abs(xy_turn)` is close to 0.5 (it's also done this way in the existing function `single_qubit_matrix_to_phased_x_z` above. But I agree this is a bit confusing; will change to use `math.isclose`.

PR #2801, opened 2020-03-03
More gates aqt

ma5x:
Added the R-gate (AQT's version of Cirq's 'PhasedXPowGate) to the set of physical gates for AQT devices.

Also fixed a bug in AQT's data return format.


vtomole:
Not related to this pull request: Are there plans for fix these issues? https://github.com/quantumlib/Cirq/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+aqt

cirq/aqt/aqt_sampler.py
```diff
@@ -82,15 +82,20 @@ def _generate_json(
...
         circuit = resolve_parameters(circuit, param_resolver)
         for op in circuit.all_operations():
             line_qubit = cast(Tuple[LineQubit], op.qubits)
             op = cast(ops.GateOperation, op)
             qubit_idx = [obj.x for obj in line_qubit]
             op_str = get_op_string(op)
             gate = cast(ops.EigenGate, op.gate)
-            seq_list.append((op_str, gate.exponent, qubit_idx))
+            if op_str == 'R':
+                seq_list.append((op_str, float(gate.exponent),
```
vtomole: Why is `float` called here and not in the `else`?

ma5x:
> Not related to this pull request: Are there plans for fix these issues? https://github.com/quantumlib/Cirq/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+aqt

Thanks for bringing this to my attention! On first sight, some of the issues seem to be resolved already, but I will have a closer look.

cirq/aqt/aqt_sampler.py
ma5x: This is an inconsistency connected to the failed mypy test. Working on it...

vtomole: Did we fix this issue yet?

ma5x: i think so: The parameter is now cast to float() in both cases.

PR #2803, opened 2020-03-03
Handle all possible 1-qubit Clifford gates in Clifford simulator

fedimser:
I used the fact that all 1-qubit gates can be generated by H and S gates, and there are only 24 of them (up to phase) - [proof](http://home.lu.lv/~sd20008/papers/essays/Clifford%20group%20[paper].pdf).
So, I can explicitly generate all those 24 gates as sequences of H and S gates (see [generation code](https://github.com/fedimser/quant_comp/blob/master/All%201-qubit%20CLifford%20gates.ipynb)).
Then, I can compare given gate with all of those 24 gates and check if they differ only in phase. If so, I can implement given gate using existing implementation of S and H gates.

So, this PR ensures that Clifford simulator supports all possible 1-qubit Clifford gates.

Also updated function in mux.py which checks if circuit is Clifford circuit.
There was one test of determinism in mux_test.py which started failing, because now it's simulated on Clifford simulator. Currently Clifford simulator doesn't support taking seed parameter, so I changed that test to use another gate, which is not Clifford gate.


fedimser:
This PR is related to issue #2423.

fedimser:
@smitsanghavi, please have a look (looks like I can't add reviewers to this PR).

cirq/sim/clifford/clifford_gate_decomposer.py
```diff
@@ -0,0 +1,73 @@
+from typing import (Tuple)
```
smitsanghavi: There is already a _decompose_into_cliffords method in pauli_string.py that uses the _decompose_into_clifford_with_qubits_ method. Can these be unified so that there is one right way to do a clifford decomposition?

cirq/sim/clifford/clifford_simulator.py
```diff
@@ -47,8 +48,16 @@ def __init__(self):
         self.init = True
 
     @staticmethod
-    def get_supported_gates() -> List['cirq.Gate']:
-        return [cirq.X, cirq.Y, cirq.Z, cirq.H, cirq.S, cirq.CNOT, cirq.CZ]
+    def is_supported_operation(op: 'cirq.Operation') -> bool:
```
smitsanghavi: This handling is being replaced with a has_stabilizer_effect protocol in #2760

cirq/sim/clifford/clifford_gate_decomposer.py
```diff
@@ -0,0 +1,73 @@
...
+        def name_to_unitary(name):
+            result = np.eye(2)
+            for ch in name:
+                if ch == 'S': result = result @ s
+                elif ch == 'H': result = result @ h
+            return result
+
+        # Calculate matrices for all candidates.
+        self._candidates = [
+            (name, name_to_unitary(name)) for name in _ALL_CLIFFORD_GATES
```
smitsanghavi: every time a CliffordgateDecomposer is initialized, this map is reconstructed. It would be more efficient to have it as a hardcoded constant map instead.

cirq/sim/clifford/clifford_simulator.py
fedimser: That PR have not been merged yet. As soon as it's merged, I'll update this PR to extend `_has_stabilizer_effect_` on all matrices and to use it to check if gate can be simulated by Clifford simulator.

By the way, I think there is more elgeant way of implementing this - instead of defining `_has_stabilizer_effect_` at every gate, we can do the following:
In has_stabilizer_effect_protocol.py check if gate has 2x2 uniary matrix (name it U).
If so, check that `U@X@U.conj().T` and `U@Z@U.conj.T` are in set `X, -X, Y, -Y, Z, -Z`.

This is also better than trying to match unitary with 24 matrices, as I am doing in this PR.

cirq/sim/clifford/clifford_gate_decomposer.py
fedimser: Yes, that's why I made it a singleton, so this computation can be done once in runtime.
This computation is not that expensive - just 2x2 matrix multiplications ~100 times.

I refactored this code to use static methods instead, so now it's clear that this computation is done only once.

cirq/sim/clifford/clifford_gate_decomposer.py
fedimser: _decompose_into_cliffords doesn't dodecomposition itself. It check sthat `op` is already a clifford gate, or it defines `_decompose_into_clifford_`.

In this class I am solving specific problem - given 2x2 unitary, check whether it's a Clifford unitary, and if so - decompose it into produt of H and S gates, and global phase. I didn't find this problem solved (for all Clifford gates) in Cirq. So, I think that this class should exist to solve this problem, and called from other places where this decomposition is needed.

fedimser:
I added a method `from_unitary` to `SingleQubitCliffordGate`.
Since `SingleQubitCliffordGate` already has method `decompose_rotation`, and we know how to decompose rotations into H and S gates, I don't need precalculated list of all possible Clifford gates anymore.
Please take another look.

cirq/sim/clifford/clifford_simulator.py
```diff
@@ -300,25 +316,50 @@ def apply_unitary(self, op: 'cirq.Operation'):
...
+                    self._apply_H(self.qubit_map[op.qubits[0]])
+                    applied_unitary = h @ s @ h @ applied_unitary
+                elif axis == pauli_gates.Y:
+                    self._apply_S(self.qubit_map[op.qubits[0]])
+                    self._apply_S(self.qubit_map[op.qubits[0]])
+                    self._apply_H(self.qubit_map[op.qubits[0]])
+                    applied_unitary = h @ s @ s @ applied_unitary
+                else:
+                    assert axis == pauli_gates.Z
+                    self._apply_S(self.qubit_map[op.qubits[0]])
```
smitsanghavi: I like this implementation much better.

A small question, would it be too messy to check for the gate to be `H` beforehand? I find it unpleasant that an H gate, even though very natively supported, will get resolved to something like `hshhshhss`.

fedimser: Added special handling for H gate.

cirq/sim/clifford/clifford_simulator.py
```diff
@@ -300,25 +316,55 @@ def apply_unitary(self, op: 'cirq.Operation'):
...
-            self.ch_form._H(self.qubit_map[op.qubits[0]])
         else:
-            raise ValueError('%s cannot be run with Clifford simulator' %
+            raise ValueError('%s cannot be run with Clifford simulator.' %
                              str(op.gate))  # type: ignore
 
+    def apply_single_qubit_unitary(self, op: 'cirq.Operation'):
+        assert op.gate is not None
+        qubit = self.qubit_map[op.qubits[0]]
+        if op.gate == cirq.H:
```
smitsanghavi: It would be nice to have a comment explaining why we did this special handling so that future readers aren't confused why `S` isn't included here or if `H` cannot be handled by the process below

fedimser: That's why I didn't want to have any special case handling in first place...

fedimser: Added a comment.

cirq/sim/clifford/clifford_simulator.py
```diff
@@ -300,25 +316,56 @@ def apply_unitary(self, op: 'cirq.Operation'):
...
 
+    def apply_single_qubit_unitary(self, op: 'cirq.Operation'):
+        assert op.gate is not None
+        qubit = self.qubit_map[op.qubits[0]]
+        # Handle H natively as optimization.
+        if op.gate == cirq.H:
+            self._apply_H(qubit)
+            return
+
+        u = unitary(op.gate)
```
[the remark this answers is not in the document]
fedimser: Done.

cirq/ops/clifford_gate.py
```diff
@@ -206,6 +217,32 @@ def _validate_map_input(required_transform_count: int,
...
+    """Creates Clifford gate with given unitary (up to global phase).
+    
+    Args:
+        u: 2x2 unitary matrix of a Clifford gate.
+    
+    Returns:
+        SingleQubitCliffordGate, whose matrix is equal to given matrix (up to
+        global phase), or `None` if `u` is not a matrix of a single-qubit 
+        Clifford gate.
+    """
```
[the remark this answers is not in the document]
fedimser: Done.

cirq/sim/clifford/clifford_simulator.py
```diff
@@ -47,8 +52,17 @@ def __init__(self):
         self.init = True
 
     @staticmethod
-    def get_supported_gates() -> List['cirq.Gate']:
-        return [cirq.X, cirq.Y, cirq.Z, cirq.H, cirq.S, cirq.CNOT, cirq.CZ]
+    def is_supported_operation(op: 'cirq.Operation') -> bool:
+        """Checks whether given operation can be simulated by this simulator."""
+        if protocols.is_measurement(op): return True
+        gate = op.gate
+        if gate is None: return False
```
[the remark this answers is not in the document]
fedimser: Done.

fedimser:
I added tests for PauliString and GlobalPhaseOperation.

GlobalPhaseOperation doesn't have any qubits, so it wasn't handled by my current implementation (which works for ops on exactly 1 qubit). So I just handled it separately (without involving single qubit Clifford gate).

fedimser:
Craig, I can't merge this PR because you requested changes, but I addressed them. I think that some action from you is required to acknowledge that changes were made.
@Strilanc

fedimser:
Can we please merge this now?
I understand that I also need to update `has_stabilizer_effect` (because PR #2760 is now merged) and use that method to determine whether operation is supported by the CLifford simulator, but I'd rather do it in a separate, simple PR.

PR #2804, opened 2020-03-04
Add a limited FSIM gateset

dstrain115:
- This will be equivalent to combining the sqrt iswap and syc gatesets.
- Combining these two gatesets will have the following advantages:
--  This can allow for parameterization of the ISWAP**0 gate
--  That allows circuits to drop out iSwaps and batch efficiently
--  Also allows for expansion if other FSim variants become available

### DIFF
--- a/cirq/google/common_serializers.py
+++ b/cirq/google/common_serializers.py
@@ -418,6 +418,68 @@ SQRT_ISWAP_DESERIALIZERS = [
         args=[]),
 ]
 
+
+#
+# FSim serializer
+# Only allows sqrt_iswap, its inverse, identity, and sycamore
+#
+def _can_serialize_limited_fsim(theta: float, phi: float):
+    # Identity
+    if _near_mod_2pi(theta, 0) and _near_mod_2pi(phi, 0):
+        return True
+    # sqrt ISWAP and inverse sqrt ISWAP
+    if _near_mod_2pi(abs(theta), np.pi / 4) and _near_mod_2pi(phi, 0):
+        return True
+    # Sycamore
+    if _near_mod_2pi(theta, np.pi / 2) and _near_mod_2pi(phi, np.pi / 6):
+        return True
+    return False
+
+
+LIMITED_FSIM_SERIALIZERS = [
+    op_serializer.GateOpSerializer(
+        gate_type=ops.FSimGate,
+        serialized_gate_id='fsim',
+        args=[
+            op_serializer.SerializingArg(serialized_name='theta',
+                                         serialized_type=float,
+                                         op_getter='theta'),
+            op_serializer.SerializingArg(serialized_name='phi',
+                                         serialized_type=float,
+                                         op_getter='phi')
+        ],
+        can_serialize_predicate=(lambda op: _can_serialize_limited_fsim(
+            cast(ops.FSimGate, op.gate).theta,
+            cast(ops.FSimGate, op.gate).phi))),
+    op_serializer.GateOpSerializer(
+        gate_type=ops.ISwapPowGate,
+        serialized_gate_id='fsim',
+        args=[
+            op_serializer.SerializingArg(
+                serialized_name='theta',
+                serialized_type=float,
+                op_getter=(lambda op: cast(ops.ISwapPowGate, op.gate).exponent *
+                           np.pi / 2)),
+            op_serializer.SerializingArg(serialized_name='phi',
+                                         serialized_type=float,
+                                         op_getter=lambda e: 0)
+        ],
+        can_serialize_predicate=(lambda op: _near_mod_n(
+            abs(cast(ops.ISwapPowGate, op.gate).exponent), 0.5, 4
+        ) or _near_mod_n(cast(ops.ISwapPowGate, op.gate).exponent, 0, 4))),
+]
+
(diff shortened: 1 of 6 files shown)


cirq/google/gate_sets.py
```diff
@@ -69,6 +71,23 @@
...
+        MEASUREMENT_SERIALIZER,
+        WAIT_GATE_SERIALIZER,
+    ],
+    deserializers=[
+        LIMITED_FSIM_DESERIALIZER,
+        *SINGLE_QUBIT_DESERIALIZERS,
+        MEASUREMENT_DESERIALIZER,
+        WAIT_GATE_DESERIALIZER,
+    ])
+document(SQRT_ISWAP_GATESET,
```
maffoo: `FSIM_GATESET`

cirq/google/common_serializers.py
```diff
@@ -418,6 +418,68 @@ def _convert_physical_z(op: ops.Operation, proto: v2.program_pb2.Operation):
...
+            cast(ops.FSimGate, op.gate).phi))),
+    op_serializer.GateOpSerializer(
+        gate_type=ops.ISwapPowGate,
+        serialized_gate_id='fsim',
+        args=[
+            op_serializer.SerializingArg(
+                serialized_name='theta',
+                serialized_type=float,
+                op_getter=(lambda op: cast(ops.ISwapPowGate, op.gate).exponent *
+                           np.pi / 2)),
```
maffoo: Need a minus sign here, because iswap**0.5 corresponds to fsim(-pi/4, 0).

cirq/google/gate_sets.py
dstrain115: Fixed.

cirq/google/common_serializers.py
dstrain115: Good catch!  I always screw up this convention.

cirq/google/common_serializers_test.py
```diff
@@ -607,3 +607,64 @@ def test_wait_gate():
     op = cirq.WaitGate(cirq.Duration(nanos=20)).on(q)
     assert gate_set.serialize_op(op) == proto
     assert gate_set.deserialize_op(proto) == op
+
+
+@pytest.mark.parametrize(('gate', 'theta', 'phi'), [
+    (cirq.ISWAP**0.5, -np.pi / 4, 0),
+    (cirq.ISWAP**-0.5, np.pi / 4, 0),
```
mrwojtek: Can you explain to me how does it work, since you didn't specify serializer for ISwapPowGate with exponent equal to -0.5?

dstrain115: The serializer checks for abs(exponent)==0.5 (mod 2pi)

cirq/google/common_serializers.py
```diff
@@ -418,6 +418,69 @@ def _convert_physical_z(op: ops.Operation, proto: v2.program_pb2.Operation):
...
+# FSim serializer
+# Only allows sqrt_iswap, its inverse, identity, and sycamore
+#
+def _can_serialize_limited_fsim(theta: float, phi: float):
+    # Identity
+    if _near_mod_2pi(theta, 0) and _near_mod_2pi(phi, 0):
+        return True
+    # sqrt ISWAP and inverse sqrt ISWAP
+    if _near_mod_2pi(abs(theta), np.pi / 4) and _near_mod_2pi(phi, 0):
+        return True
```
maffoo: There's a subtlety here with periodicity and absolute value; we want to accept theta = pi/4 and -pi/4, but also theta = 7pi/4 (which is equal to -pi/4 mod 2pi). The current code will not accept that. I think we should do:
```python
if (_near_mod_2pi(theta, np.pi / 4) or _near_mod_2pi(theta, -np.pi / 4)) and _near_mod_2pi(phi, 0):
    ...
```
Could also split these into separate if clauses and label them explicitly as sqrt iswap and inverse sqrt iswap.

cirq/google/common_serializers.py
```diff
@@ -418,6 +418,69 @@ def _convert_physical_z(op: ops.Operation, proto: v2.program_pb2.Operation):
...
+                serialized_type=float,
+                # Note that ISWAP ** 0.5 is Fsim(-pi/4,0)
+                op_getter=(lambda op: cast(ops.ISwapPowGate, op.gate).exponent *
+                           -np.pi / 2)),
+            op_serializer.SerializingArg(serialized_name='phi',
+                                         serialized_type=float,
+                                         op_getter=lambda e: 0)
+        ],
+        can_serialize_predicate=(lambda op: _near_mod_n(
+            abs(cast(ops.ISwapPowGate, op.gate).exponent), 0.5, 4
```
maffoo: here, too, need to check whether the exponent is close to 0.5 mod 4 _or_ -0.5 mod 4; taking the absolute value misses exponent 3.5, which gives the same gate as -0.5.

cirq/google/common_serializers.py
dstrain115: You are correct!  I tried to be too fancy for my own good.  I just made all the checks separate and explicit and added tests for the -7pi/4 and 7pi/4

cirq/google/common_serializers.py

PR #2808, opened 2020-03-04
Create qis subpackage and move some existing code to qis.states and qis.measures

kevinsung:
Moves the following things to `linalg/states.py` from `sim/wave_function.py` and `sim/density_matrix_utils.py`:
```
STATE_VECTOR_LIKE
bloch_vector_from_state_vector
density_matrix_from_state_vector
dirac_notation
to_valid_state_vector
validate_normalized_state
to_valid_density_matrix
von_neumann_entropy
```
This may not be their final resting place (see https://github.com/quantumlib/Cirq/pull/2797#discussion_r387915534) but it certainly makes more sense than before, given that we have a file called `states.py`. Also, it has the benefit of removing some circular dependency import hacks.


cirq/sim/__init__.py
```diff
@@ -17,8 +17,6 @@
 from cirq.sim.density_matrix_utils import (
     measure_density_matrix,
     sample_density_matrix,
```
viathor: We may need to import the names into `cirq.sim` as well to avoid breaking clients that depend on this.

kevinsung: There have been several comments on #2810 agreeing with creating `cirq.qis`. Shall I go ahead and move everything to `qis/states.py`? Also I was hoping there would be a better way to deprecate importing from old modules, but if there isn't then sure, we can just import into them.

viathor: Yes, let's go ahead with `cirq.qis`. Thanks!

I'm not aware of a better way. I think you should move the code to the new location, import into the old location and add the deprecated decorator.

kevinsung: The deprecated decorator goes over the function though. As far as I can tell, there is no way to mark _the act of importing from the old module_ as deprecated.

viathor: Decorator is just a function (that takes and returns functions). Also, you don't need to mark the act of importing as deprecated. Instead, think about a symbol under `cirq.sim` as being defined using the `deprecated` function and the new symbol under `cirq.qis`.

This may work:

In a module inside the `cirq.qis` package:

```python
def von_neumann_entropy(...):
    ...
```

In a module inside the `cirq.sim` package:

```python
from cirq.qis import von_neumann_entropy as qis_von_neumann_entropy

von_neumann_entropy = deprecated(qis_von_neumann_entropy, ...)
```

kevinsung: Thanks, that gave a way to deprecate the functions. However, this doesn't let us deprecate `STATE_VECTOR_LIKE`. I just left that in the old file.

kevinsung: In order to deprecate `one_hot` and `eye_tensor` in `linalg.states`, I need to import `qis` in that file. But since `qis` depends on `linalg`, this causes a circular import, which is causing the MacOS test to fail.

cirq/__init__.py
```diff
@@ -38,6 +38,7 @@
 from cirq import (
     # Core
     circuits,
+    qis,
```
Strilanc: This should be lower. qis will be imported by more modules than this (e.g. sim would probably use it). Lower level goes lower.

kevinsung: Lower means higher up in the file, right? I put it after linalg.

kevinsung:
I was looking into deprecating `STATE_VECTOR_LIKE` and achieved a similar thing in OpenFermion-Cirq, see https://github.com/quantumlib/OpenFermion-Cirq/pull/375. Shall I do the same thing here? It is a bit ugly ...

kevinsung:
Oh, also, I haven't figured out a way to resolve the circular imports issue. Though I haven't actually spent that much time on it; been busy with other stuff.

kevinsung:
I have deprecated `sim.STATE_VECTOR_LIKE` using a module-editing hack and restored old code to `linalg.states` to avoid the circular import issue. Also, I have moved `von_neumann_entropy` to a new file, `qis.measures`.

kevinsung:
Hmm the module-editing hack is causing problems with `check/doctest`:
```
$ check/doctest -q
  File "dev_tools/run_doctest.py", line 233, in <module>
    main()
  File "dev_tools/run_doctest.py", line 217, in main
    quiet=quiet)
  File "dev_tools/run_doctest.py", line 84, in run_tests
    quiet=quiet)
  File "dev_tools/run_doctest.py", line 140, in load_tests
    tests = [make_test(file_path) for file_path in file_paths]
  File "dev_tools/run_doctest.py", line 140, in <listcomp>
    tests = [make_test(file_path) for file_path in file_paths]
  File "dev_tools/run_doctest.py", line 128, in make_test
    mod = import_file(file_path)
  File "dev_tools/run_doctest.py", line 202, in import_file
    spec.loader.exec_module(mod)  # type: ignore
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "cirq/sim/__init__.py", line 88, in <module>
    _sys.modules[__name__] = _wrap_module(_sys.modules[__name__],
KeyError: '__init__'
The command "check/doctest -q" exited with 1.
```

kevinsung:
The issue seems to be that in the doctest, files called `__init__` are executed as modules with the name `__init__`, but they exist in the `sys.modules` dictionary under the actual name of the module (which, for instance, would be `cirq.sim` for the file `cirq/sim/__init__.py`).

cduck:
It looks like this is broken because run_doctest bypasses the normal import machinery.  run_doctest could be modified to support module wrapping the same way `import` does.

Try changing [run_doctest.py#L202-203](https://github.com/quantumlib/Cirq/blob/master/dev_tools/run_doctest.py#L202) to
```python
mod_name = '<correct module name or dummy name>'
mod.__name__ = mod_name
sys[mod_name] = mod
spec.loader.exec_module(mod)
mod = sys[mod_name]
del sys[mod_name]
return mod
```

kevinsung:
@cduck I tried your suggestion, specifically:
```
    mod_name = 'dummy'
    mod.__name__ = mod_name
    sys.modules[mod_name] = mod
    spec.loader.exec_module(mod)
    mod = sys.modules[mod_name]
    del sys.modules[mod_name]
    return mod
```
and am getting
```
Loading tests   .Traceback (most recent call last):
  File "dev_tools/run_doctest.py", line 238, in <module>
    main()
  File "dev_tools/run_doctest.py", line 222, in main
    quiet=quiet)
  File "dev_tools/run_doctest.py", line 84, in run_tests
    quiet=quiet)
  File "dev_tools/run_doctest.py", line 140, in load_tests
    tests = [make_test(file_path) for file_path in file_paths]
  File "dev_tools/run_doctest.py", line 140, in <listcomp>
    tests = [make_test(file_path) for file_path in file_paths]
  File "dev_tools/run_doctest.py", line 128, in make_test
    mod = import_file(file_path)
  File "dev_tools/run_doctest.py", line 205, in import_file
    spec.loader.exec_module(mod)
  File "<frozen importlib._bootstrap_external>", line 724, in exec_module
  File "<frozen importlib._bootstrap_external>", line 800, in get_code
  File "<frozen importlib._bootstrap_external>", line 406, in _check_name_wrapper
ImportError: loader for __init__ cannot handle dummy
```

cduck:
@kevinsung I just tested this myself.  The code you have nearly works but `mod_name = ` needs to go before `spec_from_file_location`:

```python
def import_file(file_path: str) -> ModuleType:
    mod_name = 'cirq_doctest_module'
    # Find and create the module
    spec = importlib.util.spec_from_file_location(mod_name, file_path)
    mod = importlib.util.module_from_spec(spec)
    # Run the code in the module (but not with __name__ == '__main__')
    sys.modules[mod_name] = mod
    spec.loader.exec_module(mod)  # type: ignore
    del sys.modules[mod_name]
    return mod
```

I can reproduce the error you saw by using different values for `spec_from_file_location()` and `mod.__name__ = `.

kevinsung:
Thanks, @cduck, that worked.

dev_tools/run_doctest.py
```diff
@@ -192,14 +191,15 @@ def import_file(file_path: str) -> ModuleType:
 
     Returns: The imported module.
     """
-    mod_name = os.path.basename(file_path)
-    if mod_name.endswith('.py'):
-        mod_name = mod_name[:-3]
+    mod_name = 'cirq_doctest_module'
```
viathor, line 194: This doesn't look right: we're passing "cirq_doctest_module" as module name regardless of what module we're actually importing.

cduck: This is what it was doing before.  It used to import "filename" as a top-level module.  It imports each file as if it is an independent script so it doesn't interfere with the previously imported `cirq` module.

cirq/_compat.py
```diff
@@ -180,3 +181,34 @@ def decorated_func(*args, **kwargs) -> Any:
...
+
+    Returns:
+        Wrapped module with deprecated attributes.
+    """
+
+    class Wrapped(ModuleType):
+
+        def __getattr__(self, name):
+            if name in deprecated_attributes:
+                deadline, fix = deprecated_attributes[name]
```
cduck, line 205: Nit: You can unpack tuples in one line. `if name, (deadline, fix) in deprecated_attributes.items():`
Not important to fix for this PR.

PR #2830, opened 2020-03-10
    Update google gRPC Client to include reservation calls

dstrain115:
- Updates to generated protos and GAPIC clinet code
    - Includes new calls to create and list reservations

Many thanks to @lindmarkm for the initial generation of gapic files and figuring out how the generation tools worked.


cirq/google/engine/client/quantum_v1alpha1/proto/engine.proto
```diff
@@ -0,0 +1,705 @@
...
+import "google/cloud/quantum_v1alpha1/proto/quantum.proto";
+import "google/protobuf/duration.proto";
+import "google/protobuf/empty.proto";
+import "google/protobuf/field_mask.proto";
+import "google/api/client.proto";
+
+option go_package = "google.golang.org/genproto/googleapis/cloud/quantum/v1alpha1;quantum";
+option java_package = "com.google.cloud.quantum.engine.v1alpha1";
+
+// -
```
wcourtney: Should we remove these?

cirq/google/engine/client/quantum.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright 2019 Google LLC
+# Copyright 2020 Google LLC
```
lindmarkm: Do we care about it changing the copyright years?

cirq/google/engine/client/quantum_v1alpha1/types.py
```diff
@@ -0,0 +1,55 @@
+# -*- coding: utf-8 -*-
```
lindmarkm: We ran into issues with this file. I don't remember if it was actually running it, or mypy, or tests but that's why I deleted it and created the types/__init__.py. If this works now, let's delete the types/__init__.py otherwise let's delete this and add duration_pb2 to types/__init__.py.

cirq/google/engine/client/quantum.py
dstrain115: I don't think it matters too much, but I think the newer year is more correct and the tool updated the year, so I will keep it that way.

cirq/google/engine/client/quantum_v1alpha1/types.py
dstrain115: Yes, it looks like mypy can't handle this file since it is dynamically adding items to the namespace.  I have removed it and added duration_pb2 to the types/__init__.py

cirq/google/engine/client/quantum_v1alpha1/proto/engine.proto
dstrain115: Removed

Issue #2829, opened 2020-03-10
Move jupyter notebooks from `examples` to `examples/notebooks` directory


vtomole:
`QAOA_Demo.ipynb` and `Rabi_Demo.ipynb`. Also, are notebook names supposed to be cased like that? If so we should rename `quantum_walks_tutorial.ipynb`

joshp112358:
Hi, I can do this.

vtomole:
Fixed by https://github.com/quantumlib/Cirq/pull/2831

PR #2831, opened 2020-03-11
Issue #2829 - moved Jupyter notebooks into examples/notebooks

joshp112358:


joshp112358:
@googlebot I signed it!

PR #2833, opened 2020-03-11
Fix quantum engine gRPC path

dstrain115:
- gRPC path was incorrectly changed to the cirq proto path (cirq.google.engine...)
- It should be using the cloud path (google.cloud...)


maffoo:
How is the grpc code generated? Is the script to do that checked in somewhere? Also, as an action item, we should consider adding a CI check that regenerates proto/grpc code and checks that there's no diff with what was committed.

dstrain115:
> How is the grpc code generated? Is the script to do that checked in somewhere? Also, as an action item, we should consider adding a CI check that regenerates proto/grpc code and checks that there's no diff with what was committed.

They just changed the procedure to generate this code and it is extremely complicated, involving blaze, bazel, and manual adjustments.  It would be extremely difficult to do what you are asking.

That being said, I am hoping to put directions and part of the procedure into google3.

PR #2834, opened 2020-03-12
Add grid parallel two-qubit XEB

kevinsung:
This separates data collection and data analysis using idioms suggested by @mpharrigan (though note Matt that my definitions of `save` and `load` differ from the ones we've been using internally).


cirq/experiments/grid_parallel_two_qubit_xeb.py
```diff
@@ -0,0 +1,510 @@
...
+        # Compute the numerators and denominators
+        num_processors = min(num_processors, len(arguments))
+        with multiprocessing.Pool(num_processors) as pool:
+            pool.starmap(_get_fidelity_estimator_components, arguments)
+
+        # Calculate the results
+        for qubit_pair in all_active_qubit_pairs:
+            data = []
+            for depth in cycles:
+                fidelity = (sum(numerators[(qubit_pair, depth)]) /
```
viathor, line 442: For estimating fidelity from XEB data you should use one of the estimators in [fidelity_estimation.py](https://github.com/quantumlib/Cirq/blob/master/cirq/experiments/fidelity_estimation.py). Alternatively, if none of them fit your requirements, you should add your own there and call it from here.

Ideally, one should be able to use the code with any fidelity estimator they choose(*).

We should avoid having disjoint, incompatible sets of machinery for performing related tasks.

---
(*) There are good reasons to prefer different estimators in different circumstances, e.g. estimators generally have different variance.

kevinsung: I've defined a new fidelity estimator function, but I don't see a way to make the API compatible. The reason is that the estimator I'm using uses additional information to the measured bitstring probabilities; it also uses the probabilities for all possible bitstrings. Thus, this estimator is only practical for a small Hilbert space dimension.

cirq/experiments/fidelity_estimation.py
```diff
@@ -227,3 +227,120 @@ def log_xeb_fidelity(
...
+            observable on a uniformly random bitstring. The order should match
+            the order in the other arguments.
+    """
+    numerator = 0.0
+    denominator = 0.0
+    for m, e, u in zip(measured_expectations, exact_expectations,
+                       uniform_expectations):
+        numerator += (m - u) * (e - u)
+        denominator += (e - u)**2
+    return numerator / denominator
```
viathor, line 283: Optional: It may be interesting to also compute residuals to allow the user to evaluate how well the model above fits the data.

cirq/experiments/fidelity_estimation.py
```diff
@@ -227,3 +227,120 @@ def log_xeb_fidelity(
...
+
+        f = (∑_U (m_U - u_U) * (e_U - u_U)) / (∑_U (e_U - u_U)^2).
+
+    Args:
+        measured_expectations: A list of the m_U, the experimental estimates
+            of the observable, one for each circuit U.
+        exact_expectations: A list of the e_U, the exact value of the
+            observable. The order should match the order of the
+            `measured_expectations` argument.
+        uniform_expectations: A list of the u_U, the expectation of the
```
viathor, line 273: nit: s/A list/A sequence/ for agreement with type annotations.

(Applies to `measured_expectations` and `exact_expectations` above, too.)

cirq/experiments/fidelity_estimation.py
```diff
@@ -227,3 +227,120 @@ def log_xeb_fidelity(
...
+            observable on a uniformly random bitstring. The order should match
+            the order in the other arguments.
+    """
+    numerator = 0.0
+    denominator = 0.0
+    for m, e, u in zip(measured_expectations, exact_expectations,
+                       uniform_expectations):
```
viathor, line 280: This will silently truncate to the shortest of the sequences. Consider verifying their length are equal and raising a `ValueError` otherwise.

cirq/experiments/fidelity_estimation.py
```diff
@@ -227,3 +227,120 @@ def log_xeb_fidelity(
...
+    Args:
+        measured_expectations: A list of the m_U, the experimental estimates
+            of the observable, one for each circuit U.
+        exact_expectations: A list of the e_U, the exact value of the
+            observable. The order should match the order of the
+            `measured_expectations` argument.
+        uniform_expectations: A list of the u_U, the expectation of the
+            observable on a uniformly random bitstring. The order should match
+            the order in the other arguments.
+    """
```
viathor, line 276: nit: missing "Returns:" section (and "Raises:" once you raise `ValueError` - see comment below).

cirq/experiments/fidelity_estimation.py
```diff
@@ -227,3 +227,120 @@ def log_xeb_fidelity(
...
+        all_probabilities: Ideal probabilities of all possible bitstrings.
+            A list of lists, where each inner list contains the probabilities
+            for a single circuit, and should have length equal to the Hilbert
+            space dimension. The order of the lists should correspond to that
+            of `observed_probabilities`.
+        observable_from_probability: Function that computes the observable from
+            a given probability.
+        normalize_probabilities: Whether to multiply the probabilities by the
+            Hilbert space dimension before computing the observable.
+    """
```
viathor, line 320: nit: missing Returns: section of docstring

cirq/experiments/fidelity_estimation.py
```diff
@@ -227,3 +227,120 @@ def log_xeb_fidelity(
...
+        hilbert_space_dimension: int,
+        observed_probabilities: Sequence[Sequence[float]],
+        all_probabilities: Sequence[Sequence[float]],
+        observable_from_probability: Optional[Callable[[float], float]] = None,
+        normalize_probabilities: bool = True) -> float:
+    """Least squares fidelity estimator with observable based on probabilities.
+
+    Using the notation from the docstring of
+    `least_squares_xeb_fidelity_from_expectations`, this function computes the
+    least squares fidelity estimator when the observable O_U has eigenvalue
```
viathor, line 296: nit: s/estimator/estimate/

(This function is an *estimator* and what it computes is an *estimate*. Sorry for being pedantic.)

cirq/experiments/fidelity_estimation.py
```diff
@@ -227,3 +227,120 @@ def log_xeb_fidelity(
...
+            probabilities for a single circuit.
+        all_probabilities: Ideal probabilities of all possible bitstrings.
+            A list of lists, where each inner list contains the probabilities
+            for a single circuit, and should have length equal to the Hilbert
+            space dimension. The order of the lists should correspond to that
+            of `observed_probabilities`.
+        observable_from_probability: Function that computes the observable from
+            a given probability.
+        normalize_probabilities: Whether to multiply the probabilities by the
+            Hilbert space dimension before computing the observable.
```
viathor, line 319: This argument has a very narrow use and complicates the API. We can make the API simpler and at the same time more flexible by removing `normalize_probabilities` and instead changing the expected signature of `observable_from_probability` from

```python
Callable[[float], float]
```

to

```python
Callable[[float, int], float]
```

where the second argument to `observable_from_probability` is the dimension of the Hilbert space. This means that identity which is currently specified as `lambda p: p` below becomes `lambda p, _: p`. Thus, user is free to ignore the second argument if their function is dimension-independent, but can also use it to normalize the first argument (probability) or to do something else (hence extra flexibility of this API).

cirq/experiments/fidelity_estimation.py
kevinsung: I've modified this function so that it outputs a tuple where the first element is the fidelity and the second element is the list of residuals.

cirq/experiments/fidelity_estimation.py
kevinsung: Done.

cirq/experiments/fidelity_estimation.py
kevinsung: Done.

cirq/experiments/fidelity_estimation.py
kevinsung: Done.

cirq/experiments/fidelity_estimation.py
kevinsung: Done.

cirq/experiments/fidelity_estimation.py
kevinsung: Done.

cirq/experiments/fidelity_estimation.py
kevinsung: It's crucial for performance that `observable_from_probability` is allowed to be a built-in Numpy ufunc, since it is applied to Numpy arrays that are likely to have thousands to millions of elements. Your proposed change would preclude that.

kevinsung:
pytest-cov seems unable to tell that code used by a multiprocessing pool is actually covered.

viathor:
You can use `# coverage: ignore` (with the rationale in a comment) to disable coverage check.

cirq/experiments/fidelity_estimation.py
viathor: Makes sense.

cirq/experiments/fidelity_estimation.py
```diff
@@ -227,3 +227,148 @@ def log_xeb_fidelity(
...
+
+        of the least squares minimization.
+
+    Raises:
+        ValueError: The lengths of the input sequences are not all the same.
+    """
+    if not (len(measured_expectations) == len(exact_expectations) ==
+            len(uniform_expectations)):
+        raise ValueError('The lengths of the sequences of expectations must '
+                         'all be the same.')
```
viathor, line 291: In practice, it is often helpful to have the lengths in the error message. This allows to distinguish among common causes of problems with list lengths, e.g. empty list, off-by-one errors etc.

cirq/experiments/fidelity_estimation.py
```diff
@@ -227,3 +227,148 @@ def log_xeb_fidelity(
...
+            observable_from_probability = lambda p: p
+        else:
+            observable_from_probability = np.frompyfunc(
+                observable_from_probability, 1, 1)
+    observable_from_probability = cast(Callable, observable_from_probability)
+    measured_expectations = []
+    exact_expectations = []
+    uniform_expectations = []
+    prefactor = hilbert_space_dimension if normalize_probabilities else 1.0
+    for observed, all_ in zip(observed_probabilities, all_probabilities):
```
viathor, line 360: nit: I wonder whether we can come up with a better name for "all_". Maybe "observed" -> "observed_probs" and "all_" -> "all_probs"?

cirq/experiments/fidelity_estimation.py
kevinsung: Done.

cirq/experiments/fidelity_estimation.py
kevinsung: Done.

Issue #2802, opened 2020-03-03
Offer better guidance on how to set --n_clifford_trials when doing direct fidelity estimation


tonybruguier-google:
During the review of PR#2709 Matthew Harrigan (mpharrigan) suggested that the number of Pauli states sampled (when having a Clifford circuit) should be more than just a flag that the user has to set without any guidance.

Instead, we could add an exhaustive option and do numerical studies. See the full comment there:
https://github.com/quantumlib/Cirq/pull/2709#discussion_r386717424

dabacon:
`n_clifford_trials` is now called `n_measured_operators`.

mhucka:
Due to the age of this issue and the need to prioritize other work, I regret I feel the need to close this issue. If this is wrong and someone else feels this should be reopened, please go ahead.

PR #2837, opened 2020-03-12
Add ability to do exhaustive Clifford trial

tonybruguier-google:
This is a first step in addressing issue #2802 

Introduce a way to do an exhaustive enumeration of all the states for Clifford circuit. Whether or not to use the exhaustive enumeration is not solved by this PR.


examples/direct_fidelity_estimation.py
```diff
@@ -109,7 +109,7 @@ def _estimate_pauli_traces_clifford(n_qubits: int,
         n_qubits: An integer that is the number of qubits.
         clifford_state: The basis of the Pauli states with non-zero probability.
         n_clifford_trials: An integer that is the number of Pauli states to
-            sample.
+            sample. If set to 0, we compute the exhaustive list of Pauli states.
```
Strilanc: Zero seems like the wrong value for this. I'd expect zero to return the maximally uninformed prior instead of doing exhaustive search.

Maybe what we should do instead is use a completely different value, like `float("inf")` or `"all"` or even a separate argument `exhaustive=True` that can't be specified simultaneously. Or perhaps this should even be a completely different method. This seems like a rather large change in behavior.

tonybruguier-google:
Thanks. I looked into using "inf" but that seemed a bit strange because it had to be a float. Instead, I decided to keep an int and signify exhaustivity with sys.maxsize. I hope it's OK with you, but I can change the code if you don't like this.

tonybruguier-google:
Friendly ping. No hurry at all.

examples/direct_fidelity_estimation.py
cduck: @tonybruguier-google Why not just use `None`?  This is the usual way in Python to indicate the value isn't specified.

tonybruguier-google:
Thanks. PTAL.

tonybruguier-google:
Hello,
Does anyone have further comments on this?

examples/direct_fidelity_estimation.py
```diff
@@ -129,14 +129,27 @@ def _estimate_pauli_traces_clifford(n_qubits: int,
...
 
-    pauli_traces = []
-    for _ in range(n_clifford_trials):
-        # Build the Pauli string as a random sample of the basis elements.
-        dense_pauli_string = cirq.DensePauliString.eye(n_qubits)
-        for stabilizer in stabilizer_basis:
-            if np.random.randint(2) == 1:
-                dense_pauli_string *= stabilizer
+    dense_pauli_strings = []
+    if n_clifford_trials is not None:
```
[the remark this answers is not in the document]
tonybruguier-google: Done.

tonybruguier-google:
Thanks. PTAL.

tonybruguier-google:
@Strilanc Craig, do you have any more comment on this PR? I tried to address your request for changes and I'm happy to do more if it's not what you wanted.

PR #2839, opened 2020-03-13
Fix parameter order swap for gRPC list calls.

lindmarkm:


dstrain115:
Looks like some failing tests and formatting, but, yeah we should fix this.

dstrain115:
This is now fixed in https://github.com/quantumlib/Cirq/pull/2838

PR #2850, opened 2020-03-20
Convert documentation to use nbsphinx

dstrain115:
- Converts documentation to use only ipynb where it exists
- Instead of a duplicate ipynb and md file, only a ipynb file will
exist.
- There was a problem with circuits.ipynb.  I deleted it for now (leaving the md file) and will add it back in a subsequent PR.

- Also fixes some goofy errors I was getting while trying to get sphinx
to build docs


mpharrigan:
How does readthedocs know that it needs to install nbsphinx to build the docs?

dstrain115:
> How does readthedocs know that it needs to install nbsphinx to build the docs?

I added it to dev_tools/conf/pip-list-dev-tools.txt.  That causes build_docs to work.
Will this be enough?

mpharrigan:
so `build_docs` isn't a 100% faithful test (flashbacks to when the docs were failing to build on readthedocs because the instances didn't have enough RAM to install tensorflow), but I did trace the requirements and updating pip-list-dev-tools should be sufficient.

For anyone following along:
 - readthedocs looks for `.readthedocs.yml` (or docs/requirements.txt, which we don't have)
 - `.readthedocs.yml` specified `dev_env` as an extra configuration
 - `setup.py` reads pip-list-dev-tools.txt as the extra requirements for the `dev_env` configuration

PR #2851, opened 2020-03-20
Restore single-Clifford inversion in RB benchmarking.

### LINKED ISSUE #2846: Increased length of RB circuits?
In #1712, the code to manually construct the terminal "inversion" Clifford in randomized benchmarking was replaced with a call to `protocols.inverse`:
https://github.com/quantumlib/Cirq/blob/afad82b688a93a12976b3890392e59671e0bb2bf/cirq/experiments/qubit_characterizations.py#L527-L534

In my tests, `protocols.inverse` doesn't create a single gate, but rather creates an inverted copy of the entire circuit (i.e. each gate is inverted separately and their order is reversed). I would expect a noisy simulator to see roughly half as much noise if the inversion step was a single gate - does this change affect the RB experiment when running on hardware?

95-martin-orion:
Fixes #2846.

This restores the "invert circuit with a single Clifford" behavior previously used in the benchmark without significantly increasing test runtime. For review, compare with the state of the code before #1712; the main change is memoizing subsets of the two-qubit Clifford unitaries.

Runtime comparison of `cirq/experiments/qubit_characterizations_test.py`:
- Before this change: **~1s**
- Clean rollback of #1712: **~45s**
- Partial rollback of #1712, no memoization: **~12s**
- After this change: **~1.5s**


95-martin-orion:
This is a blocker for testing the noise model generator (#2715) - please take a look.

PR #2852, opened 2020-03-21
Add stratified_circuit method

Strilanc:
This can be a handy way to get a circuit into a nice form, or to guarantee that measurements happen on their own.


cirq/optimizers/stratify.py
```diff
@@ -0,0 +1,122 @@
...
+            assert set(cur_frontier.values()).issubset({len(circuit)})
+            break
+        prev_frontier = cur_frontier
+
+    # Re-pack operations within each section, then concatenate into result.
+    result = circuits.Circuit()
+    for f1, f2 in zip(frontiers, frontiers[1:]):
+        result += circuits.Circuit(
+            op for _, op in circuit.findall_operations_between(f1, f2)
+        )
```
viathor: Optional: This would probably look better if you made consistent use of comprehensions: for both loops or for none of them. Maybe something like:

```python
return circuits.Circuit(
    op for f1, f2 in zip(frontiers, frontiers[1:])
    for _, op in circuit.findall_operations_between(f1, f2)
)
```

cirq/optimizers/stratify.py
```diff
@@ -0,0 +1,122 @@
...
+
+def stratified_circuit(
+        circuit: 'cirq.Circuit',
+        *,
+        categories: Iterable[Union[
+            'cirq.Gate',
+            'cirq.Operation',
+            Type['cirq.Gate'],
+            Type['cirq.Operation'],
+            Callable[['cirq.Operation'], bool]
```
viathor: Optional: I think it'd be useful to also allow single callable that returns a category (e.g. as `int`). This would allow you to pass in a simple thing like `lambda g: len(g.qubits)` instead of an iterable of callables.

This is surprisingly tricky.

cirq/optimizers/stratify.py
```diff
@@ -0,0 +1,122 @@
...
+
+from typing import TYPE_CHECKING, Type, Callable, Union, Iterable
+
+from cirq import ops, circuits
+
+if TYPE_CHECKING:
+    import cirq
+
+
+def stratified_circuit(
```
viathor: There is a well-known convention that calls for verb-based function names. I think we haven't consistently followed it, but here adherence is trivial: just rename to `stratify_circuit`.

This will make cirq feel more familiar to all folks who know the convention :-)

cirq/optimizers/stratify.py
Strilanc: Ah, a `key=` parameter. That's an interesting idea. It would be a pretty big change to the implementation.

cirq/optimizers/stratify.py
Strilanc: The issue with `stratify_circuit` is that circuit is mutable, and that name implies we are mutating the original circuit instead of returning a new one.

cirq/optimizers/stratify.py
Strilanc: This implementation is incorrect. It loses the stratification.

viathor: Oops, you're right.

cirq/optimizers/stratify.py
viathor: I guess `make_stratified_circuit` would work, but on second thought I think the current name fits well with python's other conventions (c.f. `sorted`, `reversed` etc).

cirq/optimizers/stratify.py
```diff
@@ -0,0 +1,108 @@
...
+from cirq import ops, circuits
+
+if TYPE_CHECKING:
+    import cirq
+
+
+def stratified_circuit(circuit: 'cirq.Circuit', *, categories: Iterable[
+        Union['cirq.Gate', 'cirq.Operation', Type['cirq.Gate'],
+              Type['cirq.Operation'], Callable[['cirq.Operation'], bool]]]
+                      ) -> 'cirq.Circuit':
```
viathor: No action needed: Yapf made the function signature less readable.

PR #2856, opened 2020-03-25
Add Pauli expansion for PhasedXZ

viathor:
Tweaks to the formulas do cause test failures which confirms that the unit tests run consistency checks.


cirq/ops/phased_x_z_gate.py
```diff
@@ -165,6 +165,23 @@ def _phase_by_(self, phase_turns, qubit_index) -> 'cirq.PhasedXZGate':
...
+        phase = np.exp(1j * (x_angle + z_angle))
+
+        cx = np.cos(x_angle)
+        sx = np.sin(x_angle)
+        return value.LinearDict({
+            'I': phase * cx * np.cos(z_angle),
+            'X': -1j * phase * sx * np.cos(z_angle + axis_angle),
+            'Y': -1j * phase * sx * np.sin(z_angle + axis_angle),
+            'Z': -1j * phase * cx * np.sin(z_angle),
+        })  # yapf: disable
```
vtomole: How ugly does this look when enabled?

viathor: Yapf breaks it after colons which means the equations no longer line up. I like the alignment since it allows the reader to quickly recognize where the formulas differ.

PR #2860, opened 2020-03-25
Use UTC time for reservation interactions.

wcourtney:


cirq/google/engine/engine_processor.py
```diff
@@ -223,7 +224,7 @@ def remove_reservation(self, reservation_id: str):
             raise ValueError('Cannot determine freeze_schedule from processor.'
                              'Call _cancel_reservation or _delete_reservation.')
         secs_until = (reservation.start_time.seconds -
-                      int(datetime.datetime.now().timestamp()))
+                      int(datetime.datetime.now(tz=utc).timestamp().timestamp()))
```
dstrain115: Why are there two timestamps() here?

dstrain115:
I am wondering why this matters.  It should be turned into epoch timestamps everywhere which should be timezone independent.

cirq/google/engine/engine_processor.py
wcourtney: Apparently I hadn't pushed the fix for this. \Done.

wcourtney:
> I am wondering why this matters. It should be turned into epoch timestamps everywhere which should be timezone independent.

The epoch timestamps are created from localized dates. Specifically, the PR is a bugfix for operation In colab, where datetime.now() provides a timestamp that is 7 hours later than running locally in PT. I've resolved that on the client side for the parameterized values, remove_timeslot tries to cancel reservations that it should delete and fails.

PR #2861, opened 2020-03-26
Add GridQid class

mmcewen-g:
Adds GridQid class to match LineQid, to support grid systems with dimension !=2


cirq/devices/grid_qubit.py
```diff
@@ -12,43 +12,27 @@
...
-
-    New GridQubits can be constructed by adding or subtracting tuples
-
-        >>> cirq.GridQubit(2, 3) + (3, 1)
-        cirq.GridQubit(5, 4)
 
-        >>> cirq.GridQubit(2, 3) - (1, 2)
-        cirq.GridQubit(1, 1)
-    """
+@functools.total_ordering
```
Strilanc: Why `total_ordering`?

cirq/devices/grid_qubit.py
```diff
@@ -82,6 +64,197 @@ def neighbors(self,
                 neighbors.add(q)
         return neighbors
 
+    @abc.abstractmethod
+    def _with_row_col(self: TSelf, row: int, col: int) -> TSelf:
+        '''Returns a qid with the same type but a different coordinate.'''
```
Strilanc: Replace all `'''` with `"""`

cirq/devices/grid_qubit.py
```diff
@@ -82,6 +64,197 @@ def neighbors(self,
...
+
+    def __add__(self: TSelf, other: Tuple[int, int]) -> 'TSelf':
+        if isinstance(other, _BaseGridQid):
+            return self._with_row_col(row=self.row + other.row, col=self.col + other.col)
```
Strilanc: Raise an exception if the dimension doesn't match.

cirq/devices/grid_qubit.py
```diff
@@ -82,6 +64,197 @@ def neighbors(self,
...
+        if not (isinstance(other, tuple) and len(other) == 2 and
+                all(isinstance(x, int) for x in other)):
+            raise TypeError(
```
Strilanc: Nit: we can now use f-strings.

cirq/devices/grid_qubit.py
```diff
@@ -82,6 +64,197 @@ def neighbors(self,
...
+    New GridQid can be constructed by adding or subtracting tuples
+
+        >>> cirq.GridQubit(2, 3, dimension=2) + (3, 1)
+        cirq.GridQubit(5, 4, dimension=2)
+
+        >>> cirq.GridQubit(2, 3, dimension=2) - (1, 2)
+        cirq.GridQubit(1, 1, dimension=2)
+    """
+
+    def __init__(self, row: int, col: int, dimension: int) -> None:
```
Strilanc: I think we should make dimension a forced keyword argument by putting `*` beforehand. This will avoid confusion where GridQid(3, 2, 4) could be seen as having dimension 3.

cirq/devices/grid_qubit.py
```diff
@@ -82,6 +64,197 @@ def neighbors(self,
...
+
+        Args:
+            diameter: Length of a side of the square
+            top: Row number of the topmost row
+            left: Column number of the leftmost row
+            dimension: The dimension of the qid, e.g. the number of quantum
+                levels.
+
+        Returns:
+            A list of GridQubits filling in a square grid
```
Strilanc: GridQids

cirq/devices/grid_qubit.py
```diff
@@ -82,6 +64,197 @@ def neighbors(self,
...
+        Args:
+            rows: Number of rows in the rectangle
+            cols: Number of columns in the rectangle
+            top: Row number of the topmost row
+            left: Column number of the leftmost row
+            dimension: The dimension of the qid, e.g. the number of quantum
+                levels.
+
+        Returns:
+            A list of GridQubits filling in a rectangular grid
```
Strilanc: GridQids

cirq/devices/grid_qubit.py
```diff
@@ -82,6 +64,197 @@ def neighbors(self,
...
+        """
+        return [
+            GridQid(row, col, dimension=dimension)
+            for row in range(top, top + rows)
+            for col in range(left, left + cols)
+        ]
+
+    @staticmethod
+    def from_diagram(diagram: str, dimension: int) -> List['GridQid']:
```
Strilanc: Force keyword argument on dimension

Strilanc: Return type is List['GridQid']

cirq/devices/grid_qubit.py
```diff
@@ -82,6 +64,197 @@ def neighbors(self,
...
+            A list of GridQid corresponding to qids in the provided diagram
+
+        Raises:
+            ValueError: If the input string contains an invalid character.
+        """
+        coords = _ascii_diagram_to_coords(diagram)
+        return [GridQid(*c, dimension=dimension) for c in coords]
+
+    def __repr__(self):
+        return 'cirq.GridQid({}, dimension={})'.format(self.x, self.dimension)
```
Strilanc: Nit: f-string

Strilanc: row, col not x

cirq/devices/grid_qubit.py
```diff
@@ -82,6 +64,197 @@ def neighbors(self,
...
+
+    def __str__(self):
+        return '({}, {}) (d={})'.format(self.row, self.col, self.dimension)
+
+    def _json_dict_(self):
+        return protocols.obj_to_dict_helper(self, ['x', 'dimension'])
```
Strilanc: row, col, dimension

You'll also need to add an entry into the json serialization code. (The tests will enforce that this happens.)

cirq/devices/line_qubit.py
```diff
@@ -66,12 +66,16 @@ def _with_x(self: TSelf, x: int) -> TSelf:
         '''Returns a qubit with the same type but a different value of `x`.'''
 
     def __add__(self: TSelf, other: int) -> TSelf:
+        if isinstance(other, _BaseLineQid):
```
Strilanc: If the dimension is different, raise an error.

cirq/devices/grid_qubit.py
mmcewen-g: jbc that's what _BaseLineQid uses; I was confused bc it seemed like Qid implements all the rich comparison functions, but figured there was a reason for it to be there: possibly removable from both _BaseLineQid and _BaseGridQid?

Strilanc: It might be a performance optimization somehow. But normally it requires you to define e.g. `__lt__`

Issue #2807, opened 2020-03-04
Support more noise types in metrics-to-noise converters


95-martin-orion:
Following PR #2715, the `simple_noise_from_calibration_metrics` will support three noise types:
- depolarization during gates
- qubit decay during readout
- error during readout (bitflip)

This issue tracks the addition of other noise types to this converter. Candidates include:
- amplitude damping during gates
- qubit excitation during readout

95-martin-orion:
+ @dstrain115 for insights into calibration metrics.

One of the challenges in supporting per-gate amplitude damping is that noise models do not have device information when they are applied to a set of moments. This prevents an accurate estimation of the duration of each moment, which makes applying time-dependent amplitude damping difficult.

95-martin-orion:
With the addition of two-qubit Pauli error metrics (PR #2761) we can model gate depolarization more precisely by controlling depolarization probability on the gate type.

95-martin-orion:
> With the addition of two-qubit Pauli error metrics (PR #2761) we can model gate depolarization more precisely by controlling depolarization probability on the gate type.

In that PR, @viathor notes that two-qubit Pauli error is defined as the chance of _at least_ one qubit experiencing error, modeled by the formula:
```
\rho -> (1 - r_p) \rho + r_p sum(k=1..d^2-1, P_k \rho P_k) / (d^2 - 1)
```

Since this model is symmetric, with each possible combination of Paulis (aside from the identity) having equal probability `r_p / (4^n  - 1)`, the single-qubit error rate `r_pk` should come out to be:
```
r_pk = [r_p / (4^n - 1)] * [3 * 4^(n-1)]
```
since there are `3 * 4^(n-1)` combinations of Pauli errors on `n` qubits which have an error on a given qubit k.

95-martin-orion:
Copying discussion from #2864:

> @mpharrigan: The noise models are _models_. All models are wrong but some models are useful. Instead of beating ourselves over the head trying to get the current zero-degrees-of-freedom model to be predictive, free the parameters and see what you can get. [Normal caveats apply: make sure you have separate test/train/validation sets]

> @95-martin-orion: This is a good point. IIUC, what you're suggesting would look something like this:
>
> 1. Create a model (e.g. the existing depol+readout bitflip/decay) with parameters m0, m1, etc.
>     a. This model will inevitably exclude some sources of error
> 2. Use metrics to get reasonable starting values for those parameters
[...]

PR #2864, opened 2020-03-27
[WIP] Add tests for generated noise models

95-martin-orion:
The intent of these tests is to assert that the error observed in calibration runs falls within 3 sigma of error in simulated runs with generated noise models. For the moment, tests simply print out result data so that assertions don't prevent later tests from running.

Current state of these tests:

- Expected fidelity (calculated from calibration data) with depolarization noise is considerably lower than the observed (simulated) fidelity. I suspect that this is caused by our use of `single_qubit_rb_total_error` as opposed to the (as-yet unavailable) `single_qubit_rb_pauli_error`.
- Expected p00 and p11 with readout error falls within 3 sigma of the simulated error.

Known issues:
- As stated above, we need to switch over to `single_qubit_rb_pauli_error` for depol noise.
- In order for these tests to have a sensible duration, we can't run many circuits; this results in a very wide 3 sigma window. (3 sigma was chosen to minimize flakes)


95-martin-orion:
Note that this includes PR #2851, which significantly affects the depol noise in calibration (in short, it cuts the number of Cliffords used in half)

95-martin-orion:
Now includes fixes from #2927.

Trial runs of this test show estimated RB error from this model ranging from "just barely inside acceptable bounds" to "wildly inaccurate". Notably, the model always falls on the side of predicting more error than there actually is.

Readout error predictions seem considerably better, although still not consistently within acceptable bounds. This may not be trustworthy, given that the only source of p00-type error from this model is readout error (i.e. no <1|U|0> state-change error).

mpharrigan:
This is extremely interesting. What do you count as acceptable? Often times having something is better than nothing.

I'm still personally in the camp of: it's a fools errand to try to shoehorn experimentally measured parameters into related-but-certainly-not-the-same-thing noise model parameters. Why not tweak the noise model parameters to make it match better? Automate this tweaking with an optimizer and now you've got a machine-learned error model :)

95-martin-orion:
> This is extremely interesting. What do you count as acceptable?

The current test determines a 3-sigma confidence interval for total error from the metrics observed in the experiment. If the model is accurate, I would expect the error in the real metrics to fall within this interval ~99% of the time; as it stands, the Pauli error consistently falls outside of the interval.

Data from the test is below. Values on the left are expected total fidelity for 10 / 100 Clifford circuits using the reported metrics; the intervals on the right are +/- 3 sigma measured from the generated model.
```
Randomized benchmark results ((5, 2)):
0.949805502854814 is not within (0.9599152606041494, 0.9708470533133249)
0.6232262123972939 is not within (0.7588668730294142, 0.7840445067288034)
Randomized benchmark results ((5, 3)):
0.9615741373641938 is not within (0.9710718621084208, 0.9802719284800496)
0.6978320705630213 is not within (0.8257194843814142, 0.8478742812295359)
Randomized benchmark results ((6, 2)):
0.9538745702825574 is not within (0.9581957525795256, 0.9693694387482794)
0.6481754507884451 is not within (0.7811969582931437, 0.8054750369111724)
Randomized benchmark results ((6, 3)):
0.9654292442489797 is not within (0.9727123939000851, 0.9816286991162003)
0.7239456298875216 is not within (0.8226408287354804, 0.84495833201984)
```

> Often times having something is better than nothing.

My main concern here is false positives: as seen above, the fidelity from metrics is noticeably below what the model reports.

> I'm still personally in the camp of: it's a fools errand to try to shoehorn experimentally measured parameters into related-but-certainly-not-the-same-thing noise model parameters.

Which parameters in particular are you referring to? I'm aware that using 1-qubit metrics to generate noise for 2-qubit gates is ill-advised - are there any others?

> Why not tweak the noise model parameters to make it match better? Automate this tweaking with an optimizer and now you've got a machine-learned error model :)

Could you clarify this proposal? The current state suggests to me that we're missing parameters (e.g. gate dephasing, proper 2-qubit gate handling, etc.), which can't be fixed by tweaking existing parameters.

mpharrigan:
Please correct me if I'm wrong on any of this, but:

 - readout noise: using p00 and p11 to parameterize an amplitude damping and bitflip channel. This isn't a valid mapping. Second order concerns: p00 and p11 measurements contain state prep errors
 - RB: compiling cliffords to hardware gates violate RB's theoretical assumptions
 - The two qubit gate thing still :) 

The noise models are *models*. All models are wrong but some models are useful. Instead of beating ourselves over the head trying to get the current zero-degrees-of-freedom model to be predictive, free the parameters and see what you can get. [Normal caveats apply: make sure you have separate test/train/validation sets]

95-martin-orion:
> readout noise: using p00 and p11 to parameterize an amplitude damping and bitflip channel.

p00 and p11 are only used for verification since they're explicitly captured in the test. The model is built using the readout separation metric. (Agreed that this fails to capture state-prep errors, though)

The other issues listed all seem valid to me.

> All models are wrong but some models are useful. {...} Free the parameters and see what you can get.

This is a good point. IIUC, what you're suggesting would look something like this:

1. Create a model (e.g. the existing depol+readout bitflip/decay) with parameters m0, m1, etc.
    a. This model will inevitably exclude some sources of error
2. Use metrics to get reasonable *starting* values for those parameters
3. Set up an optimizer to fit the parameters to observed results
    a. This will necessarily "overtune" some parameters to make up for the excluded error types

On a conceptual level this sounds good, it's just outside the scope of the current model structure. I'll copy this up to #2807 to keep tracking this discussion.

(FWIW: I'm happy to close this PR without merging - having prompting this discussion, it accomplished its purpose)

mpharrigan:
Yup, I think that would be cool! Yup, it's probably out of scope for the current work :)

balopat:
> (FWIW: I'm happy to close this PR without merging - having prompting this discussion, it accomplished its purpose)

As all the issues are opened to track this work I'm closing this. Please reopen as needed.

PR #2870, opened 2020-03-29
Refactor DFE to run samplers

tonybruguier-google:


examples/direct_fidelity_estimation.py
```diff
@@ -207,20 +207,29 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
...
+            of the activity. Useful when running experiments and we want to
+            store data.
 
     Returns:
         The estimated fidelity.
     """
     # n_trials is upper-case N in https://arxiv.org/abs/1104.3835
 
+    if log_output_filename:
+        sys.stdout = open(log_output_filename, "w")
```
viathor: This breaks the [principle of separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns). Function for direct fidelity estimation should not be responsible for setting up logging.

Perhaps we don't really need to print anything in this function. It seems that what we print are the inputs and return value anyway.

tonybruguier-google: I had a concern that it might not be the right way. I reverted the changes to the logging, and instead narrowed the scope to only the refactoring.

examples/direct_fidelity_estimation.py
```diff
@@ -63,10 +63,11 @@ def compute_characteristic_function(circuit: cirq.Circuit,
...
 
 
-async def estimate_characteristic_function(
-        circuit: cirq.Circuit, pauli_string: cirq.PauliString,
-        qubits: List[cirq.Qid], simulator: cirq.DensityMatrixSimulator,
-        samples_per_term: int):
+async def estimate_characteristic_function(circuit: cirq.Circuit,
+                                           pauli_string: cirq.PauliString,
+                                           qubits: List[cirq.Qid],
+                                           sampler: cirq.Sampler,
```
viathor: Args section of the docstring below needs to be updated.

examples/direct_fidelity_estimation.py
```diff
@@ -280,15 +281,18 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
...
                                                  samples_per_term))
         else:
             sigma_i, _ = compute_characteristic_function(
                 circuit, measure_pauli_string, qubits, noisy_density_matrix)
 
         fidelity += Pr_i * sigma_i / rho_i
 
-    return fidelity / n_trials * d
+    estimated_fidelity = fidelity / n_trials * d
+    print('Estimated fidelity: %f' % (estimated_fidelity))
```
viathor: Please remove.

examples/direct_fidelity_estimation.py

examples/direct_fidelity_estimation.py

tonybruguier-google:
Thanks. I don't think I have permissions to merge so I'll wait for the bot :)

examples/direct_fidelity_estimation.py
```diff
@@ -327,15 +328,15 @@ def main(*, n_trials: int, n_clifford_trials: int, samples_per_term: int):
...
 
     estimated_fidelity = direct_fidelity_estimation(
         circuit,
         qubits,
-        noise,
+        noisy_simulator,
         n_trials=n_trials,
         n_clifford_trials=n_clifford_trials,
         samples_per_term=samples_per_term)
-    print('Estimated fidelity: %f' % (estimated_fidelity))
```
viathor: You probably want to keep this. 

(Note that this one is different from the other printfs since it's in the main() function - outputing the estimated fidelity is the objective of this script.)

examples/direct_fidelity_estimation.py
```diff
@@ -207,7 +208,7 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
     Args:
         circuit: The circuit to run the simulation on.
         qubits: The list of qubits.
-        noise: The noise model when doing a simulation.
+        sampler: Either a noisy simulator or an engine.
```
viathor: I just noticed: you're still referring to the old 'noise' parameter on line 215 below, could you update the comment appropriately?

examples/direct_fidelity_estimation.py
```diff
@@ -260,11 +261,11 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
     # above) does bother it, so we re-normalize the probs.
     p /= np.sum(p)
 
-    noisy_simulator = cirq.DensityMatrixSimulator(noise=noise)
     fidelity = 0.0
 
     if samples_per_term == 0:
         # sigma in https://arxiv.org/abs/1104.3835
+        noisy_simulator = cast(cirq.DensityMatrixSimulator, sampler)
```
viathor: We should raise an exception if sampler is not of the right type.

examples/direct_fidelity_estimation.py

examples/direct_fidelity_estimation.py
tonybruguier-google: Added an assert

viathor: Asserts should be used for conditions that are true regardless of input parameters due to internal consistency of the algorithm (e.g. [loop invariants](https://en.wikipedia.org/wiki/Loop_invariant)). Assert failure should indicate a bug.

In this case, the condition may be false due to user input. This should be handled by raising an exception that tells the caller what they did wrong.

examples/direct_fidelity_estimation.py
```diff
@@ -207,13 +208,14 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
...
+        sampler: Either a noisy simulator or an engine.
         n_trial: The total number of Pauli measurements.
         n_clifford_trials: In case the circuit is Clifford, we specify the
             number of trials to estimate the noise-free pauli traces.
-        samples_per_term: is set to 0, we use the 'noise' parameter above and
+        samples_per_term: is set to 0, we use the 'sampler' parameter above as
+            a noise (must be of type cirq.DensityMatrixSimulator) and
             simulate noise in the circuit. If greater than 0, we ignore the
-            'noise' parameter above and instead run an estimation of the
+            'sampler' parameter above and instead run an estimation of the
```
viathor: If `samples_per_term` is greater than zero, we don't actually ignore `sampler` - we use it in the course of estimation of the characteristic function (to obtain the values of the Pauli observables involved).

tonybruguier-google:
Thanks. PTAL.

examples/direct_fidelity_estimation.py
```diff
@@ -260,11 +261,12 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
     # above) does bother it, so we re-normalize the probs.
     p /= np.sum(p)
 
-    noisy_simulator = cirq.DensityMatrixSimulator(noise=noise)
     fidelity = 0.0
 
     if samples_per_term == 0:
         # sigma in https://arxiv.org/abs/1104.3835
+        assert type(sampler) == cirq.DensityMatrixSimulator
```
viathor: Please raise an exception instead of asserting.

examples/direct_fidelity_estimation.py
viathor: This isn't done, see comment two lines above.

examples/direct_fidelity_estimation.py
```diff
@@ -207,15 +208,15 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
...
+        sampler: Either a noisy simulator or an engine.
         n_trial: The total number of Pauli measurements.
         n_clifford_trials: In case the circuit is Clifford, we specify the
             number of trials to estimate the noise-free pauli traces.
-        samples_per_term: is set to 0, we use the 'noise' parameter above and
-            simulate noise in the circuit. If greater than 0, we ignore the
-            'noise' parameter above and instead run an estimation of the
-            characteristic function.
-
+        samples_per_term: is set to 0, we use the 'sampler' parameter above as
```
viathor: nit: replace "is set to 0" with "if set to 0"

examples/direct_fidelity_estimation.py
tonybruguier-google: Like this?

examples/direct_fidelity_estimation.py
tonybruguier-google: Sorry. I should have been more careful. How about now?

examples/direct_fidelity_estimation.py

tonybruguier-google:
Sorry about this. I think I hadn't pushed on time.

I checked that the changes are in. Thank you for the careful review. PTAL, no hurry.

examples/direct_fidelity_estimation.py
```diff
@@ -260,11 +261,13 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
     # above) does bother it, so we re-normalize the probs.
     p /= np.sum(p)
 
-    noisy_simulator = cirq.DensityMatrixSimulator(noise=noise)
     fidelity = 0.0
 
     if samples_per_term == 0:
         # sigma in https://arxiv.org/abs/1104.3835
+        if not isinstance(sampler, cirq.DensityMatrixSimulator):
+            raise TypeError('sampler is not a cirq.DensityMatrixSimulator')
```
viathor: nit: maybe add "..., but samples_per_term is zero" to guide the user towards the correct parameters.

viathor: Test coverage failed. You need a unit test that verifies the exception is raised in the right circumstances.

tonybruguier-google:
Sorry for the double-push. My first one failed the check, so I was waiting to comment but now I can because the second push passed.

PTAL.

PR #2873, opened 2020-03-30
Three qubit decomposition

### LINKED ISSUE #451: Implement general decomposition of 3-qubit unitary gates
This would allow ControlledGates made from 2-qubit KnownMatrixGates to automatically run on the simulator.

balopat:
Adds cirq.three_qubit_unitary_to_operations. 

For the design doc, please see the attached [3qubit-decomposition.pdf](https://github.com/quantumlib/Cirq/files/4996582/3qubit-decomposition.pdf)

Fixes #451.


Strilanc (dismissed):
Note: Must wait until scipy release (can't merge the wheels)

balopat:
Update from scipyDev mailing list: https://mail.python.org/pipermail/scipy-dev/2020-May/024154.html

Following schedule for 1.5.0 is accepted:
- May 26/2020: branch 1.5.x
- May 29/2020: rc1
- June 11/2020: rc2 (if needed)
- June 20/2020: final release

balopat:
@viathor - the last one in the Three Qubit Decomposition Series :D Thank you for all the reviews so far!

cirq/optimizers/three_qubit_decomposition.py
```diff
@@ -0,0 +1,272 @@
...
+
+    # if it's interesting to extract the diagonal then let's do it
+    if shift_left:
+        d_w, circuit_u1u2_l = opt.two_qubit_matrix_to_diagonal_and_operations(q1, q2, w, atol=atol)
+    # if we are at the end of the circuit, then just fall back to KAK
+    else:
+        d_w = np.eye(4)
+        circuit_u1u2_l = opt.two_qubit_matrix_to_operations(
+            q1, q2, w, allow_partial_czs=False, atol=atol
+        )
```
viathor, line 222: This is a little awkward. For example, when we pass `shift=False` we ignore `d_w` so setting it to `np.eye(4)` is unnecessary (but the function signature forces us to return something...). It seems to me we can make it more elegant by moving the `if shift_left` conditional outside of this function. This also reduces the the number of arguments since it allows us to get rid of `shift_left`.

cirq/optimizers/three_qubit_decomposition.py
```diff
@@ -0,0 +1,272 @@
...
+def _two_qubit_multiplexor_to_circuit(
+    q0: ops.Qid,
+    q1: ops.Qid,
+    q2: ops.Qid,
+    u1: np.ndarray,
+    u2: np.ndarray,
+    shift_left: bool = True,
+    diagonal: np.ndarray = np.eye(4),
+    atol: float = 1e-8,
+):
```
viathor, line 158: missing return type annotation

cirq/optimizers/three_qubit_decomposition.py
```diff
@@ -0,0 +1,272 @@
...
+    r"""Converts a two qubit double multiplexor to circuit.
+    Input: u1 ⊕ u2, with select qubit a (i.e. a = |0> => u1(b,c),
+    a = |1> => u2(b,c).
+
+    We want this:
+        $$
+        u1 ⊕ u2 = v ⊕ v @ D ⊕ D^{\dagger} @ W ⊕ W
```
viathor, line 165: capitalize u->U, v->V to match D and W

(but let's keep small v in code)

viathor, line 165: Strictly speaking there should be parentheses around direct sums since matrix product takes priority.

(Probably moot due to another comment.)

viathor, line 165: All factors here are direct sums (i.e. block diagonal matrices), so this is a less readable variant of

$$
U_1 = V D W \\
U_2 = V D^\dagger W
$$

viathor, line 165: I think the latex will render better if we have underscore between letter and digit: U_1 etc.

cirq/optimizers/three_qubit_decomposition_test.py
```diff
@@ -0,0 +1,182 @@
...
+        cirq.ControlledGate(cirq.ISWAP)._unitary_(),
+        cirq.CCX._unitary_(),
+    ],
+)
+def test_three_qubit_matrix_to_operations(u):
+    a, b, c = cirq.LineQubit.range(3)
+    operations = cirq.three_qubit_matrix_to_operations(a, b, c, u)
+    final_circuit = cirq.Circuit(operations)
+    final_unitary = final_circuit.unitary(qubits_that_should_be_present=[a, b, c])
+    cirq.testing.assert_allclose_up_to_global_phase(u, final_unitary, atol=1e-9)
```
viathor, line 46: We should also test that the number of CNOTs is at most 20.

cirq/optimizers/three_qubit_decomposition.py
```diff
@@ -0,0 +1,272 @@
...
+    try:
+        from scipy.linalg import cossin
+    except ImportError:  # coverage: ignore
+        # coverage: ignore
+        raise ImportError(
+            "cirq.three_qubit_unitary_to_operations requires "
+            "SciPy 1.5.0+, as it uses the cossin function. Please"
+            " upgrade scipy in your environment to use this "
+            "function!"
+        )
```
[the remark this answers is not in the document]
balopat: Because we do not want to impose for this single functionality a lower bound on the full project until we are convinced that 1.5.0 is propagated to the majority of large, slower moving environments (e.g. ours internally, or national labs). I can create an issue for this and make a TODO for a year from now or so. 1.5.0 is very new, it just came out in July.

cirq/optimizers/three_qubit_decomposition.py
```diff
@@ -0,0 +1,272 @@
...
+        cirq.CNOT(q2, q0),
+        rzs[2],
+        cirq.CNOT(q1, q0),
+        rzs[3],
+        cirq.CNOT(q2, q0),
+    ]
+    return _optimize_multiplexed_angles_circuit(ops)
+
+
+def _multiplexed_angles(theta: Union[List[float], np.ndarray]) -> np.ndarray:
```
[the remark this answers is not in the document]
balopat: I think I'll push back on this one. I tried it and I'm not convinced - as while it might save some space, I'm not sure it will be more readable / understandable / testable. I really like the fact that the individual building blocks map to the concepts in the paper. I find that easier to follow. Also, the unit tests are thoroughly testing each primitive one by one. If I start to merge them further up, I'd have to write much more complicated tests.

cirq/optimizers/three_qubit_decomposition.py
```diff
@@ -0,0 +1,272 @@
...
+        cirq.CZ(q2, q0),
+        rys[2],
+        cirq.CZ(q1, q0),
+        rys[3],
+        cirq.CZ(q2, q0),
+    ]
+    return _optimize_multiplexed_angles_circuit(ops)
+
+
+def _optimize_multiplexed_angles_circuit(operations: List[ops.Operation]):
```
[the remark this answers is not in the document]
balopat: Yeah a general compiler that does that could be useful, but much harder task than this specific one - that's why we agreed with Craig back in the day to implement the special case here. I can open an issue though, I think it might be useful. Though a lot of cases are covered by `MergeInteractions`, this might be an interesting one to write for larger gates. 

re: `cirq.linalg.is_involutive` could be cool - we just talked about with Orion maybe introducing a `cirq.period` protocol too, where things could tell their period, which would be 2 for involutary operators.

cirq/optimizers/three_qubit_decomposition.py
```diff
@@ -0,0 +1,272 @@
...
+    cs_ops = _cs_to_ops(q0, q1, q2, theta)
+    if len(cs_ops) > 0 and cs_ops[-1] == cirq.CZ(q2, q0):
+        # optimization A.1 - merging the last CZ from the end of CS into UD
+        # cz = cirq.Circuit([cs_ops[-1]]).unitary()
+        # CZ(c,a) = CZ(a,c) as CZ is symmetric
+        # for the u1⊕u2 multiplexor operator:
+        # as u1(b,c) is the operator in case a = \0>,
+        # and u2(b,c) is the operator for (b,c) in case a = |1>
+        # we can represent the merge by phasing u2 with I ⊗ Z
+        u2 = u2 @ np.kron(np.eye(2), np.array([[1, 0], [0, -1]]))
```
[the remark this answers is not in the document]
balopat: yes, with `[1, -1, 1, -1]`

cirq/optimizers/three_qubit_decomposition.py
balopat: Yes, that's right below it. I can remove it though if we don't think it's useful.

cirq/optimizers/three_qubit_decomposition.py
balopat: I know it's a bit awkward but theoretically correct and keeps the conceptual boundary nice and clean: this method returns the operations for a two-qubit multiplexor and extracts a diagonal optionally. That optionality is expressed via the diagonal being identity or not. 

If I pull out the if statement - then we have a method that "extracts some operations from the two-qubit multiplexor, and returns the W matrix for further processing"... It will mix the levels of abstraction in my opinion. 

I think I'll have to push back on this one as well, unless you see a resolution that is clean conceptually as well.

balopat:
@viathor - ready for a final review - I hope you are okay with my reasoning around the two outstanding items.

cirq/optimizers/three_qubit_decomposition.py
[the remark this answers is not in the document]
viathor: SGTM

cirq/optimizers/three_qubit_decomposition.py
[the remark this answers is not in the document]
viathor: Fair enough.

Regarding tests: Unit tests that verify behavior of private help functions make refactoring harder. Ideally, tests would verify behavior of such helper functions via specially chosen input data. This is why many best practices docs (e.g. Google's own testing best practices) encourage testing public APIs.

cirq/optimizers/three_qubit_decomposition.py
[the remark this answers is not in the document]
viathor: OK

cirq/optimizers/three_qubit_decomposition.py
viathor: Sounds good - I agree about keeping clean mapping between code and concepts in the paper.

It still bothers me that we instantiate the identity matrix that the caller just ignores. Instantiation of this matrix creates an expectation in the mind of someone reading the code that it'll be used for something.

Suggestion: Let's use `None` in place of `np.eye(4)`. This has a nice ripple effect of changing the signature of the function by introducing explicit `Optional` that makes the optional character of the array explicit.

cirq/optimizers/three_qubit_decomposition.py
```diff
@@ -0,0 +1,276 @@
...
+def _two_qubit_multiplexor_to_ops(
+    q0: ops.Qid,
+    q1: ops.Qid,
+    q2: ops.Qid,
+    u1: np.ndarray,
+    u2: np.ndarray,
+    shift_left: bool = True,
+    diagonal: np.ndarray = np.eye(4),
+    atol: float = 1e-8,
+) -> Tuple[np.ndarray, List[ops.Operation]]:
```
viathor, line 162: Let's make it `Tuple[Optional[np.ndarray], List[ops.Operation]]` per other comment.

cirq/optimizers/three_qubit_decomposition.py
```diff
@@ -0,0 +1,276 @@
...
+
+    Returns:
+        The resulting operations will have only known two-qubit and one-qubit
+        gates based operations, namely CZ, CNOT and rx, ry, PhasedXPow gates.
+    Raises:
+        ValueError if the u matrix is non-unitary or not of shape (8,8)
+    """
+    if np.shape(u) != (8, 8):
+        raise ValueError(f"Expected unitary matrix with shape (8,8) got {np.shape(u)}")
+    if not cirq.is_unitary(u):
```
viathor, line 49: We should probably use `atol` with `is_unitary`.

viathor (approved):
LGTM with a few minor comments.

cirq/optimizers/three_qubit_decomposition.py
balopat: I like that suggestion, thank you!

PR #2875, opened 2020-03-30
Add CZ to combined FSim gate set

dstrain115:


cirq/google/common_serializers_test.py
```diff
@@ -659,6 +659,8 @@ def test_wait_gate():
     (cirq.ISWAP**0.5, -np.pi / 4, 0),
     (cirq.ISWAP**-0.5, np.pi / 4, 0),
     (cirq.ISWAP**0.0, 0, 0),
+    (cirq.CZ, 0, np.pi),
+    (cirq.CZ**-1.0, 0, np.pi),
```
maffoo: Also add test cases for `cirq.FSimGate(theta=0, phi=np.pi)` etc.

PR #2876, opened 2020-03-30
Fix images in cirq docs

dstrain115:
- Move images to an images directory
- Also use relative links to force docs to add them.


dstrain115:
> Where does `_static`(https://github.com/quantumlib/Cirq/tree/master/docs/_static) fit into this?

Changed the images to _static and moved the links back to absolute.

PR #2877, opened 2020-03-30
Move notebooks from examples to docs

dstrain115:
- Move notebooks from examples/notebooks to docs/studies
- This is a new and growing section for more advanced case studies that will appear in the docs.


dstrain115:
> Do we want all the images in one place? `_static`?

Moved to _static

PR #2878, opened 2020-04-01
Fix time bugs in EngineProcessor.get_schedule

Strilanc:
- Fix default "now" being assigned at init time instead of at call time
- Refactored method to take a datetime (absolute) or a timedelta (relative to now)
- Fixed specifying from_time but not to_time (or vice versa) resulting in a bad filter containing None
- Add 'freezegun' requirement for testing with control over 'now'


cirq/google/engine/engine_processor.py
```diff
@@ -277,29 +277,65 @@ def list_reservations(
...
 
         The schedule may be filtered by time.
 
         Time slot type will be supported in the future.
+
+        Args:
+            from_time: Filters the returned scheduled to only include entries that end no earlier than the given value.
+                Specified either as an absolute time (datetime.datetime) or as a time relative to now
+                (datetime.timedelta). Defaults to now (a relative time of 0). Set to None to omit this filter.
+            to_time: Filters the returned scheduled to only include entries that start no later than the given value.
```
viathor, line 296: nit: s/scheduled/schedule/

Also applies to l.293 above.

cirq/google/engine/engine_processor.py
```diff
@@ -277,29 +277,65 @@ def list_reservations(
...
+                Specified either as an absolute time (datetime.datetime) or as a time relative to now
+                (datetime.timedelta). Defaults to two weeks from now (a relative time of two weeks).
+                Set to None to omit this filter.
+            time_slot_type: Filters the returned schedule to only include entries with a given type (e.g. maintenance,
+                open swim). Defaults to None. Set to None to omit this filter.
+
+        Returns:
+            Schedule time slots.
         """
+        now = datetime.datetime.utcnow()
```
maffoo, line 306: I think you should use `datetime.datetime.now()` instead. Using `utcnow` gives a "bare" datetime so when you get the timestamp below it will not correspond to the current time.
```python
>>> import datetime, time
>>> print(time.time(), datetime.datetime.now().timestamp(), datetime.datetime.utcnow().timestamp())  
1585755370.6489632 1585755370.648967 1585780570.648977
```

Strilanc: Fixed. Also caught a bug in freezegun when I tried to make a test that caught this: https://github.com/spulec/freezegun/issues/344

cirq/google/engine/engine_processor.py
Strilanc: Done

cirq/google/engine/engine_processor.py
viathor: Good catch! 

Apparently `datetime.utcnow` returns a `datetime` instance with time set to the current UTC time and the timezone set to `None` (known as "naive" `datatime` instance). This is interpreted by `timestamp` as the *local* timezone. Consequently, the local-UTC offset is applied twice: first by `utcnow` and a second time by `timestamp`.

There is a warning about this behavior in the documentation [1].

---
[1] https://docs.python.org/3/library/datetime.html

wcourtney: We moved intentionally to a bare timestamp with utcnow(). When datetime.datetime.now() is used, the datetime is "localized" to your environment, but that's a problem when your environment is Colab and you can't set the environment timezone...the defaults will all be skewed by your UTC offset.

maffoo: @wcourtney, by "timestamp" we usually mean a single number (seconds since unix epoch). As I showed above, if you do `utcnow().timestamp()` to get a timestamp you get the wrong value. If you want to get a timestamp you have to use `now().timestamp()` or use datetime instances with an actual timezone set to UTC.

cirq/google/engine/engine_processor.py
```diff
@@ -277,29 +277,70 @@ def list_reservations(
 
     def get_schedule(
             self,
-            from_time: datetime.datetime = datetime.datetime.now(tz=utc),
-            to_time: datetime.datetime = datetime.datetime.now(tz=utc) +
-            datetime.timedelta(weeks=2),
-            time_slot_type: qenums.QuantumTimeSlot.TimeSlotType = None
+            from_time: Union[None, datetime.datetime, datetime.
+                             timedelta] = datetime.timedelta(),
+            to_time: Union[None, datetime.datetime, datetime.
```
wcourtney, line 282: This parameter style should be kept in sync with list_reservations() just above.

cirq/google/engine/engine_processor.py
```diff
@@ -277,29 +277,70 @@ def list_reservations(
...
+                (datetime.timedelta). Defaults to two weeks from now (a relative
+                time of two weeks). Set to None to omit this filter.
+            time_slot_type: Filters the returned schedule to only include
+                entries with a given type (e.g. maintenance, open swim).
+                Defaults to None. Set to None to omit this filter.
+
+        Returns:
+            Schedule time slots.
         """
+        now = datetime.datetime.now()
```
wcourtney, line 310: Please ensure that this works as expected in a colab environment. I expect that the timestamp is going to be skewed by the utc offset.

Strilanc: Why would you expect dates to behave differently in colab? That would be a pretty serious problem with colab.

Strilanc: In a new colab notebook the timezone seems to be set to UTC so you can't tell the difference between the two methods.

cirq/google/engine/engine_processor.py
Strilanc: Done.

Issue #2726, opened 2020-01-31
Exponents in gates should ensure that they are real


zchen088:
In many cases the exponents are periodic so we need to use `mod` to translate the gate into a hardware implementation, but this only works if the exponent is `float` rather than `complex`.

PR #2881, opened 2020-04-01
Ensure exponent in eigen gate is real.

tanujkhattar:
To solve #2726


cirq/ops/eigen_gate_test.py
```diff
@@ -88,6 +88,8 @@ def test_init():
     assert CExpZinGate(sympy.Symbol('a')).exponent == sympy.Symbol('a')
 
     assert ZGateDef(exponent=0.5).exponent == 0.5
+    with pytest.raises(TypeError):
```
Strilanc, line 91: Include a `match="real"` argument to raises.

cirq/ops/eigen_gate_test.py
```diff
@@ -203,6 +205,8 @@ def test_pow():
                     global_shift=0.5)**2 == ZGateDef(
         exponent=0.5,
         global_shift=0.5)
+    with pytest.raises(TypeError):
```
Strilanc, line 208: Include a `match="real"` argument to raises.

cirq/ops/eigen_gate.py
```diff
@@ -102,6 +102,8 @@ def __init__(
                 `cirq.rx(t)` uses a `global_shift` of -0.5, which is why
                 `cirq.unitary(cirq.rx(pi))` equals -iX instead of X.
         """
+        if not isinstance(exponent, (int, float, sympy.Basic)):
```
Strilanc, line 105: I think we should allow `complex` as long as the imaginary part is 0. Instead of doing this check, do:

```
if isinstance(exponent, complex) and exponent.imag:
    raise ValueError(...)
```

For example, this will mean that things like `fractions.Fraction` will work. It's more in line with python's duck typing philosophy.

PR #2883, opened 2020-04-02
Add JSON Serialization for GateTabulation.

KevinVillela:
This is useful for precomputing the GateTabulation for Quantum Volume metrics.


KevinVillela:
@mpharrigan would you be able to take a look at this guy, please?

KevinVillela:
Sounds good, thanks! AFAICT the test failure is unrelated to this change...

Out of curiosity, outside of Cirq what should `repr` be used for? Debugging info?

mpharrigan:
As far as I understand it `repr` should give you some idea of what type of object you're dealing with while `str` tells you more about the data within

PR #2885, opened 2020-04-03
Check 2x2 unitary in has_stabilizer_effect

fedimser:
If everything else fails (i.e. op and its gate don't declare has_stabilizer effect), and op has unitary, 2hich is 2x2 unitary, explicilty check if that unitary has stabilizer effect.
After this change `has_stabilizer_effect` returns `True` exactly for those gates which are supported by `CliffordSimulator`.
This is follow-up to PR #2803.

### DIFF
--- a/cirq/protocols/has_stabilizer_effect_protocol.py
+++ b/cirq/protocols/has_stabilizer_effect_protocol.py
@@ -76,5 +76,5 @@ def _strat_has_stabilizer_effect_from_unitary(val: Any) -> Optional[bool]:
         return None
     unitary = protocols.unitary(val)
     if unitary.shape == (2, 2):
-        return not SingleQubitCliffordGate.from_unitary(unitary) is None
+        return SingleQubitCliffordGate.from_unitary(unitary) is not None
     return None
--- a/cirq/protocols/has_stabilizer_effect_protocol.py
+++ b/cirq/protocols/has_stabilizer_effect_protocol.py
@@ -25,6 +25,9 @@ from cirq import protocols
 def has_stabilizer_effect(val: Any) -> bool:
     """
     Returns whether the input has a stabilizer effect.
+
+    For 1-qubit gates always returns correct result. For other operations relies
+    on the operation to define whether it has stabilizer effect.
     """
     strats = [
         _strat_has_stabilizer_effect_from_has_stabilizer_effect,
@@ -66,7 +69,8 @@ def _strat_has_stabilizer_effect_from_gate(val: Any) -> Optional[bool]:
 def _strat_has_stabilizer_effect_from_unitary(val: Any) -> Optional[bool]:
     """Attempts to infer whether val has stabilizer effect from its unitary.
 
-    Returns whether unitary of `val` normalizes Pauli group.
+    Returns whether unitary of `val` normalizes the Pauli group. Works only for
+    2x2 unitaries.
     """
     if not protocols.has_unitary(val):
         return None
--- a/cirq/protocols/has_stabilizer_effect_protocol.py
+++ b/cirq/protocols/has_stabilizer_effect_protocol.py
@@ -17,17 +17,19 @@ from typing import (
     Optional,
 )
 
+from cirq.ops.clifford_gate import SingleQubitCliffordGate
+
+from cirq import protocols
+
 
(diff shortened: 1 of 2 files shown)


cirq/protocols/has_stabilizer_effect_protocol.py
```diff
@@ -17,17 +17,19 @@
...
+from cirq import protocols
+
 
 def has_stabilizer_effect(val: Any) -> bool:
     """
-    Returns whether the input has a stabilizer effect. Currently only limits to
-    Pauli, H, S, CNOT and CZ gates and their Operations. Does not attempt to
-    decompose a gate into supported gates. For e.g. iSWAP or X**0.5 gate will
-    return False.
+    Returns whether the input has a stabilizer effect.
```
smitsanghavi, line 27: I'd keep the caveats still (but update them) because even after this, it doesn't generalize to include all stabilizer-effect inputs.

cirq/protocols/has_stabilizer_effect_protocol.py
```diff
@@ -59,3 +61,16 @@ def _strat_has_stabilizer_effect_from_gate(val: Any) -> Optional[bool]:
     if hasattr(val, 'gate'):
         return _strat_has_stabilizer_effect_from_has_stabilizer_effect(val.gate)
     return None
+
+
+def _strat_has_stabilizer_effect_from_unitary(val: Any) -> Optional[bool]:
+    """Attempts to infer whether val has stabilizer effect from its unitary.
+
```
smitsanghavi, line 68: Also add that this is constrained to 2x2 unitaries.

cirq/protocols/has_stabilizer_effect_protocol_test.py
```diff
@@ -111,3 +122,17 @@ def test_via_gate_of_op():
     assert not cirq.has_stabilizer_effect(NoOp1())
     assert not cirq.has_stabilizer_effect(NoOp2())
     assert not cirq.has_stabilizer_effect(NoOp3())
+
+
+def test_via_unitary():
+    op1 = OpWithUnitary(np.array([[0, 1], [1, 0]]))
```
smitsanghavi, line 128: Also add a case with an additional overall phase. iX or -X

cirq/protocols/has_stabilizer_effect_protocol.py
fedimser: Rephrazed. I don't think we need to list all gates for which it returns True, because then we will have to update this pydoc every time _has_stabilizer_effect_ is defined for a new gate.

cirq/protocols/has_stabilizer_effect_protocol.py
fedimser: Done.

cirq/protocols/has_stabilizer_effect_protocol_test.py
fedimser: Done.

fedimser:
Can we please merge this? I think we agreed that it's ok to fall back to matrix representation in 2x2 (single-qubit) case.
@Strilanc

cirq/protocols/has_stabilizer_effect_protocol.py
```diff
@@ -59,3 +64,17 @@ def _strat_has_stabilizer_effect_from_gate(val: Any) -> Optional[bool]:
...
+    Returns whether unitary of `val` normalizes the Pauli group. Works only for
+    2x2 unitaries.
+    """
+    if not protocols.has_unitary(val):
+        return None
+    unitary = protocols.unitary(val)
+    if unitary.shape == (2, 2):
+        return not SingleQubitCliffordGate.from_unitary(unitary) is None
```
viathor, line 79: nit: use `is not None` for better readability:

```python
return SingleQubitCliffordGate.from_unitary(unitary) is not None
```

fedimser: Done.

fedimser:
@googlebot I fixed it.

fedimser:
Closing this PR because I accidentally pushed commits with wrong email and I can't fix CLA. I opened PR #2938 which is identical to this.

PR #2886, opened 2020-04-03
Fix column axis rename in pandas 1.0

dstrain115:
Fix axis rename and unpin pandas

### DIFF
--- a/cirq/experiments/t1_decay_experiment.py
+++ b/cirq/experiments/t1_decay_experiment.py
@@ -75,7 +75,7 @@ def t1_decay(sampler: work.Sampler,
 
     # Cross tabulate into a delay_ns, false_count, true_count table.
     tab = pd.crosstab(results.delay_ns, results.output)
-    tab=tab.rename_axis(None, axis="columns")
+    tab = tab.rename_axis(None, axis="columns")
     tab = tab.rename(columns={0: 'false_count', 1: 'true_count'}).reset_index()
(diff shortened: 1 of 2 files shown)


cirq/experiments/t1_decay_experiment.py
```diff
@@ -75,7 +75,7 @@ def t1_decay(sampler: work.Sampler,
 
     # Cross tabulate into a delay_ns, false_count, true_count table.
     tab = pd.crosstab(results.delay_ns, results.output)
-    del tab.columns.name
+    tab = tab.rename_axis(None, axis="columns")
```
viathor, line 78: Suggestion:

```python
tab.rename_axis(None, axis="columns", inplace=True)
```

On my workstation in place rename is 6x faster. 

```
In [26]: %timeit df2.rename_axis(None, axis="columns")
83.7 µs ± 4.83 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [27]: %timeit df2.rename_axis(None, axis="columns", inplace=True)
13.9 µs ± 31.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

dstrain115: Good idea.  Done.

viathor: You should not be storing the return value back in tab.

dstrain115: Oops.  I also shouldn't be making quick fixes on Friday evening!  
Fixed.

PR #2889, opened 2020-04-03
Add return values to direct_fidelity_estimation()

tonybruguier-google:
The intent is to eventually run experiments on a real computer, and thus, it might make sense to record intermediate data for later analysis


tonybruguier-google:
Hi,

I would like to eventually run experiments, and so I was advised by mpharrigan@ to record everything since compute time is expensive. I have a quick example of how I would use it here:
https://github.com/tonybruguier-google/Cirq/blob/d3ea3159880359b179b18688f0708022ce948b8a/examples/direct_fidelity_estimation_recirq.py#L52

examples/direct_fidelity_estimation.py
```diff
@@ -218,10 +220,12 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
             'sampler' parameter directly to estimate the characteristic
             function.
     Returns:
-        The estimated fidelity.
+        The estimated fidelity and a log of the run.
     """
     # n_trials is upper-case N in https://arxiv.org/abs/1104.3835
 
+    run_log: Dict[str, Any] = {}
```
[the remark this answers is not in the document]
viathor: Strictly speaking final result of DFE is the estimate of fidelity. The quantities placed in `run_log` are intermediate results. If this is needed for debugging, then perhaps using python's `logging` library is more appropriate?

---

If however, you do require the intermediate results for some sort of analysis, then I agree with @mpharrigan's suggestion to introduce a dataclass. I'd name it `DFEIntermediateResults` though and would *not* make `estimated_fidelity` a member of it. This way it's clear what the result of DFE is. A nice practical consequence of this API is that in the most common use-case of DFE you can discard the intermediate results like this:

```python
fidelity_estimate, _ = cirq. direct_fidelity_estimation(...)
```

as opposed to having to extract the fidelity estimate from the returned structure. 

Also, rename `run_log` to `intermediate_results` or `intermediates`.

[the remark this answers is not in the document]
tonybruguier-google: * Added a dataclass, as per @mpharrigan 's suggestion, but named it `DFEIntermediateResult` as per @viathor 's suggestion
* As per @viathor 's suggestion, it does _not_ contain the fidelity. Note that I was originally returning the value in two locations, just as a plain float and in the result class.
* Renamed the variable `dfe_intermediate_result`

tonybruguier-google:
Thanks. PTAL.

examples/direct_fidelity_estimation.py
```diff
@@ -196,6 +199,23 @@ def _estimate_pauli_traces_general(qubits: List[cirq.Qid],
...
 
+@dataclass
+class DFEIntermediateResult:
+    """
+    A container for the various debug and run data from calling the function
+    direct_fidelity_estimation(). This is useful when running a long-computation
+    on an actual computer, which is expensive. This way, runs can be more easily
+    debugged offline.
+    """
+    # If the circuit is Clifford, the list of Clifford basis states. Otherwise
```
viathor, line 210: nit: comment says this is a list of Clifford states, but type annotation says it is a single optional state.

examples/direct_fidelity_estimation.py
```diff
@@ -196,6 +199,23 @@ def _estimate_pauli_traces_general(qubits: List[cirq.Qid],
...
+    # it is None.
+    clifford_state: Optional[cirq.CliffordState]
+    # The list of Pauli traces we can sample from.
+    pauli_traces: List[Dict[str, Any]]
```
viathor, line 214: Using the `Any` type essentially disables type checking, so we shouldn't use it unless there is really no other way. Could you replace `Any` with more concrete types here and in `trial_results` below?

(This does have a lot of knock-on effects, but type annotations are worth the effort of putting them in.)

examples/direct_fidelity_estimation.py
tonybruguier-google: Thanks. Modified the comment to say it's a single object but we can get a list of Pauli strings out of it. Is this OK?

examples/direct_fidelity_estimation.py
tonybruguier-google: OK. So let me sure I understand what it is that you are asking here.

I think the main issue is that the values of the dictionary have heterogeneous types, and thus having an annotation is difficult. I seems that it's possible with TypedDict:
https://stackoverflow.com/questions/48054521/indicating-multiple-value-in-a-dict-for-type-hints

But this seems to be available in Python 3.8:
https://www.python.org/dev/peps/pep-0589/

However, it seems that we are a version < 3.7
https://github.com/quantumlib/Cirq/blob/master/requirements.txt

Am I correct? I'm leaving this unresolved because I am happy to address the comment, it's just that I could use a little help when you have the time.
Am I correct?

tonybruguier-google:
Thanks. PTAL.

examples/direct_fidelity_estimation.py
tonybruguier-google: Actually, I though about it overnight and you had already given the answer: Use dataclasses, which I did. Sorry for being obtuse.

No more "Any" type.

tonybruguier-google:
PTAL.

examples/direct_fidelity_estimation.py
```diff
@@ -95,6 +96,20 @@ def compute_characteristic_function(circuit: cirq.Circuit,
...
 
+@dataclass
+class PauliTrace:
+    """
+    A class that contains the Pauli states as described on page 2 of:
+    https://arxiv.org/abs/1104.3835
+    """
+    # Pauli string.
+    P_i: cirq.PauliString
+    # Defined at bottom of left column of page 2.
```
viathor, line 107: Definition of rho_i is simple enough to include inline in addition to the reference.

(Applies to Pr_i and sigma_i below, too.)

examples/direct_fidelity_estimation.py
```diff
@@ -185,17 +197,47 @@ def _estimate_pauli_traces_general(qubits: List[cirq.Qid],
...
+@dataclass
+class DFEIntermediateResult:
+    """
+    A container for the various debug and run data from calling the function
+    direct_fidelity_estimation(). This is useful when running a long-computation
+    on an actual computer, which is expensive. This way, runs can be more easily
+    debugged offline.
+    """
+    # If the circuit is Clifford, the Clifford state from which we can extract
+    # a list of Pauli strings that for a basis of the stabilizers.
```
viathor, line 233: nit: remove "that"

examples/direct_fidelity_estimation_test.py
```diff
@@ -51,13 +49,12 @@ def test_direct_fidelity_estimation_with_noise():
...
     noisy_simulator = cirq.DensityMatrixSimulator(noise=noise)
 
-    estimated_fidelity = direct_fidelity_estimation.direct_fidelity_estimation(
-        circuit,
-        qubits,
-        noisy_simulator,
-        n_trials=10,
-        n_clifford_trials=3,
-        samples_per_term=10)
+    estimated_fidelity, _ = dfe.direct_fidelity_estimation(circuit,
```
viathor, line 52: At least one of the tests should inspect and verify the second return value (the intermediate results).

examples/direct_fidelity_estimation.py
tonybruguier-google: Done

examples/direct_fidelity_estimation_test.py
tonybruguier-google: Done. I didn't test everything because some of the results are random, but I think the coverage maybe is sufficient since I hit every object?

examples/direct_fidelity_estimation.py
tonybruguier-google: Done

tonybruguier-google:
Thanks. PTAL.

examples/direct_fidelity_estimation.py
```diff
@@ -95,6 +96,22 @@ def compute_characteristic_function(circuit: cirq.Circuit,
...
 
+@dataclass
+class PauliTrace:
+    """
+    A class that contains the Pauli states as described on page 2 of:
+    https://arxiv.org/abs/1104.3835
+    """
+    # Pauli string.
+    P_i: cirq.PauliString
+    # The pure state for the Pauli string above, formally defined at bottom of
```
viathor, line 107: Wait, this isn't true though, is it? Rho with the caret is the ideal pure state. Rho with the index i is the trace of the product of rho caret with ith Pauli string. You can also see this from the type annotation: rho_i is a float, so it can't be a pure state.

I see three ways of defining this well:

1. Experimentalist: "Average value of the ith Pauli observable in the ideal pure state."
2. Theoretician: "Coefficient of the ideal pure state expanded in the Pauli basis scaled by sqrt(dim H)".
3. Mathematician: "Hilbert-Schmidt inner product of the ideal pure state and the ith Pauli basis element."
4. Just show me the math: Simply state the formula.

My favorite is actually 2 since it hints a little bit at how DFE works behind the scenes. Your choice though.

examples/direct_fidelity_estimation.py
```diff
@@ -185,17 +199,48 @@ def _estimate_pauli_traces_general(qubits: List[cirq.Qid],
...
 
 
+@dataclass
+class TrialResult:
+    """
+    Contains the results of a trial, either by simulator or actual run
+    """
+    # The index in the list of Pauli traces.
+    i: int
+    # The measured state for that trial under the Pauli state indexed by i
```
viathor, line 221: This is not a measured state. See comment above. Whatever you decide there, please make this comment consistent with the one above.

examples/direct_fidelity_estimation.py
tonybruguier-google: Done

examples/direct_fidelity_estimation.py
tonybruguier-google: I changed it, but I wasn't as sure. Does this look reasonable?

tonybruguier-google:
Thanks. PTAL.

viathor:
Thank you for all the additional cleanup you have done in this PR! It is really nice.

tonybruguier-google:
Thanks for your patience with me and making me a better coder.

PR #2892, opened 2020-04-06
Adds cirq.num_cnots_required and cirq.to_special

balopat:
Adds cirq.num_two_qubit_gates_required and cirq.to_special. 

- `cirq.num_cnots_required`: Based on simple linear algebra users can calculate the minimum required two-qubit gates (CZ, CNOT) to implement a two-qubit unitary. 
- `cirq.to_special`: converts a unitary to a special unitary

Context: this is the first PR breaking up #2873.

For an overview of the high level design decisions of the whole project see: https://drive.google.com/open?id=1SDEtttIAaTwfV9AUs7XAxeW3AUd0qLoY.


balopat:
Argh, I pushed this from my personal computer, I'll fix the CLA soon.

balopat:
note - I'm just testing automerge

cirq/contrib/shannon_decomposition/three_qubit_decomposition.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Args:
+        u: a two-qubit unitary
+    Returns:
+        the number of two-qubit gates required to implement the unitary
+    """
+    assert np.shape(u) == (4, 4)
+    assert is_unitary(u)
+
+    poly = np.poly(_gamma(_to_special(u)))
```
viathor, line 71: By the last part of proposition III.3 in ["Recognizing..."](https://arxiv.org/pdf/quant-ph/0308045.pdf), if your goal is to check whether u requires 3 CNOTs then you don't need to compute the full characteristic polynomial of gamma. The reason is that coefficients 0, 2 and 4 are guaranteed to be real and coefficients 3 and 1 are the trace and its conjugate. Thus, we only need to check that the trace is real which is simpler.

```python
g = _gamma(_to_special(u))
return not np.allclose(np.trace(g).imag, 0, atol=tolerance)
```

cirq/contrib/shannon_decomposition/three_qubit_decomposition.py
```diff
@@ -0,0 +1,72 @@
...
+    Args:
+        u: a member of SU(4)
+    Returns:
+        u @ yy @ u.T @ yy, where yy = Y ⊗ Y
+    """
+
+    return u @ _YY @ u.T @ _YY
+
+
+def _is_three_cnot_two_qubit_unitary(u: np.ndarray) -> bool:
```
viathor, line 56: Since the function involves a check that a potentially complex quantity is real, I'd add a tolerance (with a default).

cirq/contrib/shannon_decomposition/three_qubit_decomposition.py
```diff
@@ -0,0 +1,72 @@
...
+    """Returns true if U requires 3 CNOT/CZ gates.
+
+    See Proposition III.1, III.2, III.3 in Shende et al. “Recognizing Small-
+    Circuit Structure in Two-Qubit Operators and Timing Hamiltonians to Compute
+    Controlled-Not Gates”.  https://arxiv.org/abs/quant-ph/0308045
+
+    Args:
+        u: a two-qubit unitary
+    Returns:
+        the number of two-qubit gates required to implement the unitary
```
viathor, line 66: Probably should be something like "True if u requires 3 CNOTs to implement, False if it requires 0, 1 or 2." (but I think it may be of more general interest to have a function consistent with your present Returns description, see another comment.)

cirq/contrib/shannon_decomposition/three_qubit_decomposition.py
viathor, line 56: Optional: You can use the propositions in the paper to implement a function that takes a 2-qubit unitary and returns the number of CNOTs required. I think this results in better API and could be of more general interest. I'd even make it public.

cirq/contrib/shannon_decomposition/three_qubit_decomposition_test.py
```diff
@@ -0,0 +1,58 @@
...
+    _to_special, _is_three_cnot_two_qubit_unitary)
+
+
+def test_is_three_cnot_two_qubit_unitary():
+    assert _is_three_cnot_two_qubit_unitary(
+        _two_qubit_circuit_with_cnots(3)._unitary_())
+    assert not _is_three_cnot_two_qubit_unitary(
+        _two_qubit_circuit_with_cnots(2)._unitary_())
+    assert not _is_three_cnot_two_qubit_unitary(
+        _two_qubit_circuit_with_cnots(1)._unitary_())
```
viathor, line 30: Maybe add the case with 0 CNOTs.

viathor (changes requested):
Why is this in contrib and not in `cirq.optimizers`? The original bug suggests this is going to be very useful.

cirq/contrib/shannon_decomposition/three_qubit_decomposition.py
viathor: (This comment becomes moot if you choose to make this a function that computes the number of CNOTs required.)

viathor: Actually, even in this more general case we can avoid computing the characteristic polynomial, because the symmetry of the polynomial used in Proposition III.3 remains valid and so we only need to know the coefficients by x^3 and x^2. Specifically, I think we can do something like this:

```python
def number_of_cnots_required(u: np.ndarray, tolerance: float = 1e-6) -> int:
    g = _gamma(_to_special(u))
    a3 = np.trace(g)
    if np.abs(a3 - 4) < tolerance or np.abs(a3 + 4) < tolerance:
        return 0
    a2 = (a3*a3 - np.trace(g @ g)) / 2
    if np.abs(a3) < tolerance and np.abs(a2 - 2) < tolerance:
        return 1
    if np.abs(a3.imag) < tolerance:
        return 2
    return 3
```

The first condition comes from Proposition III.1 and Pascal triangle. Note that we don't need to check whether a2 = 6 (why?). The second condition comes from Proposition III.2 and simple algebra. The final condition is the existing code.

The formula for a2 is the second iteration of [Fadeev-LaVerrier](https://en.wikipedia.org/wiki/Faddeev%E2%80%93LeVerrier_algorithm).

cirq/contrib/shannon_decomposition/three_qubit_decomposition.py
```diff
@@ -0,0 +1,72 @@
...
+import numpy as np
+
+from cirq.linalg import is_unitary
+from cirq.ops import YY
+from cirq.protocols import unitary
+
+_YY = unitary(YY)
+
+
+def _to_special(u: np.ndarray) -> np.ndarray:
```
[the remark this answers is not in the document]
balopat: Sounds good, I'll move it.

cirq/contrib/shannon_decomposition/three_qubit_decomposition_test.py
balopat: np.eye(4) was "a" 0 case, but I can add it in as well.

cirq/contrib/shannon_decomposition/three_qubit_decomposition.py
balopat: Good catch, yes, the original version was calculating the number of CNOTs required. It seems like you think there might be general interest for it. I'll make it a public utility function.

cirq/contrib/shannon_decomposition/three_qubit_decomposition.py
balopat: This is great - it is probably more efficient than my naive https://github.com/quantumlib/Cirq/pull/2873/files#diff-4a594a63b684a45eefea705168a2777bR193 following the polynomial calculation - assuming that np.poly is more expensive than calculating g @ g. 

Thank you for pointing me to the recursive Fadeev-LaVerrier formula. 

Not exactly sure why we don't need to check whether a2 = 6. We know from `g` being from SU(4) and that the product of the eigenvalues should be 1 that  a4=a0=1 and a1 = a3 = tr(g) as described in the proof of Proposition III.3. Fadeev-LaVerrier requires g^2 to be calculated, so it can't be a simple function of a3...but it sounds like you have a reason to believe that from a3 = ±4 it follows that a2 = 6. I would need to find a special unitary with char.polynomial [1,4,a3!=6,4,1] to prove you wrong, but I'm going to ask for your proof instead :)

This looks good - not that it matters for the calculations, but to stay closer to the Fadeev-LaVerrier formulae, `a3 = - np.trace(g)`. Also `np.abs(a3)-4 < tolerance` is a further simplification of the first condition.

viathor: (I misspelled the name in my comment: should be "Le" in LeVerrier. Sorry.)

Proof that a3 = ±4 implies a2 = 6: Consider the four eigenvalues of g. They are all numbers on the unit circle in the complex plane. The only way(*) for a3, i.e. the negative sum of the eigenvalues to be +4 (respectively, -4) is if they are all +1 (respectively, -1). So the condition a3 = ±4 fixes all eigenvalues and hence it also fixes the characteristic polynomial. The matrix with eigenvalues +1, +1, +1, +1 (respectively, -1, -1, -1, -1) has characteristic polynomial (x-1)^4 (respectively, (x+1)^4) which is x^4 + 4x^3 + 6x^2 + 4x + 1 (respectively, x^4 - 4x^3 + 6x^2 - 4x + 1). Thus, a2 = 6. Done.

Regarding the efficiency of np.poly. My guess is that it uses Fadeev-LeVerrier, so it's probably quite fast, but will still do more work than we need because it is unaware of highly constrained nature of our matrix. In particular, for the last coefficient it will end up computing the determinant as it doesn't know that our matrix is in SU(4)!

I agree `a3 = -np.trace(g)` is better. Thanks for catching this!

I don't think we can simplify `np.abs(a3 - 4) < tolerance or np.abs(a3 + 4) < tolerance` to `np.abs(a3)-4 < tolerance`, because we don't know that a3 is real! Counterexample is `np.diag(1j, 1j, 1j, 1j)`.

---
(*) Formally: let the real parts of the eigenvalues be 1-a, 1-b, 1-c, 1-b. Note that a, b, c, d are real numbers in [0, 2] (or else the eigenvalues are not on the unit circle). The sum of the eigenvalues is 4-a-b-c-d which equals 4 only when a=b=c=d=0.

balopat: oh such a simple proof, nice :) I should have just thought a bit harder, thank you! 
Sounds good, I'll do the modifications!

balopat:
> Why is this in contrib and not in `cirq.optimizers`? The original bug suggests this is going to be very useful.

@viathor, thank you for the thorough review - awesome suggestions!
I moved most of these functions into cirq packages, also updated the PR description to cater for them and fixed most of your comments. Let me know what you think!

cirq/linalg/decompositions.py
```diff
@@ -992,3 +1000,49 @@ def _canonicalize_kak_vector(k_vec: np.ndarray, atol: float) -> np.ndarray:
     k_vec[need_diff, 2] *= -1
 
     return k_vec
+
+
+def num_two_qubit_gates_required(u: np.ndarray, atol: float = 1e-8) -> int:
```
viathor, line 1005: I'd keep CNOT or CZ in the name since if one were to compute the minimum number of gates of a different type required to implement a given unitary one would generally obtain a different result(*).

Also, we may in future consider decompositions into other gates, so a name containing CNOT or CZ is future-proof.

---
(*) For example, SWAP requires 3 CNOTs/CZs, but (trivially) only 1 SWAP.

cirq/linalg/decompositions.py
```diff
@@ -992,3 +1000,49 @@ def _canonicalize_kak_vector(k_vec: np.ndarray, atol: float) -> np.ndarray:
...
+    """Returns the min number of 2-qubit gates required by a two-qubit unitary.
+
+    See Proposition III.1, III.2, III.3 in Shende et al. “Recognizing Small-
+    Circuit Structure in Two-Qubit Operators and Timing Hamiltonians to Compute
+    Controlled-Not Gates”.  https://arxiv.org/abs/quant-ph/0308045
+    Args:
```
viathor, line 1011: nit: need empty line above Args

cirq/linalg/decompositions.py
```diff
@@ -992,3 +1000,49 @@ def _canonicalize_kak_vector(k_vec: np.ndarray, atol: float) -> np.ndarray:
...
+    if np.abs(a3.imag) < atol:
+        return 2
+    return 3
+
+
+def _gamma(u: np.ndarray) -> np.ndarray:
+    """Gamma function to convert u to the magic basis.
+
+    See Definition IV.1 in Shende et al. "Minimal Universal Two-Qubit CNOT-based
+     Circuits." https://arxiv.org/abs/quant-ph/0308033
```
viathor, line 1040: nit: remove one of the leading spaces to align

cirq/linalg/decompositions.py
```diff
@@ -992,3 +1000,49 @@ def _canonicalize_kak_vector(k_vec: np.ndarray, atol: float) -> np.ndarray:
...
+
+    Args:
+        u: a member of SU(4)
+    Returns:
+        u @ yy @ u.T @ yy, where yy = Y ⊗ Y
+    """
+
```
viathor, line 1047: nit: remove empty line for consistency with other functions

cirq/linalg/decompositions_test.py
```diff
@@ -741,3 +741,38 @@ def test_kak_decompose(unitary: np.ndarray):
...
+
+    def random_one_qubit_gate():
+        return cirq.PhasedXPowGate(phase_exponent=random.random(),
+                                   exponent=random.random())
+
+    def one_cz():
+        return [
+            random_one_qubit_gate().on(a),
+            random_one_qubit_gate().on(b),
+            cirq.CZ.on(a, b)
```
viathor, line 772: The list should begin with CZ (or otherwise the random single-qubit gates on lines 776 and 777 below should be moved to after the list comprehension).

viathor (approved):
Nice! LGTM with a few minor comments.

cirq/linalg/decompositions.py
viathor: (SWAP is not universal, so it's a poor example, but the larger point remains.)

cirq/linalg/decompositions_test.py
balopat: Ah, nice catch!

PR #2129, opened 2019-09-19
Add initial Clifford simulator

vtomole:
Author of https://github.com/quantumlib/Cirq/pull/1927 has not been active so I will address the comments in that pull request so we can get this in.

Fixes: https://github.com/quantumlib/Cirq/issues/965

PR #2894, opened 2020-04-07
Add JSON serialization for CliffordState and CliffordTableau

tonybruguier-google:
I am unfamiliar with the section of the code base, so I'd be happy to withdraw the PR if this is misguided and/or is clearly not ready for review (pointers would be welcome in this case),


cirq/sim/clifford/clifford_simulator.py
```diff
@@ -279,6 +280,25 @@ def __init__(self, qubit_map, initial_state=0):
         self.ch_form = stabilizer_state_ch_form.StabilizerStateChForm(
             self.n, initial_state)
 
+    def _json_dict_(self):
+        return {
+            'cirq_type': self.__class__.__name__,
+            'qubit_map': dict([(v, k) for k, v in self.qubit_map.items()]),
```
mpharrigan, line 286: use a dictionary literal

```python
{v: k for k, v in ...}
```

Why are you reversing the keys and values? this might be dangerous? instead store as a list of tuples

cirq/sim/clifford/clifford_tableau.py
```diff
@@ -50,6 +51,26 @@ def bits(s):
...
+    def _from_json_dict_(cls, n, rs, xs, zs, **kwargs):
+        state = cls(n)
+        state.rs = rs.copy()
+        state.xs = xs.copy()
+        state.zs = zs.copy()
+
+        return state
+
+    def __eq__(self, other):
+        if not isinstance(other, type(self)):
```
mpharrigan, line 67: is there a reason why you use `type(self)` instead of writing out the type?

cirq/sim/clifford/stabilizer_state_ch_form.py
```diff
@@ -64,8 +66,32 @@ def bits(s):
...
+
+        return copy
+
+    def _value_equality_values_(self):
+        return (self.n, self.initial_state, self.G, self.F, self.M, self.gamma,
+                self.v, self.v, self.s, self.omega)
+
     def copy(self) -> 'cirq.StabilizerStateChForm':
-        copy = StabilizerStateChForm(self.n)
+        copy = StabilizerStateChForm(self.n, self.initial_state)
```
mpharrigan, line 94: After looking at this code: why is `initial_state` stored on the class? Like the other `initial_state` arguments it should be used to set up the initial state but the class's "data" should just be the current state. Can you do a git blame to figure out who we should tag with this question?

cirq/sim/clifford/clifford_simulator.py
```diff
@@ -279,6 +280,25 @@ def __init__(self, qubit_map, initial_state=0):
...
+            'tableau': self.tableau,
+            'ch_form': self.ch_form,
+        }
+
+    @classmethod
+    def _from_json_dict_(cls, qubit_map, tableau, ch_form, **kwargs):
+        state = cls(dict([(v, int(k)) for k, v in qubit_map.items()]))
+        state.tableau = tableau.copy()
+        state.ch_form = ch_form.copy()
```
mpharrigan, line 295: the copy is likely unnecessary?

cirq/sim/clifford/clifford_tableau.py
```diff
@@ -50,6 +51,26 @@ def bits(s):
             self.xs[i, i] = True
             self.zs[self.n + i, i] = True
 
+    def _json_dict_(self):
+        return protocols.obj_to_dict_helper(self, ['n', 'rs', 'xs', 'zs'])
+
+    @classmethod
+    def _from_json_dict_(cls, n, rs, xs, zs, **kwargs):
+        state = cls(n)
+        state.rs = rs.copy()
```
mpharrigan, line 60: unnecessary copy

cirq/sim/clifford/clifford_tableau.py
tonybruguier-google: It seems to be used elsewhere in the codebase, e.g.
https://github.com/quantumlib/Cirq/blob/740ef6c8034b25a0d2e6b763600b9f8cc5f5c899/cirq/ops/gate_operation.py#L83
or
https://github.com/quantumlib/Cirq/blob/740ef6c8034b25a0d2e6b763600b9f8cc5f5c899/cirq/value/timestamp.py#L80

I suspect it's because of inheritance, but my Python-fu is weak. I was just imitating what I had read.

cirq/sim/clifford/stabilizer_state_ch_form.py
tonybruguier-google: @vtomole in this PR:
https://github.com/quantumlib/Cirq/pull/2129

It is used later on but only for display, here:
https://github.com/quantumlib/Cirq/blob/740ef6c8034b25a0d2e6b763600b9f8cc5f5c899/cirq/sim/clifford/stabilizer_state_ch_form.py#L87

I figured it was more correct to add it, because it might confuse someone who is debugging, although I don't think it would cause incorrect simulation results. Happy to keep it in this PR, add it to another PR, or just revert.

cirq/sim/clifford/clifford_simulator.py
tonybruguier-google: Removed.

cirq/sim/clifford/clifford_tableau.py
tonybruguier-google: Removed.

cirq/sim/clifford/clifford_simulator.py
tonybruguier-google: Done.

tonybruguier-google:
I commented about the value of `initial_state` being stored in the comment above. I don't think it's catastrophic.

I added some round-trip test. Since it seems that there is only one test file for all the objects, I added the test for all three objects there.

vtomole:
@mpharrigan  Yes. you are right. That `repr` should `Return the CH form representation of the state.`, so we should be returning the state not the inital state.

tonybruguier-google:
Removed the initial_state member variable. PTAL.

cirq/sim/clifford/stabilizer_state_ch_form.py
```diff
@@ -84,7 +107,7 @@ def __str__(self):
     def __repr__(self):
         """Return the CH form representation of the state. """
         return ('StabilizerStateChForm(num_qubits={!r}, '
-                'initial_state={!r})').format(self.n, self.initial_state)
+                'to_state_vector={!r})').format(self.n, self.to_state_vector())
```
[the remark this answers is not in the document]
tonybruguier-google: Like this?

cirq/protocols/json_test_data/StabilizerStateChForm.repr
```diff
@@ -0,0 +1 @@
+cirq.StabilizerStateChForm(num_qubits=1)
```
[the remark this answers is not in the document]
tonybruguier-google: So we only have one argument for the constructor now, right?

tonybruguier-google:
Thanks. PTAL. I'm not sure I did what you suggested, but I tried. Let me know if that is still not correct. Happy to work more on it. This is not urgent. Thank you.

PR #2898, opened 2020-04-09
Use proper module name in doctest

kevinsung:
For instance, `cirq.sim.states` instead of `states` and `cirq.sim` instead of `cirq.sim.__init__`. This is not a hack but it is needed for the module-editing hack used to deprecate `sim.STATE_VECTOR_LIKE` in #2808 .

### DIFF
--- a/dev_tools/run_doctest.py
+++ b/dev_tools/run_doctest.py
@@ -192,9 +192,16 @@ def import_file(file_path: str) -> ModuleType:
 
     Returns: The imported module.
     """
-    mod_name = os.path.basename(file_path)
+    mod_name_components = []
+    head, tail = os.path.split(file_path)
+    while tail:
+        mod_name_components.append(tail)
+        head, tail = os.path.split(head)
+    mod_name = '.'.join(reversed(mod_name_components))
     if mod_name.endswith('.py'):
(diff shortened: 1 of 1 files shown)


kevinsung:
Wait, I think there's something wrong, since executing ``check/doctest` gives `Passed: 0`

kevinsung:
@cduck I wonder if you have thoughts about this. In this PR, all I do is change the value of `mod_name` used in the line https://github.com/quantumlib/Cirq/blob/68bf2994412c8b5f21b0e15b5cf9c4581ec3da52/dev_tools/run_doctest.py#L199 Instead of using the filename, I use the fully qualified module name. Strangely (to me), this causes doctest to think that the classes in the file do _not_ belong to the module, so it ignores them, and no doctests are run. Using a random string as `mod_name` is fine though. So this test currently relies on the line above not being passed the fully qualified module name. I wonder if there is a way to rewrite the test so that is not the case.

cduck:
@kevinsung The script imports all of Cirq then imports the file to test.  My guess is the file was already imported along with Cirq so the name matches something already in sys.modules and it won't be imported again.  There are ways to force it or you can delete it from sys.modules first.  I'll have more time to look into this next week.

PR #2899, opened 2020-04-09
Fix invalid argument in list_processors.

95-martin-orion:
Calling `engine.list_processors` currently fails, as in the grpc client the second argument must be a number. This fixes the issue, but does not fully test for it, as that would require an integration test.


95-martin-orion:
Opened #2900 to track better testing of this kind of issue.

PR #2906, opened 2020-04-14
Add some missing type annotations

viathor:
Also, replace calls to `format` with format strings.

### DIFF
--- a/cirq/contrib/acquaintance/gates.py
+++ b/cirq/contrib/acquaintance/gates.py
@@ -16,8 +16,8 @@ import functools
 import itertools
 import math
 import operator
-from typing import Sequence, Dict, Tuple, List, NamedTuple, Optional, \
-    TYPE_CHECKING
+from typing import (Dict, Iterable, List, NamedTuple, Optional, Sequence, Tuple,
+                    TYPE_CHECKING)
 
 from cirq import ops, protocols, value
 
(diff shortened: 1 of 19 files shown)


cirq/contrib/acquaintance/gates.py
```diff
@@ -309,7 +310,8 @@ def _decompose_(self, qubits: Sequence['cirq.Qid']) -> 'cirq.OP_TREE':
         if final_gate:
             yield final_gate(*qubits)
 
-    def _circuit_diagram_info_(self, args: 'cirq.CircuitDiagramInfoArgs'):
+    def _circuit_diagram_info_(self, args: 'cirq.CircuitDiagramInfoArgs'
+                              ) -> 'cirq.CircuitDiagramInfo':
```
dstrain115, line 314: How come one of the _circuit_diagram_info_ returns Iterable[str] and the other returns cirq.CircuitDiagramInfo?

I forget.  Does this protocol allow both?

cirq/contrib/acquaintance/permutation.py
```diff
@@ -143,12 +144,12 @@ def permutation(self) -> Dict[int, int]:
...
         yield self.swap_gate(*qubits)
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return ('cirq.contrib.acquaintance.SwapPermutationGate(' +
                 ('' if self.swap_gate == ops.SWAP else repr(self.swap_gate)) +
                 ')')
 
-    def _value_equality_values_(self):
+    def _value_equality_values_(self) -> Any:
```
dstrain115, line 152: Is Any the correct type here?  Should it be a Tuple or Iterable of Any instead?

Same with the others below.

viathor: Yes. More precisely `Any` is the most general type supported by the [value equality](https://github.com/quantumlib/Cirq/blob/master/cirq/value/value_equality.py#L27) protocol. In practice, if you like you can narrow it down to a more concrete type (since mypy allows return type covariance), but this may cause problems in derived classes (if they want to use the more general type). Since we subclass `PermutationGate` quite a lot I decided to use the most general type allowed. We did the same in quirk's [arithmetic operation class](https://github.com/quantumlib/Cirq/blob/master/cirq/interop/quirk/cells/arithmetic_cells.py#L75).

cirq/contrib/acquaintance/gates.py
viathor: Yes, mypy supports return type covariance and the protocol supports `str`, `Iterable[str]` and `CircuitDiagramInfo`, see [here](https://github.com/quantumlib/Cirq/blob/master/cirq/protocols/circuit_diagram_info_protocol.py#L182). In this case I chose the most concrete type because we don't derive any subclasses from these.

cirq/contrib/acquaintance/permutation.py
viathor: There is another reason `Any` is better here but not in `_circuit_diagram_info_`: we want values returned from `_value_equality_values_` to be treated as opaque.

PR #2910, opened 2020-04-15
Remove explicit object superclass

viathor:
Cleaning up a vestige of python 2.


viathor:
Yes, except for a few in protobuf code which is generated.

PR #2911, opened 2020-04-15
Compare pauli matrices in commute using equality instead of `is`

### LINKED ISSUE #2813: Pauli operator deserialization, and commutation checking
When you serialize and then deserialize a Pauli operator, you get a different instance, which might not be too surprising
```python
cirq.read_json(json_text=cirq.to_json(cirq.Z)) is cirq.Z

False
```
This shouldn't be a problem, however Paulis are checking commutation using `is` instead of `==`, so commutation with deserialized Paulis is broken.

tanujkhattar:
To fix #2813


cirq/ops/pauli_gates.py
```diff
@@ -52,7 +52,7 @@ def _commutes_(self, other: Any,
                    atol: float) -> Union[bool, NotImplementedType, None]:
         if not isinstance(other, Pauli):
             return NotImplemented
-        return self is other
+        return self == other
```
viathor, line 55: I think that rather than making this work with multiple Pauli instances, we could instead change `_Pauli{X,Y,Z}._from_json_dict_` to return one of the three instances stored in `Pauli._XYZ`. This way we always have a single instance of each Pauli and the `is` operator` is fine. This is also consistent with how we do things in methods like `by_index` and `third`.

Is there a reason this wouldn't work?

(One obstacle to this is that the internal classes `_Pauli{X,Y,Z}` take the exponent argument. I think they shouldn't and filed #2920 about this.)

viathor: Now that you have fixed #2920, could you restore this line to `self is other` and instead modify `_Pauli{X,Y,Z}._from_json_dict_` to return the appropriate Pauli instance from `Pauli._XYZ`? This should fix #2813.

tanujkhattar: Done. Also added  a test which fails before this change.

tanujkhattar:
Updated as per the suggestion. PTAL.

PR #2912, opened 2020-04-15
Add ZerosSampler

### LINKED ISSUE #2818: ZerosSampler
Often it is useful to test your workflow can successfully collect data without caring too much about what that data is. 

I propose a `ZerosSampler` that just immediately returns a bunch of zeros but otherwise follows the `cirq.Sampler` interface. This can also run the program through the proto serialization routines to verify that the gateset is correct. 

One final use case for this is profiling the "classical overhead" when running a complicated pipeline by removing the quantum part

fedimser:
A dummy sampler which implements `cirq.Sampler` and returns zeroes for all measurements.

Fixes #2818


cirq/work/zeros_sampler.py
```diff
@@ -0,0 +1,63 @@
...
+        Args:
+            program: The circuit to sample from.
+            params: Parameters to run with the program.
+            repetitions: The number of times to sample.
+
+        Returns:
+            TrialResult list for this run; one for each possible parameter
+            resolver.
+        """
+        zero_measurement = np.zeros((repetitions, 1), dtype=np.int8)
```
mpharrigan, line 46: this doesn't work correctly for circuits with multi-qubit measurement gates

fedimser: I didn't know Cirq support that :)
Added support for multi-qubit measurement, PTAL.

cirq/work/zeros_sampler_test.py
```diff
@@ -0,0 +1,54 @@
...
+    assert result[0].measurements.keys() == {'a', 'b,c'}
+    assert result[0].measurements['a'].shape == (3, 1)
+    assert np.all(result[0].measurements['a'] == 0)
+    assert result[0].measurements['b,c'].shape == (3, 2)
+    assert np.all(result[0].measurements['b,c'] == 0)
+
+
+def test_sample():
+    # Create a circuit whose measurements are always zeros, and check that
+    # results of ZeroSampler on this circuit are identical to results of
```
mpharrigan, line 39: 👍

cirq/work/zeros_sampler.py
```diff
@@ -0,0 +1,56 @@
...
+            program: The circuit to sample from.
+            params: Parameters to run with the program.
+            repetitions: The number of times to sample.
+
+        Returns:
+            TrialResult list for this run; one for each possible parameter
+            resolver.
+        """
+        measurements = {}  # type: Dict[str, np.ndarray]
+        for op in ops.flatten_op_tree(iter(program)):
```
mpharrigan, line 47: use `program.all_operations`?

fedimser:
Hi Matthew,

1. ZeroSampler inherits from Sampler, which has default implementation of `run_async` which  calls `run` and of `run_sweep_async` which calls `run_sweep`. So now async methods of ZerosSampler will work as expected, and I don't need to implement them. Am I right?
2. I didn't work with gate sets in Cirq before, but here is what I can suggest. We can pass `SerializableGateSet` object to ZeroSampler constructor. Then in `run_sweep` we will check for all operations whether it `is_supported_operation`. And if there is no gate set, don't do this validation. Does it make sense? Do you think it's better to do this in a follow-up PR?
Also, what are "extand qubits"?

mpharrigan:
1: ok good
2: that sounds entirely reasonable to me
2b, re: follow up PR. up to you. If it's not too drastic of a change, I'd say just put it into this because it will likely involve changing the API of `ZerosSampler` slightly
2c: re: extant qubits. real devices only have certain qubits and a common failure mode is accidentally using e.g. a LineQubit when you should use GridQubits or using GridQubit(0,0) when the physical numbering scheme starts somewhere else

fedimser:
Hi @mpharrigan,
I added gateset validation, PTAL.

Regarding extand qubits: there is a class cirq.devices.Device, which has method qubit_set. So we could pass the device to a constructor (like I did with gate set) and in run_sweep() get all qubits referenced by a device, and check that they are all in this qubit_set.
However, this can be simplified: we could just pass qubit_set. Or, if we are passing device, we can just call Device.validate_circuit. 
What do you prefer? 
Also, does Device.validate_circuit validates gateset? If so we don't need SerializableGetSet in ZerosSampler (which I just added).

dabacon:
Question to @mpharrigan : shouldn't the device and gateset validation be handled by setting an appropriate device on the circuit?  What is the primary use case for this class, is it to do this validation?  Then maybe it should be called something slightly different?

cirq/work/zeros_sampler.py
```diff
@@ -0,0 +1,70 @@
...
+import abc
+from typing import (Dict, List, TYPE_CHECKING)
+
+import numpy as np
+
+from cirq import work, study, protocols
+
+if TYPE_CHECKING:
+    import cirq
+    from cirq.google.serializable_gate_set import SerializableGateSet
```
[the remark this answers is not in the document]
fedimser: Done, except now I must violate line length limit.

cirq/work/zeros_sampler_test.py
```diff
@@ -0,0 +1,71 @@
...
+
+def test_sample_with_gate_set():
+    gate_set = SerializableGateSet('test', SINGLE_QUBIT_SERIALIZERS,
+                                   SINGLE_QUBIT_DESERIALIZERS)
+    sampler = cirq.ZerosSampler(gate_set=gate_set)
+    a, b = cirq.LineQubit.range(2)
+    circuit1 = cirq.Circuit([cirq.X(a)])
+    circuit2 = cirq.Circuit([cirq.CX(a, b)])
+
+    sampler.sample(circuit1)
```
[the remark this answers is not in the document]
fedimser: Done.

fedimser:
Can we merge this?
@mpharrigan @dabacon

balopat:
@googlebot  please rescan

PR #2914, opened 2020-04-15
More type annotations and format strings

viathor:
Follow-up to #2908.

### DIFF
--- a/cirq/circuits/_bucket_priority_queue.py
+++ b/cirq/circuits/_bucket_priority_queue.py
@@ -156,18 +156,19 @@ class BucketPriorityQueue(Generic[TItem]):
 
         return priority, item
 
-    def __str__(self):
+    def __str__(self) -> str:
         lines = [
             '{}: {},'.format(p, e)
             for p, e in self
         ]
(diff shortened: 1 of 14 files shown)


cirq/circuits/_bucket_priority_queue.py
```diff
@@ -156,18 +156,19 @@ def dequeue(self) -> Tuple[int, TItem]:
...
 
-    def __repr__(self):
-        return '{}(entries={!r}, drop_duplicate_entries={!r})'.format(
-            'cirq.circuits._bucket_priority_queue.BucketPriorityQueue',
-            list(self),
-            self._drop_set is not None)
+    def __repr__(self) -> str:
+        entries = list(self)
+        drop_duplicate_entries = self._drop_set is not None
+        return (f'cirq.circuits._bucket_priority_queue.BucketPriorityQueue('
```
dstrain115, line 169: nit: I don't think you need a f on this first part of the string.

viathor: Done.

PR #2918, opened 2020-04-16
DFE with exhaustive sampling for exhaustive Clifford Pauli strings

tonybruguier-google:
In case of a Clifford exhaustive enumeration of the Pauli strings, we can sample each of them only once (since they all have the same probability).

I think it makes sense to add, but I am more than happy to withdraw if not.


examples/direct_fidelity_estimation.py
```diff
@@ -263,6 +263,37 @@ def _estimate_pauli_traces_general(qubits: List[cirq.Qid],
     return pauli_traces
 
 
+def _sample_pauli_traces(pauli_traces, n_trials, n_clifford_trials):
```
mpharrigan, line 266: type annotations including return type annotation (for my edification!)

examples/direct_fidelity_estimation.py
```diff
@@ -263,6 +263,37 @@ def _estimate_pauli_traces_general(qubits: List[cirq.Qid],
...
+    """
+    Either samples the Pauli traces or exhaustively enumerate them.
+
+    Args:
+        pauli_traces: The list of Pauli traces for the circuit.
+        n_trial: The total number of Pauli measurements, or None to explore each
```
mpharrigan, line 272: `n_trials`

examples/direct_fidelity_estimation.py
```diff
@@ -370,10 +397,7 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
             noisy_simulator.simulate(circuit)).final_density_matrix
 
     trial_results: List[TrialResult] = []
-    for _ in range(n_trials):
-        # Randomly sample as per probability.
-        i = np.random.choice(len(pauli_traces), p=p)
-
+    for i in _sample_pauli_traces(pauli_traces, n_trials, n_clifford_trials):
```
mpharrigan, line 400: more pythonic, I think: have `_sample_pauli_traces` yield pauli traces and nix `i` (or use `enumerate` if you want it in `TrialResult`)

tonybruguier-google: Done, I think, although I did the enumerate in the `_sample_pauli_traces()` and not here because we want the index of the pauli trace pre-sampling, and putting enumerate here would get the index of the trace post-sampling.

examples/direct_fidelity_estimation.py
tonybruguier-google: Done.

examples/direct_fidelity_estimation.py
tonybruguier-google: Done.

tonybruguier-google:
Addressed the comments. Regarding the meaning of the two params, `n_trials` and `n_clifford_trials` I am also confused if I may be honest.

So for example, if `n_qubit=3`, then the Pauli basis has 3 traces, and thus we can compute `2**3=8` Pauli traces, let's call them `P_0, P_1, ...., P_7`.

Now, how do we sample them? Let's say, `n_clifford_trials=4` and `n_trials=2`. First step is we randomly sample 4 traces, and say we get: `P_0, P_1, P_2, P_0`. Then,*out of this list*, we sample 2, and get, say, `P_2, P_2`.

With the change, we guarantee that if `n_trials=None`, we guarantee that we repeat the 4 traces. I am not certain it's the desired behavior (as a matter of fact, I changed a little bit of the code by removing the exception) because upon writing this, I change my mind.

We could also merge the two params in one.

tonybruguier-google:
Thanks. PTAL.

examples/direct_fidelity_estimation.py
```diff
@@ -370,11 +380,16 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
             noisy_simulator.simulate(circuit)).final_density_matrix
 
     trial_results: List[TrialResult] = []
-    for _ in range(n_trials):
-        # Randomly sample as per probability.
-        i = np.random.choice(len(pauli_traces), p=p)
+    for i_seq in range(len(pauli_traces)):
```
mpharrigan, line 383: not blocking for this PR but keep in mind:

I still feel like this could be structured in a more readable ("pythonic") way. So you have `pauli_traces` and you need to do some of them with some probability that depends on which modality you're operating in. Why not factor the generation of the ones you want to measure outside the loop?

```python
if clifford etc:
  measured_ptraces = pauli_traces
else:
  measure_ptraces = np.choice(pauli_traces, size=len(pauli_traces))

for mpt in measure_ptraces:
  do_jazzy_stuff(mpt.P_i)
  do_more_jazzy_stuff(mpt.rho_i)
  trial_results.append(TrialResult(sigma_i=sigma_i))
```

Every time I see an `i` index in a Python for loop, I get very suspicious.

Related question: what is a user supposed to do with the `i` that you save in the `TrialResult`? is that helpful? Maybe just give them the whole pauli_trace instead of making them have to look it up?

mpharrigan:
Relevant test failure? Strangely only happens on the Mac runs of the test. Could be numerical issues, but the difference looks pretty huge (pytest says 1.09 when it should be <=1). Are you able to reproduce?

examples/direct_fidelity_estimation.py
tonybruguier-google: Done.

tonybruguier-google:
Sorry about accidentally closing this PR. I still would like to get it merged eventually (no hurry).

I did implement the Pythonesque changed you suggested.

I was able to repeat the failure, by running the tests repeatedly, so I think it was a flaky test. I think it's possible to estimate the fidelity to be above 1.0 since we divide sigma_i / rho_i. I then modified `test_direct_fidelity_estimation_with_noise_clifford()` and `test_direct_fidelity_estimation_with_noise_non_clifford()` by adding more samples_per_term nad making them exaustive. Finally, I re-ran the tests 100 times and they all passed.

PTAL.

mpharrigan:
The above ^ raises a related concern that was going to come up eventually anyways; which is it's good to provide hooks for running "random" things deterministically. See some of the simulators for example. You can provide optional seeds / `np.random.RandomState` objects to functions that do randomness. Then you can make tests like this deterministic [fixing the underlying assert statements like you did should also be done!]

viathor:
I agree with the need to run code that generates random numbers deterministically.

However, I  think we should avoid making seeds explicit in the APIs. Smaller APIs with few parameters directly related to the central concern of the code are cleaner and easier to understand, test and maintain than large APIs with a knob for everything(*).

So what can we do instead? Well, if we agree to use a particular PRNG, say the one in numpy, then whenever you need to run some code deterministically you just do this:

```python
prng_state = np.random.get_state()
np.random.seed(my_seed)  # determinism from here on!
...
# calls to functions that internally use numpy's PRNG
...
# be a good citizen and don't interfere with other code (may need try-finally)
np.random.set_state(prng_state)
```

Would this work for you?

@mpharrigan @tonybruguier-google 

---
(*) The problem with large APIs is in some sense similar to the curse of dimensionality in data analysis. Too many parameters open up a large space of possibilities where bugs can hide more easily.

tonybruguier-google:
Works for me. I will close the other PR. Thanks for taking the time to look into it.

mpharrigan:
We follow https://networkx.github.io/documentation/stable/reference/randomness.html which uses the numpy global prng by default if you don't pass anything. There are helper functions floating around in Cirq for parsing a `random_state` argument and applying suitable defaults

PR #2919, opened 2020-04-17
Speed up clifford simulator by short circuiting on known gates

mpharrigan:
This brings run-time down from 1.298s to 0.265s on a particular circuit I was investigating


cirq/sim/clifford/clifford_simulator.py
```diff
@@ -366,14 +379,26 @@ def apply_single_qubit_unitary(self, op: 'cirq.Operation'):
...
+        self.tableau._X(qubit)
+        self.ch_form._X(qubit)
+
+    def _apply_Z(self, qubit: int):
+        self.tableau._Z(qubit)
+        self.ch_form._Z(qubit)
+
+    def _apply_Y(self, qubit: int):
+        self.tableau._Y(qubit)
+        self.ch_form._Y(qubit)
```
viathor, line 400: I think we've decided that Clifford gates will define their effect on the tableau/CH form. IOW, this should be defined in the respective gates, not special cased here. Instead, here you should be invoking a protocol. The protocol would check whether a gate defines an effect on tableau/CH form and if so would use it for efficiency. If not, it would fall back to computing the effect.

This is also consistent with `_apply_unitary_`.

@Strilanc @fedimser 

(Also, I wonder whether it makes sense to keep both the tableau and CH form.)

vtomole: Thanks for the reminder. This is stated in https://github.com/quantumlib/Cirq/issues/2423. Also, with respect to your last statement; here is my take on it: https://github.com/quantumlib/Cirq/issues/2423#issuecomment-549546013

vtomole: `_clifford_tableau_` hasn't been created yet so we can block this and wait or merge and have it be fixed once `_clifford_tableau_`  has been created. I'm down for either as this speeds up the simulation.

smitsanghavi: Given that using the current implementation (which was recently added) is significantly slowing down the simulation, I'd vote for letting this go through. 

I will work on implementing the `_clifford_tableau_` method and protocol and clean this up ASAP.

cirq/sim/clifford/clifford_simulator.py
```diff
@@ -366,14 +379,26 @@ def apply_single_qubit_unitary(self, op: 'cirq.Operation'):
...
+    def _apply_X(self, qubit: int):
+        self.tableau._X(qubit)
+        self.ch_form._X(qubit)
+
+    def _apply_Z(self, qubit: int):
+        self.tableau._Z(qubit)
+        self.ch_form._Z(qubit)
+
+    def _apply_Y(self, qubit: int):
+        self.tableau._Y(qubit)
```
smitsanghavi, line 399: I think it would be cleaner if _apply_Gate methods accept the actual Qid and internally resolve that to self.qubit_map[Qid] instead of having the caller do that since the qubit map is an internal implementation detail of the state. (thinking in terms of eventually moving the tableau application to the Gate definitions)

cirq/sim/clifford/clifford_simulator.py
```diff
@@ -366,14 +379,26 @@ def apply_single_qubit_unitary(self, op: 'cirq.Operation'):
...
+        self.ch_form._Y(qubit)
+
```
smitsanghavi, line 401: Also add and use _apply_CZ and _apply_CNOT for completeness.

cirq/sim/clifford/clifford_simulator.py
smitsanghavi: @viathor are you okay with letting this optimization merge in and I can clean up later.

cirq/sim/clifford/clifford_simulator.py
smitsanghavi: @viathor Nothing really. Ideally we can take this to conclusion by removing the _GATE methods on the tableau and CH form too and let individual gates modify the internal states of tableau and the CH form. It will be more involved but I can make that change.

Do we want to have two separate protocols: `apply_clifford_tableau` and `apply_stabilizer_ch_form` or combine them into one `update_clifford_state`? (names subject to change)

@mpharrigan Feel free to ignore the initial suggestion in this comment.

mpharrigan:
Thanks all. I agree that a more principled solution is needed; but wanted to share this small change which helped my work and doesn't make anything worse :)

PR #2921, opened 2020-04-17
Loosen tolerance restrictions in Quantum Engine

dstrain115:
- It looks like serialization and deserialization of
floats can have an error up to around 5e-8, so loosening the
restriction to 1e-6 to be safe, since this is only used to determine
gates close to the desired gates.

### DIFF
--- a/cirq/google/common_serializers.py
+++ b/cirq/google/common_serializers.py
@@ -37,17 +37,25 @@ PHYSICAL_Z = 'physical'
 VIRTUAL_Z = 'virtual_propagates_forward'
 
 
-def _near_mod_n(e, t, n, atol=1e-6):
+# Default tolerance for differences in floating point
+# Note that Google protocol buffers use floats
+# which trigger a conversion from double precision to single precision
+# This results in errors possibly up to 1e-6
+# (23 bits for mantissa in single precision)
+_DEFAULT_ATOL = 1e-6
+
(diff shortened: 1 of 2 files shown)


dstrain115:
> This is concerning. Serializing and then deserializing a float should give you the exact same value since no rounding or approximation is involved. If this is not the case it is probably due to the fact that too few significant digits are saved.
> 
> For single-precision floats you need to save [9 significant digits](https://en.wikipedia.org/wiki/Single-precision_floating-point_format) and for double-precision floats you need [17 significant digits](https://en.wikipedia.org/wiki/Double-precision_floating-point_format).

Theta and phi are stored as floats in the proto, which are 32 bits, which I believe is 23 bits for the mantissa which is about 7 digits.  The page you linked to even states that "This gives from 6 to 9 significant decimal digits precision." for single precision.

I'm not totally sure where the conversion is happening, but I think it is mapping from a double in python to single precision in the proto.

We could consider using doubles in the proto, but that is a much bigger change and probably not worth it, since our errors are much bigger than 1e-7.

dstrain115:
> Could you add a comment somewhere in the file explaining that the reason for the tolerance is that serialization and deserialization entail implicit conversion between single- and double-precision?

I split out the atol into a _DEFAULT_ATOL variable and put the comment there.

PR #2922, opened 2020-04-20
A example of Qec using Shor's Code

crystalzhaizhai:
This is an example of using Shor's code to do quantum error correction and fault tolerant operations.

There is a base class for one qubit code, which encodes one physical qubit with several ancilla qubits and some gates. The Shor's Code class will be derived from the base one qubit class. Then it's easy for us to add other codes in the future. 
There are multi-qubit codes that take in multiple physical qubits and call one qubit class for each of the input qubits. The class also takes in the original circuit of the original physical qubits and transforms it into fault-tolerate operations on the encoded circuits. 

The files are under examples/qec
  | __init__.py |  
  | onequbit_qec.py |  
  | multiqubit_qec.py |  
  | shors_code.py |  
  | fault_tolerate_operations.py |  
 
The tests are added to the end of existing examples_test.py


examples/qec/multiqubit_qec.py
```diff
@@ -0,0 +1,56 @@
+"""
+This is code for multiple physical qubits
+"""
+import random
+from typing import List
+
+import cirq
+import fault_tolerate_operations as ops
+from onequbit_qec import OneQubitCode
```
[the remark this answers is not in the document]
vtomole: This is not the line. The line is in `shors_code.py`, not this file. `examples/qec/shors_code.py:26: error: Cannot find module named 'onequbit_qec'.`

dstrain115:
Can you add a description to the PR?

crystalzhaizhai:
> Can you add a description to the PR?

crystalzhaizhai:
> Can you add a description to the PR?

Just added some description. Please let me know if it looks good:)

examples/qec/multiqubit_qec.py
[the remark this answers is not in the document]
crystalzhaizhai: I realized it cannot use relative path

examples/qec/shors_code.py
```diff
@@ -0,0 +1,88 @@
+"""
```
balopat, line 1: nit: minus new line - all the examples start with `""" Text of the example` without the newline

examples/qec/onequbit_qec.py
```diff
@@ -0,0 +1,28 @@
+"""
+This is the base class for code for one physical qubit
+"""
+
+import cirq
+
+
+class OneQubitCode:
```
balopat, line 8: I feel like this is abstraction is not important for an example, can you just merge it into the `OneQubitShorsCode`?

vtomole: I think it's because we plan on adding more codes besides Shor's as examples in the future.

balopat:
Right - I think we should discuss the end goal here first then. Let's bring
it up on the Circ Cync, to clarify it!

examples/qec/onequbit_qec.py
crystalzhaizhai: base class removed

examples/qec/shors_code.py
crystalzhaizhai: done

crystalzhaizhai:
> 1. the



> Thanks for your patience on this!
> I know that this was originally intended to be part of the codebase. However now that it became an example I think we should simplify it:
> 
> 1. A single file with where we walk through the user who wants to learn about Shor's code should be sufficient. It would be great to show also the behavior of the code with a main function. For example: create circuit, create error, detect error and fix it
> 2. I think a single qubit shor code should be sufficient as a starting point / textbook example. Let's discuss on Cirq Cync whether we want to have an example for multiple qubits.
> 
> Thank you!

1. Main function added.
2. Trim the module down to a simple shor's code example

examples/shors_code.py
```diff
@@ -0,0 +1,118 @@
...
+        ])
+
+    def apply_gate(self, gate: cirq.Gate, pos: int):
+        if pos > self.num_physical_qubits:
+            raise IndexError
+        else:
+            return gate(self.physical_qubits[pos])
+
+    def correct(self):
+        yield cirq.ops.Moment([
```
balopat, line 63: This is not correct. 

To correct errors in an error correction code, you have to first measure the syndromes, and use those results. I'm really not sure what this code does - it looks like it reverses the encoding - but I don't think that's how you decode per se - I think you'll have to measure the right Zlogical =XXXXXXXX and Xlogical=ZZZZZZZZZ observables, but I really want to find a good description to prove my hunch here.

examples/shors_code.py
```diff
@@ -0,0 +1,118 @@
+""" Shor's code is a stabilizer code for quantum error correction.
+It uses 9 qubits to encode 1 logic qubit and is resistant
+to at least one bit flip and one sign flip or their combination.
+
```
balopat, line 4: I would make this text flow like this: 

- encode with phase flip a state alpha |0> + beta |1>  to alpha |+++>  + beta |---> 
- concatenate it with the bit-flip code 
- a stabilizer code is defined by the stabilizers, which are Z1Z2, Z2Z3, Z4Z5, Z5Z6 ... and X1X2...X6, X4X5...X9 plus the two encoded logical operators (Xlogical = ZZZZZZZZZ, Zlogical = XXXXXXXXX)
- let's introduce an X error 
- one or two of the ZZ syndromes should light up 
- let's introduce a Z error 
- one or both of the XXXX.. syndromes should result in -1 measurement 
- introduce recovery condititonal on syndrome values 
- measurement in stabilizer codes should be a stabilizer itself, to keep the code in the codespace as far as I understand (but I might be wrong here) and not measurement in the Z basis for each qubits...
- we should be able to show that both the phase flip and bit flip errors are corrected by the recovery method

examples/shors_code.py
```diff
@@ -0,0 +1,118 @@
...
+        yield cirq.ops.Moment(
+            [cirq.CNOT(self.physical_qubits[0], self.physical_qubits[3])])
+        yield cirq.ops.Moment(
+            [cirq.CNOT(self.physical_qubits[0], self.physical_qubits[6])])
+        yield cirq.ops.Moment([
+            cirq.CCNOT(self.physical_qubits[3], self.physical_qubits[6],
+                       self.physical_qubits[0])
+        ])
+
+    def measure(self):
```
balopat, line 95: I don't think this is correct either. Measuring the stabilizer formalism should keep the system in the codespace.

cirq/ops/pauli_string_test.py
```diff
@@ -23,7 +23,6 @@
 import cirq
 import cirq.testing
 
-
```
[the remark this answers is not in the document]
crystalzhaizhai: I don't know how it got in here. At first I didn't change anything in this file, but when I push, it raises an error. So I have to modify this file to pass lint.

examples/shors_code.py
crystalzhaizhai: For Shor's code, the "correct" step is the same as long as the errors is within the scope of the code. No need to measure before correct. 
Basically, you do the encode and the error (X, Z or Y) happens or no error, and then performs correct. Then you measure the first qubit in Z basis, the result is corrected original qubit.

Usually stablizer code needs to measure before correct because the correct will depend on the result of the measurement. But Cirq doesn't have classical register, so we are not able to implement such codes for the time being.

examples/shors_code.py
```diff
@@ -0,0 +1,118 @@
...
+(0, 4): ───────┼───────X───┼───X───┼───@───────────┼───┼───M───
+               │           │       │   │           │   │
+(0, 5): ───────┼───────────X───────X───@───────────┼───┼───M───
+               │                                   │   │
+(0, 6): ───────X───H───@───@───@───@───X───H───────X───@───M───
+                       │   │   │   │   │
+(0, 7): ───────────────X───┼───X───┼───@───────────────────M───
+                           │       │   │
+(0, 8): ───────────────────X───────X───@───────────────────M───
+"""
```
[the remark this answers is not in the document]
crystalzhaizhai: I read how the code works and made the graph. So I will just say according to Peter Shor's work

examples/shors_code.py
crystalzhaizhai: If it's measurement of syndromes, then yes for sure and should be placed before correction. But Shor's code doesn't need that step. The measurement here is just to read the final result. We actually only need to read the first qubit, so no problem if we want to get rid of measurements of other qubits. I measure all the qubits for debugging reasons.

examples/shors_code.py
crystalzhaizhai: I think the text flow you suggested looks great! I will format the steps into this.

Yes we are able to show both phase flip and bit flip errors.

examples/shors_code.py
balopat: I'm standing corrected (no pun intended). After discussions with the team and you, we can skip the stabilizer formalism aspect of the Shor code. We can focus on the encode and decode/correct form.

Regarding the classical register - that is true, however, in Cirq you can create conditional gates on measured qubits, which would work similarly well, as seen in our [quantum_teleportation.py](https://github.com/quantumlib/Cirq/blob/master/examples/quantum_teleportation.py) example

examples/shors_code.py
balopat: After our discussions: let's keep it simple as it is now - no need to introduce the stabilizer formalism related operators / syndromes.

cirq/ops/pauli_string_test.py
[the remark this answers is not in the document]
balopat: haha, that is weird - do you mind if I push commits and experiment on your branch?

examples/shors_code.py
```diff
@@ -0,0 +1,118 @@
+""" Shor's code is a stabilizer code for quantum error correction.
+It uses 9 qubits to encode 1 logic qubit and is resistant
+to at least one bit flip and one sign flip or their combination.
```
balopat, line 3:
```suggestion
at most one bit flip and one sign flip or their combination.
```

examples/shors_code.py
```diff
@@ -0,0 +1,118 @@
+""" Shor's code is a stabilizer code for quantum error correction.
+It uses 9 qubits to encode 1 logic qubit and is resistant
```
balopat, line 2:
```suggestion
It uses 9 qubits to encode 1 logic qubit and is able to correct
```

cirq/ops/pauli_string_test.py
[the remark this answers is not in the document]
crystalzhaizhai: not at all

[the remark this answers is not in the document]
crystalzhaizhai: No worries. Just reverted the file and the Format Check below doesn't raise error, so it should be fine. My local check/format-incremental still demands format change.

examples/shors_code.py
balopat: Let's skip the measure method, I think it is confusing, as with this encode/decode form the only qubit of interest to be measured would be the first qubit.

examples/shors_code.py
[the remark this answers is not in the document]
balopat: Sounds good, let's just put down Shor's paper as the reference. In the end I realized that this is fairly standard decoding for the phase-flip and bit-flip codes that is combined here.

examples/shors_code.py
```diff
@@ -0,0 +1,118 @@
...
+    def measure(self):
+        for i in range(self.num_physical_qubits):
+            yield cirq.measure(self.physical_qubits[i])
+
+
+if __name__ == '__main__':
+    # coverage: ignore
+    #create circuit
+    mycode1 = OneQubitShorsCode()
+    my_circuit1 = cirq.Circuit(mycode1.encode())
```
balopat, line 104: What do you think about this: 

```
    # coverage: ignore

    #create circuit with 9 physical qubits
    code = OneQubitShorsCode()

    circuit = cirq.Circuit(code.apply_gate(cirq.X ** (1 / 4), 0))
    print(cirq.dirac_notation(circuit.final_state_vector(initial_state=0)))

    circuit += cirq.Circuit(code.encode())
    print(cirq.dirac_notation(circuit.final_state_vector(initial_state=0)))

    #create error
    circuit += cirq.Circuit(
        code.apply_gate(cirq.X,
                           random.randint(0, code.num_physical_qubits - 1)))
    print(cirq.dirac_notation(circuit.final_state_vector(initial_state=0)))

    #correct error and decode
    circuit += cirq.Circuit(code.correct())
    print(cirq.dirac_notation(circuit.final_state_vector(initial_state=0)))

```

crystalzhaizhai: 1 Yeah it makes sense to delete measure method.
2 I put down Shor95 as reference.
3 Looks good! Merging the newly added final_state_vector is blocked right now. It seems I need to address the requested change first.

crystalzhaizhai: What is the purpose of cirq.X**(1/4) before encode?

balopat: It is to create an interesting superposition - in the end any  should be preserved over an error channel of up to one phase and bit-flip errors or the combination of them. We can show at the end that after decoding the first qubit is still in the same state. 

While you were showing this with |0> I think a more interesting superstition can drive the message home even better.

crystalzhaizhai: Great idea! 

Using dirac notation allows us to show superposition well.

crystalzhaizhai: Just pushed the changes. Also changed the test file.

crystalzhaizhai:
@balopat 
Hi, I am trying to figure out the CLA issue. It seems to need a explicit approval from your review to lift "change requested" and then change the cla to yes. Otherwise the cla is automatically reverted.

crystalzhaizhai:
> LGTM, this is now clean - I do have some more ideas to iterate on and make it more educational, but we can do that as a follow-up PR:
> 
> * represent the bit flip code and the phase flip code explicitly in the methods
> * represent the decoding of the bit flip code and the phase flip code explicitly in the methods
> * represent the concatenation between the two codes explicitly as methods somehow
> * provide more explanation in the description of the example

Thank you very much! Yeah we can do a follow-up PR to integrate these ideas!

I can see how separating the bit flip and the phase flip code can help users understand the logic behind the code better. In the description, it would be easier to explain there are two layers of code involved.

Issue #2832, opened 2020-03-11
Use relevant gate info in metrics-to-noise model


95-martin-orion:
Split off from #2807. This issue specifically tracks improvements in the parameters of noise applied to specific gate types; at the time of writing, all non-measurement gates generate the same DepolarizingChannel regardless of type or duration.

95-martin-orion:
As noted in PR #2715, since the data for this model is Google-specific, we should move it to `cirq.google` when we get the chance.

dstrain115:
@95-martin-orion Is this bug still useful?  If so, can you add more information about it so it can be done by a contributor?  Otherwise, let's close it.

95-martin-orion:
IMO #4640 and related issues and sub-PRs have resolved this. If we want further changes to the metrics-to-noise conversion process, it should be filed under a separate issue.

PR #2927, opened 2020-04-21
Update noise model to new metrics.

95-martin-orion:
Related to #2832. Per-gate Pauli error is more accurate than total error for this model, as it is fed directly into a depolarizing channel.

The other noise channels are already using their preferred metrics, and those metrics were left unchanged in the recent update.


cirq/contrib/noise_models/noise_models.py
```diff
@@ -321,11 +321,9 @@ def simple_noise_from_calibration_metrics(calibration: engine.Calibration,
     readout_error_probs: Dict['cirq.Qid', float] = {}
 
     if depol_noise:
-        # TODO: replace with Pauli error once it's available.
-        # Github issue: https://github.com/quantumlib/Cirq/issues/2832
         depol_probs = {
             qubit[0]: depol_prob[0] for qubit, depol_prob in
-            calibration['single_qubit_rb_total_error'].items()
+            calibration['single_qubit_rb_pauli_error_per_gate'].items()
```
viathor, line 326: Depolarization probability and Pauli error probability are equal only approximately. The approximation is good for a large number of qubits, but for a single-qubit the two quantities are quite different. See for example equation (42) on p.11 in the [supplement to the quantum supremacy paper](https://arxiv.org/abs/1910.11333). In the single-qubit case, it yields:

```python
depol_prob = sq_pauli_error_prob * 4 / 3
```

95-martin-orion: Thanks for catching this! Added comments and adjusted to reflect this.

cirq/contrib/noise_models/noise_models_test.py
```diff
@@ -355,6 +355,10 @@ def test_aggregate_decay_noise_after_moment():
     }]
 """, v2.metrics_pb2.MetricsSnapshot())
 
+depol_001 = .001 * 4 / 3
+depol_002 = .002 * 4 / 3
+depol_003 = .003 * 4 / 3
```
viathor, line 360: nit: Use capital letters for global constants.

cirq/contrib/noise_models/noise_models.py
```diff
@@ -321,11 +321,11 @@ def simple_noise_from_calibration_metrics(calibration: engine.Calibration,
     readout_error_probs: Dict['cirq.Qid', float] = {}
 
     if depol_noise:
-        # TODO: replace with Pauli error once it's available.
-        # Github issue: https://github.com/quantumlib/Cirq/issues/2832
+        # In the single-qubit case, Pauli error and the depolarization fidelity
+        # differ by a factor of 4/3.
```
dstrain115, line 325: Isn't this exactly the factor that average error and pauli error differ by?  Should we use average error instead?

I think you might want to use that rather than introduce a constant factor all over.

95-martin-orion: Using the recent metrics as an example, we have:
```
# 1-qubit metrics for 5_2
avg. error:   0.0014671
pauli error:  0.0022007
```
It doesn't look like using average error would be equivalent to `pauli_err * 4 / 3` here.

dstrain115: You are correct.  I took a look again at the supremacy supplement and depolarization error and average error are separately calculated.

PR #2928, opened 2020-04-21
Add ylabel argument to heatmap colorbar.

mrwojtek:


cirq/vis/heatmap.py
```diff
@@ -274,6 +277,8 @@ def _plot_colorbar(self, mappable: mpl.cm.ScalarMappable,
                                       orientation=orien,
                                       **self.colorbar_options)
         colorbar_ax.tick_params(axis='y', direction='out')
+        if self.colorbar_ylabel:
+            colorbar_ax.set_ylabel(self.colorbar_ylabel)
```
viathor, line 281: Test coverage failed for this line, but it should be easy to fix that: just set the label in `test_colorbar`.

mrwojtek:
Thanks Ping, that solves the problem!

PR #2929, opened 2020-04-21
Fix indentation on markdown

dstrain115:
- Too many spaces causes sphinx to blockquote the links instead of using an unordered list.


mpharrigan:
Are the changes to absolute URLs necessary? This isn't ideal for e.g. locally building the docs

dstrain115:
> Are the changes to absolute URLs necessary? This isn't ideal for e.g. locally building the docs

I couldn't get it to build otherwise.  The plugin for markdown for sphinx doesn't seem to want to play nice with nbsphinx.  It errors out that it can't find the target.  Maybe there is a way to get it to work, but I couldn't figure it out, and I didn't want to change the links to XYZ.html or XYZ.md, since those would also be broken locally.

mpharrigan:
ah that's unfortunate

PR #2930, opened 2020-04-21
Fix pytest for mac by replacing grep with perl

tanujkhattar:
`grep -P` does not work on mac anymore. Replace it with `perl -nle` instead.


vtomole:
`mac anymore`

Since when? What version of Mac? This is not the only place we call `grep`.

tanujkhattar:
This is the only place we call `grep -P` i.e. use perl regex with grep. 

It looks it's [been quite a while](https://stackoverflow.com/questions/16658333/grep-p-no-longer-works-how-can-i-rewrite-my-searches), so I guess for this project it might have never worked on a mac by default, unless people have a GNU grep version installed.  I can dig in more to find relevant information on the difference if that's necessary.

vtomole:
@tanujkhattar 

The macOS that we are running on Github Workflow doesn't seem to be running into this problem. What version of mac gives this error? The version we are running must be different.

tanujkhattar:
`$> grep --version`
`grep (BSD grep) 2.5.1-FreeBSD`

PR #2933, opened 2020-04-22
Add seed to DFE code and test

tonybruguier-google:
Adding a seed to DFE code and unit tests


examples/direct_fidelity_estimation.py
```diff
@@ -330,9 +332,12 @@ def direct_fidelity_estimation(circuit: cirq.Circuit, qubits: List[cirq.Qid],
             simulate noise in the circuit. If greater than 0, we instead use the
             'sampler' parameter directly to estimate the characteristic
             function.
+        seed: The random seed to use for this DFE computation.
     Returns:
         The estimated fidelity and a log of the run.
     """
+    np.random.seed(seed)
```
viathor, line 339: I'm not convinced we should do this since this alters the globally visible state of numpy's PRNG. Any random numbers generated after a call to `direct_fidelity_estimation` would be affected. Such behavior can lead to subtle bugs.

I think it is better to leave it up to the user to set the seed before they call `direct_fidelity_estimation`.

tonybruguier-google: It was suggested, I think, by @mpharrigan in the review of:
https://github.com/quantumlib/Cirq/pull/2918
But I am happy to withdraw, of course.

It is also quite possible that I didn't understand what Matt was suggesting, so I will wait for a comment. No hurry.

tonybruguier-google:
Waiting for more info. Not urgent.

tonybruguier-google:
Withdrawing this PR.

examples/direct_fidelity_estimation.py
mpharrigan: hunt around for how we do it elsewhere in the code base. Re-seeding the global random state is not ok

PR #2934, opened 2020-04-23
Allow creating device proto for specified qubits and pairs

maffoo:
Currently `cirq.google.devices.known_devices.create_device_proto_from_diagram` allows creating a `DeviceDescription` proto message from an ascii-art depiction of a grid of qubits; the function assumes that all adjacent qubits in the grid are coupled. We refactor this to introduce a new function `create_device_proto_for_qubits` which allows directly specifying the qubits and coupled pairs in a device. This could be useful for other device connectivity, for example irregular grids where some adjacent are not coupled.


cirq/google/devices/known_devices.py
```diff
@@ -61,55 +65,77 @@ def create_device_proto_from_diagram(
...
+    """Parse ASCIIart device layout into DeviceSpecification proto.
 
-    Args:
-            ascii_grid: An ASCII version of the grid, see _parse_device for
-                    details
-            gate_set: A serializable gate_set object that is used to define
-                    the translation between gate ids and cirq Gate objects
-            durations_picos: A dictionary with keys of gate ids and values of
-                    the duration of that gate in picoseconds
+    This function also assumes that all pairs of adjacent qubits are valid
```
viathor, line 72: nit s/also//

(no other assumption is given in preceding text, so it's unclear what "also" refers to)

cirq/google/devices/known_devices.py
```diff
@@ -61,55 +65,77 @@ def create_device_proto_from_diagram(
...
+    return create_device_proto_for_qubits(qubits, pairs, gate_sets,
+                                          durations_picos, out)
+
+
+def create_device_proto_for_qubits(
+        qubits: Collection['cirq.Qid'],
+        pairs: Collection[Tuple['cirq.Qid', 'cirq.Qid']],
+        gate_sets: Optional[Iterable[
+            serializable_gate_set.SerializableGateSet]] = None,
+        durations_picos: Dict[str, int] = None,
```
viathor, line 101: `Optional[Dict[...]]`

cirq/google/devices/known_devices.py
```diff
@@ -61,55 +65,77 @@ def create_device_proto_from_diagram(
         gate_sets: Optional[Iterable[
             serializable_gate_set.SerializableGateSet]] = None,
         durations_picos: Dict[str, int] = None,
```
viathor, line 67: `Optional[Dict[...]]`

cirq/google/devices/known_devices.py
```diff
@@ -61,55 +65,77 @@ def create_device_proto_from_diagram(
...
-
     qubits, _ = _parse_device(ascii_grid)
-    spec = device_pb2.DeviceSpecification()
 
-    # Create valid qubit list
     qubit_set = frozenset(qubits)
-    spec.valid_qubits.extend([v2.qubit_to_proto_id(q) for q in qubits])
+    pairs: List[Tuple['cirq.Qid', 'cirq.Qid']] = []
+    for qubit in qubits:
+        for offset in [(0, 1), (1, 0)]:
```
viathor, line 87: This duplicates a bit of logic already implemented in the `neighbors` methods of qubit classes. Could you use the method instead?

cirq/google/devices/known_devices.py
maffoo: Fixed.

cirq/google/devices/known_devices.py
maffoo: Fixed.

cirq/google/devices/known_devices.py
maffoo: Fixed.

cirq/google/devices/known_devices.py
maffoo: We could, but would have to add extra logic here to avoid adding reversed pairs (e.g. want `(0, 0), (0, 1))` but not also `((0, 1), (0, 0))`). I can do that if you think it's worth i.

maffoo: nvm, went ahead and refactored to use the `neighbors` method.

viathor: The code isn't using the `neighbors` method.

BTW: Another reason to use it is that (AFAIK) qubits may be instances of classes other than `GridQubit` and then adding a 2-tuple of ints won't work.

(Perhaps you forgot to push a commit?)

maffoo: Pushed. Note that this only works on the output of `_parse_device` which explicitly returns `cirq.GridQubit`, so adding tuples works. But in any case using `neighbors` is cleaner; thanks for the suggestion!

cirq/google/devices/known_devices.py
```diff
@@ -60,56 +64,77 @@ def create_device_proto_from_diagram(
...
+    for qubit in qubits:
+        for neighbor in sorted(qubit.neighbors()):
+            if neighbor > qubit and neighbor in qubit_set:
+                pairs.append((qubit, neighbor))
+
+    return create_device_proto_for_qubits(qubits, pairs, gate_sets,
+                                          durations_picos, out)
+
+
+def create_device_proto_for_qubits(
```
viathor, line 95: Optional: Consider adding a unit test. The function is covered by tests for `create_device_proto_from_diagram` so this is optional.

Benefits of adding a dedicated unit test:
 * we won't have to add it if/when `create_device_proto_from_diagram` is removed or changed not to rely on `create_device_proto_for_qubits`,
 * you get to test the very behavior you're interested in - filtering for specified qubits.

cirq/google/devices/known_devices.py
```diff
@@ -60,56 +64,77 @@ def create_device_proto_from_diagram(
...
     meas_targets.name = _MEAS_TARGET_SET
     meas_targets.target_ordering = device_pb2.TargetSet.SUBSET_PERMUTATION
 
     # Set up a target set for 2 qubit gates (all adjacent pairs)
-    grid_targets = spec.valid_targets.add()
+    grid_targets = out.valid_targets.add()
     grid_targets.name = _2_QUBIT_TARGET_SET
     grid_targets.target_ordering = device_pb2.TargetSet.SYMMETRIC
 
     # Create the target set as all adjacent pairs on the grid
```
dstrain115, line 129: Update the comment?

maffoo: Done.

cirq/google/devices/known_devices.py
maffoo: Added a test of an "irregular" grid.

PR #2942, opened 2020-04-28
Rename RANDOM_STATE_LIKE to RANDOM_STATE_OR_SEED_LIKE and expose at top level

Strilanc:


Strilanc:
I'm worried that a user seeing the type name without the "or seed" qualifier will not realize that they are allowed to pass in, for example, the integer 5. Because '5' is really not "like" a prng.

viathor:
I see. WDYT about clipping the "_LIKE" suffix? Given that "random state" is already fairly vague "RANDOM_STATE_OR_SEED" seems to encompass everything we need.

Strilanc:
I think that's a reasonable idea but it's not my preference. I think it fits with other "_LIKE" types in that it accepts a wide variety of things in order to make it easier on the user.

PR #2944, opened 2020-04-28
Generalize `measurement_key` to `measurement_keys` protocol

Strilanc:
- This is necessary for eventually allowing measurements inside of composite operations and other kinds of more general measurements


cirq/ops/gate_operation.py
```diff
@@ -15,7 +15,7 @@
 """Basic types defining qubits, gates, and operations."""
 
 from typing import (Any, Dict, FrozenSet, List, Optional, Sequence, Tuple, Type,
-                    TypeVar, Union, TYPE_CHECKING)
+                    TypeVar, Union, TYPE_CHECKING, Iterable)
```
viathor, line 19: nit: keep alphabetical order

cirq/protocols/measurement_key_protocol.py
```diff
@@ -48,12 +56,21 @@ def _measurement_key_(self) -> str:
         will be stored.
         """
 
+    @document
+    def _measurement_keys_(self) -> Iterable[str]:
+        """Return the key that will be used to identify this measurement.
```
viathor, line 56: nit: s/key/keys/

cirq/protocols/measurement_key_protocol.py
```diff
@@ -67,28 +84,62 @@ def measurement_key(val: Any, default: Any = RaiseTypeErrorIfNotProvided):
...
-    """Returns whether or not the given value is a measurement.
+    """Determines whether or not the given value is a measurement.
 
-    A measurement must implement the `measurement_key` protocol and have a
-    channel, as represented by the `has_channel` protocol returning true.
+    Measurements are identified by the fact that `cirq.measurement_keys` returns
+    a non-empty result for them.
     """
-    return measurement_key(val, None) is not None and has_channel(val)
+    return bool(measurement_keys(val))
```
viathor, line 140: Before we would consider objects that have a quantum channel to be measurements, probably to reflect the fact that non-trace preserving quantum operations are measurements.

I don't have the full story here, but I'm concerned we may be breaking some code (in cirq or elsewhere) that depends on this.

cirq/protocols/measurement_key_protocol.py
```diff
@@ -67,28 +84,62 @@ def measurement_key(val: Any, default: Any = RaiseTypeErrorIfNotProvided):
...
-                type(val)))
+    raise TypeError(f"object of type '{type(val)}' had no measurement keys.")
+
+
+def measurement_keys(val: Any, *,
+                     allow_decompose: bool = True) -> Tuple[str, ...]:
+    """Gets the measurement keys of measurements within the given value.
+
+    Args:
+        val: The value which has the measurement key..
```
viathor, line 104: nit: one dot

cirq/protocols/measurement_key_protocol.py
```diff
@@ -67,28 +84,62 @@ def measurement_key(val: Any, default: Any = RaiseTypeErrorIfNotProvided):
...
+            not set, composite operations will appear to have no measurement
+            keys.
+
+    Returns:
+        The measurement keys of the value. If the value has no measurement,
+        the result is the empty tuple.
+    """
+    getter = getattr(val, '_measurement_keys_', None)
+    result = NotImplemented if getter is None else getter()
+    if result is not NotImplemented and result is not None:
```
viathor, line 117: nit:

```python
if result not in (NotImplemented, None):
    ...
```

cirq/protocols/measurement_key_protocol.py
```diff
@@ -48,12 +56,21 @@ def _measurement_key_(self) -> str:
...
         """
 
+    @document
+    def _measurement_keys_(self) -> Iterable[str]:
+        """Return the key that will be used to identify this measurement.
+
+        When a measurement occurs, either on hardware, or in a simulation,
+        this is the key value under which the results of the measurement
+        will be stored.
+        """
```
viathor, line 61: One of the docstrings in this class should say whether it is ok to define both `_measurement_key_` and `_measurement_keys_` and what happens when one does this.

cirq/protocols/measurement_key_protocol.py
```diff
@@ -13,12 +13,20 @@
...
 
-from typing import Any
+from typing import Any, Iterable, Union, TYPE_CHECKING, Tuple, Dict, List, \
+    Optional, DefaultDict
 
 from typing_extensions import Protocol
 
 from cirq._doc import document
-from cirq.protocols import has_channel
+from cirq.protocols.channel import has_channel
```
viathor: I think this is unused (but see other comment).

cirq/protocols/measurement_key_protocol.py
```diff
@@ -67,28 +84,62 @@ def measurement_key(val: Any, default: Any = RaiseTypeErrorIfNotProvided):
...
+                         f'from {val!r}.')
 
     if default is not RaiseTypeErrorIfNotProvided:
         return default
 
-    if getter is None:
-        raise TypeError(
-            "object of type '{}' has no _measurement_key_ method.".format(
-                type(val)))
+    raise TypeError(f"object of type '{type(val)}' had no measurement keys.")
```
viathor, line 96: nit: Capitalize "Object" or consistency with exception on l.95 above.

cirq/protocols/measurement_key_protocol.py
```diff
@@ -67,28 +84,62 @@ def measurement_key(val: Any, default: Any = RaiseTypeErrorIfNotProvided):
...
+        the result is the empty tuple.
+    """
+    getter = getattr(val, '_measurement_keys_', None)
+    result = NotImplemented if getter is None else getter()
+    if result is not NotImplemented and result is not None:
+        return tuple(result)
+
+    getter = getattr(val, '_measurement_key_', None)
+    result = NotImplemented if getter is None else getter()
+    if result is not NotImplemented and result is not None:
```
viathor, line 122: ditto

cirq/protocols/measurement_key_protocol.py
Strilanc: The fallout of removing this check is that a bad operation that said it was a measurement but had no specified channel would fail slightly later. This is not really a useful check IMO. In particular, note that all the tests continued to pass.

Further note that some callers require the measurement to be something more specific than a channel (e.g. a stabilizer operation, an operation that can apply to a state vector, etc). Relying on this method to check that the measurement can act on your state in the way you need is coupling two things together that can be kept separate.

cirq/ops/gate_operation.py
Strilanc: Done. I think we should either write a tool to do this automatically or stop caring about it.

cirq/protocols/measurement_key_protocol.py
Strilanc: Done.

cirq/protocols/measurement_key_protocol.py
Strilanc: Done.

cirq/protocols/measurement_key_protocol.py
Strilanc: Acknowledged. That constructs a tuple as part of the check. More expensive.

cirq/protocols/measurement_key_protocol.py
Strilanc: Done.

cirq/protocols/measurement_key_protocol.py
Strilanc: Fixed.

cirq/protocols/measurement_key_protocol.py
Strilanc: Done.

cirq/protocols/measurement_key_protocol.py
Strilanc: Acknowledged.

PR #2946, opened 2020-04-29
Add `can_act_on_state_vector` protocol

Strilanc:
- Introduce `cirq.can_act_on_state_vector`
- Fix several places where `GateOperation` could delegate a protocol directly to its gate but wasn't (resulting in unnecessary work and possibly additional calls to decompose).
- Add `allow_decompose` parameter to `has_unitary` and `has_mixture`
- Implement `_can_act_on_state_vector_` for `MeasurementGate` and `ResetChannel`
- Fix `cirq.has_mixture_channel` not doing a decompose check

Note:
- Will be followed up by `cirq.act_on_state_vector`, a more general version of `apply_unitary`
- Will be followed up by `cirq.SparseSimulator` delegating `[can_]act_on_state_vector` for almost everything
- Comes after https://github.com/quantumlib/Cirq/pull/2944


Strilanc:
We discussed this at the weekly meeting and a lot of opinions about how to do this properly came up. In particular, should it be more generic (act on simulator state type T), is this just redundant with has_channel, etc.

Strilanc:
Obsoleted

PR #2950, opened 2020-04-30
Accept empty list of whitelisted_users when updating reservations.

wcourtney:
Check that whitelisted_users is actually None to accept updates where input is an empty list (i.e. clearing all users).


cirq/google/engine/engine_client.py
```diff
@@ -766,7 +766,7 @@ def update_reservation(self,
         if end:
             reservation.end_time.seconds = int(end.timestamp())
             paths.append('end_time')
-        if whitelisted_users:
+        if whitelisted_users != None:
```
dstrain115, line 769: Should we add a short comment in the doc string to denote the difference between [] and None?

wcourtney: Done.

PR #2954, opened 2020-05-01
Add get_device for engine_processor

dstrain115:
- Adds get_device() that returns a cirq.Device out of the
device specification.

- Also, cleaned up some documentation that was not using the
engine processor correctly.

### DIFF
--- a/cirq/google/engine/engine_processor.py
+++ b/cirq/google/engine/engine_processor.py
@@ -113,6 +113,12 @@ class EngineProcessor:
     def get_device(
             self,
             gate_sets: Iterable[serializable_gate_set.SerializableGateSet]):
+        """Returns a `Device` created from the processor's device specification.
+
+        This method queries the processor to retrieve the device specification,
+        which is then use to create a `SerializableDevice` that will validate
+        that operations are supported and use the correct qubits.
+        """
         spec = self.get_device_specification()
         if not spec:
             raise ValueError('Processor does not have a device specification')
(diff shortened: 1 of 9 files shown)


cirq/google/engine/engine_processor.py
```diff
@@ -108,6 +110,15 @@ def get_device_specification(
         else:
             return None
 
+    def get_device(
```
wcourtney, line 113: Please add a doc string for this method.

cirq/google/engine/engine_processor.py
```diff
@@ -108,6 +110,21 @@ def get_device_specification(
...
+            self,
+            gate_sets: Iterable[serializable_gate_set.SerializableGateSet]):
```
dabacon, line 115: Just a side comment: This probably really shows that the gate set should also be part of the device spec.

dabacon, line 115: add return type

dstrain115: Perhaps there is a better way of doing it, but the gate set is part of the device spec.  You just need a way to turn 'fsim' into 'cirq.FSimGate'.  There needs to be a place that can translate the gate ids into cirq objects.

cirq/google/engine/engine_processor.py
dstrain115: Done.

cirq/google/engine/engine_processor.py
dstrain115: Done.

PR #2955, opened 2020-05-02
Decompose cphase to two fSim gates

viathor:
This implements the decomposition described in section VII F I of the [supplement](https://arxiv.org/abs/1910.11333) to the quantum supremacy paper.

### DIFF
--- a/cirq/optimizers/cphase_to_fsim.py
+++ b/cirq/optimizers/cphase_to_fsim.py
@@ -2,7 +2,7 @@ from typing import Optional, Sequence, Tuple, TYPE_CHECKING
 
 import numpy as np
 
-from cirq import ops, circuits, devices, protocols
+from cirq import circuits, devices, ops, protocols
 
 if TYPE_CHECKING:
     import cirq
--- a/cirq/optimizers/cphase_to_fsim.py
+++ b/cirq/optimizers/cphase_to_fsim.py
@@ -1,0 +1,224 @@
+from typing import Optional, Sequence, Tuple, TYPE_CHECKING
+
+import numpy as np
+
+from cirq import ops, circuits, devices, protocols
+from cirq.optimizers.merge_single_qubit_gates import MergeSingleQubitGates
+from cirq.optimizers.drop_empty_moments import DropEmptyMoments
+
+if TYPE_CHECKING:
+    import cirq
+
+
+def _asinsin(x: float) -> float:
+    """Computes arcsin(sin(x)) for any x. Return value in [-π/2, π/2]."""
+    k = round(x / np.pi)
+    if k % 2 == 0:
+        return x - k * np.pi
+    return k * np.pi - x
+
+
+def compute_cphase_exponents_for_fsim_decomposition(
+        fsim_gate: 'cirq.FSimGate') -> Sequence[Tuple[float, float]]:
+    """Returns intervals of CZPowGate exponents valid for FSim decomposition.
+
+    Ideal intervals associated with the constraints are closed, but due to
+    numerical error the caller should not assume the endpoints themselves
+    are valid for the decomposition. See `decompose_cphase_into_two_fsim`
+    for details on how FSimGate parameters constrain the phase angle of
+    CZPowGate.
+
+    Args:
+        fsim_gate: FSimGate into which CZPowGate would be decomposed.
+
+    Returns:
+        Sequence of 2-tuples each consisting of the minimum and maximum
+        value of the exponent for which CZPowGate can be decomposed into
+        two FSimGates. The intervals are cropped to [0, 2]. The function
+        returns zero, one or two intervals.
+    """
+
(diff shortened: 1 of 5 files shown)


cirq/optimizers/cphase_to_fsim.py
```diff
@@ -0,0 +1,224 @@
...
+        ops.rx(xi).on(q0),
+
+        # Local Z rotations to convert exp(-i Z⊗Z δ/4) into desired CPhase.
+        ops.rz(-delta / 2).on(q0),
+        ops.rz(-delta / 2).on(q1),
+        ops.GlobalPhaseOperation(np.exp(-1j * delta / 4)),
+    )
+
+    MergeSingleQubitGates().optimize_circuit(circuit)
+    DropEmptyMoments().optimize_circuit(circuit)
```
mrwojtek, line 223: Could you skip those, and let user decide how to optimize circuit?

viathor: I did this originally for consistency with [another optimizer](https://github.com/quantumlib/Cirq/blob/master/cirq/optimizers/two_qubit_to_fsim.py), but on second thought I agree we should leave this up to the user.

(I also considered adding a boolean flag, but calling an additional optimizer is arguably so easy that this just pollutes the signature for little benefit.)

I'll fix the other optimizer for consistency in follow-up PRs (they are out-of-scope here).

viathor:
PTAL

cirq/optimizers/cphase_to_fsim.py
```diff
@@ -0,0 +1,218 @@
...
+    #
+    return circuits.Circuit(
+        # Local X rotations to convert Γ1⊗I − iZ⊗Γ2 into exp(-i Z⊗Z δ/4)
+        ops.rx(xi).on(q0),
+        ops.rx(eta).on(q1),
+
+        # Y(θ, φ) := exp(-i X⊗X θ/2) exp(-i Y⊗Y θ/2) exp(-i Z⊗Z φ/4)
+        fsim_gate.on(q0, q1),
+        ops.rz(phi / 2).on(q0),
+        ops.rz(phi / 2).on(q1),
```
maffoo, line 196: Consider using `PhasedXZ` instead of separate `Rx` and `Rz` gates.

cirq/optimizers/cphase_to_fsim.py
```diff
@@ -0,0 +1,218 @@
...
+        fsim_gate: FSimGate into which CZPowGate would be decomposed.
+
+    Returns:
+        Sequence of 2-tuples each consisting of the minimum and maximum
+        value of the exponent for which CZPowGate can be decomposed into
+        two FSimGates. The intervals are cropped to [0, 2]. The function
+        returns zero, one or two intervals.
+    """
+
+    def filter_out_empty_intervals(intervals: Sequence[Tuple[float, float]]
```
maffoo, line 39: nit: call this "nonempty_intervals"

cirq/optimizers/cphase_to_fsim.py
```diff
@@ -0,0 +1,218 @@
...
+    max_exponent = max(max_exponent_1, max_exponent_2)
+    return filter_out_empty_intervals([(min_exponent, max_exponent)])
+
+
+def decompose_cphase_into_two_fsim(
+        cphase_gate: 'cirq.CZPowGate',
+        *,
+        fsim_gate: 'cirq.FSimGate',
+        qubits: Optional[Sequence['cirq.Qid']] = None,
+        atol: float = 1e-8) -> 'cirq.Circuit':
```
maffoo, line 76: This will probably be easier to use if it returned an OP_TREE rather than a circuit.

viathor: Indeed. Done.

cirq/optimizers/cphase_to_fsim.py
viathor: Done.

cirq/optimizers/cphase_to_fsim.py
viathor: I like the transparency and readability of the operations here and I'm concerned that we would loose it (as a result of some gates straddling the conceptually distinct steps in the decomposition).

Perhaps we can leave such optimization up to the caller (and another optimizer) as we did with merging single-qubit gates and dropping empty moments in Wojtek's comment earlier?

viathor:
PTAL

cirq/optimizers/cphase_to_fsim.py
mrwojtek: I agree with Adam here.

cirq/optimizers/cphase_to_fsim.py
viathor: I promised a PR, but filed #3050 instead in case anyone disagrees or relies on current behavior.

PR #2960, opened 2020-05-04
Fix Problem with Y gates in Clifford Simulator

alexisshaw:
I found a bug in the Y gate for the tableau circuit simulator. Here is a fix and a test

### DIFF
--- a/cirq/sim/clifford/clifford_simulator_test.py
+++ b/cirq/sim/clifford/clifford_simulator_test.py
@@ -291,6 +291,19 @@ def test_clifford_stabilizerStateChForm_repr():
     assert repr(state) == 'StabilizerStateChForm(num_qubits=2)'
 
 
+def test_clifforf_circuit_SHSYSHS():
+    q0 = cirq.LineQubit(0)
+    circuit = cirq.Circuit(cirq.S(q0), cirq.H(q0), cirq.S(q0), cirq.Y(q0),
+                           cirq.S(q0), cirq.H(q0), cirq.S(q0), cirq.measure(q0))
+
+    clifford_simulator = cirq.CliffordSimulator()
+    wave_function_simulator = cirq.Simulator()
+
+    np.testing.assert_almost_equal(
+        clifford_simulator.simulate(circuit).final_state.wave_function(),
+        wave_function_simulator.simulate(circuit).final_state)
+
+
 def test_clifford_circuit():
     (q0, q1) = (cirq.LineQubit(0), cirq.LineQubit(1))
     circuit = cirq.Circuit()
--- a/cirq/sim/clifford/clifford_tableau.py
+++ b/cirq/sim/clifford/clifford_tableau.py
@@ -140,7 +140,7 @@ class CliffordTableau():
         self.rs[:] ^= self.zs[:, q]
 
     def _Y(self, q):
-        self.rs[:] ^= self.xs[:, q] | self.zs[:, q]
+        self.rs[:] ^= self.xs[:, q] ^ self.zs[:, q]
 
     def _Z(self, q):
         self.rs[:] ^= self.xs[:, q]


cirq/sim/clifford/clifford_simulator_test.py
```diff
@@ -291,6 +291,19 @@ def test_clifford_stabilizerStateChForm_repr():
     assert repr(state) == 'StabilizerStateChForm(num_qubits=2)'
 
 
+def test_clifforf_circuit_SHSYSHS():
```
viathor, line 294: I suppose a simple test would check whether XZ and Y yield the same state up to global phase, but this should work just as well.

cirq/sim/clifford/clifford_simulator_test.py
```diff
@@ -475,4 +488,4 @@ def test_json_roundtrip():
 
     # And the CH form isn't changed either.
     assert np.allclose(state.ch_form.wave_function(),
-                       state_roundtrip.ch_form.wave_function())
+                       state_roundtrip.ch_form.wave_function())
```
vtomole, line 491: Add newline

cirq/sim/clifford/clifford_simulator_test.py
```diff
@@ -291,6 +291,19 @@ def test_clifford_stabilizerStateChForm_repr():
...
+def test_clifforf_circuit_SHSYSHS():
+    q0 = cirq.LineQubit(0)
+    circuit = cirq.Circuit(cirq.S(q0), cirq.H(q0), cirq.S(q0), cirq.Y(q0),
+                           cirq.S(q0), cirq.H(q0), cirq.S(q0), cirq.measure(q0))
+
+    clifford_simulator = cirq.CliffordSimulator()
+    wave_function_simulator = cirq.Simulator()
+
+    np.testing.assert_almost_equal(
+        clifford_simulator.simulate(circuit).final_state.wave_function(),
```
vtomole, line 303: @viathor  Not related to this PR: What do you think about the fact that the clifford simulator needs a `.wave_function()` appended while the `wave_function_simulator` does not from a usability perpective? Should we make these consistent?

viathor: Such inconsistency is definitely a problem in the API. Would you like to file an issue about this?

Thanks for catching.

alexisshaw: Why should there be an exponential cost in this otherwise polynomial simulator?
Generating the state vector is exponential in the number of qubits, but everything else is quadratic in the number of qubits? You would need to work around this.

cirq/sim/clifford/clifford_simulator_test.py
alexisshaw: I was looking for something that could be exposed as a measurement. The error in the Y gate can only be exposed when there is a Y stabilize, and this is not the case in the proposed test.

cirq/sim/clifford/clifford_simulator_test.py
alexisshaw: Fixed

alexisshaw:
Was trying to consolidate into a single commit.

cirq/sim/clifford/clifford_simulator_test.py
viathor: Fair enough. Thank you for adding the test!

cirq/sim/clifford/clifford_simulator_test.py
viathor: I agree exponential cost computation should not be executed implicitly in a surprising context. This does not conflict with API consistency. For example, some simulators already require that a user calls `wave_function` *explicitly* if they wish to see the reconstructed wavefunction.

In any case, none of this blocks this PR.

alexisshaw:
As there seems to be a problem with the automatic testing, would you mind if I close the PR and then re-send it?

PR #2966, opened 2020-05-04
Enable serialization of non-GridQubits.

### LINKED ISSUE #1325: Qubits other than GridQubit don't work on Quantum Engine Simulator
I tried to run the following code through Quantum Engine:
```
import random

circuit = cirq.Circuit.from_ops(cirq.X(a))
engine = cirq.google.Engine(api_key=api_key,
                            api='autopush-quantum.sandbox', 
                            **colab_auth)
name = 'example-%s' % ''.join(random.choice(
    string.ascii_uppercase + string.digits) for _ in range(6))
job = engine.run_sweep(
    job_config=cirq.google.JobConfig(project_id=project_id, program_id=name),
    program=circuit,
    repetitions=1000)
```
and got
```
AttributeError                            Traceback (most recent call last)
<ipython-input-18-cef5b731c112> in <module>()
     11     job_config=cirq.google.JobConfig(project_id=project_id, program_id=name),
     12     program=circuit,
---> 13     repetitions=1000)

/usr/local/lib/python3.6/dist-packages/cirq/google/engine/engine.py in run_sweep(self, program, job_config, params, repetitions, priority, target_route)
    356             sweep in sweeps]
    357         program_dict['operations'] = [op for op in
--> 358                                       schedule_to_proto_dicts(schedule)]
    359         code = {
    360             '@type': 'type.googleapis.com/cirq.api.google.v1.Program'}

/usr/local/lib/python3.6/dist-packages/cirq/google/engine/engine.py in <listcomp>(.0)
    355             sweep_to_proto_dict(sweep, repetitions) for
    356             sweep in sweeps]
--> 357         program_dict['operations'] = [op for op in
    358                                       schedule_to_proto_dicts(schedule)]
    359         code = {

/usr/local/lib/python3.6/dist-packages/cirq/google/programs.py in schedule_to_proto_dicts(schedule)
    154         op = gate_to_proto_dict(
    155             cast(ops.GateOperation, so.operation).gate,
--> 156             so.operation.qubits)
    157         time_picos = so.time.raw_picos()
    158         if last_time_picos is None:

/usr/local/lib/python3.6/dist-packages/cirq/google/programs.py in gate_to_proto_dict(gate, qubits)
     38             # coverage: ignore
     39             raise ValueError('Wrong number of qubits.')
---> 40         return _x_to_proto_dict(gate, qubits[0])
     41 
     42     if isinstance(gate, ops.YPowGate):

/usr/local/lib/python3.6/dist-packages/cirq/google/programs.py in _x_to_proto_dict(gate, q)
     69 def _x_to_proto_dict(gate: ops.XPowGate, q: ops.QubitId) -> Dict:
     70     exp_w = {
---> 71         'target': cast(devices.GridQubit, q).to_proto_dict(),
     72         'axis_half_turns':
     73             _parameterized_value_to_proto_dict(0),

AttributeError: 'LineQubit' object has no attribute 'to_proto_dict'
```
Maybe it makes sense to force the user to convert all their qubits to GridQubit, I don't know. I feel like my example should run, at least on the simulator.

`google/programs.py` has lines like
```
def _x_to_proto_dict(gate: ops.XPowGate, q: ops.QubitId) -> Dict:
    exp_w = {
        'target': cast(devices.GridQubit, q).to_proto_dict(),
```
where we cast `q` to a GridQubit. Shouldn't we instead just change the type annotation from QubitId to GridQubit?

95-martin-orion:
Fixes #1325.

Most of the work was already done here - the only remaining effort was to remove checks and add a mux for the deserialize methods. This does have a noticeable side-effect: this code now accepts **any** proto ID as a valid qubit, since NamedQubit has no restrictions on the strings allowed as names.

As an aside, it's a little odd that we use `x_y` notation for GridQubits, when `int('1_2')` returns `12`. This makes parsing strings as different qubit types order-dependent.


dabacon:
Does the simulator accept arbitrary strings right now?

95-martin-orion:
> Does the simulator accept arbitrary strings right now?

Yes - it populates a map of qubits to string IDs with no parsing/validation on the IDs. This will probably misbehave if you mix e.g. `LineQubit(1)` with `NamedQubit('1')`, but that's ill-advised for plenty of other reasons.

cirq/google/api/v2/program_test.py
```diff
@@ -72,3 +72,15 @@ def test_line_qubit_from_proto_id_invalid():
...
     assert v2.named_qubit_from_proto_id('a') == cirq.NamedQubit('a')
+
+
+def test_generic_qubit_from_proto_id():
+    assert v2.qubit_from_proto_id('1_2') == cirq.GridQubit(1, 2)
+    assert v2.qubit_from_proto_id('1') == cirq.LineQubit(1)
+    assert v2.qubit_from_proto_id('a') == cirq.NamedQubit('a')
+
+    # int(1_2_3) = 123
+    assert v2.qubit_from_proto_id('1_2_3') == cirq.LineQubit(123)
```
dabacon, line 83: This feels strange because it is ambiguous with grid qubit definition and how python treats string ints with underscores.

My suggestion would be to write this code as 
1) split on '_',
2) If two parts, try int cast and grid qubit
3) If one part, attempt to cast to int and a line qubit.
4) make all other cases named.
This case would then be a named qubit.

95-martin-orion: Done.

Issue #1642, opened 2019-05-15
Add links to issues next to TODO comments


vtomole:
There're currently ~20 TODOs in the repository.
 For example:
https://github.com/quantumlib/Cirq/blob/7fcda9361a3fb2b58ef914a0fce7573b2a83c6f9/cirq/sim/sparse_simulator.py#L295

We should link to the corresponding issues alongside these TODO comments or create the issues and then link them if they don't exist.

Hancockan:
Assign this to me

cjcarey:
Assign this to me

vtomole:
@cjcarey Creating 20 issues at once might be too much. Open 5 for now. Link to this issue when creating.

cjcarey:
@vtomole I created 5 new issues and made a new pull request for the spelling error. When should I have the rest of the TODO statements completed

vtomole:
@cjcarey Creating that many issues at once was not as spammy as i thought it would be. How about we just finish it all right now?

cjcarey:
I think I got everything.. let me know if I missed something!

vtomole:
@cjcarey That's it. Thanks for doing this.

mpharrigan:
A lot of the issue titles don't make sense out of context

vtomole:
Hey @cjcarey, please add more context to the issues you created. It's a good exercise in learning the code base.

dabacon:
We need to reaudit these todo issues.

dabacon:
I went over all TODO and added issues for all of them.  I also edited the issues to make sure they were clear on the issue.  Closing.

PR #2971, opened 2020-05-04
Cleanup TODOs and corresponding Github issues

dabacon:
This includes making all TODOs refer to github issues.  I've also audited all of these github issues to make sure they include enough information about the issues.


dabacon:
This will fix #1642

PR #2974, opened 2020-05-05
Update protobuf version

dabacon:
Fixes #2239


dabacon:
PTAL

Bumped up to latests stable which is 3.11.

PR #2976, opened 2020-05-05
Update asteroid version

dabacon:
Previously we were blocked on upgrading asteroid due to a bug
https://github.com/PyCQA/astroid/issues/650
However the bug now says this is fixed, so verifying that our tests pass.

...and it turns out they don't pass.

Fixes:
* `ndarray.reshape` technically takes lists of arguments like `a.reshape(1, 3, 4)`.  However sometimes lint gets mad about this.  Fixed this and changed other uses in this test to use same pattern for consistency.
* Found a missing assertion right nearby and fixed this.  Needed to fix order bug in test.
* Didn't fix a lint error about the result of getattr not being callable.  I don't think there is any fix besides this as there are lots of bugs about pylint messing this up.  So added a disable.

### DIFF
--- a/cirq/linalg/transformations_test.py
+++ b/cirq/linalg/transformations_test.py
@@ -442,7 +442,7 @@ def test_subwavefunction():
 def test_subwavefunction_bad_subset():
     a = cirq.testing.random_superposition(4)
     b = cirq.testing.random_superposition(8)
-    state = np.kron(a, b).reshape(2, 2, 2, 2, 2)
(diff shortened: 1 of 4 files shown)


dabacon:
Thanks!  Turns out that this ended up exposing a bunch of corner case and even *gasp* a test that had a missing assert that when inserted made the test fail.

dabacon:
PTAL.

Updated first comment above to reflect what changes were required to get this to pass.

PR #2932, opened 2020-04-22
Adding a DFE Jupyter notebook

tonybruguier-google:
As requested, I added a Jupyter notebook that showcases the DFE code.

Note that the present version relies on under-review PR:
https://github.com/quantumlib/Cirq/pull/2918

I've never gone through a review for a notebook, so I am attaching a PDF of what it currently looks like.
[direct_fidelity_estimation.pdf](https://github.com/quantumlib/Cirq/files/4517980/direct_fidelity_estimation.pdf)

PR #2977, opened 2020-05-05
std dev on fidelity

tonybruguier-google:
Adding some bounds on DFE


examples/direct_fidelity_estimation.py
```diff
@@ -276,6 +277,24 @@ def _estimate_pauli_traces_general(qubits: List[cirq.Qid],
...
 
+def _estimate_std_devs_clifford(fidelity: float, N: int):
+    """
+    Estimates the standard deviation of the measurement for Clifford circuits.
+
+    Args:
+        fidelity: The measured fidelity
+        N: the number of measurements
+    Returns:
+        The standard deviation (estimated and a bound)
```
dabacon, line 288: Could you add a note on what the bound means (i.e. it is statistical bound for within one standard deviation?).

examples/direct_fidelity_estimation.py
```diff
@@ -276,6 +277,24 @@ def _estimate_pauli_traces_general(qubits: List[cirq.Qid],
     return pauli_traces
 
 
+def _estimate_std_devs_clifford(fidelity: float, N: int):
```
dabacon, line 280: add return type of Tuple[float, float].  This will help as the doc string does not make it immediately obvious that this is a Tuple returned.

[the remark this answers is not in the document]
tonybruguier-google: Done

examples/direct_fidelity_estimation.py
```diff
@@ -276,6 +277,24 @@ def _estimate_pauli_traces_general(qubits: List[cirq.Qid],
...
+    """
+    Estimates the standard deviation of the measurement for Clifford circuits.
+
+    Args:
+        fidelity: The measured fidelity
+        N: the number of measurements
+    Returns:
```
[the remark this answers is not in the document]
tonybruguier-google: Done.

examples/direct_fidelity_estimation.py
tonybruguier-google: Like this?

examples/direct_fidelity_estimation.py
tonybruguier-google: Done.

tonybruguier-google:
Thanks. PTAL.

mpharrigan:
Can you motivate this? What is this? Is there an equation in the papers you can reference? Why do you have to "clip" the fidelity to 0 and 1, and is this ok to do

tonybruguier-google:
@mpharrigan 

I had this little write up with equations:
https://drive.google.com/file/d/1cuRtxOaLBprEukeTRt5PuPJOhf_AlHGn/view?usp=sharing

I don't think they are in the papers referenced (possibly because it is too obvious).

mpharrigan:
That's a helpful write-up. I'd definitely try to (very) briefly summarize it in the docstring and mention you are using the Bhatia Davis inequality

tonybruguier-google:
Added a comment mentioning the Bhatia Davis inequality.

I also took the liberty of modifying my Jupyter notebook in this PR:
https://github.com/quantumlib/Cirq/pull/2932

You can see the preview there:
https://github.com/quantumlib/Cirq/blob/2963e9f04a23b1c358913b2935d78ddc28e1e3db/examples/direct_fidelity_estimation.ipynb

examples/direct_fidelity_estimation.py
```diff
@@ -276,6 +277,35 @@ def _estimate_pauli_traces_general(qubits: List[cirq.Qid],
...
+
+    # We use the Bhatia Davis inequality to estimate the variance of the
+    # fidelity. This gives that:
+    # Var[\hat{F}] <= (1 - F) . F / N
+    # StdDev[\hat{F}] <= \sqrt{(1 - F) . F / N}
+    #
+    # By further using the fact that 0 <= F <= 1 we get:
+    # StdDev[\hat{F}] <= \frac{1}{2 \sqrt{N}}
+
+    fidelity_bounded = max(0.0, min(1.0, fidelity))
```
mpharrigan, line 301: is clipping the right thing to do here? or should we throw an exception if the input is outside of the bounds it's supposed to be in

tonybruguier-google: I think it's because we use the *estimated* fidelity, which can be outside the range [0, 1], even if we know that the true fidelity cannot be outside of it, right?

If I comment out the range breaking, I can see some unit tests failing like here:
```
   def test_calling_main():
        dfe.main(n_measured_operators=3, samples_per_term=0)
>       dfe.main(n_measured_operators=3, samples_per_term=10)

examples/direct_fidelity_estimation_test.py:180: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
examples/direct_fidelity_estimation.py:498: in main
    estimated_fidelity, _ = direct_fidelity_estimation(
examples/direct_fidelity_estimation.py:448: in direct_fidelity_estimation
    std_dev_estimate, std_dev_bound = _estimate_std_devs_clifford(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

fidelity = -0.06666666666666667, n = 3
```

I looked at the code + math a little bit more, but with some noise, it's not impossible that we flip the sign right? I'm still not sure at all and there could be a bug in the existing code of course. Also, I'm not sure it's a good idea to use the estimated fidelity to estimate its variance. I don't have the math to support it, but it seems like it'd be reasonable if the estimate is good enough.

For these two reasons, I'm not resolving the conversation. Thanks for the very careful review!

tonybruguier-google:
Thanks. PTAL.

examples/direct_fidelity_estimation.py
mpharrigan: 0.06 seems like a pretty big deviation, although for 3 operators and 10 samples per term, you'd also expect a high variance. I can think of three options and I'll defer to you what you think is best:

1. Add an additional `tolerance` parameter with a reasonably small default so an error can be thrown if someone inputs something blatantly wrong (like a fidelity of 23) and tests can expand the tolerance when statistics are known to be bad
2. Do something fancier like throw an exception if the deviation from [0,1] exceeds the computed standard deviation :)
3. Make an argument that since this is a private helper function we know that the input will be a fidelity so the chance of blatant abuse is lessened or would be caught by the rest of the fidelity tests

tonybruguier-google:
I decided to punt the issue and not report any estimate if the fidelity is outside [0, 1]. If the estimate is that off, it might make more sense to just say nothing than say something wrong.

mpharrigan:
maybe dave's review was blocking cirqbot? although now it's not picking this PR up at all

dabacon:
Cancelled automerge of PR#2977 ('std dev on fidelity'): Failed to update branch (incorrect expected_head_sha).

dabacon:
This looks like it is out of sync with master.

mpharrigan:
cirqbot is supposed to merge master, run tests, and merge. I.e. this is not a problem on any other PR!

balopat:
There is also no `Update branch` button - I think there is something fishy going on here. 
I can merge master locally into the branch. @tonybruguier-google can you try to push a commit where you merge master in?

tonybruguier-google:
@balopat Yes, I had pushed the update button but I'm not sure whether it helped. I am still confused as to what happened.

Thanks everyone who's helped getting this in.

PR #2979, opened 2020-05-06
Enforce argument types in engine client.

### LINKED ISSUE #2900: Add list_processors to integration tests
For reference: PR #2899 fixes a bug due to argument types in list_processors. Issues of this type cannot be caught by mock-based unit tests (likely how it escaped notice to begin with) which means we need to check for it in our integration tests.

Additionally, there are several other methods in the `engine_client.py` file which have similarly-fragile type-agnostic tests. If any of those are not covered in integration tests, they probably should be.

95-martin-orion:
Fixes #2900.

Experimentally verified that this catches issues like the one fixed by #2899.

### DIFF
--- a/cirq/google/engine/client/quantum_v1alpha1/gapic/quantum_engine_service_client.py
+++ b/cirq/google/engine/client/quantum_v1alpha1/gapic/quantum_engine_service_client.py
@@ -17,6 +17,9 @@
 
 import functools
 import pkg_resources
+from typing import (
+    Dict, List, Optional, Sequence, Tuple, Union, TYPE_CHECKING
+)
 import warnings
 
 from google.oauth2 import service_account
@@ -30,6 +33,7 @@ import google.api_core.page_iterator
 import google.api_core.protobuf_helpers
 import grpc
 
+from cirq.google.engine.client.quantum_v1alpha1 import types as pb_types
 from cirq.google.engine.client.quantum_v1alpha1.gapic import enums
 from cirq.google.engine.client.quantum_v1alpha1.gapic import quantum_engine_service_client_config
 from cirq.google.engine.client.quantum_v1alpha1.gapic.transports import quantum_engine_service_grpc_transport
@@ -75,12 +79,12 @@ class QuantumEngineServiceClient(object):
     from_service_account_json = from_service_account_file
 
     def __init__(self,
-                 transport=None,
-                 channel=None,
-                 credentials=None,
-                 client_config=None,
-                 client_info=None,
-                 client_options=None):
+                 transport: Union[quantum_engine_service_grpc_transport.QuantumEngineServiceGrpcTransport] = None,
+                 channel: grpc.Channel = None,
+                 credentials: service_account.Credentials = None,
+                 client_config: Dict = None,
+                 client_info: google.api_core.gapic_v1.client_info.ClientInfo = None,
+                 client_options: Union[Dict, google.api_core.client_options.ClientOptions] = None):
         """Constructor.
 
         Args:
@@ -179,16 +183,16 @@ class QuantumEngineServiceClient(object):
         # These are the actual callables which invoke the proper
         # transport methods, wrapped with `wrap_method` to add retry,
         # timeout, and the like.
-        self._inner_api_calls = {}
+        self._inner_api_calls: Dict = {}
 
     # Service calls
     def create_quantum_program(self,
-                               parent=None,
-                               quantum_program=None,
-                               overwrite_existing_source_code=None,
-                               retry=google.api_core.gapic_v1.method.DEFAULT,
-                               timeout=google.api_core.gapic_v1.method.DEFAULT,
-                               metadata=None):
+                               parent: str = None,
+                               quantum_program: Union[Dict, pb_types.QuantumProgram] = None,
+                               overwrite_existing_source_code: bool = None,
+                               retry: Optional[google.api_core.retry.Retry] = google.api_core.gapic_v1.method.DEFAULT,
+                               timeout: Optional[float] = google.api_core.gapic_v1.method.DEFAULT,
+                               metadata: Optional[Sequence[Tuple[str, str]]] = None):
         """
         -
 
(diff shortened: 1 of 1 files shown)


cirq/google/engine/client/quantum_v1alpha1/gapic/quantum_engine_service_client.py
```diff
@@ -74,13 +76,16 @@ def from_service_account_file(cls, filename, *args, **kwargs):
...
-    def __init__(self,
-                 transport=None,
-                 channel=None,
-                 credentials=None,
-                 client_config=None,
-                 client_info=None,
-                 client_options=None):
+    def __init__(
+            self,
+            transport: Union[quantum_engine_service_grpc_transport.
```
viathor, line 81: This union has only one element. I think you meant `Optional[...]`?

cirq/google/engine/client/quantum_v1alpha1/gapic/quantum_engine_service_client.py
```diff
@@ -74,13 +76,16 @@ def from_service_account_file(cls, filename, *args, **kwargs):
...
+                             QuantumEngineServiceGrpcTransport] = None,
+            channel: grpc.Channel = None,
```
viathor, line 83: Since you're allowing `None` this should be `Optional[grpc.Channel]`. This applies throughout the PR.

Why hasn't mypy caught any of those?

cirq/google/engine/client/quantum_v1alpha1/gapic/quantum_engine_service_client.py
```diff
@@ -74,13 +76,16 @@ def from_service_account_file(cls, filename, *args, **kwargs):
...
+            credentials: service_account.Credentials = None,
+            client_config: Dict = None,
```
viathor, line 85: Is there really nothing we know about key and value type? Perhaps `Dict[str, Any]` applies?

cirq/google/engine/client/quantum_v1alpha1/gapic/quantum_engine_service_client.py
```diff
@@ -179,16 +184,18 @@ def __init__(self,
...
-    def create_quantum_program(self,
-                               parent=None,
-                               quantum_program=None,
-                               overwrite_existing_source_code=None,
-                               retry=google.api_core.gapic_v1.method.DEFAULT,
-                               timeout=google.api_core.gapic_v1.method.DEFAULT,
-                               metadata=None):
+    def create_quantum_program(
+            self,
+            parent: str = None,
```
viathor, line 192: `None` is definitely not an instance of the `str` type, so this should be `Optional[str]`.

cirq/google/engine/client/quantum_v1alpha1/gapic/quantum_engine_service_client.py
95-martin-orion: From the function doc, I think I missed adding a `Callable` option to the union. Will fix.

cirq/google/engine/client/quantum_v1alpha1/gapic/quantum_engine_service_client.py
95-martin-orion: Wrapped with `Optional`.

The note just above [this header](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#disabling-strict-optional-checking) in the mypy docs explains this: arguments with a default value of `None` aren't required to be `Optional` unless `--no-implicit-optional` is enabled.

cirq/google/engine/client/quantum_v1alpha1/gapic/quantum_engine_service_client.py
95-martin-orion: On the first pass I directly copied the types in the docstrings for each argument - I'll go back through now and narrow these down if possible.

95-martin-orion: Done - it looks like these are all for protobufs in `Dict[str, Any]` format.

cirq/google/engine/client/quantum_v1alpha1/gapic/quantum_engine_service_client.py
```diff
@@ -75,12 +78,18 @@ def from_service_account_file(cls, filename, *args, **kwargs):
     from_service_account_json = from_service_account_file
 
     def __init__(self,
-                 transport=None,
-                 channel=None,
-                 credentials=None,
-                 client_config=None,
-                 client_info=None,
-                 client_options=None):
+                 transport: Union[
```
dabacon, line 81: For a type this big would recommend creating a custom type.  QUANTIM_ENGINE_SERVICE_GRPC_TRANSPORT_LIKE?  Oh that feels like Java. Does that fit on one line?

95-martin-orion:
The failed check appears to have timed out - there may be a high density of PRs running checks right now. I'll re-request the check later today.

wcourtney:
Driveby...

1. I think that this file is auto-generated from the service protos, so if there are significant changes that require another autogen (it's already happened once since this was first added ~6 months ago), then this might be a painful merge. Could this be enforced at a layer outside of the generated code that we can maintain more easily?

2. I think that this is still error-prone if the user skips kwargs and the types happen to work out ok. Should we force kwargs with a leading * param?

95-martin-orion:
> Have we looked into using mypy-protobuf?

The files in question aren't the usual proto-generated Python files, but more readable client libraries for GAPIC. I don't think mypy-protobuf has support for generating these, but I haven't looked too deeply.

I'm currently focused on checking if the GAPIC tools have support for mypy annotation.

dstrain115:
I am fine with this approach, but just realize we may accidentally blow the type info away if we regenerate these files.  You can comment on that hypothetical future PR ti retain the typing info though!

95-martin-orion:
Discussed offline - since the GAPIC tools don't document any support for mypy, I'll instead add a note on how to re-populate mypy annotations after an update. The good news here is that GAPIC tools _do_ provide type information in the function docstrings, so cleanup is mostly painless.

dabacon:
I would be great if we can create a tool that does these changes for us.  Also this might be something we should raise as an issue over on the grpc repo.

Issue #2817, opened 2020-03-06
EngineJob.results() shouldn't give up


mpharrigan:
Right now if you check for results for more than 500 seconds, it just gives up. In practice, this causes more harm than good. It can take down a whole run (by raising an exception) if someone or many someones slots in a sufficient amount of work in front of you in the queue

dabacon:
On the other hand we have had loser who have "lost their server" and could not figure out how to shut down their job.

Maybe the solution is to set this at a day or something like that.

dstrain115:
Timeout is now configurable and defaults to not timing out.

PR #2996, opened 2020-05-12
Add timeout to engine object

dstrain115:
- Adds timeout to Engine object and makes it configurable
- This plumbs the timeout down to results object so that the
amount of polling is configurable.

Solves #2817

### DIFF
--- a/cirq/google/engine/engine.py
+++ b/cirq/google/engine/engine.py
@@ -71,7 +71,8 @@ class EngineContext:
                  proto_version: Optional[ProtoVersion] = None,
                  service_args: Optional[Dict] = None,
                  verbose: Optional[bool] = None,
-                 client: 'Optional[engine_client.EngineClient]' = None) -> None:
+                 client: 'Optional[engine_client.EngineClient]' = None,
+                 timeout: Optional[int] = None) -> None:
         """Context and client for using Quantum Engine.
 
         Args:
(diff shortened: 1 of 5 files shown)


cirq/google/engine/engine.py
```diff
@@ -138,6 +139,8 @@ def __init__(
                 configure options on the underlying client.
             verbose: Suppresses stderr messages when set to False. Default is
                 true.
+            timeout: Timeout for polling for results, in seconds.  Default is
+                to never timeout
```
dabacon, line 143: nit on period at end of comments.

cirq/google/engine/engine.py
```diff
@@ -150,6 +153,7 @@ def __init__(
                                     service_args=service_args,
                                     verbose=verbose)
         self.context = context
```
dabacon, line 155: does it make sense to put this on the context?

cirq/google/engine/engine.py
dstrain115: Done.

cirq/google/engine/engine.py
dstrain115: Sounds like a good idea to me.  Done.

cirq/google/engine/engine_job.py
```diff
@@ -262,10 +262,15 @@ def results(self) -> List[study.TrialResult]:
...
+            total_seconds_waited = 0.0
+            timeout = self.context.timeout
+            while True:
+                if timeout and total_seconds_waited >= timeout:
+                    break
                 if job.execution_status.state in TERMINAL_STATES:
                     break
                 time.sleep(0.5)
+                total_seconds_waited += 0.5
                 job = self._refresh_job()
```
dabacon, line 274: I guess this will not be exactly the timeout time since refresh job takes time.  But probably ok.

PR #2999, opened 2020-05-13
Fix some documentation for engine

dabacon:
* Clarified that not specifying the proto version gives you version 2
* Removed protoversion defaults from example code
* Fixed remaining use of `get_latest_calibration` in docs, which no longer exists.


dabacon:
Updated top level comment with list of changes.  Sorry should have included up front.

Issue #2313, opened 2019-10-10
Circuit._repr_svg_ with nice diagrams for ipython


mpharrigan:
https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display

Also see https://github.com/perak/quantum-circuit for a MIT licensed package that makes nice SVG circuits

mpharrigan:
Partially done in #2323 

I'm not going to work on it any more. Someone could pick this up and address the following things

 - make it prettier
 - actually distribute the horizontal spacing / don't assume constant row-height
 - thorough testing
 - I think it breaks if you have empty moments (?)
 - Generalize the textdiagramdrawer / blockdiagramdrawer to reduce duplicated code
 - Move from contrib to main trunk
 - Add `_repr_svg_` to `cirq.Circuit` instead of requiring wrapping in a helper class (this one must come after the others)

mpharrigan:
- fix spacing routines to not require integer coordinates

mpharrigan:
- deal with `MatrixGate`

vanhalen42:
Hey @mpharrigan, I am right now working on the following issues of #4499.

- [ ] Implement a fix that no longer requires circuits to not require integer coordinates for spacing routines.

For this, I changed the `int` to `float` and it still worked fine for me. What specific bug were you facing while using non-integer coordinates? 

https://github.com/vanhalen42/Cirq/blob/baa4310b7dddd8603f530771ee63e37701cf859b/cirq-core/cirq/contrib/svg/svg.py#L189-L217

- [ ] Implement a dynamic horizontal spacing framework for the rows in the SVGs of the circuits.

https://github.com/vanhalen42/Cirq/blob/baa4310b7dddd8603f530771ee63e37701cf859b/cirq-core/cirq/contrib/svg/svg.py#L166-L171

I see that currently, the implementation has hard-coded values. Do you have any idea how we could make this dynamically space the lines based on the user's dimensions? May I know what exactly do you prefer for the second task listed?

westurner:
See also:
- https://github.com/quantumlib/Cirq/issues/1492
- https://github.com/quantumlib/Cirq/issues/2313
- https://github.com/quantumlib/Cirq/issues/5981
- https://github.com/ipython/ipython/blob/main/IPython/core/display_functions.py#L207-L240

PR #3007, opened 2020-05-14
[SVG Circuits] Give up gracefully on gates with big labels

mpharrigan:
`MatrixGate` tries to print its whole matrix which is just huge. This PR hacks in a check that uses `?` for any gate whose label includes a newline. I will quickly remind reviewers that this functionality is in contrib, xref #2313 if anyone wants to "productionize" this functionality.


dabacon:
Coverage doesn't like you.

Issue #3004, opened 2020-05-14
Move capture_logging from _compat_testing.py to cirq.testing


dabacon:
Code that wants to test deprecation uses this and currently we import from _compat_testing which is a bit ugly

dabacon:
Or alternatively create a test for testing deprecation.

dabacon:
Fixed in #3009 

We now have `cirq.testing.assert_logs`

PR #3009, opened 2020-05-15
ContextManager for testing logs

dabacon:
Adds a helper for testing logs.
```
with cirq.testing.assert_logs('bacon', 'eggs', 'celery', count=2):
    logging.error('bacon and eggs is not available')
    logging.error('only celery is available')
```
This removes capture code from `cirq._compat_test` which was used to test deprecation warnings across our codebase.

Features:

* Ability to set log level.
* Ability to capture or not capture from python's `warnings` module.

PR also cleans up two things encountered while using:

* Use `import cirq` pattern in `moments_test.py`
* Fix unused type annotations in parity_gates.py.

### DIFF
--- a/cirq/_compat_test.py
+++ b/cirq/_compat_test.py
@@ -12,14 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import logging
 import types
-from typing import ContextManager, List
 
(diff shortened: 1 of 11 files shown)


dabacon:
Fixes #3004

Issue #2360, opened 2019-10-16
Only run build proto during continuous integration when protos change


dabacon:

dabacon:
This requires modifying dev_tools/build-protos.sh

PR #3012, opened 2020-05-16
Add increment proto build and use in CI

dabacon:
Adds a script to only build changed proto files in `cirq/google/api`.

Updates Google Actions to use this script instead of the full build.

Currently includes two changed proto files to test.  Will revert these before submission. 

Could have also made one script but opted for separate. Feedback welcome :) 

Also updates the full build tool to automatically grab all protos in `cirq/google/api` instead of having a hard coded list.

Fixes some documentation.


dabacon:
Fixes #2360

dabacon:
Fixes #2360

.github/workflows/ci.yml
```diff
@@ -129,8 +129,8 @@ jobs:
           python -m pip install -r dev_tools/conf/pip-list-dev-tools.txt
           wget https://github.com/bazelbuild/bazel/releases/download/0.26.0/bazel_0.26.0-linux-x86_64.deb
           sudo dpkg -i bazel_0.26.0-linux-x86_64.deb
-      - name: Build protos
-        run: check/build-protos
+      - name: Build changed protos
+        run: check/build-changed-protos
```
Strilanc, line 133: We're positive we can't break the proto build except by editing the protos themselves? For example, if we update the build-changed-protos file but not the protos themselves...

dabacon: This is a goodpoint.  If the build files are changed, then I think we should build the corresponding proto.

dabacon: PTAL.  Added that it builds all protos if the BUILD rule has changed.

check/build-changed-protos
```diff
@@ -0,0 +1,99 @@
...
+if [[ ! -z "$changed" ]]; then
+    for base in $changed:
+    do
+        echo -e "Found changed proto: '${base}.proto'"
+    done
+fi
+
+# Check if any of the build files have changed.
+build_changed=$(git diff --name-only ${rev} -- \
+    | grep "google/api/.*BUILD$" \
```
Strilanc, line 76: Also this script file itself could change.

Strilanc:
The logic in this file is getting quite complicated.  Check out `bash_scripts_test.py` if you want to test it.

dabacon:
The replaced Github action passes, but for some reason the status check is not reporting it as so.

dabacon:
This is stuck on the old action not being reported.  Is there further config that needs to be done for actions?  May try renaming it to old name.

PR #3015, opened 2020-05-17
Disentangle MS gate and XXPowGate. 

### LINKED ISSUE #2958: Mølmer–Sørensen gate should be disentangled from XX parity gate
The Mølmer–Sørensen gate is nothing more than 
`ops.XXPowGate(exponent=rads*2/np.pi, global_shift=-0.5)`
But in general that gate is usually only called that by the ion trap community (since it referes to a specific very cool trick to couple via an oscillator mode).  Right now we special case XXPowGate names when it is this gate.  

This also leads to a strange coupling between ops and ion, not in code but in naming.

I think we could make the MS gate just the `XXPowGate` with the name change in ion.  If that were done then consumers who don't know about ions could consume it as a `XXPowGate` while it would show up ion trap code as desired.

tanujkhattar:
To fix #2958


cirq/ion/ion_gates_test.py
```diff
@@ -44,9 +48,12 @@ def test_ms_matrix():
 
 
 def test_ms_repr():
+    ms = cirq.ms(np.pi / 2)
```
dabacon, line 51: nit: moved down to right where this is used to make it more readable.  I was wondering what you were doing with it for a while ;)

cirq/ion/ion_gates_test.py
```diff
@@ -20,12 +20,16 @@
 def test_ms_arguments():
     eq_tester = cirq.testing.EqualsTester()
     eq_tester.add_equality_group(cirq.ms(np.pi / 2),
-                                 cirq.XXPowGate(global_shift=-0.5))
+                                 cirq.ion.ion_gates.MSGate(rads=np.pi / 2))
```
dabacon, line 24: add another equality group with the XXPowGate (since it shouldn't be equal).

cirq/ion/ion_gates_test.py
tanujkhattar: done

tanujkhattar:
Resolved the comments. PTAL.

PR #3016, opened 2020-05-17
Modify PointOptimizer to handle multiple modified gates acting on the same qubit in a moment

### LINKED ISSUE #2990: Changing qubits in PointOptimizer causes ValueError
The cirq.PointOptimizer class assumes that an optimizer won't change qubits (or at least won't change qubits in a way that pushes the new gate backwards).   If two gates are modified in the same moment to use the same qubits, a ValueError is raised.

A check similar to https://github.com/quantumlib/Cirq/blob/master/cirq/circuits/optimization_pass.py#L128
needs to be added at or around Line 148 to detect this condition and adjust the 'i' index or frontier accordingly.

A minimal example that causes the issue is below.
```
class EverythingIs42(cirq.PointOptimizer):
        if len(op.qubits) == 1:
          gate = op.gate
          new_op = op.gate(cirq.LineQubit(42))
          return cirq.PointOptimizationSummary(clear_span=1,
                                             clear_qubits=op.qubits,
                                             new_operations=new_op)

c=cirq.Circuit(cirq.X(cirq.LineQubit(0)), cirq.X(cirq.LineQubit(1)))
EverythingIs42().optimize_circuit(c)
```

tanujkhattar:
To fix #2990


dstrain115:
Would it be possible to warn the user if they have done this, so they get a useful warning message or a useful exception?  After considering it, I think the point you brought up is rather critical.  If the two operations are non-commuting and act on the same qubit, we should alert the user that they have done something irreconcilable.

cirq/circuits/optimization_pass.py
```diff
@@ -145,6 +145,18 @@ def optimize_circuit(self, circuit: Circuit):
...
                 new_operations = self.post_clean_up(
                     cast(Tuple[ops.Operation], opt.new_operations))
-                circuit.insert_at_frontier(new_operations, i, frontier)
 
+                flat_new_operations = tuple(ops.flatten_to_ops(new_operations))
+
+                new_qubits = set()
+                for flat_op in flat_new_operations:
+                    for q in flat_op.qubits:
+                        new_qubits.add(q)
```
balopat, line 154: This logic to flatten the new ops and collecting the qubits from them is exactly repeated at the beginning of `insert_at_frontier`, I think we should just add this logic there: 

https://github.com/quantumlib/Cirq/blob/b3728aaf6d64477e0913015edec2223748aab54e/cirq/circuits/circuit.py#L1320-L1323

WDYT?

tanujkhattar: You mean add the `new_qubits.issubset(set(opt.clear_qubits))` condition in `insert_at_frontier`? Though `insert_at_frontier` is currently only called from `optimization_pass`, I think the condition of checking that new qubits should be a subset of original qubits should logically belong in the `PointOptimizer` and not in `insert_at_frontier`.

balopat: Yeah, that's what I meant. I am not confident that the abstractions are crisp here - the whole concept of a frontier only comes up in optimizers / compilers - so why does the Circuit class need to know about it. 
But you are right that this logic is an optimizer concern - also I totally missed that for my suggestion to work you'd have to pass `opt` as well to `insert_at_frontier` which makes the situation worse, leaking even more optimizer concepts to the Circuit class. 

So let's keep it this way, and we'll live with this small amount of duplication for now.

PR #3017, opened 2020-05-17
Make it easier to implement cirq.Sampler

Strilanc:
- Add `sample_dict` / `sample_dicts` methods to Sampler
- Use `cirq.ABCMetaImplementAnyOneOf` to derive any of run/run_sweep/sample_dict/sample_dicts from any of the others
- Fix deprecation warning from pytest by bumping version of pytest-asyncio
- Fix `cirq.to_sweep` not saying it takes `cirq.Sweepable`

Discussion points:

The implementer still has to deal with boilerplate parameter resolution. Is there some nice way to expose a `params` argument to the user without requiring the implementer to deal with it? I came up with something interesting but possibly confusing:

```
    @value.boilerplate(lambda wrapped_function, self, program, *, params = None:
                       wrapped_function(self, program=protocols.resolve_parameters(program, params)))
    @value.alternative(requires='sample_dicts',
                       implementation=_sample_dict_from_sample_dicts)
    def sample_dict(
            self,
            program: 'cirq.Circuit',
    ) -> Dict[str, Any]:
```

### DIFF
--- a/cirq/study/sweepable.py
+++ b/cirq/study/sweepable.py
@@ -14,13 +14,15 @@
 
 """Defines which types are Sweepable."""
 
-from typing import Dict, Iterable, Iterator, List, Union, cast
+from typing import Dict, Iterable, Iterator, List, Union, cast, TYPE_CHECKING
 import itertools
 
 from cirq._doc import document
 from cirq.study.resolver import ParamResolver, ParamResolverOrSimilarType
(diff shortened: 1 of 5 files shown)


balopat:
Discussed offline - closing it.

mpharrigan:
Can we delete this branch?

PR #3024, opened 2020-05-19
Add `with_probability` modifier to Gate and Operation

Strilanc:
- Define `cirq.ProbableGate`

### DIFF
--- a/cirq/ops/probable_gate.py
+++ b/cirq/ops/probable_gate.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import numbers
-from typing import Tuple, TYPE_CHECKING, Dict, Any
+from typing import Tuple, TYPE_CHECKING, Dict, Any, cast, SupportsFloat
 
 import numpy as np
 
@@ -30,8 +30,8 @@ class ProbableGate(raw_types.Gate):
     """Applies a sub operation with some probability."""
 
     def __init__(self, *, sub_gate: 'cirq.Gate', probability: value.TParamVal):
-        if isinstance(probability,
-                      numbers.Number) and not 0 <= float(probability) <= 1:
+        if isinstance(probability, numbers.Number) and not 0 <= float(
+                cast(SupportsFloat, probability)) <= 1:
             raise ValueError("not 0 <= probability <= 1")
 
         self.sub_gate = sub_gate
--- a/cirq/ops/probable_gate.py
+++ b/cirq/ops/probable_gate.py
@@ -1,4 +1,4 @@
-# Copyright 2018 The Cirq Developers
+# Copyright 2020 The Cirq Developers
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -27,7 +27,7 @@ if TYPE_CHECKING:
 
 @value.value_equality(approximate=True)
 class ProbableGate(raw_types.Gate):
-    """Applies a sub operation with some probability."""
+    """Applies a sub gate with some probability."""
 
     def __init__(self, *, sub_gate: 'cirq.Gate', probability: value.TParamVal):
         if isinstance(probability, numbers.Number) and not 0 <= float(
--- a/cirq/ops/probable_gate.py
+++ b/cirq/ops/probable_gate.py
@@ -1,0 +1,132 @@
+# Copyright 2018 The Cirq Developers
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numbers
+from typing import Tuple, TYPE_CHECKING, Dict, Any
+
+import numpy as np
+
+from cirq import protocols, value
+from cirq.ops import raw_types
+from cirq._compat import proper_repr
+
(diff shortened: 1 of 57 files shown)


cirq/ops/probable_gate.py
```diff
@@ -0,0 +1,132 @@
+# Copyright 2018 The Cirq Developers
```
dabacon, line 1: nit: 2020

cirq/ops/probable_gate.py
```diff
@@ -0,0 +1,132 @@
...
+
+if TYPE_CHECKING:
+    import cirq
+
+
+@value.value_equality(approximate=True)
+class ProbableGate(raw_types.Gate):
+    """Applies a sub operation with some probability."""
+
+    def __init__(self, *, sub_gate: 'cirq.Gate', probability: value.TParamVal):
```
dabacon, line 32: why not just `gate`?  I'm not sure what it's `sub` to.

cirq/ops/probable_gate.py
```diff
@@ -0,0 +1,132 @@
...
+from cirq.ops import raw_types
+from cirq._compat import proper_repr
+
+if TYPE_CHECKING:
+    import cirq
+
+
+@value.value_equality(approximate=True)
+class ProbableGate(raw_types.Gate):
+    """Applies a sub operation with some probability."""
```
dabacon, line 30: Mixes `operation` with `gate` which is confusing.  Suggest that you also be explicit about the do nothing part.

cirq/ops/probable_gate.py
```diff
@@ -0,0 +1,132 @@
...
+            self.sub_gate = self.sub_gate.sub_gate
+
+    def _qid_shape_(self) -> Tuple[int, ...]:
+        return protocols.qid_shape(self.sub_gate)
+
+    def _value_equality_values_(self):
+        return self.sub_gate, self.probability
+
+    def _has_unitary_(self):
+        return False
```
dabacon, line 52: what about probability=0 and gate is unitary?

cirq/ops/probable_gate.py
```diff
@@ -0,0 +1,132 @@
...
+from cirq import protocols, value
+from cirq.ops import raw_types
+from cirq._compat import proper_repr
+
+if TYPE_CHECKING:
+    import cirq
+
+
+@value.value_equality(approximate=True)
+class ProbableGate(raw_types.Gate):
```
dabacon, line 29: So far we have named things that can have mixtures or channels as `Channel`.  I'm a bit worried about breaking that consistency, though it is fairly clear what this can't be just a unitary so probably fine.

cirq/ops/raw_types.py
```diff
@@ -259,6 +260,12 @@ def __pow__(self, power):
     def __call__(self, *args, **kwargs):
         return self.on(*args, **kwargs)
 
+    def with_probability(self, probability: Union[float, sympy.Basic, float]):
```
dabacon, line 263: add return type

cirq/ops/raw_types.py
```diff
@@ -459,11 +466,18 @@ def controlled_by(self,
                 default to 1.
         """
         # Avoids circular import.
-        from cirq.ops import ControlledOperation
+        from cirq.ops.controlled_operation import ControlledOperation
         if len(control_qubits) == 0:
             return self
         return ControlledOperation(control_qubits, self, control_values)
 
+    def with_probability(self, probability: Union[float, sympy.Basic, float]):
```
dabacon, line 474: shouldn't this be on GateOperation and not on the general Operation?

cirq/ops/probable_gate.py
Strilanc: Done.

cirq/ops/probable_gate.py
Strilanc: Acknowledged. I don't think there's risk of confusion in this case. `ProbableGateChannel` seems worse..

cirq/ops/probable_gate.py
Strilanc: Fixed. I originally wrote this as an operation and this one slipped through.

cirq/ops/probable_gate.py
Strilanc: It's `sub` to `self`. Same as in `ControlledGate`.

cirq/ops/probable_gate.py
Strilanc: Which it's technically true that when probability is 0 or 1 that there is a unitary effect, I think it's fair to treat these cases uniformly with the rest. Similar to how the bit flip channel isn't considered unitary when p=1 and the various damping channels aren't unitary when gamma=0.

cirq/ops/raw_types.py
Strilanc: Done.

cirq/ops/raw_types.py
Strilanc: No, any operation can specify an underlying gate.

cirq/ops/probable_gate.py
dabacon: `RandomGateChannel`?  Good either way.

cirq/ops/probable_gate.py
dabacon: Make sense, except since the reader doesn't know it's a channel it's not clear its part of the channel ;) I find `sub` a throw away word, from context it's clear there is only one object necessary for construction of the object.

cirq/ops/probable_gate.py
dabacon: SGTM

kevinsung:
I know we discussed this at Cirq sync but I'd like to make a case against the name RandomGateChannel. RandomGate sounds like a gate drawn randomly from an ensemble of gates. I think the best name is PossibleGateChannel.

cduck:
I agree.  Alternative to arguing over the name, a follow up PR can add support for this to randomly apply one of a list of gates with given probabilities without changing the basic functionality.

PR #3026, opened 2020-05-19
Deprecate mixture_channel in favor of just mixture

Strilanc:
- mixture's behavior now matches what mixture_channel used to do (i.e. unitary objects have implicit mixtures)


cirq/protocols/mixture_protocol.py
```diff
@@ -82,63 +83,6 @@ def mixture(val: Any, default: Any = RaiseTypeErrorIfNotProvided
...
-    Args:
-        val: The value whose mixture_channel is being computed.
-        default: A default value if val does not support mixture.
-
-    Returns:
-        An iterable of tuples of size 2. The first element of the tuple is a
-        probability (between 0 and 1) and the second is the unitary that occurs
-        with that probability. The probabilities will sum to 1.0.
-    """
     mixture_getter = getattr(val, '_mixture_', None)
```
dabacon, line 86: not sure why it let me, but the docstring needs to be update to move away from the generic (prob, object) and to the specific (prob, unitary).

Strilanc: Done.

Strilanc:
Not sure why the tests failed. Not able to reproduce on my machine.

PR #3027, opened 2020-05-20
Fix PhaseDampingChannel's channel

Strilanc:
- I checked for equivalence with the definition from Preskill's notes (http://www.theory.caltech.edu/~preskill/ph219/chap3_15.pdf page 28) and found the existing definition was wrong
- Redefined the channel as a mixture of I and Z
- Added a test checking equivalence with the 3-term definition
- The old channel appears to still be a dephasing channel, but the damping had the wrong dependence on gamma.

### DIFF
--- a/cirq/ops/common_channels.py
+++ b/cirq/ops/common_channels.py
@@ -638,15 +638,9 @@ class PhaseDampingChannel(gate_features.SingleQubitGate):
 
             $$
             \begin{aligned}
-            M_0 =& \begin{bmatrix}
-                    1 & 0 \\
-                    0 & \sqrt{1 - \gamma}
-                  \end{bmatrix}
+            M_0 =& \sqrt{1 - \gamma / 2} I
             \\
-            M_1 =& \begin{bmatrix}
-                    0 & 0 \\
-                    0 & \sqrt{\gamma}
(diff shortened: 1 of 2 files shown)


Strilanc:
The old definition _behaves observably differently_ from the new one. In particular, it fails the test I added. They cannot be equivalent.

viathor:
Clarification: There are two questions here:
1. The choice of Kraus operators - there are many equivalent possibilities here. The ones before and after this PR are all correct.
2. The choice of parameter - this is a matter of convention.

I think it'd be good to follow a well-known text such as Mike&Ike, but as long as we document what we're doing I'm fine with it.

Strilanc:
Hmmm I see... so we're using a different convention. In that case I guess this is not an error since we do document the behavior.

viathor:
Yes, I agree. Also, our convention follows the main set of equations in MIke&Ike so it's a very decent choice.

dabacon:
https://quantumcomputing.stackexchange.com/questions/4809/confusion-on-the-definition-of-the-phase-damping-channel

dabacon:
Always have to check these question on the stack exchange to make sure it wasn't @Strilanc who asked or answered the question :)

PR #3029, opened 2020-05-21
Updates random_circuit for single qubit circuits

### LINKED ISSUE #3028: Random single qubit circuits always have empty moments
Single qubit circuits returned from [testing.random_circuit](https://github.com/quantumlib/Cirq/blob/master/cirq/testing/random_circuit.py#L38) are always empty with the default gate domain. For example:

```python
from cirq.testing import random_circuit

circuit = random_circuit(qubits=1, n_moments=4, op_density=0.99)
print(circuit)  # Blank

print(circuit.moments)
[cirq.Moment(), cirq.Moment(), cirq.Moment(), cirq.Moment()]
```

The reason is that the [default gate domain](https://github.com/quantumlib/Cirq/blob/master/cirq/testing/random_circuit.py#L23) contains two-qubit gates, so [`max_arity = 2`](https://github.com/quantumlib/Cirq/blob/master/cirq/testing/random_circuit.py#L89) which prevents any gates from being added to single qubit circuits.

This can be fixed by passing in a `gate_domain` with only one-qubit gates, but it seems useful to have the default behavior return non-empty circuits.

**Proposed change:** If the number of qubits is one, only select from single-qubit gates in the provided gate domain.

For example:

```python
if n_qubits == 1:
    gate_domain = {key: val for key, val in gate_domain.items() if val == 1}
```

Does this sound reasonable? If so I can submit a PR with this change.

rmlarose:
Fixes #3028.


rmlarose:
@googlebot I signed it!

cirq/testing/random_circuit.py
```diff
@@ -70,13 +75,15 @@ def random_circuit(qubits: Union[Sequence[ops.Qid], int],
...
     if not gate_domain:
         raise ValueError('gate_domain must be non-empty')
-    max_arity = max(gate_domain.values())
 
     if isinstance(qubits, int):
         qubits = tuple(ops.NamedQubit(str(i)) for i in range(qubits))
     n_qubits = len(qubits)
     if n_qubits < 1:
         raise ValueError('At least one qubit must be specified.')
+    if n_qubits == 1:
```
dabacon, line 84: actually now that I think about this, we should probably just make it select the gates that are less than the number of qubits.

gate_domain = {key: val for key, val in gate_domains.itmes() if val <= n_qubits}

PR #3037, opened 2020-05-26
Fix empty gate domain problem in random_circuit

### LINKED ISSUE #3033: Random circuit can fail due to empty gate domain
Random circuit was changed recently to make sure that gates that have too many qubits are not used in the random circuit.  This had the side effect that the sometimes removed all of the gates from the domain, which causes `test_random_circuit` to fail.

dabacon:
Fixes #3033 

Fixes a bug where an empty gate domain occurs due to there not being any gates that operate on the proper number of qubits and the circuit has fewer number of qubits.

Also cleans up random circuit tests
* Adds `DEFAULT_GATE_DOMAIN` to `cirq.testing`, users might want to use it for the circuits they generate
* Allows for an `op_density` of 1.0.
* Correct an incorrect documentation assertion about what `op_density` means and added a test that demonstrates why that document assertion was incorrect.
* Moved to standard style for importing modules and not methods.

### DIFF
--- a/cirq/testing/random_circuit_test.py
+++ b/cirq/testing/random_circuit_test.py
@@ -12,46 +12,46 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from typing import Optional, Dict, Sequence, Union, cast
-from random import randint, random, sample, choice
+import random
 
 import numpy as np
 import pytest
 
 import cirq
-from cirq.testing.random_circuit import random_circuit, DEFAULT_GATE_DOMAIN
+import cirq.testing
 
 
 def test_random_circuit_errors():
-    with pytest.raises(ValueError):
-        random_circuit(randint(1, 10), randint(1, 10), -1)
-    with pytest.raises(ValueError):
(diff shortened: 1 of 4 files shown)


cirq/testing/random_circuit_test.py
```diff
@@ -12,81 +12,93 @@
...
-from cirq.testing.random_circuit import random_circuit, DEFAULT_GATE_DOMAIN
+import cirq.testing
 
 
 def test_random_circuit_errors():
-    with pytest.raises(ValueError):
-        random_circuit(randint(1, 10), randint(1, 10), -1)
-    with pytest.raises(ValueError):
-        random_circuit(randint(1, 10), randint(1, 10), 1.)
+    with pytest.raises(ValueError, match='-1'):
```
dstrain115, line 25: Should we add more to the matching statement, since this could be any error that prints out this argument?

dabacon: Done

PR #3038, opened 2020-05-26
Adds pylint on only changed files.

### LINKED ISSUE #2021: Add `check/pylint-changed-files` script
Linting the code base is beginning to take an unreasonably large amount of time. It would be convenient to be able to only lint files that changed.

Basically, this change amounts to rebasing the diffs of `check/pytest-changed-files` vs `check/pytest` onto `check/pylint`.

Note that `check/all` has an `--only-changed-files` flag, and when this flag is set it should call the only-changes version of pylint.

Should also update `docs/development.md`

dabacon:
Only run pylint on changed files via a new check script.  Uses this in github CI. 

Fixes #2021


dabacon:
> Can we have it run the normal pylint unless that flag is passed, then run the incremental pylint?

I think the difference here is that the testing only changed files can miss a lot of broken tests, whereas running pylint with only the changed files shouldn't allow any lint to get into the code.  I guess this then begs the difference of why include the full one, and maybe I should just have one which defaults to changed, but can be made to run on all.

Strilanc:
This will miss lint failures such as... you remove a keyword from a function, and that keyword is still used in another file.

dabacon:
Move to keep pylint for full and piped into check all.  PTAL

dev_tools/bash_scripts_test.py
```diff
@@ -64,6 +64,8 @@ def run(
 dir=$(git rev-parse --show-toplevel)
 cd {}
 git init --quiet
+git config --global user.name 'Me'
```
vtomole, line 67: When this is executed locally, it overrides the local git config. Your most recent commits were done by `Me`: https://github.com/quantumlib/Cirq/pull/3038/commits

dabacon: LOL doh.

dabacon:
CLA yes: googler (me) made commit that was incorrectly named due to git config being set during test.

dabacon:
@googlebot see previous statement

dabacon:
> For the `check/all`, we have
> 
> ```
> If --only-changed-files is specified, pytest-changed-files will be run instead of pytest-and-incremental-coverage.
> ```
> 
> Can we have it run the normal pylint unless that flag is passed, then run the incremental pylint?

PTAL

dabacon:
If you look at that action, you see that the github action for the full test passes, but the changed one does not.  I think there is a bug in the @only_on_posix function, created https://github.com/quantumlib/Cirq/issues/3059 to look into this

dabacon:
Removed git config from actions since it really only applies to this one test.

dabacon:
Fixes #3059

PR #3042, opened 2020-05-27
Add _mixture_ support to Controlled{Operation,Gate}

Strilanc:


cirq/ops/controlled_operation.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from typing import (Any, cast, Collection, Dict, Optional, Sequence, Tuple,
-                    Union, TYPE_CHECKING)
+                    Union, TYPE_CHECKING, List)
```
dabacon, line 15: nit: abc

Strilanc: Acknowledged. I think if we want to enforce sorted imports we should use a tool to automate it.

dabacon: Indeed.  https://github.com/timothycrosley/isort

PR #3043, opened 2020-05-27
Add support for monte carlo channel simulation (quantum trajectories)

Strilanc:


cirq/sim/act_on_state_vector_args.py
```diff
@@ -227,3 +228,33 @@ def _strat_act_on_state_vector_from_mixture(action: Any,
...
+            left_matrix=kraus_tensors[i],
+            right_target=args.target_tensor,
+            target_axes=args.axes,
+            out=args.available_buffer,
+        )
+        weight = np.linalg.norm(args.available_buffer)**2
+        p -= weight
+        if p < 0:
+            break
+
```
dkafri, line 258: I would also assert p<0 after the loop, as a sanity check.

Strilanc: It may not be the case that `p < 0`, due to floating point error, though it should be very small if not.

cirq/sim/act_on_state_vector_args.py
```diff
@@ -227,3 +228,35 @@ def _strat_act_on_state_vector_from_mixture(action: Any,
...
+        return NotImplemented
+
+    shape = protocols.qid_shape(action)
+    kraus_tensors = [
+        e.reshape(shape * 2).astype(args.target_tensor.dtype)
+        for e in kraus_operators
+    ]
+    p = args.prng.random()
+    weight = None
+    for i in range(len(kraus_tensors)):
```
dabacon, line 247: is there a version of this in which `i` is recorded?

cirq/sim/act_on_state_vector_args.py
```diff
@@ -227,3 +228,35 @@ def _strat_act_on_state_vector_from_mixture(action: Any,
...
+            target_axes=args.axes,
+            out=args.available_buffer,
+        )
+        weight = np.linalg.norm(args.available_buffer)**2
+        p -= weight
+        if p < 0:
+            break
+
+    assert weight is not None, "No Kraus operators"
+    args.available_buffer /= np.sqrt(weight)
```
dabacon, line 260: I think we can get weight 0 due to a floating point error?  

Say there are two outcomes, one with probability evaluating to 0.99998 and the other to 0, then if p=0.99999 then we would skip the first result and get to the second, in which case this would divide by zero.

cirq/sim/act_on_state_vector_args.py
Strilanc: Not at the moment.

cirq/sim/act_on_state_vector_args.py
Strilanc: Fixed and added a unit test triggering it.

PR #3045, opened 2020-05-28
Add dataclass annotation to CrossEntropyResult

dabacon:
Fairly certain this was not meant to be a class level variable.

### DIFF
--- a/cirq/experiments/cross_entropy_benchmarking.py
+++ b/cirq/experiments/cross_entropy_benchmarking.py
@@ -62,6 +62,7 @@ class CrossEntropyDepolarizingModel:
     covariance: Optional[np.ndarray] = None
 
 
+@dataclasses.dataclass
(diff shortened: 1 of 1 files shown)


cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -62,6 +62,7 @@ class CrossEntropyDepolarizingModel:
     covariance: Optional[np.ndarray] = None
 
 
+@dataclasses.dataclass
 @protocols.json_serializable_dataclass(frozen=True)
```
kevinsung, line 66: The line below it, `@protocols.json_serializable_dataclass(frozen=True)`, already turns the class into a dataclass.

dabacon:
Oh thanks I hadn't noticed that.  For some reason this class is giving the documentation a big problem.

Issue #2978, opened 2020-05-06
sphynx/readthedocs not listing/generating class method documentation


Strilanc:
For example, https://cirq.readthedocs.io/en/stable/generated/cirq.Moment.html has no list of the methods of `cirq.Moment`.

Strilanc:
The relevant file to edit is probability `docs/conf.py`, which is the sphynx configuration code. For example, this is where the "skip member" method is defined ( https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#skipping-members ).

I recommend that, when testing changes, you temporarily reduce `api.rst` to only mention a single class. This makes document generation *way* faster. Taking seconds instead of minutes.

dabacon:
I've looked into this and it is really wanky.  One problem is that the skip method is not correct as it ignores the `skip` variable.  But even when you fix that sphix seems to run the skip method twice, and the first time it runs it ignores the result.  This results in generated every method for a class, even after correctly configuring sphinx to only keep some methods.   

Going to poke around a bit more, but right now the only real option is either a) include all methods (which includes all possible dunder methods) or b) roll back the @documentation annotation to remove the skip logic.

dabacon:
I've confirmed that this was broken by #2659

PR #3048, opened 2020-05-29
Fix methods missing from doc strings

dabacon:
This restores summaries of methods on doc strings. These are missing from 0.8.0, so we should consider a bug fix branch and update the docs on ReadTheDocs.

Skipping the private methods is required or else all methods (all of those from `__dir__`).

Note that this introduces a new problem that I've been unable to fix yet, which is that you cannot assign classes to one another and have both of them appear in the api documentation.  Will open an issue for this after submitting this.

### DIFF
--- a/cirq/ops/common_gates.py
+++ b/cirq/ops/common_gates.py
@@ -1094,10 +1094,7 @@ document(
          [. . . -1]]
     """)
 
-
 CNotPowGate = CXPowGate
-
-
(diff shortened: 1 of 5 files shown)


dabacon:
Ah this broke test.  Sorry will see if I can fix.

dabacon:
Fixes #2978

dabacon:
Ugh still some breakage.  Investigating.

Issue #2905, opened 2020-04-14
Tagging noise operations clutters the circuit output


mpharrigan:
The extra `[]` are from the tag, I think

xref #2719 

```
0: ───X───D(0.005)[]───H────────D(0.005)[]───

1: ───X───D(0.005)[]───H────────D(0.005)[]───

2: ───────D(0.005)[]───H────────D(0.005)[]───

3: ───────D(0.005)[]───X^-0.5───D(0.005)[]───
```

dabacon:
This is when you add the noise to the circuit?  I think that should remove the tags.

mpharrigan:
Interestingly, this is more broken than I first thought. In the terminal it displays as

```
(0, 0): ───I───D(0.001)[<virtual>]───I───D(0.001)[<virtual>]───I───D(0.001)[<virtual>]───

(0, 1): ───I───D(0.001)[<virtual>]───I───D(0.001)[<virtual>]───I───D(0.001)[<virtual>]───

(0, 2): ───I───D(0.001)[<virtual>]───I───D(0.001)[<virtual>]───I───D(0.001)[<virtual>]───
```

but the `<virtual>` tags are interpreted as badly formed html tags in e.g. Jupyter notebook and are not displayed. In the SVGCircuit implementation, the virtual tag hides the closing `]`

dabacon:
For physical versus virtual Z gates, clearly seeing the tag is important.  But for noice, I'm not sure this is correct, and as I say in #2984 I don't think we should reuse virtual in this form for noise.  It is re-using a tag for no real reason.

mpharrigan:
Based on the above, the "virtual" specifier probably gets eaten in jupyter notebook for virtual Z gates then

mpharrigan:
The behavior has changed. It now does `D(0.001)[cirq.VirtualTag()]`

mpharrigan:
I'm contributing a hack in https://github.com/quantumlib/Cirq/pull/3076/files#diff-d949b9f3211c62c5defeb9f7d1439416R16 which should be reverted after this is fixed

balopat:
[[ Note to self: don't include issue reference in a commit message. It will keep spamming the issue from every fork that rebases the commit :) ]]

MichaelBroughton:
Is this still relevant @mpharrigan ?

mpharrigan:
The notebook isn't broken, but it's still pretty cluttery

```
[...]

PR #3051, opened 2020-05-30
Fix VirtualTag

### LINKED ISSUE #3006: Why is cirq.VirtualTag not at top level
Currently you have to say `cirq.ops.VirtualTag`.  Is there a reason it doesn't appear at the top level as `cirq.VirtualTag`?

viathor:
Fixes #3006 and repr and adds serialization.


mpharrigan:
xref #2905

Issue #2948, opened 2020-04-29
Update CliffordSimulator state through method dispatch on Gates


smitsanghavi:
Spun out from #2423.

Clifford Simulator currently special cases handling for multiple Gates through hardcoding operations on the CliffordTableau and StabilizerCHForm in the simulator. The CliffordTableau and StabilizerCHForm logic also, in turn, has hardcoded handling for these gates as private methods.

This is unpythonic; the goal is to implement method dispatch and have this handling exported to individual gates and just use a protocol/protocols in the Clifford Simulator. The protocol(s) would update the internal state of the tableau or the ch_form directly through the gate's defined method. It would also have a generalized, but more expensive, fallback to decomposing single-qubit unitaries into corresponding Cliffords if the method is not defined on that Gate.

Whether to have 2 separate protocols for `update_clifford_tableau` and `update_ch_form` or unify it to `update_clifford_state` is similar to the discussion around generality vs specificity in context of #2946 .

PR #3063, opened 2020-06-03
Add ActOnCliffordTableauArgs. Support it in _act_on_ for common gates and measurements.

smitsanghavi:
#2948 #2423


smitsanghavi:
Friendly Ping!

cirq/ops/common_gates_test.py
```diff
@@ -272,6 +261,232 @@ def test_h_str():
     assert str(cirq.H**0.5) == 'H^0.5'
 
 
+def test_x_act_on():
```
Strilanc, line 264: Create a testing method `def assert_clifford_tableau_effect_matches_unitary():`

Implement the method as follows:

1. Create a circuit that prepares a Bell pair for each qubit to be input into the gate (H + CNOT), then runs the gate (acting on one qubit from each Bell pair).
2. Create a method `stabilizers_to_state_vector` which creates the state stabilized by a given list of commuting stabilizers. Do this by generating a random state and then iteratively projecting it onto each stabilizer's +1 eigenstate.
3. Compare the result of `cirq.final_state_vector(circuit)` to (the equivalent of) `stabilizers_to_state(cirq.final_clifford_tableau(circuit))`.

Alternatively, you can check that the state returned by the state vector simulator is in fact stabilized by each of the claimed stabilizers in the Clifford tableau. That would be more efficient in general.

The key thing is that we want to be able to base the correctness of the Clifford simulation on the correctness of the state vector simulator.

Strilanc: Maybe we could implement a utility method `state_vector_has_stabilizer(np.ndarray, PauliString)`.

smitsanghavi: Sounds good. Went with the state_vector_has_stabilizer approach.

smitsanghavi:
Tests failing due to #2468

smitsanghavi:
Please review

cirq/testing/consistent_act_on.py
```diff
@@ -0,0 +1,81 @@
...
+from cirq.circuits.circuit import Circuit
+from cirq.devices import LineQubit
+from cirq.ops import common_gates
+from cirq.ops.dense_pauli_string import DensePauliString
+from cirq import protocols
+from cirq.sim import act_on_state_vector_args, final_state_vector
+from cirq.sim.clifford import act_on_clifford_tableau_args, clifford_tableau
+
+
+def state_vector_has_stabilizer(state_vector: np.ndarray,
```
Strilanc, line 28: This method currently mutates the `state_vector` argument and makes a copy that it doesn't mutate. Switch it around so the input argument is not mutated but the copy is mutated.

cirq/testing/consistent_act_on.py
```diff
@@ -0,0 +1,81 @@
...
+                                stabilizer: DensePauliString) -> bool:
+    original_state_vector = state_vector.copy()
+    args = act_on_state_vector_args.ActOnStateVectorArgs(
+        target_tensor=state_vector,
+        available_buffer=np.empty(state_vector.shape, dtype=np.complex64),
```
Strilanc, line 33: Prefer `np.empty_like(state_vector)`.

cirq/testing/consistent_act_on.py
Strilanc, line 28: Add a docstring. Note that the stabilizer has to be correct even including global phase (e.g. Z is a stabilizer of |0> but not |1>; that would be stabilized by -Z instead).

cirq/testing/consistent_act_on.py
```diff
@@ -0,0 +1,81 @@
...
+        circuit.append([
+            common_gates.H(qubits[2 * i]),
+            common_gates.CNOT(qubits[2 * i], qubits[2 * i + 1])
+        ])
+    circuit.append(val.on(*qubits[::2]))
+    state_vector = np.reshape(final_state_vector(circuit, qubit_order=qubits),
+                              protocols.qid_shape(qubits))
+
+    tableau = clifford_tableau.CliffordTableau(len(qubits))
+    for op in circuit.all_operations():
```
Strilanc, line 65: Extract this into a local private `final_tableau` method that we can later make globally accessible.

smitsanghavi: Done.

cirq/testing/consistent_act_on.py
smitsanghavi: Done.

cirq/testing/consistent_act_on.py
smitsanghavi: Thanks for the suggestion!

cirq/testing/consistent_act_on.py
smitsanghavi: Done

smitsanghavi:
> Add the "clifford matches" test to the `_assert_meets_standards_helper` method.

This also helped me find that Operations were not supported. Fixed that.

Strilanc:
Very nice!

PR #3064, opened 2020-06-03
Add Batched requests to Google's cirq messages

dstrain115:
- Adds BatchedProgram, BatchedRunContext, and BatchedResult
- Each is a repeated message so that circuits can be bundled
together as one request for increased efficiency.


cirq/google/api/v2/result.proto
```diff
@@ -16,6 +16,17 @@ message Result {
   repeated SweepResult sweep_results = 1;
 }
 
+// The result returned by a BatchedProgram
+message BatchedResult {
+
+  // The results of a BatchedProgram
+  // Each result in the message will directly correspond to a Program
+  // in the BatchedProgram request.  There will be one Result for each
```
dabacon, line 24: I think this should be slightly reworded, since there is a result per Program/RunContext pair.  I think it is important in the protos to convey the state of the resources, users may thinking about this as "run a program", but we should make it clear in the api that just creating a batched program does not ended up with results.

cirq/google/api/v2/program.proto
```diff
@@ -24,6 +24,14 @@ message Program {
   }
 }
 
+// A Batch of multiple circuits that should be run as one program
+message BatchedProgram {
```
maffoo, line 28: Nit: call this `BatchProgram` instead? Similarly for `BatchRunContext` and `BatchResult`.

cirq/google/api/v2/program.proto
```diff
@@ -24,6 +24,14 @@ message Program {
...
+
+  // The circuits that should be bundled togather as one program
```
maffoo, line 30: typo: "together"

cirq/google/api/v2/program.proto
```diff
@@ -24,6 +24,14 @@ message Program {
   }
 }
 
+// A Batch of multiple circuits that should be run as one program
```
maffoo, line 27: Might want to clarify here that "as one program" is at the level of the quantum engine api. These will not _necessarily_ be executed all at once on the hardware.

cirq/google/api/v2/run_context.proto
```diff
@@ -12,6 +12,19 @@ message RunContext {
   repeated ParameterSweep parameter_sweeps = 1;
 }
 
+// A batch of contexts for running a bundled batch of programs
+// To be used in conjunction with BatchedProgram
+message BatchedRunContext {
```
maffoo, line 17: You might consider moving all these into a new `batch.proto` file, so they are closer together and the relationship between them should be easier to see:
```python
# batch.proto

message BatchProgram { ... }
message BatchRunContext { ... }
message BatchResult { ... }
```

cirq/google/api/v2/result.proto
dstrain115: Reworded.  Take another look and let me know if it is more clear.  I wasn't sure if I understood your concerns correctly.

cirq/google/api/v2/program.proto
dstrain115: Done.

cirq/google/api/v2/program.proto
dstrain115: Done.

cirq/google/api/v2/program.proto
dstrain115: Done.

cirq/google/api/v2/result.proto
```diff
@@ -16,6 +16,17 @@ message Result {
   repeated SweepResult sweep_results = 1;
 }
 
+// The result returned by a BatchedProgram
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/api/v2/run_context.proto
dstrain115: Done.

cirq/google/api/v2/batch.proto
```diff
@@ -0,0 +1,51 @@
...
+  // The circuits that should be bundled together as one program
+  repeated Program programs = 1;
+}
+
+// A batch of contexts for running a bundled batch of programs
+// To be used in conjunction with BatchProgram
+message BatchRunContext {
+
+  // Run contexts for each program in the BatchProgram
+  // Each RunContext should map directly to a Program in the corresponding
```
maffoo, line 28: Nit: maybe add lines between paragraphs as you've done for BatchResult:
```
// Run contexts for ...
//
// Each RunContext should map ...
```

cirq/google/api/v2/batch.proto
```diff
@@ -0,0 +1,52 @@
...
+package cirq.google.api.v2;
+
+option java_package = "com.google.cirq.google.api.v2";
+option java_outer_classname = "BatchProto";
+option java_multiple_files = true;
+
+// A Batch of multiple circuits that should be run together
+// as one QuantumProgram within Quantum Engine.
+// (Note that circuits may not necessarily be executed
+//  simultaneously on hardware based on circumstances)
```
dabacon, line 16: nit: "executed as a batch" instead of "executed simultaneously"

dstrain115: I think the wording you proposed adds more confusion.  I rephrased again.

cirq/google/api/v2/batch.proto
```diff
@@ -0,0 +1,55 @@
...
+}
+
+
+// The result returned from running a BatchProgram/BatchRunContext
+message BatchResult {
+
+  // Results returned from executing a BatchProgram/BatchRunContext pair.
+  //
+  // After a BatchProgram and BatchRunContext is successfully run in
+  // Quantum Engine, the expected result if successful will be a BatchResult.
```
wcourtney, line 47: Must success be for _every_ program/context combo, or can this include partial results?

cirq/google/api/v2/batch.proto
```diff
@@ -0,0 +1,55 @@
...
+
+// A batch of contexts for running a bundled batch of programs
+// To be used in conjunction with BatchProgram
+message BatchRunContext {
+
+  // Run contexts for each program in the BatchProgram
+  // Each RunContext should map directly to a Program in the corresponding
+  // BatchProgram.
+  //
+  // This message should have one RunContext for each Program in the
```
wcourtney, line 34: "should" should be "must" when describing the relationship between run contexts and programs.

cirq/google/api/v2/batch.proto
dstrain115: I suspect we would not initially support partial result, but put in a paragraph on how it would be supported in the future.

cirq/google/api/v2/batch.proto
dstrain115: Done.

PR #3072, opened 2020-06-04
Make LinearDict json serializable

dabacon:
* Fails for linear dicts that have a validator (since this is a function).
* Changes default to None from an explicit lambda. Behavior of not specifying validator is exactly the same.

My first time adding json protocol, let me know if I'm not following good practices.


cirq/protocols/json_test_data/LinearDict.repr
```diff
@@ -0,0 +1 @@
+[cirq.LinearDict({}), cirq.LinearDict({'a': 1}), cirq.LinearDict({'a': 1, 'b': 2}), cirq.LinearDict({cirq.X: 1.0, cirq.Y: 2})]
```
mpharrigan, line 1: Why doesn't this cause the repr roundtrip test to fail? As I understand it this is the repr for a list of LinearDict rather than a LinearDict

dabacon: The test allows you to write a list of reprs.  See line 488 in json_serialization_test.py

mpharrigan: TIL. Thanks!

PR #3075, opened 2020-06-05
Add Circuit.zip method

Strilanc:
Handy to have around when building circuits up in tiled pieces and wanting to guarantee the moment structure comes out right.


cirq/circuits/circuit_test.py
```diff
@@ -3764,3 +3764,56 @@ def test_deprecated():
...
+    assert cirq.Circuit.zip() == cirq.Circuit()
+    assert cirq.Circuit.zip(cirq.Circuit()) == cirq.Circuit()
+    assert cirq.Circuit().zip(cirq.Circuit()) == cirq.Circuit()
+    assert circuit1.zip(cirq.Circuit()) == circuit1
+    assert cirq.Circuit(cirq.Moment()).zip(cirq.Circuit()) == cirq.Circuit(
+        cirq.Moment())
+    assert cirq.Circuit().zip(cirq.Circuit(cirq.Moment())) == cirq.Circuit(
+        cirq.Moment())
+
+    assert circuit1.zip(circuit2) == circuit2.zip(circuit1) == cirq.Circuit(
```
dstrain115, line 3789: Do you want to have a line comment above this and below here, since there is a heading above for empty cases, but it is not visually clear where the section of "Empty cases" ends.

Strilanc: Done.

cirq/circuits/circuit.py
```diff
@@ -1235,6 +1235,73 @@ def _insert_operations(self, operations: Sequence['cirq.Operation'],
...
+                The zipped circuits have overlapping operations occurring at the
+                same moment index.
+
+        Examples:
+            >>> import cirq
+            >>> a, b, c, d = cirq.LineQubit.range(4)
+            >>> circuit1 = cirq.Circuit(cirq.H(a), cirq.CNOT(a, b))
+            >>> circuit2 = cirq.Circuit(cirq.X(c), cirq.Y(c), cirq.Z(c))
+            >>> circuit3 = cirq.Circuit(cirq.Moment(), cirq.Moment(cirq.S(d)))
+            >>> print(circuit1.zip(circuit2))
```
dabacon, line 1269: Having this as an instance method feels strange because this isn't how it is done in python?

cduck: It's a bit unusual but makes sense to me.  Because of how Python binds methods, both `Circuit.merge(c1, c2)` and `c1.merge(c2)` will work.

dabacon: I just start getting scared when we hit line 1269 in a single class definition :)

Issue #2984, opened 2020-05-08
Virtual tag does not feel like correct tag for noise


dabacon:
From the doc string:
>Operations marked with this tag are presumed to have zero duration of their own, although they may have a non-zero duration if run in the same Moment as a non-virtual operation.

This is very confusing because noise certainly has "duration".  Of course we only support "noise" in simulation (well on the quantum computers too haha).

Do we have an example of a noise composition that is useful?  Right now noise composition feels dangerous since noise does not commute and the new model is highly dependent on this ordering.

dabacon:
Discussed at Cirq Cync.  Decision was to remove automatic tagging but allow optional tagging for noise models.

dstrain115:
@dabacon Can we close this issue?  It does not seem very relevant anymore.

dabacon:
I believe noise models still do this, we should convert that to optional.

dabacon:
Thinking more about this, I actually think the virtual tag is OK for the noise models.  Submitted a PR to clarify the tag docstring which I think will close this.

PR #3076, opened 2020-06-05
[SVG] Noise hack and font fix

mpharrigan:
- Specify the font. This should not be a change for viewing in Jupyter notebook but fixes opening the svg data in e.g. inkscape where the default font is different and the boxes aren't the right size
 - Don't display virtual tags (for noise) xref #2905

### DIFF
--- a/cirq/contrib/svg/svg.py
+++ b/cirq/contrib/svg/svg.py
@@ -6,12 +6,22 @@ if TYPE_CHECKING:
     import cirq
 
 QBLUE = '#1967d2'
+FONT = "Arial"
+
+
+def fixup_text(text: str):
+    if '\n' in text:
+        return '?'
+    if '[<virtual>]' in text:
+        # https://github.com/quantumlib/Cirq/issues/2905
+        # TODO: escape angle brackets when you actually want to display tags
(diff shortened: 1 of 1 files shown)


cirq/contrib/svg/svg.py
```diff
@@ -6,12 +6,25 @@
     import cirq
 
 QBLUE = '#1967d2'
+FONT = "Arial"
+
+
+def fixup_text(text: str):
+    if '\n' in text:
+        return '?'
+    if '[<virtual>]' in text:
```
Strilanc, line 15: This seems awfully specific.

mpharrigan: We can wait for https://github.com/quantumlib/Cirq/issues/2984#issuecomment-638427120 and then it will be unnecessary but I'm using this patch currently

dabacon: This is contrib, so I'd just make sure we have an issue to fix this before moving to cirq, and maybe a link from that issue.

mpharrigan: there's a comment linking to the issue; I'll put a backref on the issue itself

PR #3077, opened 2020-06-06
Fix Quirk import extra gates bug

matpompili:
When more than one extra gate is provided only the last one was used for all of them.

Example: 
```Python
extra_gates = {
    "Foo": cirq.rx(np.pi/2),
    "Bar": cirq.ry(np.pi/2),
}

cirq.quirk_url_to_circuit(
    'http://algassert.com/quirk#circuit={"cols":[["Foo"], ["Bar"]]}', 
    extra_cell_makers=extra_gates)

0: ───Ry(0.5π)───Ry(0.5π)───
```

This is because the the extra_makers are defined as list comprehension, so only the last value is used for all the lambdas.
More details here: [On Python's lambda and closures](https://louisabraham.github.io/articles/python-lambda-closures.html)


matpompili:
@googlebot I signed it!

matpompili:
Should be done!

Issue #3083, opened 2020-06-09
random_circuit_test is flakey


dabacon:
See pytest ubuntu here: https://github.com/quantumlib/Cirq/runs/755627222?check_suite_focus=true

PR #3084, opened 2020-06-09
Fix flakes in random_circuit_test

dabacon:
Sometimes test data (about 1 in 100) had no gates that act on few enough qubits to fit with number of qubits.  

* Cleans up the generating function which was a bit of a mess to read. For tests prefer explicitness over conciseness.
* Fixes missing `f` in error and improves test case for that error.


dabacon:
Fixes  #3083

PR #3085, opened 2020-06-10
Docs site setup

lamberta:
Set up docs/ directory for site. Will work on `docs_site` branch.

Create `/dev_tools/docs/` for doc tooling.
Move Sphinx config/assets to `dev_tools/docs/sphinx/`
Remove `build_docs` CI test.
Rebased with master branch.

Context: b/158584984


dstrain115:
It looks like this breaks a bunch of the tests.  Can we fix that?  
Formatting changes should be easy to fix (you can use check/format.py or fix them manually)
Looks like docs/docs_coverage_test.py changed locations, so the tests need to reflect that.  Probably we should disable the build docs test on this branch as well.

lamberta:
@dstrain115 Ok, updated to pass CI tests, PTAL (and apologies for the incremental noise from this PR)

Moved files but did not delete---and updated any references.
Removed the `build_docs` CI test since I was running into config problems and this won't be relevant anyway.

lamberta:
After rebasing with https://github.com/quantumlib/Cirq/pull/3089 the CLA bot is picking up some commits can't verify from me. Looks like they're all Googlers and this should be fine.

@dstrain115 @MichaelBroughton Can you double-check or override? (And merge this since I can't.) Thanks!

PR #3093, opened 2020-06-16
Fix cirq.Simulator assuming a reset is safe to cache before sampling

Strilanc:
- This resulted in some circuits with random behavior giving long lists of identical samples when repeated
- Change the general strategy used by the simulator to one of extracting a unitary prefix that can be cached, and then looking at the rest to decide whether sampling can be done efficiently or not


cirq/sim/sparse_simulator.py
```diff
@@ -134,34 +135,39 @@ def _run(self, circuit: circuits.Circuit,
...
-
-    def _run_sweep_repeat(self, circuit: circuits.Circuit,
+        assert step_result is not None
+
+        # When an otherwise unitary circuit ends with non-demolition computation
+        # basis measurements, we can sample the results more efficiently.
+        general_ops = list(general_suffix.all_operations())
+        if all(isinstance(op.gate, ops.MeasurementGate) for op in general_ops):
+            return step_result.sample_measurement_ops(measurement_ops=cast(
+                List[ops.GateOperation], general_ops),
```
dabacon, line 157: autoformat is on drugs

cirq/sim/sparse_simulator.py
```diff
@@ -331,3 +336,27 @@ def sample(self,
                                                     self, None),
                                                 repetitions=repetitions,
                                                 seed=seed)
+
+
+def _split_into_unitary_then_general(circuit: 'cirq.Circuit'
```
dabacon, line 341: There is probably a case to be made for a good way to handle this sort of splitting of circuits by disjoint filter as part of Cirq.  No need to do now but maybe we should have an issue for this?

cirq/sim/sparse_simulator.py
```diff
@@ -134,34 +135,39 @@ def _run(self, circuit: circuits.Circuit,
...
-
-    def _run_sweep_terminal_sample(self, circuit: circuits.Circuit,
-                                   repetitions: int) -> Dict[str, np.ndarray]:
-        for step_result in self._base_iterator(
-                circuit=circuit,
-                qubit_order=ops.QubitOrder.DEFAULT,
-                initial_state=0,
-                perform_measurements=False):
+        qubit_order = sorted(resolved_circuit.all_qubits())
+
```
dabacon, line 139: just to make sure I understand: the bug was that intermediate channel ops that could be monte carlo sampled were allowed to not be terminal, and the new way this works it splits into the biggest possible unitary chunk and if then "others" to figure out whether this is possible.

cirq/sim/sparse_simulator.py
```diff
@@ -134,34 +135,39 @@ def _run(self, circuit: circuits.Circuit,
...
+        intermediate_state = step_result.state_vector().reshape(qid_shape)
+        return self._run_sweep_repeat(initial_state=intermediate_state,
+                                      circuit=general_suffix,
+                                      repetitions=repetitions,
+                                      qubit_order=qubit_order)
+
+    def _run_sweep_repeat(self, initial_state: np.ndarray,
+                          circuit: circuits.Circuit,
+                          qubit_order: 'cirq.QubitOrderOrList',
                           repetitions: int) -> Dict[str, np.ndarray]:
```
dstrain115, line 171: Is it worth adding a developer-centric docstring?  It's a private method, so probably not mandatory, but might be nice to have.

cirq/sim/sparse_simulator.py
```diff
@@ -331,3 +336,27 @@ def sample(self,
                                                     self, None),
                                                 repetitions=repetitions,
                                                 seed=seed)
+
+
+def _split_into_unitary_then_general(circuit: 'cirq.Circuit'
+                                    ) -> Tuple['cirq.Circuit', 'cirq.Circuit']:
```
dstrain115, line 342: Ditto.  I would either expand on the comment above "Simulate as many unitary operations as possible..." or write a docstring to explain what this is doing and returning.

cirq/sim/sparse_simulator_test.py
```diff
@@ -147,7 +147,7 @@ def test_run_invert_mask_measure_not_terminal(dtype):
                 np.testing.assert_equal(result.measurements,
                                         {'m': [[1 - b0, b1]] * 3})
                 assert result.repetitions == 3
-        assert mock_sim.call_count == 12
+        assert mock_sim.call_count > 4
```
dstrain115, line 150: For this and below, would it be worth a line comment to explain what the assertion is testing and/or why ">4" is expected?

cirq/sim/sparse_simulator_test.py
```diff
@@ -900,3 +899,44 @@ def test_random_seed_mixture_deterministic():
                   [[1], [0], [0], [0], [1], [0], [0], [1], [1], [1], [1], [1],
                    [0], [1], [0], [0], [0], [0], [0], [1], [0], [1], [1], [0],
                    [1], [1], [1], [1], [1], [0]])
+
+
+def test_entangled_reset_does_not_break_randomness():
+    # A previous version made the mistake of assuming that it was okay to cache
```
dstrain115, line 905: Do you think this should use the triple quote doc string instead of the line comments?

cirq/sim/sparse_simulator.py
Strilanc: Yes, that was the issue.

cirq/sim/sparse_simulator.py
Strilanc: Acknowledged.

cirq/sim/sparse_simulator.py
Strilanc: We cannot question it, we can only follow.

cirq/sim/sparse_simulator.py
Strilanc: Acknowledged. In this case I think the method name is pretty descriptive.

cirq/sim/sparse_simulator_test.py
Strilanc: Huh. I never considered putting a doc string on a test method.

cirq/sim/simulator.py
```diff
@@ -471,29 +463,43 @@ def sample_measurement_ops(self,
...
+            if key in seen_measurement_keys:
+                raise ValueError(f'Duplicate MeasurementGate with key {key}')
+            seen_measurement_keys.add(key)
+
+        # Find measured qubits, ensuring a consistent ordering.
+        measured_qubits = []
+        seen_qubits: Set[cirq.Qid] = set()
+        for op in measurement_ops:
+            for q in op.qubits:
+                if q not in seen_qubits:
```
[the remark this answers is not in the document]
Strilanc: This is what I initially did, and it resulted in several fixed-seed tests being flaky. The problem is it results in a non-deterministic ordering because `set` doesn't guarantee insertion order ordering.

cirq/sim/sparse_simulator.py
Strilanc: Done. Also renamed it.

cirq/sim/sparse_simulator.py
Strilanc: Added a docstring.

cirq/sim/sparse_simulator_test.py
Strilanc: Added comments

cirq/sim/sparse_simulator_test.py
Strilanc: Made it a doc string.

PR #3096, opened 2020-06-18
Implement unitary protocol for linear combinations

viathor:


cirq/ops/linear_combinations.py
```diff
@@ -351,6 +387,36 @@ def copy(self) -> 'PauliSum':
...
+        result = np.zeros((2,) * (2 * num_qubits), dtype=np.complex128)
+        for vec, coeff in self._linear_dict.items():
+            op = _pauli_string_from_unit(vec)
+            identity = np.eye(num_dim,
+                              dtype=np.complex128).reshape(result.shape)
+            workspace = np.empty_like(identity)
+            axes = tuple(qubit_to_axis[q] for q in op.qubits)
+            u = protocols.apply_unitary(
+                op, protocols.ApplyUnitaryArgs(identity, workspace, axes))
+            result += coeff * u
```
mpharrigan, line 408: shouldn't the body of this loop be factored into a method on `PauliString`? Is the problem tensoring in the right number of identities to each PauliString? Presumably the `PauliString.matrix()` method should take a `qubits` argument that (a) sets the order of the qubits and (b) will tensor in appropriate identities if the string doesn't have explicit support on a given input qubit

viathor: Good suggestion. Done.

Also, filed #3099 to make `matrix` more official and useful.

viathor:
PTAL

cirq/ops/pauli_string.py
```diff
@@ -171,7 +171,7 @@ def __getitem__(self, key: 'cirq.Qid') -> pauli_gates.Pauli:
 
     # pylint: disable=function-redefined
     @overload
-    def get(self, key: 'cirq.Qid') -> pauli_gates.Pauli:
+    def get(self, key: 'cirq.Qid', default: None = None) -> pauli_gates.Pauli:
```
mpharrigan, line 174: Can you explain this change? I assume it's for the type checking? How is it different than the final definition
```python
def get(self, key: 'cirq.Qid', default=None):
```

cirq/ops/pauli_string.py
```diff
@@ -344,14 +344,28 @@ def __str__(self) -> str:
 
         return prefix + '*'.join(factors)
 
+    def matrix(self,
+               qubits: Optional[Iterable[raw_types.Qid]] = None) -> np.ndarray:
```
mpharrigan, line 348: elsewhere we use the `QubitOrdering` class to have a more type-safe sentinel value for "default". That might be overkill here and/or not in the same spirit as this argument; just wanted to make a note of it

cirq/ops/pauli_string.py
viathor: Type-checking succeeds either way, but without explicit specification of the `default` parameter in one of the overloads pylint complains when we try to use it.

cirq/ops/pauli_string.py
viathor: In general, a `QubitOrder` describes a relation in a potentially unbounded set of qubits (see description of the second argument to `QubitOrder.explicit`), but we want the user to give us both: the set of qubits and their order, so this isn't a good fit.

PR #3107, opened 2020-06-24
Fix some test incompatibilities with latest numpy

Strilanc:
- Also fix ZPowGate and PhasedXPowGate not supporting approximate equality
- Also fix some tests using deprecated methods

### DIFF
--- a/cirq/ops/common_gates.py
+++ b/cirq/ops/common_gates.py
@@ -377,7 +377,7 @@ class YPowGate(eigen_gate.EigenGate,
         ).format(proper_repr(self._exponent), self._global_shift)
 
 
-@value.value_equality(approximate=True)
+@value.value_equality
 class ZPowGate(eigen_gate.EigenGate,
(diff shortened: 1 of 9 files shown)


cirq/ops/common_gates.py
```diff
@@ -377,7 +377,7 @@ def __repr__(self) -> str:
         ).format(proper_repr(self._exponent), self._global_shift)
 
 
-@value.value_equality
```
dabacon, line 380: so the issue here is that this override eigen_gate's decorator?  Do we need to do this for XPowGate and YPowGate?

Strilanc: Actually.. turns out the test I added passed anyways. Removed this change.

dabacon: OK, I think there is still a bug here from one of the cases I found, but will follow up separately.

dabacon:
api test doesn't like you :)

Issue #3101, opened 2020-06-23
Density Matrix Simulator seems to mutate initial states sometimes


trevormccrt:
Found this while running density matrix simulations in a loop passing in the same python object as initial state each time. Fails input density matrix validity check on the second iteration. not sure if bug or hack for higher performance. I remember pointing out something similar in the wavefunction simulator to @dabacon  late last year and I think it was deemed a bug. Thoughts in this case?

dabacon:
Yeah I think initial states should be copied.  There is a similar problem in that the state is reused in the moment by moment iterator: 3100

dabacon:
@trevormccrt can you give an example of this failing?  Is it the initial state that is being mutated?

trevormccrt:
Yes. What i did to produce this error was very simple. start with an initial state density matrix of type np.complex64. Then in a for loop run various circuits starting from the same initial state. I found that on the second iteration the initial state object had been mutated. This would lead to a "density matrix is not hermetian" error at the start of the second iteration during input checking. Basically just the following snippet in a loop for various values of circuit:

` actual_final = cirq.DensityMatrixSimulator().simulate(
        circuit, initial_state=qubit_rho_zero,
        qubit_order=qubits).final_density_matrix`

Wrapping qubit_rho_zero in np.copy solved the issue for me.

Issue #3100, opened 2020-06-19
simulate_moment_steps return invalid density matrices


CirqUser199:
When I try to simulate the density matrix through a circuit using simulate_moment_steps, I have some density matrices that have a trace different than 1 for some steps. For instance, when I execute the simple code below, the last line returns 1/2. If I did no mistake, this means that the simulator returns a density matrix that has trace less than 1. From what I can see, this could be caused by the insert strategy (maybe it interferes with the moment by moment simulator). Did you already hear about such an issue?

```
import cirq
import numpy as np
from cirq.circuits import InsertStrategy
q0, q1, q2, q3, q4, q5, q6 = cirq.LineQubit.range(7)

example_circuit=cirq.Circuit()
example_circuit.append([cirq.rx(3*np.pi/7)(q0),cirq.rx(3*np.pi/5)(q1),cirq.rx(np.pi/4)(q2)])
example_circuit.append([cirq.ry(1*np.pi/5)(q0),cirq.ry(3*np.pi/2)(q1),cirq.ry(2*np.pi/5)(q2)])
example_circuit.append([cirq.H(q3),cirq.H(q4)], strategy=InsertStrategy.NEW_THEN_INLINE)
example_circuit.append([cirq.CNOT(q3,q5),cirq.CNOT(q4,q6)], strategy=InsertStrategy.NEW_THEN_INLINE)

sim = cirq.DensityMatrixSimulator()
example_densities=[]

for i, step in enumerate(sim.simulate_moment_steps(example_circuit)):
    example_densities.append(step.density_matrix())
    
np.trace(example_densities[2])
```

cduck:
It appears that the simulator is re-using and modifying the matrix returned by `.density_matrix()` during later steps.  This was probably intended behavior to improve speed but can cause strange usability issues like you found.  If others have done this I would consider it a bug.

Try this instead.  You get the correct result if you use the matrix before the next simulation step.
```python
for i, step in enumerate(sim.simulate_moment_steps(example_circuit)):
    print(i, np.trace(step.density_matrix()))
```
```
0 (1+0j)
1 (1-1.1175871e-08j)
2 (1-1.1175869e-08j)
3 (1-1.117587e-08j)
```

If you do need to store all the density matrices (I would not recommend this for circuits much larger than your example), just make copies of the matrix:
```python
    example_densities.append(np.copy(step.density_matrix()))
```

CirqUser199:
Thanks a lot for your quick answer!

[...]

PR #3109, opened 2020-06-24
Defensively copy density matrices in simulator

dabacon:
This makes it so that by default when you access the density matrix during simulation, it makes a copy of the matrix.  This can be overridden if one wants to get the speed benefits of not copying.

### DIFF
--- a/cirq/sim/density_matrix_simulator.py
+++ b/cirq/sim/density_matrix_simulator.py
@@ -256,8 +256,9 @@ class DensityMatrixSimulator(simulator.SimulatesSamples,
         initial_matrix = qis.to_valid_density_matrix(initial_state,
                                                      len(qid_shape),
                                                      qid_shape=qid_shape,
-                                                     dtype=self._dtype,
-                                                     copy=True)
+                                                     dtype=self._dtype)
+        if np.may_share_memory(initial_matrix, initial_state):
+            initial_matrix = initial_matrix.copy()
         measured = collections.defaultdict(
             bool)  # type: Dict[Tuple[cirq.Qid, ...], bool]
         if len(circuit) == 0:
@@ -438,10 +439,10 @@ class DensityMatrixStepResult(simulator.StepResult):
 
         Args:
             copy: If True, then the returned state is a copy of the density
-            matrix. If False, then the density matrix is not copied, potentially
-            saving memory. If one only needs to read derived parameters from the
-            density matrix and store then using False can speed up simulation by
-            eliminating a memory copy.
+                matrix. If False, then the density matrix is not copied,
+                potentially saving memory. If one only needs to read derived
+                parameters from the density matrix and store then using False
+                can speed up simulation by eliminating a memory copy.
         """
         size = np.prod(self._qid_shape, dtype=int)
         matrix = self._density_matrix.copy() if copy else self._density_matrix
@@ -536,7 +537,7 @@ class DensityMatrixTrialResult(simulator.SimulationTrialResult):
                          final_simulator_state=final_simulator_state)
         size = np.prod(protocols.qid_shape(self), dtype=int)
         self.final_density_matrix = np.reshape(
-            final_simulator_state.density_matrix, (size, size)).copy()
+            final_simulator_state.density_matrix.copy(), (size, size))
 
     def _value_equality_values_(self) -> Any:
         measurements = {
--- a/cirq/sim/density_matrix_simulator.py
+++ b/cirq/sim/density_matrix_simulator.py
@@ -256,7 +256,7 @@ class DensityMatrixSimulator(simulator.SimulatesSamples,
         initial_matrix = qis.to_valid_density_matrix(initial_state,
                                                      len(qid_shape),
                                                      qid_shape=qid_shape,
-                                                     dtype=self._dtype)
+                                                     dtype=self._dtype).copy()
         measured = collections.defaultdict(
             bool)  # type: Dict[Tuple[cirq.Qid, ...], bool]
         if len(circuit) == 0:
@@ -407,7 +407,7 @@ class DensityMatrixStepResult(simulator.StepResult):
         density_matrix = np.reshape(density_matrix, sim_state_matrix.shape)
         np.copyto(dst=sim_state_matrix, src=density_matrix)
 
-    def density_matrix(self):
+    def density_matrix(self, copy=True):
         """Returns the density matrix at this step in the simulation.
 
         The density matrix that is stored in this result is returned in the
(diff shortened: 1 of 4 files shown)


dabacon:
Fixed #3101 and #3100

cirq/sim/density_matrix_simulator.py
```diff
@@ -256,7 +256,7 @@ def _base_iterator(self,
         initial_matrix = qis.to_valid_density_matrix(initial_state,
                                                      len(qid_shape),
                                                      qid_shape=qid_shape,
-                                                     dtype=self._dtype)
+                                                     dtype=self._dtype).copy()
```
cduck, line 259: This will make an unnecessary copy twice if to_valid_density_matrix already made a copy to change the dtype.

95-martin-orion: It looks like `to_valid_density_matrix` only changes the dtype if `initial_state` is an int, in which case a double-copy isn't particularly expensive. For matrices, a mismatched dtype results in an error instead of copying the matrix.

I think this `copy` should be OK.

cirq/sim/density_matrix_simulator.py
```diff
@@ -528,7 +535,7 @@ def __init__(self, params: study.ParamResolver,
                          final_simulator_state=final_simulator_state)
         size = np.prod(protocols.qid_shape(self), dtype=int)
         self.final_density_matrix = np.reshape(
-            final_simulator_state.density_matrix, (size, size))
+            final_simulator_state.density_matrix, (size, size)).copy()
```
95-martin-orion, line 538: IIUC, this state is only created once the simulation is complete, so it should be safe to not copy in this situation - no further mutation is possible.

dabacon: Yes but there are two ways to get this, one from the density_matrix on the last StepResult and also here.  One could manipulate the former and without this copy it could appear here on the final.  I think this would be odd because you'd have to opt into the no copy behavior for the StepResult, but I wanted to insure that never happens.

95-martin-orion: Ahh, that makes sense. SGTM.

cirq/sim/density_matrix_simulator.py
cduck: I think there are other cases too, and it could change in the future.  An easy solution would be to add
```python
if np.may_share_memory(initial_matrix, initial_state):
    initial_matrix = initial_matrix.copy()

File: cirq/sim/density_matrix_simulator.py [hunk 3]
```diff
@@ -435,9 +435,16 @@ def density_matrix(self):
                 |  6  |   1    |   1    |   0    |
                 |  7  |   1    |   1    |   1    |
 
+        Args:
+            copy: If True, then the returned state is a copy of the density
+            matrix. If False, then the density matrix is not copied, potentially
+            saving memory. If one only needs to read derived parameters from the
+            density matrix and store then using False can speed up simulation by
+            eliminating a memory copy.
```
cduck, line 444: Nit: Do these need one more indent?

cirq/sim/density_matrix_simulator.py
dabacon: I move the copy as an optional parameter in `to_valid_density_matrix` to handle this.

cirq/sim/density_matrix_simulator.py
```diff
@@ -435,9 +436,16 @@ def density_matrix(self):
...
         """
         size = np.prod(self._qid_shape, dtype=int)
-        return np.reshape(self._density_matrix, (size, size))
+        matrix = self._density_matrix.copy() if copy else self._density_matrix
+        return np.reshape(matrix, (size, size))
```
cduck, line 448: Reshape can also return a copy but this is is probably too unlikely to worry about.

cirq/sim/density_matrix_simulator.py
```diff
@@ -528,7 +536,7 @@ def __init__(self, params: study.ParamResolver,
                          final_simulator_state=final_simulator_state)
         size = np.prod(protocols.qid_shape(self), dtype=int)
         self.final_density_matrix = np.reshape(
-            final_simulator_state.density_matrix, (size, size))
+            final_simulator_state.density_matrix, (size, size)).copy()
```
cduck, line 539: Nit: I would switch the order of copy and reshape to reduce the chance of reshape also making a copy.

cirq/sim/density_matrix_simulator_test.py
```diff
@@ -1112,3 +1112,36 @@ def test_simulate_noise_with_terminal_measurements():
...
+def test_density_matrix_copy():
+    sim = cirq.DensityMatrixSimulator()
+
+    q = cirq.LineQubit(0)
+    circuit = cirq.Circuit(cirq.H(q), cirq.H(q))
+
+    matrices = []
+    for step in sim.simulate_moment_steps(circuit):
+        matrices.append(step.density_matrix(copy=True))
+    assert all(np.isclose(np.trace(x), 1.0) for x in matrices)
```
cduck, line 1126: Just testing the trace might not catch all non-copying conditions.  You could save each matrix and a copy then check they are all pairwise equal afterward.  Or you could check np.may_share_memory/np.shares_memory over all pairs.

cirq/sim/density_matrix_simulator_test.py
```diff
@@ -1112,3 +1112,36 @@ def test_simulate_noise_with_terminal_measurements():
...
+
+
+def test_final_density_matrix_is_not_last_object():
+    sim = cirq.DensityMatrixSimulator()
+
+    q = cirq.LineQubit(0)
+    initial_state = np.array([[1, 0], [0, 0]], dtype=np.complex64)
+    circuit = cirq.Circuit(cirq.WaitGate(0)(q))
+    result = sim.simulate(circuit, initial_state=initial_state)
+    assert result.final_density_matrix is not initial_state
```
cduck, line 1146: This won't catch aliased matrices.  Also use np.may_share_memory/np.shares_memory.
```python
>>> matrix.reshape((...)) is not matrix.reshape((...))
True
```

cirq/sim/density_matrix_simulator.py
dabacon: done

cirq/sim/density_matrix_simulator.py
dabacon: done

cirq/sim/density_matrix_simulator_test.py
dabacon: done

cirq/sim/density_matrix_simulator_test.py
dabacon: done

cirq/sim/density_matrix_simulator_test.py
```diff
@@ -1134,6 +1136,8 @@ def test_density_matrix_copy():
         traces.append(np.trace(step.density_matrix(copy=False)))
     assert any(not np.isclose(np.trace(x), 1.0) for x in matrices)
     assert all(np.isclose(x, 1.0) for x in traces)
+    assert any(not np.shares_memory(x, y) for x, y in
```
cduck: You mean `all(not` or `not any(`?

dabacon: fixed

dabacon:
@clabot this should be fine

PR #3112, opened 2020-06-25
Change Operation.with_qubits return type to match self type

maffoo:
Also change the return type of `transform_qubits` in the same way. AFAICT this should always be correct, but it's sometimes tricky to convince mypy of this (have to cast in `GateOperation.with_qubits`, for example).

### DIFF
--- a/cirq/ops/gate_operation.py
+++ b/cirq/ops/gate_operation.py
@@ -15,8 +15,8 @@
 """Basic types defining qubits, gates, and operations."""
 
 import re
-from typing import (Any, Dict, FrozenSet, Iterable, List, Optional, Sequence,
-                    Tuple, TypeVar, Union, TYPE_CHECKING)
+from typing import (Any, cast, Dict, FrozenSet, Iterable, List, Optional,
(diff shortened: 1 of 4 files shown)


cirq/ops/gate_operation.py
```diff
@@ -52,8 +55,8 @@ def qubits(self) -> Tuple['cirq.Qid', ...]:
         """The qubits targeted by the operation."""
         return self._qubits
 
-    def with_qubits(self, *new_qubits: 'cirq.Qid') -> 'cirq.Operation':
-        return self.gate.on(*new_qubits)
+    def with_qubits(self: TSelf, *new_qubits: 'cirq.Qid') -> TSelf:
+        return cast(TSelf, self.gate.on(*new_qubits))
```
viathor, line 59: Optional: Have you tried making `GateOperation` parametrized with the type of gate it wraps? I think this should let us remove the cast.

maffoo: I didn't try that, but my experience with "genericizing" classes in python is that it is not trivial, especially for something this widely used. The relationship between gates and operations is also complicated by various specialized operation subclasses that we have.

PR #3113, opened 2020-06-26
Only override control when global phase is zero

### LINKED ISSUE #3111: Gate inconsistencies between versions when using controlled interface
It seems there is a (unintentional?) change to what the controlled interface does before and after the `8179aa2d678bcce480c0ce14bd2852068657abf6` commit in certain cases, possibly related to how the global phase is handled.

Running the following circuit (and inspecting the states at each moment):
```
q0 = cirq.GridQubit(0, 0)
q1 = cirq.GridQubit(0, 1)
q2 = cirq.GridQubit(0, 2)
circuit = cirq.Circuit()
circuit.append(cirq.X(q0))
circuit.append(cirq.rx(1).controlled()(q0,q1))
circuit.append(cirq.rx(1).controlled()(q1,q2))
simulator = cirq.Simulator()
for i, step in enumerate(simulator.simulate_moment_steps(circuit))
    temp = np.around(step.state_vector(), 3)
    print('state at step %d: %s %s' % (i, temp[temp != 0], temp.nonzero()))
```

gives the following output before the `8179aa2d678bcce480c0ce14bd2852068657abf6` commit
```
state at step 0: [1.+0.j] (array([4]),)
state at step 1: [0.878+0.j    0.   -0.479j] (array([4, 6]),)
state at step 2: [ 0.878+0.j     0.   -0.421j -0.23 +0.j   ] (array([4, 6, 7]),)
```
and this output after:
```
state at step 0: [1.+0.j] (array([4]),)
state at step 1: [0.878+0.j    0.   -0.479j] (array([4, 6]),)
state at step 2: [ 0.77-0.421j  0.  -0.421j -0.23+0.j   ] (array([4, 6, 7]),)
```
The output I would expect to see is the former one, as it matches other implementations of the same circuit.

dabacon:
Fixes #3111 

Presently the control override on XPowGate results in copying over the global phase shift of the XPowGate to the CXPowGate (similar for other control constuctions).  This is inconsistent with most people's notion of "control" where the identity component has no phase. 

This makes this override occur only when the global phase is zero.

### DIFF
--- a/cirq/ops/common_gates.py
+++ b/cirq/ops/common_gates.py
@@ -136,11 +136,16 @@ class XPowGate(eigen_gate.EigenGate,
         a generic qudit) and where the control is satisfied by the qubit being
         ON, as opposed to OFF.
 
+        Note that this only transforms into a CXPowGate (or controlled version
+        of that gate) if the global shift of the XPowGate is 0, otherwise
+        it produces a normal ControlledGate.
+
         (Note that a CXPowGate is, by definition, a controlled-XPowGate.)
         """
         result = super().controlled(num_controls, control_values,
                                     control_qid_shape)
-        if (isinstance(result, controlled_gate.ControlledGate) and
+        if (self._global_shift == 0 and
+                isinstance(result, controlled_gate.ControlledGate) and
                 result.control_values[-1] == (1,) and
                 result.control_qid_shape[-1] == 2):
             return cirq.CXPowGate(exponent=self._exponent,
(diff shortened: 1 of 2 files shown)


cirq/ops/common_gates.py
```diff
@@ -454,11 +459,16 @@ def controlled(self,
         a generic qudit) and where the control is satisfied by the qubit being
         ON, as opposed to OFF.
 
+        Note that this only transforms into a CZPowGate (or controlled version
+        of that gate) if the global shift of the ZPowGate is 0, otherwise
+        it produces a normal ControlledGate.
+
```
smitsanghavi, line 465: Looks like the above two paragraphs can actually be combined more concisely into a single paragraph listing the preconditions for the specialization to occur.

dabacon: Combined for this and all other gates.

smitsanghavi: I was suggesting more along the lines of:
```
This method will specialize the control only if:
1. global shift is 0
2. the last of the controls is a default-type control qubit (followed by the clarifying what that means)
```
It would make the prerequisites clearer to the reader. But it's a minor documentation nit and not worth blocking the PR.

PR #3114, opened 2020-06-29
Add batch support to cirq Engine client

dstrain115:
- Adds a run_batch() call to engine that can handle
lists of circuits and lists of programs.
- Creates BatchProgram and BatchRunContext objects.
- Pipes the objects through engine program and engine job.


cirq/google/engine/engine.py
```diff
@@ -316,6 +385,52 @@ def create_program(
...
+        """
+        gate_set = gate_set or gate_sets.XMON
+
+        if not program_id:
+            program_id = _make_random_id('prog-')
+
+        code = qtypes.any_pb2.Any()
+        batch = v2.batch_pb2.BatchProgram()
+        for program in programs:
+            batch.programs.append(gate_set.serialize(program))
```
maffoo, line 421: One thing you can do here is to instead serialize directly into the existing message:
```python
for program in programs:
    gate_set.serialize(program, msg=batch.programs.add())
```

cirq/google/engine/engine_job.py
```diff
@@ -285,6 +285,10 @@ def results(self) -> List[study.TrialResult]:
                   result_type == 'cirq.api.google.v2.Result'):
                 v2_parsed_result = v2.result_pb2.Result.FromString(result.value)
                 self._results = self._get_job_results_v2(v2_parsed_result)
+            elif result_type == 'cirq.google.api.v2.BatchResult':
```
maffoo, line 288: Could use the built-in Any mechanism for checking types here:
```python
elif result.Is(v2.batch_pb2.BatchResult.DESCRIPTOR):
    ...
```
We could change other cases as well though we'd have to leave in code to handle old `cirq.api.google` types if we still need that.

cirq/google/engine/engine_job.py
```diff
@@ -310,6 +314,18 @@ def _get_job_results_v1(result: v1.program_pb2.Result
...
+    def _get_batch_results_v2(results: v2.batch_pb2.BatchResult
+                             ) -> List[study.TrialResult]:
+        trial_results = []
+        # Flatten to single list to match to sampler api.
+        for result in results.results:
+            sweep_results = v2.results_from_proto(result)
+            for sweep_result in sweep_results:
+                for trial_result in sweep_result:
+                    trial_results.append(trial_result)
+        return trial_results
```
maffoo, line 327: Could use the existing `_get_job_results_v2` function here:
```python
@classmethod
def _get_batch_results_v2(cls, ...):
    trial_results = []
    for result in results.results:
        trial_results.extend(cls._get_job_results_v2(result))
    return trial_results
```

cirq/google/engine/engine_job.py
```diff
@@ -285,6 +285,10 @@ def results(self) -> List[study.TrialResult]:
                   result_type == 'cirq.api.google.v2.Result'):
                 v2_parsed_result = v2.result_pb2.Result.FromString(result.value)
                 self._results = self._get_job_results_v2(v2_parsed_result)
+            elif result_type == 'cirq.google.api.v2.BatchResult':
+                v2_parsed_result = v2.batch_pb2.BatchResult.FromString(
+                    result.value)
+                self._results = self._get_batch_results_v2(v2_parsed_result)
```
maffoo, line 291: It seems rather odd to flatten this down to a single list of results across all programs in the batch. I would think we might want a new API to access batch results in a more structured manner.

cirq/google/engine/engine.py
```diff
@@ -276,6 +277,74 @@ def run_sweep(
...
                                         labels=job_labels)
 
+    def run_batch(
+            self,
+            programs: List['cirq.Circuit'],
+            program_id: Optional[str] = None,
+            job_id: Optional[str] = None,
+            params_list: List[study.Sweepable] = None,
+            repetitions: int = 1,
+            processor_ids: Sequence[str] = ('xmonsim',),
```
dabacon, line 287: suggest not having a default and yelling if not supplied.  this is different from other cases, but there is an issue out to not have this default (it is too dangerous, you could think you were doing an experiment, but actually be running on a simulator).

cirq/google/engine/engine.py
```diff
@@ -276,6 +277,74 @@ def run_sweep(
...
+            params_list: Parameter sweeps to use with the circuits.  The number
+                of sweeps should match the number of circuits and will be
+                paired in order with the circuits.
+            repetitions: Number of circuit repetitions to run.  Each sweep value
+                of each circuit in the batch will run with the same repetitions.
+            processor_ids: The engine processors that should be candidates
+                to run the program. Only one of these will be scheduled for
+                execution.
+            gate_set: The gate set used to serialize the circuit. The gate set
+                must be supported by the selected processor.
```
dabacon, line 324: not a fan of the default being xmon here, but we should at least document (yes run doesn't document this too)

cirq/google/engine/engine.py
dstrain115: Yeah, I guess this would get rid of the need for a copy.  Done.

cirq/google/engine/engine_job.py
dstrain115: Done.  I will leave the other cases how they are for now.

cirq/google/engine/engine_job.py
dstrain115: Done.

cirq/google/engine/engine_job.py
dstrain115: Yeah, my plan is to have something like that in a future PR that extends the interface.  This will fit into the current implementation, and then we can have a new API to grab batch results in a follow-up.  

Sound good?

cirq/google/engine/engine.py
dstrain115: Done.

cirq/google/engine/engine.py
dstrain115: Changed it to raise an error, like the other parameters.

dstrain115:
> Do we want to make `EngineProgram` and `EngineJob` be able to know if they were batch or not? Right now I don't think you won't be able to tell this easily?

Acknowledged but have not addressed yet.  Will fix before submitting.

dstrain115:
> Acknowledged but have not addressed yet. Will fix before submitting.

I have added batch_mode to the classes.  For `EngineJob`, it is purely informational.

dstrain115:
FYI: I gave this a try over the weekend and it worked!  Woo woo!

After this is submitted, I think we are good to try this out with a limited set of testers.

cirq/google/engine/engine_program.py
```diff
@@ -102,6 +107,77 @@ def run_sweep(
...
+        import cirq.google.engine.engine as engine_base
+        if not self.batch_mode:
+            raise ValueError('Can only use run_batch() in batch mode.')
+        if not job_id:
+            job_id = engine_base._make_random_id('job-')
+        if not params_list:
+            raise ValueError('No parameter list specified')
+
+        # Pack the run contexts into batches
+        batch_context = qtypes.any_pb2.Any()
```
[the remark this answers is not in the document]
dstrain115: Moved batch below the for loop.  Also changed this to import any_pb2 directly.

cirq/google/engine/engine_program.py
```diff
@@ -102,6 +107,77 @@ def run_sweep(
         return engine_job.EngineJob(self.project_id, self.program_id,
                                     created_job_id, self.context, job)
 
+    def run_batch(
+            self,
+            job_id: Optional[str] = None,
+            params_list: List[study.Sweepable] = None,
+            repetitions: int = 1,
+            processor_ids: Sequence[str] = ('xmonsim',),
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/engine/engine.py
```diff
@@ -276,6 +277,77 @@ def run_sweep(
...
 
+    def run_batch(
+            self,
+            programs: List['cirq.Circuit'],
+            program_id: Optional[str] = None,
+            job_id: Optional[str] = None,
+            params_list: List[study.Sweepable] = None,
+            repetitions: int = 1,
+            processor_ids: Sequence[str] = (),
+            gate_set: serializable_gate_set.SerializableGateSet = None,
```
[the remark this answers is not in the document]
dstrain115: Done.  Fixed in a bunch of places in this file.

cirq/google/engine/engine.py
```diff
@@ -276,6 +277,77 @@ def run_sweep(
...
+            program_labels: Optional set of labels to set on the program.
+            job_description: An optional description to set on the job.
+            job_labels: Optional set of labels to set on the job.
+
+        Returns:
+            An EngineJob. If this is iterated over it returns a list of
+            TrialResults, one for each parameter sweep.
+        """
+        if not gate_set:
+            raise ValueError('Gate set must be specified.')
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/engine/engine.py
```diff
@@ -316,6 +388,55 @@ def create_program(
...
+                alphanumeric and YYMMDD is the current year, month, and day.
+            gate_set: The gate set used to serialize the circuit. The gate set
+                must be supported by the selected processor
+            description: An optional description to set on the program.
+            labels: Optional set of labels to set on the program.
+
+        Returns:
+            A EngineProgram for the newly created program.
+        """
+        gate_set = gate_set or gate_sets.XMON
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/engine/engine.py
```diff
@@ -316,6 +388,55 @@ def create_program(
...
+
+        if not program_id:
+            program_id = _make_random_id('prog-')
+
+        code = qtypes.any_pb2.Any()
```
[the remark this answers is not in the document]
dstrain115: Implemented the first suggestion.

I hesitate to mess with the client too much, since it is partially generated by the gapic tools, so, if we regenerate the library, we don't want to mess with any customizations to the greatest extent possible.

cirq/google/engine/engine.py
```diff
@@ -276,11 +280,80 @@ def run_sweep(
...
+            params_list: List[study.Sweepable] = None,
+            repetitions: int = 1,
+            processor_ids: Sequence[str] = (),
+            gate_set: Optional[sgs.SerializableGateSet] = None,
+            program_description: Optional[str] = None,
+            program_labels: Optional[Dict[str, str]] = None,
+            job_description: Optional[str] = None,
+            job_labels: Optional[Dict[str, str]] = None,
+    ) -> engine_job.EngineJob:
+        """Runs the supplied Circuit via Quantum Engine.Creates
```
dabacon, line 297: nit: Circuit -> Circuits

PR #3115, opened 2020-06-29
Defensively copy state vector in simulator

dabacon:
The state vector simulator version of #3109  One difference is that the code process the initial state already make a copy of the state.

### DIFF
--- a/cirq/sim/sparse_simulator_test.py
+++ b/cirq/sim/sparse_simulator_test.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import itertools
 import random
 from unittest import mock
 import numpy as np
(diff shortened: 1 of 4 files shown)


cirq/sim/state_vector_simulator.py
```diff
@@ -212,7 +212,7 @@ def state_vector(self):
                 |  6  |   1    |   1    |   0    |
                 |  7  |   1    |   1    |   1    |
         """
-        return self._final_simulator_state.state_vector
+        return self._final_simulator_state.state_vector.copy()
```
95-martin-orion, line 215: Should we also have `test_final_state_vector_is_not_last_object` for this object?

cirq/sim/sparse_simulator_test.py
```diff
@@ -963,3 +964,46 @@ def test_separated_measurements():
...
+    q = cirq.LineQubit(0)
+    circuit = cirq.Circuit(InplaceGate()(q), InplaceGate()(q))
+
+    vectors = []
+    for step in sim.simulate_moment_steps(circuit):
+        vectors.append(step.state_vector(copy=True))
+    for x, y in itertools.combinations(vectors, 2):
+        assert not np.shares_memory(x, y)
+
+    # If the state vector is not copied, then applying second Hadamard
```
95-martin-orion, line 988: `Hadamard` -> `InplaceGate`

dabacon: Done

cirq/sim/state_vector_simulator.py
dabacon: It is actually tested in sparse_vector_simulator_test (result.state_vector). But I've added a test to state_vector_simulator_test to test the copy here.

dabacon:
PTAL

PR #3119, opened 2020-07-01
Propose a bug report issue

dabacon:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,17 @@
...
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+Pleas give a clear and concise description of what the bug is here.
```
balopat, line 11:
```suggestion
Please give a clear and concise description of what the bug is you see.
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,17 @@
...
+
+**To Reproduce**
+If you can describe the steps needed to reproduce here that will help us squash the bug faster.  If you can reproduce the bug using simple code, please put that code here.
```
viathor, line 14: Optional: Since you mention code samples, consider adding that github supports markdown and perhaps linking to https://docs.github.com/en/github/writing-on-github/basic-writing-and-formatting-syntax or somesuch.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,17 @@
...
+
+**Cirq version**
+Please describe what version of Cirq you are using.  If you installed cirq via pip you can get the version by running `pip show cirq`.
```
viathor, line 17: Optional: Something like `python -c 'import cirq; print(cirq.__version__)'` has the benefit of working regardless of how cirq was obtained.

.github/ISSUE_TEMPLATE/bug_report.md
viathor: +1

.github/ISSUE_TEMPLATE/bug_report.md
dabacon: done

.github/ISSUE_TEMPLATE/bug_report.md
dabacon: done

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,17 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
```
Strilanc, line 10: This is a bit over-general, like "How to solve the problem. Step 1) describe the solution to the problem. Step 2) ...". Maybe "What went wrong?".

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,17 @@
...
+Please give a clear and concise description of what the bug is.
+
+**To Reproduce**
+If you can describe the steps needed to reproduce here that will help us squash the bug faster.  If you can reproduce the bug using simple code, please put that code here.  Github supports [markdown](https://docs.github.com/en/github/writing-on-github/basic-writing-and-formatting-syntax), you can surround code snippets by three backticks \`\`\` to show the code elegantly in the bug.
```
Strilanc, line 14: Since this is a template, you can put formatting elements in for the user instead o describing how to put formatting elements in:

    ```
    import cirq
    cirq.cause_bug()
    ```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,17 @@
...
+
+**Cirq version**
+Please describe what version of Cirq you are using.  You can learn this by running `python -c 'import cirq; print(cirq.__version__)'`.
```
Strilanc, line 17: Suddenly we're in the command line context; could be confusing. Just `print(cirq.__version__)` might be better.

.github/ISSUE_TEMPLATE/bug_report.md
balopat: How about "Describe your expectation and what Cirq does instead"?

.github/ISSUE_TEMPLATE/bug_report.md
balopat: I agree - we can cut most of the description here with the code template. 
Plus, for the output, you can do: 

```
<details>
my output
</details> 
```

Which will create a foldable output for long outputs, keeping the bug tidy.

.github/ISSUE_TEMPLATE/bug_report.md
balopat: also, do we know a good command that would result in us knowing: 
- OS 
- python version 
- versions for cirq dependencies in the current Python PATH? 

Or should we write one?

dabacon:
How about this approach?  More minimalist and guides users to use code blocks and detail blocks.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+
+---
+
+**Description of the issue**
+
+**How to reproduce the issue**
+
+\`\`\`
+
+put code in code blocks
```
viathor, line 16: nit: "put code in code blocks *like this*" is a little more concrete (it implicitly explains what a "code block" is) and sounds even less formal

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+
+\`\`\`
+
+\<details\>
+
+put long logs in details blocks
```
viathor, line 22: ditto: "put long logs in details blocks *like this*"

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+
+\<\/details\>
+
+**Cirq version (optional)**
```
viathor, line 26: Why make version optional? Not specifying the version potentially makes issues harder to reproduce. Also, for users reading or searching through our issues version is also a good guide to how far cirq may have changed since something was first reported (and is therefore a better indicator of staleness than dates on github posts).

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+You can get the cirq version by printing `cirq.__version__`. From the command line this can be obtained by running
```
viathor, line 27: nit: "From the command line:" is shorter, less formal and conveys the same information.

.github/ISSUE_TEMPLATE/bug_report.md
dabacon: removed optional, mostly this was just to reduce friction.

.github/ISSUE_TEMPLATE/bug_report.md
dabacon: done

dabacon:
PTAL @balopat @Strilanc 

Address @viathor comments.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+labels: ''
+assignees: ''
+
+---
+
+**Description of the issue**
+
+**How to reproduce the issue**
+
+\`\`\`
```
balopat, line 14: no need to escape - this is a template!

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+
+put code in code blocks *like this*
+
+\`\`\`
+
+\<details\>
```
balopat, line 20: no need to escape - this is a template!

Issue #1943, opened 2019-08-13
invert_mask doesn't flip measurement results.


sammcardle30:
In the following example, I would expect setting `invert_mask` to `True` to make these two circuits equivalent. They are deemed equivalent by `cirq.testing.assert_circuits_with_terminal_measurements_are_equivalent()`, but lead to different measurement results.

```
import cirq
simulator = cirq.Simulator()

# Circuit with identity gate, and then an inverted measurement.
# Expect to measure '1' every time.
# Actually measure '0' every time.
qubit_invert = cirq.GridQubit(0, 0)
circuit_invert = cirq.Circuit()

circuit_invert.append(cirq.I(qubit_invert))
circuit_invert.append(cirq.measure(
        qubit_invert, key=str(qubit_invert), invert_mask=(True,)))

result_invert = simulator.run(circuit_invert, repetitions=5)

print('Circuit with inverted measurement:')
print(circuit_invert)
print('Measurement results: ', result_invert)
print()


# Circuit with X gate before normal measurement result.
# Expect to measure '1' every time.
# We do indeed measure '1' every time.
qubit_flip = cirq.GridQubit(0, 0)
circuit_flip = cirq.Circuit()

circuit_flip.append(cirq.I(qubit_flip))
circuit_flip.append(cirq.X(qubit_flip))
circuit_flip.append(cirq.measure(qubit_flip, key=str(qubit_flip)))

result_flip = simulator.run(circuit_flip, repetitions=5)

print('Circuit with bit flip before normal measurement:')
print(circuit_flip)
print('Measurement results: ', result_flip)
print()


# Despite obtaining different results, circuits are evaluated as equivalent.
print('Differences between circuits: ', 
      cirq.testing.assert_circuits_with_terminal_measurements_are_equivalent(
        circuit_invert, circuit_flip, atol=1E-8))
```

This produces the following output:
```
Circuit with inverted measurement:
(0, 0): ───I───!M───
Measurement results:  (0, 0)=00000

Circuit with bit flip before normal measurement:
(0, 0): ───I───X───M───
Measurement results:  (0, 0)=11111

Differences between circuits:  None
```

dabacon:
It looks like the fast simulator doesn't invert but the slow one does.  If you append an X to the first circuit it gives the correct result.

dabacon:
I have a fix for this.

PR #3123, opened 2020-07-07
Add MeasurementGate.with_key to change measurement key

maffoo:
The "rekey" operation is useful for some circuit transformations we want to implement; defining it on the MeasurementGate ensures that all the gate properties get copied properly when rekeying.

### DIFF
--- a/cirq/google/api/v2/results.py
+++ b/cirq/google/api/v2/results.py
@@ -90,15 +90,6 @@ def _grid_qubits(op: 'cirq.Operation') -> List['cirq.GridQubit']:
     return cast(List['cirq.GridQubit'], list(op.qubits))
 
 
-def _full_mask(op: 'cirq.GateOperation') -> List[bool]:
-    invert_mask = list(cast(ops.MeasurementGate, op.gate).invert_mask)
-    len_missing_mask = len(op.qubits) - len(invert_mask)
-    if len_missing_mask > 0:
-        return invert_mask + [False] * len_missing_mask
-    else:
-        return invert_mask
-
-
 def pack_bits(bits: np.ndarray) -> bytes:
     """Pack bits given as a numpy array of bools into bytes."""
     # Pad length to multiple of 8 if needed.
--- a/cirq/google/api/v2/results.py
+++ b/cirq/google/api/v2/results.py
@@ -51,7 +51,7 @@ class MeasureInfo(
     """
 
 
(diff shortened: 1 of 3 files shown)


cirq/google/api/v2/results.py
```diff
@@ -77,12 +77,11 @@ def find_measurements(program: 'cirq.Circuit',) -> List[MeasureInfo]:
 def _circuit_measurements(circuit: 'cirq.Circuit') -> Iterator[MeasureInfo]:
     for i, moment in enumerate(circuit):
         for op in moment:
-            if (isinstance(op, ops.GateOperation) and
-                    isinstance(op.gate, ops.MeasurementGate)):
+            if isinstance(op.gate, ops.MeasurementGate):
                 yield MeasureInfo(key=op.gate.key,
                                   qubits=_grid_qubits(op),
                                   slot=i,
-                                  invert_mask=_full_mask(op))
```
balopat, line 85: This leaves `_full_mask` unused, should we remove it?

maffoo: Good call. Done.

cirq/ops/measurement_gate.py
```diff
@@ -81,7 +88,8 @@ def with_bits_flipped(self, *bit_positions: int) -> 'MeasurementGate':
             new_mask[b] = not new_mask[b]
         return MeasurementGate(self.num_qubits(),
                                key=self.key,
-                               invert_mask=tuple(new_mask))
+                               invert_mask=tuple(new_mask),
+                               qid_shape=self._qid_shape)
```
balopat, line 92: This is fine here - but it popped a question in my head which might result in a new issue: how do we (do we?) apply an "invert_mask" on qudits?

maffoo: I don't think we do that anywhere. Frankly I'm not sure that invert mask as a thing is particularly useful, but we added it early on (well before we had any notion of qudits in cirq).

balopat: Got it! 

Re qudits: I found https://github.com/quantumlib/Cirq/pull/2079 explaining that it flips |0> with |1> and leaves |2+> alone. 
Re general usefulness: It seems (at least based on https://github.com/quantumlib/Cirq/issues/1943) that some users use invert mask

maffoo:
Is there a way to tell the cla bot to report its status?

PR #3125, opened 2020-07-07
Add KakDecomposition._decompose_ by fixing factor of 2 error in exp(PauliString)

Strilanc:
- I compared against results from WolframAlpha to make sure the modified tests are now correct

### DIFF
--- a/cirq/linalg/decompositions_test.py
+++ b/cirq/linalg/decompositions_test.py
@@ -724,3 +724,16 @@ def test_kak_vector_negative_atol():
 def test_kak_vector_input_not_unitary():
     with pytest.raises(ValueError, match='must correspond to'):
         cirq.kak_vector(np.zeros((4, 4)))
+
+
+@pytest.mark.parametrize('unitary', [
+    cirq.testing.random_unitary(4),
+    cirq.unitary(cirq.ISWAP),
+    cirq.unitary(cirq.CZ**0.5),
+])
+def test_kak_decompose(unitary: np.ndarray):
+    kak = cirq.kak_decomposition(unitary)
+    circuit = cirq.Circuit(kak._decompose_(cirq.LineQubit.range(2)))
+    np.testing.assert_allclose(cirq.unitary(circuit), unitary, atol=1e-8)
+    assert len(circuit) == 5
+    assert len(list(circuit.all_operations())) == 8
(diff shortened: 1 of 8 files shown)


cirq/ops/pauli_string_test.py
```diff
@@ -532,6 +533,21 @@ def test_pow():
...
+def test_rpow():
+    a, b = cirq.LineQubit.range(2)
+
+    u = cirq.unitary(np.exp(1j * np.pi / 2 * cirq.Z(a) * cirq.Z(b)))
+    np.testing.assert_allclose(u, np.diag([1j, -1j, -1j, 1j]), atol=1e-8)
+
+    u = cirq.unitary(np.exp(-1j * np.pi / 4 * cirq.Z(a) * cirq.Z(b)))
+    cirq.testing.assert_allclose_up_to_global_phase(u,
+                                                    np.diag([1, 1j, 1j, 1]),
+                                                    atol=1e-8)
```
viathor, line 545: Stronger assertion:

```python
cirq.testing. assert_allclose(u, np.diag(np.sqrt(-1j), np.sqrt(1j), np.sqrt(1j), np.sqrt(-1j)), atol=1e-8)
```

cirq/linalg/decompositions_test.py
```diff
@@ -724,3 +724,16 @@ def test_kak_vector_negative_atol():
 def test_kak_vector_input_not_unitary():
     with pytest.raises(ValueError, match='must correspond to'):
         cirq.kak_vector(np.zeros((4, 4)))
+
+
+@pytest.mark.parametrize('unitary', [
+    cirq.testing.random_unitary(4),
```
viathor, line 730: Optional: I wonder what distribution we get here on the elements of KAK (e.g. on the three interaction coefficients). A good addition to the random unitary test would be one synthesized from known coefficients, e.g.

```python
u = np.exp(1j*np.pi*(0.2*cirq.unitary(cirq.XX) + 0.3*cirq.unitary(cirq.YY) + 0.4*cirq.unitary(cirq.ZZ)))
```

cirq/ops/pauli_string_test.py
Strilanc: I intentionally only did it up to global phase to make the test simpler to write, to make sure the complications weren't hiding a mistake.

cirq/linalg/decompositions_test.py
Strilanc: I added a couple more cases. The "arbitrary fixed coefficients" case is very well covered by the random unitary; the corner cases have to do with interaction coefficients being zero or maxed.

PR #3126, opened 2020-07-08
Updating Pasqal classes

HGSilveri:
Significant modifications to the first version of the `pasqal` module. Most relevantly:
- `PasqalDevice` was modified to be the parent class of all Pasqal devices. Also allows submission of nearly unconstrained circuits to Pasqal processors.
- `PasqalVirtualDevice` enforces the constraints of a near-term Pasqal QPU.
- `ThreeDGridQubit` was renamed `ThreeDQubit` (and `TwoDQubit` was created as a subclass).
- More flexibility in the accepted qubits (`PasqalVirtualDevice` can also work with `GridQubit` and `LineQubit`)
- More adequate accepted gate set, now consistent with the hardware

New additions:
- Documentation (docs/pasqal) and "getting_started" tutorial


HGSilveri:
The check that is failing has to do with the new documentation files (in docs/pasqal) not being in any toctree. I don't want to mess with the structure of your documentation, so I think it should be up to you to decide where to list them. Please, feel free to do so! Otherwise, you can just tell me where you would like me to add them and I will do so. Thanks!

cirq/ops/pauli_string.py
```diff
@@ -259,26 +259,6 @@ def keys(self) -> KeysView[raw_types.Qid]:
     def qubits(self) -> Tuple[raw_types.Qid, ...]:
         return tuple(sorted(self.keys()))
 
-    def _circuit_diagram_info_(self, args: 'cirq.CircuitDiagramInfoArgs'
```
balopat, line 262: I'm not sure why we removed the PauliString diagram. Now we have

```
0: ───X───-X───X(0)*Z(2)───1j*X(0)*Y(1)───────────(0.7071067811865476+0.7071067811865475j)*X(0)*Y(1)───
               │           │                      │
1: ────────────┼───────────#2─────────────-1j*Y───#2───────────────────────────────────────────────────
               │
2: ────────────#2──────────────────────────────────────────────────────────────────────────────────────

```

instead of 

```
0: ───PauliString(+X)───PauliString(-X)───PauliString(+X)───PauliString(iX)──────────────────────PauliString((0.707+0.707i)*X)───
                                          │                 │                                    │
1: ───────────────────────────────────────┼─────────────────Y─────────────────PauliString(-iY)───Y───────────────────────────────
                                          │
2: ───────────────────────────────────────Z──────────────────────────────────────────────────────────────────────────────────────
```

is this intentional?

cirq/testing/random_circuit_test.py
```diff
@@ -37,44 +37,21 @@ def test_random_circuit_errors():
...
-
 @pytest.mark.parametrize(
     'n_qubits,n_moments,op_density,gate_domain,pass_qubits',
-    _cases_for_random_circuit())
+    [(random.randint(1, 20), random.randint(1, 10), random.random(),
+      (None if random.randint(0, 1) else dict(
+          random.sample(
+              tuple(cirq.testing.DEFAULT_GATE_DOMAIN.items()),
+              random.randint(1, len(cirq.testing.DEFAULT_GATE_DOMAIN))))),
+      random.choice((True, False))) for _ in range(10)])
```
balopat, line 54: This seems like an unrelated change, also, it is good practice to extract long expressions like this to a named private method for better readability.

cirq/testing/random_circuit.py
```diff
@@ -96,8 +96,8 @@ def random_circuit(qubits: Union[Sequence[ops.Qid], int],
         raise ValueError('At least one qubit must be specified.')
     gate_domain = {k: v for k, v in gate_domain.items() if v <= n_qubits}
     if not gate_domain:
-        raise ValueError(f'After removing gates that act on less than '
-                         f'{n_qubits} qubits, gate_domain had no gates.')
+        raise ValueError(f'After removing gates that act on less that '
```
balopat, line 99: I think this might be accidental - `less than` was correct.

cirq/testing/random_circuit.py
```diff
@@ -96,8 +96,8 @@ def random_circuit(qubits: Union[Sequence[ops.Qid], int],
...
+                         '{n_qubits}, gate_domain had no gates.')
```
balopat, line 100: And this was correct as an F-string, please revert.

dev_tools/docs/build_api_docs.py
```diff
@@ -1,64 +0,0 @@
-# Copyright 2020 The Cirq Authors. All Rights Reserved.
```
balopat, line 1: why did we delete this file?

cirq/protocols/circuit_diagram_info_protocol.py
```diff
@@ -153,19 +153,6 @@ def format_real(self, val: Union[sympy.Basic, int, float]) -> str:
             return str(val)
         return f'{float(val):.{self.precision}}'
 
-    def format_complex(self,
```
balopat, line 156: We shouldn't remove this formatter from the diagrams. Why do you think it needs to go?

cirq/protocols/circuit_diagram_info_protocol_test.py
```diff
@@ -200,29 +200,6 @@ def test_format_real():
     assert args.format_real(sympy.Symbol('t') * 2 + 1) == '2*t + 1'
 
 
-def test_format_complex():
```
balopat, line 203: If we bring back the formatter - we should bring back the test too.

cirq/protocols/circuit_diagram_info_protocol_test.py
```diff
@@ -183,7 +183,7 @@ def test_circuit_diagram_info_args_repr():
 
 
 def test_format_real():
-    args = cirq.CircuitDiagramInfoArgs.UNINFORMED_DEFAULT.copy()
```
balopat, line 186: Why remove copy()? Otherwise the same instance is going to be mutated across multiple tests that can result in surprising errors.

cirq/ops/pauli_string_test.py
```diff
@@ -1600,25 +1600,3 @@ def test_deprecated():
         _ = cirq.PauliString({
             a: 'x'
         }).expectation_from_state_vector(state=state_vector, qubit_map={a: 0})
-
-
-def test_circuit_diagram_info():
```
balopat, line 1605: We should bring back this test too after we brought back the diagram info for PauliStrings.

HGSilveri:
Hi @balopat, thank you very much for the remarks! You are absolutely right, these are all accidental and unrelated to our classes. I will  clean up every change that didn't come from our side ASAP, thank you for your patience.

HGSilveri:
Hi again @balopat, it should be okay now. I apologise for the trouble!

docs/pasqal/sampler.md
```diff
@@ -0,0 +1,45 @@
...
+p_circuit = cirq.Circuit(device=p_device)
+p_circuit.append(cirq.X(qubit))                        # NOT gate.
+p_circuit.append(cirq.measure(qubit, key='result'))    # Measurement.
+
+
+# Create a PasqalSampler object to use.
+# Replace 'my_token' with the access token and uncomment next lines.
+
+# PASQAL_API_ACCESS_TOKEN = 'my_token'
+# sampler = cirq.pasqal.PasqalSampler(remote_host='http://34.98.71.118/v0/pasqal', access_token=PASQAL_API_ACCESS_TOKEN)
```
balopat, line 31: is this IP something you want to expose?

cirq/pasqal/pasqal_qubits.py
```diff
@@ -11,184 +11,199 @@
...
+
+    def __rsub__(self, other: Tuple[float, float, float]) -> 'ThreeDQubit':
+        return -self + other
+
+    def __neg__(self) -> 'ThreeDQubit':
+        return ThreeDQubit(x=-self.x, y=-self.y, z=-self.z)
+
+
+class TwoDQubit(ThreeDQubit):
+    """A qubit in 2d."""
```
balopat, line 144: I am a little bit confused about this at a first read, let's take this apart a bit - why does a 2D qubit inherits from a 3D qubit? Also GridQubit _is_ a 2D qubit... It would be great to clarify these design choices.

docs/pasqal/sampler.md
HGSilveri: Yes, no problem with that.

cirq/pasqal/pasqal_qubits.py
HGSilveri: So, the rationale behind this is that 3D qubit is the most general qubit in our devices, which can be placed in arbitrary positions in 3D. The 2D qubit is then a particular case of a 3D qubit in which the z-coordinate is fixed to 0. Some devices will only allow a 2D layout, so it's helpful to have a specific object for these devices. It also allows us to incorporate convenient static methods to create layouts in specific 2D patterns (e.g. in a triangular lattice), which would be ambivalent to have on ThreeDQubit.

As for the GridQubit, it is true that it is a 2D qubit, but it restricts the qubit positions to a grid, while we accept arbitrary positions on a plane in TwoDQubit. We are accepting GridQubit and LineQubit to give the users more flexibility, but in fact they could be completely substituted by the TwoDQubit.

balopat: This is great, makes sense now, thank you! 
It would be very beneficial for users to be able to see this information in the pydoc for the class, so that they can more easily navigate the different options.

cirq/pasqal/pasqal_device.py
```diff
@@ -11,52 +11,45 @@
...
+        """Initializes a device with some qubits.
 
         Args:
-            control_radius: the maximum distance between qubits for a controlled
-                gate. Distance is measured in units of the indices passed into
-                the qubit constructor.
-            qubits: Qubits on the device, identified by their x, y, z position.
-                Must be of type ThreeDGridQubit.
-
+            qubits (NamedQubit): Qubits on the device, exclusively unrelated to
```
balopat, line 33: Why does it needs to be `NamedQubit`? If this is a recommendation, explain in the doc. If it's a requirement, it should be represented in the type signature.

balopat: I see the issue here, we can't represent it easily in the type signature.

HGSilveri: `NamedQubit`was the only qubit native to Cirq (`Qid`being to general for practical purposes) that would not misleadingly make the user position the qubit in space. We could have created another qubit class specifically for this, but it seemed overkill. I realise this is not optimal, but it felt better than accepting (necessarily) spatially positioned qubits when those positions will ultimately be ignored, or cluttering the module with yet another qubit type. Do you agree?

To be clear, we are requiring the qubits to be `NamedQubit`, so I will represent this in the type signature.

HGSilveri:
Hi again @balopat, thank you for your time and help so far. Regarding your recommendation, I understand your point. However, these changes are very much intertwined, so it is easier said that done. Let me elaborate on what I think, realistically, can be done:

- Separating the docs is an easy split, which should already remove 8 files from the PR.
- The remaining 20 are all interconnected and it is essentially impossible to contain a change to only 10 (e.g. changing `ThreeDGridQubit` to `ThreeDQubit` alone will touch nearly all of them). However, I should point out that out of these 20 files, 11 are small changes related to json serialization (either files that are deleted, added or very slightly changed) and are thus very straightforward to review. 
- That being said, an effort could still be made into focusing each PR on specific changes. I agree with your division, but in reality points 1-4 are very hard to dissociate. If you insist, I believe we could make 5. into a separate PR though. 

To summarise,  I suggest we divide the PR into a big PR (points 1-4 or 1-5), followed by one or two smaller PR's (just 6 or 5 and 6). I understand this is not ideal, but it is really impossible to naturally dissociate the changes in the Qubit classes and the Device classes, since they influence one another. Plus, forcing this split would still entail a disappointingly small reduction in the number of files per PR.

cirq/pasqal/pasqal_device.py
HGSilveri: > I see the issue here, we can't represent it easily in the type signature.

You're right, I remember now why I couldn't fix the type to restrict the input to NamedQubit. I'm still checking the type upon initialization, so I hope that suffices.

balopat:
Hey @HGSilveri, 

Thanks for thinking this through with me! This sounds good, let's do 1-4, 5, 6 separately then. It is not only about the number of files. It is also about the number of concepts that change or are introduced in a PR. Last year when I developed the QASM Importer, I ran into the same issue with only 11 files that I broke out and got much faster reviews - well it still took some time, but at the end it was really easier on the reviewers (https://github.com/quantumlib/Cirq/pulls?q=is%3Apr+is%3Aclosed+author%3Abalopat+qasm+import). Thank you!

HGSilveri:
As discussed above, the PR will be closed in favour of a split into smaller PR's.

PR #3128, opened 2020-07-09
Add getitem to EngineJob

### LINKED ISSUE #3127: EngineJob should not just be iterable but also indexable
Currently only implements `__iter__` but think this should also implement `__getitem__`.

dabacon:
Currently `EngineJob` can be iterated over, but results cannot be accessed by index.  This adds the ability to look up by index.

Fixes #3127


cirq/google/engine/engine_job.py
```diff
@@ -365,6 +368,9 @@ def _raise_on_failure(job: quantum.types.QuantumJob) -> None:
     def __iter__(self) -> Iterator[study.TrialResult]:
         return iter(self.results())
 
+    def __getitem__(self, item) -> study.TrialResult:
+        return list(self.results())[item]
```
viathor, line 372: I we believe type annotations then `results` returns a list, so we shouldn't need to invoke the constructor.

```python
def results(self) -> List[study.TrialResult]:
```

cirq/google/engine/engine_job.py
```diff
@@ -365,6 +368,9 @@ def _raise_on_failure(job: quantum.types.QuantumJob) -> None:
     def __iter__(self) -> Iterator[study.TrialResult]:
         return iter(self.results())
 
+    def __getitem__(self, item) -> study.TrialResult:
```
viathor, line 371: Missing type annotation for `item`.

maffoo, line 371: Could also use overloads to tell mypy that this acceps `int` or `slice` and returns `study.TrialResult` in the first case but `List[study.TrialResult]` in the latter.

cirq/google/engine/engine_job.py
dabacon: oop fixed.

cirq/google/engine/engine_job.py
dabacon: yeah should be union of TrialResult and List of TrialResults I think. fixed

dabacon: fixed

dabacon: fixed

dabacon: fixed

cirq/google/engine/engine_job.py
dabacon: fixed

cirq/google/engine/engine_job.py
viathor: Union will successfully pass type-checking of this function but for the purposes of type-checking the callers it's somewhat inferior in that it hides from mypy the relationship that exists between argument and return types (namely that `int` gives you a single `study.TrialResult` while a `slice` gives you a sequence).

You can fix this by using [overloads](https://mypy.readthedocs.io/en/stable/more_types.html#function-overloading), probably like:

```python
@overload
def __getitem__(self, item: int) -> study.TrialResult: ...
@overload
def __getitem__(self, item: slice) -> Sequence[study.TrialResult]: ...

def __getitem__(self, item):
    return self.results()[item]
```

cirq/google/engine/engine_job.py
```diff
@@ -365,6 +369,18 @@ def _raise_on_failure(job: quantum.types.QuantumJob) -> None:
     def __iter__(self) -> Iterator[study.TrialResult]:
         return iter(self.results())
 
+    # pylint: disable=function-redefined
```
viathor, line 372: I think we should re-enable on l.383.

cirq/google/engine/engine_job.py
```diff
@@ -365,6 +369,18 @@ def _raise_on_failure(job: quantum.types.QuantumJob) -> None:
...
+    @overload
+    def __getitem__(self, item: int) -> study.TrialResult:
```
viathor, line 374: In practice, the client code will probably also want to know the number of results available so that it can decide which indices are valid, so we should probably also implement `__len__`.

cirq/google/engine/engine_job.py
```diff
@@ -42,6 +43,9 @@ class EngineJob:
     executing on a machine, or it may have entered a terminal state
     (either succeeding or failing).
 
+    `EngineJob`s can be iterated over, returning `TrialResult`s. These
```
dstrain115, line 46: Optional: Should we mention that this will force retrieval of the results?  (or is that obvious by returning TrialResults?)

dabacon: done

cirq/google/engine/engine_job.py
dabacon: done

dabacon:
Should not code during meetings apparently.

viathor:
Dunno about your meeting, but some are so boring the itch to hack away is irresistible.

PR #3132, opened 2020-07-10
Bump protobuf requirement to 3.12.x

### LINKED ISSUE #3110: Incompatible requirement: protobuf==3.8.0
Hello

When installing cirq 0.8.1 using pip

```ERROR: google-api-core 1.21.0 has requirement protobuf>=3.12.0, but you'll have protobuf 3.8.0 which is incompatible.```

cirq has a requirement ```protobuf == 3.8.0 ```. this is incompatible with what other dependency wants.

```
> cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04 LTS (Focal Fossa)"
...
> pip --version
pip 20.0.2 from /usr/lib/python3/dist-packages/pip (python 3.8)
```

### PR DESCRIPTION
**Pull request by balopat**
Fixes #3110.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ matplotlib~=3.0
 networkx~=2.4
 numpy~=1.16, < 1.19
 pandas
-protobuf~=3.11.0
+protobuf~=3.12.0
 requests~=2.18
 sortedcontainers~=2.0
 scipy
--- a/cirq/google/arg_func_langs_test.py
+++ b/cirq/google/arg_func_langs_test.py
@@ -105,7 +105,12 @@ def test_serialize_sympy_constants():
                                        including_default_value_fields=True,
                                        preserving_proto_field_name=True,
                                        use_integers_for_enums=True)
-    assert packed == {'arg_value': {'float_value': float(np.float32(sympy.pi))}}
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by balopat** [comment]
@googlebot please rescan

File: requirements.txt [hunk 1]
```diff
@@ -7,7 +7,7 @@ matplotlib~=3.0
 networkx~=2.4
 numpy~=1.16, < 1.19
 pandas
-protobuf~=3.11.0
+protobuf~=3.12.0
```
dabacon, line 10: Do we want this to be 3.12 to possibly pick up bug fix releases?

dabacon:
@googlebot has turned into "flakeybot"

requirements.txt
balopat: In my understanding the `~=` does exactly that: https://www.python.org/dev/peps/pep-0440/#compatible-release
I.e. `~=3.12.0` is equivalent to `>= 3.12.0, == 3.12.*`. We could argue that maybe can start at 3.12.3 as that's today's latest, but I thought we can live with that lower boundary to reduce noise for those who have [3.12.0. 3.12.2] installed.

balopat:
Thank you - let's hold off the release until I understand the implications of the float truncation better.

balopat:
@googlebot please rescan

PR #3133, opened 2020-07-13
[Pasqal updates] New qubit classes

HGSilveri:
This the first child-PR of #3126 .

- Introduces new Pasqal qubit classes: `ThreeDQubit`and `TwoDQubit` (a subclass of `ThreeDQubit`)
- These new classes are intended to replace the original `ThreeDGridQubit`, which will be deleted on the next PR (where other classes' dependencies will be resolved).
- In comparison with `ThreeDGridQubit`, these new classes allow for arbitrary placement of qubits in 3D or 2D space, instead of restricting them to a grid. This more closely resembles the capabilities of Pasqal devices.

### DIFF
--- a/cirq/pasqal/pasqal_qubits.py
+++ b/cirq/pasqal/pasqal_qubits.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 from typing import Any, Dict, Iterable, List, Optional, Set, Tuple
 from numpy import sqrt, isclose
+import numpy as np
 
 import cirq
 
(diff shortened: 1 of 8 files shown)


HGSilveri:
Good news, @balopat! I managed to make a smaller first PR than previously anticipated by only adding the new qubit classes. `ThreeDGridQubit` had to be left here and will be deleted once `PasqalDevice` is updated, but I still think this will facilitate the review process. Thank you for all your help so far, I hope this helps!

cirq/pasqal/pasqal_qubits_test.py
```diff
@@ -219,3 +219,195 @@ def test_to_json():
         'col': 1,
         'lay': 1,
     }
+
+# NEW QUBITS TESTS
```
balopat, line 223: nit: this comment seems unnecessary

cirq/pasqal/pasqal_qubits.py
```diff
@@ -192,3 +193,194 @@ def __rsub__(self, other: Tuple[int, int, int]) -> 'ThreeDGridQubit':
...
+
+    def _comparison_key(self):
+        return round(self.z, 9), round(self.y, 9), round(self.x, 9)
+
+    @property
+    def dimension(self) -> int:
+        return 2
+
+    def distance(self, other: cirq.ops.Qid) -> float:
+        """Returns the distance between two qubits in 3D."""
```
balopat, line 230: Nit: In other places we use `3d`, `2d`. 
Having said that I realize that we have inconsistencies within Cirq - which I will probably cleanup :)

cirq/pasqal/pasqal_qubits.py
```diff
@@ -192,3 +193,194 @@ def __rsub__(self, other: Tuple[int, int, int]) -> 'ThreeDGridQubit':
...
+        pasqal.ThreeDQubit(1.4, 1.1, 1.9)
+    """
+
+    def __init__(self, x: float, y: float, z: float):
+        self.x = x
+        self.y = y
+        self.z = z
+
+    def _comparison_key(self):
+        return round(self.z, 9), round(self.y, 9), round(self.x, 9)
```
balopat, line 223: Why rounding the coordinates? And why to 9 digits?

cirq/pasqal/pasqal_qubits.py
```diff
@@ -192,3 +193,194 @@ def __rsub__(self, other: Tuple[int, int, int]) -> 'ThreeDGridQubit':
...
+        < ThreeDQubit(0, 0, 1) < ThreeDQubit(1, 0, 1)
+        < ThreeDQubit(0, 1, 1) < ThreeDQubit(1, 1, 1)
+
+    New ThreeDQubit can be constructed by adding or subtracting tuples
+
+        >>> cirq.pasqal.ThreeDQubit(2.5, 3, 4.7) + (3, 1.2, 6)
+        pasqal.ThreeDQubit(5.5, 4.2, 10.7)
+
+        >>> cirq.pasqal.ThreeDQubit(2.4, 3.1, 4) - (1, 2, 2.1)
+        pasqal.ThreeDQubit(1.4, 1.1, 1.9)
```
balopat, line 214: Is it worth mentioning what unit these floats represent? Also, is it important for them to be floats?

cirq/pasqal/pasqal_qubits.py
```diff
@@ -192,3 +193,194 @@ def __rsub__(self, other: Tuple[int, int, int]) -> 'ThreeDGridQubit':
...
+        Returns:
+            A list of TwoDQubits filling in a rectangular grid
+        """
+        return [
+            TwoDQubit(x0 + x, y0 + y) for y in range(cols) for x in range(rows)
+        ]
+
+    @staticmethod
+    def triangular_lattice(l: int, x0: float = 0, y0: float = 0):
+        """Returns a triangular lattice of TwoDQubit.
```
balopat, line 361:
```suggestion
        """Returns a triangular lattice of TwoDQubits.
```

cirq/pasqal/pasqal_qubits.py
```diff
@@ -192,3 +193,194 @@ def __rsub__(self, other: Tuple[int, int, int]) -> 'ThreeDGridQubit':
...
+
+        Args:
+            l: Number of qubits along one direction
+            x0: x-coordinate of the first qubit
+            y0: y-coordinate of the first qubit
+
+        Returns:
+            A list of TwoDQubit filling in a triangular lattice.
```
balopat, line 369:
```suggestion
            A list of TwoDQubits filling in a triangular lattice.
```

cirq/pasqal/pasqal_qubits_test.py
HGSilveri: I agree, it was just a mark for the time the old tests are removed. I have no problem removing it.

cirq/pasqal/pasqal_qubits.py
HGSilveri: The units are arbitrary, they only matter with relation to the control radius. It makes sense for them to be floats to reiterate that we can place them in arbitrary positions.

cirq/pasqal/pasqal_qubits.py
HGSilveri: I don't recall exactly what prompted this, but I'm assuming it was some machine precision error. We end up using a lot of irrational numbers from triangular lattice coordinates, so I'm assuming there was some false negative.
As for the rounding, I think it is arbitrary, it could have more digits. I'm happy to change it to a higher value.

HGSilveri:
That's good to hear @balopat ! Thanks a lot for the quick review, I've addressed all your questions and nits.

cirq/pasqal/pasqal_qubits.py
```diff
@@ -192,3 +193,194 @@ def __rsub__(self, other: Tuple[int, int, int]) -> 'ThreeDGridQubit':
...
+
+        Args:
+            rows: Number of rows in the rectangle
+            cols: Number of columns in the rectangle
+            x0: x-coordinate of the first qubit
+            y0: y-coordinate of the first qubit
+            z0: z-coordinate of the first qubit
+
+        Returns:
+            A list of ThreeDQubits filling in a 3D grid
```
balopat, line 276: nit: 3d

cirq/pasqal/pasqal_qubits.py
balopat: Great - if you could recover that use case that would be useful - we should add it to the tests to protect this functionality. Otherwise someone might just remove this without failing tests.

cirq/pasqal/pasqal_qubits.py
```diff
@@ -192,3 +193,194 @@ def __rsub__(self, other: Tuple[int, int, int]) -> 'ThreeDGridQubit':
...
+        return 'pasqal.ThreeDQubit({}, {}, {})'.format(self.x, self.y, self.z)
+
+    def __str__(self):
+        return '({}, {}, {})'.format(self.x, self.y, self.z)
+
+    def _json_dict_(self):
+        return cirq.protocols.obj_to_dict_helper(self, ['x', 'y', 'z'])
+
+    def __add__(self, other: Tuple[float, float, float]) -> 'ThreeDQubit':
+        if not (isinstance(other, tuple) and len(other) == 3 and
```
balopat, line 293: one last question: `GridQubit` does have the functionality to add/substract `GridQubit`s together in here: 

https://github.com/quantumlib/Cirq/blob/adc8964697feb9f97474ab13d60c36a6f27f3bd4/cirq/devices/grid_qubit.py#L76-L82

Do you mind adding that with tests to be consistent?

HGSilveri: Actually, I've been discussing with the rest of the team and we reached the conclusion that these dunder methods for addition and subtraction of qubits don't really make sense for our case. 
They were first added because the `ThreeDGridQubit` class tried to mimic `GridQubit` and were just left there for no particular reason. As the `ThreeDGridQubit` class turned into `ThreeDQubit`, it made increasingly less sense to have them, so I am now advocating for their removal. However, before I do so, I would like to know,  do you have any objection or counter-argument to this?

cirq/pasqal/pasqal_qubits.py
HGSilveri: Turns out rounding to 9 was excessive, I increased the number of digits to 15 and introduced a test to back that up.

cirq/pasqal/pasqal_qubits.py
balopat: No counter argument / objection on my end - I prefer not having it at all then having it half-way implemented!

PR #3134, opened 2020-07-13
Add docs for circuit batching

dstrain115:
- Add documentation under Google docs for
batching circuits together using engine.run_batch()

### DIFF
--- a/docs/google/best_practices.md
+++ b/docs/google/best_practices.md
@@ -137,7 +137,7 @@ of your program so that more repetitions are completed per second.
 
 In order to efficiently batch circuits, the circuits in these batches must
 measure the same qubits and have the same number of repetitions for each
-circuit.  Otherwise, the circuits will not be able to batched together
+circuit.  Otherwise, the circuits will not be batched together
 on the device, and there will be no gain in efficiency.
 
 ## Keep qubits busy
@@ -201,4 +201,3 @@ print(list(flat_sweep.param_tuples()))
 #  (('<2**t - 1>', 0.681792830507429),),
 #  (('<2**t - 1>', 1.0),)]
 ```
-
--- a/docs/google/best_practices.md
+++ b/docs/google/best_practices.md
@@ -49,7 +49,7 @@ the measurement gate after all optimizers have run.
 
 In the current NISQ (noisy intermediate scale quantum) era, gates and devices still
 have significant error. Both gate errors and T1 decay rate can cause long circuits
-to have noise that overwhelms any signal in the circuit. 
+to have noise that overwhelms any signal in the circuit.
 
 The recommended gate depths vary significantly with the structure of the circuit itself
 and will likely increase as the devices improve. Total circuit fidelity can be roughly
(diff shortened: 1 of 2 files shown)


docs/google/best_practices.md
```diff
@@ -129,9 +129,18 @@ One word of caution is there is a limit to the total number of repetitions.  Tak
 that your parameter sweeps, especially products of sweeps, do not become so excessively large
 that they overcome this limit.
 
+## Use batches if sweeps are not possible
+
+The engine has a method called `run_batch()` that can be used to send multiple
+circuits in a single request.  This can be used to increase the efficiency
+and rep rate of your program.
```
balopat, line 136: repetition rate? Maybe we want to define that concept too?

docs/google/best_practices.md
```diff
@@ -129,9 +129,18 @@ One word of caution is there is a limit to the total number of repetitions.  Tak
...
+
+In order to efficiently batch circuits, the circuits in these batches must
+measure the same qubits and have the same number of repetitions for each
+circuit.
```
balopat, line 140: What happens when there is a difference between measurements? How bad is it going to be? 
What amount of repetition difference are we tolerating? Can you explain a little bit more why these constraints result in the most efficient batching?

docs/google/engine.md
```diff
@@ -54,7 +70,7 @@ circuit = cirq.Circuit(
     cirq.measure(qubit, key='result')   # Measurement.
 )
 
-# Create an Engine object to use.
+# Creates an Engine object.
```
balopat, line 73: You are using `Create` further below

dstrain115: Done.

docs/google/best_practices.md
dstrain115: I added a short note onto this.  However, the reasons for this are fairly deep in the hardware and I don't think it's appropriate to put too much more information on the public page.

docs/google/best_practices.md
dstrain115: Done.

docs/google/best_practices.md
```diff
@@ -129,9 +129,19 @@ One word of caution is there is a limit to the total number of repetitions.  Tak
...
 
+## Use batches if sweeps are not possible
+
+The engine has a method called `run_batch()` that can be used to send multiple
+circuits in a single request.  This can be used to increase the efficiency
+of your program so that more repetitions are completed per second.
+
+In order to efficiently batch circuits, the circuits in these batches must
+measure the same qubits and have the same number of repetitions for each
+circuit.  Otherwise, the circuits will not be able to batched together
```
balopat, line 140:
```suggestion
circuit.  Otherwise, the circuits will not be batched together
```

docs/google/best_practices.md
```diff
@@ -129,9 +129,19 @@ One word of caution is there is a limit to the total number of repetitions.  Tak
 that your parameter sweeps, especially products of sweeps, do not become so excessively large
 that they overcome this limit.
 
+## Use batches if sweeps are not possible
+
+The engine has a method called `run_batch()` that can be used to send multiple
+circuits in a single request.  This can be used to increase the efficiency
+of your program so that more repetitions are completed per second.
+
+In order to efficiently batch circuits, the circuits in these batches must
```
balopat, line 138:
```suggestion
The circuits that are meant to be executed in batches must
```
I.e. there is no such thing as "efficiently batching" and "less efficiently batching" - it seems that this is more binary - you must measure the same qubits otherwise there is no batching.

docs/google/best_practices.md
balopat: That sounds good - that was what I was looking for!

PR #3137, opened 2020-07-14
release process nits

### LINKED ISSUE #3136: release process nits
As I went through releasing v0.8.2 I noticed a following small things: 

1) _version.py is under the cirq folder, the code snippets should reflect that
2) -r dev_tools/conf/pip-list-dev-tools.txt might not be actually required, and it was failing on python 3.8 (typed-ast fails to compile) but succeeded on 3.7 
3) ./dev_tools/packaging/verify-published-package.sh fails on MacOSX - it would be useful to mention that the instructions are for Linux only - or fix the script for Mac 
4) `# NOTE: THE FIRST RUN WILL LIKELY FAIL`  I didn't understand what the reason for this comment is - but in the --prod version I saw that the package wasn't indexed yet - maybe we could clarify that. 
5) there is no need to update the master branch in case of a bugfix release

balopat:
Fixes #3136.

### DIFF
--- a/release.md
+++ b/release.md
@@ -53,12 +53,12 @@ git pull origin master  # or upstream master
 git status  # should be no pending changes
 ```
 
-For PATCH update: Make sure you're in the last version 
+For PATCH update: Make sure you checked out the version you want to patch. 
+Typically this will be something like `${MAJOR}.${MINOR}.${LAST_PATCH}` 
 
 ```bash
 git fetch origin # or upstream - to fetch all tags
-export LATEST_TAG=$(git tag | tail -n 1) # check with echo $LATEST_TAG 
-git checkout $LATEST_TAG  
+git checkout <desired tag to patch>   
 git status  # should be no pending changes
 ```
 
@@ -86,6 +86,19 @@ Create a release branch called "v${VERSION}-dev":
 
 ```bash
 git checkout -b "v${VER}-dev"
+```
+
+If you are doing a PATCH update, also cherrypick the commits for the fixes 
+you want to include in your update and resolve all potential merge conflicts 
+carefully: 
+
+```bash
+git cherry-pick <commit> 
+```
+
+Bump the version on the release branch: 
+
+```bash
 vi ./cirq/_version.py   # Remove .dev from version
 git add ./cirq/_version.py
 git commit -m "Bump cirq version to ${NEXT_VER}"
--- a/release.md
+++ b/release.md
@@ -42,11 +42,23 @@ release.
 
 ### Preparation
 
-Make sure you're on an up-to-date master branch and in cirq's root directory.
+System requirements: Linux, python3.7
+
+For MINOR / MAJOR release: Make sure you're on an up-to-date master branch and 
+in cirq's root directory.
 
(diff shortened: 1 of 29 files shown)


release.md
```diff
@@ -42,21 +42,33 @@ release.
...
+For MINOR / MAJOR release: Make sure you're on an up-to-date master branch and 
+in cirq's root directory.
 
 ```bash
 git checkout master
 git pull origin master  # or upstream master
 git status  # should be no pending changes
 ```
 
+For PATCH update: Make sure you're in the last version
```
dstrain115, line 56: Maybe say "in the last version's branch" to make it more clear?

release.md
```diff
@@ -42,21 +42,33 @@ release.
...
 git checkout master
 git pull origin master  # or upstream master
 git status  # should be no pending changes
 ```
 
+For PATCH update: Make sure you're in the last version 
+
+```bash
+git fetch origin # or upstream - to fetch all tags
+export LATEST_TAG=$(git tag | tail -n 1) # check with echo $LATEST_TAG
```
dstrain115, line 60: This seems risky, since it assumes that the last tag is what we want.  What if we patch both 0.8.2 and 0.7.1? Or create a tag for a non-release reason?

At the very least, I would change the echo to an actual command, like echo "You are in branch $LATEST_TAG".

release.md
```diff
@@ -65,28 +77,31 @@ Also define these variables for the versions you are releasing:
...
 Create a release branch called "v${VERSION}-dev":
 
 ```bash
-git checkout master -b "v${VER}-dev"
-vi _version.py   # Remove .dev from version
-git add _version.py
-git commit -m "bla bla"
+git checkout -b "v${VER}-dev"
+vi ./cirq/_version.py   # Remove .dev from version
+git add ./cirq/_version.py
```
dstrain115, line 90: I don't think you need to add it, since the file already exists, correct?

release.md
balopat: We create the release branch later. However, I could clarify: "Make sure you checked out the version you want to patch. Typically this will be something like `v${MAJOR}.${MINOR}.${LAST_PATCH}`"

balopat: Oh, also, I think there might be a bit of misunderstanding here - I believe the tag is a better starting point than the release branch for the last version, because the branch is mutable and tags aren't (well, technically are, but shouldn't be).

release.md
balopat: I agree that this is risky - especially after 1.0 this might be a real situation. I am just going to remove this part and introduce the variable for "$VERSION_TO_PATCH".

release.md
balopat: `git add` is required even if the file is already tracked by git, this adds select modifications to the index, so when using `git commit -m "message"`, only the `git add`-ed changes are committed. It's a conservative approach compared to `git commit -am "message"` which is equivalent to "add modifications to _all_ changed tracked files to the index" - which could result accidentally committing other files if you accidentally changed them in the repo.

dabacon:
@googlebot cla is a-okay

Issue #2957, opened 2020-05-03
Change default serializable gate set in Engine from XMON


dabacon:
1. This isn't documented in doc strings
2. Should this be upgraded to something more modern?
3. I assume lots of uses are having to specify this for every call, is this something that we should have an Engine level default for?  I think we have steered away from that pattern.

maffoo:
Internally, we have code where we default the gateset to `None`. When a circuit is submitted and no gateset is specified, we try each gateset that the engine says it supports in turn until we find one that can serialize the given circuit and then use that. In practice, the first gateset that the engine claims to support should be the full "fsim" gateset, which can basically serialize any allowed circuit, so there's little overhead. This could possibly be made cleaner if there were a way to ask a gateset whether it can serialize a given circuit instead of just attempting to serialize and capturing any exceptions.

PR #3138, opened 2020-07-14
Remove XMON as engine default

dstrain115:
- Engine now raises ValueError if you don't specify gate_set.
- While technically backwards incompatible, I don't think anyone
will have working code that doesn't specify a gate_set here.

Solves #2957

### DIFF
--- a/cirq/google/engine/engine.py
+++ b/cirq/google/engine/engine.py
@@ -32,7 +32,6 @@ from typing import Dict, List, Optional, Sequence, TypeVar, Union, TYPE_CHECKING
 from google.protobuf import any_pb2
 
 from cirq import circuits, study, value
-from cirq.google import gate_sets
 from cirq.google import serializable_gate_set as sgs
 from cirq.google.api import v1, v2
 from cirq.google.engine import (engine_client, engine_program, engine_job,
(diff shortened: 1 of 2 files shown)


dstrain115:
> We should somehow add this to a "breaking changes" list that we can point out in the next release. We could add a "staging" area for release notes for the next planned minor release where we can build up these deprecation notices.

I will leave that up to a later PR, since we don't yet have a current place to put that.  If you have a suggestion on where to add it, I can put up a PR to add a first draft of the "upcoming release" notes.

dabacon:
I guess the question is why have a default if you have to specify it now?

dstrain115:
> I guess the question is why have a default if you have to specify it now?

I am not sure I understand the question.  Do you mean: why have a default of `None`?  I think we need that in order to make it a named parameter instead a positional parameter.

balopat:
> > We should somehow add this to a "breaking changes" list that we can point out in the next release. We could add a "staging" area for release notes for the next planned minor release where we can build up these deprecation notices.
> 
> I will leave that up to a later PR, since we don't yet have a current place to put that. If you have a suggestion on where to add it, I can put up a PR to add a first draft of the "upcoming release" notes.

Sounds good - this is my suggestion - https://github.com/quantumlib/Cirq/issues/3143 - as soon as I'll get around it, I'll create it.

Issue #3116, opened 2020-07-01
Upgrade float to double in cirq protos


jaeyoo:
Hi, I am suffering from the precision issue due to google protobuf.

In fact, google protobuf float is float32, and double is float64. However, python uses float only for float64, and this makes discrepancy when I saved into protobuf and reload the circuit from the proto. As it happens to others (https://github.com/quantumlib/Cirq/issues/2728, for example) I suggest that upgrading the float to double in cirq protos.

For example,
https://github.com/quantumlib/Cirq/blob/master/cirq/google/api/v2/program.proto#L146
```
float float_value = 1; # -> double float_value = 1;
```
OR at least, I hope to have `double_value` as one of the variables,

https://github.com/quantumlib/Cirq/blob/master/cirq/google/api/v2/program.proto#L145-L149
```
oneof arg_value {
    float float_value = 1;
    RepeatedBoolean bool_values = 2;
    string string_value = 3;
    double double_value = 4; # <- like this.
  }
```

dabacon:
Is this coming from Tensorflow quantum integration?  Changing double to float isn't really an option without breaking the api, but we could add a double value.

jaeyoo:
Yes, it happens when I tested with TFQ. Thank you for considering the second option!

dabacon:
I think we need to 
1. add double value to proto
2. make `_arg_from_proto` support double value
3. get `_arg_from_proto` version of cirq out to dependencies
4. make `_proto_from_arg` support double value (should ints be made to doubles?)

@maffoo

dabacon:
@dstrain115 for comment as well.

maffoo:
Note that we also use floats in the sweep protos for Points and Linspace: https://github.com/quantumlib/Cirq/blob/master/cirq/google/api/v2/run_context.proto#L104.
We might want to modify those as well to allow specifying doubles for people who want the higher precision.

I don't feel particularly strongly either way; it seems like the few tests that fail that can be handled with approximate equality (or might be fixed with proto 3.12 which truncates floats differently) and 32-bit floats are plenty of precision for the hardware control. On the other hand, using 64-bit doubles might reduce some surprises at the cost of a little more complexity in handling protos.

balopat:
Can we clarify the problem here a little bit? 
@jaeyoo so you are serializing and deserializing a Circuit and that doesn't end being equal to the original Circuit? Can you provide example code? 
Can you provide some clarity around what use case does this influence in TFQ?

[...]

PR #3140, opened 2020-07-15
Create double_val in oneof for program

dstrain115:
- Creates an option to store double as an ArgValue for serialization.
- This is currently only one-way for backwards compatibility.
- The conversion of python floats will still go to float_value.


balopat:
I assume this is to help with https://github.com/quantumlib/Cirq/issues/3116 - but it doesn't really yet, right because we are not converting python floats to proto double? What's the plan/rationale for introducing this double field then?

dstrain115:
> I assume this is to help with #3116 - but it doesn't really yet, right because we are not converting python floats to proto double? What's the plan/rationale for introducing this double field then?

That's correct.  We need to do this in two steps for backwards compatibility.  We first introduce the field and the code to handle it.

Once this code is fully pushed out, then we can start converting python floats into the new double_val.

Make sense?  I welcome alternatives if you have a different idea to preserve compatibility.

balopat:
Yeah this is not possible with protobuf 3 gracefully. We are essentially changing the type of a field which is not backward compatible in this case. 

Have we considered introducing a v3 api for this? Then we have the option to fully implement the proper solution. Clients have the option to stay on v2 or switch over to v3 at their own leisure.

dstrain115:
> Yeah this is not possible with protobuf 3 gracefully. We are essentially changing the type of a field which is not backward compatible in this case.

Upgrading to v3 is a way way way way bigger change.  I don't think it is worth it for this change.
I mean, I think the plan would be to support deserializing here, then start serializing into this field in cirq 0.9.0.

The only thing we need to worry about is people deserializing programs with an old version, and I think that would be really unlikely, since clients would deserialize results, not programs.  Only "servers" would need to deserialize programs, so we really only need to worry about QE and TFQ servers to upgrade here.

What do you think?

balopat:
> > Yeah this is not possible with protobuf 3 gracefully. We are essentially changing the type of a field which is not backward compatible in this case.
> 
> Upgrading to v3 is a way way way way bigger change. I don't think it is worth it for this change.
> I mean, I think the plan would be to support deserializing here, then start serializing into this field in cirq 0.9.0.
> 
> The only thing we need to worry about is people deserializing programs with an old version, and I think that would be really unlikely, since clients would deserialize results, not programs. Only "servers" would need to deserialize programs, so we really only need to worry about QE and TFQ servers to upgrade here.
> 
> What do you think?

That makes a lot of sense to me, thanks for explaining! LGTM!

PR #3141, opened 2020-07-16
[Pasqal updates] PasqalDevice class generalization

HGSilveri:
This the second child-PR of #3126 , following #3133 .

 - Generalizes `PasqalDevice`, removing ties to the physical placement of the qubits.
 - `PasqalDevice` now serves as the parent class of all future Pasqal devices, only enforcing constraints expected to exist on all future devices.
 - `PasqalDevice`can also be used as the device for submission of explicitly unconstrained circuits to Pasqal's devices.

A new device class, `PasqalVirtualDevice`, will subclass the new `PasqalDevice` and recover the hardware constraints of the current `PasqalDevice` on a future PR.

### DIFF
--- a/cirq/pasqal/pasqal_device.py
+++ b/cirq/pasqal/pasqal_device.py
@@ -11,51 +11,35 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import Any, Dict, FrozenSet, Iterable
-from numpy import sqrt
+from typing import FrozenSet, Sequence
 
 import cirq
-
-from cirq.pasqal import ThreeDGridQubit
+from cirq.ops import NamedQubit
 
 
 @cirq.value.value_equality
 class PasqalDevice(cirq.devices.Device):
-    """A Pasqal Device with qubits placed on a 3D grid."""
+    """A generic Pasqal device."""
 
-    def __init__(self, control_radius: float,
-                 qubits: Iterable[ThreeDGridQubit]) -> None:
-        """ Initializes the description of the device.
+    def __init__(self, qubits: Sequence[cirq.ops.Qid]) -> None:
+        """Initializes a device with some qubits.
 
         Args:
-            control_radius: the maximum distance between qubits for a controlled
-                gate. Distance is measured in units of the indices passed into
-                the qubit constructor.
-            qubits: Qubits on the device, identified by their x, y, z position.
-                Must be of type ThreeDGridQubit.
-
+            qubits (NamedQubit): Qubits on the device, exclusively unrelated to
+                a physical position.
         Raises:
-            ValueError: if the wrong qubit type is provided or if invalid
-                parameter is provided for control_radius. """
-        us = cirq.value.Duration(micros=1)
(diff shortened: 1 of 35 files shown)


HGSilveri:
Hi @balopat ! Here is the second PR. Also, I was wondering: is it okay to do two PR's simultaneously if they are unrelated? I suppose it is not recommended, I was just wondering since it could help speed up the process.

balopat:
> Hi @balopat ! Here is the second PR. Also, I was wondering: is it okay to do two PR's simultaneously if they are unrelated? I suppose it is not recommended, I was just wondering since it could help speed up the process.

Absolutely, feel free to open multiple ones! I'll have a look at this one today!

cirq/pasqal/pasqal_device.py
```diff
@@ -11,52 +11,45 @@
...
 import cirq
-
-from cirq.pasqal import ThreeDGridQubit
+from cirq.ops import NamedQubit
 
 
 @cirq.value.value_equality
 class PasqalDevice(cirq.devices.Device):
-    """A Pasqal Device with qubits placed on a 3D grid."""
+    """A generic Pasqal device."""
```
balopat, line 22: A bit more documentation here would be very helpful (you can just rephrase a bunch of the text you had in the documentation) for users to understand the intent of this class: When is this useful? Why would someone want to use an unconstrained Pasqal device vs the 3D one?

cirq/pasqal/pasqal_device.py
```diff
@@ -11,52 +11,45 @@
...
-        self._max_parallel_xy = 2
-        self._max_parallel_c = 10
-        self._max_parallel_t = 1
+            TypeError: if the wrong qubit type is provided.
+        """
 
         for q in qubits:
-            if not isinstance(q, ThreeDGridQubit):
+            if not isinstance(q, self.supported_qubit_type):
                 raise TypeError('Unsupported qubit type: {!r}'.format(q))
```
balopat, line 36: To make the error message more helpful, you could list the supported qubit types.

cirq/pasqal/pasqal_device.py
```diff
@@ -11,52 +11,45 @@
...
-        self.control_radius = control_radius
         self.qubits = qubits
 
+    @property
+    def supported_qubit_type(self):
+        return (NamedQubit,)
+
+    @property
+    def maximum_qubit_number(self):
+        return 100
```
balopat, line 51: we are so in the NISQ era ;)

cirq/pasqal/pasqal_device.py
```diff
@@ -99,96 +92,56 @@ def validate_operation(self, operation: cirq.ops.Operation):
...
-        if not PasqalDevice.is_pasqal_device_op(operation):
+        if not self.is_pasqal_device_op(operation):
             raise ValueError('{!r} is not a supported '
                              'gate'.format(operation.gate))
 
         for qub in operation.qubits:
-            if not isinstance(qub, ThreeDGridQubit):
-                raise ValueError('{} is not a 3D grid qubit '
+            if not isinstance(qub, self.supported_qubit_type):
+                raise ValueError('{} is not a valid qubit '
```
balopat, line 106: Similar to __init__, please clarify what are the accepted types for this device.

PR #3142, opened 2020-07-16
Use device instead of gateset validation in ZerosSampler

dabacon:
This introduces a dependency of cirq on cirq.google, which we want to avoid.  If we want a gate set validating version of this we should put it in cirq.google.


cirq/work/zeros_sampler.py
```diff
@@ -13,27 +13,16 @@
...
 
-    def __init__(self, gate_set: 'cirq.google.SerializableGateSet' = None):
-        """
-        Args:
-            gate_set: `SerializableGateSet`. If set, sampler will validate that
-                all gates in the circuit are from the given gate set.
-        """
-        self.gate_set = gate_set
-
     def run_sweep(
```
dstrain115, line 26: I am fine with getting rid of validation altogether, but one alternative I just thought of was to add an optional device parameter.  That could also validate the circuit in a more natural way.

dabacon: OK added device validation.  Could also just do this by adding it to a circuit.

Issue #2972, opened 2020-05-04
Sampler should document big endian nature of Pandas dataframe


dabacon:
We document this in `TrialResul` but no in sample.  Documentation should also reference unpacking and packing methods like `cirq.big_endian_bits_to_int`

dstrain115:
Closed via #3146

PR #3146, opened 2020-07-17
Adding documentation about big endian values

dstrain115:
- Adjusting documentation about big endian representations
for TrialResult and Sampler.
- Solves #2972


cirq/work/sampler.py
```diff
@@ -69,9 +69,13 @@ def sample(
...
         Returns:
             A `pandas.DataFrame` with a row for each sample, and a column for
-            each measurement result as well as a column for each symbolic
-            parameter. There is an also index column containing the repetition
-            number, for each parameter assignment.
+            each measurement key as well as a column for each symbolic
+            parameter.  Measurement results are stored as a big endian integer
+            representation with one bit for each measured qubit in the key.
+            See `cirq.big_endian_int_to_bits` and similar functions for how
+            to convert this integer into bits.
```
viathor, line 76: You could also reference the second example below which illustrates endianness of the representation of measurement results.

PR #3155, opened 2020-07-20
Fix calibration metric names in docs

dabacon:
The two qubit metrics had extra "cycle" in them.  

I confirmed these names with those currently returned from a device.  I also double checked it with the name on the internal spreadsheet.


dabacon:
> Thanks for catching this!

Copy-pasta failure FTW :)

PR #3157, opened 2020-07-20
Adds getitem to TrialResult

### LINKED ISSUE #1288: Define __getitem__ for TrialResult
Should obtain measurements.

dstrain115:
- returns the measurements with that key.
- Fixes #1288


cirq/study/trial_result.py
```diff
@@ -277,6 +281,12 @@ def __eq__(self, other):
             return NotImplemented
         return self.data.equals(other.data) and self.params == other.params
 
+    def __getitem__(self, key: str) -> np.ndarray:
+        return self._measurements[key]
```
viathor, line 285: Regarding the choice of `measurements` vs `data` (discussed on #1288), I think we should choose the one that is the more common case. My own experience is that the pandas DataFrame is easier to use (because it is more self-documenting), so it would be my preference.

@dabacon @kevinsung @Strilanc WDYT?

dstrain115: That would trigger the creation of the data frame.  I am not sure how expensive an option that is, but it would force the creation of the whole ndarray though.  If people are going to do that anyway though, maybe it doesn't matter.

viathor: We're caching the DataFrame, so it's only going to hit the user once.

Also, DataFrame is well-suited for interactive use (e.g. in ipython, colab etc) while ndarray is a better choice in code where we care about performance. In the latter - since it's generally non-interactive - one can easily write out the longer reference like `result.measurements[key]`, but in interactive use it's nice to be able to write the short `result[key]` and get a DataFrame.

cirq/study/trial_result_test.py
```diff
@@ -45,6 +45,20 @@ def test_str():
...
+    results_c = np.array([[0], [1], [1], [0]])
+    result = cirq.TrialResult.from_single_parameter_set(
+        params=cirq.ParamResolver({}),
+        measurements={
+            'ab': results_ab,
+            'c': results_c
+        })
+    assert len(result) == 2
+    assert (result['ab'] == results_ab).all()
+    assert (result['c'] == results_c).all()
```
dabacon, line 59: nit: consistency with most other tests that use np.all(arr1 == arr2).

dabacon:
Re: dataframe versus array.   My main worry is not so much time here, but that there are considerably less people familiar with pandas than numpy.  Or at least that is my impression.

viathor:
@dabacon I'm not convinced that this is a valid argument. If we accept it as valid then we may want to reconsider integrating pandas in cirq in general, not just in the context of this particular PR.

dabacon:
> @dabacon I'm not convinced that this is a valid argument. If we accept it as valid then we may want to reconsider integrating pandas in cirq in general, not just in the context of this particular PR.

There are many other examples of places where we return objects that aren't familiar to a lot of people, but this doesn't preclude us of giving the option to return them (example: QASM).

Note that the original filer of the issue thought that this ambiguity meant that maybe we shouldn't do either of these.  From #1288
> Well, we could also make trial_result[key] return trial_result.data[key] instead of trial_result.measurements[key]. The return type would then be a Pandas Series. Now that I think of it, this ambiguity makes a case for not implementing this.

dabacon:
FWI: https://theatlas.com/charts/rJ9sZ5syf

viathor:
Haven't we already decided that pandas is sufficiently familiar to our users in allowing DataFrames to be returned from `Sampler.sample`? Should we revisit that API? Note that it is probably early enough since there aren't many other places where we use pandas in the API.

Regarding @kevinsung's comment: I agree the ambiguity is a good reason not to do this at all. This raises the question about alternatives. WDYT about two methods (in `TrialResult`): `as_array` and `as_dataframe` (each accepting a measurement key, like `__getitem__` in the present PR)?

dstrain115:
Due to disagreements and ambiguity on what __getitem__ would do, if implemented, I am withdrawing this PR from consideration and will close the bug once we revert the similar change in EngineJob.

PR #3159, opened 2020-07-21
CI check for uncommitted generated proto files

### LINKED ISSUE #3087: build_proto ci check does not enforce that generated proto code is up to date
We recently added cirq/google/api/v2/batch.proto and the ci check did not catch that generated code was not checked in. Original PR: #3064; follow-up to add generated code: #3086.

balopat:
Fixes #3087.

demo: 
Unnamed screen capture (1)


check/build-changed-protos
```diff
@@ -94,6 +94,17 @@ do
     (set -x; bazel build "${base}_proto"; \
     bazel build "${base}_py_proto"; \
     bazel build "${base}_cc_proto")
+
+    python -m grpc_tools.protoc -I=. --python_out=. --mypy_out=. ${base}.proto
 done
 
+untracked=$(git status --porcelain 2>/dev/null| grep "^?? cirq/google" | cut -d " " -f 2)
```
viathor, line 101: nit: add space before the first |

viathor, line 101: We should either make this (and the loop below) work with files that may include space in the name or add a comment saying we don't support this.

balopat: I added the comment!

balopat: Thanks!

PR #3160, opened 2020-07-21
[Pasqal updates] Updated gate set

HGSilveri:
This the third child-PR of #3126 , following #3141 .

- The accepted gate set is updated to:
  - Accept only specific multi-qubit gates
  - Accept the Hadamard gate
- The noise model is updated accordingly and is now device dependent, in preparation for the addition of new devices whose noise model will naturally differ from the standard
- A specific converter for non-native gate decomposition is adapted from the `neutral_atoms` module


balopat:
Sorry for the wait on this @HGSilveri - I was distracted in the last two days with an event. Will be back to this in the coming days.

HGSilveri:
No worries @balopat , thanks for all your help! By the way, I believe that the failed CI check is unrelated to this PR.

cirq/pasqal/pasqal_noise_model.py
```diff
@@ -67,16 +74,15 @@ def noisy_moment(self, moment: cirq.ops.Moment,
...
         return list(moment) + noise_list
 
+    def get_op_string(self, cirq_op: cirq.ops.Operation) -> str:
+        """Find the string representation for a given operation
+        Args:
+            cirq_op: one cirq operation
+        Returns:
+            String representing the gate operations
+        """
+        if not self.device.is_pasqal_device_op(cirq_op) \
```
balopat, line 84: nit: prefer parentheses over `\` newline separator

cirq/pasqal/pasqal_device.py
```diff
@@ -77,21 +78,31 @@ def decompose_operation(self,
...
+                               cirq.ops.PhasedXPowGate, cirq.ops.XPowGate,
+                               cirq.ops.YPowGate, cirq.ops.ZPowGate))
+
+        if not valid_op:  # To prevent further checking if already passed
+            if isinstance(
+                    op.gate,
+                (cirq.ops.HPowGate, cirq.ops.CNotPowGate, cirq.ops.CZPowGate,
+                 cirq.ops.CCZPowGate, cirq.ops.CCXPowGate)):
+                expo = op.gate.exponent
+                valid_op = np.isclose(expo, np.around(expo, decimals=0))
```
balopat, line 103: How would this work with parametrized gates? The type of exponent is `Union[float, sympy.Basic]` - it could be parametrized. I think you should make sure you're handling that here and cover that flow with tests.

HGSilveri: You're right, I made the appropriate changes.

cirq/pasqal/pasqal_device.py
```diff
@@ -77,21 +79,32 @@ def decompose_operation(self,
...
+                               cirq.ops.PhasedXPowGate, cirq.ops.XPowGate,
+                               cirq.ops.YPowGate, cirq.ops.ZPowGate))
+
+        if not valid_op:  # To prevent further checking if already passed
+            if isinstance(
+                    op.gate,
+                (cirq.ops.HPowGate, cirq.ops.CNotPowGate, cirq.ops.CZPowGate,
+                 cirq.ops.CCZPowGate, cirq.ops.CCXPowGate)):
+                expo = op.gate.exponent
+                if not isinstance(expo, sympy.Basic):
```
balopat, line 104: 1) Use `cirq.is_parametrized(op)` - that should take care of other parameters as well. 
2) Is the Pasqal device going to handle non-integer exponent and/or parametrized gates from PhasedXPowGate, XPowGate, YPowGate, ZPowGate?

HGSilveri: 1. Ok, will do.
2. Yes, we'll handle non-integer exponents for these. As for the parametrization, we won't necessarily handle that, but we support the parameter substitution for actual values in `PasqalSampler.run_sweep`, so we have to allow those here.

PR #3161, opened 2020-07-21
[Pasqal updates] Removing ThreeDGridQubit

HGSilveri:
This the fourth child-PR of #3126 , following #3141  and in parallel with #3160 .

`ThreeDGridQubit` is removed from the `pasqal` module, as all dependencies have been eliminated and it can be fully replaced by `ThreeDQubit`.


HGSilveri:
Hi @balopat ! Thank you very much for all your help so far, I'm happy to say we are nearly done. This PR should be very straightforward to merge, #3160 probably not so much but it is still okay. Please take a look whenever you can, thanks!

PR #3163, opened 2020-07-22
Remove q from calibrations.py

### LINKED ISSUE #2963: Update code for when Engine metric qubits are formatted differently
There is currently code in calibration.py that looks at qubit ids that start with `q` and removes this `q` before getting the grid qubit.  If the API no longer returns `q` prefixed code this should be removed, otherwise if we are going to continue to support this we should add it to `grid_qubit_from_proto`.

dstrain115:
- Qubits from engine have not had the 'q' prefix
in a while.  Removing this prefix check from the
calibration code.

- Fixes #2963


dabacon:
Also I cannot resist.  "There is no q in calibrations".

Issue #3170, opened 2020-07-23
`final_wavefunction` is deprecated wrong


mpharrigan:
If you use `cirq.final_wavefunction` it says:

```
  /.../: DeprecationWarning: final_state_vector was used but is deprecated.
  It will be removed in cirq v0.10.0.
  Use `cirq.final_state_vector` instead.
```

mpharrigan:
there is a test: `mux_test:test_deprecated` that must be broken as well. It uses `cirq.testing.assert_logs`

balopat:
This is an issue with the deprecation function's text itself - I'll have a look

E.g.

```
>>> cirq.QFT(*cirq.LineQubit.range(3))
<input>:1: DeprecationWarning: qft was used but is deprecated.
It will be removed in cirq v0.10.0.
Use cirq.qft instead.
```

balopat:
The tests are just matching on the substring of the new function name to be there but not the structure of the whole text.

mpharrigan:
the mux test deprecated searches for both the new name *but also* the old function name which *isn't* (wasn't ... I see you've already fixed the underlying issue) in the message

Issue #1985, opened 2019-08-19
`_Pauli{X,Y,Z}` maybe shouldn't be private


mpharrigan:
Unlike every other gate, with `cirq.X`, `cirq.Y`, and `cirq.Z`, it's very easy to end up with an instance of a class that is not importable from top-level-`cirq`

```python
x1 = cirq.X
x2 = cirq.X**0.99
type(x1)
Out[6]: cirq.ops.pauli_gates._PauliX
type(x2)
Out[7]: cirq.ops.common_gates.XPowGate
```

consider renaming to `PauliX` and exposing as `cirq.PauliX`

mpharrigan:
Actually, this happens with a lot of singletons and is probably by design. Running into this when trying to automate testing for #1880

dabacon:
Do we know this is the right pattern?  We don't use singletons anywhere else and this is a pattern that may confuse a lot of people.

mpharrigan:
There's a couple other places:

```python
import inspect
import cirq

def _get_all_public_things():
    for cls_name, cls_cls in inspect.getmembers(cirq):
        if inspect.isfunction(cls_cls) or inspect.ismodule(cls_cls):
            continue

        if not inspect.isclass(cls_cls):
            # singletons, for instance
            cls_cls = cls_cls.__class__

        yield cls_name, cls_cls


for cls_name, cls_cls in _get_all_public_things():
    if cls_cls.__name__.startswith('_'):
        print(f'{cls_name:20s} {cls_cls.__name__:20s}')
```
NO_NOISE             _NoNoiseModel       
OP_TREE              _Union              
ParamResolverOrSimilarType _Union              
PauliSumLike         _Union              
QubitOrderOrList     _Union              
Sweepable            _Union              
UnconstrainedDevice  _UnconstrainedDeviceType
UnitSweep            _Unit               
X                    _PauliX             
Y                    _PauliY             
Z                    _PauliZ         
```

Strilanc:
I would omit the _Union types from consideration, since they are just shorthands for type annotations.

I am not opposed to making the singleton value types public. They pollute the autocomplete namespace with strings that are not useful to users, but on the otherhand it makes sense that if an instance is public then its type should also be public.

mpharrigan:
For the purposes of #1880 automated testing, it no longer matters to me. 

The `_Union` types are not ours, but shows that even the python-powers-that-be sometimes gives you an instantiation of a private class

[...]

PR #3171, opened 2020-07-23
Add TensorProductState, a state-vector-like object

mpharrigan:
This is factored out of #2781 with @Strilanc 's initial comments addressed, per comments on that PR.

`cirq.STATE_VECTOR_LIKE` accepts basis-state indices, sequences of per-qudit basis indices, or arrays of amplitudes. This PR extends this union type to accept `TensorProductState` -- an object representing a state vector which can be decomposed into a tensor product of single-qubit states. The structure shares a lot of similarity with `PauliString`. Whereas PauliString is a dictionary from qubits to (single-qubit) `Pauli` objects; here it is a mapping from qubits to single-qubit named states. Here we include `|0>, |1>, |+>, |->, |i>, |-i>`. As an example:

```python
# start from the |+++> state
q = cirq.LineQubit.range(3)
initial_state = cirq.KET_PLUS(q[0]) * cirq.KET_PLUS(q[1]) * cirq.KET_PLUS(q[2])
```

Since most things already use `STATE_VECTOR_LIKE` and its associated helper function `to_valid_state_vector`, only minor modifications to this function was necessary to enable specification of a tensor product state wherever initial states are sold.


cirq/sim/mux.py
```diff
@@ -161,9 +156,11 @@ def final_state_vector(
     return cast(sparse_simulator.SparseSimulatorStep, result).state_vector()
 
 
-final_wavefunction = deprecated(
+@deprecated(
     deadline='v0.10.0',
-    fix='Use `cirq.final_state_vector` instead.')(final_state_vector)
+    fix='Use `cirq.final_state_vector` instead.')
+def final_wavefunction(*args, **kwargs):
+    return final_state_vector(*args, **kwargs)
```
mpharrigan, line 163: hacky fix of #3170

mpharrigan:
The mypy check is pulling the wrong commit ..

cirq/sim/mux.py
mpharrigan: No apologies necessary; thanks for the fix!

cirq/value/tensor_product_state.py
```diff
@@ -0,0 +1,345 @@
...
+            'states': list(self.states.items()),
+        }
+
+    @classmethod
+    def _from_json_dict_(cls, states, **kwargs):
+        return cls(states=dict(states))
+
+    def state_vector(self, qubit_order: 'cirq.QubitOrder' = None):
+        """The state-vector representation of this state."""
+        from cirq import ops
```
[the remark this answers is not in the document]
mpharrigan: circular dependency; specifically `value_equality` lives in `value` and is used in a lot of places (and I've put this code in `value`, which seemed like the best fit to me)

cirq/value/tensor_product_state.py
```diff
@@ -0,0 +1,345 @@
...
+
+        mat = 1.0 + 0.j
+        for qubit in qubits:
+            oneq_state = self[qubit]
+            oneq_proj = oneq_state.projector()
+            mat = np.kron(mat, oneq_proj)
+        return mat
+
+
+class _KetPlus(_Named1QState):
```
tanujkhattar, line 174: What is the benefit of having `_Named1QState` as an abstract base class and then defining each of the derived classes individually over, say, the `_Named1QState` base class taking in 4 arguments (str, repr, state_vector, stabilized_by) and then directly defining `KET_PLUS` etc. as objects of the `_Named1QState` base class? The later has an advantage of shorter code and reduced boilerplate. 

IIUC, the derived classes `_KetPlus` etc. are not really singletons here (`id(cirq.KET_PLUS) != id(cirq.value.tensor_product_state._KetPlus())`) but the use of `__class__` in the base class makes it work fine. Since we are anyways exposing only the final initialized variables, I'm curious to understand the benefits of this approach versus differentiating the different objects simply based on parameters passed to the base class, as described above.

tanujkhattar: NVM, found the answer here: https://github.com/quantumlib/Cirq/pull/2781#discussion_r459166296

> big problem: stabilized by needs to import cirq.ops which would cause a circular import because of value_equality.

This could be addressed by passing an index to `_Named1QState` and using `ops.Pauli.by_index()`

> Small problem: we'd be initializing the arrays at import time. We don't do this anywhere else

Can you provide link to relevant reading material which explains the difference? (tried searching for relevant keywords but couldn't find anything relevant).

cirq/value/tensor_product_state.py
```diff
@@ -0,0 +1,345 @@
...
+
+        The returned object is a TensorProductState of length 1.
+        """
+        return TensorProductState({qubit: self})
+
+    def __call__(self, *args, **kwargs):
+        return self.on(*args, **kwargs)
+
+    @abc.abstractmethod
+    def state_vector(self):
```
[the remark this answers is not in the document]
mpharrigan: This is an interesting idea and I think it hints at the overall reality which is: there are a lot of ways you can unambiguously specify a single qubit state! What I had in mind is that `TensorProductState` can be built with any objects that expose a `state_vector` property. Right now, the only such objects are `_Named1QState`s (because this PR is built with #2781 in mind) and there's utility in having an explicitly enumerated list of possibilities. 

Using a circuit to specify the state could work but I'd argue just because you could doesn't mean you should :) I think calling something the plus state and being done with it is easier than saying "it's the state you get by starting in |0> and doing a Hadamard"! 

Like I said, though, there's ample opportunity to expand the scope of states further once we have a use case

cirq/value/tensor_product_state.py
mpharrigan: Thanks for linking to the previous thread. I'll think about this suggestion very seriously since you and @Strilanc both raised it :) 

Like I said, this whole file mimics the structure of the Pauli gates (i.e. `cirq.X`)

> IIUC, the derived classes _KetPlus etc. are not really singletons here
Sure; this is a Python limitation as far as I can tell. They're idiomatic singletons

> This could be addressed by passing an index to _Named1QState and using ops.Pauli.by_index()

This is true and something I will seriously consider. I think then I'd have to actually refactor `_Named1QState` into an abstract base class `OneQubitState` and the suggested class `PauliOneQubitState`

> Can you provide link to relevant reading material which explains the difference?

I suppose you could briefly read how Python's import system works. In summary: when you import a "module" the file is executed in its entirety. Because most files are just a bunch of function `def`s, you don't see this in action. But! this behavior is why you need to put `if __name__ == '__main__'` in scripts; otherwise every time you import that module it would execute the script! If you have a global constant e.g. `MY_ARRAY = np.zeros(50000)` it would actually allocate the array when you `import cirq` instead of how gates (and current states) work when an array is only allocated when you call `unitary()` or `state_vector()`. Obviously these arrays for states are tiny, so it's probably not a huge deal in practice but I was just noting that we don't do this elsewhere. For example, the X gate is *not* defined as

```python
X = MatrixGate(np.array([[0,1],[1,0]]))
```

mpharrigan: here's a strawman implementation
```python
# @value_equality
class PauliOneQubitState(OneQubitState):
    def __init__(self, pauli:'cirq.Pauli', eigenvalue:int):
        if eigenvalue not in [-1, 1]:
            raise ValueError("Eigenvalue must be -1 or +1")

        self.pauli = pauli
        self.eigenvalue=eigenvalue

        return (self.pauli, self.eigenvalue)

    def _json_dict_(self):
        return protocols.obj_to_dict_helper(self, ['pauli', 'eigenvalue'])

    def __str__(self):
        pm = '+' if self.eigenvalue == 1 else '-'
        return f'{pm}{self.pauli}'

    def state_vector(self):
        from cirq.ops.pauli_gates import _PauliX, _PauliZ
        if isinstance(self.pauli, _PauliX):
            if self.eigenvalue == 1:
                return np.array([1, 1]) / np.sqrt(2)
            elif self.eigenvalue == -1:
                return np.array([1, -1]) / np.sqrt(2)
            else:
                raise ValueError("Eigenvalue must be -1 or +1")
        if isinstance(self.pauli, _PauliZ):
            if self.eigenvalue == 1:
                return np.array([1,0])
            elif self.eigenvalue == -1:
                return np.array([0,1])
            else:
                raise ValueError("Eigenvalue must be -1 or +1")
        # ...
```

at the end of the day we need to explicitly write out the state vectors for these states just like how we have to write the unitaries for named gates somewhere, so I'm still leaning towards modeling this after how `cirq.X` is made

cirq/value/tensor_product_state.py
```diff
@@ -0,0 +1,345 @@
...
+import numpy as np
+
+from cirq import protocols
+from cirq._doc import document
+
+if TYPE_CHECKING:
+    import cirq
+
+
+class _Named1QState(metaclass=abc.ABCMeta):
```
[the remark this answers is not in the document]
mpharrigan: done, pending further discussion below

cirq/value/tensor_product_state.py
```diff
@@ -0,0 +1,345 @@
...
+        return np.array([0, 1])
+
+    def stabilized_by(self):
+        from cirq import ops
+        return -1, ops.Z
+
+
+KET_PLUS = _KetPlus()
+document(
+    KET_PLUS, """The |+> State
```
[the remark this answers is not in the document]
mpharrigan: very fancy; done

cirq/value/tensor_product_state_test.py
```diff
@@ -0,0 +1,175 @@
...
+                      cirq.X(q1)]))
+
+    s01 = cirq.KET_ZERO(q0) * cirq.KET_ONE(q1)
+    psi2 = cirq.final_state_vector(cirq.Circuit([cirq.I.on_each(q0, q1)]),
+                                   initial_state=s01)
+
+    np.testing.assert_allclose(psi1, psi2)
+
+
+def test_tp_projector():
```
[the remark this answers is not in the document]
mpharrigan: added

cirq/value/tensor_product_state.py
tanujkhattar: ^ Thanks for the explanation! For some reason, I completely overlooked the `state_vector` arrays in argument and was thinking about `PAULI_STATES` list below. My bad.

Regarding the implementation above, if we have strong opinions about not initialising the `state_vector` array at import time, I would also prefer the original implementation similar to `cirq.X` instead of the above one where we check all cases in `state_vector` function. 

My original suggestion was mainly to reduce code redundancy. To get around the problem of allocating numpy arrays, we could also wrap the array in a lambda while passing it as an argument to the class.  eg: 

```python
KET_PLUS = _Named1QState(
    name='KET_PLUS',
    str='+X',
    state_vector = lambda: np.array([1,1])/np.sqrt(2),  # state_vector wrapped in a lambda to avoid allocation at import time.
    stabilized_by=(+1, 0) # 0 instead of ops.X to avoid circular import
)
```


Again, I don't have very strong opinions here. Maybe @Strilanc can chime in since he also originally pointed this out.

mpharrigan: yeah your solution looks nice as well. I don't have strong preferences. Consensus should be reached on these questions:

1. Do we actually care that the current implementation is fairly verbose and produces singletons that aren't so single?
2. Do we actually care to avoid making numpy arrays at import time?
3. Do we mind that passing a pauli operator index instead of using `ops.X` is mildly less readable?

Maybe we can discuss briefly at cirq cync if people don't comment here

mpharrigan: Discussed over video call: going to stick with current implementation but make the base class public so others can inherit from it. (side note: maybe we should make the pauli base class public (for gates) #1985)

mpharrigan:
@viathor to check for forwards-compatibility with qudits

mpharrigan:
@viathor do you know when you might have time to look at this?

cirq/value/tensor_product_state.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def __repr__(self):
+        return 'cirq.KET_IMAG'
+
+    def state_vector(self):
+        return np.array([1, 1j]) / np.sqrt(2)
+
+    def stabilized_by(self):
+        from cirq import ops
+        return 1, ops.Y
```
viathor, line 220: The more I think about it the more I'm convinced that we should avoid defining a class to represent every interesting quantum state. Instead, we should define a class for each interesting basis.

The core of the argument is that all states in a given basis are so similar that the classes end up nearly identical. Repeating them is unnecessary and results in verbosity. Admittedly, verbosity inside the library is a minor issue. In this case, however, the building blocks we offer are so low level that the client code ends up verbose, too.

For example, suppose we wish to represent the state stabilized by XII, IYI and -IIZ. Using the code in this PR, we would say

```python
qs = cirq.LineQubit.range(3)
state = cirq.KET_PLUS(qs[0]) * cirq.KET_IMAG(qs[1]) * cirq.KET_ONE(qs[2])
```

A possible alternative may be:

```python
qs = cirq.LineQubit.range(3)
state = cirq.X.basis[+1] * cirq.Y.basis[+1] * cirq.Z.basis[-1]
```

where I assumed that:
1. We represent each basis as a class.
2. Each basis class is accessed via an operator using the ".basis" property.
3. The class supports indexing where the index is the corresponding eigenvalue.

The code is not just shorter, but is more amenable to programmatic generation of states, e.g.:

```python
def func(operators, eigenvalue):
    state = 1
    for op in operators:
        state *= op.basis[eigenvalue]
    return state

s = func([cirq.X, cirq.Y, cirq.Z], -1)
```

Also, the choice to use existing gates such as cirq.X provides a familiar hook for this functionality: once the user learns the ".basis" trick they know how to easily build complicated states without having to remember names we invent for the various states (which as the set of states grows are bound to become increasingly weird and idiosyncratic).

Side note: Interestingly, it seems that `EigenGate._eigen_components ` could be used readily to implement the `projector` member function that you need.

mpharrigan: what is the type of `x.basis[i]`? This looks nifty but looks like an alternative way of constructing OneQubitStates and does not obviate the need for such a class

viathor: Ah, I was imprecise. I meant a situation where we'd have one class per interesting basis, but each instance of a basis class would represent a basis element (just like your instances do presently). In particular, it would be appropriate to name these classes as `*State` as you do currently (`__init__` arguments would distinguish basis elements). For example, in place of two classes such as `_KetPlus` and `_KetMinus`, we'd have a single class, say `_XEigenState` (and there would be two instances of this class). This reduces verbosity by a factor of two for qubits (and more for qudits).

cirq/value/tensor_product_state.py
```diff
@@ -0,0 +1,346 @@
...
+
+@dataclass(frozen=True)
+class TensorProductState:
+    """A quantum state that is a tensor product of one qubit states.
+
+    For example, the |00⟩ state is `cirq.KET_ZERO(q0) * cirq.KET_ZERO(q1)`.
+    The |+⟩ state is a length-1 tensor product state and can be constructed
+    with `cirq.KET_PLUS(q0)`.
+    """
+    states: Dict['cirq.Qid', _NamedOneQubitState]
```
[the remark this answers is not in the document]
mpharrigan: `STATE_VECTOR_LIKE` is a highly-permissive type annotation that should be kept to our API surface and everything should be canonicalized inside containers, like this. `STATE_VECTOR_LIKE` also is generally on > 1 qubit, which wouldn't work here. 

I'd also push back pretty strongly against allowing mixed states here. `STATE_VECTOR_LIKE` and friends use the convention that "state vector" deals with pure states.

cirq/value/tensor_product_state.py
mpharrigan: Yes, I think this is a good happy medium among all the various levels of verbosity that have been proposed here! I'll work on this change

cirq/value/tensor_product_state.py
```diff
@@ -0,0 +1,346 @@
...
+        return self.__class__ == other.__class__
+
+    def __hash__(self):
+        # Descendants should be singletons determined solely by the class name.
+        # Otherwise, you must override this method.
+        return hash(self.__class__.__name__)
+
+
+@dataclass(frozen=True)
+class TensorProductState:
```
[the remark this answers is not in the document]
mpharrigan: done

mpharrigan:
alright, check out the new form. Now you can do either `cirq.KET_ZERO` or `cirq.Z.basis[+1]` and get a `_ZEigenState(eigenvalue=+1)` object, and most of the logic is factored into a base class with only the specification of the actual ndarray's left to the base classes.

cirq/value/product_state.py
```diff
@@ -0,0 +1,319 @@
...
+            return np.array([0, 1])
+        # coverage: ignore
+        raise ValueError("Bad eigenvalue: {}".format(self.eigenvalue))
+
+    def stabilized_by(self):
+        from cirq import ops
+        return self.eigenvalue, ops.Z
+
+
+KET_PLUS = _XEigenState(eigenvalue=+1)
```
viathor, line 244: These symbols lead to two different ways of accessing the 1-qubit stabilizer states: by names (when available) and by the ".basis" trick. This goes against the [Zen of python](https://www.python.org/dev/peps/pep-0020/). If adherence to the Zen isn't a sufficient argument against them, then consider that they will also lead to expressions that mix the two ways - these will look ugly - and may delay the users becoming familiar with the ".basis" trick - which we want them to learn as soon as possible.

Suggestion: Remove all KET_* symbols (but keep `PAULI_STATES` - this is useful, e.g. for iterating over all 1-qubit stabilizer states).

cirq/ops/pauli_gates.py
```diff
@@ -119,6 +117,14 @@ def _from_json_dict_(cls, exponent, global_shift, **kwargs):
         assert exponent == 1
         return Pauli._XYZ[0]
 
+    @property
+    def basis(self):
```
viathor, line 121: Missing type annotation (applies to other basis methods, too.)

viathor, line 121: WDYT about moving this functionality into the `EigenGate` (where I think you may be able to benefit from the gate knowing the projectors onto their eigenspaces)?

Also, I imagine it may be useful sometimes to have a way to get *a* non-stabilizer state and something like `cirq.H.basis[+1]` is an obvious thing to try.

cirq/value/product_state.py
viathor: (Relevant excerpt from the Zen: "There should be one-- and preferably only one --obvious way to do it.")

mpharrigan: I see the current construct analogous to how we have both `cirq.CNotPowGate(exponent=1)` and `cirq.CNOT`. I also think there's educational value in presenting e.g. |+> as either the 0'th basis element of the X pauli operator or "the plus state"

cirq/ops/pauli_gates.py
mpharrigan: You can get the projectors, yes but not the state vectors; and that's the important part. In particular, the state vector ndarrays are in the `_[_]EigenState` classes which are only implemented for `X`, `Y`, `Z`. Ie, we'd have to put another one in for `H`.

cirq/value/tensor_product_state.py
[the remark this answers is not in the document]
viathor: SGTM

cirq/value/tensor_product_state.py
```diff
@@ -0,0 +1,346 @@
...
+        return {
+            'cirq_type': self.__class__.__name__,
+            'states': list(self.states.items()),
+        }
+
+    @classmethod
+    def _from_json_dict_(cls, states, **kwargs):
+        return cls(states=dict(states))
+
+    def state_vector(self, qubit_order: 'cirq.QubitOrder' = None):
```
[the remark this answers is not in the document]
viathor: Not yet addressed.

cirq/value/product_state.py
```diff
@@ -0,0 +1,319 @@
...
+        for qubit in qubits:
+            oneq_state = self[qubit]
+            oneq_proj = oneq_state.projector()
+            mat = np.kron(mat, oneq_proj)
+        return mat
+
+
+class _PauliEigenState(_NamedOneQubitState):
+
+    def __init__(self, eigenvalue):
```
viathor, line 161: This should also take op (as returned by `stabilized_by`) and state vector (as returned by `state_vector`). This way you'll remove even more duplication from the `_*EigenState` classes below (i.e. you can probably move definitions of `state_vector` and `stabilized_by` into `_PauliEigenState`). This will make `_*EigenState` classes trivial which is good since conceptually there isn't much difference between the bases to justify per-basis logic.

cirq/value/tensor_product_state.py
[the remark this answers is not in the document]
mpharrigan: Hey @viathor . Ack! I was on vacation last week :)

cirq/ops/pauli_gates.py
mpharrigan: This is such a good idea I already thought of it and the test is here! https://github.com/quantumlib/Cirq/pull/3171/files#diff-c5d792e951ba985282f75339939b6e14R60

(And indeed the projector is computed from the outer product of the state vector)

viathor: Cool! Thanks! I expect over time we'll add more eigenbases and we'll want to test each for consistency with the projectors, so we should move this into `assert_eigengate_implements_consistent_protocols`.

mpharrigan: I mean I don't consider this a "protocol" at this point. After a quick look at the plumbing for `assert_eigengate_implements_consistent_protocols` it's not clear how to add this check other than shoe-horning it in for X, Y, Z gates. I find it much clearer to have this of the projector method alongside the projector method

cirq/__init__.py
```diff
@@ -410,6 +410,14 @@
     TParamVal,
     validate_probability,
     value_equality,
+    KET_PLUS,
+    KET_MINUS,
+    KET_IMAG,
+    KET_MINUS_IMAG,
+    KET_ZERO,
+    KET_ONE,
```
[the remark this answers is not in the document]
mpharrigan: I'll once again appeal to the analogy to gates where you could argue that more structure could be carried in namespacing but in practice it's quite nice to have everything at the top level and have useful aliases of common gates.

cirq/ops/pauli_gates.py
viathor: Fair enough. I think a protocol would be appropriate here (especially since most gates do not support `.basis`), but it's reasonable to say it's out-of-scope for this PR.

BTW: Should I have another look or do you plan some more changes here?

cirq/value/product_state.py
mpharrigan: This was one of the original suggestions. I don't like it because (1) you'd run into circular import errors with `ops` and (2) gates' unitary method specifies their matrix form; it's not provided as a constructor. This gives the gate class some meaning. There would actually be *no* `_*EigenState` classes if you take this to its logical end, and now you've lost the ability to quickly check what basis you're dealing with (e.g. to add basis-change rotations prior to measurement). To abuse the metaphor: It's the same reason why we have gate classes and not just `MatrixGate` and a bunch of constants

cirq/ops/pauli_gates.py
mpharrigan: added!

mpharrigan: Just added type annotations. Have a look!

Since most gates don't support basis doesn't that mean it's *not* appropriate for a protocol? I know there's no widely accepted definition for what makes a good Cirq protocol but usually it's "an abstract api with many fallback implementations" and I don't see that here.

The real sticking point is that you have to write down the state vector representations "by hand" somewhere

mpharrigan:
@viathor @balopat have a look

balopat (approved):
This looks good to me. Thanks for your patience on this.

cirq/ops/pauli_gates.py
viathor: Most gates don't provide their eigenbasis, but they do provide enough other information that it isn't hard to imagine various fallbacks that derive the eigenbasis *for certain types of gates*. Hence, a protocol. It might look something like this:

```
if the gate provides its eigenbasis explicitly then return that (e.g. X)
else if the gate is a tensor product of gates then return tensor product of their eigenbases (e.g. XX)
else if the gate is an EigenGate then use its projector (e.g. H)
else compute eigendecomposition or fail or whatever
```

Sure, we need to write down the eigenbasis somewhere for *some* gates, but not all. For other gates we can compute it from existing information (e.g. the projector for EigenGates) and we can compute it recursively (e.g. XX).

In any case, my point wasn't really about implementing a new protocol, but about testing consistency of eigenbases with projectors which you're doing in the test. That's good enough for now.

cirq/value/product_state.py
viathor: I'm not convinced by the metaphor. Many gates have efficient implementations of `apply_unitary` which exploit the particular form of their unitary matrix. I don't see anything like this here.

Here is my concern: this PR doesn't just add a fairly new thing to cirq but it also establishes a *new practice*: "how to define new states in cirq". I would like to see us choose an efficient approach that requires relatively little new code and little repetition/verbosity. As it is, the code isn't too bad, but now imagine we add ~5 more classes like this. I think our future selves would prefer we did something like this:

```python
class _ZEigenState(_PauliEigenState):
    def __init__(self):
        super().__init__(self,
                         eigendict={+1: np.array([1, 0]), -1: np.array([1, 0])},
                         op=ops.Z,
                         symbol='Z')
```

cirq/value/product_state.py
```diff
@@ -0,0 +1,319 @@
...
+
+        mat = 1.0 + 0.j
+        for qubit in qubits:
+            oneq_state = self[qubit]
+            oneq_proj = oneq_state.projector()
+            mat = np.kron(mat, oneq_proj)
+        return mat
+
+
+class _PauliEigenState(_NamedOneQubitState):
```
viathor, line 159: For future extensibility: does this need to have "Pauli" in the name? AFAICT, the implementation is appropriate more generally (modulo a single type annotation below) .

viathor (commented):
Almost there, two last comments.

cirq/__init__.py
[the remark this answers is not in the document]
viathor: OK

cirq/value/product_state.py
balopat: There is dependency in str() and init() on having only +1/-1 eigenvalues available for this state. This is specific to involutory matrices - but in this context I think it's fine to call it PauliEigenState - if we need a more generic EigenState, we can extract it. I would like to push this in as I think it is in a good enough shape to be iterated on if needed. @viathor do you have any big blockers still outstanding?

viathor: Good point about nature of the eigenvalues. No other blockers.

viathor (approved):

PR #3173, opened 2020-07-23
Add option to preserve tags for TaggedOperations when resolving parameters

sabrinahong:


maffoo:
Ha, I had the same idea: https://github.com/quantumlib/Cirq/pull/3172 :-)

In that case I instead opted to make it non-optional since I can think of a reason to _not_ preserve tags, and I also went ahead and allowed for tags to themselves be parameterized.

sabrinahong:
Oh haha totally missed your PR @maffoo , thanks for addressing it! I'll go ahead and close this one.

PR #3174, opened 2020-07-23
fixing broken deprecation namings

### LINKED ISSUE #3170: `final_wavefunction` is deprecated wrong

balopat:
Fixes #3170. 

The deprecated decorator should not be used in a way it was directly used. 
The pattern was: 

```
old_function = deprecated(
    deadline='v0.9', fix='Use cirq.new_func instead.')(
        cirq.new_func)
```

The main problem is that we pass in the _new_ function into the deprecated decorator...now it does return a new function instance that is a decorated one and hence the old function gets a good wrapper around it. However, the name calculation in case `name` arg is not defined cannot determine what the name of `old_function` is. 

There are two ways to fix this AFAICT (unless I'm missing some python magic that can tell what the name of the variable is that is calling the current function - which I found to be not possible): 
1.) set `name` and duplicate the variable (function) name 
```
old_function = deprecated(
    deadline='v0.9', fix='Use cirq.new_func instead.', name = 'old_function')(
        cirq.new_func)
```
2.) use the decorator on a proper def
```
@deprecated(
    deadline='v0.9', fix='Use cirq.new_func instead.', name = 'old_function')
def old_function(*args,**kwargs):
     return cirq.new_func(*args,**kwargs)
```

I chose 2 in my solution, a bit more verbose but there is no duplication at least.

### DIFF
--- a/cirq/_compat_test.py
+++ b/cirq/_compat_test.py
@@ -76,11 +76,27 @@ def f(a, b):
     return a + b
 
 
-def test_deprecated():
+def test_deprecated_with_name():
     with cirq.testing.assert_logs('test_func was used',
                                   'will be removed in cirq vNever',
                                   'Roll some dice.'):
         assert f(1, 2) == 3
+
+
+def new_func(a, b):
+    return a + b
+
+
+@deprecated(deadline='vNever', fix='Roll some dice.')
+def resolved_name(*args, **kwargs):
+    return new_func(*args, **kwargs)
+
+
+def test_deprecated_with_name_resolution():
+    with cirq.testing.assert_logs('resolved_name was used',
+                                  'will be removed in cirq vNever',
+                                  'Roll some dice.'):
+        assert resolved_name(1, 2) == 3
 
 
 def test_deprecated_parameter():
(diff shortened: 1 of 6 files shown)


cirq/_compat_test.py
```diff
@@ -76,13 +76,29 @@ def f(a, b):
...
                                   'Roll some dice.'):
         assert f(1, 2) == 3
 
 
+def new_func(a, b):
+    return a + b
+
+
+@deprecated(deadline='vNever', fix='Roll some dice.')
+def resolved_name(*args, **kwargs):
```
maffoo, line 91: nit: maybe call this `old_func` to make the intent more clear

cirq/_compat_test.py
```diff
@@ -76,13 +76,29 @@ def f(a, b):
...
+    return new_func(*args, **kwargs)
+
+
+def test_deprecated_with_name_resolution():
```
maffoo, line 95: It's not clear to me what "name resolution" means here. This seems to me like it should be the default case so could perhaps just call this `test_deprecated`?

balopat:
@googlebot please rescan

Issue #3175, opened 2020-07-24
cirq-unstable updates are broken


maffoo:
cirq-unstable has failed to update since #3167. The packages are failing to upload to pypi, and from the [build log](https://travis-ci.com/github/quantumlib/Cirq/builds/176794690) it looks like pypi can't handle something in the new README.rst:

> HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/
> The description failed to render in the default format of reStructuredText. See https://pypi.org/help/#description-content-type for more information.

maffoo:
Fixed by #3176.

PR #3176, opened 2020-07-24
Fix readme section headers

maffoo:
I'm not sure if this will fix the cirq-unstable upload (#3175), but it does say in the [reStructuredTest spec](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#sections) that
> An underline/overline is a single repeated punctuation character that begins in column 1 and forms a line extending at least as far as the right edge of the title text.

I also added an extra blank line before section headers which improves readability of the plain text file, at least to my eyes.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,6 +17,7 @@ circuits and running them against quantum computers and simulators.
     :target: https://readthedocs.org/projects/cirq/versions/
     :alt: Documentation Status
 
+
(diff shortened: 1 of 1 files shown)


maffoo:
Looks like the pypi upload [succeeded](https://travis-ci.com/github/quantumlib/Cirq/builds/177057561), so I'll close #3175.

PR #3177, opened 2020-07-24
adding rstcheck for rst files

balopat:
Adds rstcheck to the doc tests CI step.

This is to avoid accidentally breaking pypi publishing with syntactically incorrect. See https://github.com/quantumlib/Cirq/issues/3175 for when this happened.

### DIFF
--- a/dev_tools/conf/pip-list-dev-tools.txt
+++ b/dev_tools/conf/pip-list-dev-tools.txt
@@ -30,4 +30,4 @@ ipython
 ipykernel
 
 # For verifying rst
-rstcheck
+rstcheck~=3.3.1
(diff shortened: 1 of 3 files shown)


dev_tools/conf/pip-list-dev-tools.txt
```diff
@@ -23,7 +23,11 @@ recommonmark >= 0.4.0
...
 sphinx_rtd_theme
 sphinx-markdown-tables
+
 # For generating notebooks in documentation
 nbsphinx
 ipython
 ipykernel
+
+# For verifying rst
+rstcheck
```
maffoo, line 33: Consider pinning the version here, as we do for other packages that check for correctness like mypy, pylint, yapf, etc. It can be frustrating when a local build and ci disagree because they're using different package versions.

balopat: that makes total sense, thank you!

PR #3181, opened 2020-07-29
[Pasqal updates] Adding PasqalVirtualDevice

HGSilveri:
This is the fifth child-PR of #3126, following  #3160.

`PasqalVirtualDevice`, which enforces the constraints typically found in a physical device, is added to the module. These constraints were present in `PasqalDevice` before its generalisation in #3141 . They are now reincorporated here.

This is the last code PR of the Pasqal updates, after which all that is left is to incorporate the documentation.


HGSilveri:
Awesome, @balopat ! That's the end of the code PR's, only the docs are left now. Thank you very much for your time and help!

Issue #2492, opened 2019-11-05
Implement cirq.eigencomponents()


jaeyoo:
Conclusion of discussion in #2481 .

tonybruguier-google:
Sent:
https://github.com/quantumlib/Cirq/pull/3197

Happy to withdraw if misguided.

balopat:
To clarify, an implementation of `cirq.eigen_projectors/vectors/values` would be something that is fast for `EigenGate`s (using its eigencomponents method) but it falls back to numpy's linear algebra methods in the general case.

MichaelBroughton:
This looks like it's in now. Going to close.

PR #3195, opened 2020-08-06
Approximately equal circuits with symbols not detected by cirq.approx_eq

### LINKED ISSUE #3192: Approximately equal circuits with symbols not detected by cirq.approx_eq
**Description of the issue**
Circuits with symbols are determined to be unequal by `cirq.approx_eq` when they should be approximately equal.

```python
q = cirq.GridQubit(0, 0)
s = sympy.Symbol("s")

no_symbol_1 = cirq.Circuit(cirq.rz(1.515)(q))
no_symbol_2 = cirq.Circuit(cirq.rz(1.510)(q))
print(cirq.approx_eq(no_symbol_1, no_symbol_2, atol=0.1))

symbol_1 = cirq.Circuit(cirq.rz(1.515*s)(q))
symbol_2 = cirq.Circuit(cirq.rz(1.510*s)(q))
print(cirq.approx_eq(symbol_1, symbol_2, atol=0.1))
```
First output is `True`, second output is `False`

**Cirq version**
0.8.2

tonybruguier-google:
Approximately equal circuits with symbols not detected by cirq.approx_eq (Fixes #3192.)

We should use sympy to make comparisons when available.


cirq/protocols/approximate_equality_protocol.py
```diff
@@ -94,6 +95,11 @@ def approx_eq(val: Any, other: Any, *, atol: Union[int, float] = 1e-8) -> bool:
     if isinstance(val, str):
         return val == other
 
+    if isinstance(val, sympy.core.mul.Mul) or isinstance(
+            other, sympy.core.mul.Mul):
+        delta = sympy.simplify(sympy.Abs(other - val))
```
balopat, line 100: Why do we need the simplify here?

cirq/protocols/approximate_equality_protocol_test.py
```diff
@@ -152,6 +153,23 @@ def test_approx_eq_list():
     assert not cirq.approx_eq([1.1, 1.2, 1.3], [1, 1, 1], atol=0.2)
 
 
+def test_approx_eq_symbol():
+    q = cirq.GridQubit(0, 0)
+    s = sympy.Symbol("s")
+    t = sympy.Symbol("t")
+
+    no_symbol_1 = cirq.Circuit(cirq.rz(1.515)(q))
```
balopat, line 161: these should be covered elsewhere, no need to repeat in this test

tonybruguier-google: Done

cirq/protocols/approximate_equality_protocol.py
tonybruguier-google: If I don't do that, the unit test fails. Upon inspection, if I print `delta`, I see `delta=Abs((s + 1.51)/pi - (s + 1.515)/pi)`

tonybruguier-google:
Thanks. PTAL.

cirq/protocols/approximate_equality_protocol.py
```diff
@@ -94,6 +95,11 @@ def approx_eq(val: Any, other: Any, *, atol: Union[int, float] = 1e-8) -> bool:
     if isinstance(val, str):
         return val == other
 
+    if isinstance(val, sympy.core.mul.Mul) or isinstance(
```
balopat, line 98: use sympy.Basic instead Mul is way too restrictive

We should also add a test case maybe directly against other expressions e.g.
```
cirq.approx_eq(t+1.51+s, t+1.50+s, atol=0.1)
```
Would evaluate to False with this code.

cirq/protocols/approximate_equality_protocol.py
balopat: Got it - however, this is still not kosher. There might be a case where you can't evaluate to a number literal and then this question is undecidable. We should handle that and throw an error instead of silently returning False.

```sympy.Abs(val - other).simplify().is_number``` should give you a good indication if the simplification worked.

Please add a test case covering that too.

File: cirq/protocols/approximate_equality_protocol.py [hunk 4]
```diff
@@ -94,6 +95,11 @@ def approx_eq(val: Any, other: Any, *, atol: Union[int, float] = 1e-8) -> bool:
...
+            other, sympy.core.mul.Mul):
+        delta = sympy.simplify(sympy.Abs(other - val))
+        return sympy.LessThan(delta, atol) == True
```
balopat, line 101: Let's compare to `sympy.true` to make this a bit less confusing - it does work but it does take a bit of a sympy knowledge to understand what's happening here.

cirq/protocols/approximate_equality_protocol.py
tonybruguier-google: Changed to Basic

Added the direct test case, although since the tolerance was right on the threshold, I decided to add two tests, once with tolerate 2.0, one with tolerance 0.5

cirq/protocols/approximate_equality_protocol.py
tonybruguier-google: So I know very little abot Sympy. Did you mean `sympy.LessThan(delta, atol) == sympy.true`? If so, I see an error:
`
E       assert not True
E        +  where True = <function approx_eq at 0x7f9a3f624ee0>(((t + 1.51) + s), ((t + 1.5) + s), atol=0.05)
E        +    where <function approx_eq at 0x7f9a3f624ee0> = cirq.approx_eq
`

I suspect it's because we need to convert to a plain Python True. Although I am not sure I understood you correctly and since I did not change the code, I'm not resolving this comment.

cirq/protocols/approximate_equality_protocol.py
balopat: Why was the tolerance on the threshold? 1.51-1.50 = 0.01 < 0.1

tonybruguier-google: It was a mistake. Changed to 0.2 for the other tests.

tonybruguier-google:
Fixes #2492

cirq/protocols/approximate_equality_protocol.py
```diff
@@ -94,6 +95,12 @@ def approx_eq(val: Any, other: Any, *, atol: Union[int, float] = 1e-8) -> bool:
     if isinstance(val, str):
         return val == other
 
+    if isinstance(val, sympy.Basic) or isinstance(other, sympy.Basic):
+        delta = sympy.Abs(other - val).simplify()
+        if not delta.is_number:
+            return False
```
balopat, line 101: We should throw an error, rather than False - they might be equal but we don't know!

tonybruguier-google: Done

cirq/ops/eigen_gate_test.py
```diff
@@ -132,22 +134,22 @@ def test_approx_eq():
...
         CExpZinGate(sympy.Symbol('a')),
         CExpZinGate(sympy.Symbol('a')),
         atol=0.1
     )
-    assert not cirq.approx_eq(
-        CExpZinGate(sympy.Symbol('a')),
-        CExpZinGate(sympy.Symbol('b')),
-        atol=0.1
-    )
+    with pytest.raises(AttributeError,
```
balopat, line 148: it's awesome that we caught this! This was a bad test - as how can we tell if a == b? we can't!

cirq/protocols/approximate_equality_protocol.py
```diff
@@ -94,6 +95,12 @@ def approx_eq(val: Any, other: Any, *, atol: Union[int, float] = 1e-8) -> bool:
     if isinstance(val, str):
         return val == other
 
+    if isinstance(val, sympy.Basic) or isinstance(other, sympy.Basic):
+        delta = sympy.Abs(other - val).simplify()
+        if not delta.is_number:
+            raise AttributeError('Cannot reduce Sympy expression to number.')
```
balopat, line 101: one last nit: I would make the error message more descriptive to include contextual information for the user: 
`f"Insufficient information to decide whether expressions are approximately equal [{val}] vs [{other}]"`

cirq/ops/eigen_gate_test.py
tonybruguier-google: Ack. Thanks!

cirq/protocols/approximate_equality_protocol.py
tonybruguier-google: Done.

Issue #2763, opened 2020-02-12
Distance measures for quantum states


mpharrigan:
- fidelity
- process fidelity
- purity
- diamond norm
- others?

viathor:
Well, trace distance, of course.

Regarding other distance measures to consider, it is surprisingly easy for them to proliferate. Given a distance measure on quantum states you can define a distance measure on quantum operations by:
1. Applying the former to the (trace-normalized) process matrices.
2. Computing the average over input states.
3. Computing the maximum over input states.

The latter has some cool properties, but is not invariant under tensoring with the identity. The problem can be remedied by a small tweak to the definition (first tensor with the identity on a sufficiently large Hilbert space, then compute maximum).

Now, the diamond norm is the distance measure for quantum processes obtained by applying maximization over input states (with the above tweak) to the trace distance. You can apply the same procedure to other distance measures. At least one of them is actually interesting and that is when you apply maximization over input states (again, with the above tweak) to the fidelity. I don't think this has an established name. I certainly haven't seen anyone refer to this as "diamond fidelity". I have seen it called "stabilized fidelity" or "S fidelity".

Also, since you've included purity, which is not a distance measure, we may consider normalized linear entropy, tangle, Schmidt rank, von Neumann entropy and maybe other measures of entanglement.

So, there is potentially a lot of stuff to do here. However, there are reasons for parsimony. If anything the field suffers from too many rather than too few distance measures and people came up with well-justified recommendations to settle on. We should probably endorse one (or few) of those and stick to them. Moreover, some of distance measures one can find in the literature (e.g. the Bures metric) are simple functions of the fidelity anyway so users who really need them can very easily compute them from the fidelity (trace distance is an exception). 

Also, regarding distance measures for operations, not every combination of procedures 1-3 above and a distance measure for quantum states yields a reasonable distance measure for quantum operations. Process fidelity and diamond norm SGTM. Process trace distance and the S fidelity may be reasonable additions. And we should probably stop there.

Regarding measures of entanglement, many of them are NP-hard to compute.

viathor:
There is another, related concern that came up in a discussion I had with @maffoo. We probably want to avoid developing and maintaining a full quantum information package inside cirq. Many such packages already exist, e.g. [qiskit](https://qiskit.org/) and [qutip](http://qutip.org/) and have a lot more functionality than we are likely to build. Rather than writing our own, we may want to consider partnering with an existing project and ensuring easy interoperability.

@dabacon @Strilanc

[...]

PR #3198, opened 2020-08-08
Add a process fidelity utility function

tonybruguier-google:
This is for issue #2763

### DIFF
--- a/cirq/__init__.py
+++ b/cirq/__init__.py
@@ -316,6 +316,7 @@ from cirq.qis import (
     eye_tensor,
     fidelity,
     one_hot,
+    process_fidelity,
     STATE_VECTOR_LIKE,
     to_valid_density_matrix,
     to_valid_state_vector,
--- a/cirq/qis/__init__.py
+++ b/cirq/qis/__init__.py
@@ -1,5 +1,6 @@
 from cirq.qis.measures import (
     fidelity,
+    process_fidelity,
     von_neumann_entropy,
 )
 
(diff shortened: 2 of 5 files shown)


mpharrigan:
What are U and V? time-evolution operators that are nearly unitary? I skimmed the linked preprint and I don't think this is exactly what we'd want as the main `process_fidelity` function. It should be defined on two CPTP maps aka quantum channels aka something that turns one density matrix into another. 

The function from qiskit is probably a good starting point: https://qiskit.org/documentation/stubs/qiskit.quantum_info.process_fidelity.html#qiskit.quantum_info.process_fidelity. Of course, this would require introducing a superoperator representation for processes. The different representations are described here, for example https://arxiv.org/abs/1111.6950

viathor:
I agree with @mpharrigan. Any utility for process fidelity needs to be explicit about what representation of a CPTP map it accepts. Also, a few [common maps](https://github.com/quantumlib/Cirq/blob/master/cirq/ops/common_channels.py) are already implemented in cirq. We should use one representation consistently throughout the library and if you find that the present representation is insufficient we should generalize it appropriately rather than adding another.

tonybruguier:
Hello,

Sorry for the delay. I have now a new user name and it took me a while to understand the paper you gave me (sorry and thank you for the help). I would suggest closing this PR and moving the review to:
https://github.com/quantumlib/Cirq/pull/3226

balopat:
> I would suggest closing this PR and moving the review to:
> #3226

Sounds good - closing this one.

PR #3203, opened 2020-08-10
Support ActOnStabilizerCHFormArgs in the common gates and testing

smitsanghavi:
Also some minor bug fixes and refactoring in StabilizerStateCHForm, ActOnCliffordTableauArgs and consistent_act_on.

#2948 #2423


cirq/ops/common_gates.py
```diff
@@ -321,6 +333,43 @@ def _act_on_(self, args: Any):
...
+                if ZPowGate()._act_on_(args) is True:
+                    if H._act_on_(args) is True:
+                        state.omega *= (1 + 1j) / (2**0.5)  # type: ignore
+                        return True
+                    # Only one of the delegated _act_on_ succeeded, so the
+                    # state might have been corrupted, return False instead of
+                    # NotImplemented.
+                    return False  # coverage: ignore
+            elif effective_exponent == 1:
+                if ZPowGate()._act_on_(args) is True:
```
balopat, line 351: instead of these embedded `if` statements, I would do something like this:
```
for op in [ZPowGate(), H, ZPowGate(), H]:
    if not op._act_on(args): 
        # fail fast in case there is a failure
        return False 
state.omega *= 1j 
return True 
```

cirq/sim/clifford/act_on_stabilizer_ch_form_args.py
```diff
@@ -0,0 +1,95 @@
...
+
+class ActOnStabilizerCHFormArgs:
+    """State and context for an operation acting on a stabilizer state in CH
+    form. To act on this object, directly edit the `state` property, which is
+    storing the stabilizer state of the quantum system with one axis per qubit.
+    Measurements are currently not supported on this object.
+    """
+
+    def __init__(self, state: StabilizerStateChForm, axes: Iterable[int]):
+        """
```
balopat, line 39: Please fix the docs string according to our [style guide](https://github.com/quantumlib/Cirq/blob/master/docs/dev/development.md#writing-docstrings-and-generating-documentation).

Namely: no single line intro or description here.

cirq/sim/clifford/act_on_stabilizer_ch_form_args.py
```diff
@@ -0,0 +1,95 @@
...
+from cirq.ops.clifford_gate import SingleQubitCliffordGate
+from cirq.protocols import has_unitary, num_qubits, unitary
+from cirq.sim.clifford.stabilizer_state_ch_form import StabilizerStateChForm
+
+if TYPE_CHECKING:
+    import cirq
+
+
+class ActOnStabilizerCHFormArgs:
+    """State and context for an operation acting on a stabilizer state in CH
```
balopat, line 32: Please fix the docs string according to our [style guide](https://github.com/quantumlib/Cirq/blob/master/docs/dev/development.md#writing-docstrings-and-generating-documentation).

Namely: no single line intro or description here (I know that it's not consistent with ActOnCliffordTableau but that should be fixed too).

cirq/testing/consistent_act_on.py
```diff
@@ -41,19 +44,32 @@ def state_vector_has_stabilizer(state_vector: np.ndarray,
...
 
-def assert_act_on_clifford_tableau_effect_matches_unitary(val: Any) -> None:
-    """Checks that act_on with CliffordTableau generates stabilizers that
-    stabilize the final state vector. Does not work with Operations or Gates
-    expecting non-qubit Qids."""
+def assert_all_implemented_act_on_effects_match_unitary(
+        val: Any,
+        assert_tableau_implemented: bool = False,
+        assert_ch_form_implemented: bool = False) -> None:
+    """
```
balopat, line 51: Please fix the docs string according to our [style guide](https://github.com/quantumlib/Cirq/blob/master/docs/dev/development.md#writing-docstrings-and-generating-documentation).

Namely: no single line intro

cirq/testing/consistent_act_on.py
```diff
@@ -104,3 +134,24 @@ def _final_clifford_tableau(circuit: Circuit, qubit_map
         except TypeError:
             return None
     return tableau
+
+
+def _final_stabilizer_state_ch_form(
+        circuit: Circuit,
+        qubit_map) -> Optional[stabilizer_state_ch_form.StabilizerStateChForm]:
+    """Initializes a StabilizerStateChForm with default args for the given
```
balopat, line 142: Please fix the docs string according to our [style guide](https://github.com/quantumlib/Cirq/blob/master/docs/dev/development.md#writing-docstrings-and-generating-documentation).

Namely: no single line intro and or description of args and return values.

cirq/sim/clifford/act_on_clifford_tableau_args.py
```diff
@@ -86,8 +86,7 @@ def _act_on_fallback_(self, action: Any, allow_decompose: bool):
 
 def _strat_act_on_clifford_tableau_from_single_qubit_decompose(
         val: Any, args: 'cirq.ActOnCliffordTableauArgs') -> bool:
-    num_qubits_getter = getattr(val, 'num_qubits', None)
-    if num_qubits_getter is not None and num_qubits_getter() == 1:
+    if num_qubits(val) == 1:
```
balopat, line 89: This change should be technically split out to a separate PR - do you mind doing that?

cirq/sim/clifford/act_on_stabilizer_ch_form_args.py
```diff
@@ -0,0 +1,95 @@
...
+        if not has_unitary(val):
+            return NotImplemented
+        u = unitary(val)
+        clifford_gate = SingleQubitCliffordGate.from_unitary(u)
+        if clifford_gate is not None:
+            for axis, quarter_turns in clifford_gate.decompose_rotation():
+                if axis == pauli_gates.X:
+                    if common_gates.XPowGate(exponent=quarter_turns /
+                                             2)._act_on_(args) is not True:
+                        return False  # coverage: ignore
```
balopat, line 77: I'm really worried about these coverage: ignore-s all around - this should be only the case if it is impossible to cover.

cirq/sim/clifford/act_on_clifford_tableau_args.py
smitsanghavi: Makes sense. I'll split out the small changes that don't really fit here and send separate PRs for them. Will hold off on this one for now since the small ones would need to go in first.

cirq/ops/common_gates.py
balopat: We can make it even more pythonic:

```
if not all(cirq.act_on(op, args) for op in [ZPowGate(), H, ZPowGate(), H])
```

balopat: Ah ignore that last one - `cirq.act_on` does not return the bool value.

I also don't understand why you would not return False in case the first act_on fails?

cirq/ops/common_gates.py
```diff
@@ -103,6 +103,18 @@ def _act_on_(self, args: Any):
                 tableau.xs[:, q] ^= tableau.zs[:, q]
             return True
 
+        if isinstance(args, clifford.ActOnStabilizerCHFormArgs):
+            if protocols.is_parameterized(self) or self.exponent % 0.5 != 0:
+                return NotImplemented
+            if H._act_on_(args) is True:
```
Strilanc, line 109: I would replace these with explicit calls to the methods of ChForm, or just assert that they succeeded.

cirq/ops/common_gates.py
```diff
@@ -322,6 +334,43 @@ def _act_on_(self, args: Any):
...
+            state = args.state
+            if effective_exponent == 0.5:
+                if ZPowGate()._act_on_(args) is True:
+                    if H._act_on_(args) is True:
+                        state.omega *= (1 + 1j) / (2**0.5)  # type: ignore
+                        return True
+                    # Only one of the delegated _act_on_ succeeded, so the
+                    # state might have been corrupted, return False instead of
+                    # NotImplemented.
+                    return False  # coverage: ignore
```
Strilanc, line 350: Make a test that triggers it, or assert it doesn't happen.

cirq/ops/common_gates.py
```diff
@@ -322,6 +334,43 @@ def _act_on_(self, args: Any):
...
+            elif effective_exponent == 1.5:
+                if H._act_on_(args) is True:
+                    if ZPowGate()._act_on_(args) is True:
+                        state.omega *= (1 - 1j) / (2**0.5)  # type: ignore
+                        return True
+                    # Only one of the delegated _act_on_ succeeded, so the
+                    # state might have been corrupted, return False instead of
+                    # NotImplemented.
+                    return False  # coverage: ignore
```
Strilanc, line 370: Does the global protocol method recognize these False results? Won't this cause a misleading error about the gate class returning an invalid result, instead of an error about not supporting the operation?

cirq/sim/clifford/act_on_stabilizer_ch_form_args_test.py
```diff
@@ -0,0 +1,69 @@
+# Copyright 2018 The Cirq Developers
```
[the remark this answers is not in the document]
smitsanghavi: Done

cirq/ops/common_gates.py
smitsanghavi: I was creating a tri-state where `False` denotes a corruption of state or an error and results in https://github.com/quantumlib/Cirq/blob/master/cirq/protocols/act_on_protocol.py#L102

NotImplemented allows the protocol to try the fallback strategy.

If the first `act_on` fails, the state has not been modified and we are free to try the fallback. If the first `act_on` succeeds but some subsequent one fails, the state might be unintentionally modified and we should fail fast.

But based on comments by @Strilanc and you, this tri-state behavior might be too messy while adding minimal value. 

I decided to do away with it and `assert` in the pythonic way you suggest here. This will cause us to fail when any of the listed gates fails to `act_on`(including the first one). But given that these are well-established Clifford gates that should have `_act_on_` implemented, even if the first one fails, it is actually more prudent to assert fail right here rather than keeping an option of a fallback strategy.

cirq/sim/clifford/act_on_stabilizer_ch_form_args.py
smitsanghavi: Done

cirq/sim/clifford/act_on_stabilizer_ch_form_args.py
smitsanghavi: Done

cirq/testing/consistent_act_on.py
smitsanghavi: Done

cirq/testing/consistent_act_on.py
smitsanghavi: Done

cirq/sim/clifford/act_on_stabilizer_ch_form_args.py
smitsanghavi: Removed them all

cirq/ops/common_gates.py
smitsanghavi: The methods of the ChForm are to be deleted as per #2948.

cirq/ops/common_gates.py
smitsanghavi: Assert-ed!

I was creating a tri-state where `False` denotes a corruption of state or an error and results in https://github.com/quantumlib/Cirq/blob/master/cirq/protocols/act_on_protocol.py#L102

NotImplemented allows the protocol to try the fallback strategy.

If the first `act_on` fails, the state has not been modified and we are free to try the fallback. If the first `act_on` succeeds but some subsequent one fails, the state might be unintentionally modified and we should fail fast.

But based on comments by @balopat and you, this tri-state behavior might be too messy while adding minimal value. 

I decided to do away with it and `assert` in the way you suggest here. This will cause us to fail when any of the listed gates fails to `act_on`(including the first one). But given that these are well-established Clifford gates that should have `_act_on_` implemented, even if the first one fails, it is actually more prudent to assert fail right here rather than keeping an option of a fallback strategy.

cirq/ops/common_gates.py
smitsanghavi: Agreed, I was trying to have it trigger that error but asserting as you mentioned above seems like a better way since these gates must have `_act_on_` implemented. So failing right there would be better both design and debugging wise.

Thanks!

cirq/ops/common_gates.py
```diff
@@ -103,6 +103,16 @@ def _act_on_(self, args: Any):
                 tableau.xs[:, q] ^= tableau.zs[:, q]
             return True
 
+        if isinstance(args, clifford.ActOnStabilizerCHFormArgs):
+            if protocols.is_parameterized(self) or self.exponent % 0.5 != 0:
+                return NotImplemented
+            assert all(
+                gate._act_on_(args) for gate in  # type: ignore
+                [HPowGate(),
```
Strilanc, line 111: You should be able to use `cirq.H` (except from the file it's defined in, not the global scope).

cirq/ops/common_gates.py
```diff
@@ -322,6 +332,29 @@ def _act_on_(self, args: Any):
...
+        if isinstance(args, clifford.ActOnStabilizerCHFormArgs):
+            if protocols.is_parameterized(self) or self.exponent % 0.5 != 0:
+                return NotImplemented
+            effective_exponent = self._exponent % 2
+            state = args.state
+            if effective_exponent == 0.5:
+                assert all(
+                    gate._act_on_(args)  # type: ignore
+                    for gate in [ZPowGate(), HPowGate()])
+                state.omega *= (1 + 1j) / (2**0.5)  # type: ignore
```
Strilanc, line 344: Why is the type ignore needed?

Strilanc (approved):

cirq/ops/common_gates.py
smitsanghavi: mypy assumes state.omega is an int since it was initialized by an int. It complains when I try to assign a complex value to it. Maybe adding `+ 0j` to the initialization will fix it. Let me know if you think that's better.

cirq/ops/common_gates.py
smitsanghavi: Just used `HPowGate` to be consistent with `ZPowGate` (since importing pauli_gate for `Z` leads to circular dependency)

balopat (changes requested):
Left a bunch of comments! 
Also, I'm a bit worried about the fact that phases just go unnoticed...
In `assert_all_implemented_act_on_effects_match_unitary` shouldn't we use 

```
      np.testing.assert_allclose(np.reshape(
            stabilizer_ch_form.state_vector(), protocols.qid_shape(qubits)),
                                                   state_vector,
                                                   atol=1e-07, err_msg="inconsistent stabilizer ch form")
```
 instead of 

```
        testing.assert_allclose_up_to_global_phase(np.reshape(
            stabilizer_ch_form.state_vector(), protocols.qid_shape(qubits)),
                                                   state_vector,
                                                   atol=1e-07)
```
? 
The whole benefit of the CH-form is that it is phase sensitive. Why are we checking unitary consistency in a phase insensitive way?

cirq/ops/common_gates.py
smitsanghavi: Fixed it

cirq/ops/common_gates.py
```diff
@@ -322,6 +332,29 @@ def _act_on_(self, args: Any):
...
+                assert all(
+                    gate._act_on_(args) for gate in  # type: ignore
+                    [ZPowGate(), HPowGate(),
+                     ZPowGate(), HPowGate()])
+                state.omega *= 1j  # type: ignore
+            elif effective_exponent == 1.5:
+                assert all(
+                    gate._act_on_(args)  # type: ignore
+                    for gate in [HPowGate(), ZPowGate()])
+                state.omega *= (1 - 1j) / (2**0.5)  # type: ignore
```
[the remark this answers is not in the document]
balopat: > Left a bunch of comments!
> Also, I'm a bit worried about the fact that phases just go unnoticed...
> In `assert_all_implemented_act_on_effects_match_unitary` shouldn't we use
> 
> ```
>       np.testing.assert_allclose(np.reshape(
>             stabilizer_ch_form.state_vector(), protocols.qid_shape(qubits)),
>                                                    state_vector,
>                                                    atol=1e-07, err_msg="inconsistent stabilizer ch form")
> ```
> 
> instead of
> 
> ```
>         testing.assert_allclose_up_to_global_phase(np.reshape(
>             stabilizer_ch_form.state_vector(), protocols.qid_shape(qubits)),
>                                                    state_vector,
>                                                    atol=1e-07)
> ```
> 
> ?
> One of the benefits of the CH-form is that it is phase sensitive. Why are we checking unitary consistency in a phase insensitive way?

It seems that all the gate implementations for some reason are off by a factor of 1j. I.e if I test with `np.testing.assert_allclose` and `state_vector * 1j` then they all succeed...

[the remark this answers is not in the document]
balopat: For the record: I was wrong with the -1 phase for the YPowGate - it seems to be fine. I made a mistake `u(Circuit(H(a), Z(a))` should have been instead of `u(Circuit(H(a), Z(a))`

cirq/ops/common_gates_test.py
```diff
@@ -326,14 +326,45 @@ def test_x_act_on():
...
 
+def test_x_act_on_ch_form():
+    with pytest.raises(TypeError, match="Failed to act"):
+        cirq.act_on(cirq.X, object())
+    original_state = cirq.StabilizerStateChForm(num_qubits=5, initial_state=31)
+    flipped_state = cirq.StabilizerStateChForm(num_qubits=5, initial_state=23)
+
+    args = cirq.ActOnStabilizerCHFormArgs(state=original_state.copy(), axes=[1])
+
+    cirq.act_on(cirq.X**0.5, args, allow_decompose=False)
```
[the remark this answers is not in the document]
smitsanghavi: That's a good idea! Done, much cleaner now. Will probably do the same for the existing CliffordTableau tests in a subsequent PR.

cirq/ops/common_gates.py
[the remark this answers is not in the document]
smitsanghavi: I could not find any issue with the implementation of the gates or off-by `1j` factor issue. Even after taking phases into account, only the tests with `global_shift` being introduced by `assert_eigengate_implements_consistent_protocols` were failing. Added additional accounting based on that to `common_gates` instances.

Also, `decompose_rotation` in `SingleQubitCliffordGate` was messing up the global phase. Added a (slightly expensive) post-correction there.

cirq/ops/common_gates.py
```diff
@@ -1098,6 +1179,23 @@ def _act_on_(self, args: Any):
...
             return True
 
+        if isinstance(args, clifford.ActOnStabilizerCHFormArgs):
+            if protocols.is_parameterized(self) or self.exponent % 1 != 0:
+                return NotImplemented
+            q1 = args.axes[0]
+            q2 = args.axes[1]
+            state = args.state
+            if self._exponent % 2 == 1:
+                # Prescription for CZ left multiplication.
```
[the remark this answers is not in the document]

cirq/testing/consistent_act_on.py
```diff
@@ -41,19 +54,40 @@ def state_vector_has_stabilizer(state_vector: np.ndarray,
...
 
     # pylint: disable=unused-variable
     __tracebackhide__ = True
     # pylint: enable=unused-variable
 
     num_qubits_val = protocols.num_qubits(val)
 
-    if protocols.is_parameterized(val) or not protocols.has_unitary(
-            val) or protocols.qid_shape(val) != (2,) * num_qubits_val:
+    if protocols.is_parameterized(val) or \
```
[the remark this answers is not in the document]
smitsanghavi: Done. Merge issue.

cirq/testing/consistent_act_on.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019 The Cirq Developers
+# Copyright 2020 The Cirq Developers
```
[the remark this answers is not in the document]
smitsanghavi: Done. File was created in 2020 though.

cirq/ops/common_gates_test.py
```diff
@@ -516,38 +659,69 @@ def test_cz_act_on():
...
-        cirq.CX**4,
-        cirq.CZ**4,
-        # Gates not supported by CliffordTableau should not fail too.
-        cirq.X**0.25,
-        cirq.Y**0.25,
-        cirq.Z**0.25,
-        cirq.H**0.5,
-        cirq.CX**0.5,
-        cirq.CZ**0.5
+        (cirq.X, True, True),
```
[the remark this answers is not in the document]
smitsanghavi: Done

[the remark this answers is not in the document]
smitsanghavi: Keeping them separate in the consistent_act_on file though. The idea is to enable adding incremental support for `CliffordTableau` or `StabilizerStateChForm` to new gates without having to add both at once.

smitsanghavi:
> Left a bunch of comments!
> Also, I'm a bit worried about the fact that phases just go unnoticed...
> In `assert_all_implemented_act_on_effects_match_unitary` shouldn't we use
> 
> ```
>       np.testing.assert_allclose(np.reshape(
>             stabilizer_ch_form.state_vector(), protocols.qid_shape(qubits)),
>                                                    state_vector,
>                                                    atol=1e-07, err_msg="inconsistent stabilizer ch form")
> ```
> 
> instead of
> 
> ```
>         testing.assert_allclose_up_to_global_phase(np.reshape(
>             stabilizer_ch_form.state_vector(), protocols.qid_shape(qubits)),
>                                                    state_vector,
>                                                    atol=1e-07)
> ```
> 
> ?
> One of the benefits of the CH-form is that it is phase sensitive. Why are we checking unitary consistency in a phase insensitive way?

That makes sense. There were some tests adding extra global phases that were breaking it previously. Added support for `global_shift` to fix this.

cirq/testing/consistent_act_on.py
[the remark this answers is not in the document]
balopat: ohhhh, then 2020 is better :D

balopat (approved):
LGTM, thank you for all the hard work on this!!! Can you just address the two last nits? Thank you!

cirq/testing/consistent_act_on.py
```diff
@@ -41,19 +53,39 @@ def state_vector_has_stabilizer(state_vector: np.ndarray,
...
+    Args:
+        val: A gate or operation that may be an input to protocols.act_on.
+        assert_tableau_implemented: asserts that protocols.act_on() works with
+          val and ActOnCliffordTableauArgs inputs.
+        assert_ch_form_implemented: asserts that protocols.act_on() works with
+          val and ActOnStabilizerStateChFormArgs inputs.
+    """
 
     # pylint: disable=unused-variable
-    __tracebackhide__ = True
```
balopat, line 50: Why did we remove this?

cirq/ops/common_gates.py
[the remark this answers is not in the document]
balopat: Good find! That was the culprit that I saw! :)

cirq/testing/consistent_act_on.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 The Cirq Developers
```
balopat: this is okay to roll back given that it was created this year - sorry for my confusion

cirq/testing/consistent_act_on.py
smitsanghavi: My bad, was trying to induce some tests to fail here to verify it was working properly. Reverted.

cirq/testing/consistent_act_on.py
smitsanghavi: Np. Done

PR #3204, opened 2020-08-11
Remove cirq v1 programs from namespace

dstrain115:
- Cirq v1 protos are deprecated.
- No one should have been actively using these functions for over a year.

### DIFF
--- a/cirq/google/api/v1/programs_test.py
+++ b/cirq/google/api/v1/programs_test.py
@@ -20,7 +20,6 @@ import cirq
 import cirq.google as cg
 import cirq.google.api.v1.programs as programs
 from cirq.google.api.v1 import operations_pb2
-from cirq.google.api.v1.programs import (_parameterized_value_from_proto)
 
 
 def assert_proto_dict_convert(gate: cirq.Gate, proto: operations_pb2.Operation,
@@ -92,7 +91,7 @@ def test_pack_results():
             [1, 0],
         ])),
     ]
-    data = cg.pack_results(measurements)
+    data = programs.pack_results(measurements)
     expected = make_bytes("""
         000 00
         001 01
@@ -108,7 +107,7 @@ def test_pack_results():
 
 
 def test_pack_results_no_measurements():
-    assert cg.pack_results([]) == b''
+    assert programs.pack_results([]) == b''
 
 
 def test_pack_results_incompatible_shapes():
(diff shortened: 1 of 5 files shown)


cirq/google/api/v1/programs_test.py
```diff
@@ -20,7 +20,6 @@
 import cirq.google as cg
 import cirq.google.api.v1.programs as programs
```
wcourtney, line 21: Do you want v2 programs here?

cirq/google/optimizers/convert_to_xmon_gates.py
```diff
@@ -65,6 +64,19 @@ def _convert_one(self, op: 'cirq.Operation') -> 'cirq.OP_TREE':
...
+    def _is_native_xmon_op(self, op: 'cirq.Operation') -> bool:
+        """Check if the gate within an operation is a native xmon gate.
+
+        Args:
+            op: Input operation.
+
+        Returns:
+            True if the operation is native to the xmon, false otherwise.
+        """
+        return (isinstance(op, ops.GateOperation) and isinstance(
```
wcourtney, line 76: Does this differ from the XMON gateset? https://github.com/quantumlib/Cirq/blob/059006d951541d399bbd8fba0f3eb43553dcc18b/cirq/google/gate_sets.py#L92

If it doesn't, can we reference that as the source of truth for the supported gates?

cirq/google/api/v1/programs_test.py
dstrain115: Presumably that is tested here:
https://github.com/quantumlib/Cirq/blob/master/cirq/google/api/v2/program.py

dstrain115:
> Is the plan to leave the proto definitions in place for posterity, or will they be removed in a future PR?

I hope to remove the v1 protos eventually, but probably not for a few months.  I'll at least let the rest of the changes soak for a while.  Plus, we need to get rid of them in engine side as well.

cirq/google/optimizers/convert_to_xmon_gates.py
dstrain115: Done.  Had to import within the function to avoid circular deps, but otherwise should be okay.

dstrain115: Note: it actually needs to reference cirq.google.XmonDevice (the device) not cirq.google.XMON (the gateset) since those two are actually slightly out of sync, due to PhasedXZ.

cirq/google/optimizers/convert_to_xmon_gates.py
```diff
@@ -65,6 +64,19 @@ def _convert_one(self, op: 'cirq.Operation') -> 'cirq.OP_TREE':
...
+    def _is_native_xmon_op(self, op: 'cirq.Operation') -> bool:
+        """Check if the gate within an operation is a native xmon gate.
+
+        Args:
+            op: Input operation.
+
+        Returns:
+            True if the operation is native to the xmon, false otherwise.
+        """
+        from cirq.google.devices import XmonDevice
```
wcourtney, line 76: Is an inlined import acceptable style for this project? I feel like I was once critiqued for that, but I can't recall for sure :-)

balopat: it is - unfortunately circular dependencies are forcing us to do this in quite a few places

PR #3212, opened 2020-08-13
Disable cirq v1 api docs

dstrain115:
The v1 api is deprecated and no one should be using it.

### DIFF
--- a/dev_tools/docs/build_api_docs.py
+++ b/dev_tools/docs/build_api_docs.py
@@ -54,7 +54,8 @@ def main(unused_argv):
             "cirq.google.engine.client.quantum.QuantumEngineServiceClient":
             ["enums"],
             "cirq.google.engine.client.quantum_v1alpha1.QuantumEngineServiceClient":
-            ["enums"]
+            ["enums"],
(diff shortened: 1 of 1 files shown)


lamberta:
Just FYI, when generating API docs, we use the `build_api_docs.py` on the `master` branch. So that will need to be updated to preview before this lands in master

balopat:
So then, if I understand correctly this should be merged against master? I'm okay with that too. @dstrain115 do you want to change the base of this PR to be master?

dstrain115:
The two branches are too dissimilar.  Maybe a git whiz could figure out how to rebase, but for me, it is just easier to make a new PR to merge to master:

This is continued in this PR:
https://github.com/quantumlib/Cirq/pull/3215

lamberta:
> The two branches are too dissimilar. Maybe a git whiz could figure out how to rebase, but for me, it is just easier to make a new PR to merge to master

Yes, this is a concern for me as we continue to work. The git workflow is a bit confusing and I'm worried about keeping things up-to-date while working on separate branches, and the final merge.

Issue #3105, opened 2020-06-24
Test in convert_to_ion_gates_test fails due to not using approximate equality testing


dabacon:
test_convert_to_ion_gates fails at least sporadically

```
E       assert [cirq.PhasedX...t(0, 1)), ...] == [cirq.PhasedX...t(0, 1)), ...]
E         At index 0 diff: cirq.PhasedXPowGate(phase_exponent=-0.7500000000000002, exponent=0.5).on(cirq.GridQubit(0, 0)) != cirq.PhasedXPowGate(phase_exponent=-0.75, exponent=0.5).on(cirq.GridQubit(0, 0))
E         Full diff:
E           [
E         -  cirq.PhasedXPowGate(phase_exponent=-0.75, exponent=0.5).on(cirq.GridQubit(0, 0)),
E         +  cirq.PhasedXPowGate(phase_exponent=-0.7500000000000002, exponent=0.5).on(cirq.GridQubit(0, 0)),
E         ?                                          ++++++++++++++
E         -  cirq.PhasedXPowGate(phase_exponent=1, exponent=0.25).on(cirq.GridQubit(0, 1)),
E         +  cirq.PhasedXPowGate(phase_exponent=1.0, exponent=0.25).on(cirq.GridQubit(0, 1)),
E         ?                                      ++
E         -  cirq.T(cirq.GridQubit(0, 0)),
E         +  (cirq.Z**0.25000000000000044).on(cirq.GridQubit(0, 0)),
E            cirq.ms(-0.5*np.pi/2).on(cirq.GridQubit(0, 0), cirq.GridQubit(0, 1)),
E         -  (cirq.Y**0.5).on(cirq.GridQubit(0, 0)),
E         +  cirq.PhasedXPowGate(phase_exponent=0.5, exponent=0.5).on(cirq.GridQubit(0, 0)),
E         -  cirq.PhasedXPowGate(phase_exponent=1, exponent=0.25).on(cirq.GridQubit(0, 1)),
E         +  cirq.PhasedXPowGate(phase_exponent=1.0, exponent=0.25).on(cirq.GridQubit(0, 1)),
E         ?                                      ++
E            (cirq.Z**-0.75).on(cirq.GridQubit(0, 0)),
E           ]

cirq/ion/convert_to_ion_gates_test.py:68: AssertionError
```

Issue #3103, opened 2020-06-23
Broken decompositions_test


dabacon:
test_single_qubit_op_to_framed_phase_form_output_on_example_case has flaked out on me.

dabacon:
This looks like it is blocking submissions to master, so may not even be a flake.

dabacon:
```
E       Mismatched elements: 2 / 4 (50%)
E       Max absolute difference: 1.
E       Max relative difference: 1.41421356
E        x: array([[ 7.071068e-01-2.236690e-17j, -2.707207e-16-7.071068e-01j],
E              [-7.071068e-01+3.735782e-16j, -2.707207e-16-7.071068e-01j]])
E        y: array([[7.071068e-01+5.551115e-17j, 5.551115e-17-7.071068e-01j],
E              [5.551115e-17-7.071068e-01j, 7.071068e-01+5.551115e-17j]])
```

dabacon:
it looks like the code that fixes for global phase isn't working for this case?  Cannot repo on my local box.

dabacon:
Hypothesis is that this might be a numpy update.  https://pypi.org/project/numpy/#history was just updated to 1.19.  I confirmed that the last successful build was on 1.18 and the new failures are 1.19.  Checking to see if pinning fixes.

dabacon:
Confirmed that penning to 1.16 numpy makes test pass.

dabacon:
confirmed that is also fine a numpy 1.18

tai271828:
confirmed that
  - numpy 1.19.0rc1 breaks decompositions_test
  - numpy 1.18.5 passes decompositions_test

tai271828:
The same test case will pass with numpy=1.19.0 on my local machine (amd64, ubuntu 16.04.6). I dived into the difference between my local run and the github action CI, and the eigen vector from `np.linalg.eig(mat)` of  https://github.com/quantumlib/Cirq/blob/master/cirq/optimizers/decompositions.py#L142 differs.

mpharrigan:
This test is failing for me. It's currently xfailed on windows with a reference to #2468 which is closed

tai271828:
@mpharrigan may you please share your debugging information to us? e.g.

1. The output of `pip list`, or any method that could tell us your numpy version
2. Which cirq version you are using?

maffoo:
Is this the only blocker for numpy 1.19? @dabacon, are you actively working on fixing the test?

balopat:
[...]

PR #3214, opened 2020-08-14
Updating the requirement for numpy 

exAClior:
### Purpose
Updates requirement.txt to let numpy be >= 1.19.1 
Tested for issues 
### Related issues
#3105 #3103 

### Future Lookouts
quimb imports from numpy.matlib , it is deprecated since numpy 1.19.0


requirements.txt
```diff
@@ -5,7 +5,7 @@ freezegun~=0.3.15
 google-api-core[grpc] >= 1.14.0, < 2.0.0dev
 matplotlib~=3.0
 networkx~=2.4
-numpy~=1.16, < 1.19
+numpy>=1.19.1
```
balopat, line 8: Thanks for opening this! 
I would prefer keeping it `~=1.19.1` so that we don't get auto-upgraded suddenly to the next minor version.

exAClior: My bad, modified! A side note though, the previous `numpy>=1.19.1` would somehow break the pytest ubuntu and the coverage test too. I have tested on my machine which runs manjaro with python 3.7 and 3.6 with numpy of 1.19.1. No error shows up when I run check/all command.

balopat: Interesting! We should get to the bottom of this - do you mind testing a couple of times with `>=1.19.1` and `~=` by pushing different versions? It shouldn't matter at this moment!!! 

It seems that this is a flaky situation but the [failing build on your previous commit](https://github.com/quantumlib/Cirq/runs/982916285) did reproduce https://github.com/quantumlib/Cirq/issues/3103. We shouldn't merge until we confirm that there is no flake - i.e. if you can run this test on your ubuntu machine a 100 times and it won't fail, then we are confident that it works.

balopat: Oh one more thing - the best course of action in case of flakes is to understand the root cause of why these flakes are happening and then resolve that. If you can pinpoint the reason why that test failed in the first place - that would give us enough confidence to upgrade to numpy 1.19+ without introducing flakes. Thank you for your investigations on this!

exAClior: > Oh one more thing - the best course of action in case of flakes is to understand the root cause of why these flakes are happening and then resolve that. If you can pinpoint the reason why that test failed in the first place - that would give us enough confidence to upgrade to numpy 1.19+ without introducing flakes. Thank you for your investigations on this!

Got it, will investigate later!

balopat:
Friendly ping @exAClior - did you have time to look into the flakes?

exAClior:
@balopat  terribly sorry, I was very occupied with other things  since it is the beginning of the semester. Maybe I could find sometime to work on that in the next few days or is there anyone willing to take on this?

balopat:
> @balopat terribly sorry, I was very occupied with other things since it is the beginning of the semester. Maybe I could find sometime to work on that in the next few days or is there anyone willing to take on this?

Absolutely no worries - I'd suggest closing this PR in the meantime. I will pick this up myself or find someone in the team :)

PR #3216, opened 2020-08-14
Add Moment.__sub__ and generalize Moment.__add__

Strilanc:
- `Moment.__add__` now accepts any `OP_TREE`
- `Moment.__sub__` takes an `OP_TREE` and returns a moment with those operations removed
- If there is an operation-to-remove that was not present, an error is raised
- Also bump mypy version to fix an incompatibility with python 3.8 and fix new warnings
    - Add explicit !r to format statements touching bytes
    - Rename `value_equality.py` to avoid name collision with the decorator it defines
    - Ignore false positives related to abstract base type bounds

### DIFF
--- a/dev_tools/auto_merge.py
+++ b/dev_tools/auto_merge.py
@@ -56,7 +56,7 @@ class PullRequestDetails:
 
         if response.status_code != 200:
             raise RuntimeError(
-                'Pull check failed. Code: {}. Content: {}.'.format(
+                'Pull check failed. Code: {}. Content: {!r}.'.format(
                     response.status_code, response.content))
 
         payload = json.JSONDecoder().decode(response.content.decode())
@@ -139,7 +139,7 @@ def check_collaborator_has_write(repo: GithubRepository, username: str
 
     if response.status_code != 200:
         raise RuntimeError(
-            'Collaborator check failed. Code: {}. Content: {}.'.format(
+            'Collaborator check failed. Code: {}. Content: {!r}.'.format(
                 response.status_code, response.content))
 
     payload = json.JSONDecoder().decode(response.content.decode())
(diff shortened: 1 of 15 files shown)


dev_tools/auto_merge.py
```diff
@@ -509,7 +508,7 @@ def attempt_sync_with_master(pr: PullRequestDetails
     data = {
         'base': pr.branch_name,
         'head': master_sha,
-        'commit_message': 'Update branch (automerge)'.format(pr.branch_name)
+        'commit_message': 'Update branch (automerge)'
```
viathor, line 511: The original looks like it was accidentally missing a {}. Suggestion:

```python
'commit_message': 'Update branch {pr.branch_name} (automerge)'
```

dev_tools/auto_merge.py
```diff
@@ -534,8 +533,8 @@ def attempt_sync_with_master(pr: PullRequestDetails
             "'Update Branch' for me before trying again.")
 
     raise RuntimeError('Sync with master failed for unknown reason. '
-                       'Code: {}. Content: {}.'.format(response.status_code,
-                                                       response.content))
+                       'Code: {}. Content: {!r}.'.format(
+                           response.status_code, response.content))
```
viathor, line 537: Use a format string?

(Applies throughout.)

dev_tools/auto_merge.py
Strilanc: No, it was intentional. When I changed it to not mention the branch name I just forgot to remove the format call.

dev_tools/auto_merge.py
Strilanc: Acknowledged. Wanted to keep the refactoring minimal.

PR #3217, opened 2020-08-14
Minor Correction to Dev Docs.

abhik-99:
The Developer Documentation had the two lines below which needed change.

```bash
git clone git@github.com:quantumlib/cirq.git
cd cirq
```

and 

```bash
git clone git@github.com:USERNAME/cirq.git
cd cirq
```
The directory name after cloning is ```Cirq``` and not ```cirq```. Changed the casing to match the directory name.


abhik-99:
@googlebot I signed it!

PR #3226, opened 2020-08-18
Process fidelity using the Kraus representation

tonybruguier:
issue #2763

Add a utility function that computes the process fidelity. It uses the Kraus representation for the computation of the traces.


tonybruguier:
Chatted with @Strilanc , @balopat and @cduck during the meeting. If I understand correctly:

- It seems the representation format of open circuit is not completely settled, so we might have to wait.
- The code to compose Kaus representation is indeed exponential (as is in this PR)
- this apply_channel code could be extended to circuit (please point to where, if it is indeed the way to go).

Thanks everyone for the help. Happy to wait for advice.

cirq/qis/measures.py
```diff
@@ -69,3 +74,53 @@ def von_neumann_entropy(density_matrix: np.ndarray) -> float:
...
+    d = 2**n
+    qubit_map = {q.with_dimension(1): i for i, q in enumerate(qubits)}
+
+    kraus_operations = [clean_circuit.unitary().reshape([2] * (2 * n))]
+    for op in noisy_circuit.all_operations():
+        target_axes = [qubit_map[q.with_dimension(1)] for q in op.qubits]
+
+        next_kraus_operations = []
+
+        for op_kraus in protocols.channel(op, default=None):
```
mpharrigan, line 110: Am I right in thinking of this loop as basically a new noisy simulator? This should likely be factored out. Of course @viathor will not hesitate to point out that we need to nail down the story on our representations for channels. I know our existing noise channels describe themselves with Kraus operators, but as you can see here: they get a little clunky to compose

viathor: +1 to both factoring this out (good to do in this PR) and representation (out of scope for this PR).

Also, I suppose "a little clunky" is a euphemism for "exponentially clunky" ;-)

balopat:
This is blocked on https://github.com/quantumlib/Cirq/issues/3248

cirq/linalg/operator_sum_utils.py
```diff
@@ -0,0 +1,54 @@
...
+# limitations under the License.
+"""Utils for the computation of operator sum (Kraus operators)."""
+
+import numpy as np
+
+from cirq.linalg.transformations import targeted_left_multiply
+from cirq import protocols
+
+
+def compute_kraus_operations(initial_density_matrix, noisy_circuit, qubits):
```
vtomole, line 22: Add types for mypy.

cirq/qis/measures.py
```diff
@@ -69,3 +71,37 @@ def von_neumann_entropy(density_matrix: np.ndarray) -> float:
...
+
+
+def process_fidelity(clean_circuit, noisy_circuit, qubits) -> float:
+    """Calculates the average fidelity of a noisy circuit.
+
+    The code uses the Kraus representation for open circuits, when decomposing
+    into noisy channels. The formula for process fidelity can be found at
+    equation (2) of "Quantum Gate Fidelity in Terms of Choi Matrices" by
+    Nathaniel Johnston and David W. Kribs which can be found at:
+    https://arxiv.org/pdf/1102.0948.pdf
```
vtomole, line 83: Let's link to the abstract instead of the pdf: https://arxiv.org/abs/1102.0948

cirq/qis/measures.py
```diff
@@ -69,3 +71,37 @@ def von_neumann_entropy(density_matrix: np.ndarray) -> float:
...
+
+    Another useful reference is "A simple formula for the average gate fidelity
+    of a quantum dynamical operation" by Michael A. Nielsen which can be found
+    at:
+    https://arxiv.org/pdf/quant-ph/0205035.pdf
```
vtomole, line 88: Change link to https://arxiv.org/abs/quant-ph/0205035

cirq/qis/measures.py
```diff
@@ -69,3 +71,37 @@ def von_neumann_entropy(density_matrix: np.ndarray) -> float:
...
+    Returns:
+        A scalar that is the average (process) entropy
+    """
+    n = len(qubits)
+    d = 2**n
+
+    kraus_operations = compute_kraus_operations(
+        clean_circuit.unitary().reshape([2] * (2 * n)), noisy_circuit, qubits)
+
+    EiT = [x.reshape(d, d) for x in kraus_operations]
```
vtomole, line 103: Change `EiT` to all lower case

cirq/qis/measures.py
```diff
@@ -69,3 +71,37 @@ def von_neumann_entropy(density_matrix: np.ndarray) -> float:
...
+
+    sum_traces = sum([abs(np.trace(x))**2 for x in EiT])
```
vtomole, line 105: Change `EiT` to all lower case

cirq/qis/measures_test.py
```diff
@@ -129,3 +130,36 @@ def test_von_neumann_entropy():
         np.diag([0, 0, 0.1, 0, 0.2, 0.3, 0.4, 0])),
                       1.8464,
                       atol=1e-04)
+
+
+def test_process_fidelity_one_qubit():
+    q = cirq.LineQubit(0)
```
vtomole, line 136: Change to `qubit = cirq.LineQubit.range(1)`

cirq/qis/measures_test.py
```diff
@@ -129,3 +130,36 @@ def test_von_neumann_entropy():
...
+
+    clean_circuit = cirq.Circuit(cirq.X(q)**0.123)
+
+    noise_model = cirq.ConstantQubitNoiseModel(
+        cirq.AmplitudeDampingChannel(0.5))
+    noisy_circuit = cirq.Circuit(
+        noise_model.noisy_moments(clean_circuit.moments, [q]))
```
vtomole, line 143: Change `[q]` to `qubit`

cirq/qis/measures_test.py
```diff
@@ -129,3 +130,36 @@ def test_von_neumann_entropy():
...
+
+    assert np.isclose(cirq.process_fidelity(clean_circuit, clean_circuit, [q]),
```
vtomole, line 145: Change `[q]` to `qubit`

cirq/qis/measures_test.py
```diff
@@ -129,3 +130,36 @@ def test_von_neumann_entropy():
...
+                      1.0)
+    assert np.isclose(cirq.process_fidelity(clean_circuit, noisy_circuit, [q]),
```
vtomole, line 147: Change `[q]` to `qubit`

cirq/linalg/operator_sum_utils.py
```diff
@@ -0,0 +1,54 @@
+# Copyright 2019 The Cirq Developers
```
vtomole, line 1: Nit: change Copyright year to `2020`

vtomole (changes requested):

tonybruguier:
Thanks. I tried to address the comments, but I wasn't able to add the mypy annotations because they trigger circular imports. PTAL.

vtomole:
@tonybruguier Please paste those circular import errors and the types you added so that i may replicate the issue on my machine.

tonybruguier:
@vtomole 
This is the change:
https://github.com/quantumlib/Cirq/pull/3226/commits/a53b30fedef5cbe558dade71dcd90fb35d52743b

And I get this error:
`
ImportError while loading conftest '/home/ubuntu/Dropbox/AWS_EC2_shared/git/process_fidelity/cirq/conftest.py'.
cirq/__init__.py:30: in <module>
    from cirq import (
cirq/linalg/__init__.py:66: in <module>
    from cirq.linalg.operator_sum_utils import compute_kraus_operations
cirq/linalg/operator_sum_utils.py:18: in <module>
    from cirq.circuits import Circuit
cirq/circuits/__init__.py:21: in <module>
    from cirq.circuits.qasm_output import (
cirq/circuits/qasm_output.py:22: in <module>
cirq/ops/__init__.py:21: in <module>
    from cirq.ops.clifford_gate import (
cirq/ops/clifford_gate.py:21: in <module>
    from cirq.ops import common_gates, gate_features, named_qubit, pauli_gates
cirq/ops/common_gates.py:36: in <module>
    from cirq.ops import controlled_gate, eigen_gate, gate_features, raw_types
cirq/ops/controlled_gate.py:21: in <module>
    from cirq.ops import raw_types, controlled_operation as cop
cirq/ops/controlled_operation.py:31: in <module>
    from cirq import protocols, qis, value
undo
cirq/qis/__init__.py:1: in <module>
    from cirq.qis.measures import (
cirq/qis/measures.py:20: in <module>
    from cirq.linalg.operator_sum_utils import compute_kraus_operations
E   ImportError: cannot import name 'compute_kraus_operations' from partially initialized module 'cirq.linalg.operator_sum_utils' (most likely due to a circular import) (/home/ubuntu/Dropbox/AWS_EC2_shared/git/process_fidelity/cirq/linalg/operator_sum_utils.py)
`

vtomole:
@tonybruguier  I see that type annotation is in this pull request and the format check is passing.

tonybruguier:
Oh, I undid the failure. Here was the breakage:
https://github.com/quantumlib/Cirq/runs/1477821679?check_suite_focus=true

Thanks for all the help!

cirq/linalg/operator_sum_utils.py
```diff
@@ -0,0 +1,56 @@
...
+# limitations under the License.
+"""Utils for the computation of operator sum (Kraus operators)."""
+
+import numpy as np
+
+from cirq.linalg.transformations import targeted_left_multiply
+from cirq import protocols
+
+
+def compute_kraus_operations(initial_density_matrix: np.ndarray, noisy_circuit, qubits):
```
vtomole, line 22: Add type annotations to the `noisy_circuit`, `qubits` variables and also the return type "list of Kraus operators"

rtd_docs/api.rst
```diff
@@ -0,0 +1,773 @@
+
```
vtomole, line 1: This file looks like it doesn't belong in this PR. Please remove it.

cirq/qis/measures.py
```diff
@@ -276,3 +277,38 @@ def von_neumann_entropy(
     if validate:
         _ = quantum_state(state, qid_shape=qid_shape, copy=False, validate=True, atol=atol)
     return 0.0
+
+
+def process_fidelity(clean_circuit, noisy_circuit, qubits) -> float:
```
vtomole, line 282: Add type annotations to `clean_circuit`, and `noisy_circuit` and `qubits`

cirq/qis/measures.py
```diff
@@ -276,3 +277,38 @@ def von_neumann_entropy(
...
+    https://arxiv.org/abs/quant-ph/0205035
+
+    Args:
+        clean_circuit: The perfect circuit (no noise, closed).
+        noisy_circuit: The circuit with noise gates (open circuit).
+        qubits: The list of qubits.
+    Returns:
+        A scalar that is the average (process) entropy
+    """
+    n = len(qubits)
```
vtomole, line 303: Do we need to pass in qubits? Why not call `circuit.all_qubits` instead?

vtomole (changes requested):
I know this PR is blocked: Just wanted to add a couple of comments i noticed while skimming it.

tonybruguier:
Thanks. PTAL.

cirq/qis/measures.py
daxfohl: Could we change this to a pure linalg function like `compose_kraus_operations`, that takes the initial density matrix and like `operations: Iterable[Tuple[Sequence[ndarray], Sequence[int]]]`, (i.e. a list of (kraus, axes) tuples)? Then this would be a useful general function, not bound to our definition of a circuit. We could have another function that calculates the fidelity of that result.

If we have those, then we don't necessarily need to "standardize" on a representation. We can leave this as pure linalg, and anyone can use it by writing a couple lines to extract the (kraus, axes) list from their noisy circuit and feed it into here. (Maybe we offer that extraction as a separate utility function elsewhere).

tonybruguier:
@daxfohl Like so?

daxfohl:
Yes, I think that's better with the circuit bits removed from linalg. I was thinking the reshaping line could go into the linalg function too. But I don't know the math here so well; I only suggest that because it happens to be a piece of pure linalg so it seems like it could be pushed lower. If doing so would make the resulting function seem wonky and less reusable, then keep it as is.

PR #3236, opened 2020-08-19
Feature request template

balopat:
Inspired by the SciPy FR template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_requet.md
+++ b/.github/ISSUE_TEMPLATE/feature_requet.md
@@ -1,0 +1,19 @@
+---
+name: Feature request
+about: Submit a feature request 
+title: ''
+labels: 'kind/feature-request'
+assignees: ''
+
+---
+
+**Is your feature request related to a use case or problem? Please describe.**
+
+
+**Describe the solution you'd like**
+
+
+**Describe alternatives you've considered**
+
+
+**Additional context (e.g. screenshots)**
--- a/.github/ISSUE_TEMPLATE/feature_requet.md
+++ b/.github/ISSUE_TEMPLATE/feature_requet.md
@@ -13,7 +13,18 @@ assignees: ''
 **Describe the solution you'd like**
 
 
-**Describe alternatives you've considered**
+**Describe alternatives/workarounds you've considered**
 
 
 **Additional context (e.g. screenshots)**
+
+
+**What is the urgency from your perspective for this issue? Is it blocking important work?**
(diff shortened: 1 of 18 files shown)


.github/ISSUE_TEMPLATE/feature_requet.md
```diff
@@ -0,0 +1,30 @@
+---
```
dstrain115, line 1: Name of this file is mispelled.

.github/ISSUE_TEMPLATE/feature_requet.md
```diff
@@ -0,0 +1,30 @@
...
+**Describe alternatives/workarounds you've considered**
+
+
+**Additional context (e.g. screenshots)**
+
+
+**What is the urgency from your perspective for this issue? Is it blocking important work?**
+<!-- Please choose one and remove the others -->
+
+P1 - this should land no later than a week
```
dstrain115, line 25: It seems like all of these priorities should be one less.  P0 = less than a week, etc.  That seems closer to how we rank things internally.

.github/ISSUE_TEMPLATE/feature_requet.md
```diff
@@ -0,0 +1,30 @@
...
+
+---
+
+**Is your feature request related to a use case or problem? Please describe.**
+
+
+**Describe the solution you'd like**
+
+
+**Describe alternatives/workarounds you've considered**
```
dstrain115, line 16: Should we mark this and the next one as optional?  I don't want to put too many barriers in the way of people asking for features.

balopat: SGTM!

.github/ISSUE_TEMPLATE/feature_requet.md
balopat: ooops! Good eye!

.github/ISSUE_TEMPLATE/feature_requet.md
balopat: I'm open to it. 
My thinking was that I'd use a single priority labelling for both bugs and features - and I just wanted to hold P0 for hotfixes which I thought would only be for bugs - I can hardly imagine the lack of a new feature being critical/harmful that needs a hotfix immediately (the class I would use for P0) while bugs can reach that level.

dstrain115: Ah, that's a good point.  In the same vein, do we really have features that need to be done within a week?!?

PR #3251, opened 2020-08-20
[automerge/cirqbot] paging for PR events

### LINKED ISSUE #3250: Fix CirqBot event retrieval logic on long PRs
In `check_auto_merge_labeler` we only pull the first page of events. For longer PRs this makes it seem that Github is inconsistent - https://api.github.com/repos/quantumlib/cirq/issues/3227/events doesn't contain any of the events for labelling with `automerge` - but CirqBot expects it - maybe we can improve the error message here. 

We need to implement paging logic. 

_Originally posted by @balopat in https://github.com/quantumlib/Cirq/pull/3227#issuecomment-676828018_

balopat:
Fixes #3250.

Adds paging capability. 
Adds per_page=100 for the events retrieval in a PR, that should avoid paging in most cases (though there might be some PRs with 100+ events)  

As we don't have tests yet for this - I tested it manually on my fork using my personal API token. 
FYI - this version is running currently - so if it works on this PR, it will be a live test ;)


balopat:
> How did the semi-colon separator before access token work in the past..???

My guess is that the per-page parameter's value got tossed away and this was an unauthenticated call.

balopat:
Actually, I'm wrong, semicolon is "recommended" to be supported as a query parameter separator! 
https://stackoverflow.com/questions/3481664/semicolon-as-url-query-separator#:~:text=So%20W3C%20recommends%20that%20web,easier%20to%20write%20these%20URLs.&text=To%20make%20matters%20worse%2C%20until,must%20use%20%26%20for%20others.

PR #3252, opened 2020-08-20
Fixing a bunch of typos

balopat:


balopat:
Haha, I did use pycharm's spell checker. It was very noisy but I filtered out some signal. I'm sure there are other typos still.

PR #3255, opened 2020-08-20
Add CPMG to T2 experiment

### LINKED ISSUE #3001: Add CPMG to T2 experiment
The T2 decay experiment currently supports T2 and T2*, also known as Hahn echo and Ramsey, but it would be nice to also support a CPMG style T2 experiment.

dstrain115:
- This adds Carr-Purcell-Meiboom-Gill pulse sequences to the T2
experiment.
- This sequence has the following sequence:
π/2, t, π, 2t, π, ... 2t, π, t

Since there are two variables (N pulses and t delay), this allows
the user to modify either variable or both to create a 2d scan.

In order to accomplish this efficiently, the pi-pulses are encoded as
parameterized X gates, e.g. X ** sympy.symbol('pulse_3') which is
only turned on for more than 3 pulses.

- Fixes #3001


cirq/experiments/t2_decay_experiment.py
```diff
@@ -163,21 +200,77 @@ def t2_decay(
...
+
+    The X gates are paramterizd by 'pulse_N' symbols so that pulses can be
+    turned on and off.  This is done to combine circuits with different pulses
+    into the same paramterized circuit.
+    """
+    circuit = circuits.Circuit(
+        ops.Y(qubit)**0.5,
+        ops.WaitGate(value.Duration(nanos=delay_var))(qubit), ops.X(qubit))
+    for n in range(max_pulses):
+        pulse_n_on = sympy.Symbol(f'pulse_{n}')
```
viathor, line 250: No action needed: I think I understand why you're doing this. We shouldn't have to. Filed #3266.

cirq/experiments/t2_decay_experiment.py
```diff
@@ -155,6 +170,28 @@ def t2_decay(
...
+        circuit.append(ops.X(qubit)**inv_x_var)
+        circuit.append(ops.Y(qubit)**inv_y_var)
+        circuit.append(ops.measure(qubit, key='output'))
+        tomography_sweep = study.Zip(
+            study.Points('inv_x', [0.0, 0.5]),
+            study.Points('inv_y', [-0.5, 0.0]),
+        )
+        pulse_sweep = _cpmg_sweep(num_pulses)
+        sweep = study.Product(delay_sweep, pulse_sweep, tomography_sweep)
+
```
viathor, line 194: The experiments are very similar to each other and consequently there is a lot of repetition here. For example, `tomography_sweep` is defined three times.

Also, something that makes conceptual sense and will help reduce repetition is the separation of T2 and tomography circuits (the latter is exactly the same in each experiment).

I think it should be possible to have only one block of code to handle all three experiments in a readable way.

cirq/experiments/t2_decay_experiment.py
```diff
@@ -214,13 +307,21 @@ def _expectation(self, data) -> pd.DataFrame:
...
         zeros = data[0]
         pauli_expectation = (2 * (ones / (ones + zeros))) - 1.0
-        return pd.DataFrame({'delay_ns': xs, 'value': pauli_expectation})
+        if 'num_pulses' in data.columns:
+            return pd.DataFrame({
+                'delay_ns': delay,
+                'num_pulses': data['num_pulses'],
+                'value': pauli_expectation
+            })
+        return pd.DataFrame({'delay_ns': delay, 'value': pauli_expectation})
```
viathor, line 324: missing type annotation on `data` argument up in the function signature (github won't let me comment there).

cirq/experiments/t2_decay_experiment.py
```diff
@@ -163,21 +200,77 @@ def t2_decay(
...
+    tabulation = pd.crosstab(cols, measurements.output).reset_index()
+    # If all measurements are 1 or 0, fill in the missing column with all zeros.
+    for col_index, name in [(1, 0), (2, 1)]:
+        if name not in tabulation:
+            tabulation.insert(col_index, name, [0] * tabulation.shape[0])
+    return tabulation
+
+
+def _cpmg_circuit(qubit: devices.GridQubit, delay_var: sympy.Symbol,
+                  max_pulses: int):
```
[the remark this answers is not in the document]
viathor: This will also remove inconsistency in conventions for the interpretation of `delay_var` between your Hahn echo and CPMG code (in the former delay_var is *half* the time between the initial pi/2 pulse and the subsequent pi pulse, but in CPMG it is exactly the time between the pi/2 and pi).

Not sure whether we care about backwards compatibility here. If not, I think your CPMG convention is to be preferred.

cirq/experiments/t2_decay_experiment.py
```diff
@@ -163,21 +200,77 @@ def t2_decay(
...
+    be 0.
+    """
+    pulse_points = []
+    for n in range(max(num_pulses)):
+        pulse_points.append(
+            study.Points(f'pulse_{n}', [1 if p > n else 0 for p in num_pulses]))
+    return study.Zip(*pulse_points)
+
+
 class T2DecayResult:
```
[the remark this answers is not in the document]
dstrain115: My understanding is that this is the trickiest and most subtle piece.  The calibration team is reluctant to even give me a standard way to calculate it, since it is vulnerable to error in several ways.  

I have been avoiding doing it and considering leaving the final calculation to the user unless I can get consensus on how to do this properly.  

For now, out of scope.

cirq/experiments/t2_decay_experiment.py
```diff
@@ -70,8 +71,20 @@ def t2_decay(
     The same method of measuring the final state as Ramsey experiment is applied
     after the second half of the delay period.
 
-    CPMG, or the Carr-Purcell-Meiboom-Gill sequence, is currently not
-    implemented.
+    CPMG, or the Carr-Purcell-Meiboom-Gill sequence, involves sending a sequence
+    of pi pulses (X gates) in a specific timing pattern:
+    π/2, t, π, 2t, π, ... 2t, π, t
```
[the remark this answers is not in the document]
dstrain115: I just included the final reference, since that is what we are actually doing.  I also expanded the explanation.

I have not found a good paper that summarizes the procedure from a quantum computing perspective, which is sad for poor computer science people like me :(

cirq/experiments/t2_decay_experiment.py
```diff
@@ -87,6 +100,8 @@ def t2_decay(
              of nanoseconds.  If specified, this will override the max_delay and
              min_delay parameters.  If not specified, the experiment will sweep
              from min_delay to max_delay with linear steps.
+        num_pulses: For CPMG, a list of the number of pulses to use.
+             If multiple pulses are specified, each will be swept on.
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/experiments/t2_decay_experiment.py
```diff
@@ -163,21 +200,77 @@ def t2_decay(
...
+            axis=1)
+
+    x_basis_tabulation = _create_tabulation(x_basis_measurements)
+    y_basis_tabulation = _create_tabulation(y_basis_measurements)
 
     # Return the results in a container object
     return T2DecayResult(x_basis_tabulation, y_basis_tabulation)
 
 
+def _create_tabulation(measurements):
```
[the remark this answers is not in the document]
dstrain115: Added.

cirq/experiments/t2_decay_experiment.py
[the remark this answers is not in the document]
dstrain115: Done.  I doubt people are using this, but. if so. I have broken their timing.

cirq/experiments/t2_decay_experiment.py
dstrain115: I combined the echo code and CPMG.  I am not sure if it is worth the hit in readability to try to combine ramsey as well.  I think the preference would be to have a bit of verboseness since this is basically an extended example.

cirq/experiments/t2_decay_experiment.py
dstrain115: Added.

cirq/experiments/t2_decay_experiment.py
```diff
@@ -70,8 +71,20 @@ def t2_decay(
     The same method of measuring the final state as Ramsey experiment is applied
     after the second half of the delay period.
 
-    CPMG, or the Carr-Purcell-Meiboom-Gill sequence, is currently not
-    implemented.
+    CPMG, or the Carr-Purcell-Meiboom-Gill sequence, involves sending a sequence
+    of pi pulses (X gates) in a specific timing pattern:
```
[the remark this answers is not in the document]
dstrain115: I have changed the description.  Let me know if find further inaccuracies.

cirq/experiments/t2_decay_experiment.py
```diff
@@ -68,10 +69,36 @@ def t2_decay(
...
-
-    CPMG, or the Carr-Purcell-Meiboom-Gill sequence, is currently not
-    implemented.
+    after the second half of the delay period.  See the animation on the wiki
+    page https://en.wikipedia.org/wiki/Spin_echo for a visual illustration
+    of this experiment.
+
+    CPMG, or the Carr-Purcell-Meiboom-Gill sequence, involves using a sqrt(Y)
+    followed by a sequence of pi pulses (X gates) in a specific timing pattern:
+        π/2, t, π, 2t, π, ... 2t, π, t
```
viathor, line 78: nit: maybe add empty line before and after for readability and consistency with other docstrings in cirq.

cirq/experiments/t2_decay_experiment.py
```diff
@@ -68,10 +69,36 @@ def t2_decay(
...
+    The first pulse, a sqrt(Y) gate, will put the qubit's state on the Bloch
+    equator.  After a delay, successive X gates will refocus dehomogenous
+    phase effects by causing them to precess in opposite directions and
+    averaging their effects across the entire pulse train.
+
+    This pulse pattern has two variables that can be adjusted.  The first,
+    denoted as 't' in the above sequence, is delay, which can be specified
+    with `delay_min` and `delaxy_max` or by using a `delay_sweep`, similar to
+    the other experiments.  The second variable is the number of pi pulses
+    (X gates)  This can be specified as a list of integers using the
```
viathor, line 88: nit: missing dot after "(X gates)"

cirq/experiments/t2_decay_experiment.py
```diff
@@ -68,10 +69,36 @@ def t2_decay(
...
+    `num_pulses` parameter.  If multiple different pulses are specified,
+    the data will presented in a data frame with two
```
viathor, line 90: nit: missing "be" after "will"

cirq/experiments/t2_decay_experiment.py
```diff
@@ -68,10 +69,36 @@ def t2_decay(
...
+
+    See the following reference for more information about CPMG pulse trains:
+    Meiboom, S., and D. Gill, “Modified spin-echo method for measuring nuclear
+    relaxation times”, Rev. Sci. Inst., 29, 688–691 (1958).
+    https://doi.org/10.1063/1.1716296
+
+    Note that interpreting T2 data is fairly tricky and subtle, as it can
+    include other effects that need to be accounted for.  For instance,
+    amplitude damping (T1) will present as T2 noise and needs to be
+    appropriately compensated for to find a true measure of T2.
```
viathor, line 101: Maybe add a remark saying that this is why the class does not actually return a T2 estimate at present.

cirq/experiments/t2_decay_experiment.py
viathor: Yes, leaving Ramsey circuit on its own is ok.

That said, you should move `tomography_sweep` out of the conditionals since it is literally identical for all experiments.

You should also move `sweep` outside the experiment-type-dependent blocks since the initialization you do for CPMG will work for Ramsey too (with a minor tweak to the condition on max_pulses).

cirq/experiments/t2_decay_experiment.py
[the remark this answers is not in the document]
viathor: SGTM

cirq/experiments/t2_decay_experiment.py
```diff
@@ -133,51 +164,120 @@ def t2_decay(
...
+            tabulation.insert(col_index, name, [0] * tabulation.shape[0])
+    return tabulation
+
+
+def _cpmg_circuit(qubit: devices.GridQubit, delay_var: sympy.Symbol,
+                  max_pulses: int) -> 'cirq.Circuit':
+    """Creates a CPMG circuit for a given qubit.
+
+    The circuit will look like:
+      sqrt(Y) - wait(delay_var)  - X - wait(2*delay_var) - ... - wait(delay_var)
```
viathor, line 246: nit: too many whitespaces after first wait gate

dstrain115: Done.

cirq/experiments/t2_decay_experiment.py
dstrain115: Done I guess.

cirq/experiments/t2_decay_experiment.py
dstrain115: Done.

cirq/experiments/t2_decay_experiment.py
dstrain115: Done.

cirq/experiments/t2_decay_experiment.py
dstrain115: Done.

dstrain115 (commented):
>That said, you should move tomography_sweep out of the conditionals since it is literally identical for all experiments.

>You should also move sweep outside the experiment-type-dependent blocks since the initialization you do for CPMG will work for Ramsey too (with a minor tweak to the condition on max_pulses).

Tomography sweep is not equal, since there is an extra Y in Hahn echo and CPMG so the X has to flip in the other direction in order to use the same condition when calculating pauli expectations.

I did move the sweep calculation out of the conditional though.

cirq/experiments/t2_decay_experiment.py
```diff
@@ -68,10 +69,41 @@ def t2_decay(
...
+        π/2, t, π, 2t, π, ... 2t, π, t
+
+    The first pulse, a sqrt(Y) gate, will put the qubit's state on the Bloch
+    equator.  After a delay, successive X gates will refocus dehomogenous
+    phase effects by causing them to precess in opposite directions and
+    averaging their effects across the entire pulse train.
+
+    This pulse pattern has two variables that can be adjusted.  The first,
+    denoted as 't' in the above sequence, is delay, which can be specified
+    with `delay_min` and `delaxy_max` or by using a `delay_sweep`, similar to
```
viathor, line 88: nit: delaxy_max -> delay_max

cirq/experiments/t2_decay_experiment.py
```diff
@@ -132,52 +169,117 @@ def t2_decay(
...
+        # Performs the following sequence
+        # π/2 - wait(t) - π - wait(2t) - ... - π - wait(t)
+        # There will be N π pulses (X gates)
+        # where N sweeps over the values of num_pulses
+        #
+        if not num_pulses:
+            raise ValueError('At least one value must be given '
+                             'for num_pulses in a CPMG experiment')
+        circuit = _cpmg_circuit(qubit, delay_var, max_pulses)
+        circuit.append(ops.X(qubit)**inv_x_var)
```
viathor, line 194: This part:

```python
circuit.append(ops.X(qubit)**inv_x_var)
circuit.append(ops.Y(qubit)**inv_y_var)
circuit.append(ops.measure(qubit, key='output'))
tomography_sweep = study.Zip(
    study.Points('inv_x', [0.0, 0.5]),
    study.Points('inv_y', [-0.5, 0.0]),
)
```

is state tomography and AFAIU is independent of which type of experiment is performed. It is also duplicated above for Ramsey experiment (modulo a minus sign which does not seem to matter).

Perhaps I'm missing something, but I don't see why we need to implement tomography twice in this function. If indeed we don't then please reduce the indent from here until line 200 and remove lines 164-166 and 168-171.

This will reduce code verbosity and make the code better reflect what is happening here by introducing a clear separation between the core of each T2 experiment and state tomography (which is just a generic method of characterizing quantum states).

---

It would be even better if we did not implement tomography here at all and instead called [`cirq. single_qubit_state_tomography`](https://github.com/quantumlib/Cirq/blob/master/cirq/experiments/qubit_characterizations.py#L357). WDYT?

viathor (commented):
This looks good, except one final point of code duplication, see below.

cirq/experiments/t2_decay_experiment.py
viathor: Let's defer. Filed #3316.

viathor (approved):
Deferred the duplication issue and filed #3316. Remaining issues are nits. LGTM!

cirq/experiments/t2_decay_experiment_test.py
```diff
@@ -297,8 +369,33 @@ def test_bad_args():
                                       delay_sweep=product)
 
 
-def test_str():
+def test_cpmg_circuit():
+    """Tests sub-component to make sure CPMG circuit is generated correctly."""
+    q = cirq.GridQubit(1, 1)
+    t = sympy.Symbol('t')
+    circuit = t2._cpmg_circuit(q, t, 2)
```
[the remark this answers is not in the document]
dstrain115: I understand the sentiment, but I think I will leave these in for now.  It's hard to come up with good noise models that are also testable, which makes testing the output of the experiment tricky.  I think, in this case, it's worth testing the intermediate values of the circuit and sweeps to make sure it is doing the correct experiment, rather than just testing the result of the experiment.

cirq/experiments/t2_decay_experiment.py
```diff
@@ -132,52 +169,117 @@ def t2_decay(
...
+            tabulation.insert(col_index, name, [0] * tabulation.shape[0])
+    return tabulation
+
+
+def _cpmg_circuit(qubit: devices.GridQubit, delay_var: sympy.Symbol,
+                  max_pulses: int) -> 'cirq.Circuit':
+    """Creates a CPMG circuit for a given qubit.
+
+    The circuit will look like:
+      sqrt(Y) - wait(delay_var) - X - wait(2*delay_var) - ... - wait(delay_var)
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/experiments/t2_decay_experiment.py
dstrain115: Done.

cirq/experiments/t2_decay_experiment.py
dstrain115: I'll leave using cirq. single_qubit_state_tomography to another PR.  I moved the tomography outside the conditional.

cirq/experiments/t2_decay_experiment_test.py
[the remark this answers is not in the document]
viathor: Fair enough.

PR #3256, opened 2020-08-20
Fix sweep behavior on dictionary of lists

### LINKED ISSUE #2698: Expand SWEEP_LIKE to work for {'t': [0, 1, 3]}

kevinsung:
Fixes #2698 .

### DIFF
--- a/cirq/study/sweepable.py
+++ b/cirq/study/sweepable.py
@@ -30,7 +30,7 @@ document(
     """Dictionary from symbols to sequence of values taken.""")
 
 Sweepable = Union[ParamResolverOrSimilarType, Sweep, SweepableDictType,
-                  Iterable[Union[ParamResolverOrSimilarType, Sweep]], None]
+                  Iterable[Union[ParamResolverOrSimilarType, Sweep, SweepableDictType]], None]
 document(
     Sweepable,  # type: ignore
     """An object or collection of objects representing a parameter sweep.""")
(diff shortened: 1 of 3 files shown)


cirq/study/sweepable.py
```diff
@@ -16,35 +16,44 @@
...
-        yield from sweepable
-    elif isinstance(sweepable, dict):
-        yield ParamResolver(cast(Dict, sweepable))
-    elif isinstance(sweepable, Iterable) and not isinstance(sweepable, str):
-        for item in cast(Iterable, sweepable):
-            yield from to_resolvers(item)
-    else:
-        raise TypeError(f'Unrecognized sweepable type: {type(sweepable)}.\n'
-                        f'sweepable: {sweepable}')
+    for sweep in to_sweeps(sweepable):
```
balopat, line 42: I believe that the intuition here is good to merge `to_resolvers` logic with `to_sweeps` as `Sweepable` -> `List[Sweep]` -> `Iterator[ParamResolver]` does make sense to me, and that unifies the behavior. 

However, I think we have to be a bit more careful when merging the logic: 

1. when `sweepable` is a ParamResolver, now it will be converted to a Sweep then back to ParamResolver
2. if it's a Sweep we should just yield paramResolvers from it...

kevinsung:
@balopat Thanks for the analysis. I'll implement something along those lines, though it will be a bit different due to what we agreed on in #2698 . I'll probably do it in separate PRs, which will make this one obsolete.

kevinsung:
Obsoleted by #3330, #3333, and #3343.

PR #3259, opened 2020-08-21
Split right multiplication into separate methods in StabilizerStateCHForm

smitsanghavi:
#2423 

This will allow each method to have a definite action which will allow us to move the left multiplication logic to individual gates #2948. Also makes the update_sum method public so that the gates can call it instead of duplicating the logic.


cirq/sim/clifford/stabilizer_state_ch_form.py
```diff
@@ -296,4 +295,4 @@ def project_Z(self, q, z):
         if np.all(t == u):
             self.omega /= np.sqrt(2)
 
-        self._update_sum(t, u, delta=delta)
+        self.update_sum(t, u, delta=delta)
```
balopat, line 298: why do we need to make `_update_sum` public?

cirq/sim/clifford/stabilizer_state_ch_form.py
```diff
@@ -159,27 +161,16 @@ def _Y(self, q):
         self._X(q)
         self.omega *= 1j
 
-    def _CZ(self, q, r, right=False):
```
balopat, line 162: I'm not sure I understand why you are removing the right vs left multiplication, can you explain, I don't get the "This will allow each gate to have a definite action" argument if that's why?
As far as I see, this structure maps nicely to the paper's definition of  and   .

cirq/sim/clifford/stabilizer_state_ch_form.py
```diff
@@ -54,7 +54,13 @@ def __init__(self,
         self.v = np.zeros(self.n, dtype=bool)
         self.s = np.zeros(self.n, dtype=bool)
 
-        self.omega = 1
+        # Keep initial state positive by transferring the negative part to the
+        # global phase.
+        if initial_state < 0:
```
balopat, line 59: When does negative state occur?

cirq/sim/clifford/stabilizer_state_ch_form.py
balopat: Oh I just read your description: you want to expose this so that the gates can call it directly.

cirq/sim/clifford/stabilizer_state_ch_form.py
smitsanghavi: #2948  These methods defining individual Clifford gate behavior are going to be deleted and we will rely on the implementation of `CZPowGate._act_on` to control how the `StabilizerStateCHForm` evolves. `_act_on_` does not have access to an extra right-ness argument like here and has to stick to the default "left" implementation.

So we either move the `right=True` logic to the caller (`update_sum`) right now (as this PR does) or split it out into a `_CZ_right` function which I believe is more messy.

cirq/sim/clifford/stabilizer_state_ch_form.py
smitsanghavi: Since StabilizerStateCHForm keeps track of the global phase, a negative initial state should just be the positive one with a -1 phase. Currently, it was using the negative number bit representation which is strictly worse IMO.

cirq/sim/clifford/stabilizer_state_ch_form.py
balopat: OK, got it. 
I have two thoughts here: 
- act_on is a "left multiplication"-only, do we want to have a right multiplication version of act_on? 
- I don't think that having these lines in themselves are readable: they are completely disconnected from the gates, and will be very hard to link them back to the original paper. At least a comment should indicate that this is the right multiplication implementation of CZ, but I think it would be better to keep it in a named private method (_CZ_right) - why do you think that it's more messy?
```
self.G[:, q] ^= self.G[:, i]
self.F[:, i] ^= self.F[:, q]
self.M[:, q] ^= self.M[:, i]
```

balopat (changes requested):
Let's resolve the comments before going forward.

cirq/sim/clifford/stabilizer_state_ch_form.py
balopat: Got it - as discussed let's make this into an error case instead - let's force computational basis state initial states.

cirq/sim/clifford/stabilizer_state_ch_form.py
smitsanghavi: Created _right methods

cirq/sim/clifford/stabilizer_state_ch_form.py
smitsanghavi: Will make that change in a separate PR since initial_state is an arg for multiple classes. Would prefer to do put them all in a single PR.

balopat (approved):

Issue #3248, opened 2020-08-19
Representing quantum channels


viathor:
At the moment the [`channel` protocol](https://github.com/quantumlib/Cirq/blob/master/cirq/protocols/channel.py) returns a sequence of Kraus operators. This representation of quantum channels is cumbersome. For example, naively composing k channels leads to growth in the number of operators that is exponential in k (to arrest the growth of the number of operators one needs to re-express them in an operator basis manually after composition). Another drawback of the representation is that it is ambiguous (different sets of Kraus operators may describe the same channel). This makes it tricky to compare channels.

It seems that a more practical representation arises from the Choi–Jamiołkowski isomorphism (aka state-channel duality). For a start, its mypy type is simpler: `np.ndarray` rather than `Sequence[np.ndarray]`. Also, it makes composition and comparisons easier. Finally, it simplifies tasks such as the computation of process fidelity(*).

One way to deal with the issue is to switch to the state representation of channels. Another, would be to encapsulate channels in a type that hides the representation and provides methods for tasks such as the application of the channel to a state, the computation of process fidelity, etc.

Filing for discussion.

Context: #3226.

---
(*) Process fidelity can be computed from the process matrix (in computational basis) with one matrix-vector product and one dot product (and in the Pauli basis with a single lookup).

maffoo:
> Another, would be to encapsulate channels in a type that hides the representation and provides methods for tasks such as the application of the channel to a state, the computation of process fidelity, etc.

I like this idea, though there are details to be worked out. I was thinking about this in the context of https://github.com/quantumlib/Cirq/pull/3249 where in some cases we might want to have a decomposed representation of the channel, even if the operation can't be decomposed. For example, the [measurement channel](https://github.com/quantumlib/Cirq/blob/fb9fbac0a534bba10df0501e3652d4a4edad5b5a/cirq/ops/measurement_gate.py#L111) as currently implemented requires a number of Krauss operators that is exponentially large in the number of qubits being measured. This channel can be straightforwardly decomposed into a phase damping channel on each measured qudit, but we don't want to implement `_decompose_` on the `MeasurementGate` itself because that would split it into multiple measurements with the same key.

maffoo:
[...]

PR #3262, opened 2020-08-22
Generalize asymmetric depolarization channel to multiple qubits

### LINKED ISSUE #3220: Multi-qubit channels
In [cirq/ops/common_channels.py](https://github.com/quantumlib/Cirq/blob/master/cirq/ops/common_channels.py) cirq implements a number of common quantum operations. All of them act on single-qubits only, but some of them (e.g. depolarizing channel) have clear definitions for multiple qubits. We should generalize them to handle multiple qubits.

tonybruguier:
Adds multi qubit generalization for asymmetric depolarization channel. 

Fixes #3220.


tonybruguier:
Thanks. PTAL.

cirq/ops/common_channels.py
```diff
@@ -229,6 +230,110 @@ def depolarize(p: float) -> DepolarizingChannel:
     return DepolarizingChannel(p)
 
 
+@value.value_equality
+class MultiAsymmetricDepolarizingChannel(gate_features.SingleQubitGate):
```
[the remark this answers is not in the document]
viathor: We still have two classes: `AsymmetricDepolarizingChannel` (pre-existing) and `MultiAsymmetricDepolarizingChannel` (new). Instead of adding `MultiAsymmetricDepolarizingChannel` this PR should add the new functionality to `AsymmetricDepolarizingChannel`.

tonybruguier:
Thanks. I am not sure that I understood what you exactly meant, but it's now all under a single class. Happy to change further if that's not what you had in mind. Apologies for the delay in getting back to you.

PTAL.

cirq/ops/common_channels.py
```diff
@@ -31,10 +31,20 @@
 class AsymmetricDepolarizingChannel(gate_features.SingleQubitGate):
     """A channel that depolarizes asymmetrically along different directions."""
 
-    def __init__(self, p_x: float, p_y: float, p_z: float) -> None:
+    def __init__(self, *args, **kwargs) -> None:
         r"""The asymmetric depolarizing channel.
 
         This channel applies one of four disjoint possibilities: nothing (the
```
viathor, line 37: "four" is incorrect here, the maximum number of terms is 4**n.

cirq/ops/common_channels.py
```diff
@@ -31,10 +31,20 @@
...
+        identity channel) or one of the 2**n pauli gates. The disjoint
```
viathor, line 38: The number of Pauli strings on n qubits is 4\*\*n, not 2\*\*n. This figure includes the identity.

cirq/ops/common_channels.py
```diff
@@ -31,10 +31,20 @@
...
+        probabilities of the 2**n gates.
+
+        This channel evolves a density matrix via
+
+            $$
+            \sum_i p_I Pi \rho Pi
```
viathor, line 44: p_I -> p_i

viathor, line 44: Add explanation of the symbols.

cirq/ops/common_channels.py
```diff
@@ -50,90 +60,156 @@ def __init__(self, p_x: float, p_y: float, p_z: float) -> None:
             $$
```
viathor, line 60: This can be removed now since a more general formula is added above.

cirq/ops/common_channels.py
```diff
@@ -31,10 +31,20 @@
 class AsymmetricDepolarizingChannel(gate_features.SingleQubitGate):
     """A channel that depolarizes asymmetrically along different directions."""
 
-    def __init__(self, p_x: float, p_y: float, p_z: float) -> None:
+    def __init__(self, *args, **kwargs) -> None:
```
viathor, line 34: The arguments should be explicit and type annotations should be added. Also, for backwards compatibility we should default num_qubits to 1:

```python
def __init__(self,
             p_x: Optional[float] = None,
             p_y: ...,
             p_z: ...,
             num_qubits: int = 1,
             ps: Optional[Dict[str, float]] = None) -> None:

File: cirq/ops/common_channels.py [hunk 7]
```diff
@@ -50,90 +60,156 @@ def __init__(self, p_x: float, p_y: float, p_z: float) -> None:
             $$
 
         Args:
+            num_qubits: Number of qubits
+            pis: Dictionary of string (Pauli operator) to its probability
```
viathor, line 64: Let's rename this to "ps".

cirq/ops/common_channels.py
```diff
@@ -50,90 +60,156 @@ def __init__(self, p_x: float, p_y: float, p_z: float) -> None:
...
             ValueError: if the args or the sum of args are not probabilities.
         """
-
-        self._p_x = value.validate_probability(p_x, 'p_x')
-        self._p_y = value.validate_probability(p_y, 'p_y')
-        self._p_z = value.validate_probability(p_z, 'p_z')
-        self._p_i = 1 - value.validate_probability(p_x + p_y + p_z,
-                                                   'p_x + p_y + p_z')
+        assert ('pis' in kwargs or 'p_x' in kwargs or 'p_y' in kwargs or
+                'p_z' in kwargs or len(args) > 0)
```
viathor, line 73: Let's remove the assertion. Once the argument list is explicit (see another comment) mypy will take care of checking this.

cirq/ops/common_channels.py
```diff
@@ -50,90 +60,156 @@ def __init__(self, p_x: float, p_y: float, p_z: float) -> None:
...
+        if 'pis' in kwargs:
```
viathor, line 74:
```python
if ps is not None:
    ...
```

cirq/ops/common_channels.py
```diff
@@ -50,90 +60,156 @@ def __init__(self, p_x: float, p_y: float, p_z: float) -> None:
...
+            assert 'num_qubits' in kwargs
```
viathor, line 75: Instead of asserting, we should raise `ValueError` when `ps` contains a key whose length is different than `num_qubits`. Note that it is ok not to specify `num_qubits` when we're creating a single-qubit channel.

cirq/ops/common_channels.py
```diff
@@ -50,90 +60,156 @@ def __init__(self, p_x: float, p_y: float, p_z: float) -> None:
...
-        return self._p_x, self._p_y, self._p_z
+        return self._num_qubits, hash(tuple(sorted(self._pis.items())))
 
     def __repr__(self) -> str:
-        return 'cirq.asymmetric_depolarize(p_x={!r},p_y={!r},p_z={!r})'.format(
-            self._p_x, self._p_y, self._p_z
-        )
+        return (
+            'cirq.asymmetric_depolarize' +
+            '(num_qubits={!r},pis={!r})'.format(self._num_qubits, self._pis))
```
viathor, line 125: Let's remove "num_qubits={!r}," if num_qubits is one (the default).

cirq/ops/common_channels.py
```diff
@@ -50,90 +60,156 @@ def __init__(self, p_x: float, p_y: float, p_z: float) -> None:
...
-                                                        self._p_z)
-        return 'A({!r},{!r},{!r})'.format(self._p_x, self._p_y, self._p_z)
+        if self._num_qubits == 1:
+            if args.precision is not None:
+                f = '{:.' + str(args.precision) + 'g}'
+                return 'A({},{},{})'.format(f, f,
+                                            f).format(self.p_x, self.p_y,
+                                                      self.p_z)
+            return 'A({!r},{!r},{!r})'.format(self.p_x, self.p_y, self.p_z)
+        else:
```
viathor, line 140: nit: Remove `else` and unindent the following block (since the "then" block above ends in return.

cirq/ops/common_channels.py
```diff
@@ -50,90 +60,156 @@ def __init__(self, p_x: float, p_y: float, p_z: float) -> None:
...
+                    pauli + ':' + '{!r}'.format(p)
+                    for pauli, p in self._pis.items()
+                ]
+            return 'A({})'.format(', '.join(ps))
 
     @property
     def p_x(self) -> float:
         """The probability that a Pauli X and no other gate occurs."""
-        return self._p_x
+        assert self._num_qubits == 1
```
viathor, line 157: This raises `AssertionError` which should be raised for conditions which we believe should not arise unless there is a bug.

We should raise `ValueError` which signifies that the user of this class made a mistake.

(Applies below, too.)

cirq/ops/common_channels.py
```diff
@@ -50,90 +60,156 @@ def __init__(self, p_x: float, p_y: float, p_z: float) -> None:
...
+        return self._pis.get('Z', 0.0)
+
+    @property
+    def num_qubits(self) -> int:
+        """The number of qubits"""
+        return self._num_qubits
+
+    @property
+    def pis(self) -> Dict[str, float]:
```
viathor, line 178: Let's call this "error_probabilities". This is more user friendly and consistent with cirq which general tends towards longer, more descriptive names.

viathor (changes requested):
This is much better! Thanks! I still have some comments, mostly about argument handling plus a few nits.

cirq/ops/common_channels.py
tonybruguier: Done.

cirq/ops/common_channels.py
tonybruguier: Done.

cirq/ops/common_channels.py
tonybruguier: Done.

tonybruguier: Done.

cirq/ops/common_channels.py
tonybruguier: Done.

cirq/ops/common_channels.py
tonybruguier: Done.

cirq/ops/common_channels.py
tonybruguier: Done.

cirq/ops/common_channels.py
tonybruguier: Just the function right?

cirq/ops/common_channels.py
tonybruguier: Done.

cirq/ops/common_channels.py
tonybruguier: Done.

tonybruguier:
Thanks. PTAL.

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,196 @@
...
 
-        This channel applies one of four disjoint possibilities: nothing (the
-        identity channel) or one of the three pauli gates. The disjoint
-        probabilities of the three gates are p_x, p_y, and p_z and the
-        identity is done with probability 1 - p_x - p_y - p_z. The supplied
-        probabilities must be valid probabilities and the sum p_x + p_y + p_z
-        must be a valid probability or else this constructor will raise a
-        ValueError.
+        This channel applies one of 4**n disjoint possibilities: nothing (the
+        identity channel) or one of the 4**n pauli gates. The disjoint
```
viathor, line 44: nit: "one of the 4**n - 1 Pauli gates" (minus one since you are counting identity separately)

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,196 @@
...
+        probabilities of the 4**n gates.
```
viathor, line 45: nit: "The disjoint probabilities of the 4**n gates." looks like an incomplete sentence? Maybe remove?

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,196 @@
...
 
         This channel evolves a density matrix via
 
             $$
-            \rho \rightarrow (1 - p_x - p_y - p_z) \rho
-                    + p_x X \rho X + p_y Y \rho Y + p_z Z \rho Z
+            \sum_i p_i Pi \rho Pi
             $$
 
+        Where i varies from 0 to 4**n-1, representing the 4**n Pauli gates,
```
viathor, line 53: nit: don't capitalize "where" since it continues the sentence above

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,196 @@
...
+        represented by Pi. The input \rho is the density matrix before the
```
viathor, line 54: nit: you're using "represent" twice in the sentence, maybe "where i varies from 0 to 4**n-1 and Pi represents n-qubit Pauli operator (including identity)".

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,196 @@
...
+        depolarization.
+
         Args:
             p_x: The probability that a Pauli X and no other gate occurs.
             p_y: The probability that a Pauli Y and no other gate occurs.
             p_z: The probability that a Pauli Z and no other gate occurs.
+            num_qubits: Number of qubits
+            error_probabilities: Dictionary of string (Pauli operator) to its
+                probability
```
viathor, line 63: nit: maybe add an example, so that it's clear to the user what they're supposed to pass in without having to read the code below

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,196 @@
...
+                        k, num_qubits))
+            self._num_qubits = num_qubits
+            self._error_probabilities = error_probabilities
+        else:
+            if p_x is None:
+                p_x = 0.0
+            if p_y is None:
+                p_y = 0.0
+            if p_z is None:
+                p_z = 0.0
```
viathor, line 81: Optional: you can reduce each of these conditionals to a one-liner like this:

```python
p_z = 0.0 if p_z is None else p_z
```

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,196 @@
...
+
+            self._num_qubits = 1
+            self._error_probabilities = {'I': p_i, 'X': p_x, 'Y': p_y, 'Z': p_z}
 
     def _mixture_(self) -> Sequence[Tuple[float, np.ndarray]]:
-        return ((self._p_i, protocols.unitary(identity.I)),
-                (self._p_x, protocols.unitary(pauli_gates.X)),
-                (self._p_y, protocols.unitary(pauli_gates.Y)),
-                (self._p_z, protocols.unitary(pauli_gates.Z)))
+        Pis = []
```
viathor, line 93: nit: We follow a convention that uses lower case letters for variable names. Perhaps just call it `ps` (this looks like plural of `p` which is the mathematical convention for probability).

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,196 @@
...
-        self._p_x = value.validate_probability(p_x, 'p_x')
-        self._p_y = value.validate_probability(p_y, 'p_y')
-        self._p_z = value.validate_probability(p_z, 'p_z')
-        self._p_i = 1 - value.validate_probability(p_x + p_y + p_z,
-                                                   'p_x + p_y + p_z')
+        if error_probabilities is not None:
+            for k in error_probabilities.keys():
+                if len(k) != num_qubits:
+                    raise ValueError('{} does not have {} entries'.format(
+                        k, num_qubits))
```
viathor, line 72: Also raise a `ValueError` if one of the letters is not in {I, X, Y, Z}.

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,196 @@
...
             ValueError: if the args or the sum of args are not probabilities.
         """
-
-        self._p_x = value.validate_probability(p_x, 'p_x')
-        self._p_y = value.validate_probability(p_y, 'p_y')
-        self._p_z = value.validate_probability(p_z, 'p_z')
-        self._p_i = 1 - value.validate_probability(p_x + p_y + p_z,
-                                                   'p_x + p_y + p_z')
+        if error_probabilities is not None:
+            for k in error_probabilities.keys():
```
viathor, line 69: Actually, we can remove the `num_qubits` argument and infer the number of qubits from the length of the first key in `error_probabilities`. Sorry for missing this earlier.

(We should still keep the for loop here to ensure all keys are of the same length. We can also keep the internal `_num_qubits` since you're making use of it elsewhere.)

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,196 @@
...
     def __repr__(self) -> str:
-        return 'cirq.asymmetric_depolarize(p_x={!r},p_y={!r},p_z={!r})'.format(
-            self._p_x, self._p_y, self._p_z
-        )
+        if self._num_qubits == 1:
+            return (
+                'cirq.asymmetric_depolarize(error_probabilities={!r})'.format(
+                    self._error_probabilities))
+        return ('cirq.asymmetric_depolarize' +
+                '(num_qubits={!r},error_probabilities={!r})'.format(
```
viathor, line 124: We can actually simplify this since we don't need the `num_qubits` argument (see other comment).

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,196 @@
...
+                    self._num_qubits, self._error_probabilities))
 
     def __str__(self) -> str:
-        return 'asymmetric_depolarize(p_x={!r},p_y={!r},p_z={!r})'.format(
-            self._p_x, self._p_y, self._p_z
-        )
+        if self._num_qubits == 1:
+            return 'asymmetric_depolarize(error_probabilities={!r})'.format(
+                self._error_probabilities)
+        return ('asymmetric_depolarize(num_qubits={!r},' +
```
viathor, line 131: ditto about `num_qubits`

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,196 @@
...
+            $$
+            \sum_i p_i Pi \rho Pi
+            $$
 
-    Args:
-        p_x: The probability that a Pauli X and no other gate occurs.
-        p_y: The probability that a Pauli Y and no other gate occurs.
-        p_z: The probability that a Pauli Z and no other gate occurs.
+        Where i varies from 0 to 4**n-1, representing the 4**n Pauli gates,
+        represented by Pi. The input \rho is the density matrix before the
```
viathor, line 208: nit: make the same style improvements suggested in another comment above

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,196 @@
...
+        depolarization.
+
+        Args:
+            p_x: The probability that a Pauli X and no other gate occurs.
+            p_y: The probability that a Pauli Y and no other gate occurs.
+            p_z: The probability that a Pauli Z and no other gate occurs.
+            num_qubits: Number of qubits
```
viathor, line 215: As above, this can be removed.

viathor (commented):
This looks really nice now. I have made one more comment that should simplify the API (we can remove num_qubits). The rest are minor nits.

Thank you for bearing with me here :-)

tonybruguier:
Thanks. PTAL.

cirq/ops/common_channels.py
```diff
@@ -31,109 +32,202 @@
...
         Args:
             p_x: The probability that a Pauli X and no other gate occurs.
             p_y: The probability that a Pauli Y and no other gate occurs.
             p_z: The probability that a Pauli Z and no other gate occurs.
+            error_probabilities: Dictionary of string (Pauli operator) to its
+                probability
+
+        Examples of calls:
+            * Single qubit: AsymmetricDepolarizingChannel(0.2, 0.1, 0.3)
+            * Single qubit: AsymmetricDepolarxizingChannel(p_z=0.3)
```
viathor, line 65: AsymmetricDepolarxizingChannel -> AsymmetricDepolarizingChannel (no x in "depolarizing')

cirq/ops/common_channels.py
```diff
@@ -31,109 +32,202 @@
...
 
-        This channel applies one of four disjoint possibilities: nothing (the
-        identity channel) or one of the three pauli gates. The disjoint
-        probabilities of the three gates are p_x, p_y, and p_z and the
-        identity is done with probability 1 - p_x - p_y - p_z. The supplied
-        probabilities must be valid probabilities and the sum p_x + p_y + p_z
-        must be a valid probability or else this constructor will raise a
-        ValueError.
+        This channel applies one of 4**n disjoint possibilities: nothing (the
+        identity channel) or one of the 4**n - 1 pauli gates.x
```
viathor, line 44: nit: remove trailing "x"

cirq/ops/common_channels.py
```diff
@@ -31,109 +32,202 @@
...
+        Examples of calls:
+            * Single qubit: AsymmetricDepolarizingChannel(0.2, 0.1, 0.3)
+            * Single qubit: AsymmetricDepolarxizingChannel(p_z=0.3)
+            * Two qubits: AsymmetricDepolarizingChannel(
+                                error_probabilities={'XX': 0.2})
 
         Raises:
             ValueError: if the args or the sum of args are not probabilities.
         """
+        if error_probabilities is not None and bool(error_probabilities):
```
viathor, line 72: This condition can be simplified:

```python
if error_probabilities:
    ...
```

cirq/ops/common_channels.py
```diff
@@ -31,109 +32,202 @@
...
+            num_qubits = len(list(error_probabilities.keys())[0])
```
viathor, line 73: No need to call `keys`.

cirq/ops/common_channels.py
```diff
@@ -31,109 +32,202 @@
...
+            regex = re.compile('(I|X|Y|Z){%d}' % (num_qubits))
```
viathor, line 74: Regular expressions are powerful, but expensive. You can easily check that a string is made up of a small set of characters and has a given length without regex:

```python
for k in error_probabilities:
    if not set(k).issubset({'I', 'X', 'Y', 'Z'}):
        raise ...
    if len(k) != num_qubits:
        raise ...
```

cirq/ops/common_channels.py
```diff
@@ -31,109 +32,202 @@
...
+            for k in error_probabilities.keys():
```
viathor, line 75: no need to call `keys`

cirq/ops/common_channels.py
```diff
@@ -31,109 +32,202 @@
...
+                if not regex.fullmatch(k):
+                    raise ValueError(
+                        ('{} does not have {} entries made ' +
```
viathor, line 78: This error message is confusing. Note that `error_probability` may contain entries made up of IXYZ and still fail if one of the other entries contains an illegal letter. Better error message:

"{} is not made solely of I, X, Y, Z."

cirq/ops/common_channels.py
```diff
@@ -31,109 +32,202 @@
...
+                         'of I, X, Y, and Z').format(k, num_qubits))
```
viathor, line 79: Use f-strings instead of the `format` function:

```python
raise ValueError(f"{k} is not made solely of I, X, Y, Z.")
```

cirq/ops/common_channels.py
```diff
@@ -31,109 +32,202 @@
...
+            self._num_qubits = 1
+            self._error_probabilities = {'I': p_i, 'X': p_x, 'Y': p_y, 'Z': p_z}
 
     def _mixture_(self) -> Sequence[Tuple[float, np.ndarray]]:
-        return ((self._p_i, protocols.unitary(identity.I)),
-                (self._p_x, protocols.unitary(pauli_gates.X)),
-                (self._p_y, protocols.unitary(pauli_gates.Y)),
-                (self._p_z, protocols.unitary(pauli_gates.Z)))
+        ps = []
+        for pauli in self._error_probabilities.keys():
```
viathor, line 98: no need to call `keys`

cirq/ops/common_channels.py
```diff
@@ -31,109 +32,202 @@
...
+            tuple(sorted(self._error_probabilities.items())))
 
     def __repr__(self) -> str:
-        return 'cirq.asymmetric_depolarize(p_x={!r},p_y={!r},p_z={!r})'.format(
-            self._p_x, self._p_y, self._p_z
-        )
+        if self._num_qubits == 1:
+            return (
+                'cirq.asymmetric_depolarize(error_probabilities={!r})'.format(
+                    self._error_probabilities))
```
viathor, line 126: This is identical to the return statement below. You can remove the entire conditional.

cirq/ops/common_channels.py
```diff
@@ -31,109 +32,202 @@
...
+        return ('cirq.asymmetric_depolarize(error_probabilities={!r})'.format(
+            self._error_probabilities))
 
     def __str__(self) -> str:
-        return 'asymmetric_depolarize(p_x={!r},p_y={!r},p_z={!r})'.format(
-            self._p_x, self._p_y, self._p_z
-        )
+        if self._num_qubits == 1:
+            return 'asymmetric_depolarize(error_probabilities={!r})'.format(
+                self._error_probabilities)
```
viathor, line 133: ditto

cirq/ops/common_channels.py
```diff
@@ -31,109 +32,202 @@
...
+            self._error_probabilities)
 
     def _circuit_diagram_info_(self,
                                args: 'protocols.CircuitDiagramInfoArgs') -> str:
+        if self._num_qubits == 1:
+            if args.precision is not None:
+                f = '{:.' + str(args.precision) + 'g}'
+                return 'A({},{},{})'.format(f, f,
+                                            f).format(self.p_x, self.p_y,
+                                                      self.p_z)
```
viathor, line 144: Use f-strings:

```python
return f"A({self.p_x:.{args.precision}g, {self.p_y:.{args.precision}g, {self.p_z:.{args.precision}g)"
```

(Applies below, too.)

viathor (commented):
Few more comments, mostly abut avoiding the use of regex for simple string analysis and about more idiomatic string formatting using f-strings.

tonybruguier:
Thanks. PTAL.

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,192 @@
...
         """
+        if error_probabilities:
+            num_qubits = len(list(error_probabilities)[0])
+            for k in error_probabilities.keys():
+                if not set(k).issubset({'I', 'X', 'Y', 'Z'}):
+                    raise ValueError(f"{k} is not made solely of I, X, Y, Z.")
+                if len(k) != num_qubits:
+                    raise ValueError(f"{k} must have {num_qubits} Pauli gates.")
+            self._num_qubits = num_qubits
+            self._error_probabilities = error_probabilities
```
viathor, line 79: Perform validation of each probability individually and the overall sum in a way that generalizes the validation below for the case of single-qubit.

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,192 @@
...
     def _value_equality_values_(self):
-        return self._p_x, self._p_y, self._p_z
+        return self._num_qubits, hash(
+            tuple(sorted(self._error_probabilities.items())))
 
     def __repr__(self) -> str:
-        return 'cirq.asymmetric_depolarize(p_x={!r},p_y={!r},p_z={!r})'.format(
-            self._p_x, self._p_y, self._p_z
-        )
+        return ('cirq.asymmetric_depolarize(error_probabilities={!r})'.format(
```
viathor, line 121: use f-string

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,192 @@
...
+            self._error_probabilities))
 
     def __str__(self) -> str:
-        return 'asymmetric_depolarize(p_x={!r},p_y={!r},p_z={!r})'.format(
-            self._p_x, self._p_y, self._p_z
-        )
+        return ('asymmetric_depolarize(error_probabilities={!r})').format(
```
viathor, line 125: ditto

cirq/ops/common_channels.py
```diff
@@ -31,109 +31,192 @@
...
+            error_probabilities = [
+                f"{pauli}:{p:.{args.precision}g}"
+                for pauli, p in self._error_probabilities.items()
+            ]
+        else:
+            error_probabilities = [
+                f"{pauli}:{p}"
+                for pauli, p in self._error_probabilities.items()
+            ]
+        return 'A({})'.format(', '.join(error_probabilities))
```
viathor, line 146: f-string

viathor (approved):
LGTM with two more comments: validate probabilities and use f-strings.

tonybruguier:
Thanks. PTAL.

viathor (approved):
LGTM! Thank you!

BackOnTrackIce:
Nice feature, I have one question how you could use it? The MatrixDensitySimulator is not capable of using it, i mean if you increase the number of n_qubits=2 you create the 16 Krauss operators but the simulator cannot handle it, is there another issue open somewhere?

tonybruguier:
I'm not an expert, but I understand it's because of the representation. I think there is some discussion there:
https://github.com/quantumlib/Cirq/issues/3248

Maybe ask there and the real experts can answer? I'm happy to do more work if you have suggestions, of course.

BackOnTrackIce:
Thanks a  lot for the hint!
Have a nice weekend

PR #3265, opened 2020-08-25
Add run_batch to Sampler and QuantumEngineSampler

### LINKED ISSUE #3224: Add run_batch to Sampler interface
This would be similar to `run_async`, in that it simply defers to `run` or `run_sweep`, and should be overridden by parent classes that can sample batches more efficiently. It would be overridden in QuantumEngineSampler.

If this should be done, I'm willing to work on it.

kevinsung:
Resolves #3224 .


cirq/work/sampler.py
```diff
@@ -199,3 +199,42 @@ async def run_sweep_async(
...
+        each (circuit, parameter sweep) pair. Child classes that are capable of
+        sampling batches more efficiently should override it to use other
+        strategies.
+
+        Args:
+            programs: The circuits to execute as a batch.
+            params_list: Parameter sweeps to use with the circuits. The number
+                of sweeps should match the number of circuits and will be
+                paired in order with the circuits.
+            repetitions: Number of circuit repetitions to run. Each sweep value
```
zaqqwerty, line 225: Quick question here, was wondering what the reasoning is around keeping the number of repetitions fixed across all circuits, rather than able to be specified per-circuit?  I imagine there are situations where a user might want to get a different number samples from each circuit.  For example, in equation 60 of [this paper](https://iopscience.iop.org/article/10.1088/1367-2630/18/2/023023/pdf) there is an expected number of samples required to achieve a desired precision that is different for each circuit.

kevinsung: Perhaps @dstrain115 or @maffoo can answer that question. Here I'm just following the interface of the already existing method `cirq.google.Engine.run_batch`.

dstrain115: Currently, you do not receive any speed up if you specify different numbers of repetitions per circuit since the hardware cannot batch them together.  This is specific to quantum engine, which makes me think that maybe we should not add this to Sampler.

balopat: @dstrain115, I agree that the batching is very specific _currently_ to Quantum Engine, and it seems that it might be too early to create this abstraction as we simply don't know how/whether other NISQ devices will utilize batching (we should probably ask the Pasqal/AQT/IQM folks!). 

On the other hand I see two drivers for this to happen: 
1) simulator/hardware gap and hardware/hardware gap - switching from simulation to hardware now has a friction point - at the extreme the user will have to maintain 5 different circuit execution strategies for simulation, Pasqal, AQT, IQM, Google Quantum Engine...it would be great to increase the overlap and reduce the amount of special code required for each architecture
2) A similar question comes up in case of downstream users - e.g. how will TFQ leverage batching in QE? It seems that the Sampler interface is the main entrypoint, it would simplify maintenance for them. If we keep run_batch on the QE level, then they'll have to implement logic that goes below the Sampler abstraction level. 

@kevinsung what is your main use case for this?

kevinsung: I was planning to use this to speed up XEB benchmarking: https://github.com/quantumlib/Cirq/blob/master/cirq/experiments/grid_parallel_two_qubit_xeb.py.
As is typical of the code in the `experiments/` directory, that code works at the Sampler abstraction level.

balopat: After a lot of thinking I came to the conclusion that the best course of action is to leave this functionality in Quantum Engine. The sole reasons this method was introduced was to exploit the QE specific optimized execution patterns. 
Bringing it up to the Sampler abstraction level creates confusion as we cannot easily answer the question "What is batching for?" in general. If the answer is "sometimes it's inconvenience without additional benefit, but achieves speedup on QE in these and these specific cases" then we're not ready to pull this up. As @zaqqwerty already raised it - it could be interpreted as a sweep over repetitions - but that is a very different interpretation. 

I would close this and introduce `cirq.google` specific logic in the XEB benchmarking code for optimal execution.

zaqqwerty: Sending batches of circuits is also useful for simulators: knowing all circuits upfront, rather than seeing them one at a time, allows various optimizations to be made.  For example the batch of circuits can be parallelized: in TFQ, in order to use the cirq density matrix simulator (which inherits from cirq.Sampler), we use the code [here](https://github.com/tensorflow/quantum/blob/4d36041d88b92d06ae08adc4ee78aae3c3627cbe/tensorflow_quantum/core/ops/batch_util.py#L575) to parallelize over batches of circuits.  If there were a `run_batch` call in the API of cirq.Sampler, then this parallelization logic could be moved into the implementation of the simulator.

zaqqwerty: So in our use case an answer to the question "What is batching for?" is to take advantage of multi-threaded machines to speed up simulation, and I think the reason to do it in the simulator itself is to reduce code duplication in the case that users besides us look to take advantage of this parallelization.

zaqqwerty: I wanted to clear up confusion around the phrase "it could be interpreted as a sweep over repetitions": I think it is not quite a sweep over repetitions, rather I meant that each circuit in the batch would be paired with a number of samples to extract from it; so that `len(repetitions) == len(programs)`.  The API could also accept a single number, in which case it pads out repetitions to contain `len(programs)` copies of the passed number.

zaqqwerty: Sorry for the several comments, just want to be sure our use case is understood; if it still seems too specific to merit the API change I understand.

zaqqwerty: Sorry for the several comments, just want to be sure our use case is understood; if it still seems too specific to merit the API change I understand.

mpharrigan:
how do you plan to support a development flow where you can test code on a simulator and then jump to "the real thing" while using batching?

balopat:
> how do you plan to support a development flow where you can test code on a simulator and then jump to "the real thing" while using batching?

Great question. 

My main hypothesis is that in the NISQ era we will not be able to invent/enforce a perfect abstraction that will work across all simulators and all quantum platforms that also allows for optimal execution on all these backends. 

To your question: anywhere outside the boundaries of "common functionality" the user will have to drop down to backend specific logic to leverage the specific features of that backend. I.e I might be completely off track here but something like: 

```
def get_samples(device) -> cirq.TrialResult: 
    circuits = create_circuits_sweeps()
    if device == cirq.google.Engine(): 
       batches = _prepare_batches(circuits)
       device.run_batch(batches)
    else: 
       device.run_sweep(circuits)
``` 

i.e. the user code will have to handle explicitly the cases of different platforms. 

Now, the Google code could offer a simulated version - to try out. That might be a good idea, so that you can test the batching logic in your code without having to run it against the real service. 

---- 
The question is where to draw that boundary and how. I think the boundary is around "unoptimized, generic workflows and/or universally applicable optimizations". If we can do optimization in a generic way, we should include that in our abstract classes. If we cannot, we should keep it in the subclasses. 

If we see that a certain optimization appears in most of our backends (e.g. batching turns out to be exactly the same for all our platforms) then it's probably worthwhile to pull it up the stack. But before we have this confidence it will just create confusion.

kevinsung:
Yes, as @maffoo said, there is no extra cost to using `run_batch` if the backend doesn't natively support it. So the answer to the question "What is batching for?" is simply that it may lead to faster executions on some backends. I don't think that is a confusing answer. The analogous answer could be given to the question "What is async for?" to justify the `run_async` method of Sampler.

kevinsung:
@maffoo 's comment suggests another answer to the question "What is batching for?" which is that `run_batch` is actually just a natural operation to perform in many situations. It can be used simply to make code nicer. Note that this suggests we should add it to the Sampler interface *even if it is never supported by any hardware*.

cirq/work/sampler.py
balopat: Thanks for your comments and providing the TFQ/simulator perspective @zaqqwerty and sorry for the confusion about repetitions.

balopat:
Thank you for all the comments - I'm _almost_ convinced to introduce this. The final question is still around the seeming difference in the backends: the repetitions. Is it a single number or one for each circuit? Does TFQ batching benefit from having a single number or is it actually a limitation? I would go with the more generic case in the interface and then the user can discover that in order to get any speed advantage on different platforms they'll have to parametrize batching accordingly, which in QE case will be the same repetition for each circuit.

kevinsung:
>I would go with the more generic case in the interface and then the user can discover that in order to get any speed advantage on different platforms they'll have to parametrize batching accordingly, which in QE case will be the same repetition for each circuit.

I agree.

zaqqwerty:
I also agree about going with the more generic case in the interface, since for the TFQ use case at least it will be most useful if the repetitions are able to be specified per-circuit in the batch.

balopat (changes requested):
Updated review: based on the discussion let's make this work in the Sampler interface, but let's change the repetitions to be per program. We should also add ample explanation in the docstring to encourage users to check the docs on the actual device / simulator they are using to figure out what parametrization will lead to actual speed up.

kevinsung:
Cool. It would be great if I can also address #3285 here. The proposal is that the user should not be forced to pass in a list of parameter resolvers; it should just work if no resolvers are passed.

balopat:
> I have updated this PR to allow the number of repetitions to be a list and documented that child classes may have requirements that must be met to obtain a speedup. I have also updated the behavior of the method so that it works if no sweeps are provided. However, I didn't modify the behavior of `Engine.run_batch`, so this does not fix #3285 .

Sounds good - I forgot to reply: let's do that in a separate PR to keep things clean / small.

balopat:
This looks good to me. @dstrain115 can you please have a look too at the final result?

balopat (approved):
LGTM

cirq/google/engine/engine_sampler.py
```diff
@@ -61,6 +61,34 @@ def run_sweep(
                                          gate_set=self._gate_set)
         return job.results()
 
+    def run_batch(
+            self,
+            programs: List['cirq.Circuit'],
+            params_list: Optional[List['cirq.Sweepable']] = None,
+            repetitions: Union[int, List[int]] = 1,
+    ) -> List['cirq.TrialResult']:
```
dstrain115, line 69: I am not sure if the sampler should flatten the results.  I think returning an List[List['cirq.TrialResult']] or maybe even an object would be better.

I realize that engine.results() only returns List['cirq.TrialResult'], but I was considering going back and adding a engine.batch_results() that returned an unflattened list.

cirq/google/engine/engine_sampler_test.py
```diff
@@ -50,6 +50,87 @@ def test_run_engine_program():
...
+    params1 = [cirq.ParamResolver({'t': 1})]
+    params2 = [cirq.ParamResolver({'t': 2})]
+    circuits = [circuit1, circuit2]
+    params_list = [params1, params2]
+    sampler.run_batch(circuits, params_list, 5)
+    engine.run_batch.assert_called_with(gate_set=cg.XMON,
+                                        params_list=params_list,
+                                        processor_ids=['tmp'],
+                                        programs=circuits,
+                                        repetitions=5)
```
dstrain115, line 70: Should these also test the return value?

cirq/work/sampler.py
```diff
@@ -199,3 +199,57 @@ async def run_sweep_async(
             An awaitable TrialResult.
         """
         return self.run_sweep(program, params=params, repetitions=repetitions)
+
+    def run_batch(
+            self,
+            programs: List['cirq.Circuit'],
+            params_list: Optional[List['cirq.Sweepable']] = None,
+            repetitions: Union[int, List[int]] = 1,
+    ) -> List['cirq.TrialResult']:
```
dstrain115, line 208: Same comment here.  I think flattening the list is probably a bad idea for the main interface.

cirq/google/engine/engine_sampler.py
kevinsung: I agree that a list of lists is better, but yes, here I was just following the structure of job.results(). I'll change the Sampler interface to return a list of lists. For now, I could just un-flatten the list here, and when you implement batch_results we can change it to use that. Does that sound good? If you'll implement that soon I could also wait.

cirq/google/engine/engine_sampler_test.py
kevinsung: The return value is a mock, so there's nothing to test (right?).

cirq/work/sampler.py
kevinsung: I agree; I'll change it to a list of lists.

cirq/google/engine/engine_sampler_test.py
dstrain115: Yeah, I guess it could test the pass-through behavior, but I guess there's not much point.  Agreed.

cirq/google/engine/engine_sampler.py
dstrain115: I put out #3303 for this.  What do you think?  Can we use it here?

kevinsung: Done.

cirq/work/sampler.py
kevinsung: Done.

kevinsung:
@dstrain115 I've addressed your comments; could you take another look?

cirq/google/engine/engine_sampler.py
```diff
@@ -61,6 +61,34 @@ def run_sweep(
...
+    ) -> List[List['cirq.TrialResult']]:
+        """Runs the supplied circuits.
+
+        In order to gain a speedup from using this method instead of other run
+        methods, the number of circuit repetitions must be the same for all
+        circuits. That is, the `repetitions` argument must be an integer, or
+        else a list with identical values.
+        """
+        if isinstance(repetitions, List) and len(programs) != len(repetitions):
+            raise ValueError('Number of circuits and repetitions must match')
```
dstrain115, line 78: Maybe put "Number of circuits and repetitions specified must match." so that people don't get confused between the number of reps in each circuit and the number of elements in the repetition list?  I am not sure how to make that more clear.

cirq/work/sampler.py
```diff
@@ -199,3 +199,57 @@ async def run_sweep_async(
...
+            programs: List['cirq.Circuit'],
+            params_list: Optional[List['cirq.Sweepable']] = None,
+            repetitions: Union[int, List[int]] = 1,
+    ) -> List[List['cirq.TrialResult']]:
+        """Runs the supplied circuits.
+
+        Each circuit provided in `programs` will pair with the optional
+        associated parameter sweep provided in the `params_list`, and be run
+        with the associated repetitions provided in `repetitions` (if
+        `repetitions` is an integer then all runs will have that number of
```
dstrain115, line 214: nit: I think there should be a comma after integer.

dstrain115 (approved):
LGTM with a few small nits.

cirq/google/engine/engine_sampler.py
```diff
@@ -61,6 +61,34 @@ def run_sweep(
...
+            programs: List['cirq.Circuit'],
+            params_list: Optional[List['cirq.Sweepable']] = None,
+            repetitions: Union[int, List[int]] = 1,
+    ) -> List[List['cirq.TrialResult']]:
+        """Runs the supplied circuits.
+
+        In order to gain a speedup from using this method instead of other run
+        methods, the number of circuit repetitions must be the same for all
+        circuits. That is, the `repetitions` argument must be an integer, or
+        else a list with identical values.
```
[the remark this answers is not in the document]
kevinsung: Done.

cirq/google/engine/engine_sampler.py
kevinsung: I've made the messages more clear.

cirq/work/sampler.py
kevinsung: Done.

PR #3268, opened 2020-08-26
Update book/index

lamberta:
Preview in cl/328544931

Updates from cl/327133340

### DIFF
--- a/docs/_index.yaml
+++ b/docs/_index.yaml
@@ -12,12 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This file desribes the main splash page for the Cirq sub-site on the 
-# larger Google Quantum site. 
-# It is used by "devsite", a Google internal CMS platform for documentation. 
+# This file desribes the main splash page for the Cirq sub-site on the
+# larger Google Quantum site.
+# It is used by "devsite", a Google internal CMS platform for documentation.
 
-book_path: /quark/cirq/_book.yaml
-project_path: /quark/cirq/_project.yaml
+book_path: /cirq/_book.yaml
+project_path: /cirq/_project.yaml
 description: >
   Cirq is a Python library for writing, manipulating, and optimizing quantum
   circuits and running them against quantum computers and simulators.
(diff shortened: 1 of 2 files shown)


docs/_index.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 The Cirq Developers. All Rights Reserved.
+# Copyright 2020 Google LLC. All Rights Reserved.
```
viathor, line 1: I think we should say "The Cirq Developers" as we do throughout cirq?

lamberta: Thanks for approval. I removed this license header since it's not necessary (should default to project license).
I'll check with the open source folks to clarify but I believe Google LLC is what they want.

viathor: SGTM. Note that "Google LLC" is in the [AUTHORS file](https://github.com/quantumlib/Cirq/blob/master/AUTHORS) at the root of the repo.

docs/_index.yaml
```diff
@@ -1,23 +1,5 @@
...
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
-# This file desribes the main splash page for the Cirq sub-site on the 
-# larger Google Quantum site. 
-# It is used by "devsite", a Google internal CMS platform for documentation. 
-
```
balopat, line 18: Why does the license have to go? 
How about the "usage" comment - does that have to go too? 
Does devsite have issues with yaml comments?

docs/_index.yaml
balopat: @lamberta please see go/opensource/releasing/authors.md for guidance - we chose the AUTHORS file method, which is kind of means The Cirq Authors / The Cirq Developers => list of authors. This allows for future extension of the authors without the need to update all the license headers.

docs/_index.yaml
lamberta: I'll fix and keep the license as-is. Your link OSS makes sense, thanks.
Re-adding usage section. That was accidentally removed when I copied over an earlier file.

balopat: Ah great! Thank you!

PR #3269, opened 2020-08-26
Serialization for gate tokens

dstrain115:
- Allows the ability to add a token to a gate.
- This will add the ability to specify a specific tuned version of a
gate identified by a string token.

Now adjusted so that gates without tokens are unaffected.


dstrain115:
> One concern I have with this approach is that we will be repeating the same token string (which could be longish like a UUID) on every op that uses it. What if instead of adding the token as an arg on every op, we were to add a list of tokens in, say, the Program message, and then a dedicated proto field on the ops to store the index in that list of the calibration token for the op, if any. Our serialization is already quite inefficient for large circuits and I worry about bloating it even more.

Hmm... this would save some space I guess.  It would cause some changes to the Circuit proto (and to the the serialization code) though that would be specific to this particular feature, and there would still be a field on every op (though I guess it would be an int instead of a long string).

Another idea would be to discard empty values of args, since that would be beneficial to the default case of no focused calibrations.

I wonder if I should look into making the circuits more efficient.

dstrain115:
I have now adjusted this PR so that gates without tokens are unaffected.  The only bloating to the serialization happens when using this new feature that no one has yet used.

As captured offline, I will look at streamlining the proto serialization in a separate task after this (likely next quarter though).

maffoo (commented):
Could we just add a `token` string field on the `Operation` proto? This has the desired semantics of not taking any space on the wire if not used, but also avoids repeating the string "token" if we _do_ add calibration tokens (the token strings themselves will still be repeated, but we're not going to solve that here). Seems simpler since we add the token facility in one place instead of having to add new arg serializers on every gate serializer.

dstrain115:
Closing this in favor of the approach in #3480

Issue #3263, opened 2020-08-23
cirq.contrib.quimb fails with quimb 1.2.0


dstrain115:
**Description of the issue**

Quimb fails with version 1.2.0.  We should either require 1.3 or (ideally) fix whatever was broken in 1.2

**How to reproduce the issue**
```
pip install quimb==1.2.0
pytest cirq/contrib/quimb/
```

<details>
test_tensor_expectation_value
...
 AttributeError: 'numpy.complex128' object has no attribute 'add_owner'
</details>

**Cirq version**
cirq from HEAD and quimb == 1.2.0

mpharrigan:
It fails in `rank_simplify`. I think the fix is this line: https://github.com/jcmgray/quimb/commit/492201519157355b751d64af73e297a318f0adff#diff-382c633ca92381427ca0bc74c1ac3081R3361

One option is to comment out `state_vector.py:150` which calls `rank_simplify`. It's possible EITHER that (1) this will cause performance to deteriorate in a major way for gridlike circuits because lots of things can be fused ahead of any true contraction OR (2) `contract` already finds a path that does this so it's fine.

Instead of truly investigating this, I say we set the requirements to `>=1.3.0` and if someone is desperate to use `1.2.0` they can patch out that specific line. What do you think?

dstrain115:
So, it looks like this doesn't quite work in our favorite environment, since it looks like the version number is not super well specified.  Maybe it got pulled from master or something:

This is the error:
"invalid literal for int() with base 10: '0+unknown'"

I think it is fine for it to assume unknown versions are not compatible, but it should not error out.

PR #3272, opened 2020-08-26
Minimum required version of quimb.

mpharrigan:
Alternatively, you can comment out `state_vector.py` line 150.

Fixes gh-3263

### DIFF
--- a/cirq/contrib/contrib-requirements.txt
+++ b/cirq/contrib/contrib-requirements.txt
@@ -4,7 +4,7 @@ ply>=3.4
 pylatex~=1.3.0
 
 # quimb
-quimb>=1.3.0
+quimb
 opt_einsum
 autoray
 
--- a/cirq/contrib/contrib-requirements.txt
+++ b/cirq/contrib/contrib-requirements.txt
@@ -4,7 +4,7 @@ ply>=3.4
 pylatex~=1.3.0
 
 # quimb
-quimb
+quimb>=1.3.0
 opt_einsum
 autoray
 
--- a/cirq/contrib/quimb/state_vector.py
+++ b/cirq/contrib/quimb/state_vector.py
@@ -151,6 +151,7 @@ def tensor_expectation_value(circuit: cirq.Circuit,
     ]
     tn = qtn.TensorNetwork(tensors + end_bras)
     if QUIMB_VERSION < (1, 3):
+        # coverage: ignore
         warnings.warn('Please use quimb>=1.3 for optimal performance in '
                       '`tensor_expectation_value`. '
                       'See https://github.com/quantumlib/Cirq/issues/3263')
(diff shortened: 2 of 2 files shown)


dstrain115:
Ideally, I would like to support quimb 1.2 if possible, since a certain repo ending in "3" has that version.  (Either that, or we need to upgrade 1.3)

balopat:
It feels odd to have a constraint on our contrib by g3 and vica versa. Can we make our optional package optional in g3 as well?

dstrain115:
> It feels odd to have a constraint on our contrib by g3 and vica versa. Can we make our optional package optional in g3 as well?

Yeah, it's not a hard constraint.  I don't know if there are optional requirements in g3 (the tests fail for quimb 1.2, so it would be tough to detect correctly).  There are work-arounds, but stuff like this always sets off warning bells for me.  The more restrictive our requirements, the harder it will be to install.  If there is a legitimate requirement that we need, that's fine.  However, if it is just because we are lazy, then we are pushing the work off on someone else (me and everyone else who installs this package on an older system).  It's not just Google either.  Many bigger systems, like national labs, for instance, have standardized environments where they try to keep a set of packages that will work for scientific software.

For quimb (in contrib) though, it's not really that big of a deal, so if the requirement is 1.3, I guess I'll deal with it.  It looks like quimb 1.2 is already over a year old, so 1.3 is probably not too unreasonable.

Since we know which line causes the issue, can we just do a quick compatibility check on that line and skip it (maybe with a warning) if the version is old?  That would be more user friendly than forcing an upgrade to the newest version.

mpharrigan:
Yeah, I could add a conditional skip of that line. Maybe we could solicit a community contribution to do a little investigation to figure out if that line is necessary or not.

balopat:
Great points @dstrain115, this is an ever green tension - users have a wide range of versions of our dependencies, however the wider version range we support the larger the cost to maintain it. 

We should definitely do our due diligence respectfully and not take it lightly (we shouldn't be lazy about it) to limit our versions. 

We _know_ that it works on quimb 1.3.0 and quimb provides advantage over the DensityMatrixSimulator in some cases, which was the primary reason to introduce this feature. 

The complete investigation for this particular failure is going to take some time, as one would have to test whether there are cases where removing this line (and using quimb 1.2.0, which might bring other degradations!) is going to cause performance degradation and whether that nullifies the advantage or not. The functionality will still work, the performance is not guaranteed, only on 1.3.0. 

I don't think I want to burn engineer work on exploring the performance guarantees of a year old system. 

Thus, my recommendation: 
- let's try to keep quimb unrestricted, 
- add a conditional skip on the line AND print a warning that performance is not guaranteed, things might be slower than expected, we recommend upgrading to the latest quimb version 

How does that sound?

mpharrigan:
done

PR #3275, opened 2020-08-27
docstrings for type aliases in devsite API docs

balopat:
The devsite API docs generation script doesn't pick up docstrings for our type aliases currently. This fixes it. 
Inspired by https://github.com/tensorflow/tensorflow/blob/40a314cab8d6c7afbc62456433a033c1cab8d74f/tensorflow/python/types/doc_typealias.py#L24:5

### DIFF
--- a/cirq/_doc.py
+++ b/cirq/_doc.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Workaround for associating docstrings with public constants."""
+import sys
+
 from typing import Any, Dict, NamedTuple, Optional
 
 DocProperties = NamedTuple(
@@ -44,4 +46,20 @@ def document(value: Any, doc_string: Optional[str] = None):
     """
     docs = DocProperties(doc_string=doc_string)
     RECORDED_CONST_DOCS[id(value)] = docs
+
+    ## this is how the devsite API generator picks up
+    ## docstrings for type aliases
+    if sys.version_info >= (3, 7):
+        try:
+            value.__doc__ = doc_string
+        except AttributeError:
+            # we have a couple (~ 7) global constants of type list, tuple, dict,
+            # that fail here as their __doc__ attribute is read-only.
+            # For the time being these are not going to be part of the generated
+            # API docs. See b/166663013 for more info.
+
+            # to list them, uncomment these lines and run `import cirq`:
+            # print(f"WARNING: {e}")
+            # print(traceback.format_stack(limit=2)[0])
+            pass
     return value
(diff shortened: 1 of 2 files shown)


cirq/google/gate_sets.py
```diff
@@ -111,7 +111,7 @@
 }
 
 document(NAMED_GATESETS,
-         """"A convenience mapping from gateset names to gatesets""")
```
mpharrigan, line 114: how did this ever work?

cirq/_doc.py
```diff
@@ -44,4 +46,20 @@ def document(value: Any, doc_string: Optional[str] = None):
...
+    ## this is how the devsite API generator picks up
+    ## docstrings for type aliases
+    if sys.version_info >= (3, 7):
+        try:
+            value.__doc__ = doc_string
+        except AttributeError:
+            # we have a couple (~ 7) global constants of type list, tuple, dict,
+            # that fail here as their __doc__ attribute is read-only.
+            # For the time being these are not going to be part of the generated
+            # API docs. See b/166663013 for more info.
```
dabacon, line 59: a bit strange to have b/XX in external repo.  maybe make a cirq issue and the include this reference there?

cirq/_doc.py
```diff
@@ -44,4 +46,20 @@ def document(value: Any, doc_string: Optional[str] = None):
     """
     docs = DocProperties(doc_string=doc_string)
     RECORDED_CONST_DOCS[id(value)] = docs
+
+    ## this is how the devsite API generator picks up
+    ## docstrings for type aliases
+    if sys.version_info >= (3, 7):
```
dabacon, line 52: why the version check?

mpharrigan, line 52: Is this a true restriction? After some mild googling (and https://docs.python.org/3.6/reference/datamodel.html) it looks like it was fine in 3.6. If you're catching attribute errors anyways, why the need for a version check?

cirq/google/gate_sets.py
dabacon: It just adds the quote to the front, so the doc string was
```"A convenience mapping from gateset names to gatesets```

mpharrigan: ah. I was way overthinking it

cirq/_doc.py
balopat: The difference is `Union`, `Any`, `List` and others will fail under 3.7. 
🤦  you are right about the error catching though :D

cirq/_doc.py
balopat: Fixing!

cirq/_doc.py
balopat: Copy paste for the TF codebase - before 3.7 type aliases wouldn't work with this. But as @mpharrigan pointed out actually using this _and_ the exception handling is superfluous.

PR #3280, opened 2020-08-28
Adding default values to op serializers and deserializers

dstrain115:
- This will allow us to have more efficient passing of values over
the wire for common or missing values.
- This will also allow tokens for focused calibrations (see PR #3269)
to not add a token arg for every gate.
- This will also add defaults for deserialization only.

Note: a second PR will later be needed to add the defaults for
serializers.  This is done in two steps for backwards compatibility.
We need the ability to correct for defaults in deserialization on
all servers before we can add support for defaults in serialization.


balopat (approved):
LGTM with a nit

balopat (changes requested):
I would hold this back after until the release of v0.9.0 just so that we don't have the asymmetry between the deserializer and serializer in the released version.

dstrain115:
> I would hold this back after until the release of v0.9.0 just so that we don't have the asymmetry between the deserializer and serializer in the released version.

I disagree.  For backwards compatibility, we need the deserialization broadly available before we start serializing into this form.  Otherwise, we break all v0.9 clients when we upgrade the server to use the new serialization.

balopat:
> > I would hold this back after until the release of v0.9.0 just so that we don't have the asymmetry between the deserializer and serializer in the released version.
> 
> I disagree. For backwards compatibility, we need the deserialization broadly available before we start serializing into this form. Otherwise, we break all v0.9 clients when we upgrade the server to use the new serialization.

Sounds good - I guess we would be okay with the asymmetry in cirq-unstable for a couple of weeks as well. Thanks for chiming in. Merging.

balopat (approved):

PR #3282, opened 2020-08-30
Add Moment.with_operations and use in Circuit.batch_insert_into

maffoo:
Previously batch_insert_into could only take one Operation per
insertion, unlike batch_insert. With this change, batch_insert_into can
also take an OP_TREE for each insertion, but still ensures that all the
insertions go into existing moments instead of creating new ones (that
is, the inserted OP_TREE cannot have multiple ops on a given qubit).


cirq/ops/moment.py
```diff
@@ -116,6 +116,32 @@ def with_operation(self, operation: 'cirq.Operation') -> 'cirq.Moment':
 
         return m
 
+    def with_operations(self, *contents: 'cirq.OP_TREE') -> 'cirq.Moment':
```
Strilanc, line 119: This is equivalent to what the `__add__` operation does.

It's probably reasonable to have an autocomplete-discoverable name for it. But mention the `__add__` approach as an alternative, and delegate it instead of duplicating code.

cirq/circuits/circuit.py
```diff
@@ -1411,8 +1411,8 @@ def batch_insert_into(self,
             Inserted into a moment index that doesn't exist.
         """
         copy = self.copy()
-        for i, op in insert_intos:
-            copy._moments[i] = copy._moments[i].with_operation(op)
+        for i, insertions in insert_intos:
+            copy._moments[i] = copy._moments[i].with_operations(insertions)
         self._device.validate_circuit(copy)
         self._validate_op_tree_qids(copy)
```
Strilanc, line 1417: Add unit tests exercising the new functionality.

Strilanc (dismissed):

cirq/ops/moment.py
maffoo: Done. I kept the code in `with_operations` and made `__add__` delegate to this since it doesn't re-index the existing operations and so will be more efficient in the case of adding a small additional number of operations into a large moment. Also kept the special behavior of `__add__` of delegating to `Circuit.__radd__` when adding to a circuit, which we don't want for `with_operations`.

cirq/circuits/circuit.py
maffoo: Done.

dstrain115 (approved):

balopat (approved):
LGTM

PR #3284, opened 2020-08-31
fix hanging docs build + bad cleanup logic

balopat:


balopat:
You're right I should have updated it - it would have been nicer to have a more substantial commit message - I just haven't read your comment in time, and couldn't stop Cirqbot merging (we should have an undo label?:D) I added the description at least it will be in the PR description historically - I don't want to force update to master to amend the commit msg.

PR #3287, opened 2020-09-02
Add more information about gates and best practices

dstrain115:
- Added information about supported gates and their unitaries
- Organized best practices and added more information
  -- organized into running at all, improving error, and rep rate
  -- Added info about delaying initialization
  -- Added more details about aligning gates.


docs/google/devices.md
```diff
@@ -24,8 +24,179 @@ Several factors can influence this limit, but this can be estimated at about 40
...
+applying a physical Z gate that performs a hardware operation
+to affect the frequency of the qubit.
+
+This can be done by applying a PhysicalZTag to the Z gate,
+such as in the following example:
+
+```
+cirq.Z(cirq.GridQubit(5, 5)).with_tags(cirq.google.PhysicalZTag())
+```
+
```
dabacon, line 107: Are these implemented using two microwave pulses?

docs/google/best_practices.md
```diff
@@ -58,7 +74,69 @@ using a error rate of 0.5% per gate, a circuit of depth 20 and width 20 could be
...
+
+See the function `cirq.stratified_circuit` for an automated way to pack gates
+into moments with similar gates.
+
+## Running circuits faster
+
+The following sections give tips and tricks that allow you to improve your
+repetition rate (how many repetitions per second the device will run).
+
+These tips will not improve your results or your fidelity but will help you
```
dabacon, line 129: This could improve fidelity if you are doing your own characterization, since drift can be smaller.

docs/google/devices.md
```diff
@@ -24,8 +24,179 @@ Several factors can influence this limit, but this can be estimated at about 40
...
+The inverse (`cirq.ISWAP ** -0.5`) is also available
+
+The unitary of this gate, which can also be found via the `cirq.unitary`
+function, is:
+
+$$
+\left[
+\begin{matrix}
+1 & 0 & 0 & 0 \\
+0 & 0.5sqrt{2} 0 & 0.5 i sqrt{2} & 0 \\
```
dabacon, line 156: does this latex render correctly?  Think you need `\sqrt`?

docs/google/devices.md
```diff
@@ -24,8 +24,179 @@ Several factors can influence this limit, but this can be estimated at about 40
...
+0 & 0.5 i sqrt{2} 0 & 0.5 sqrt{2} & 0 \\
```
dabacon, line 157: suggest making these `\frac{1}{\sqrt{2}}` or `\frac{i}{\sqrt{2}}`

docs/google/devices.md
dabacon, line 156: extra 0 in `0.5sqrt{2} 0`

docs/google/devices.md
```diff
@@ -24,8 +24,179 @@ Several factors can influence this limit, but this can be estimated at about 40
...
+
+This gate is equivalent to FSimGate(0, π)
+
+#### FSim gateset
+
+The `cirq.FSIM_GATESET` provides all three of the above gates in one set.
+In addition, by using this combined gate set, the FSimGate can be parameterized,
+which allows for efficient sweeps across varying two-qubit gates.
+Note that providing a theta/phi combination that
+is not one of the above gates, the circuit will not execute properly.
```
dabacon, line 180: Will it actually serialize?  "Execute properly" makes it sound like it will silently run, just not properly.

docs/google/devices.md
dstrain115: I'm not totally sure actually.  I think I recall that it was some sort of frequency detuning, but I don't remember exactly what the hardware people told me.  Anyway, we don't have implementation details for other gates, so I think it is okay not including this here.

docs/google/best_practices.md
dstrain115: That's true.  Changed.

docs/google/devices.md
dstrain115: Done

docs/google/devices.md
dstrain115: Done

docs/google/devices.md
dstrain115: Done

docs/google/devices.md
dstrain115: Wording changed.  If the theta/phi combination is explicitly wrong, it will fail to serialize.  If the invalid combination is hidden in a sweep, the error will occur server side.

docs/google/best_practices.md
```diff
@@ -4,7 +4,28 @@ This section lists some best practices for creating a circuit that performs well
 on Google hardware devices. This is an area of active research, so users are
 encouraged to try multiple approaches to improve results.
 
-## Use built-in optimizers as a first pass
+This guide is split into three parts:
+*  Getting your circuit to run
+*  Lowering error
+*  Making it run faster
```
mpharrigan, line 10: change the order of these? i.e. 1) get circuits to run 2) make them fast 3) lower error. The thought being: it's easier to try error reduction techniques if it's already faster :) 

otherwise: splitting this doc up into meaningful sections like these is a very good idea

docs/google/best_practices.md
```diff
@@ -4,7 +4,28 @@ This section lists some best practices for creating a circuit that performs well
...
+
+
+## Getting a circuit to run on hardware
+
+In order to run on hardware, the circuit must only use qubits and gates that the
+device supports.  Using inactive qubits, non-adjacent qubits, or non-native
+gates will immediately cause a circuit to fail.
+
+Attaching a `cirq.Device` to a circuit will test a lot of these conditions for
```
mpharrigan, line 19: Is this something we want to lean in to? I always considered the fact that circuits have a device to be a vestigial design choice that no one uses in practice and IMO doesn't make much sense. From my point of view, it would make sense for devices to validate circuits: `cirq.google.Sycamore.validate_circuit(circuit)`. You could even imagine a scheme where instead of just raising an exception at the first problem it could return a structured list of problems and suggestions on how to fix them.

docs/google/best_practices.md
```diff
@@ -58,7 +74,69 @@ using a error rate of 0.5% per gate, a circuit of depth 20 and width 20 could be
...
+Make sure that measurements are kept in the same moment as the final moment in
+the circuit.  Make sure that any circuit optimizers do not alter this by
+incorrectly pushing measurements forward. This behavior can be avoided by
+measuring all qubits with a single gate or by adding
+the measurement gate after all optimizers have run.
+
+Currently, only terminal measurements are supported by the hardware.  If you
+absolutely need intermediate measurements for your application, reach out to
+your Google sponsor to see if they can help devise a proper circuit using
+intermediate measurements.
```
mpharrigan, line 88: there's no checks against this and in fact you can get intermediate measurements to work (where work is defined as return bits without crashing)

docs/google/best_practices.md
```diff
@@ -201,3 +260,10 @@ print(list(flat_sweep.param_tuples()))
 #  (('<2**t - 1>', 0.681792830507429),),
 #  (('<2**t - 1>', 1.0),)]
 ```
+
+### Measure less qubits
```
mpharrigan, line 264: fewer

mpharrigan: also this might be subtly bad advice? I always have a sneaking suspicion that measuring spectator qubits could be useful. For example: what if you throw out any repetitions where spectator qubits are excited? I don't know if anyone's actually tried this though

docs/google/best_practices.md
dstrain115: Yeah, but I don't think we want people doing this unless they have help.  It will be difficult to get working correctly (where correctly is defined as getting return bits that are not random noise).

docs/google/best_practices.md
dstrain115: Yeah, I was on the fence for including this advice anyway.  I'll drop it.

docs/google/devices.md
```diff
@@ -24,8 +24,179 @@ Several factors can influence this limit, but this can be estimated at about 40
...
+The unitary of this gate, which can also be found via the `cirq.unitary`
+function, is:
+
+$$
+\left[
+\begin{matrix}
+1 & 0 & 0 & 0 \\
+0 & 0 & -i & 0 \\
+0 & -i & 0 & 0 \\
+0 & 0 & 0 & e^{-i \pi \over 6}
```
dabacon, line 135: slight pref for `e^{-i \frac{\pi}{6}}.  Puts the `-i` out front instead of the minus in the numerator, and also LaTeX people yell at me for using `\over`.

docs/google/best_practices.md
dstrain115: Is this something we are trying to get rid of?  I guess either way doesn't matter much to me.  I changed it to validate using the device.

docs/google/best_practices.md
dstrain115: Done.

docs/google/devices.md
dstrain115: Sure.  Done.

docs/google/best_practices.md
```diff
@@ -58,7 +74,69 @@ using a error rate of 0.5% per gate, a circuit of depth 20 and width 20 could be
...
+
+Devices generally operate in the Z basis, so that rotations around the Z axis will become
+book-keeping measures rather than physical operations on the device. The EjectZ optimizer
+included in optimizer lists for each device will generally compile these operations out
+of the circuit by pushing them back to the next non-commuting operator. If the resulting
+circuit still contains Z operations, they should be aggregated into their own moment,
+if possible.
+
+See the function `cirq.stratified_circuit` for an automated way to pack gates
+into moments with similar gates.
```
mrwojtek: As far as I understand the stratifier works to other way around - it unpacks moments when they contain mixed gates, it doesn't merge them together.

docs/google/devices.md
```diff
@@ -24,8 +24,179 @@ Several factors can influence this limit, but this can be estimated at about 40
...
+cirq.Circuit(
+  cirq.Moment(cirq.Z(cirq.GridQubit(4,4))),
+  cirq.Moment(cirq.X(cirq.GridQubit(4,5)))
+)
+```
+
+## Gates supported
+
+The following lists the gates supported by Google devices.
+Please note that gate durations are subject to change as hardware isxi
```
mrwojtek, line 69: typo ("isxi")

docs/google/devices.md
```diff
@@ -24,8 +24,179 @@ Several factors can influence this limit, but this can be estimated at about 40
...
+
+Google devices support arbitrary one-qubit gates of any rotation.
+The full complement of these rotations can be accessed by using the
+`cirq.PhasedXZGate`.  More restrictive one-qubit gates, such as
+the Pauli gates `cirq.X`, `cirq.Y`, `cirq.Z`, as well as the
+gate `cirq.PhasedXPowGate` can also be natively executed.
+
+#### Virtual Z gates
+
+Rotation around the Z axis is generally not a hardware operation.
```
mrwojtek, line 85: This is not strictly true. Virtual Z gates are realized on the hardware, they are just bundled together with two-qubit gates and microwaves. Their presence changes the signals applied to the gates.

docs/google/devices.md
```diff
@@ -24,8 +24,179 @@ Several factors can influence this limit, but this can be estimated at about 40
...
+which allows for efficient sweeps across varying two-qubit gates.
+Note that providing a theta/phi combination that
+is not one of the above gates will cause an error when run on hardware.
+
+
+### Wait gate
+
+For decay experiments and other applications, a WaitGate is provided
+that causes the device to idle for a specified amount of time.
+This can be accomplished by specifying a `cirq.WaitGate`.
```
mrwojtek, line 187: What is the behavior of wait gate when run with other gates? Will the moment last the maximum of the gates in that moment, including a wait gate?

docs/google/devices.md
```diff
@@ -24,8 +24,179 @@ Several factors can influence this limit, but this can be estimated at about 40
...
 ## Specific Device Layouts
 
+The following devices are provided as part of cirq and can help you get your
+circuit ready for running on hardware by verifying that you are using
+appropriate qubits.
+
+Note that real hardware does not always have all qubits enabled, and it
+is important to check the device specification for the processor that you
+will attempt to run on to make sure that the qubits your circuit uses
+are actually active.
```
mrwojtek, line 198: It would be good to mention here that some qubits might be disabled and re-enabled at different days (result of maintenance).

mrwojtek (commented):
Hi Doug, I put some comments related to our email conversation.

balopat (approved):

docs/google/best_practices.md
```diff
@@ -58,7 +74,69 @@ using a error rate of 0.5% per gate, a circuit of depth 20 and width 20 could be
...
+since the duration of a moment is its longest gate.  Making sure that each layer
+contains similar gates of the same duration can be challenging, but it will
+likely have a measurable impact on the fidelity of your circuit.
+
+Devices generally operate in the Z basis, so that rotations around the Z axis will become
+book-keeping measures rather than physical operations on the device. The EjectZ optimizer
+included in optimizer lists for each device will generally compile these operations out
+of the circuit by pushing them back to the next non-commuting operator. If the resulting
+circuit still contains Z operations, they should be aggregated into their own moment,
+if possible.
```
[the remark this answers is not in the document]
dstrain115: If you have extra moments because of the Z gates, it can lengthen the duration of the circuit.  I tried to clarify my comments.  See if you like the new wording better.

Like, if you have
```
---Z---X---
---X-------
```

docs/google/best_practices.md
dstrain115: I thought someone optimizer that aligned one and two qubit layers.  If this isn't the way to do it, I am not sure what to recommend.

docs/google/devices.md
dstrain115: Done.

docs/google/devices.md
dstrain115: Adjusted wording to say that it is not a hardware operation on its own.

docs/google/devices.md
dstrain115: Yes, I believe so.  I clarified that for the general case in the moment section above (that the duration of a moment is its longest gate) but did not specifically call it out in this section.

docs/google/devices.md
dstrain115: Done.

docs/google/best_practices.md
mpharrigan: opened https://github.com/quantumlib/Cirq/issues/3295 to discuss the broader question

PR #3291, opened 2020-09-02
Implement speckle purity benchmarking as part of grid XEB

kevinsung:
I followed the procedure outlined in Section VI.C.4 of https://arxiv.org/pdf/1910.11333.pdf.


cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -25,6 +25,52 @@
 
 CrossEntropyPair = NamedTuple('CrossEntropyPair', [('num_cycle', int),
                                                    ('xeb_fidelity', float)])
+SpecklePurityPair = NamedTuple('SpecklePurityPair', [('num_cycle', int),
+                                                     ('purity', float)])
+
+
+def purity_from_probabilities(
```
viathor, line 32: This function should live in separate file "purity_estimation.py" for consistency with "fidelity_estimation.py".

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -25,6 +25,52 @@
...
+        hilbert_space_dimension: int,
+        probabilities: Sequence,
```
viathor, line 34: `Sequence[float]`

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -25,6 +25,52 @@
...
+        hilbert_space_dimension: Dimension of the Hilbert space on which the
+            quantum circuits acts.
+        probabilities: Empirical probabilities of bitstrings observed in
+            experiment.
+    Returns:
+        Estimate of the purity of the state resulting from the experimental
+        implementation of a quantum circuit.
+    """
+    D = hilbert_space_dimension
+    return np.var(probabilities) * D**2 * (D + 1) / (D - 1)
```
viathor, line 73: Optional: I assume you're creating the local variable just to shorten the return statement. Here's an alternative that has an additional benefit of explaining to the reader what's happening:

```python
pt_var = (hilbert_space_dimension - 1) / (hilbert_space_dimension + 1) / hilbert_space_dimension**2
return np.var(probabilities) / pt_var
```

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -107,44 +182,87 @@ def plot(self, ax: Optional[plt.Axes] = None,
...
         and f is the cross entropy fidelity for that number of cycles,
         using nonlinear least squares.
 
         Returns:
-            A DepolarizingModel object, which has attributes `coefficient`
-            representing the value S, `decay_constant` representing the value
-            p, and `covariance` representing the covariance in the estimation
-            of S and p in that order.
+            A CrossEntropyDepolarizingModel object, which has attributes
+            `coefficient` representing the value S, `decay_constant`
```
viathor, line 191: The original text you modified appears outdated. The attributes are "spam_depolarization" and "cycle_depolarization", not "coefficient" and "decay_constant". Please update.

(Applies to Returns: section in the docstring for `purity_depolarizing_model` as well.)

viathor (changes requested):

cirq/experiments/cross_entropy_benchmarking.py
kevinsung: Done.

cirq/experiments/cross_entropy_benchmarking.py
kevinsung: Done. I was originally using this on a list of numpy arrays, but I've modified that code to convert to a flattened numpy array to match Sequence[float].

cirq/experiments/cross_entropy_benchmarking.py
kevinsung: Even with your suggestion we would have to split the formula for `pt_var` over multiple lines, which I think is much less readable. I've taken your suggestion to create a `pt_var` variable, but kept the variable `D`.

cirq/experiments/cross_entropy_benchmarking.py
kevinsung: Done.

cirq/experiments/grid_parallel_two_qubit_xeb.py
```diff
@@ -35,9 +35,11 @@
 from cirq import devices, ops, protocols, sim, value
 from cirq.experiments.cross_entropy_benchmarking import (CrossEntropyResult,
                                                          CrossEntropyResultDict,
-                                                         CrossEntropyPair)
+                                                         CrossEntropyPair,
+                                                         SpecklePurityPair)
 from cirq.experiments.fidelity_estimation import (
     least_squares_xeb_fidelity_from_probabilities)
+from cirq.experiments.purity_estimation import (purity_from_probabilities)
```
viathor, line 42: nit: no need for parens

viathor (approved):
Thank you! LGTM!

cirq/experiments/grid_parallel_two_qubit_xeb.py
kevinsung: Done.

PR #3294, opened 2020-09-02
removing readthedocs.io references

balopat:
We should not refer to readthedocs.io from our docs for content, unless it's absolutely necessary. 
This removes the last readthedocs.io references.

### DIFF
--- a/docs/tutorials/examples.md
+++ b/docs/tutorials/examples.md
@@ -100,22 +100,22 @@ Quantum circuit to prepare the BCS ground states for superconductors/superfluids
 
 ## Advanced Tutorials
 
-*    [Variational Algorithm](./variational_algorithm)
+*    [Variational Algorithm](./variational_algorithm.ipynb)
 Case study demonstrating construction of an ansatz for a two-dimensional Ising
 model and how to simulate and optimize it.
 
-*    [QAOA](./QAOA_Demo)
+*    [QAOA](./QAOA_Demo.ipynb)
 Demonstration of optimizing cost of a max-cut problem using quantum
 approximation optimization algorithm (QAOA)
 
-*    [Hidden Linear Function](./hidden_linear_function)
+*    [Hidden Linear Function](./hidden_linear_function.ipynb)
 Demonstration of a problem similar to Bernstein-Vazirani that uses a hidden
 function rather than using an Oracle.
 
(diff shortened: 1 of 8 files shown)


balopat:
Note that from md files we shouldn't use file extensions, but from notebooks we should - this inconsistency is weird, but that's how sphinx, recommonmark and nbsphinx work - I hope that this works with devsite as well, @lamberta?

lamberta:
> we shouldn't use file extensions, but from notebooks we should

Since we try to make the OSS GitHub browsing experience as pleasant as possible, include file extensions for everything (`.md` and `.ipynb`). Should work in Colab, as well. These extensions are stripped when imported into the site.

The exception is the `_book.yaml` nav file. No extension there since it's for the site only.

docs/tutorials/examples.md
```diff
@@ -100,22 +100,22 @@ Quantum circuit to prepare the BCS ground states for superconductors/superfluids
 
 ## Advanced Tutorials
 
-*    [Variational Algorithm](https://cirq.readthedocs.io/en/latest/tutorials/variational_algorithm.html)
+*    [Variational Algorithm](./variational_algorithm)
```
dstrain115, line 103: Do we need an extension for this (ipynb)?  Same with below.  Your comment suggests we should add it.

dstrain115 (approved):

docs/tutorials/examples.md
lamberta: Yes. You can test this in the [github preview](https://github.com/quantumlib/Cirq/blob/e59015ee42620ab9221a9b2dda9191d05e015d7d/docs/tutorials/examples.md). This link doesn't work in GitHub unless you add `variational_algorithm.ipynb`
It will work on the site since we strip extensions anyway, but, as a general style, using the extension means the link works everywhere.

balopat: Okay, thanks for the input - I agree, I'll change this back to .ipynb.
However, this will need a bit more work due to a bug in recommonmark - I'm running into https://github.com/readthedocs/recommonmark/issues/130. 

The current version of this page https://cirq.readthedocs.io/en/latest/docs/tutorials/examples.html is broken both on RTD and in devsite. I'm having promising results with [myst](https://myst-parser.readthedocs.io/en/latest/index.html)...

PR #3296, opened 2020-09-03
Add information about batching restrictions

dstrain115:


docs/google/engine.md
```diff
@@ -129,6 +129,14 @@ has a method `run_batch()` that functions similar to `run()` but accepts a
 list of circuits and parameter sweeps.  Each circuit must have a corresponding
 parameter sweep.  If the circuit does not use a sweep, pass in `None`.
 
+There are some restrictions on the circuits that can be batched together.
+All circuits in the same batch must measure the same set of qubits.  They
+must all contain the same number of repetitions.
+Batching circuits together that do not follow this restriction may not
```
dabacon, line 135: "may not" or "will not"

dabacon (approved):

dabacon:
You will still get speedup because of lack of latencies, right?

docs/google/engine.md
dstrain115: Well, I put "may not" since I am not totally sure how the batch sampler stuff will land.  If you specify different repetitions in the Sampler interface, will it raise an exception or silently let it slide?   Right now it is still TBD.

dstrain115:
> You will still get speedup because of lack of latencies, right?

Maybe a slight speed-up, but I think the main benefit would be if the programs can be batched together server-side.

docs/google/engine.md
balopat, line 135:
```suggestion
Batching circuits together that do not follow these restrictions may not
```

docs/google/engine.md
```diff
@@ -129,6 +129,14 @@ has a method `run_batch()` that functions similar to `run()` but accepts a
 list of circuits and parameter sweeps.  Each circuit must have a corresponding
 parameter sweep.  If the circuit does not use a sweep, pass in `None`.
 
+There are some restrictions on the circuits that can be batched together.
+All circuits in the same batch must measure the same set of qubits.  They
```
balopat, line 133: Use bullet points and spell out the rules, makes it more memorable: e.g.: 

* **Same qubits**: All circuits in the same batch must measure the same set of qubits.
* **Same repetitions**: All circuits in the same batch must have the same number of repetitions.

dstrain115: Good idea.  Done.

PR #3298, opened 2020-09-03
Add cirq.QubitPermutationGate

### LINKED ISSUE #2671: Permutation Gate
That permutes qubits. There's `cirq.interop.QuirkQubitPermutationGate` which does what I want but a) it's not part of the main set of cirq ops and b) it requires an "identifier" and "name" which are superfluous

balopat:
Adds cirq.QubitPermutationGate. This is a refactor of cirq.interop.quirk.QuirkPermutationGate. 
I also pulled out testing utilities for it from the quirk pacakge into `cirq.testing.assert_equivalent_computational_basis_map`.

Fixes #2671.


cirq/ops/permutation_gate.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+        Args:
+            permutation: A shuffled sequence of integers from 0 to
+                len(permutation) - 1. The entry at offset `i` is the result
+                of permuting `i`.
+        """
+        if not permutation:
+            raise ValueError(f"Invalid permutation (empty): {permutation}")
+
+        if len(Counter(permutation)) < len(permutation):
```
viathor, line 38: Built-in set should do:

```python
if len(set(permutation)) < len(permutation):
```

cirq/ops/permutation_gate.py
```diff
@@ -0,0 +1,85 @@
...
+            raise ValueError(f"Invalid permutation {permutation} "
+                             f"Each index must appear only once.")
+
+        invalid_indices = list(
+            filter(lambda x: x >= len(permutation) or x < 0, permutation))
+        if len(invalid_indices) > 0:
```
viathor, line 44: Use `any` and comprehension expression:

```python
if any(not 0 <= a < len(permutation) for a in permutation):
    ...
```

cirq/ops/permutation_gate.py
```diff
@@ -0,0 +1,85 @@
...
+            raise ValueError(f"All indices have to satisfy "
+                             f"0 < i < {len(permutation)-1}. "
```
viathor, line 46: Inequalities are incorrect, should be: `f"0 <= i < {len(permutation)}. "

cirq/interop/quirk/cells/qubit_permutation_cells.py
```diff
@@ -39,31 +39,11 @@ def __init__(self, identifier: str, name: str, permutation: Sequence[int]):
...
-            ai = args.axes[i]
-            aj = args.axes[j]
-            assert args.target_tensor.shape[ai] == args.target_tensor.shape[aj]
-            permuted_axes[aj] = ai
-
-        # Delegate to numpy to do the permuted copy.
-        args.available_buffer[...] = args.target_tensor.transpose(permuted_axes)
-        return args.available_buffer
-
     def _circuit_diagram_info_(self, args: 'cirq.CircuitDiagramInfoArgs'
```
viathor, line 47: Shouldn't this be removed, too?

cirq/testing/equivalent_basis_map.py
```diff
@@ -0,0 +1,75 @@
...
+from cirq import circuits
+
+
+def assert_equivalent_computational_basis_map(maps: Dict[int, int],
+                                              circuit: circuits.Circuit):
+    """Ensure equivalence of basis state mapping.
+
+    Args:
+        maps: dictionary of test computational basis input states and
+            the output computational basis state that they should be mapped to.
```
viathor, line 28: nit: state -> states (plural)

cirq/ops/permutation_gate.py
```diff
@@ -0,0 +1,85 @@
...
+
+from cirq import protocols, value
+from cirq.ops import raw_types
+
+if TYPE_CHECKING:
+    import cirq
+
+
+@value.value_equality
+class QubitPermutationGate(raw_types.Gate):
```
viathor, line 25: Out of scope for this PR: I initially thought this would be a gate that permutes computational basis. Instead, it permutes qubits which is a less general operation. I think we should also implement the former, but not necessarily in this PR. I guess the reason to have both is that QubitPermutationGate has an efficient `_apply_unitary_ `.

viathor (changes requested):

cirq/ops/permutation_gate.py
balopat: ack

cirq/ops/permutation_gate.py
balopat: I did want to put the invalid indices on the error message though

cirq/ops/permutation_gate.py
balopat: ack

cirq/interop/quirk/cells/qubit_permutation_cells.py
balopat: No that is still valid for the quirk version

cirq/testing/equivalent_basis_map.py
balopat: ack

cirq/ops/permutation_gate.py
balopat: I agree. I was just following almost mechanically @mpharrigan's requirements from #2671 - thus this one is the same as the quirk gate minus the name + id. Having a permutation gate on the computational basis states might be 1) valuable for use cases (although I'm not aware of those just yet) 2) an even better option to consolidate the permutation gate hierarchy in the contrib.aqcuaintance package.

cirq/ops/permutation_gate.py
viathor: Good idea. You should use list comprehension though:

```python
invalid_indices = [x for x in permutation if not 0 <= x < len(permutation)]
```

It's simpler and [considered more idiomatic](https://google.github.io/styleguide/pyguide.html#s2.15-deprecated-language-features) in python that filter, map et al.

viathor (approved):
LGTM with one comment. Feel free to go ahead and merge once addressed.

cirq/interop/quirk/cells/qubit_permutation_cells.py
viathor: Oops, missed the gate name.

cirq/ops/permutation_gate.py
viathor: Regarding use-cases of the more general permutation gate: unitary corresponding to any bijective function from n-bitstrings to n-bitstrings (including e.g. arithmetic operations) is a permutation matrix. This may not be sufficient grounds for a generic permutation gate since it's reasonable to implement specific gates for such operations instead.

We can defer until we see more use-cases.

PR #3303, opened 2020-09-09
Add batched_results() to retrieve BatchResult

dstrain115:
- results() returns a List[TrialResult]
- added batched_result() to return the data as a List[List[TrialResult]]


cirq/google/engine/engine_job.py
```diff
@@ -296,7 +309,10 @@ def results(self) -> List[study.TrialResult]:
             elif result.Is(v2.batch_pb2.BatchResult.DESCRIPTOR):
                 v2_parsed_result = v2.batch_pb2.BatchResult.FromString(
                     result.value)
-                self._results = self._get_batch_results_v2(v2_parsed_result)
+                self.batch_mode = True
```
kevinsung, line 312: This should already be set, right? Since this wasn't previously set, maybe don't do it here.

kevinsung (approved):
LGTM, just one unimportant comment.

cirq/google/engine/engine_job.py
dstrain115: Agreed.  Removed.

PR #3306, opened 2020-09-10
Fix a bug in validating computational basis limits

smitsanghavi:
Could potentially cause out of bounds access for `one_hot`


viathor (approved):
Thanks!

balopat (approved):

PR #3307, opened 2020-09-10
Adjust docs for Getting started guide

dstrain115:
- Minor issues found through walk-through
- Google colabs are meant to be downloaded,
   so image links need to be absolute
- Rearrange text so that project creation/selection
   happens before trying to enable the API
- Add link to reservation colab on getting started guide.

### DIFF
--- a/docs/tutorials/google/colab.ipynb
+++ b/docs/tutorials/google/colab.ipynb
@@ -35,7 +35,7 @@
         "You can retrieve ipython notebooks in the cirq repository by\n",
         "going to the [doc directory](https://github.com/quantumlib/Cirq/tree/master/docs).  For instance, this colab template can be found [here](https://github.com/quantumlib/Cirq/blob/master/docs/tutorials/google/colab.ipynb).  Select the file that you would like to download and then click the \"Raw\" button in the upper right part of the window:\n",
         "\n",
-        "![Raw button](https://cirq.readthedocs.io/en/latest/_images/colab_github.png)\n",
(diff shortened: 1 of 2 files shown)


docs/tutorials/google/colab.ipynb
```diff
@@ -35,7 +35,7 @@
         "You can retrieve ipython notebooks in the cirq repository by\n",
         "going to the [doc directory](https://github.com/quantumlib/Cirq/tree/master/docs).  For instance, this colab template can be found [here](https://github.com/quantumlib/Cirq/blob/master/docs/tutorials/google/colab.ipynb).  Select the file that you would like to download and then click the \"Raw\" button in the upper right part of the window:\n",
         "\n",
-        "![Raw button](../../images/colab_github.png)\n",
+        "![Raw button](https://cirq.readthedocs.io/en/latest/_images/colab_github.png)\n",
```
balopat, line 38: Why are we referring to the RTD URL when this will end up on our devsite engine? 
Are you running this locally? Or in Colab? Where does the relative image not work? 
if we absolutely need absolute links (pun intended) we could link to the raw image on github on master (a particular commit would be better) - but referring to the Readthedocs URL is definitely odd.

dstrain115: The relative links won't work if you download the colab and then run it through colab, which you may want to do for the getting started guide and the colab template.  I changed it to github absolute URLs.

balopat (approved):
LGTM.

I'm not happy with this image URL workflow yet, opened an issue to tackle it #3308.

PR #3309, opened 2020-09-11
Improve initial_state handling for CliffordTableau and StabilizerStateChForm

smitsanghavi:
#2423 

- Use standard `big_endian_int_to_digits` method instead of local `bits` logic
- Get bounds validation for the computational basis state for free
- Improve comments in `CliffordTableau`, `StabilizerStateChForm` and `CliffordSimulator`
- Remove `np.ndarray` as a valid type for `StabilizerStateChForm` initial_state since arbitrary state vectors are not really supported.
- Add some tests


balopat (approved):
Nice, it's much more understandable this way. Thanks! 
One tiny nit and then feel free to merge it.

PR #3312, opened 2020-09-11
Update gates.ipynb

shaswata56:
there'll be `exponent=t` instead of `exponent=y`


shaswata56:
@googlebot I signed it!

balopat (approved):
Good catch, thank you!

shaswata56:
> Good catch, thank you!

I was just reading. Recently switching from Q# to Cirq. :sweat_smile:

PR #3313, opened 2020-09-11
Add field for reference strings

dstrain115:
- Draft proposal for adding a field for strings that get repeated often.
- This could save space in the serialized proto by allowing strings that
are repeated often (Such as tokens for focused calibration) to avoid
bloating the size of the proto.


dstrain115:
@maffoo Is this what you were looking for in the review of #3269?  I want to get consensus on the proto before I build out the implementation.

cirq/google/api/v2/program.proto
```diff
@@ -133,10 +137,15 @@ message Arg {
...
   // how to combine numbers and symbols mathematically.
+  //
+  // Reference indices are used to avoid repeating long strings in the
+  // serialized proto.  This is a number that refers to the index of
+  // the reference string in the program.
   oneof arg {
     ArgValue arg_value = 1;
     string symbol = 2;
     ArgFunction func = 3;
+    int32 reference_index = 4;
```
maffoo, line 148: Might want to call the underlying table `symbols` and the field here `symbol_index`; that makes clear that this is an alternative to specifying `symbol` as a string directly. It's actually kind of nice to have a single top-level "symbol table" where we can look to see all symbols that need to be resolved for a given program.

maffoo (commented):
Thanks for proposing this, @dstrain115. This is definitely along the lines of what I was looking for in #3269. One minor annoyance is that with a field `int32 reference_index` there's no way to distinguish between the field being set to 0 and not being set, so we need a way to handle that (always put an unused empty string as the first element of `reference_strings`, for example).

It's always worth considering whether we can address the same issue a bit more generally, for example have tables of `gates` and `qubits` so that in `Operation` we can refer to gates and qubits by index instead of as submessages with repeated strings. There are also the strings keys in the `args` map on `Operation` which will be repeated many times in a program and will similarly be repeated (default values for args could help somewhat, but it'd be nice if there were a way to avoid repeating arg names even when values themselves are set to non-default values). One possibility would be for the `gates` table to contain gate "prototypes" that specify the gate name and a list of arg names, then in the operation you could refer to the gate prototype by index and within that the args by index in the prototype's list of arg names.

dstrain115:
> int32 reference_index

Note that the default value of this would generally not be set.  Since this is an "oneof" field, then you can query which field was set.  If the field is supposed to be a blank string, you could set arg_value.string_value to be '' for instance.

I don't know if we should worry about setting whole gates to symbols.  I think this will start to make serialization quite tricky and possibly cause a loss of time in exchange for the decrease of size that you will get, since you will need to keep track of which gates that you have seen before (or make educated guesses?)  In any case, I will make it into a "Constant" class that be looked up on.  Look for a future PR where I combine this with serialization of tokens.

dstrain115:
Closing this in favor of the approach in #3480.

PR #3325, opened 2020-09-15
Make PauliString's key type generic

Strilanc:
- Currently the key type is required to derive from cirq.Qid
- This is a stepping stone towards allowing non-qid key types


cirq/ops/pauli_string.py
```diff
@@ -163,27 +164,52 @@ def _value_equality_values_cls_(self):
...
-    def __mul__(self, other) -> 'PauliString':
+    @overload
+    def __mul__(  # type: ignore
+            self, other: 'cirq.PauliString[TKey2]'
+    ) -> 'cirq.PauliString[Union[TKey, TKey2]]':
+        pass
+
+    @overload
+    def __mul__(self, other: Mapping[TKey2, 'cirq.PAULI_GATE_LIKE']
+               ) -> 'cirq.PauliString[Union[TKey, TKey2]]':
```
balopat, line 194: this Union could be pulled out as a type alias for "AllKeyTypes" or similar?

cirq/ops/pauli_string.py
```diff
@@ -41,10 +41,14 @@
 if TYPE_CHECKING:
     import cirq
 
+TDefault = TypeVar('TDefault')
+TKey = TypeVar('TKey', bound=raw_types.Qid)
+TKey2 = TypeVar('TKey2', bound=raw_types.Qid)
```
balopat, line 46: Nit: Add some docstring to explain what these mean.
Question: Why TKey2? Is there going to be a Tkey3, ... TkeyN eventually as the key types grow?

cirq/ops/pauli_string.py
```diff
@@ -163,27 +164,52 @@ def _value_equality_values_cls_(self):
...
+        pass
+
+    @overload
+    def __mul__(self, other: Iterable['cirq.PAULI_STRING_LIKE']
+               ) -> 'cirq.PauliString[Union[TKey, cirq.Qid]]':
+        pass
+
+    @overload
+    def __mul__(self, other: 'cirq.Operation'
+               ) -> 'cirq.PauliString[Union[TKey, cirq.Qid]]':
```
balopat, line 204: why use cirq.Qid here directly? Wasn't that the point with TKeyX?

cirq/ops/pauli_string.py
```diff
@@ -697,8 +723,8 @@ def __rpow__(self, base):
                 exponent_pos=-half_turns / 2)
         return NotImplemented
 
-    def map_qubits(self, qubit_map: Dict[raw_types.Qid, raw_types.Qid]
-                   ) -> 'PauliString':
+    def map_qubits(self,
+                   qubit_map: Dict[TKey, TKey2]) -> 'cirq.PauliString[TKey2]':
```
balopat, line 727: Why is it Tkey -> Tkey2? Can't it be Tkey -> Tkey?

balopat (changes requested):
I added some questions and docs nits

cirq/ops/pauli_string.py
Strilanc: The issue is that `other` is an operation over qubits, and is not required to be over the same type of qubits as the pauli string. We would need to make Operation generic to guarantee keeping `TKey`, which I do not think is viable since Operation is inherited by so many things.

cirq/ops/pauli_string.py
Strilanc: The input type is the dictionary key type, the output type is the dictionary value type. This is the most precise way of specifying what the method does. Using a union would give less type information.

cirq/ops/pauli_string.py
Strilanc: It doesn't make sense to define that constant, because it is used few times and also the meaning of `TKey2` is method specific (in the sense that it is "bound" separately by each method). `Union[TKey, TKey2]` really doesn't mean "all key types", it means "the result's keys come from either this pauli string or the given mapping".

balopat: Thanks for explaining. I get it now. TKey2 is used to allow for interactions with other PauliStrings that will be keyed potentially with a different type.

cirq/ops/pauli_string.py
balopat: Got it, makes sense.

cirq/ops/pauli_string.py
balopat: Got it, makes sense. Thanks!

balopat (approved):
LGTM, I would still add docstrings to the type variables.

cirq/ops/pauli_string.py
Strilanc: I split them into `TKeyOther` and `TKeyNew` for clarity.

PR #3328, opened 2020-09-16
Change _book.yaml to point to /reference/python

yashk2810:


docs/_book.yaml
```diff
@@ -97,7 +97,7 @@ upper_tabs:
       skip_translation: true
```
lamberta, line 97: While you're here, can you change the `name` from "API" to "Reference"

yashk2810: Done

balopat (approved):

PR #3333, opened 2020-09-17
Add methods for converting dict to product or zip sweep and deprecate implicit cartesian product

### LINKED ISSUE #2698: Expand SWEEP_LIKE to work for {'t': [0, 1, 3]}

### LINKED ISSUE #3018: Deprecate implicit cartesian product behavior in `to_sweeps`.
This one seems tricky to deprecate, because its arguments are often ultimately from a user but are provided by intermediate methods such as `cirq.sample`. This makes it difficult to print a relevant error message in terms of what the user did, instead of the internal details of the library.

kevinsung:
Fixes #2698 .
Fixes #3018 .

### DIFF
--- a/cirq/study/sweeps.py
+++ b/cirq/study/sweeps.py
@@ -26,6 +26,8 @@ if TYPE_CHECKING:
     import cirq
 
 Params = Iterable[Tuple['cirq.TParamKey', 'cirq.TParamVal']]
+ProductOrZipSweepLike = Dict[
+    'cirq.TParamKey', Union['cirq.TParamVal', Sequence['cirq.TParamVal']]]
 
 
 def _check_duplicate_keys(sweeps):
@@ -437,10 +439,7 @@ def _params_without_symbols(resolver: resolver.ParamResolver) -> Params:
         yield cast(str, sym), cast(float, val)
 
 
-def dict_to_product_sweep(
-        factor_dict: Dict['cirq.TParamKey',
-                          Union['cirq.TParamVal', Sequence['cirq.TParamVal']]]
-) -> Product:
+def dict_to_product_sweep(factor_dict: ProductOrZipSweepLike) -> Product:
     """Cartesian product of sweeps from a dictionary.
 
     Each entry in the dictionary specifies a sweep as a mapping from the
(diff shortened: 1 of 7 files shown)


cirq/study/sweeps.py
```diff
@@ -439,3 +435,43 @@ def _params_without_symbols(resolver: resolver.ParamResolver) -> Params:
         if isinstance(sym, sympy.Symbol):
             sym = sym.name
         yield cast(str, sym), cast(float, val)
+
+
+def dict_to_product_sweep(
+        factor_dict: Dict['cirq.TParamKey',
```
balopat, line 441: I think having a named ZipOrProductSweepLike type alias would make this more readable, and signal it more clearly that the two methods are working off of the same type.

balopat (approved):
LGTM with one nitpick.

@crystalzhaizhai please have a look as well

cirq/study/sweeps.py
kevinsung: Done.

crystalzhaizhai (approved):
LGTM. 
quick question: why use cirq.TParamkey?

kevinsung:
Parameter keys can be of type sympy.Basic in addition to string. We use TParamKey to represent the type of a parameter so we can define it in one place and also make the code more readable.

PR #3334, opened 2020-09-17
Various updates to automerge script deployment

balopat:
Makes automerge script singleton, adds skaffold config, fixes cloud build deployment. 

- The current version was not deploying the new version of the image on pushes, due to an issue with GKE deploy. I switched it out to use skaffold. This also comes with a benefit of iterating against GKE directly. 
- I realized that it can be problematic to run multiple instances of the automerge bot, so I changed it to become a singleton application using StatefulSet 
- added docs to cover the changes
- optimized the Dockerfile to become smaller, the build faster by pointing the build context to dev_tools instead of the repo root
- also added a linebreak in the ... logs in automerge after every 100 dots to have a bit more feedback in the stackdriver logs even when there are no PR events


dev_tools/auto_merge/Dockerfile
```diff
@@ -16,23 +16,22 @@
...
-#     docker build -t cirqbot -f dev_tools/automerge/Dockerfile .
+#     docker build -t cirqbot -f dev_tools/automerge/Dockerfile dev_tools
 #     docker run -it -e CIRQ_BOT_GITHUB_ACCESS_TOKEN=[access token] cirqbot
 #
 # Alternatively, if your container has access to the GCP Secret Manager, it can take the
 # value of the cirqbot-api-key secret.
 ########################################################################################
 
-FROM python:3.7-slim
+FROM python:3.7-slim@sha256:1689dea3fe5e2f230c3b2d0cc18b64e96f8214082c78792c67dbc1bda9b6c67c
```
vtomole, line 26: Why this?

balopat: To achieve reproducible builds. It's basically pinning to a particular version.

Strilanc:
The `.` logging is much sillier now that it's being run on a server inside of from my command line.

Strilanc (approved):

PR #3337, opened 2020-09-17
Explaining Engine API access

balopat:
More explanation around the Engine API.

### DIFF
--- a/docs/google/engine.md
+++ b/docs/google/engine.md
@@ -1,9 +1,17 @@
 # Quantum Engine
 
-The Quantum Engine, via the `cirq.google.Engine` class, executes programs and jobs using the
-Quantum Engine API.
+Google's Quantum Computing Service provides the Quantum Engine API to execute 
+circuits on Google's quantum processor or simulator backends and 
+to access or manage the jobs, programs, reservations and calibrations. Cirq is 
+the only supported client for this API, using the `cirq.google.Engine` class. 
 
-Note that the Quantum Engine API is not yet open for public access.
+Note: the Quantum Engine API is not yet open for public access.
+
+Note: while Cirq is the only supported way of using to the API currently, 
+if you are interested in integrating directly with the Quantum Engine API,
+instead of using Cirq, e.g. from a different language, we'd like to hear about 
+your proposal at 
+[cirq-maintainers@googlegroups.com](mailto:cirq-maintainers@googlegroups.com).
 
 ## Authenticating to Google Cloud
 
--- a/docs/google/engine.md
+++ b/docs/google/engine.md
@@ -2,16 +2,13 @@
 
 Google's Quantum Computing Service provides the Quantum Engine API to execute 
 circuits on Google's quantum processor or simulator backends and 
-to access or manage the jobs, programs, reservations and calibrations. Cirq is 
+to access or manage the jobs, programs, reservations and calibrations. As of Cirq is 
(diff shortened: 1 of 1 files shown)


karlunho (approved):
Looks great. It might be useful to mention that we have weekly open meeting to review proposals.

docs/google/engine.md
```diff
@@ -1,9 +1,17 @@
...
-The Quantum Engine, via the `cirq.google.Engine` class, executes programs and jobs using the
-Quantum Engine API.
-
-Note that the Quantum Engine API is not yet open for public access.
+Google's Quantum Computing Service provides the Quantum Engine API to execute 
+circuits on Google's quantum processor or simulator backends and 
+to access or manage the jobs, programs, reservations and calibrations. Cirq is 
+the only supported client for this API, using the `cirq.google.Engine` class. 
+
+Note: the Quantum Engine API is not yet open for public access.
```
dabacon, line 8: add timestamp to "not yet"

docs/google/engine.md
```diff
@@ -1,9 +1,17 @@
 # Quantum Engine
 
-The Quantum Engine, via the `cirq.google.Engine` class, executes programs and jobs using the
-Quantum Engine API.
-
-Note that the Quantum Engine API is not yet open for public access.
+Google's Quantum Computing Service provides the Quantum Engine API to execute 
+circuits on Google's quantum processor or simulator backends and 
+to access or manage the jobs, programs, reservations and calibrations. Cirq is 
+the only supported client for this API, using the `cirq.google.Engine` class.
```
dabacon, line 6: (maybe timestamp this statement, i.e. currently (as of XX))

dabacon (approved):

docs/google/engine.md
```diff
@@ -1,9 +1,17 @@
...
-
-Note that the Quantum Engine API is not yet open for public access.
+Google's Quantum Computing Service provides the Quantum Engine API to execute 
+circuits on Google's quantum processor or simulator backends and 
+to access or manage the jobs, programs, reservations and calibrations. Cirq is 
+the only supported client for this API, using the `cirq.google.Engine` class. 
+
+Note: the Quantum Engine API is not yet open for public access.
+
+Note: while Cirq is the only supported way of using to the API currently,
```
dstrain115, line 10: nit:  It's slightly awkward to have two "Note:" in a row.  Maybe move this note up to the previous paragraph and condense it, such as:

Cirq is the only supported client for this API, using the `cirq.google.Engine` class.  For other use cases (e.g. from a different language), contact [cirq-maintainers@googlegroups.com](mailto:cirq-maintainers@googlegroups.com) with a proposal.

docs/google/engine.md
balopat: Interesting idea - I'm guessing you're trying to avoid the situation where we open up the doors, and then this will be forgotten and will be weird? 

But what you're suggesting (timestamp / date) gets outdated immediately - i.e. for 9/15/2020 it will raise questions in the heads of users in November - is this still accurate? So to avoid that, we'd need to update these timestamps at least monthly?  I'm not convinced yet about timestamping. I think when we open the doors, we'll have a big revamp anyway and make sure the user journey's make sense, minimizing the risk of the failure scenario above.

balopat:
> Looks great. It might be useful to mention that we have weekly open meeting to review proposals.

I added the RFC process as an alternative to mailing the maintainers, that should cover the meeting review as well.

Issue #3196, opened 2020-08-06
Glitched Circuit Diagram after flatten


AlkaidCheng:
**Description of the issue**
When a circuit has a parameterised gate as a product of symbols, flattening the circuit will produce a glitched circuit diagram where the parameterised expression is gone. This will further affect the circuit diagram if two qubit gates are also present in the circuit. 
**How to reproduce the issue**

```
import sympy as sp
from cirq import CNOT, rz, Circuit
cq = cirq.Circuit()
qubits = cirq.GridQubit.rect(2,1)
x = sp.symarray('x', 2)
cq.append([rz(x[0]*x[1])(qubits[-1]), CNOT(*tuple(qubits))])
new_cq, _= cirq.flatten(cq)
new_cq
```
This will product circuit diagram like this

This will product circuit diagram like this

(0, 0): ─────────────────────────@───
　　　　　　　　　　　│
(1, 0): ───Rz(pi*)───X───

**Cirq version**
0.8.2

balopat:
Hi @AlkaidCheng  - the result on the CLI is 

```
(0, 0): ─────────────────────────@───
                                 │
(1, 0): ───Rz(pi*<x_0*x_1/pi>)───X───

```

I'm guessing that the angle brackets `<` `>` create the problem in a browser for you - maybe Jupyter notebooks?

balopat:
Yeah I just confirmed that that's the issue. In Jupyter notebooks this shows up incorrectly, but only if the object is returned as the result of the cell - if you use print, it works
```
print(new_cq)
```
Gives a workaround, but it might be a bit annoying. Not sure yet what the right solution is here.

AlkaidCheng:
Right, I was using Colab notebooks and doing print will solve the problem.

AlkaidCheng:
I have been using the circuit is returned as the result of the cell for printing circuits because for jupyter notebooks, doing print(circuit) will mess up the diagram by introducing extra newlines if the circuit is too long. Interestingly doing print(circuit) in Colab notebooks does not have this problem.

balopat:
For someone who wants to look at this should review the possibility of replacing `<>` angle brackets with another pair of symbols (e.g. `{}` or `[]`) here: https://github.com/quantumlib/Cirq/blob/master/cirq/study/flatten_expressions.py

abhik-99:
Hey! Can I contribute to this issue?

balopat:
[...]

PR #3343, opened 2020-09-19
Refactor to_resolvers to use to_sweeps

kevinsung:
This seems logical. Note, however, that since `to_sweeps` currently contains deprecated code, this introduces deprecated behavior into `to_resolvers`. To avoid that we could wait for the next release before merging this.

### DIFF
--- a/cirq/study/sweepable.py
+++ b/cirq/study/sweepable.py
@@ -35,20 +35,8 @@ document(
 
 def to_resolvers(sweepable: Sweepable) -> Iterator[ParamResolver]:
     """Convert a Sweepable to a list of ParamResolvers."""
-    if sweepable is None:
-        yield ParamResolver({})
-    elif isinstance(sweepable, ParamResolver):
-        yield sweepable
-    elif isinstance(sweepable, Sweep):
-        yield from sweepable
-    elif isinstance(sweepable, dict):
-        yield ParamResolver(cast(Dict, sweepable))
-    elif isinstance(sweepable, Iterable) and not isinstance(sweepable, str):
-        for item in cast(Iterable, sweepable):
-            yield from to_resolvers(item)
-    else:
-        raise TypeError(f'Unrecognized sweepable type: {type(sweepable)}.\n'
-                        f'sweepable: {sweepable}')
+    for sweep in to_sweeps(sweepable):
+        yield from sweep
 
 
 def to_sweeps(sweepable: Sweepable) -> List[Sweep]:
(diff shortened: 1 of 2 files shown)


cirq/study/sweepable.py
```diff
@@ -35,20 +35,8 @@
...
-        yield from sweepable
-    elif isinstance(sweepable, dict):
-        yield ParamResolver(cast(Dict, sweepable))
-    elif isinstance(sweepable, Iterable) and not isinstance(sweepable, str):
-        for item in cast(Iterable, sweepable):
-            yield from to_resolvers(item)
-    else:
-        raise TypeError(f'Unrecognized sweepable type: {type(sweepable)}.\n'
-                        f'sweepable: {sweepable}')
+    for sweep in to_sweeps(sweepable):
```
balopat, line 38: Something is not quite right here. Shouldn't this be a double loop? A Sweep is an iterator over param resolvers. 
Also, the fact that this is not breaking any tests in itself makes me wonder if I'm wrong or the tests :)

balopat: Oh ignore me - it is `yield from sweep` - not `yield sweep`

cirq/study/sweepable.py
```diff
@@ -35,20 +35,8 @@
...
 def to_resolvers(sweepable: Sweepable) -> Iterator[ParamResolver]:
     """Convert a Sweepable to a list of ParamResolvers."""
-    if sweepable is None:
-        yield ParamResolver({})
-    elif isinstance(sweepable, ParamResolver):
-        yield sweepable
-    elif isinstance(sweepable, Sweep):
-        yield from sweepable
-    elif isinstance(sweepable, dict):
-        yield ParamResolver(cast(Dict, sweepable))
```
balopat, line 45: I would like to note that this change is not exactly equivalent: 

This is passing with the new version: 
```
    assert [cirq.ParamResolver({'a': 2, 'b':1})] == list(cirq.to_resolvers({'a':[2], 'b': [1]}))
```
but fails in master: 

```
E       AssertionError: assert [cirq.ParamResolver({'a': 2, 'b': 1})] == [cirq.ParamResolver({'a': [2], 'b': [1]})]
```
However, I don't think this is a big issue. We do not officially support list values here, so I don't think our users used array values at all. I saw one example in https://github.com/quantumlib/Cirq/issues/3196 - but I can't figure out how to resolve that array value there anyway.

balopat (approved):
I added a comment as the change is not completely equivalent, however the new version is strictly better - as we have a single pathway to reason about behavior. I am not worried about the deprecation for the same reason I'm not worried about the inequivalent behavior issue: I don't think people used list valued dicts with to_resolvers anyway. Thus I am comfortable merging this.

Issue #1808, opened 2019-07-12
Rename TrialResult to just Result (with deprecated TrialResult for compatibility)


Strilanc:

dabacon:
This seems reasonable.  I suspect not many people use TrialResult directly as a name, only as a consumer (Though maybe if they are doing type annotations).

balopat:
Related #3257

tonybruguier:
Would https://github.com/quantumlib/Cirq/pull/3344 be a reasonable attempt?

balopat:
Closed in #3344.

PR #3344, opened 2020-09-20
s|TrialResult|Result

tonybruguier:
Issue #1808


cirq/__init__.py
```diff
@@ -393,7 +393,7 @@
     to_resolvers,
     to_sweep,
     to_sweeps,
-    TrialResult,
+    Result,
```
viathor, line 396: Instead of removing `TrialResult`, let's keep it [marked](https://github.com/quantumlib/Cirq/blob/8e30cf6815dcb1ebbd0244243f49e64b7cd08366/cirq/_compat.py#L85) as deprecated (see [example](https://github.com/quantumlib/Cirq/blob/33194ad3a2f4c28c48417876ff7dd1a6732d3e41/cirq/linalg/states.py#L23)).

viathor (changes requested):

cirq/__init__.py
tonybruguier: Like this? Or do you just want me to mark it as deprecated for now? Happy either way, just trying to understand what you prefer.

Also, I realized, thanks to your comment, that I should have renamed a file which I just did.

tonybruguier:
Like this? There are tests that area failing and I will take a look if it's the correct approach. Otherwise, please let me know what you had in mind. Thanks for all the guidance!

cirq/__init__.py
balopat: We should mark it as deprecated - i.e. anytime someone uses TrialResult, they should get a warning. The target deprecation release would be 0.11.0 for this.

tonybruguier: Moved to 0.11.0

Any suggestion on how to fix the errors? I am a bit confused, because it's complaining about the old object, which I intend to delete, right?

tonybruguier (commented):
Thanks. PTAL.

rtd_docs/api.rst
```diff
@@ -286,7 +287,6 @@ results.
     cirq.StepResult
     cirq.Sweep
     cirq.Sweepable
-    cirq.TrialResult
```
balopat, line 289: instead of removing this you should move it to the Deprecated section below: https://github.com/quantumlib/Cirq/pull/3344/files#diff-c69e423203b6c17f4ed493b8ae22952fL742

cirq/study/visualize.py
```diff
@@ -16,10 +16,10 @@
 
 import numpy as np
 
-from cirq.study import trial_result
+from cirq.study import result
```
balopat, line 19:
```
if TYPE_CHECKING:
    from cirq.study import result
```

balopat (changes requested):
I think this is looking much better, for the tests you'll have to fix visualize.py and api.rst - I added details.
Also, please add a deprecation test for TrialResult. A simple instantiation of cirq.TrialResult is sufficient and check for the logs with ` with cirq.testing.assert_logs(...):`

cirq/study/visualize.py
balopat: Also, you'll need to use single quotes for the type`'result.Result'`  in the `plot_state_histogram` signature below

rtd_docs/api.rst
tonybruguier: Done.

cirq/study/visualize.py
tonybruguier: Done.

tonybruguier:
Thanks. PTAL.

balopat:
> Thanks. PTAL.

Looks good, the only thing missing is a test for the deprecation. Please add a test with a simple instantiation of cirq.TrialResult is sufficient and check for the deprecation logs with with cirq.testing.assert_logs(...):

tonybruguier:
Thanks. PTAL.

balopat (approved):

cirq/protocols/json_serialization.py
```diff
@@ -159,7 +159,7 @@ def two_qubit_matrix_gate(matrix):
                 'SycamoreGate': cirq.google.SycamoreGate,
                 'TaggedOperation': cirq.TaggedOperation,
                 'ThreeDQubit': cirq.pasqal.ThreeDQubit,
-                'TrialResult': cirq.TrialResult,
+                'Result': cirq.Result,
```
mpharrigan, line 162: for backwards compatibility, both "Result" and "TrialResult" should be present and point to `cirq.Result`

cirq/protocols/json_test_data/Result.json
```diff
@@ -1,6 +1,6 @@
 [
```
mpharrigan, line 1: for backwards compatibility testing, the old TrialResult.json should be left *unmodified* and a new Result.json should be added

Issue #3347, opened 2020-09-21
Process: follow numpy's python support cadence


balopat:
**Is your feature request related to a use case or problem? Please describe.**

We need clarity around what python versions we support. As python evolves, we get issues for incompatibility. (see issues labeled with python3.8), and we should have CI tests for the supported python versions. 

**Describe the solution you'd like**

As we depend on NumPy, we could follow their support process of 42 months window: https://numpy.org/neps/nep-0029-deprecation_policy.html

**What is the urgency from your perspective for this issue? Is it blocking important work?**

P2 - we should do it in the next couple of quarters

balopat:
No objections at Cirq Cynq!

balopat:
We have to revisit this policy, as we are committed to be able to run on Google Colab. This forces us to support 3.6 at the moment as their hosted runtime is lagging behind. Eventually that constraint will be lifted, but until then we'll have to take it into consideration.

balopat:
Also Google internally is running on python 3.6 for now. So at minimum we are tied to support these two systems: Google3 + colab.

95-martin-orion:
> @dabacon, does the before-1.0 label mean we plan to drop support for 3.6 before releasing 1.0?

+1 to this, Google internals were upgraded recently (to Python 3.9, I think) and there's a number of cleanup actions we can take once 3.6 support is dropped.

maffoo:
Colab is now using 3.7.13 AFAICT. Are there any other blockers to dropping 3.6 support?

mpharrigan:
Now we should be 3.8+ (but blocked by colab, probably). What does the before-1.0 tag mean here?

mpharrigan:
Bump!

maffoo:
We're still blocked on colab, so I think the only thing we can really do here is write down our intended policy in the docs, something like "cirq supports python versions specified by numpy NEP-29, in addition to the current version used by google colab (3.7)". Other than that I don't think there's much we can do before 1.0.

PR #3348, opened 2020-09-21
adding 3.7 to the tests

balopat:
To cover all python runtimes that we support, adding coverage for 3.7 besides 3.6.
With #3194 we can add 3.8 as well, and soon 3.9 hopefully to the mix. 
If we agree on #3347, then we can drop 3.6 as well.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 # Runtime requirements for the python 3 version of cirq.
 
-dataclasses; python_version < '3.7'
+dataclasses; python_version < '3.8'
 freezegun~=0.3.15
 google-api-core[grpc] >= 1.14.0, < 2.0.0dev
 matplotlib~=3.0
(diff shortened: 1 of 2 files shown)


Untitled.ipynb
```diff
@@ -0,0 +1,43 @@
+{
```
vtomole, line 1: Remove this file.

vtomole (changes requested):

Untitled.ipynb
balopat: ahh :D thanks.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 # Runtime requirements for the python 3 version of cirq.
 
-dataclasses; python_version < '3.7'
```
vtomole, line 3: Is this change relevant to this PR? I haven't been keeping track on if we've are doing `3.8`.

balopat: This is PR is just raising the level to 3.7 i.e. < 3.8 should be fine. I've been using cirq on 3.7 for a long time now. The 3.8 work will build on top of this.

vtomole (approved):
LGTM

PR #3355, opened 2020-09-23
Add cirq.StabilizerSampler

Strilanc:
- Minimal interface
- Supports some random operations (depolarize, probabilistic gates)
- This is a potential replacement for cirq.CliffordSimulator (which is in a bad state right now, tracking two redundant state objects in a fashion observable to callers)


cirq/sim/clifford/stabilizer_sampler.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def _run(self, circuit: circuits.Circuit,
+             repetitions: int) -> Dict[str, np.ndarray]:
+
+        measurements: Dict[str, List[int]] = {
+            key: [] for key in protocols.measurement_keys(circuit)
+        }
+        axes_map = {q: i for i, q in enumerate(circuit.all_qubits())}
+
+        for _ in range(repetitions):
```
viathor, line 64: Question: Why not propagate the stabilizer once and sample from the resulting stabilizer state at the end? I suppose it's simply deferred to a follow-up?

viathor (approved):

cirq/ops/random_gate_channel.py
```diff
@@ -109,6 +109,20 @@ def _trace_distance_bound_(self) -> float:
             result *= float(self.probability)
         return result
 
+    def _act_on_(self, args):
```
balopat, line 112: missing type annotations here as well

balopat (approved):
LGTM with nits

cirq/sim/clifford/stabilizer_sampler.py
Strilanc: Measurements can return random results which play out in different ways. In principle we could check if all measurements are terminal, or translate the circuit into a graph dataflow form that was faster to sampler from (assuming all classically controlled operations were Paulis). I wasn't planning to do these optimizations.

viathor: The optimization would mostly target circuits with terminal measurements. We could also support measurements commuting with stabilizer, but that may be excessive and not useful. Finally, in some applications you want probability of a prescribed (Pauli) measurement result and in this case there is a unique update to the stabilizer generators consistent with the result.

cirq/ops/common_channels.py
```diff
@@ -183,6 +183,17 @@ def __repr__(self) -> str:
     def __str__(self) -> str:
         return 'depolarize(p={!r})'.format(self._p)
 
+    def _act_on_(self, args):
```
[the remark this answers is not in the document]
Strilanc: Done.

Issue #3056, opened 2020-06-01
Cirq routing is hanging again for some circuits


yourball:
Hi guys,

I have a concern similar to the one from issue #2282. The problem was fixed later and when running from `6d81ddbb` commit everything works just fine. However, there were further changes in apply_next_swaps(...) function, in particular a new parameter `max_num_empty_steps` was introduced.

In the last version of code when running routing via route_circuit_greedily(...) and with parameters `max_search_radius=1`, `max_num_empty_steps=1` everything also works well. However, when choosing different values for these two parameters (e.g. `max_search_radius=2`, `max_num_empty_steps=1`) routing hangs on some circuit instances.

Should there be any constraint on how to choose values for the parameters `max_search_radius` and `max_num_empty_steps`? Or maybe there is some bug in the routing code?

Thanks!

dabacon:
@bryano I think you added this feature.  Thoughts?

bryano:
@wcourtney Could you provide some circuits (or code to generate them) on which the router hangs?

yourball:
This is the code snippet.
It works fine if number of qubits involved in the random circuit <16 (number of qubits in the hardware is 16) or for small value of n_moments.

```
import networkx as nx
import cirq
from cirq import LineQubit
from cirq.contrib.routing.greedy import route_circuit_greedily

edges = [
        [0, 1],
        [0, 7],
        [0, 15],
        [1, 2],
        [7, 6],
        [7, 8],
        [15, 8],
        [15, 14],
        [2, 3],
        [3, 4],
        [4, 5],
        [5, 6],
        [8, 9],
        [9, 10],
        [10, 11],
        [11, 12],
        [12, 13],
        [13, 14],
        ]

def create_cirq_hardware(edges):
    graph = nx.Graph()
    for edge in edges:
        graph.add_edges_from([(LineQubit(edge[0]), LineQubit(edge[1]))])
    return graph

cirq_hardware = create_cirq_hardware(edges)
circuit = cirq.testing.random_circuit(qubits=16, n_moments=200, 
                                      op_density=.5, random_state=1)

swap_network = route_circuit_greedily(circuit, 
                                      cirq_hardware,
                                      max_search_radius=2, 
                                      max_num_empty_steps=1,
                                      random_state=1)
```

balopat:
This is definitely odd: 

In colab works up to 8 qubits (~ 1 minute at that point) and then it hangs (it's been running for 12+ minutes now).

[...]

PR #3360, opened 2020-09-24
Greedy router

yourball:
Fixes issue #3056. Fixes hanging greedy router on some circuit instances by partially reverting changes from 6d81ddbb.


yourball:
@googlebot I signed it!

balopat (changes requested):
Thank you for taking the time digging this out!! The fix looks good to me, as it is covered by tests. However, it would be great to put a test around it so that we can prevent regression (like what we have now).

The test would run a minimal instance where we can prove that the algorithm would run on infinite loops without this fix, but fast (<1s) with the fix. The algorithm would run in a separate process that we should measure if it finishes within the expected time. Can you find a graph instance like that?

balopat (approved):
LGTM - One more nit 
Thanks for adding the test, I just tried it out and it works nicely! :) 
We should make the timeout 5 seconds to avoid flakes, on my Mac it runs just above 2 seconds.

cirq/contrib/routing/greedy_test.py
```diff
@@ -20,10 +22,47 @@
...
+def test_router_hanging():
+    """Run a separate process and check if greedy router hits timeout (1s)."""
+    circuit, device_graph = create_circuit_and_device()
+    process = Process(target=create_hanging_routing_instance,
+                      args=[circuit, device_graph])
+    process.start()
+    process.join(timeout=1)
+    alive = process.is_alive()
+    if alive:
+        process.terminate()
```
balopat, line 67: I would do something like this instead - this takes care of the coverage issue as well

```
    try:
        assert not process.is_alive(), "Greedy router timeout"
    finally:
        process.terminate()
```

yourball: Cool, thanks! Was happy to contribute. It looks like all tests passed now.:)

PR #3361, opened 2020-09-26
Add symmetric depol

tonybruguier:
For issue #3220

### DIFF
--- a/cirq/ops/common_channels.py
+++ b/cirq/ops/common_channels.py
@@ -305,7 +305,6 @@ class DepolarizingChannel(gate_features.SingleQubitGate):
             return f"D({self._p:.{args.precision}g},{self._n_qubits})"
         return f"D({self._p},{self._n_qubits})"
 
-
     @property
     def p(self) -> float:
         """The probability that one of the Pauli gates is applied.
--- a/cirq/ops/common_channels.py
+++ b/cirq/ops/common_channels.py
@@ -268,7 +268,7 @@ class DepolarizingChannel(gate_features.SingleQubitGate):
         p_depol = p / (4**n_qubits - 1)
         p_identity = 1.0 - p
         for pauli_tuple in itertools.product(['I', 'X', 'Y', 'Z'],
-                                              repeat=n_qubits):
+                                             repeat=n_qubits):
             pauli_string = ''.join(pauli_tuple)
             if pauli_string == 'I' * n_qubits:
                 error_probabilities[pauli_string] = p_identity
--- a/cirq/ops/common_channels.py
+++ b/cirq/ops/common_channels.py
@@ -80,7 +80,8 @@ class AsymmetricDepolarizingChannel(gate_features.SingleQubitGate):
                 value.validate_probability(v, f"p({k})")
             sum_probs = sum(error_probabilities.values())
             if abs(sum_probs - 1.0) > 1e-6:
-                raise ValueError(f"Probabilities do not add up to 1 but to {sum_probs}")
+                raise ValueError(
+                    f"Probabilities do not add up to 1 but to {sum_probs}")
             self._num_qubits = num_qubits
             self._error_probabilities = error_probabilities
         else:
--- a/cirq/ops/common_channels.py
+++ b/cirq/ops/common_channels.py
@@ -262,9 +262,9 @@ class DepolarizingChannel(gate_features.SingleQubitGate):
 
         p_depol = p / (4**n_qubits - 1)
         # For numerical noise, we compute the probability of the identity at the
-        # very last so that the probabilities add up to 1.0 exactly.
+        # very last so that the probabilities add up to 1.0 exactly. We use the
+        # strange formulation below, which is mathematically the same as 1 - p.
         p_identity = 1.0 - sum([p_depol] * (4**n_qubits - 1))
-
 
         for coefficients in itertools.product(['I', 'X', 'Y', 'Z'],
                                               repeat=n_qubits):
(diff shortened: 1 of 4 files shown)


tonybruguier:
A few remarks to the reviewer:

- I changed the check for the probability in the base class. I think I had made a mistake in my previous PR #3262 where I checked that the sum of the probabilities was a probability (i.e. between 0 and 1). In fact, the sum should be exactly 1, if I am not mistake. I thus changed the test here. The reason I noticed it is that because of numerical noise, it was adding up to 1.0000002 which failed.
- I named the new parameter 'n_qubits'. I had originally named it 'num_qubits' but it was making this test fail examples/direct_fidelity_estimation_test.py. I am not sure why. I am actually also to blame for that file, so I am happy to fix it.

Thanks for all the help so far!

cirq/ops/common_channels.py
```diff
@@ -227,33 +230,53 @@ def asymmetric_depolarize(p_x: Optional[float] = None,
...
+                the Pauli gates is applied independently with probability
+                p / (2**n - 1).
+            n_qubits: the number of qubits.
 
         Raises:
             ValueError: if p is not a valid probability.
         """
 
+        error_probabilities = {}
+        all_identity = ''.join(['I'] * n_qubits)
```
viathor, line 263:
```python
all_identity = 'I' * n_qubits
```

Also, you don't need to set this in a variable - just inline in the conditional below.

cirq/ops/common_channels.py
```diff
@@ -227,33 +230,53 @@ def asymmetric_depolarize(p_x: Optional[float] = None,
...
+
+        p_depol = p / (4**n_qubits - 1)
+        p_identity = 1.0 - p
+        for coefficients in itertools.product(['I', 'X', 'Y', 'Z'],
+                                              repeat=n_qubits):
+            pauli_string = ''.join([g for g in coefficients])
+            if pauli_string == all_identity:
+                error_probabilities[pauli_string] = p_identity
+            if pauli_string != all_identity:
```
viathor, line 272: `else`

cirq/ops/common_channels.py
```diff
@@ -227,33 +230,53 @@ def asymmetric_depolarize(p_x: Optional[float] = None,
...
         Raises:
             ValueError: if p is not a valid probability.
         """
 
+        error_probabilities = {}
+        all_identity = ''.join(['I'] * n_qubits)
+
+        p_depol = p / (4**n_qubits - 1)
+        p_identity = 1.0 - p
+        for coefficients in itertools.product(['I', 'X', 'Y', 'Z'],
```
viathor, line 267: coefficients -> pauli_tuple

(I think most people would think that a coefficient is a number, but these are tuples of strings.)

cirq/ops/common_channels.py
```diff
@@ -227,33 +230,53 @@ def asymmetric_depolarize(p_x: Optional[float] = None,
...
+                                              repeat=n_qubits):
+            pauli_string = ''.join([g for g in coefficients])
```
viathor, line 269: No need for a list comprehension:

```python
pauli_string = ''.join(pauli_tuple)
```

cirq/ops/common_channels.py
```diff
@@ -265,55 +288,75 @@ def _value_equality_values_(self):
...
+        if self._n_qubits == 1:
+            if args.precision is not None:
+                return f"D({self._p:.{args.precision}g})"
+            return f"D({self._p})"
         if args.precision is not None:
-            f = '{:.' + str(args.precision) + 'g}'
-            return 'D({})'.format(f).format(self._p)
-        return 'D({!r})'.format(self._p)
+            return f"D({self._p:.{args.precision}g},{self._n_qubits})"
+        return f"D({self._p},{self._n_qubits})"
```
viathor, line 308: I think in a circuit diagram the number of qubits can be inferred from the number of lines this operation touches, so we don't need to include it at all.

cirq/ops/common_channels.py
```diff
@@ -265,55 +288,75 @@ def _value_equality_values_(self):
...
 
     @property
     def p(self) -> float:
         """The probability that one of the Pauli gates is applied.
 
-        Each of the Pauli gates is applied independently with probability p / 3.
+        Each of the Pauli gates is applied independently with probability
+        p / (2**n_qubits - 1).
```
viathor, line 315: 4\*\*n

cirq/ops/common_channels.py
```diff
@@ -265,55 +288,75 @@ def _value_equality_values_(self):
...
         """
         return self._p
 
+    @property
+    def n_qubits(self) -> int:
+        """The number of qubits
+        """
```
viathor, line 322: All quotes should be on the same line, I think (though the formatter takes precedence over my opinion!).

viathor (changes requested):
Thank you! Some comments below. In particular, there is a lot of comments that give an incorrect number of Pauli operators on n qubits (the code seems fine, though).

cirq/ops/common_channels.py
```diff
@@ -77,8 +78,10 @@ def __init__(self,
                     raise ValueError(f"{k} must have {num_qubits} Pauli gates.")
             for k, v in error_probabilities.items():
                 value.validate_probability(v, f"p({k})")
-            value.validate_probability(sum(error_probabilities.values()),
-                                       'sum(error_probabilities)')
+            sum_probs = sum(error_probabilities.values())
+            if abs(sum_probs - 1.0) > 1e-6:
+                raise ValueError(
+                    f"Probabilities do not add up to 1 but to {sum_probs}")
```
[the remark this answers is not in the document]
tonybruguier: Added a TODO.

tonybruguier:
Thanks. PTAL.

viathor (approved):
Thank you!

balopat (approved):

PR #3363, opened 2020-09-29
Execute TODO to have the prob mass auto added

tonybruguier:
Issue #3220

### DIFF
--- a/cirq/ops/common_channels.py
+++ b/cirq/ops/common_channels.py
@@ -199,7 +199,6 @@ def asymmetric_depolarize(
         p_z: Optional[float] = None,
         error_probabilities: Optional[Dict[str, float]] = None,
         tol: float = 1e-8) -> AsymmetricDepolarizingChannel:
-                          tol: float = 1e-8) -> AsymmetricDepolarizingChannel:
     r"""Returns a AsymmetricDepolarizingChannel with given parameter.
 
         This channel applies one of 4**n disjoint possibilities: nothing (the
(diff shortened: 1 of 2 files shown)


cirq/ops/common_channels.py
```diff
@@ -79,10 +79,11 @@ def __init__(self,
             for k, v in error_probabilities.items():
                 value.validate_probability(v, f"p({k})")
             sum_probs = sum(error_probabilities.values())
-            # TODO(tonybruguier): Instead of forcing the probabilities to add up
-            # to 1, check whether the identity is missing, and if that is the
-            # case, automatically add it with the missing probability mass.
-            if abs(sum_probs - 1.0) > 1e-6:
+            tol = 1e-6
```
viathor, line 82: Let's make this an argument (and document it).

viathor (approved):

cirq/ops/common_channels.py
```diff
@@ -36,8 +36,8 @@ def __init__(self,
                  p_x: Optional[float] = None,
                  p_y: Optional[float] = None,
                  p_z: Optional[float] = None,
-                 error_probabilities: Optional[Dict[str, float]] = None
-                ) -> None:
+                 error_probabilities: Optional[Dict[str, float]] = None,
+                 tol: Optional[float] = None) -> None:
```
viathor, line 40: Since below you make `None` equivalent to 1e-6, we don't actually need this to be an optional type. Let's change type to `float` and write the default explicitly.

viathor, line 40: BTW: Maybe let's use 1e-8 as the default for consistency with other code in cirq.

viathor (approved):

tonybruguier:
Thanks. PTAL.

balopat (approved):

PR #3365, opened 2020-09-29
Python 3.8 support

### LINKED ISSUE #3194: Python 3.8 support
This issue is to track the problems with Python 3.8.
It can be closed as soon as we can develop and run Cirq against 3.8, meaning that we have CI jobs against it as well.

### LINKED ISSUE #3318: Broken Jupyter notebook with current dev tools
**Description of the issue**

With the current dev tools setup, in a fresh virtual env there is a jupyter installation, however jupyter-notebook fails to connect to the kernel. 

image

Javascript console: 

```
WebSocket connection to 'ws://localhost:8888/api/kernels/f55b095d-e862-48c2-af3c-a104b8568761/channels?session_id=0901642FF6394942A2B33531093F227A' failed: Error during WebSocket handshake: Unexpected response code: 200
Kernel.start_channels @ kernel.js:461
```
On the notebook logs:

```
/Users/balintp/envs/cirq37/lib/python3.7/site-packages/notebook/base/zmqhandlers.py:284: RuntimeWarning: coroutine 'WebSocketHandler.get' was never awaited
  super(AuthenticatedZMQStreamHandler, self).get(*args, **kwargs)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

**Workaround**

This looks like a tornado issue https://github.com/jupyter/notebook/issues/4439 that requires `pip3 install --force-reinstall --upgrade jupyter` - which works. 

```
pip install -r requirements.txt -r cirq/contrib/contrib-requirements.txt -r dev_tools/conf/pip-list-dev-tools.txt
jupyter-notebook
```

**Cirq version**

master

balopat:
Ensures that packages are python 3.8 compatible. 
Adds CI tests for 3.8. 
This was a work based on @seunomonije's investigations.

Fixes #3194.
Fixes #3318.


seunomonije:
I see grpcio 1.26 was added in the latest commit. That's the only difference from what's working on my local machine to here.

dev_tools/conf/pip-list-dev-tools.txt
```diff
@@ -37,8 +35,5 @@ nbsphinx
 ipython
 ipykernel
 
-# For google notebooks
-google-colab
```
dstrain115, line 41: Will this break the Colabs that have engine auth examples?

requirements.txt
```diff
@@ -1,6 +1,6 @@
 # Runtime requirements for the python 3 version of cirq.
 
-dataclasses; python_version < '3.8'
+dataclasses; python_version < '3.9'
```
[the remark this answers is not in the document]
balopat: you know...actually this is silly. I modified it for 3.7 as well. But actually it is a version modifier for the dataclasses package, which means "only install this if python version is less than 3.9" - which doesn't make sense as dataclasses are included since 3.7 in python. So this is backport thing for python 3.6. I'll fix it. Great catch :D

dev_tools/conf/pip-list-dev-tools.txt
balopat: No it won't on colab. 
On jupyter locally we have issues anyways because of exactly google-colab: #3318 
On jupyter I'd suggest using something lighterweight, e.g. this works very nicely: 

```
# The Google Cloud Project id to use.
project_id = 'your project' #@param {type:"string"}


def setup_auth():
  """Runs the user through the Colab OAuth process.
  
  Sets the local Application Default Credentials. For more information on on 
  using Application Default Credentials see 
  https://cloud.google.com/docs/authentication/production
  """
  import pydata_google_auth
  credentials = pydata_google_auth.get_user_credentials(
    ['https://www.googleapis.com/auth/cloud-platform'],
  )


print("Getting OAuth2 credentials.")
print("Press enter after entering the verification code.")
setup_auth()
print("Authentication complete.")

try: 
    import cirq 
except:     
    !pip install cirq

print("Cirq is installed.")
```

I think I'll just update our colab/notebooks with this new header.

dstrain115: Yeah, I guess that's okay then.  I think I added it since readthedocs needed this for one of the notebooks.  If docs build and you can run it in colab, I guess it's fine, assuming we switch to that lightweight alternative.

dstrain115 (approved):

dev_tools/conf/pip-list-dev-tools.txt
balopat: Yeah, I remember that issue - specifically this build failed because of lack of the dependency on your first PR introducing these colabs: https://github.com/quantumlib/Cirq/runs/1052329100

I'm not sure what changed besides 3.6 -> 3.7 in the CI - but that doesn't explain why it works on RTD which still builds it on 3.6...
It is building now both on our CI and in RTD: https://balopat-cirq-rtd.readthedocs.io/en/latest/docs/tutorials/google/start.html (https://readthedocs.org/projects/balopat-cirq-rtd/builds/11997856/). I'm genuinely puzzled why the error happened on your branch then - I don't see any changes that would explain why it went away. 

Added #3368 for the lighter alternative.

balopat: I'm going to go ahead and merge. If something goes wrong, I will fix it.

PR #3366, opened 2020-09-29
Extending types for symbol resolution fast pass-through

### LINKED ISSUE #3359: float32 parameters leads to significant performance loss of parameter resolver
**Description of the issue**
Calling `cirq.resolve_parameters` to resolve the parameters of a symbolic circuit with `np.float32` parameters leads to a significant performance loss. I ran into this bug while working on loading circuit parameters from a `.npy` file on [this TensorQuantum Dataset PR](https://github.com/tensorflow/quantum/pull/304). Verifying that the circuits produced the expected wave functions took ~30 minutes on the CI, since we have to check hundreds of circuits. On further investigation, I discovered that the parameters were 32-bit, and casting them to 64-bit decreased the parameter resolve time by a large margin.

```python

import cirq
import sympy
import time
import numpy as np
from collections import namedtuple


def unique_name():
    """Generator to generate an infinite number of unique names.

    Yields:
        Python `str` of the form "theta_<integer>".

    """
    num = 0
    while True:
        yield "theta_" + str(num)
        num += 1


def time_resolve_param(dtype, nspins, max_depth):
    depth_grid = list(range(2, max_depth, 2))
    times = []
    for depth in depth_grid:
        qubits = cirq.GridQubit.rect(nspins, 1)

        # Symbolic parameters
        name_generator = unique_name()
        symbol_names = [next(name_generator) for _ in range(depth)]
        symbols = [sympy.Symbol(name) for name in symbol_names]

        # Random values with dtype
        params = np.random.randn(2 * depth).astype(dtype)
        params = dict(zip(symbol_names, params.flatten()))

        # Define the circuit.
        circuit = cirq.Circuit(cirq.H.on_each(qubits))
        zipped_qubits = list(zip(qubits, qubits[1:]))
        for d in range(depth):
            for q1, q2 in zipped_qubits:
                circuit.append(cirq.ZZ(q1, q2)**(symbols[d]))

        # Resolve parameters and time
        resolve_start = time.time()
        param_resolver = cirq.resolve_parameters(circuit, params)
        t_param = time.time() - resolve_start
        times.append(t_param)

    return depth_grid, times


if __name__ == "__main__":
    NSPINS = 10
    MAXDEPTH = 50
    grid, times_32 = time_resolve_param(np.float32, NSPINS, MAXDEPTH)
    _, times_64 = time_resolve_param(np.float64, NSPINS, MAXDEPTH)

    plt.plot(grid, times_32, label='np.float32')
    plt.plot(grid, times_64, label='np.float64')
    plt.xlabel('circuit depth')
    plt.ylabel('time (s)')
    plt.legend()
    plt.grid()
    plt.show()


```
This produces the following figure

float32_vs_float64_paramresolver

As we can see, with increasing circuit depth the time it takes to resolve the 32-bit parameters creates a massive slowdown.

**Cirq version**
Cirq 0.8.2

balopat:
Extends ParamResolver's logic to circumvent sympy's slowness to members of numbers.Number. 
It also generalizes sympy constants instead of only handling pi and NegativeOne.

Fixes #3359.

### DIFF
--- a/cirq/study/resolver.py
+++ b/cirq/study/resolver.py
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 """Resolves ParameterValues to assigned values."""
-
-from typing import Any, Dict, Iterator, Optional, TYPE_CHECKING, Union, cast
+import numbers
+from typing import Any, Dict, Iterator, Optional, TYPE_CHECKING, Union, cast, SupportsFloat
 import numpy as np
 import sympy
 from cirq._compat import proper_repr
@@ -64,7 +64,7 @@ class ParamResolver:
                                {} if param_dict is None else param_dict)
 
     def value_of(self,
-                 value: Union['cirq.TParamKey', float]) -> 'cirq.TParamVal':
+                 value: Union['cirq.TParamKey', numbers.Number]) -> 'cirq.TParamVal':
         """Attempt to resolve a parameter to its assigned value.
 
         Floats are returned without modification.  Strings are resolved via
@@ -90,7 +90,7 @@ class ParamResolver:
             The value of the parameter as resolved by this resolver.
         """
         # Input is a float, no resolution needed: return early
-        if isinstance(value, float):
+        if isinstance(value, (float, np.float32)):
             return value
 
         # Handles 2 cases:
@@ -99,7 +99,7 @@ class ParamResolver:
         # In both cases, return it directly.
         if value in self.param_dict:
             param_value = self.param_dict[value]
-            if isinstance(param_value, (float, int)):
+            if isinstance(param_value, (float, np.float32, int)):
                 return param_value
 
         # Input is a string and is not in the dictionary.
--- a/cirq/study/resolver.py
+++ b/cirq/study/resolver.py
@@ -63,8 +63,8 @@ class ParamResolver:
         self.param_dict = cast(ParamDictType,
                                {} if param_dict is None else param_dict)
 
-    def value_of(self,
-                 value: Union['cirq.TParamKey', numbers.Number]) -> 'cirq.TParamVal':
+    def value_of(self, value: Union['cirq.TParamKey', numbers.Number]
+                ) -> 'cirq.TParamVal':
         """Attempt to resolve a parameter to its assigned value.
 
         Floats are returned without modification.  Strings are resolved via
--- a/cirq/study/resolver.py
+++ b/cirq/study/resolver.py
@@ -14,7 +14,7 @@
 
 """Resolves ParameterValues to assigned values."""
 import numbers
-from typing import Any, Dict, Iterator, Optional, TYPE_CHECKING, Union, cast, SupportsFloat
+from typing import Any, Dict, Iterator, Optional, TYPE_CHECKING, Union, cast
 import numpy as np
 import sympy
 from cirq._compat import proper_repr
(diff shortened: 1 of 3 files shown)


cirq/study/resolver.py
```diff
@@ -90,7 +90,7 @@ def value_of(self,
             The value of the parameter as resolved by this resolver.
         """
         # Input is a float, no resolution needed: return early
-        if isinstance(value, float):
+        if isinstance(value, (float, np.float32)):
```
maffoo, line 93: Is there a reason not to have `int` here as well? In that case, could move this tuple of "pass-through" types to a module level variable and reuse it below.

viathor: What about `np.int`? Also, why just 32-bit numpy floats? Perhaps this tuple should be a constant to ensure it's consistent throughout and easily updated.

balopat: Alright, I went down a rabbit hole and made it work for numbers but also catered for sympy constants, and provided extensive tests for the cases that we care about. I think this is now a bit clearer and more robust than before.

cirq/study/resolver.py
```diff
@@ -132,10 +148,7 @@ def value_of(self,
         if isinstance(value, sympy.Pow) and len(value.args) == 2:
             return np.power(self.value_of(value.args[0]),
                             self.value_of(value.args[1]))
-        if value == sympy.pi:
-            return np.pi
-        if value == sympy.S.NegativeOne:
-            return -1
+
```
dstrain115, line 151: nit: extra space

cirq/study/resolver.py
```diff
@@ -89,18 +89,33 @@ def value_of(self,
         Returns:
             The value of the parameter as resolved by this resolver.
         """
-        # Input is a float, no resolution needed: return early
-        if isinstance(value, float):
-            return value
+
+        def pass_through(val) -> Optional[Any]:
```
dstrain115, line 93: What are your thoughts about making this a private function (e.g. _sympy_pass_through) with a short docstring rather than an inline nested function?  I personally dislike nested functions, though I don't know if this is a global sentiment.

Google style guide at least says "Avoid nested functions or classes except when closing over a local value."

cirq/study/resolver_test.py
```diff
@@ -13,33 +13,102 @@
...
+def test_value_of_transformed_types(val, resolved):
+    _assert_consistent_resolution(val, resolved)
+
+
+@pytest.mark.parametrize('val,resolved', [(sympy.I, 1j)])
+def test_value_of_substituted_types(val, resolved):
+    _assert_consistent_resolution(val, resolved, True)
+
+
+def _assert_consistent_resolution(v, resolved, subs_called=False):
```
dstrain115, line 55: Would it be worth having a short docstring here or in SubsAwareSymbol explaining why we have "subs_called" and why we are trying to avoid it?

dstrain115 (approved):
LGTM with some minor comments.

cirq/study/resolver.py
balopat: I'm open to it - the method is not used anywhere else that's why I kept it nested but I'm happy to move it out.

cirq/study/resolver_test.py
balopat: I'll add some explanation both here and to the helper class.

PR #3369, opened 2020-09-30
Change names of focused calibration protos for consistency.

dstrain115:
- Change name of FocusedCalibrationResults to FocusedCalibrationResult
to match the other Result proto messages.


wcourtney (approved):
Thanks for the quick fix!

balopat (approved):

PR #3370, opened 2020-09-30
Do not perform extra optimizations in "via B" decomposition

### LINKED ISSUE #3050: Avoid calling optimizers in optimizers
In review of #2955, @mrwojtek pointed out that optimizers should avoid calling other optimizers and instead leave any further optimization (such as merging single-qubit gates) up to the caller.

I agree with this since it makes it more explicit in the client code what optimizations actually take place and hence what output is produced. It can also reduce the number of calls to optimizers in some cases (e.g. when multiple optimizers are used).

In fact we already follow this approach in all optimizers except `decompose_two_qubit_interaction_into_four_fsim_gates_via_b`.

This issue is a request for comments on whether we want to remove

```python
    MergeSingleQubitGates().optimize_circuit(circuit)
    DropEmptyMoments().optimize_circuit(circuit)
```

from the end of that optimizer and whether we should adopt the rule of leaving such additional optimization steps up to the user in future optimizers.

viathor:
Fixes #3050.

### DIFF
--- a/cirq/optimizers/two_qubit_to_fsim_test.py
+++ b/cirq/optimizers/two_qubit_to_fsim_test.py
@@ -91,40 +91,40 @@ def test_decompose_two_qubit_interaction_into_four_fsim_gates(
                           atol=1e-6)
 
 
-def test_decompose_two_qubit_interaction_into_four_fsim_gates_validate():
+def test_decompose_two_qubit_interaction_into_four_fsim_gates_validate_via_b():
     iswap = cirq.FSimGate(theta=np.pi / 2, phi=0)
     with pytest.raises(ValueError, match='fsim_gate.theta'):
-        cirq.decompose_two_qubit_interaction_into_four_fsim_gates(
+        cirq.decompose_two_qubit_interaction_into_four_fsim_gates_via_b(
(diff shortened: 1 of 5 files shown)


cirq/optimizers/two_qubit_to_fsim_test.py
```diff
@@ -76,55 +76,55 @@ def test_decompose_xx_yy_into_two_fsims_ignoring_single_qubit_ops_fail():
 
 @pytest.mark.parametrize('obj,fsim_gate',
```
balopat, line 77: I would still have some tests for the *via_b method to protect the functionality of the (now deprecated) method as it still has logic in there, which is now untested (though accidentally gets covered by the deprecation check).

balopat (approved):
LGTM with a nit around tests

cirq/optimizers/two_qubit_to_fsim_test.py
viathor: Restored testing of the old function in most cases. This way we test both old and new versions. The only case where I changed the test to test the new function is the test where an assert needs to be modified due to different length of the resulting circuit. This way removing deprecated function will be trivial.

PR #3372, opened 2020-09-30
Observable Measurement - 2 - Grouping

mpharrigan:
Part 2 of #2781 

Follows #3371


cirq/work/observable_grouping_test.py
```diff
@@ -0,0 +1,54 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import cirq
+
+
+def test_group_settings_greedy():
```
balopat, line 18: I'd love to see some more unit tests to document the detailed behavior for edge cases and simple cases as well, e.g. empty list, single items, single group, multi-group (that is the current one) situations.

cirq/work/observable_grouping_test.py
```diff
@@ -0,0 +1,54 @@
...
+        cirq.Y(q0) * cirq.X(q1) * cirq.X(q2) * cirq.Y(q3),
+        cirq.Y(q0) * cirq.Y(q1) * cirq.X(q2) * cirq.X(q3),
+        cirq.X(q0) * cirq.X(q1) * cirq.Y(q2) * cirq.Y(q3),
+        cirq.X(q0) * cirq.Y(q1) * cirq.Y(q2) * cirq.X(q3),
+        cirq.Z(q0) * cirq.Z(q1) * cirq.Z(q2) * cirq.Z(q3)
+    ]
+    group_max_settings_should_be = cirq.work.observables_to_settings(
+        group_max_obs_should_be, qubits)
+
+    assert list(grouped_settings.keys()) == list(group_max_settings_should_be)
```
balopat, line 48: It's interesting that you are using a list here - do we really care about the insertion order? Was that intentional?

cirq/work/observable_grouping_test.py
```diff
@@ -0,0 +1,54 @@
...
+        0.1711977489805745 * cirq.Z(q0), 0.17119774898057447 * cirq.Z(q1),
+        -0.2227859302428765 * cirq.Z(q2), -0.22278593024287646 * cirq.Z(q3),
+        0.16862219157249939 * cirq.Z(q0) * cirq.Z(q1),
+        0.04532220205777764 * cirq.Y(q0) * cirq.X(q1) * cirq.X(q2) * cirq.Y(q3),
+        -0.0453222020577776 * cirq.Y(q0) * cirq.Y(q1) * cirq.X(q2) * cirq.X(q3),
+        -0.0453222020577776 * cirq.X(q0) * cirq.X(q1) * cirq.Y(q2) * cirq.Y(q3),
+        0.04532220205777764 * cirq.X(q0) * cirq.Y(q1) * cirq.Y(q2) * cirq.X(q3),
+        0.12054482203290037 * cirq.Z(q0) * cirq.Z(q2), 0.16586702409067802 *
+        cirq.Z(q0) * cirq.Z(q3), 0.16586702409067802 * cirq.Z(q1) * cirq.Z(q2),
+        0.12054482203290037 * cirq.Z(q1) * cirq.Z(q3),
```
balopat, line 31: there is too much happening in this test - do we really need 8 elements for the ZZZZ group? Why are these particular coeffs interesting?

cirq/work/observable_grouping_test.py
```diff
@@ -0,0 +1,54 @@
...
+    ]
+    group_max_settings_should_be = cirq.work.observables_to_settings(
+        group_max_obs_should_be, qubits)
+
+    assert list(grouped_settings.keys()) == list(group_max_settings_should_be)
+    groups = list(grouped_settings.values())
+    assert len(groups[0]) == 1
+    assert len(groups[1]) == 1
+    assert len(groups[2]) == 1
+    assert len(groups[3]) == 1
```
balopat, line 53: can we assert which elements go to which group instead of just the size?

cirq/work/observable_grouping.py
```diff
@@ -0,0 +1,75 @@
...
+            trial_grouped_settings = simul_settings + [setting]
+            new_max_weight_state = _max_weight_state(
+                stg.init_state for stg in trial_grouped_settings)
+            new_max_weight_obs = _max_weight_observable(
+                stg.observable for stg in trial_grouped_settings)
+            # max_weight_xxx returns None if the set of xxx's aren't compatible,
+            # so the following conditional is True if setting can
+            # be inserted into the current group.
+            if (new_max_weight_state is not None and
+                    new_max_weight_obs is not None):
```
balopat, line 60: I usually prefer local variable names over extended documentation in comments: 

```suggestion
            compatible_init_state = new_max_weight_state is not None
            compatible_observable = new_max_weight_obs is not None
            can_be_inserted = (compatible_init_state and compatible_observable)
            if can_be_inserted:
```

cirq/work/observable_grouping.py
```diff
@@ -0,0 +1,75 @@
...
+from cirq.work.observable_settings import (InitObsSetting, _max_weight_state,
+                                           _max_weight_observable)
+
+if TYPE_CHECKING:
+    pass
+
+
+def group_settings_greedy(settings: Iterable[InitObsSetting]) \
+        -> Dict[InitObsSetting, List[InitObsSetting]]:
+    """
```
balopat, line 26: nit: no new line

cirq/work/observable_grouping.py
```diff
@@ -0,0 +1,75 @@
...
+            if (new_max_weight_state is not None and
+                    new_max_weight_obs is not None):
+                del grouped_settings[max_setting]
+                new_max_setting = InitObsSetting(new_max_weight_state,
+                                                 new_max_weight_obs)
+                grouped_settings[new_max_setting] = trial_grouped_settings
+                break
+
+        else:
+            # made it through entire dict without finding an existing group
```
balopat, line 68:
```suggestion
            # made it through entire dict without finding a compatible group, 
            # thus a new group needs to be created
```

balopat (changes requested):
This looks good, I added a bunch of nits

cirq/work/observable_grouping_test.py
mpharrigan: otherwise the `keys()` method returns an iterator so you can't use `==` in an assert

cirq/work/observable_grouping_test.py
mpharrigan: this is the hydrogen hamiltonian from the openfermion example

cirq/work/observable_grouping_test.py
mpharrigan: added a bunch of tests

cirq/work/observable_grouping_test.py
mpharrigan: changed test name to clarify

cirq/work/observable_grouping_test.py
mpharrigan: changed to `set()` comparison

cirq/work/observable_grouping_test.py
mpharrigan: might clutter the test too much. The simpler tests which I added now cover checking the group values as well, so maybe less necessary now

cirq/work/observable_grouping.py
mpharrigan: done

mpharrigan:
Your suggested change to the `None` handling broke mypy!

mpharrigan:
Also the formatter insists on a formatting that breaks pylint (line too long)

balopat (approved):
Awesome, thanks for the quick fixes!

PR #3373, opened 2020-09-30
Un-invert B gate

### LINKED ISSUE #2624: _BGate defined as complex conjugate of B gate
The B gate class in two_qubit_to_fsim.py appears to have been defined as the complex conjugate of the B gate as defined in the original reference (unless I'm making a sign error somewhere). It should be `XX(a, b)**-0.5, YY(a, b)**-0.25`,

https://github.com/quantumlib/Cirq/blob/2b88d34dc2c4dc5dc1c59d21810114112286c225/cirq/optimizers/two_qubit_to_fsim.py#L137

https://arxiv.org/abs/quant-ph/0312193

In a sense this doesn't really matter, since _BGate is only used internally to the module, and B and its conjugate are related by 1-qubit gates. But it might trip someone up down the line.

viathor:
Fixes #2624.

### DIFF
--- a/cirq/optimizers/two_qubit_to_fsim.py
+++ b/cirq/optimizers/two_qubit_to_fsim.py
@@ -169,8 +169,8 @@ class _BGate(ops.Gate):
     def _decompose_(self, qubits):
         a, b = qubits
         return [
-            ops.XX(a, b)**0.5,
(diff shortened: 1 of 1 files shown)


balopat (approved):
LGTM, thanks for the fix :)

PR #3378, opened 2020-10-01
Fix quimb version checking

### LINKED ISSUE #3263: cirq.contrib.quimb fails with quimb 1.2.0

balopat:
Making quimb version check more robust.

In some systems the version might not be parseable as (int,int), e.g `0+unknown`. 
This PR fixes these scenarios.

Fixes #3263.

### DIFF
--- a/cirq/contrib/quimb/state_vector.py
+++ b/cirq/contrib/quimb/state_vector.py
@@ -7,7 +7,16 @@ import quimb.tensor as qtn
 
 import cirq
 
-QUIMB_VERSION = tuple(int(x) for x in quimb.__version__.split('.'))
+
+def _get_quimb_version():
+    version = quimb.__version__
+    try:
+        return tuple(int(x) for x in version.split('.')), version
+    except:
+        return (0, 0), version
+
+
+QUIMB_VERSION = _get_quimb_version()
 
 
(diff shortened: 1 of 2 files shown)


cirq/contrib/quimb/state_vector.py
```diff
@@ -7,7 +7,17 @@
 
 import cirq
 
-QUIMB_VERSION = tuple(int(x) for x in quimb.__version__.split('.'))
+
+# coverage: ignore
+def _get_quimb_version():
```
dstrain115, line 12: Should we have a short docstring (specifically what it is returning)?

dstrain115 (approved):
Thanks for the quick fix!

cirq/contrib/quimb/state_vector.py
balopat: added!

mpharrigan:
Thanks!

Issue #3354, opened 2020-09-23
Physical Z length may be incorrect in the documentation


btchiaro:
**Description of the issue**

In the documentation the physical z is listed as having a typical length of 12 ns: https://github.com/quantumlib/Cirq/blob/master/docs/google/devices.md#physical-z-gates

I believe that the physical z is typically 10 ns long, with 5 ns pads before and after.

btchiaro:
@dstrain115

balopat:
Discussed on Cirq Cynque: 
 - This should be changed to 20ns + we should mention that this is subject to change 
 - when we can retrieve this from the device, we should have a snippet retrieving gate lengths

dstrain115:
Sure, I will take care of this shortly.

We can retrieve gate lengths, but the problem is that some gates have multiple different durations depending on parameters (z can be 0 or 20ns and Fsim can be 0, 12 or 32)

PR #3385, opened 2020-10-03
Fix duration documentation for Physical Z

dstrain115:
- Adjust duration of physical Z to 20ns.
- Add disclaimer and pointer to gate duration docs
- Fix other random indenting that sphinx is complaining about.

Fixes: #3354

### DIFF
--- a/cirq/experiments/cross_entropy_benchmarking.py
+++ b/cirq/experiments/cross_entropy_benchmarking.py
@@ -93,7 +93,7 @@ class CrossEntropyResult:
     May also include results from speckle purity benchmarking (SPB) performed
     concomitantly.
 
-    Attributes::
+    Attributes:
         data: A sequence of NamedTuples, each of which contains two fields:
             num_cycle: the circuit depth as the number of cycles, where
             a cycle consists of a layer of single-qubit gates followed
--- a/cirq/experiments/cross_entropy_benchmarking.py
+++ b/cirq/experiments/cross_entropy_benchmarking.py
@@ -93,7 +93,7 @@ class CrossEntropyResult:
     May also include results from speckle purity benchmarking (SPB) performed
     concomitantly.
 
-    Args:
+    Attributes::
         data: A sequence of NamedTuples, each of which contains two fields:
             num_cycle: the circuit depth as the number of cycles, where
             a cycle consists of a layer of single-qubit gates followed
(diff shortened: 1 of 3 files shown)


cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -93,19 +93,19 @@ class CrossEntropyResult:
     May also include results from speckle purity benchmarking (SPB) performed
     concomitantly.
 
-    Attributes:
+    Args:
```
viathor, line 96: This section describes the attributes of the dataclass, so "Args" looks confusing. Is there a tool that complains about this?

viathor (approved):

cirq/experiments/cross_entropy_benchmarking.py
dstrain115: It looks like the problems were with the multiple indenting, not the section title, so I changed it back to "Atteributes:".  
Thanks for the quick review.

cirq/ops/common_channels.py
```diff
@@ -356,12 +351,9 @@ def depolarize(p: float, n_qubits: int = 1) -> DepolarizingChannel:
...
     This channel evolves a density matrix via
 
-        $$
-        \rho \rightarrow (1 - p) \rho
-            + 1 / (4**n - 1) \sum _i P_i X P_i
-        $$
+    $ \rho \rightarrow (1 - p) \rho + 1 / (4**n - 1) \sum _i P_i X P_i $
 
-        where P_i are the 4**n - 1 Pauli gates (excluding the identity).
+    where P_i are the $4^n - 1$ Pauli gates (excluding the identity).
```
balopat, line 356:
```suggestion
    where $P_i$ are the $4^n - 1$ Pauli gates (excluding the identity).
```

cirq/experiments/cross_entropy_benchmarking.py
```diff
@@ -93,19 +93,19 @@ class CrossEntropyResult:
     May also include results from speckle purity benchmarking (SPB) performed
     concomitantly.
 
-    Attributes:
+    Attributes::
```
balopat, line 96:
```suggestion
    Attributes:
```

nit: double colon

balopat (approved):
LGTM - found two nits

cirq/experiments/cross_entropy_benchmarking.py
balopat: We could experiment with the [docparams and docstyle](https://github.com/PyCQA/pylint/blob/master/pylint/extensions/docparams.rst) pylint plugins.

Issue #3288, opened 2020-09-02
Support for Projectors


MichaelBroughton:
**Is your feature request related to a use case or problem? Please describe.**

This is related to expanding functionality in TFQ here ( https://github.com/tensorflow/quantum/issues/357 )

**Describe the solution you'd like**

It would be nice to support projectors in `cirq.PauliSum` so that more complex (and still valid) expressions in `cirq.PauliSum` can be compressed to a reasonable length. Maybe something like:

```python3
# Normal functionality
a = 5.0 * cirq.Z(q0) - 2.0 * cirq.X(q0)

# New functionality
a = 5.0 * cirq.Z(q0) - 2.0 * cirq.X(q0) + cirq.OneProjector(q0) * cirq.ZeroProjector(q1)
```

Doing this right now in Cirq does not scale well:

```python3
def one_proj(q):
  return (1 - cirq.Z(q)) / 2

def zero_proj(q):
  return (1 + cirq.Z(q)) / 2

a = one_proj(q0) * one_proj(q1)
print(len(a._linear_dict.keys()))  # => 4

a = one_proj(q0) * one_proj(q1) * one_proj(q2)
print(len(a._linear_dict.keys()))  # => 8

a = one_proj(q0) * one_proj(q1) * one_proj(q2) * one_proj(q3)
print(len(a._linear_dict.keys()))  # => 16

.
.
.
# Anything past 20 is already 1 million elements and not all that usable anymore.
```

**What is the urgency from your perspective for this issue? Is it blocking important work?**

P1 - I need this no later than the next release (end of quarter)

balopat:
Discussed on Cirq Cynq: 

- this would be a valuable addition
- probably not in PauliSum but its own type - let's iterate on the design here and then create a PR

MichaelBroughton:
Alrighty if we want to separate out this functionality, how do people like @viathor and @mpharrigan feel about something like this:

1st PR: Make a new module `cirq/ops/projectors.py`:

```python3
class StateProjector:
   """A projector onto a particular state and associated stuff.""""

```

2nd PR: In `cirq/ops/linear_combinations.py` we add:

```python3
class ProjectorPauliSum(cirq.PauliSum): # Anyone got a better name ?
  """Many similar features to PauliSum only now incorporating StateProjector support."""
```

viathor:
Regarding 1st PR: drop the "State", just "Projector", it's cooler?

[...]

PR #3386, opened 2020-10-03
KetBraSum

tonybruguier:
This could be the first PR towards closing issue #3288.


tonybruguier:
Hello,
Is this PR on the right overall direction? I'm happy to withdraw or start over if not.

viathor (commented):
Had a quick look and one thing I noticed is that the new class only allows to define projects onto one state in the computational basis. In particular, we cannot project onto the plus state in single qubit case or onto the span(|01>, |10>) subspace in two-qubit case.

Added @MichaelBroughton to comment on what we need for the TFQ use-case.

tonybruguier:
Sorry I had misunderstood the requirements. I changed the API to accept more than one projection basis and also allow vectors like |+>.

Happy to modify/fix further.

cirq/ops/projectors.py
```diff
@@ -0,0 +1,69 @@
...
+if TYPE_CHECKING:
+    import cirq
+
+
+@value.value_equality
+class Projector(raw_types.Gate):
+    """A non-unitary gate that projects onto a single qubit."""
+
+    def __init__(self,
+                 projection_basis: List[List[float]],
```
MichaelBroughton, line 17: Will this support numpy arrays ?

cirq/ops/projectors.py
```diff
@@ -0,0 +1,69 @@
...
+                 qid_shape: Tuple[int, ...] = (2,)):
+        """
+        Args:
+            projection_basis: a (2**num_qubits, p) matrix that lists the
+                proection vectors.
```
MichaelBroughton, line 22: Nit: proection -> projection

cirq/ops/projectors.py
```diff
@@ -0,0 +1,69 @@
...
+                2 (qubit).
+
+        Raises:
+            ValueError: If the basis vector is empty.
+        """
+        if np.prod(qid_shape) != np.array(projection_basis).shape[1]:
+            raise ValueError(
+                "Invalid shape " +
+                f"{np.array(projection_basis).shape} for qid_shape {qid_shape}")
+        self._projection_basis = projection_basis
```
MichaelBroughton, line 33: Might want to store this as a numpy array since most comparisons and logic involving it appear to be using `np.array(self._projection_basis)` right away. Thoughts ?

MichaelBroughton (commented):
This looks great, before getting into a detailed review, I am curious how easily this could be extended to do things like:

```python
q = cirq.GridQubit(0, 0)
my_paulisum = 0.5 * cirq.X(q) * cirq.Z(q) + 5 * cirq.Y(q)
my_paulisum_and_projector = my_paulisum + cirq.Projector([[1,0]])  # This maintains a compact representation.
```
This was an important role for projectors when we first opened the issue :)

tonybruguier:
> This looks great, before getting into a detailed review, I am curious how easily this could be extended to do things like:
> 
> ```python
> q = cirq.GridQubit(0, 0)
> my_paulisum = 0.5 * cirq.X(q) * cirq.Z(q) + 5 * cirq.Y(q)
> my_paulisum_and_projector = my_paulisum + cirq.Projector([[1,0]])  # This maintains a compact representation.
> ```
> 
> This was an important role for projectors when we first opened the issue :)

I do not know yet (sorry, I'm still learning). In the original issue #3288 it was discussed to do in two PRs, so I was planning to wait for that. However, it would be nice for @viathor to comment. I'm happy to make any change, either here or in a subsequent PR.

cirq/ops/projectors.py
tonybruguier: Done

cirq/ops/projectors.py
tonybruguier: It did not, but I modified the code so that it does now. Also added a test.

cirq/ops/projectors.py
tonybruguier: Done

tonybruguier:
> This looks great, before getting into a detailed review, I am curious how easily this could be extended to do things like:
> 
> ```python
> q = cirq.GridQubit(0, 0)
> my_paulisum = 0.5 * cirq.X(q) * cirq.Z(q) + 5 * cirq.Y(q)
> my_paulisum_and_projector = my_paulisum + cirq.Projector([[1,0]])  # This maintains a compact representation.
> ```
> 
> This was an important role for projectors when we first opened the issue :)

I do not know yet (sorry, I'm still learning). In the original issue #3288 it was discussed to do in two PRs, so I was planning to wait for that. However, it would be nice for @viathor to comment. I'm happy to make any change, either here or in a subsequent PR.

For the other comments, I tried to address them so PTAL.

cirq/ops/projectors.py
```diff
@@ -0,0 +1,75 @@
...
+        return False
+
+    def _is_parameterized_(self) -> bool:
+        return False
+
+    def _channel_(self):
+        A = self._projection_basis
+        AH = np.transpose(np.conjugate(A))
+        result = AH * np.linalg.inv(np.matmul(A, AH)) * A
+        return (result,)
```
MichaelBroughton, line 57: Does this break if I give an overcomplete basis ?
```python
A = [[1,0], [0,1], [1,1]]
```
Should we care about such a case @viathor ?

cirq/ops/projectors_test.py
```diff
@@ -0,0 +1,79 @@
...
+                               ([[0.0, 0.0], [0.0, 1.0]],))
+
+
+def test_projector_from_np_array():
+    zero_projector = cirq.Projector(np.array([[1.0, 0.0]]))
+    np.testing.assert_allclose(cirq.channel(zero_projector),
+                               ([[1.0, 0.0], [0.0, 0.0]],))
+
+
+def test_projector_plus():
```
MichaelBroughton, line 26: Could we throw in one more complicated test ? Maybe one where you need to follow the math instead of just eyeballing it :)

cirq/ops/projectors.py
```diff
@@ -0,0 +1,75 @@
...
+    def _qid_shape_(self) -> Tuple[int, ...]:
+        return self._qid_shape
+
+    def _has_unitary_(self) -> bool:
+        return False
+
+    def _is_parameterized_(self) -> bool:
+        return False
+
+    def _channel_(self):
```
MichaelBroughton, line 53: Nit: Type annotate with: ` -> Iterable[np.ndarray]:`

cirq/ops/projectors.py
```diff
@@ -0,0 +1,75 @@
...
+        A = self._projection_basis
+        AH = np.transpose(np.conjugate(A))
+        result = AH * np.linalg.inv(np.matmul(A, AH)) * A
+        return (result,)
+
+    def _has_channel_(self):
+        return True
+
+    def __repr__(self):
```
MichaelBroughton, line 62: Nit: type annotations.

MichaelBroughton (commented):
Left some more comments. This looks good to me, but I will delegate to @viathor for final say on things.

cirq/ops/projectors.py
```diff
@@ -0,0 +1,75 @@
...
+            qid_shape: Specifies the dimension of the projection. The default is
+                2 (qubit).
+
+        Raises:
+            ValueError: If the basis vector is empty.
+        """
+        if type(projection_basis) == np.ndarray:
+            projection_array = cast(np.ndarray, projection_basis)
+        else:
+            projection_array = np.array(projection_basis)
```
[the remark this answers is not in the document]
tonybruguier: Done.

cirq/ops/projectors.py
```diff
@@ -0,0 +1,75 @@
...
+from cirq import value
+from cirq.ops import raw_types
+
+if TYPE_CHECKING:
+    import cirq
+
+
+@value.value_equality
+class Projector(raw_types.Gate):
+    """A non-unitary gate that projects onto a single qubit."""
```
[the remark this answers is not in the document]
tonybruguier: Done

cirq/ops/projectors.py
tonybruguier: Yes it does break. I guess We could do an SVD instead, but maybe it's better to let the user fix it?

Leaving unresolved since @viathor hasn't chimed in.

Regardless, I added a test for it.

cirq/ops/projectors_test.py
tonybruguier: Thank you so much for that comment! I had mistakenly used * instead of np.matmul and I was computing the wrong answer. Thanks to you suggestion for a more advanced test, I caught it.

The test is [1, 0] and [1, 1] which should result in the identity projection.

cirq/ops/projectors.py
tonybruguier: Done.

cirq/ops/projectors.py
```diff
@@ -0,0 +1,75 @@
...
+
+        if np.prod(qid_shape) != projection_array.shape[1]:
+            raise ValueError(
+                "Invalid shape " +
+                f"{np.array(projection_array).shape} for qid_shape {qid_shape}")
+        self._projection_basis = projection_array
+        self._qid_shape = qid_shape
+
+    def _projection_basis_(self):
+        return self._projection_basis.tolist()
```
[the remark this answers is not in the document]
tonybruguier: Did both.

cirq/ops/projectors.py
tonybruguier: Done.

cirq/ops/projectors.py
```diff
@@ -0,0 +1,75 @@
...
+
+        Raises:
+            ValueError: If the basis vector is empty.
+        """
+        if type(projection_basis) == np.ndarray:
+            projection_array = cast(np.ndarray, projection_basis)
+        else:
+            projection_array = np.array(projection_basis)
+
+        if np.prod(qid_shape) != projection_array.shape[1]:
```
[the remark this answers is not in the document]
tonybruguier: I'm not sure what you mean by 'rank' here. Do you mean that we should check that the basis spans a 2D space (i.e. the rank of the matrix is 2) or do you mean that the input array has shape[0] == 2?

In either case, I don't understand why it would be a requirement (not contesting, but rather not understanding).

Leaving unresolved.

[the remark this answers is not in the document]
MichaelBroughton: Sorry I wasn’t all that clear. I was just thinking in terms of error reporting if the user accidentally called the function with  a 1D or 3D array they might get a cryptic failure is all and should we catch that case or not ?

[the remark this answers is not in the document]
tonybruguier: Oh, that makes a lot of sense. Done + added a test. Thank you.

tonybruguier:
Thanks. PTAL.

viathor (changes requested):
Sorry for the delay. See comments below.

cirq/ops/projectors.py
```diff
@@ -0,0 +1,75 @@
...
+    def _has_unitary_(self) -> bool:
+        return False
+
+    def _is_parameterized_(self) -> bool:
+        return False
+
+    def _channel_(self) -> Iterable[np.ndarray]:
+        A = self._projection_basis
+        AH = np.transpose(np.conjugate(A))
+        result = np.matmul(np.matmul(AH, np.linalg.inv(np.matmul(A, AH))), A)
```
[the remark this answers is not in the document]
tonybruguier: I'm not quite sure I follow. I think in my notation:
`
A = V^\dagger
`

I used this convention since it's easier to feed inputs. So in short, my formula is:
`
V (V^\dagger V)^-1 V^\dagger
`

In case the basis is made of orthogonal vectors, it does indeed reduce to your formula:
`
V^\dagger V = I_p
`

But the formula that I have is more general, because I don't think your reasoning works with non-orthogonal vector.

If the basis has p=d vectors, then yes the formula reduces to the identity, but that's kind of expected, right? The simplification that you have does work, but in the general case A/V is not square and thus cannot be inverted, right?

I just pulled the formula from Wikipedia, by the way:
https://en.wikipedia.org/wiki/Projection_(linear_algebra)#Formulas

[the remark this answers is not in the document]
tonybruguier: Done.

cirq/ops/projectors.py
```diff
@@ -0,0 +1,75 @@
...
+
+        Raises:
+            ValueError: If the basis vector is empty.
+        """
+        projection_array = np.asarray(projection_basis)
+
+        if len(projection_array.shape) != 2:
+            raise ValueError('The input projection_basis must be a 2D array')
+
+        if np.prod(qid_shape) != projection_array.shape[1]:
```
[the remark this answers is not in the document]
tonybruguier: Don't I already have the test in there?
`
test_projector_overcomplete_basis
`

It throws an LinAlgError.

cirq/ops/projectors.py
```diff
@@ -0,0 +1,75 @@
...
+from cirq import value
+from cirq.ops import raw_types
+
+if TYPE_CHECKING:
+    import cirq
+
+
+@value.value_equality
+class Projector(raw_types.Gate):
+    """A non-unitary gate that projects onto one or more qubits."""
```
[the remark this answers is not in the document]
tonybruguier: Added some more text.

cirq/ops/projectors.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def __init__(self,
+                 projection_basis: Union[List[List[float]], np.ndarray],
+                 qid_shape: Tuple[int, ...] = (2,)):
+        """
+        Args:
+            projection_basis: a (2**num_qubits, p) matrix that lists the
+                projection vectors.
```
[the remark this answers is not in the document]
tonybruguier: Done.

tonybruguier:
Thanks for the review, @viathor 

I replied to your comments but I left some things unresolved because I feel that I was pushing back a little bit. I think the crux of the discussion is whether we want require the input basis to be orthogonal or not. In my original implementation, I was *not* assuming that it was the case, hence the formula that I was using.

I did keep my original formula since I think it's a tad more general, and has built-in sanity checks. Still, I added an optional parameter that allows to check that input basis is orthogonal.

Either way, this is a very weak preference on my part and happy to restrict to only bases that are orthogonal. Just let me know.

cirq/ops/projectors.py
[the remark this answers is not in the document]
viathor: You're right! Your algorithm is more general than the suggestion, so let's keep as is (I have a different comment though I'll add separately).

cirq/ops/projectors.py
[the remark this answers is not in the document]
viathor: Indeed.

cirq/ops/projectors.py
```diff
@@ -0,0 +1,95 @@
...
+    """A non-unitary gate that projects onto one or more qubits.
+
+    The input is a matrix representing the basis of the space we project onto.
+    The basis vectors need not be orthogonal, but they must be independent (i.e.
+    the matrix has full rank).
+
+    For example, if you want to project on |0>, you would provide the basis
+    [[1, 0]]. To project onto |10>, you would provide the basis [[0, 0, 1, 0]].
+    If you want to project on the space spanned by |10> and |11>, you could
+    provide the basis [[0, 0, 1, 0], [0, 0, 1, 0]].
```
viathor, line 23: `[[0, 0, 1, 0], [0, 0, 0, 1]]`

cirq/ops/projectors.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+
+    def __init__(self,
+                 projection_basis: Union[List[List[float]], np.ndarray],
+                 qid_shape: Tuple[int, ...] = (2,),
+                 enfore_orthogonal_basis: bool = False):
+        """
+        Args:
+            projection_basis: a (2**num_qubits, p) matrix that lists the
```
viathor, line 32: The class docstring above and the code in `_channel_` below suggest that `projection_basis` should be pxd (where d=2^n), but here the comment says dxp. IOW, the class docstring and `_channel_` code suggest that one should provide the basis vectors of the projection subspace as *rows* of `projection_basis`, but this comment suggests they should be *columns*.

cirq/ops/projectors.py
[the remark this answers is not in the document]
viathor: Actually, it'd be better to check whether projection basis is linearly independent in the constructor rather than wait until an error is triggered in `_channel_`.

cirq/ops/projectors.py
```diff
@@ -0,0 +1,95 @@
...
+        return False
+
+    def _is_parameterized_(self) -> bool:
+        return False
+
+    def _channel_(self) -> Iterable[np.ndarray]:
+        A = self._projection_basis
+        AH = np.transpose(np.conjugate(A))
+        result = AH @ np.linalg.inv(A @ AH) @ A
+        return (result,)
```
viathor, line 77: nit: An important intermediate value in your algorithm is the pseudoinverse matrix. You can make the code a little more readable by making this explicit:

```python
A = self._projection_basis
AH = A.T.conj()
pseudoinverse = AH @ np.linalg.inv(A @ AH)
return (pseudoinverse @ A,)
```

viathor (commented):
I misread your code. Sorry. You're right that your algorithm is correct and more general than what I suggested. Let's keep the algorithm as is.

I have a few more minor comments.

cirq/ops/projectors.py
tonybruguier: Done.

cirq/ops/projectors.py
tonybruguier: Thanks for catching it. It is indeed that each vector is stored in a row. I changed the description and expanded the comment. I tried to look at the rest of the code but I think it's consistent now.

cirq/ops/projectors.py
tonybruguier: Done.

tonybruguier:
Thanks. PTAL.

cirq/ops/projectors.py
```diff
@@ -0,0 +1,98 @@
...
+        Raises:
+            ValueError: If the basis vector is empty.
+        """
+        projection_array = np.asarray(projection_basis)
+
+        if len(projection_array.shape) != 2:
+            raise ValueError('The input projection_basis must be a 2D array')
+
+        if enfore_orthogonal_basis:
+            B = projection_array @ np.transpose(np.conjugate(projection_array))
```
viathor, line 50:
```python
B = projection_array @ projection_array.T.conj()
```

cirq/ops/projectors.py
```diff
@@ -0,0 +1,98 @@
...
+    For example, if you want to project on |0>, you would provide the basis
+    [[1, 0]]. To project onto |10>, you would provide the basis [[0, 0, 1, 0]].
+    If you want to project on the space spanned by |10> and |11>, you could
+    provide the basis [[0, 0, 1, 0], [0, 0, 0, 1]].
+    """
+
+    def __init__(self,
+                 projection_basis: Union[List[List[float]], np.ndarray],
+                 qid_shape: Tuple[int, ...] = (2,),
+                 enfore_orthogonal_basis: bool = False):
```
viathor, line 29: typo: missing "c" in "enforce"

viathor, line 29: nit: Since you're actually enforcing orthonormality, not just orthogonality (IOW, you're enforcing orthogonality and unit norm), this should be named "enforce_orthonormal_basis".

cirq/ops/projectors.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _has_unitary_(self) -> bool:
+        return False
+
+    def _is_parameterized_(self) -> bool:
+        return False
+
+    def _channel_(self) -> Iterable[np.ndarray]:
+        A = self._projection_basis
+        AH = np.transpose(np.conjugate(A))
```
viathor, line 78:
```python
AH = A.T.conj()
```

(Optional: You can also inline this in two places below where you use `AH`. Up to you.)

cirq/ops/projectors.py
```diff
@@ -0,0 +1,98 @@
...
+
+from cirq import value
+from cirq.ops import raw_types
+
+if TYPE_CHECKING:
+    import cirq
+
+
+@value.value_equality
+class Projector(raw_types.Gate):
```
viathor, line 13: What is this gate gonna look like in a circuit? It may be a good idea to override `_circuit_diagram_info_`.

viathor (commented):
This is very nice, thank you! I have a few minor nits and two more things we need to consider before merging this:
1. Does this address the original use-case? I'm not familiar with it, so I'd like @MichaelBroughton input.
2. We should make sure that the new gate looks well when rendered in a circuit, see comment below.

cirq/ops/projectors.py
tonybruguier: Done.

cirq/ops/projectors.py
tonybruguier: Done.

cirq/ops/projectors.py
tonybruguier: Done.

cirq/ops/projectors.py
tonybruguier: Done.

cirq/ops/projectors.py
tonybruguier: Done. Also added a test.

tonybruguier:
Thanks. PTAL.

viathor (approved):
Thank you!

cirq/ops/projectors.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def _json_dict_(self) -> Dict[str, Any]:
+        return {
+            'cirq_type': self.__class__.__name__,
+            'projection_basis': self._projection_basis.tolist(),
+            'qid_shape': self._qid_shape,
+        }
+
+    def _value_equality_values_(self) -> Any:
+        return self._projection_basis.tolist(), self._qid_shape
```
balopat, line 103: This needs to be hashable

balopat (changes requested):
1. Value equality protocol is incorrect. 
2. Where are we with regards to @MichaelBroughton's requirements? 
I.e. I'd like to understand the plan for supporting sums and products of projectors and paulis with an efficient representation.

cirq/ops/projectors_test.py
```diff
@@ -0,0 +1,114 @@
...
+import numpy as np
+import pytest
+
+import cirq
+
+
+def test_projector_circuit_diagram():
+    q = cirq.NamedQubit('q')
+    projector = cirq.Projector([[1.0, 0.0]])
+    cirq.testing.assert_has_diagram(cirq.Circuit(projector(q)),
```
balopat, line 12: How do two-qubit projectors look like?

cirq/ops/projectors.py
```diff
@@ -0,0 +1,103 @@
...
+        Raises:
+            ValueError: If the basis vector is empty.
+        """
+        projection_array = np.asarray(projection_basis)
+
+        if len(projection_array.shape) != 2:
+            raise ValueError('The input projection_basis must be a 2D array')
+
+        if enforce_orthonormal_basis:
+            B = projection_array @ projection_array.T.conj()
```
balopat, line 51: You apply conjugation here and there is no testing of it - only real valued basis vectors are tested

balopat: Do we really need the conjugation here?

cirq/ops/projectors.py
```diff
@@ -0,0 +1,103 @@
...
+            if not np.allclose(B, np.eye(projection_array.shape[0])):
+                raise ValueError('The basis must be orthogonal')
```
balopat, line 53:
```suggestion
                raise ValueError('The basis must be orthonormal')
```

cirq/ops/projectors_test.py
```diff
@@ -0,0 +1,114 @@
...
+    with pytest.raises(ValueError,
+                       match="Vectors in basis must be linearly independent"):
+        cirq.Projector([[1.0, 0.0], [0.0, 1.0], [1.0, 1.0]])
+
+
+def test_projector_non_orthogonal_basis():
+    cirq.Projector([[1.0, 0.0]], enforce_orthonormal_basis=True)
+    cirq.Projector([[1.0, 0.0], [0.0, 1.0]], enforce_orthonormal_basis=True)
+
+    with pytest.raises(ValueError, match="The basis must be orthogonal"):
```
balopat, line 60: I would add another test case: to test for the bases being of norm 1.  
You are testing against both norm 1 and orthogonality here, maybe it's worthwhile separating out the two cases.

cirq/ops/projectors.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def _has_unitary_(self) -> bool:
+        return False
+
+    def _is_parameterized_(self) -> bool:
+        return False
+
+    def _channel_(self) -> Iterable[np.ndarray]:
+        A = self._projection_basis
+        pseudoinverse = A.T.conj() @ np.linalg.inv(A @ A.T.conj())
```
balopat, line 79: I'm confused: `A @ A.T.conj()` (again conj() seems unnecessary) is identity, isn't it? We are ensuring that in the constructor. In this case pseudoinverse is `A.T.conj()`?

balopat:
Also, how does this whole thing relate to the projectors we already have? 
E.g. `p01 = (cirq.KET_ZERO(q0) * cirq.KET_ONE(q1)).projector()` gives you a projector to `|01>`, although the matrix of the projector. Can we reuse / combine things somehow? 

Also they don't agree on the matrices: 

```
    mx = (cirq.KET_IMAG(a) * cirq.KET_IMAG(b)).projector()
    ii_proj = cirq.Projector([[.5, .5j, .5j, -.5]], qid_shape=(2, 2,))

    np.testing.assert_allclose(mx, cirq.channel(ii_proj))
```

fails.

cirq/ops/projectors_test.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def test_projector_bad_rank():
+    with pytest.raises(ValueError,
+                       match="The input projection_basis must be a 2D array"):
+        cirq.Projector(np.array([1.0, 0.0]))
+
+
+def test_projector_plus():
+    plus_projector = cirq.Projector([[1.0, 1.0]])
```
balopat, line 44: This should fail if we are checking for orthonormality...
This base vector does not have norm 1!! 
That would be `[1/np.sqrt(2), 1/np.sqrt(2)]`, right?

cirq/ops/projectors.py
viathor: Remember that A is not a square matrix. It has 2^n columns, but only p rows where p < 2^n is the dimension of the subspace we're projecting onto.

(I got confused about this, too. Suggested to name the intermediate result "pseudoinverse" to improve readability. Other ideas welcome!)

cirq/ops/projectors.py
viathor: We expect (i, j) entry of B to be the inner product of the i-th and j-th row of `projection_array`. Since elements of `projection_array` can be complex(*) we need conjugation, because without it the elements of B aren't an inner product(**).

---
(*) For example the user may want to specify a projector onto one of the eigenvectors of Pauli Y operator, e.g. `projection_array=np.array([[1, 1j]])/np.sqrt(2)`.
(**) One of the requirements on the inner product is that <a, a> = 0 if and only if a = 0, but without conjugation <[1, 1j], [1, 1j]> = 0.

cirq/ops/projectors.py
balopat: Oh I see now the orthonormality is not enforced always :) I'm getting there slowly.

cirq/ops/projectors_test.py
balopat: Ignore this one - I see that orthonormality is not always enforced.

cirq/ops/projectors.py
balopat: Agreed - thanks for the explanation @viathor, I realized it right at the same time as well what's happening here :) 
However we do want to test it so I'm going to keep this thread open.

cirq/ops/projectors.py
balopat: I'm wondering if `numpy.linalg.pinv` would be better or not?

balopat: I'm also confused about using our basis vectors as rows. Most places I read the projector to the _range_ (i.e. column space) of the matrix A is calculated as `A @ left_inverse` - however here we calculate the `right_inverse @ A` - which seems to be the range of `A.conj().T` - which is not what we want - and might explain the difference the cirq.KET_IMAGE discrepancy as well.

balopat: Here's my suggestion: 

1. Make basis vectors column vectors 
2. Calculate left inverse using np.linalg.pinv
3. Add a comment explaining exactly what's happening: "As A's column vectors are linearly independent, A@A.conj().T is invertible, and we can calculate the left pseudoinverse. A @ left_inverse is then the orthogonal projector to the range of A (i.e. the subspace spanned by A's column vectors)

WDYT?

balopat: In short this is the correct solution if we don't want to change the basis vector representation altogether.

```
       # make rows into columns 
        A = self._projection_basis.T
        # left pinverse 
        pseudoinverse = np.linalg.pinv(A)
        # projector to the range of A 
        return (A @ pseudoinverse,)
```

cirq/ops/projectors.py
tonybruguier: I added two entries in the test: one that would (incorrectly) fail if the conjugation were missing and another that would (incorrectly) succeed if the conjugation were missing.

Given your remarks, I also took the liberty of adding a test in a different location. Maybe we want to make sure that even if we don't check for orthogonality, the computation of the projector is also working OK with complex number (I didn't have a test case, but now I have one).

cirq/ops/projectors.py
tonybruguier: Added a test that revealed the failure, then modified the code to make it pass.

cirq/ops/projectors_test.py
tonybruguier: Added a test for 2 qubits, code was crashing, fixed the code.

cirq/ops/projectors_test.py
tonybruguier: Done.

cirq/ops/projectors.py
tonybruguier: Done.

Also added the consistency check you had suggested below.

tonybruguier:
Thanks. PTAL.

cirq/ops/projectors_test.py
```diff
@@ -0,0 +1,166 @@
...
+
+def test_projector2d_circuit_diagram():
+    q1 = cirq.NamedQubit('q1')
+    q2 = cirq.NamedQubit('q2')
+    projector = cirq.Projector([[1.0, 0.0, 0.0, 0.0]], qid_shape=(
+        2,
+        2,
+    ))
+    cirq.testing.assert_has_diagram(cirq.Circuit(projector(
+        q1, q2)), ("q1: ───Proj([[1.0, 0.0, 0.0, 0.0]])───\n" + "       │\n" +
```
balopat, line 29: typically we use the multiline string for this, so it's more readable, i.e.
```
"""
q1: ───Proj([[1.0, 0.0, 0.0, 0.0]])───
                 |
q2: ───Proj────────────"""[1:]
```
or similar, check the code there are ton of examples. 

Also, I think we can do better than this, in case the vectors are easy to write in the dirac formalism, then we should write them out that way (using `cirq.dirac_notation`), e.g. this particular example could be something like this - also P would be shorter...though it might be confused with a phase gate?  

```
q1: ───P(|00⟩)──
                  |
q2: ───P──
```

What do you think?

cirq/ops/projectors.py
```diff
@@ -0,0 +1,107 @@
...
+        return self._qid_shape
+
+    def _has_unitary_(self) -> bool:
+        return False
+
+    def _is_parameterized_(self) -> bool:
+        return False
+
+    def _channel_(self) -> Iterable[np.ndarray]:
+        # Make rows into columns:
```
balopat, line 78:
```suggestion
        # Make rows into columns
```

cirq/ops/projectors.py
```diff
@@ -0,0 +1,107 @@
...
+        A = self._projection_basis.T
+        # Left pseudo-inverse:
```
balopat, line 80:
```suggestion
        # Left pseudo-inverse
```

cirq/ops/projectors.py
```diff
@@ -0,0 +1,107 @@
...
+        pseudoinverse = np.linalg.pinv(A)
+        # Projector to the range of A:
```
balopat, line 82:
```suggestion
        # Projector to the range (column space) of A
```

cirq/ops/projectors.py
```diff
@@ -0,0 +1,107 @@
...
+
+@value.value_equality
+class Projector(raw_types.Gate):
+    """A non-unitary gate that projects onto one or more qubits.
+
+    The input is a matrix representing the basis of the space we project onto.
+    The basis vectors need not be orthogonal, but they must be independent (i.e.
+    the matrix has full rank).
+
+    For example, if you want to project on |0>, you would provide the basis
```
balopat, line 21:
```suggestion
    For example, if you want to project on |0⟩, you would provide the basis
```

cirq/ops/projectors.py
```diff
@@ -0,0 +1,107 @@
...
+    [[1, 0]]. To project onto |10>, you would provide the basis [[0, 0, 1, 0]].
```
balopat, line 22:
```suggestion
    [[1, 0]]. To project onto |10⟩, you would provide the basis [[0, 0, 1, 0]].
```

cirq/ops/projectors.py
```diff
@@ -0,0 +1,107 @@
...
+    If you want to project on the space spanned by |10> and |11>, you could
```
balopat, line 23:
```suggestion
    If you want to project on the space spanned by |10⟩ and |11⟩, you could
```

cirq/ops/projectors_test.py
tonybruguier: Done

tonybruguier:
Thanks. PTAL.

tonybruguier:
Friendly ping. I think I have address your comments, but I'm happy to do more, of course.

tonybruguier:
Hello,
Is there any more work for this PR? I'm happy to do more, as always, but if not, maybe we could merge?

balopat:
> Hello,
> Is there any more work for this PR? I'm happy to do more, as always, but if not, maybe we could merge?

The only thing outstanding is a confirmation that this at all will solve the original problem of exponentially growing terms. Will it? I didn't have time to work it out - would love to see some design thoughts that we are on the right track.

tonybruguier:
@MichaelBroughton 
I think the question is whether this PR solves this problem:
https://github.com/quantumlib/Cirq/issues/3288#issuecomment-685986810

I am not sure I understand the question actually, so if you have the time, could you comment?

Everyone:
I am not in a hurry at all. I just don't want to forget about this, that's all. Thanks for all the help so far.

balopat:
@tonybruguier thank you for the hard work on this one. I think we are close but we still need some changes. I caught up with Mike and Adam, and captured some thoughts as a design proposal that was missing. As soon as we agree on the direction, we can continue fixing this PR up and then we'll have clarity for the next steps too. Thank you!

cirq/ops/projectors.py
```diff
@@ -26,7 +26,7 @@ class Projector(raw_types.Gate):
     """
 
     def __init__(self,
-                 projection_basis: Union[List[List[float]], np.ndarray],
+                 projection_basis: Union[STATE_VECTOR_LIKE],
```
balopat, line 29: These should be a sequence
```suggestion
                 projection_basis: Sequence[STATE_VECTOR_LIKE],
```

tonybruguier: Done.

tonybruguier:
Hi,
I'm still working on your comments. Feel free to comment or skip for now.

tonybruguier:
Hi,
Thanks for the discussions. I didn't understand everything, but I think one of the requests was to be able to add an arbitrary matrix, which I just did. PTAL?

balopat:
The conversation was more about introducing a different general representation for operators that allows for a compact representation of projectors, the "butterfly operators", or the dirac notation for matrices. A projector to the |0> state can be written as `|0><0|`. This is basically a sparse representation of the matrices. Or a projector to any particular state `|psi>` similarly works like that, i.e. `|psi><psi|`. So the thought was, that instead of introducing a special construct for projectors, we could introduce the butterfly operators and somehow have that work well with `PauliSum`s. Unfortunately this would mean scratching the current direction and re-starting the design :/

tonybruguier:
Sorry about misunderstanding. I reverted it (although kept the factoring out of the matrices.

tonybruguier:
Hi,
I have this RFC here:
https://tinyurl.com/ketbra

It's still under discussion, but I had some time, so I decided to implement it here. Of course, this is not mean I'm pushing for the current design and I'm more than happy to revise the RFC and the PR further. Thanks for all the help so far.

tonybruguier:
Hi,

I've added the KetBra object both in the RFC and the current version of the code. Not sure whether you prefer to comment on the RFC or this PR, but either way feedback is welcome when you all have some time.

95-martin-orion:
Paging @balopat for re-review: what is the plan for this PR?

viathor:
Friendly ping, @balopat, this probably needs your attention :-)

balopat:
Yes, I need to sit down and think this through again. 
Issues: 
- naming: ketbrasum doesn't sit well 
- I still don't have any answers around how this would really work when we mix operator representations as @MichaelBroughton wants to I.e. `cirq.X(q0) + cirq.I(q2) + cirq.Z(q1) * |001><001|` - do we support products? or just sums? I feel like instead of trying to get in something incremental we should prototype the final solution, play with it until we like the design, and then based on that get either incremental PRs out or just the whole thing in one go.

95-martin-orion:
@balopat, should we bring this (or the related RFC: https://tinyurl.com/ketbra) to either the weekly sync or the internal design discussion for re-review? It's been open for a long time now, but seems to be chronically stuck behind design considerations.

balopat:
> @balopat, should we bring this (or the related RFC: https://tinyurl.com/ketbra) to either the weekly sync or the internal design discussion for re-review? It's been open for a long time now, but seems to be chronically stuck behind design considerations.

As @MichaelBroughton requested the projectors originally and is taking over the helm, I actually asked him to have a look into this and make a call.

tonybruguier:
I added the entry for next week's meeting. I would like to have a net positive contribution (more work produced than help received), so happy to continue working on it, happy to hand it off to someone else, happy to drop it.

MichaelBroughton (commented):
I think unfortunately we may have had too many cooks in the kitchen on this one and became a little sidetracked on things. I think this might be worth starting over with a new PR chain with the following things:

For  TFQ, we don't need the ability to do arbitrary KetBraSums we just need the ability to do projectors with computational basis states and nothing more (in fact general KetBraSums would be too hard for us to handle). The nice things about these projectors is that their expectation value is also very easy to evaluate on a real device by just counting the number of times you find the projector bit-string in your measurements. 

With all that being said I think we should revert back to something like this where `KetBras` -> `ProjectorStrings` (like PauliStrings) and `KetBraSum` -> `ProjectorSum` (like PauliSum), so that we can keep hardware compatibility, with potential future integrations to things like Observable measurements, cirq.ProductStates, as well as the original behavior we wanted in TFQ. If we need more general outer product sums I think we should revisit at a later date. What do you think @viathor , @balopat ?

tonybruguier:
@MichaelBroughton I am withdrawing in favor of #4331

Issue #3368, opened 2020-09-30
Use pydata_google_auth instead of colab for Engine auth


balopat:
# Problems
Using google-colab as the dependency is problematic because: 
 - On jupyter locally we have issues because of exactly google-colab: #3318 
 - it has all sorts of issues with python 3.8 
 - it's heavyweight

# Recommendation 

I'd suggest using instead pydata_google_auth for auth. It is light weight and works on both colab (installed by default) and jupyter notebooks (we can install it as part of the notebook).

```python
# The Google Cloud Project id to use.
project_id = 'your project' #@param {type:"string"}

def setup_auth():
  """Runs the user through the Google OAuth process """
  try: 
     import pydata_google_auth
  except: 
     !pip install pydata_google_auth
  import pydata_google_auth
  return pydata_google_auth.get_user_credentials(
    ['https://www.googleapis.com/auth/cloud-platform'],
  )


print("Getting OAuth2 credentials.")
print("Press enter after entering the verification code.")
credentials = setup_auth()
print("Authentication complete.")

try: 
    import cirq 
except:     
    !pip install cirq

print("Cirq is installed.")

import cirq

# Create an Engine object to use, providing the project id and the args
# used for authentication (produced by running the authentication above).
engine = cirq.google.Engine(project_id=project_id, service_args = {'credentials': credentials} )
```

# Caveat

This creates a global var called "credentials" which is an OAuth2 token. This might be undesirable, we can try to hide it and work with the `GOOGLE_APPLICATION_CREDENTIALS` env var - which is a bit better as the user's code won't have direct access to the creds - it needs to read the creds file which the env var is pointing to. 

_Originally posted by @balopat in https://github.com/quantumlib/Cirq/pull/3365#r497614032_

balopat:
cc @dstrain115 wdyt?

dstrain115:
I think this is fine.  It is a BSD-3 clause license, which I think is okay, but it might be worth checking out to verify that this is an acceptable license.

balopat:
The problem I have with pydata_google_auth is that it uses PyData for the consent screen which is unfamiliar for the users. 
![image](https://user-images.githubusercontent.com/635073/97620988-8124b500-19df-11eb-8c06-203ca9a8ae9f.png)

Colab uses Google Cloud SDK which is also different from the Quantum Computing Service. @dstrain115 have we ever thought of creating an oauth consent page for QCS? I guess I can look into that.

dstrain115:
[...]

PR #3393, opened 2020-10-06
Docs frontmatter and some cleanup

lamberta:
Add frontmatter, buttons, and clean up some notebook structure. Not much for content change.
Using some placeholder URLs for some buttons so will need to make a pass later but that's a search/replace vs adding new notebook cells (rather tedious). This gets the structure in place.

Normalized notebooks with [nbfmt](https://github.com/tensorflow/docs/tree/master/tools/tensorflow_docs/tools):

```
$ python3 -m pip install -U --user git+https://github.com/tensorflow/docs

$ python3 -m tensorflow_docs.tools.nbfmt --indent=1 docs/
```


lamberta:
cc: @balopat

balopat:
Hi @lamberta - adding in google-colab as requirement is problematic - it's going to break all the python 3.8 builds and totally breaks locally jupyter. I opened https://github.com/quantumlib/Cirq/issues/3368 instead what do you think about using pydata-google-auth?

lamberta:
Hey @balopat ,

Adding `pydata-google-auth` instead is fine by me (I'll revert my last commit). Currently the *Build docs* CI test is failing because the `google.colab` module is used in a few notebooks. Can we get around this?

balopat:
If you allow me to add commits to this PR, I'll also setup a check that will ensure that notebooks are formatted correctly on every PR. I will resolve the authentication blob in a separate PR and then we can rebase this on top of that. 

Alternatively, one way I can see getting around this is to comment out the google colab authentication for now and provide details in the comments...but let me think about this a bit more.

balopat:
> If you allow me to add commits to this PR, I'll also setup a check that will ensure that notebooks are formatted correctly on every PR. I will resolve the authentication blob in a separate PR and then we can rebase this on top of that.
> 
> Alternatively, one way I can see getting around this is to comment out the google colab authentication for now and provide details in the comments...but let me think about this a bit more.

Actually setting `nbsphinx_allow_errors = True` is not the worst. I added a comment to turn error checking back on when we switch over to `pydata_google_auth`.

balopat:
Alright, I think we should be good to go now - the new script actually caught an error in my formatting (https://github.com/quantumlib/Cirq/pull/3393/checks?check_run_id=1215735034) when I used PyCharm to edit my notebook to fix a sphinx warning regarding inconsistent headers. It's working nicely!

balopat (approved):

lamberta:
Ah, great. Thanks, @balopat !

balopat:
Argh. I'm looking at the last weird error. Will be fixed soon.

balopat:
There are some other issues as well with the transformation logic (it might be related):

image

lamberta:
For the readthedocs pages, can you strip the `<table class=\"tfo-notebook-buttons\" ... </table>` ?

balopat:
> <table class=\"tfo-notebook-buttons\" ... </table>

working on it...

PR #3395, opened 2020-10-06
Add optional arguments to calibration API

dstrain115:
- Add a dictionary of arguments, so that calibrations can be
specified further with parameters.


cirq/google/api/v2/calibration.proto
```diff
@@ -38,6 +38,16 @@ message CalibrationLayer {
...
      Program layer = 2;
+
+     // Arguments that can be specified to the calibration procedure,
+     // such as the number of layers, which angles to optimize, etc
+     map<string, Arg> args = 4;
+
+     // (Optional) Token to specify an existing calibration that
+     // should be modified or overwritten in order to chain calibrations.
+     // Can also be used to specify a gate in a following circuit when
+     // circuit is specified within the current batch
```
mrwojtek, line 49: It reads that this token plays different roles: in the first part it is a reference to another calibration but in the second part it assigns id to this particular calibration. Is that right? If yes, I can imagine situations where both of those functions would be used simultaneously.

On the other hand, I don't think this token is something which is going to be supported very soon (I'm not sure if it is worth adding it now).

dstrain115: That's true.  It does serve two purposes.  However, I think the two purposes are exclusive.

If you are chaining onto a previous calibration, then you know the token, since you have to tell the API which calibration you are chaining onto.  In this case, you don't need to assign an id to this calibration, since it already has an id (the one you are updating).

I put this on the proto since it was a feature asked for in the last meeting, but it probably will not be implemented for a while, so we can drop it off until we figure out how we will do batching and chaining if there is controversy here.

mrwojtek: Since this is incremental change and will be very easy to add later, I'd vote for dropping it for now.

dstrain115: Done.

mrwojtek (approved):

PR #3400, opened 2020-10-06
fix protobuf serialization test issue

balopat:
Fix protobuf serialization test issue - as in some environments protobuf might not truncate to 4 bytes the test should cater for approximate equality.

### DIFF
--- a/cirq/google/arg_func_langs_test.py
+++ b/cirq/google/arg_func_langs_test.py
@@ -130,7 +130,7 @@ def test_serialize_sympy_constants():
     assert len(packed['arg_value']) == 1
     # protobuf 3.12+ truncates floats to 4 bytes
     assert np.isclose(packed['arg_value']['float_value'],
-                      np.float32(sympy.pi), atol=1e-8)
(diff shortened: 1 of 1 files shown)


cirq/google/arg_func_langs_test.py
```diff
@@ -126,12 +126,12 @@ def test_serialize_sympy_constants():
...
+    assert len(packed) == 1
+    assert len(packed['arg_value']) == 1
     # protobuf 3.12+ truncates floats to 4 bytes
-    assert packed == {
-        'arg_value': {
-            'float_value': float(str(np.float32(sympy.pi)))
-        }
-    }
+    assert np.isclose(packed['arg_value']['float_value'],
+                      np.float32(sympy.pi), atol=1e-8)
```
dstrain115: Is 1e-8 too tight?  Shouldn't it be 1e-6?

dstrain115 (approved):

cirq/google/arg_func_langs_test.py
balopat: Updated to 1e-7

PR #3401, opened 2020-10-06
Remove unnecessary conditional

kevinsung:
`state_like` is definitely a numpy array (see if statement on line 304).


vtomole (commented):
Do we want to change the statement in the `ValueError` to reflect this too? The string there talks about `state_like` maybe not being a numpy array which it will be if the executed code makes it to this point.

kevinsung:
I've improved the error message.

vtomole (approved):
LGTM

PR #3404, opened 2020-10-08
strip license info for RTD pages

### LINKED ISSUE #3402: Read the docs has goofy copyright
**Description of the issue**
Copyright shows up on cirq tutorials on readthedocs.

Go to https://cirq.readthedocs.io/en/latest/docs/gates.html or other pages.  Top part says "Copyright 2020 The Cirq Developers" with apache license.

**Cirq version**
0.9.1.  It probably broke when we adjusted all the copyright notices.

balopat:
Strips license info for readthedocs generation. 

Fixes #3402.

### DIFF
--- a/rtd_docs/conf.py
+++ b/rtd_docs/conf.py
@@ -133,7 +133,12 @@ def autodoc_process(app, what: str, name: str, obj: Any, options,
 
 def source_read(app, docname, source):
     source[0] = re.sub(r'"##### (Copyright 20\d\d The Cirq Developers)"',
-                       r'"**\1**"', source[0])
+                       r'""', source[0])
+    license_pattern = r'(\{\s*?"cell_type": "code".*?"#@title.*License.".*?\},)'
+    if re.search(license_pattern, source[0], flags=re.S):
+        source[0] = re.sub(license_pattern,
+                           r'', source[0], flags=re.S)
+
     source[0] = re.sub(r'"<table.*tfo-notebook-buttons.*"</table>"',
                        r'""',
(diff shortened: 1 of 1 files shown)


rtd_docs/conf.py
```diff
@@ -132,8 +132,14 @@ def autodoc_process(app, what: str, name: str, obj: Any, options,
 
 
 def source_read(app, docname, source):
-    source[0] = re.sub(r'"##### (Copyright 20\d\d The Cirq Developers)"',
-                       r'"**\1**"', source[0])
+    source[0] = re.sub(r'"##### (Copyright 20\d\d The Cirq Developers)"', r'""',
+                       source[0])
+    source[0] = re.sub(
+        r'(\{\s*?"cell_type": "code".*?"#@title.*License.".*?\},)',
```
dstrain115, line 138: nit:  Why have the question mark in ".*?"  Doesn't "*" mean zero or more already?

dstrain115 (approved):

rtd_docs/conf.py
balopat: `?` is for non-greedy matching - i.e. it is trying to find the shortest possible match for `.*` which is by default greedy.

balopat: I guess I could remove the `?` from the first one but the second and the third one need it otherwise it would match larger, cross cell strings.

dstrain115: Ah, gotcha.  Thanks for the explanation.

PR #3410, opened 2020-10-12
Add SerializableGateSet.is_supported

maffoo:
This method checks if a given object containing operations (a circuit, moment, or other arbitrary op tree) is supported by a gateset. Note that this does not address separating the notions of belonging to a gateset from serializability, but at it provides a way to check if a circuit is supported without actually serializing, which can be expensive.


balopat (approved):
LGTM, thank you!

dstrain115 (approved):

PR #3414, opened 2020-10-13
Update single-qubit cliffords to fix XZ decomposition.

### LINKED ISSUE #3411: Single qubit clifford XZ decomposition is missing two Cliffords
**Description of the issue**
See title

```python
import cirq.experiments.qubit_characterizations as ceqc
print(len(ceqc._single_qubit_cliffords().c1_in_xz))  # should be 24, but is 22
```

**Cirq version**
0.9.0.dev

maffoo:
Also adds a test to ensure we are building the full group with either
decomposition, and use ZXZ to ensure that the XZ decomposition uses only
a single microwave gate per clifford.

Fixes #3411

### DIFF
--- a/cirq/experiments/qubit_characterizations.py
+++ b/cirq/experiments/qubit_characterizations.py
@@ -15,7 +15,7 @@
 import dataclasses
 import itertools
 
-from typing import Any, Iterator, List, NamedTuple, Optional, Sequence, Tuple
+from typing import Any, Iterator, List, Optional, Sequence, Tuple
 import numpy as np
 import sympy
 
--- a/cirq/experiments/qubit_characterizations.py
+++ b/cirq/experiments/qubit_characterizations.py
@@ -19,7 +19,6 @@ import numpy as np
 import sympy
 
 from matplotlib import pyplot as plt
-from mpl_toolkits.mplot3d import Axes3D  # type: ignore # pylint: disable=unused-import
 from cirq import circuits, devices, ops, protocols, study, work
 
 Cliffords = NamedTuple('Cliffords', [('c1_in_xy', List[List[ops.Gate]]),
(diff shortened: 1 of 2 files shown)


cirq/experiments/qubit_characterizations_test.py
```diff
@@ -40,6 +42,37 @@ def test_rabi_oscillations():
     assert rms_err < 0.1
 
 
+def test_single_qubit_cliffords():
```
viathor, line 45: Optional: It seems easy to write a unit test that verifies the key property of the Clifford group - that it fixes the Pauli group - but it seems we don't have it. WDYT about adding it?

(Strictly speaking this isn't related to this PR, but if we had this test then observing it pass before and after would provide a lot of confidence in the new logic.)

cirq/experiments/qubit_characterizations.py
```diff
@@ -744,33 +743,56 @@ def _single_qubit_gates(gate_seq: Sequence[ops.Gate],
...
+    s1_x = [
+        [X**0.5],
+        [X**0.5, Y**0.5, X**0.5],
+        [Y**-0.5],
+    ]  # type: List[List[ops.Gate]]
+    s1_y = [
+        [Y**0.5],
+        [X**-0.5, Y**-0.5, X**0.5],
+        [Y, X**0.5],
+    ]  # type: List[List[ops.Gate]]
```
viathor, line 796: nit: Maybe add comments up when `Cliffords` named tuple type is defined explaining the meaning of the five fields.

cirq/experiments/qubit_characterizations_test.py
viathor, line 45: Since one of your goals was to ensure at most one microwave gate in each decomposition, could you add a test for this? Otherwise, this can change on you later unexpectedly.

maffoo: I agree we should have some better abstractions around the pauli group and clifford group, but I think I'll leave that as future work because we'll probably need some design discussion around that.

maffoo: Good idea. Done.

cirq/experiments/qubit_characterizations.py
maffoo: Changed `Cliffords` to a dataclass and added some documentation about what these are.

cirq/experiments/qubit_characterizations_test.py
```diff
@@ -40,6 +42,75 @@ def test_rabi_oscillations():
...
+
+    # Check that XZ decomposition has at most one X gate per clifford.
+    for gates in cliffords.c1_in_xz:
+        num_x = len([gate for gate in gates if isinstance(gate, cirq.XPowGate)])
+        num_z = len([gate for gate in gates if isinstance(gate, cirq.ZPowGate)])
+        assert num_x + num_z == len(gates)
+        assert num_x <= 1
+
+
+def test_single_qubit_cliffords_xz_at_most_one_x_gate_per_clifford():
```
viathor, line 83: The contents of this test duplicate the one above. I suppose you meant to move the for loop on line 76 above into here (and remove what we have in this test now)?

viathor (approved):
Test is duplicated, see comment. Otherwise, LGTM.

cirq/experiments/qubit_characterizations_test.py
maffoo: Whoops! Cleaned this up.

viathor (approved):
The new `Cliffords` type is really nice! Thank you!

PR #3417, opened 2020-10-14
Verify that Cliffords fix Pauli group

viathor:
Follow-up to #3414. Sorry, couldn't resist.

### DIFF
--- a/cirq/experiments/qubit_characterizations_test.py
+++ b/cirq/experiments/qubit_characterizations_test.py
@@ -84,7 +84,7 @@ def test_single_qubit_cliffords():
     # Check that each unitary fixes the Pauli group.
     for u in xy_unitaries:
         for p in PAULIS:
-            assert is_pauli(u @ p @ u.conj().T), f'{u}'
+            assert is_pauli(u @ p @ u.conj().T), str(u)
 
     # Check that XZ decomposition has at most one X gate per clifford.
     for gates in cliffords.c1_in_xz:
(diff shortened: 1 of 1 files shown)


cirq/experiments/qubit_characterizations_test.py
```diff
@@ -72,6 +81,11 @@ def check_distinct(unitaries):
         assert any(
             cirq.allclose_up_to_global_phase(Uxy, Uxz) for Uxz in xz_unitaries)
 
+    # Check that each unitary fixes the Pauli group.
+    for u in xy_unitaries:
+        for p in PAULIS:
+            assert is_pauli(u @ p @ np.conjugate(np.transpose(u))), f'{u}'
```
balopat, line 87: I.conj().T would be a but shorter

viathor: Done.

cirq/experiments/qubit_characterizations_test.py
```diff
@@ -72,6 +81,11 @@ def check_distinct(unitaries):
         assert any(
             cirq.allclose_up_to_global_phase(Uxy, Uxz) for Uxz in xz_unitaries)
 
+    # Check that each unitary fixes the Pauli group.
+    for u in xy_unitaries:
+        for p in PAULIS:
+            assert is_pauli(u @ p @ u.conj().T), f'{u}'
```
maffoo, line 87: nit: instead of `f'{u}'` can just do `str(u)`

maffoo (approved):
LGTM

cirq/experiments/qubit_characterizations_test.py
viathor: Done.

PR #3418, opened 2020-10-14
Adding citation doc for devsite

balopat:
Adding citation doc for devsite.

### DIFF
--- a/docs/citation.md
+++ b/docs/citation.md
@@ -1,0 +1,12 @@
+# Citing Google Quantum AI Open Source Software
+
+When using our projects, please cite them according to the following guide:
+
+| Project | Citation reference | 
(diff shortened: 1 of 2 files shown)


viathor (approved):
There is a risk that these URLs will break since there is nothing in the target repos indicating that a file in cirq needs to be updated when the URLs change. That said, these files change slowly and it's nice to have these in one place, so LGTM.

balopat:
> There is a risk that these URLs will break since there is nothing in the target repos indicating that a file in cirq needs to be updated when the URLs change. That said, these files change slowly and it's nice to have these in one place, so LGTM.

Agreed that there is risk, it's hard to keep 5 projects' info together. Thanks for the LGTM.

Issue #3394, opened 2020-10-06
Fix validate_normalized_state requiring complex64


kevinsung:
~~Inconsistent use of np.complex64 vs np.complex128 results in undesirable behavior such as~~

Calling `validate_normalized_state` fails despite not specifying a required dtype: 

```python
state = cirq.testing.random_superposition(4)
cirq.validate_normalized_state(state, qid_shape=(2, 2))
```
giving a ValueError:
```
ValueError: state_vector has invalid dtype. Expected <class 'numpy.complex64'> but was complex128
```

kevinsung:
We should probably use complex64 due to space and time savings. The drawback is that numpy uses complex128 by default, so users creating numpy arrays using numpy will either have to tell numpy to use complex64, or tell Cirq that they are using complex128.

Strilanc:
To me it looks like the bug here is "validate" assuming complex64 is required, instead of only doing that if the user asks for it.

satw1knandala:
hi, I can tackle this. please assign me if you need it

kevinsung:
@satw1knandala I've already submitted #3428, which fixes this.

PR #3419, opened 2020-10-15
Add and use QuantumState class

### LINKED ISSUE #2904: Define new state type that encompasses STATE_VECTOR_LIKE plus density matrices and use it in fidelity function

kevinsung:
Fixes #2904.


cirq/qis/states.py
```diff
@@ -35,6 +36,435 @@
...
+                        atol=atol)
+
+
+def density_matrix(
+        state: np.ndarray,
+        *,  # Force keyword arguments
+        validate: bool = True,
+        dtype: Type[np.number] = np.complex64,
+        atol: float = 1e-7) -> QuantumState:
+    if state.ndim != 2 or len(set(state.shape)) != 1:
```
viathor, line 223: Constructing a set is a little heavy for our simple purpose here. You can just say:

```python
if state.ndim != 2 or state.shape[0] != state.shape[1]:
```

cirq/qis/states.py
```diff
@@ -35,6 +36,435 @@
...
+    return QuantumState(data=state,
+                        qid_shape=qid_shape,
+                        dtype=dtype,
+                        validate=validate,
+                        atol=atol)
+
+
+def _infer_qid_shape_from_dimension(dim: int) -> Tuple[int, ...]:
+    if dim != 0 and dim & dim - 1 == 0:
+        # dim is a power of 2, assume qubits
```
viathor, line 237: This is a fair assumption, but we should provide a way for users to say that they know their qid_shape already. This would allow users to resolve the ambiguity in your ququad example [here](https://github.com/quantumlib/Cirq/issues/2904#issuecomment-694232713) where a 4x4 density matrix may describe either a state of two qubits or one ququad.

I think we should add qid_shape argument to `density_matrix`. Then we'd call `_infer_qid_shape_from_dimension` only if the user hasn't provided qid_shape explicitly.

cirq/qis/states.py
```diff
@@ -35,6 +36,435 @@
...
+        return self._qid_shape
+
+    @property
+    def dtype(self) -> np.ndarray:
+        return self._data.dtype
+
+    def is_state_vector(self) -> bool:
+        return self.data.shape == (self._dim,)
+
+    def is_state_tensor(self) -> bool:
```
viathor, line 106: Since this function name does not correspond to existing terminology, we should document it, e.g.

```python
"""Returns true iff self is a tensor product of state vectors."""
```

In fact, it'd be good to add docstrings to all functions.

cirq/qis/states.py
```diff
@@ -35,6 +36,435 @@
...
+        return self.data.shape == self.qid_shape
+
+    def is_density_matrix(self) -> bool:
+        return self.data.shape == (self._dim, self._dim)
+
+    def is_pure_state(self) -> bool:
```
viathor, line 112: I'm uneasy about this function because it may mislead the caller when `data` is a density matrix that correspond to a pure state such as [[1, 0], [0, 0]].

It seems to me that we don't actually need it: we could replace calls to `is_pure_state` with `not is_density_matrix`.

cirq/qis/measures_test.py
```diff
@@ -30,40 +30,45 @@ def test_fidelity_symmetric():
...
                                np.sum(np.sqrt(d1 * d2))**2)
 
 
 def test_fidelity_known_values():
-    vec1 = np.array([1, 1j, -1, -1j]) * 0.5
-    vec2 = np.array([1, -1, 1, -1]) * 0.5
-    mat1 = np.outer(vec1, vec1.conj())
-    mat2 = np.outer(vec2, vec2.conj())
-    mat3 = 0.3 * mat1 + 0.7 * mat2
+    vec1 = np.array([1, 1j, -1, -1j], dtype=np.complex128) * 0.5
```
viathor, line 67: I'm surprised we need to specify dtype here. Do you know why?

cirq/qis/measures.py
```diff
@@ -38,26 +46,72 @@ def fidelity(state1: np.ndarray, state2: np.ndarray) -> float:
...
+        ])**2
+
+    if qid_shape is None:
+        qid_shape = infer_qid_shape(state1, state2)
+
+    state1 = quantum_state(state1,
+                           qid_shape=qid_shape,
+                           dtype=state1.dtype,
+                           validate=validate)
+    state2 = quantum_state(state2,
```
viathor, line 79: For performance, we should not be wrapping input states to simplify control flow in the computation of fidelity. Instead, we should choose the appropriate fidelity formula based on the shape of the input numpy arrays (when the inputs are numpy arrays) and have additional branch(es) for the case when some inputs are QuantumStates. This way, most of the codepaths will completely avoid our QuantumState machinery, but at the same time in the rare cases where ambiguity arises users will be able to use `density_matrix` to tell our fidelity code what to do.

(In fact, I think we should adopt this approach more generally for all code that implements functions on quantum states, e.g. trace distance, etc.)

viathor (commented):
For now I only had a quick first look so sorry if some comments don't make sense.

cirq/qis/states.py
kevinsung: Done.

cirq/qis/states.py
kevinsung: Done.

cirq/qis/states.py
kevinsung: Done.

cirq/qis/measures_test.py
kevinsung: Yeah, it wasn't necessary on this line. I'm actually considering removing the default dtype specification of complex64 in QuantumState and related methods, and instead by default reading the dtype from the numpy array passed in. I think that's more in line with the change proposed at https://github.com/quantumlib/Cirq/issues/3394.

cirq/qis/states.py
kevinsung: I've removed this method since I felt it the underlying data representation should be mostly transparent to the user, and in any case it's preferable to start out with a minimal public API that we can expand later if needed.

I've added docstrings to all functions.

cirq/qis/measures_test.py
kevinsung: I've removed unnecessary dtype specifications and made specifying the dtype optional, following #3428 .

cirq/qis/measures.py
kevinsung: I've added a special case to handle when the two inputs are complex numpy arrays.

kevinsung:
@viathor I've addressed your comments. I've also marked this PR ready for review (it's no longer a draft).

balopat:
Sorry for letting this sit for so long @kevinsung  - I'll have a look tomorrow - can you update the branch?

kevinsung:
@balopat no worries. I've updated the branch.

cirq/qis/measures.py
```diff
@@ -38,36 +56,211 @@ def fidelity(state1: np.ndarray, state2: np.ndarray) -> float:
...
+        ValueError: The qid shape of the given states was not specified and
+            could not be inferred.
+        ValueError: Invalid quantum state.
     """
-    if len(state1.shape) == 1 and len(state2.shape) == 1:
+    # Two ints
+    if isinstance(state1, int) and isinstance(state2, int):
+        if validate and qid_shape is not None:
+            dim = np.prod(qid_shape)
+            if state1 >= dim:
```
balopat, line 75: Do we want to guard against negative states too?

cirq/qis/measures.py
```diff
@@ -38,36 +56,211 @@ def fidelity(state1: np.ndarray, state2: np.ndarray) -> float:
...
+                raise ValueError(
+                    'Invalid state for given qid shape: '
+                    'The maximum computational basis state for qid shape '
+                    f'{qid_shape} is {dim - 1}, but {state1} was given.'
+                )
+            if state2 >= dim:
```
balopat, line 81: I'd dedupe these two to a private `_validate_int_state` method

cirq/qis/measures.py
```diff
@@ -38,36 +56,211 @@ def fidelity(state1: np.ndarray, state2: np.ndarray) -> float:
...
+        return float(state1 == state2)
+
+    # Two ProductStates
+    if isinstance(state1, value.ProductState) and isinstance(state2, value.ProductState):
+        if len(state1) != len(state2):
+            raise ValueError(
+                'Mismatched number of qubits in product states: '
+                f'{len(state1)} and {len(state2)}.'
+            )
+        if validate and qid_shape is not None and qid_shape != (2,) * len(state1):
```
balopat, line 96: Are we missing the validation for `state2`?

cirq/qis/measures.py
```diff
@@ -38,36 +56,211 @@ def fidelity(state1: np.ndarray, state2: np.ndarray) -> float:
...
+
+
+def _numpy_arrays_to_state_vectors_or_density_matrices(
+    state1: np.ndarray,
+    state2: np.ndarray,
+    qid_shape: Optional[Tuple[int, ...]],
+    validate: bool,
+    atol: float,
+) -> Tuple[np.ndarray, np.ndarray]:
+    if state1.ndim > 2 or state1.ndim == 2 and state1.shape[0] != state1.shape[1]:
```
balopat, line 138: add parens to make the clauses a bit clearer - I had to look twice ("if state1.dim is greater or equal to 2, why not use >=...?:)")

cirq/qis/measures.py
```diff
@@ -38,36 +56,211 @@ def fidelity(state1: np.ndarray, state2: np.ndarray) -> float:
...
+            qid_shape = (dim1,)
+        else:
+            expected_dim = np.prod(qid_shape)
+            if dim1 != expected_dim:
+                raise ValueError(
+                    'Invalid state dimension for given qid shape: '
+                    f'Expected dimension {expected_dim} but '
+                    f'got dimension {dim1}.'
+                )
+        if state1.ndim == 2:
```
balopat, line 188: I'd combine these two ifs into a single for to make it more concise

cirq/qis/states.py
```diff
@@ -36,6 +37,503 @@
...
+        actual_qid_shape = (2,) * len(state)
+        if qid_shape is not None and qid_shape != actual_qid_shape:
+            raise ValueError(
+                'The specified qid shape must be the same as the '
+                'qid shape of the given state.\n'
+                f'Specified shape: {qid_shape}\n'
+                f'Shape of state: {actual_qid_shape}.'
+            )
+        if dtype is None:
+            dtype = np.complex64
```
balopat, line 227: can we pull this out to a constant, like DEFAULT_COMPLEX_PRECISION? It feels like a decision that requires shotgun surgery to extract. I know it exists elsewhere in the codebase, but maybe we could eventually get all of them to a single point?

cirq/qis/states.py
```diff
@@ -36,6 +37,503 @@
...
+
+        The set of qid shapes is represented as the union of a set of shapes
+        specified explicitly in `explicit_qid_shapes`, a set of shapes
+        specified implicitly in `unfactorized_total_dimension`, and a set of
+        shapes specified implicitly in `min_qudit_dimensions`.
+
+        Args:
+            explicit_qid_shapes: An explicit set of qid shapes.
+            unfactorized_total_dimension: A number representing the dimension
+                of Hilbert space. The associated qid shapes are those compatible
```
balopat, line 428: of >the< Hilbert space

cirq/qis/states.py
```diff
@@ -36,6 +37,503 @@
...
+def _infer_qid_shape_from_dimension(dim: int) -> Tuple[int, ...]:
+    if dim != 0 and dim & dim - 1 == 0:
+        # dim is a power of 2, assume qubits
+        n_qubits = dim.bit_length() - 1
+        return (2,) * n_qubits
+    # dim is not a power of 2, assume a single qudit
+    return (dim,)
+
+
+def infer_qid_shape(*states: 'cirq.QUANTUM_STATE_LIKE') -> Tuple[int, ...]:
```
balopat, line 314: can we split qid shape inference to another PR? We can throw an error in the meantime from `quantum_state()`

balopat (changes requested):
Thank you so much for your patience on this Kevin! Added some comments, but otherwise looks pretty good to me! Thanks!

cirq/qis/measures.py
kevinsung: Done.

cirq/qis/measures.py
kevinsung: Done.

cirq/qis/measures.py
kevinsung: Done.

cirq/qis/measures.py
kevinsung: Done.

cirq/qis/measures.py
kevinsung: Done.

cirq/qis/states.py
kevinsung: Done.

cirq/qis/states.py
kevinsung: Done.

cirq/qis/states.py
kevinsung: Yes, that makes sense. I've removed it from this PR and will open a separate one.

balopat (approved):

balopat:
cla cirqbot issue surfaces again: https://github.com/quantumlib/Cirq/issues/3736

PR #3426, opened 2020-10-20
Dedup noise doc

95-martin-orion:
This PR removes duplicated sections in `docs/noise.ipynb`.


docs/noise.ipynb
```diff
@@ -218,130 +218,6 @@
...
-    "`cirq.amplitude_damp` to construct these noisy gates."
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {
-    "id": "f3933141f64e"
-   },
-   "source": [
-    "### cirq.mixture and def _mixture_\n",
```
viathor, line 259: Optional: I think underscores will turn this into italics instead of displaying as `_mixture_`, so there should be backticks around `"def _mixture_"`.

While you're at it, consider fixing in the copy which you're keeping.

viathor (approved):

docs/noise.ipynb
95-martin-orion: Done.

viathor (approved):

Issue #3169, opened 2020-07-22
Exponential of a (commuting) PauliSum


tanujkhattar:
Exponentiating a `PauliSum` where individual terms commute is a fairly common utility function. Both [TFQ](https://github.com/tensorflow/quantum/blob/1f050a8ca3217adb56bdd3bf29255a4e8d610ceb/tensorflow_quantum/python/util.py#L419-L520) and [OpenFermion](https://github.com/quantumlib/OpenFermion-Cirq/pull/395) need/implement this functionality. 

Given all the related nice functions which already exist in Cirq (`to_z_basis_ops`, `PauliSumCollector` etc.), I think we should also add a utility to achieve this.

Thoughts? I can send a PR if we have a consensus.

balopat:
Thank you for opening this Tanuj! I'm personally open to it, but I'd like to get the opinion of the rest of the maintainers - let's discuss on Cirq Cynq tomorrow!

balopat:
Cirq Cynque: 
- let's do it
- implement first only for commuting terms 
- if they don't commute what do we do? 
    - first step: throw an error 
    - next PRs - for a few anticommuting terms you can do a tableau based exponentiation

dabacon:
There are sort of three cases:
1) commuting Pauli terms.
2) case where there are only a few anticommuting terms
3) no real structure.

Let H=c_1P_1+c_2P_2 +... +c_k P_k

In case 1) you can write this as a series of PauliStringPhasors, i.e. exp(iH) = exp(ic_1P) exp(ic_2P_2)...exp(ic_kP_k)

In case 2) you can use a stabilizer tableau (https://pdfs.semanticscholar.org/67af/c372369240e202ec902fb11ab25832d403af.pdf).  The basic idea would be to iterate over the P_i and build up a tableau.  If the next P_i commutes with the everything in the current tableau, and is not in the tableau, you add it as a z stabilizer for a new logical qubit.  If it anticommutes with some of the elements in the tableau, you think about it as an encoded operation on the logical qubits, or whether you can add a new logical stabilizer to make it a product of an existing x/y stabilizer and this new stabilizer.  Note that it may span multiple logical qubits.  Then you look at the topological sort of z stabilizers and these logical qubits.  Different disjoint sets can then be exponentiated.  You have to do this numerically, except probably you could be explicit about the case where these are single qubits since exp(iH) for H a single qubit is easy to write down.  


Case 3) is where you just do the full exponentiation numerically.

[...]

PR #3427, opened 2020-10-20
Add PauliSumExponential for commuting Pauli terms

tanujkhattar:
First PR for #3169


cirq/ops/pauli_sum_exponential.py
```diff
@@ -0,0 +1,131 @@
...
+from cirq import linalg, protocols, value
+from cirq.ops import linear_combinations, pauli_string_phasor
+
+if TYPE_CHECKING:
+    import cirq
+
+
+@value.value_equality(approximate=True)
+class PauliSumExponential:
+    """Represents operator defined by exponential of a PauliSum.
```
balopat, line 28:
```suggestion
    """Represents an operator defined by the exponential of a PauliSum.
```

cirq/ops/pauli_sum_exponential.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Given a hermitian/anti-hermitian PauliSum PS_1 + PS_2 + ... + PS_N, this
+    class returns an operation which is equivalent to
+    exp(j * exponent * (PS_1 + PS_2 + ... + PS_N)).
+
+    This class currently supports only exponential of commuting Pauli Terms.
```
balopat, line 34:
```suggestion
    This class only supports commuting Pauli terms.
```

cirq/ops/pauli_sum_exponential.py
```diff
@@ -0,0 +1,131 @@
...
+    def _resolve_parameters_(self,
+                             param_resolver: 'cirq.ParamResolverOrSimilarType'
+                            ) -> 'PauliSumExponential':
+        return PauliSumExponential(self._pauli_sum,
+                                   exponent=protocols.resolve_parameters(
+                                       self._exponent, param_resolver),
+                                   is_anti_hermitian=self._is_anti_hermitian)
+
+    @staticmethod
+    def _all_pauli_strings_commute_(pauli_sum: 'cirq.PauliSum') -> bool:
```
balopat, line 82: Convert this to a package level method instead of static method.

cirq/ops/pauli_sum_exponential.py
```diff
@@ -0,0 +1,131 @@
...
+                                   is_anti_hermitian=self._is_anti_hermitian)
+
+    def __repr__(self) -> str:
+        class_name = self.__class__.__name__
+        return f'cirq.{class_name}({self._pauli_sum!r}, {self._exponent!r}, '\
+               f'is_anti_hermitian={self._is_anti_hermitian!r})'
+
+    def __str__(self) -> str:
+        coeff = '1.0' if self._multiplier != 1 else 'j'
+        return f'exp({(coeff)!s} * {self._exponent!s} * ({self._pauli_sum!s}))'
```
balopat, line 131:
```suggestion
        return f'exp({coeff!s} * {self._exponent!s} * ({self._pauli_sum!s}))'
```

cirq/ops/pauli_sum_exponential.py
```diff
@@ -0,0 +1,131 @@
...
+                                   self._exponent * exponent,
+                                   is_anti_hermitian=self._is_anti_hermitian)
+
+    def __repr__(self) -> str:
+        class_name = self.__class__.__name__
+        return f'cirq.{class_name}({self._pauli_sum!r}, {self._exponent!r}, '\
+               f'is_anti_hermitian={self._is_anti_hermitian!r})'
+
+    def __str__(self) -> str:
+        coeff = '1.0' if self._multiplier != 1 else 'j'
```
balopat, line 130: We could even remove 1.0?

```suggestion
        coeff = '' if self._multiplier != 1 else 'j'
```

cirq/ops/pauli_sum_exponential.py
```diff
@@ -0,0 +1,131 @@
...
+        self._exponent = exponent
+        self._is_anti_hermitian = is_anti_hermitian
+        if is_anti_hermitian:
+            self._multiplier = -1j
+        else:
+            self._multiplier = 1
+        for pauli_string in pauli_sum:
+            normalised_coeff = complex(pauli_string.coefficient *
+                                       self._multiplier)
+            if normalised_coeff.imag != 0:
```
balopat, line 55: I think we should check this up to an absolute tolerance, right?

cirq/ops/pauli_sum_exponential.py
```diff
@@ -0,0 +1,131 @@
...
+            for y in pauli_sum:
+                if not protocols.commutes(x, y):
+                    return False
+        return True
+
+    def __iter__(self) -> Iterator['cirq.PauliStringPhasor']:
+        for pauli_string in self._pauli_sum:
+            theta = pauli_string.coefficient * self._multiplier
+            theta *= self._exponent / np.pi
+            if isinstance(theta, complex):
```
balopat, line 93: Why are we just simply truncating here? Also, how can this be complex? Didn't we ensure in the constructor that for all coeffs coeff * multiplier is real? Also self._exponent should be real, right?

cirq/ops/pauli_sum_exponential.py
```diff
@@ -0,0 +1,131 @@
...
+    exp(j * exponent * (PS_1 + PS_2 + ... + PS_N)).
+
+    This class currently supports only exponential of commuting Pauli Terms.
+    """
+
+    def __init__(self,
+                 pauli_sum_like: 'cirq.PauliSumLike',
+                 exponent: Union[int, float, sympy.Basic] = 1,
+                 *,
+                 is_anti_hermitian=False):
```
balopat, line 41: can't we just determine this based on the coefficients? Why does the user need to define this?

balopat (changes requested):
Thank you for working on this Tanuj, I added some comments and questions. 
@viathor, I would love to have your eyes on this too!

cirq/ops/pauli_sum_exponential.py
tanujkhattar: Let res = coeff * multipler. We ensured that res is real (i.e. res.imag ~= 0) but the type of the object res is still complex and hence we simply truncate it to get the real part because the exponent passed to PauliStringPhasor should be float.

cirq/ops/pauli_sum_exponential.py
tanujkhattar: done

cirq/ops/pauli_sum_exponential.py
tanujkhattar: done.

cirq/ops/pauli_sum_exponential.py
tanujkhattar: done

cirq/ops/pauli_sum_exponential.py
tanujkhattar: done

cirq/ops/pauli_sum_exponential.py
tanujkhattar: Modified to determine based on coefficients. The original thought was to specify explicitly so that the action is more clear on the call site but I think the docstring is probably enough to convey the action.

cirq/ops/pauli_sum_exponential.py
tanujkhattar: Fixed.

cirq/ops/pauli_sum_exponential.py
```diff
@@ -0,0 +1,126 @@
...
+    def __init__(
+        self, pauli_sum_like: 'cirq.PauliSumLike', exponent: Union[int, float, sympy.Basic] = 1
+    ):
+        pauli_sum = linear_combinations.PauliSum.wrap(pauli_sum_like)
+        if not _all_pauli_strings_commute(pauli_sum):
+            raise ValueError("PauliSumExponential defined only for commuting pauli sums.")
+        self._multiplier = None
+        for pauli_string in pauli_sum:
+            coeff = pauli_string.coefficient
+            curr_multiplier = -1j if abs(coeff.imag) > 1e-8 else 1.0
```
balopat, line 54: We typically extract this into a `atol` parameter with a default value of `1e-8`. 
```suggestion
            curr_multiplier = -1j if abs(coeff.imag) > atol else 1.0
```

tanujkhattar: done

cirq/ops/pauli_sum_exponential.py
balopat: Sounds good.

balopat (approved):
LGTM, let's fix the conflicts and we can merge this

PR #3430, opened 2020-10-20
Removing Quantum chess tutorial from cirq

dstrain115:
- This will be moving to ReCirq.

See https://github.com/quantumlib/ReCirq/pull/82


balopat (approved):
Thank you!
Goodbye Quantum Chess! See you in ReCirq!

PR #3437, opened 2020-10-21
Docs nav fixes

lamberta:


balopat (approved):

balopat:
Thanks for the fix! Yeah, it would be great to have a script validating the _book.yaml.

PR #3439, opened 2020-10-22
Replace `Study` with `sweep` in docs

95-martin-orion:
From what I can tell, the nearest equivalent to the old concept of "Study" is a sweep. Strangely, the example code seems to be up-to-date for this.

### DIFF
--- a/docs/simulation.ipynb
+++ b/docs/simulation.ipynb
@@ -535,12 +535,10 @@
     "Here we see that the ``Symbol`` is used in two gates, and then the resolver provides this value at run time.\n",
     "\n",
     "Parameterized values are most useful in defining what we call a\n",
-    "``sweep``.  A ``sweep`` is a collection of trials, where each\n",
-    "trial is a run with a particular set of configurations and which\n",
-    "may be run repeatedly.  \n",
+    "``sweep``.  A ``sweep`` is a sequence of trials, where each\n",
+    "trial is a run with a particular set of parameter values.\n",
     "\n",
-    "Running a `sweep` returns a\n",
-    "``Result`` per set of fixed parameter values and repetitions.  \n",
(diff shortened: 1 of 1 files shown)


docs/simulation.ipynb
```diff
@@ -535,14 +535,12 @@
     "Here we see that the ``Symbol`` is used in two gates, and then the resolver provides this value at run time.\n",
     "\n",
     "Parameterized values are most useful in defining what we call a\n",
-    "``Study``.  A ``Study`` is a collection of trials, where each\n",
+    "``sweep``.  A ``sweep`` is a collection of trials, where each\n",
```
balopat, line 538: Sweep != Study

I think Study was a larger concept, see https://github.com/quantumlib/Cirq/commit/42faf4bc9e8d2d8acec696c1162923a5e9aecfa8#diff-d00dd130390b1095e7f1f95c4bd9ff4e20bb2d2b093674252d6b85dda0c3ba12R39

As such I think these sentences become confusing as Sweep is nothing else just an iterator over different values of the given parameters for a Circuit. Can you rephrase?

balopat (changes requested):
Thanks for filing! I think we need to reflect the conceptual difference between study and sweep.

docs/simulation.ipynb
balopat: Another thing: maybe we could capture the type of Sweeps somewhere - but that could be another PR. 
I described the different Sweep types here: https://github.com/quantumlib/Cirq/pull/3256#pullrequestreview-474013593 WDYT?

95-martin-orion: Cleaned this up a bit. Personally, I think "sweeps as a data structure" should be folded under the concept of subcircuits - a sweep is just a loop over an entire circuit with different parameters for each iteration, after all.

balopat (approved):

PR #3440, opened 2020-10-23
Document private protocol methods for devsite

### LINKED ISSUE #2953: Protocol documentation are missing methods
Likely because they are private!  

It looks like a workaround might be to include
```
.. automethod:: __magic_method_
```
in the docstring

balopat:
Fixes #2953.

Also adds a preview Dockerfile that is a seed for a potential preview service that could run as a GithubAction, deploy an ephemeral Cloud Run container to preview the generated docs and comment on the PR with the link. An example link for this particular change is: https://api-docs-fijpqdu4wa-uc.a.run.app/

While the preview is not "the real thing" - a fullblown devsite preview is not realistic at this point. This preview gives an indication about the intermediate markdown representation generated by tf-docs markdown generator. It's good for:
- latex checking in docstrings - e.g. https://api-docs-fijpqdu4wa-uc.a.run.app/cirq/experiments/two_qubit_state_tomography.md
- checking whether the private methods really show up: https://api-docs-fijpqdu4wa-uc.a.run.app/cirq/protocols/SupportsActOn.md


cirq/protocols/qid_shape_protocol.py
```diff
@@ -17,11 +17,11 @@
 from typing_extensions import Protocol
 
 from cirq import ops
-from cirq._doc import document
+from cirq._doc import document, doc_private
 from cirq.type_workarounds import NotImplementedType
 
 if TYPE_CHECKING:
-    import cirq
+    pass
```
95-martin-orion, line 24: Remove this block?

95-martin-orion (approved):
All looks good to me. As a potential follow-up item: on the api-docs preview, would it be feasible to show diffs for the source PR?

cirq/_doc.py
```diff
@@ -54,3 +54,33 @@ def document(value: Any, doc_string: str = ''):
         # print(traceback.format_stack(limit=2)[0])
         pass
     return value
+
+
+# The below is based on tools/tensorflow_docs/api_generator/doc_controls.py#L295
+# in the https://github.com/tensorflow/docs repo
```
viathor, line 60: Optional nit: I'd better to provide direct link to a commit. As it is, the reader is required to do a small copy-paste dance to get there.

cirq/_doc.py
```diff
@@ -54,3 +54,33 @@ def document(value: Any, doc_string: str = ''):
...
+    """A decorator: Generates docs for private methods/functions.
+
+    For example:
+    ```
+    class Try:
+      @doc_private
+      def _private(self):
+        ...
+    ```
+    As a rule of thumb, private(beginning with `_`) methods/functions are
```
viathor, line 75: nit: space before (

viathor (approved):

cirq/_doc.py
balopat: Agreed - that's how I started, but then the linter complained ... hmm I think there might be a way - if the line starts with `http*`...let me check.

cirq/protocols/qid_shape_protocol.py
balopat: Good catch - I thought I cleaned all those up - thanks!

cirq/_doc.py
mpharrigan: we should start keeping a list of times when the linter makes us make things less readable

balopat: @mpharrigan added a "had-to-fight-with-the-linter" label :) Feel free to use it

Issue #3235, opened 2020-08-19
Classical Control: Subcircuits


balopat:
**Problem**: Cirq currently supports circuits, which are made of a series of moments, each moment is made up of different operations that act on disjoint sets of qubits.  These are meant to represent as closely as possible the execution of the circuit on hardware (the exception being that some gates are virtual and compiled out the circuit at runtime).  In addition there is the concept of a noise model, which, at runtime, adds on extra operations to a moment.  Stepping through a simulation with noise has the same iteration order as stepping through the moments, but with extra noise model operations per moment.  This feature would be to add some notion of grouping or subcircuits.  This is likely a necessary feature for the Flow Control issues listed above.  It would also be useful for compilation in order to delineate structures that could and could not be broken (in addition to the pragmas supplied by tags).  Such structures could presumably be used to replace noise models.

**Is coupled to feature:** Flow control (#3234), Feedforward (#3232)

**Rough requirements:**

- Add an abstraction to group portions of circuits
- Be able to deal with disjoint groups of qubits
- Maintain the ability to write unrolled moment by moment circuits
- Be flexible enough to supply information to executors which, may interpret the grouping as it executes the subgroup.
- Potentially be able to specify execution time information that can change the ordering of the execution
- Be serializable

95-martin-orion:
The core feature here has been resolved, and this is currently acting as an umbrella issue. The only remaining sub-issue marked for resolution before 1.0 is #3613; let's get that sorted and then reassign this to after 1.0.

dstrain115:
This umbrella issue has been mostly or totally resolved.  Closing this issue.

PR #3443, opened 2020-10-26

95-martin-orion:
`AbstractCircuit` contains all of the non-mutating methods of the Circuit class, and is inherited by both `Circuit` (which is mutable) and the new `FrozenCircuit` class (which is immutable).

This PR is part of #3235.


cirq/circuits/frozen_circuit.py
```diff
@@ -0,0 +1,157 @@
...
+    @property
+    def moments(self) -> Sequence['cirq.Moment']:
+        return self._moments
+
+    @property
+    def device(self) -> devices.Device:
+        return self._device
+
+    @staticmethod
+    def freeze(circuit: Circuit) -> 'FrozenCircuit':
```
maffoo, line 57: Instead of a static method I'd suggest having a `freeze` method on `AbtractCircuit` that returns `FrozenCircuit`. This would create a new frozen circuit if called on a mutable be circuit, and return `self` if called on a frozen circuit.

cirq/circuits/frozen_circuit.py
```diff
@@ -0,0 +1,157 @@
...
+        return FrozenCircuit(circuit.moments, device=circuit.device)
+
+    def base_circuit(self) -> Circuit:
```
maffoo, line 60: Maybe call this `unfreeze` instead? As above, could also make `unfreeze` be a method on `AbstractCircuit` if desired which would create a copy of a mutable circuit, or a new mutable instance if called on a frozen circuit.

cirq/circuits/frozen_circuit.py
```diff
@@ -0,0 +1,157 @@
...
+        return Circuit(self.moments, device=self.device)
+
+    def __repr__(self):
+        return super().__repr__().replace('AbstractCircuit', 'FrozenCircuit')
```
maffoo, line 64: Instead of string replacement here can just implement `__repr__` on the base class using `self.__class__.__name__` as suggested above.

cirq/circuits/frozen_circuit.py
```diff
@@ -0,0 +1,157 @@
...
+            p.text('FrozenCircuit(...)')
+        else:
+            super()._repr_pretty_(p, False)
+
+    def __hash__(self):
+        return hash((self.moments, self.device))
+
+    # Memoized methods for commonly-retrieved properties.
+
+    def _num_qubits_(self):
```
maffoo, line 77: There's a decorator for implementing cached properties like this in python 3.8: [`functools.cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property). It looks like a package is available to backport this to 3.7: https://pypi.org/project/backports.cached-property/.

maffoo (commented):
This is great! Very excited to have this functionality.

vtomole (commented):
There are some very old commits here: https://github.com/quantumlib/Cirq/pull/3443/commits

maffoo:
On the subject of commit history, something that might help here is to do a first PR that just reorders the methods in `Circuit` to move all the mutable ones to the end of the class, but does not make any logical changes. Then a second PR to split up `AbstractCircuit` and `Circuit` would be a much smaller diff and easier to review.

95-martin-orion:
> There are some very old commits here: https://github.com/quantumlib/Cirq/pull/3443/commits

I'm suffering from Git confusion: many of these commits appear in PR #2715, which was merged. Why do they appear again here?

maffoo:
@95-martin-orion, you probably just need to rebase your branch on master and force-push.

95-martin-orion:
Discussion moved to #3445, with commits cleaned up for easier review :)

PR #3445, opened 2020-10-27

95-martin-orion:
Reupload of #3443, this time with a cleaner commit history :) This PR is part of #3235.

I recommend reviewing this PR one commit at a time, as it's been split up based on comments from #3443 for cleaner diffs.

RE `flatten_to_ops`, `_validate_op_tree_qids`, `[find]all_operations` - these were only moved, not modifed by this PR. I can address these in a follow-up if you'd like.

Some comments from @maffoo on that PR that still need to be addressed:

> Instead of a static method I'd suggest having a freeze method on AbtractCircuit that returns FrozenCircuit. This would create a new frozen circuit if called on a mutable be circuit, and return self if called on a frozen circuit.

> There's a decorator for implementing cached properties like this in python 3.8: [`functools.cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property). It looks like a package is available to backport this to 3.7: https://pypi.org/project/backports.cached-property/.

I also need to revisit the failed checks from that PR - I think my YAPF settings are incorrect.


95-martin-orion:
Replying to @maffoo's comment:

> There's a decorator for implementing cached properties like this in python 3.8: functools.cached_property. It looks like a package is available to backport this to 3.7: https://pypi.org/project/backports.cached-property/.

Since these aren't actually properties, the `@lru_cache` decorator would work here...except for a small gotcha. These caching operators will return the same value on subsequent calls if the arguments are equal. This breaks the `all_measurement_keys` behavior, since:

1. Circuits are equal if their moments and devices are equal.
2. Moments are equal if their _sorted_ operations are equal.
3. `all_measurement_keys` is expected to return keys in the order they were added.

Thus, the following case from `test_all_measurement_keys` produces unexpected behavior, since the two circuits are equal but should return different values for `all_measurement_keys`.
```python
    # Output order matches insertion order, not qubit order.
    assert cirq.FrozenCircuit(
        cirq.Moment([
            cirq.measure(a, key='x'),
            cirq.measure(b, key='y'),
        ])).all_measurement_keys() == ('x', 'y')

    assert cirq.FrozenCircuit(
        cirq.Moment([
            cirq.measure(b, key='y'),
            cirq.measure(a, key='x'),
        ])).all_measurement_keys() == ('y', 'x')  # returns ('x', 'y')
```
This seems like a bug in the Circuit-equality behavior to me, which is outside the scope of this PR. I've filed #3452 to track it.

maffoo:
I think using `@lru_cache` is generally a bad idea for instance methods, because it creates a shared cache on the class for data that should more naturally be memoized per instance. Here that's technically ok but in most cases that I've seen the author intended to have a per-instance cache keyed on the other parameters besides `self` and didn't realize that the cache would be shared. I'd suggest we avoid using `@lru_cache` on instance methods to avoid such confusion. Also there's rarely a good way to pick an appropriate size for such caches. Ack on methods vs. properties, that's too bad.

balopat (changes requested):
Thank you for this - looks very promising! I'm going to have to experiment with it a little probably today but tomorrow definitely. I added an initial scan of nits.

cirq/circuits/frozen_circuit.py
```diff
@@ -0,0 +1,148 @@
...
+
+    @property
+    def moments(self) -> Sequence['cirq.Moment']:
+        return self._moments
+
+    @property
+    def device(self) -> devices.Device:
+        return self._device
+
+    @staticmethod
```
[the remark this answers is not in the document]
95-martin-orion: Moved these to AbstractCircuit.

cirq/circuits/frozen_circuit_test.py
```diff
@@ -0,0 +1,2780 @@
...
+                              cirq.Moment([cirq.X(a)]))
+
+    assert cirq.approx_eq(cirq.FrozenCircuit([cirq.Moment([cirq.X(a)])]),
+                          cirq.FrozenCircuit([cirq.Moment([cirq.X(a)])]))
+    assert not cirq.approx_eq(cirq.FrozenCircuit([cirq.Moment([cirq.X(a)])]),
+                              cirq.FrozenCircuit([cirq.Moment([cirq.X(b)])]))
+
+    assert cirq.approx_eq(
+        cirq.FrozenCircuit([cirq.Moment([cirq.XPowGate(exponent=0)(a)])]),
+        cirq.FrozenCircuit([cirq.Moment([cirq.XPowGate(exponent=1e-9)(a)])]))
```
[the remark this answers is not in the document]
95-martin-orion: Is the `approx` in `approx_eq` referring to types? I don't think a Circuit should register as equal to a FrozenCircuit otherwise - if users want to compare the two, they can `[un]freeze` as necessary.

95-martin-orion:
Pinging for review.

balopat:
> Pinging for review.

Ping acknowledged! I'll do this first thing.

cirq/circuits/frozen_circuit.py
```diff
@@ -0,0 +1,166 @@
...
+
+        # These variables are memoized when first requested.
+        self._num_qubits = None
+        self._qid_shape = None
+        self._all_qubits = None
+        self._all_operations = None
+        self._unitary = None
+        self._has_measurements = None
+        self._all_measurement_keys = None
+        self._are_all_measurements_terminal = None
```
maffoo, line 50: nit: add type annotations on these members and return type annotations on the corresponding methods.

cirq/circuits/frozen_circuit.py
```diff
@@ -0,0 +1,166 @@
...
+    def __rmul__(self, other):
+        return (other * self.unfreeze()).freeze()
+
+    def __pow__(self, other):
+        try:
+            return (self.unfreeze()**other).freeze()
+        except:
+            return NotImplemented
+
+    # pylint: disable=function-redefined
```
maffoo, line 127: Don't need this lint unless you are going to actually redefine the function to specify type overloads. I think it'd be a good idea to add the overloads.

cirq/circuits/frozen_circuit.py
```diff
@@ -0,0 +1,166 @@
...
+            new_device: 'cirq.Device',
+            qubit_mapping: Callable[['cirq.Qid'], 'cirq.Qid'] = lambda e: e,
+    ) -> 'FrozenCircuit':
+        return FrozenCircuit.freeze(self.unfreeze().with_device(
+            new_device, qubit_mapping))
+
+    def _resolve_parameters_(self, param_resolver: 'cirq.ParamResolver'
+                            ) -> 'FrozenCircuit':
+        return FrozenCircuit.freeze(
+            self.unfreeze()._resolve_parameters_(param_resolver))
```
maffoo, line 166: nit: Use the `.freeze()` method call instead:
```python
return self.unfreeze()._resolve_parameters(param_resolver).freeze()
```

cirq/circuits/circuit_test.py
```diff
@@ -918,6 +922,33 @@ def test_insert_into_range():
...
+    cirq.testing.assert_has_diagram(
+        c, """
+x: ───X───X───Z───@───X───X───X───────
+                  │       │
+y: ───T───Y───Y───@───Y───@───H───H───
+""")
+
+    with pytest.raises(IndexError):
+        c.insert_into_range([cirq.CZ(x, y)], 10, 10)
+
```
maffoo, line 951: What are these new insertion tests about?

cirq/circuits/frozen_circuit_test.py
```diff
@@ -0,0 +1,2782 @@
...
+    c = cirq.FrozenCircuit(cirq.Z(cirq.GridQubit(0, 0)), device=cg.Foxtail)
+    cirq.testing.assert_equivalent_repr(c)
+    assert repr(c) == """cirq.FrozenCircuit([
+    cirq.Moment(
+        cirq.Z(cirq.GridQubit(0, 0)),
+    ),
+], device=cirq.google.Foxtail)"""
+
+
+def test_empty_moments():
```
maffoo, line 220: This is almost entirely duplicated from `test_empty_moments` in `circuit_test.py`. It seems like a better way to handle this would be to write a single test that is parameterized over whether to use a frozen circuit, e.g.:
```python
@pytest.mark.parametrize('frozen', [False, True])
def test_empty_moments(frozen):
    ...
    circuit = cirq.Circuit(...)
    if frozen:
        circuit = circuit.freeze()

    cirq.testing.assert_has_diagram(...)
```
This same thing likely applies to much of the code below as well. It would be nice to avoid too much duplication so that going forward we have an easier time keeping tests in sync between Circuit and FrozenCircuit and don't miss testing common behavior on one or the other.

cirq/circuits/frozen_circuit.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def are_all_measurements_terminal(self):
+        if self._are_all_measurements_terminal is None:
+            self._are_all_measurements_terminal = super(
+            ).are_all_measurements_terminal()
+        return self._are_all_measurements_terminal
+
+    # End of memoized methods.
+
+    def __add__(self, other):
```
maffoo, line 108: Should add type annotations on these special methods.

maffoo:
@95-martin-orion, I was able to review the changes other than the method reordering by selecting a range of commits to view in the diff, but it seems there are a few formatting changes mixed in there that didn't get included in the initial reordering. Could we make a separate PR for the reordering to make the diff here easier to review?

cirq/circuits/circuit.py
```diff
@@ -120,63 +97,70 @@ def __init__(self,
...
+    @property
+    @abc.abstractmethod
+    def device(self) -> devices.Device:
+        pass
 
-    def copy(self) -> 'Circuit':
-        copied_circuit = Circuit(device=self._device)
-        copied_circuit._moments = self._moments[:]
-        return copied_circuit
+    def freeze(self) -> 'cirq.FrozenCircuit':
```
balopat, line 111: `freeze()` and `unfreeze()` do not feel right here. 

I see two problems with having them on `AbstractCircuit`:

1. now the abstract class needs to know about its implementors which is really not typical, and introduces a circular dependency in the type hierarchy - AbstractCircuit is now depending on FrozenCircuit which implements it.
2. Also if ever a new type of circuit would be added, this method (thus the class) would have to change - this is a bit against the [Open Closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle) - theoretically you could just plop in a new subclass implement the right methods and be done with it. With the current design you have to somehow "know" that you have to change these two methods as well. 
3. even for discoverability it is confusing: even in the current code you call it a couple of ways `AbstractCircuit.freeze` and `FrozenCircuit.freeze`

My suggestion:  We could make freeze/unfreeze into a protocol: 
`__freeze__`, `__unfreeze__` would be the magic methods: 
```
@abstractmethod
AbstractCircuit.__freeze__
@abstractmethod
AbstractCircuit.__unfreeze__
``` 
And then the two subclasses can implement these magic methods to their liking. 

The freeze function would check for the existence of the private methods like in other Cirq protocols.

balopat (changes requested):
I have two remarks for now: 

1. freeze, unfreeze should become protocols or at least package level methods 
2. there are no tests protecting immutability for FrozenCircuit - I can just add a the method below to FrozenCircuit - making it mutable, and there are no tests broken. 

```
 @device.setter
    def device(self, new_device: 'cirq.Device') -> None:
        new_device.validate_circuit(self)
        self._device = new_device
```

cirq/circuits/circuit.py
maffoo: @balopat, do you imagine other types being freezable/unfreezable besides possible future subclasses of `AbstractCircuit`? Seems like it doesn't merit a full-on protocol because I don't see this as being a very generic operation that we will use over multiple types. Also, FWIW I like instance methods instead of class methods for freezing (e.g. call `my_circuit.freeze()` instead of `AbstractCircuit.freeze(my_circuit)`.

cirq/circuits/frozen_circuit_test.py
95-martin-orion: This is true of most tests in frozen_circuit_test - there's a commit on this PR that literally copy-pastes the tests and replaces `Circuit` with `FrozenCircuit`. I agree that shared tests would be preferable - I'll apply this change.

cirq/circuits/circuit_test.py
95-martin-orion: Our test coverage tools are flagging sections of the `circuit.py` code that I moved in this PR. I assume this code predates our enforcement of test coverage.

cirq/circuits/frozen_circuit.py
95-martin-orion: IIRC, mypy complains about these if I include the overloads: `Overloaded method has both abstract and real implementations`, or something along those lines. I'll remove the pylint-disable comment, though.

cirq/circuits/circuit.py
balopat: Well, technically, I can image a mutable version of Devices, Moments and Operations as well but it is a stretch and I have no evidence for a need to introduce them. So I think you're right about a protocol being overkill here. 

The instance methods make sense - but then would you enforce an implementation for them from the `AbstractCircuit` with an abstract method? I would.

balopat: Actually I just discovered that `PauliString` has a `frozen()` and `mutable_copy()` method. It might be worthwhile to name them similarly - but maybe two types are already enough to warrant a protocol?

balopat: + there is a `freeze_op_tree` helper method that freezes the OP_TREE type.

maffoo: Interesting to have those other examples. I'd still suggest using instance methods here for now, but we could look into making a freeze protocol in the future. I'd want to be sure the semantics are consistent across all the freezable types, e.g. w.r.t. returning new instances vs. changing internal flags to mark things as frozen. The name `mutable_copy` has the advantage over `unfreeze` that it indicates it will make a copy, even if given an object that is already mutable. So perhaps we should use that here instead of `unfreeze`? I don't know which is more important, symmetry between `freeze`/`unfreeze` or indicating to the caller that it makes a copy.

95-martin-orion:
@maffoo:

> @95-martin-orion, I was able to review the changes other than the method reordering by selecting a range of commits to view in the diff, but it seems there are a few formatting changes mixed in there that didn't get included in the initial reordering. Could we make a separate PR for the reordering to make the diff here easier to review?

Which sets of commits are you inspecting? The large-scale reformatting steps should be isolated to three commits:
- `Moves const methods to top of Circuit class.`
- `Run format-incremental.` (Had some YAPF weirdness here)
- `Pacify yapf, mypy, and pylint.` (Fixed YAPF weirdness)

If excluding those from the review doesn't resolve things, I can split off the reordering and restack this PR's commits to be cleaner.

maffoo:
The last last yapf commit you mentioned changes new and old code, so can't be excluded cleanly. I do think splitting off the reordering and rebasing this PR will simplify the review.

95-martin-orion:
> The last last yapf commit you mentioned changes new and old code, so can't be excluded cleanly. I do think splitting off the reordering and rebasing this PR will simplify the review.

Reordering is up as PR #3478. I anticipate some additional changes being required to keep mypy and test coverage happy.

95-martin-orion:
I've opened the back half of this PR as a separate change (#3481) since cleaning the commit history required a force-push.

PR #3450, opened 2020-10-28
Add ion device tutorial.

rmlarose:
Modified from original [Colab](https://colab.research.google.com/drive/1p_SLX83UzudhHLeZ6UXx_GAp67ElxMXW) by Muyuan Li and Ken Brown. Removed BV example as it seemed out of scope.

@balopat @karlunho besides validating and decomposing ops/circuits, is there anything else we want to highlight here?


balopat (changes requested):
Since then we chatted - this should go under educators

balopat (approved):
Actually we can add this in and then move it later.

PR #3453, opened 2020-10-28
Make args_from_proto public

dstrain115:
- This is good to have visible so that end clients can
deserialize args when necessary.
- Came about as part of calibration API work.


balopat (approved):
LGTM

PR #3456, opened 2020-10-28
Move to method dispatch using act_on for CliffordState, Tableau and ChForm

smitsanghavi:
Delete all the now-unneeded methods.

Remaining work: refactor the measurement code

#2948 #2423


cirq/sim/clifford/clifford_simulator.py
```diff
@@ -336,94 +335,20 @@ def wave_function(self):
...
-                               self.qubit_map[op.qubits[1]])
-            self.ch_form._CNOT(self.qubit_map[op.qubits[0]],
-                               self.qubit_map[op.qubits[1]])
-        elif op.gate == cirq.CZ:
-            self.tableau._CZ(self.qubit_map[op.qubits[0]],
-                             self.qubit_map[op.qubits[1]])
-            self.ch_form._CZ(self.qubit_map[op.qubits[0]],
-                             self.qubit_map[op.qubits[1]])
-        else:
+        from cirq.sim.clifford import (ActOnCliffordTableauArgs,
```
balopat, line 338: why local import?

balopat: Oh I guess the `*Args` are referring to this module?

cirq/sim/clifford/clifford_simulator.py
```diff
@@ -37,12 +37,13 @@
...
 import cirq
 from cirq import circuits, study, ops, protocols, value
-from cirq.ops import pauli_gates
 from cirq.ops.clifford_gate import SingleQubitCliffordGate
 from cirq.ops.dense_pauli_string import DensePauliString
-from cirq.protocols import unitary
+from cirq.protocols import act_on, unitary
 from cirq.sim import simulator
-from cirq.sim.clifford import clifford_tableau, stabilizer_state_ch_form
+from cirq.sim.clifford import (act_on_clifford_tableau_args,
```
balopat, line 44: it seems that we could just import the types from these packages in `cirq.sim.clifford.__init__.py` and then use them directly. It is the more typical style in Cirq allowing for shorter names and easier code completion. Then you can do `from cirq.sim import clifford as clifford` and `clifford.ActOn...`

cirq/sim/clifford/clifford_simulator.py
smitsanghavi: I had assumed so, but apparently I was wrong there

cirq/sim/clifford/clifford_simulator.py
smitsanghavi: You're right, and the style guide agrees with you https://cirq.readthedocs.io/en/latest/docs/dev/style.html.

But I feel I have seen the reverse pattern being followed, the most specific subpaths are imported instead of the top-level module (except in tests where it's `cirq.Whatever`). But I'll at least fix the clifford stuff for this PR to follow the style guide.

smitsanghavi: Had to shuffle the `clifford.__init__.py` to avoid circular dependencies. Since these all classes are very closely intertwined, there is a high chance of circular dependencies if using `cirq.sim.clifford` directly.

balopat (approved):

cirq/sim/clifford/clifford_simulator.py
balopat: Yeah now that happens. The init module orderings is a dark art.

PR #3457, opened 2020-10-28
Polish QAOA tutorial

rmlarose:
- Fix links at top of notebook (to Colab, source, etc.)
- Add cost Hamiltonian and short explanation.
- Set seed for reproducibility (get the same problem instance).
- Change text from 2nd person to 3rd person to match other tutorials.
- Small wording changes, typo fixes, added comments.


balopat (changes requested):
I would keep the outputs for now - it is better for staging purposes. Otherwise it looks good!

balopat (approved):
LGTM

PR #3459, opened 2020-10-28
Fix misaligned ``` fences.

MarkDaoust:


balopat:
Thank you for the fix!!! In the meantime the docker section was removed so I think we can close this!

PR #3468, opened 2020-10-29
Fix typo in custom gates guide

rmlarose:
@balopat I found a small typo when checking that the matrix equations rendered correctly (which they indeed do now).


balopat (approved):
Wow, I had to look pretty hard to see that extra `_` :)

PR #3472, opened 2020-11-01
[3 qubit decomposition] Extract diagonal

balopat:
Adds `cirq.two_qubit_matrix_to_diagonal_and_operations` that extracts a diagonal if it can from a 3CNOT unitary to a diagonal and a 2 CNOT unitary. 

Another part of #2873.


cirq/linalg/decompositions.py
```diff
@@ -1045,3 +1046,34 @@ def _gamma(u: np.ndarray) -> np.ndarray:
...
+
+    Args:
+        u: three-CNOT two-qubit unitary
+        atol: the absolute tolerance for avoiding division by zero
+    Returns:
+        diagonal extracted from U
+    """
+    t = _gamma(transformations.to_special(u).T).T.diagonal()
+    k = np.real(t[0] + t[3] - t[1] - t[2])
+    if np.abs(k) < atol:
```
viathor, line 1068: I think you shouldn't need this conditional (and the atol parameter). Working around this singularity is exactly why people came up with [arctan2](https://en.wikipedia.org/wiki/Atan2)! This should work:

```python
psi = np.arctan2(np.imag(np.sum(t)), k)
```

(i.e. remove the conditional, unindent arctan below, replace arctan with arctan2 and "/ k" with ", k"). Note that [numpy's arctan2](https://numpy.org/doc/stable/reference/generated/numpy.arctan2.html) is also well-behaved when both arguments are zero (how cool is that?!).

cirq/linalg/decompositions.py
```diff
@@ -1045,3 +1046,34 @@ def _gamma(u: np.ndarray) -> np.ndarray:
...
+        # exp(-i*psi) (t[0]+t[3]) + exp(i*psi) (t[1]+t[2]) is real
+        # both pi/2 or 3pi/2 can work
+        psi = np.pi / 2
+    else:
+        psi = np.arctan(np.imag(np.sum(t)) / k)
+
+    cnot = block_diag(np.eye(2), np.array([[0, 1], [1, 0]]))
+    rz = np.diag([1, np.exp(1j * psi)])
+
+    return cnot @ combinators.kron(np.eye(2), rz) @ cnot
```
viathor, line 1079: CNOT is a permutation gate so matrix multiplications are unnecessary: it's action on a diagonal matrix is just a swap of the last two elements. Therefore, we can remove `cnot` and `rz` and can instead say

```python
f = np.exp(1j * psi)
return np.diag([1, f, f, 1])
```

cirq/linalg/decompositions.py
```diff
@@ -1045,3 +1046,34 @@ def _gamma(u: np.ndarray) -> np.ndarray:
         u @ yy @ u.T @ yy, where yy = Y ⊗ Y
     """
     return u @ YY @ u.T @ YY
+
+
+def extract_right_diag(u: np.ndarray, atol=1e-15):
```
viathor, line 1051: need type annotation for return value

cirq/linalg/decompositions.py
```diff
@@ -1045,3 +1046,34 @@ def _gamma(u: np.ndarray) -> np.ndarray:
...
+    See Proposition V.2 in Minimal Universal Two-Qubit CNOT-based Circuits.
+    https://arxiv.org/abs/quant-ph/0308033
+
+    Args:
+        u: three-CNOT two-qubit unitary
+        atol: the absolute tolerance for avoiding division by zero
+    Returns:
+        diagonal extracted from U
+    """
+    t = _gamma(transformations.to_special(u).T).T.diagonal()
```
viathor, line 1066: No need to transpose before taking diagonal.

cirq/optimizers/two_qubit_decompositions.py
```diff
@@ -61,6 +64,60 @@ def two_qubit_matrix_to_operations(
...
+        d = np.eye(4)
+        rest = two_qubit_matrix_to_operations(
+            q0,
+            q1,
+            mat,
+            allow_partial_czs=allow_partial_czs,
+            atol=atol,
+            clean_operations=clean_operations)
+
+    return d, rest
```
viathor, line 118: nit: It reads a little better if you employ early return in each branch (since the reader doesn't need to scroll up and down to see what happens next). Early return will also allow you to unindent the final else branch.

cirq/optimizers/two_qubit_decompositions.py
```diff
@@ -61,6 +64,60 @@ def two_qubit_matrix_to_operations(
...
+    Returns:
+        tuple(ops,D): operations `ops`, and the diagonal `D`
+    """
+    if predicates.is_diagonal(mat, atol=atol):
+        rest = []
+        d = mat
+    elif num_cnots_required(mat) == 3:
+        right_diag = extract_right_diag(mat)
+        two_cnot_unitary = mat @ right_diag
+        # note that this implies that two_cnot_unitary @ d = mat
```
viathor, line 99: Optional: If you think we can accept the performance cost, you can say that in python using an assert. I'm not sure whether this is going to be called from hot code or not.

viathor (changes requested):
Sorry for the delay in review! It's very cool to have this in cirq! Thank you! See comments below.

cirq/linalg/decompositions.py
balopat: Dude, that is so cool, very nice, I didn't know about that!!! :)

cirq/linalg/decompositions.py
```diff
@@ -21,8 +21,9 @@
 import math
 import cmath
 import numpy as np
-import scipy
 import matplotlib.pyplot as plt
+import scipy
+from scipy.linalg.special_matrices import block_diag
```
viathor, line 26: remove line to placate linter

cirq/linalg/decompositions.py
```diff
@@ -1045,3 +1046,25 @@ def _gamma(u: np.ndarray) -> np.ndarray:
         u @ yy @ u.T @ yy, where yy = Y ⊗ Y
     """
     return u @ YY @ u.T @ YY
+
+
+def extract_right_diag(u: np.ndarray, atol=1e-15) -> np.ndarray:
```
viathor, line 1051: remove unused atol from signature and Args section of docstring below

viathor (approved):
This is very cool! Thank you!

balopat:
Thank you for the thorough and insightful review @viathor!!

PR #3478, opened 2020-11-04
Reorder const methods in Circuit

95-martin-orion:
Preliminary step to ease review on PR #3445. This PR moves const methods to the top of the `Circuit` class to ease the transition to AbstractCircuits.


cirq/contrib/noise_models/noise_models_test.py
```diff
@@ -26,6 +26,7 @@
 from cirq import ops
 
 
+
```
maffoo, line 29: nit: remove this change.

maffoo (approved):
LGTM

cirq/contrib/noise_models/noise_models_test.py
95-martin-orion: Done. Thanks for the quick review!

PR #3480, opened 2020-11-04
Add serialization of tokens

dstrain115:
- Adds tokens as a seperate proto field.
- Adds a constant table for lookup of tokens.
- Constant is a symbol table where we can store constant strings.
It is its own message, so it could potentially be used for other
objects (such as gates) in the future, but for now, it is only used
to store tokens.

Note: symbol is already used for sympy symbols, so used the
phrase Constant instead.


cirq/google/api/v2/program.proto
```diff
@@ -99,6 +111,16 @@ message Operation {
 
   // Which qubits the operation acts on.
   repeated Qubit qubits = 3;
+
+  // Token that can be used to specify a version of a gate.
+  // For instance, a gate that has been calibrated for a circuit.
+  //
+  // The token can be specified as a string or as a reference to
+  // the constant table of the circuit.
+  oneof arg {
```
maffoo, line 120: Maybe call the overall oneof field "token" instead? Also might be nice to use a common prefix on the fields to make clear they relate to the token:
```
oneof token {
  string token_value = 4;
  int32 token_constant_index = 5;
}
```

cirq/google/op_deserializer.py
```diff
@@ -85,25 +88,38 @@ def __init__(
...
         qubits = [v2.qubit_from_proto_id(q.id) for q in proto.qubits]
         args = self._args_from_proto(
             proto, arg_function_language=arg_function_language)
         if self.num_qubits_param is not None:
             args[self.num_qubits_param] = len(qubits)
         gate = self.gate_constructor(**args)
-        return self.op_wrapper(gate.on(*qubits), proto)
+        op = self.op_wrapper(gate.on(*qubits), proto)
+        if self.auto_deserialize_tokens:
+            if proto.HasField('constant_index') and constants:
```
maffoo, line 116: I would suggest checking which case is populated with `WhichOneof`:
```python
which = proto.WhichOneof('token')
if which == 'token_constant_index':
    ...
elif which == 'token_value':
    ...
```
Also, if a constant index is specified but `constants` is empty, we should probably raise an error.

cirq/google/op_serializer.py
```diff
@@ -68,7 +71,8 @@ def __init__(self,
                  serialized_gate_id: str,
                  args: List[SerializingArg],
                  can_serialize_predicate: Callable[['cirq.Operation'], bool] =
-                 lambda x: True):
+                 lambda x: True,
+                 auto_serialize_tokens: Optional[bool] = True):
```
maffoo, line 75: nit: It seems to me "auto" does not add much here; could we call this just `serialize_tokens`? Similarly for the deserializer above, could rename `auto_deserialize_tokens` -> `deserialize_tokens`.

cirq/google/op_serializer.py
```diff
@@ -128,6 +137,22 @@ def to_proto(
...
+                    if constants is not None:
+                        try:
+                            constant = v2.program_pb2.Constant()
+                            constant.string_value = tag.token
+                            msg.constant_index = constants.index(constant)
+                        except ValueError:
+                            # Token not found, add it to the list
+                            msg.constant_index = len(constants)
+                            new_constant = v2.program_pb2.Constant()
+                            new_constant.string_value = tag.token
```
maffoo, line 152: Could reuse the same constant object instead of recreating:
```python
constant = v2.program_pb2.Constant()
constant.string_value = tag.token
try:
    msg.constant_index = constants.index(constant)
except ValueError:
    msg.constant_index = len(constants)
    constants.append(constant)
```

cirq/google/op_serializer_test.py
```diff
@@ -454,3 +454,69 @@ def test_defaults_not_serialized():
...
+            'id': '1_2'
+        }],
+        'constant_index': 0
+    })
+    constants = []
+    assert expected == serializer.to_proto(
+        GateWithAttribute(0.125)(q).with_tags(tag), constants=constants)
+    constant = v2.program_pb2.Constant()
+    constant.string_value = 'my_token'
+    assert constants == [constant]
```
maffoo, line 522: Would be good to test that if you serialize an op and already have a matching constant it will use the same index.

cirq/google/serializable_gate_set.py
```diff
@@ -114,9 +116,13 @@ def serialize(self,
...
         msg.language.gate_set = self.gate_set_name
         if isinstance(program, circuits.Circuit):
+            constants: Optional[List[v2.program_pb2.Constant]] = (
+                [] if use_constants_table_for_tokens else None)
             self._serialize_circuit(program,
                                     msg.circuit,
-                                    arg_function_language=arg_function_language)
+                                    arg_function_language=arg_function_language,
+                                    constants=constants)
+            msg.constants.extend(constants)
```
maffoo, line 125: Should guard here:
```python
if constants is not None:
    msg.constants.extend(constants)
```

maffoo (commented):
This looks great, just a few minor comments.

cirq/google/api/v2/program.proto
dstrain115: Done.

cirq/google/op_deserializer.py
dstrain115: Done.

cirq/google/op_serializer.py
dstrain115: Done.

cirq/google/op_serializer.py
dstrain115: Done.

cirq/google/op_serializer_test.py
dstrain115: Done.

cirq/google/serializable_gate_set.py
dstrain115: Done.

maffoo (approved):
LGTM!

balopat (approved):
LGTM

PR #3481, opened 2020-11-04

95-martin-orion:
Partial reupload of #3445, following the method reordering submitted in #3478.

Commit history has been cleaned up to remove a misconfigured YAPF run from the previous upload. The following items have also been addressed:

@balopat 
> There are no tests protecting immutability for FrozenCircuit - I can just add a the method below to FrozenCircuit - making it mutable, and there are no tests broken.

Tests have been added to ensure `moments` and `devices` cannot be set with `=`.

@maffoo 
> [frozen_circuit_test.py] is almost entirely duplicated from {...} circuit_test.py. It seems like a better way to handle this would be to write a single test that is parameterized over whether to use a frozen circuit, e.g. {...}

Shared tests have been converted to parameterized tests in `circuit_test.py`. `frozen_circuit_test.py` now only contains tests for FrozenCircuit-specific behavior.


cirq/circuits/circuit.py
```diff
@@ -68,74 +73,159 @@ class Circuit:
...
+                             strategy=InsertStrategy.EARLIEST,
+                             device=self.device)
+
+    def unfreeze(self) -> 'cirq.Circuit':
+        '''Creates a Circuit from this circuit.
+
+        If 'self' is a Circuit, the original object is returned.
+        '''
+        if isinstance(self, Circuit):
+            return self
```
maffoo, line 129: There was an earlier [discussion](https://github.com/quantumlib/Cirq/pull/3445#discussion_r516197448) about a general protocol for "freezing" objects. The analogous method on `PauliString` is [`mutable_copy`](https://github.com/quantumlib/Cirq/blob/master/cirq/ops/pauli_string.py#L330) which makes it clear that a copy will be produced if needed, so that the resulting object can be modified as desired without affecting any existing objects. Without that guarantee I think `unfreeze` will be hard to use safely, so I'd suggest returning a copy here. We can think about a more generic protocol but I think that is out of scope for this PR.

cirq/circuits/circuit.py
```diff
@@ -68,74 +73,159 @@ class Circuit:
...
+    def __getitem__(self, key: Tuple[int, 'cirq.Qid']) -> 'cirq.Operation':
+        pass
+
+    @overload
+    def __getitem__(self,
+                    key: Tuple[int, Iterable['cirq.Qid']]) -> 'cirq.Moment':
+        pass
+
+    @overload
+    def __getitem__(self, key: slice):
```
maffoo, line 178: Would be nice to add return type annotation here. We'd need a self-type annotation:
```python
Self = TypeVar('Self', bound='AbstractCircuit')
...

@overload
def __getitem__(self: Self, key: slice) -> Self:
    pass
```

cirq/circuits/circuit.py
```diff
@@ -68,74 +73,159 @@ class Circuit:
...
+
+    def __getitem__(self, key):
+        if hasattr(key, '__index__'):
+            return self.moments[key]
+        if isinstance(key, tuple):
+            if len(key) != 2:
+                raise ValueError('If key is tuple, it must be a pair.')
+            moment_idx, qubit_idx = key
+            # qubit_idx - Qid or Iterable[Qid].
+            return self.moments[moment_idx][qubit_idx]
```
maffoo, line 197: Does this actually work if `moment_idx` is a slice? It seems we need to slice the qubits on each moment in the sequence `self.moments[moment_idx]` and then reassemble into the correct circuit type.

Edit: ok, I see that `Circuit` and `FrozenCircuit` implement some of the `__getitem__` logic and then fallback to this `__getitem__` implementation. I think it might be more clear to put all the logic here and call into helper methods that must be implemented on the implementing classes as needed, e.g. we build all the moments here and then call `self._with_sliced_moments(moments)` to construct the new instance.

cirq/circuits/circuit.py
```diff
@@ -68,74 +73,159 @@ class Circuit:
...
 
+    def __repr__(self) -> str:
+        cls_name = self.__class__.__name__
+        if not self.moments and self.device == devices.UNCONSTRAINED_DEVICE:
+            return f'cirq.{cls_name}()'
+
+        if not self.moments:
+            return f'cirq.{cls_name}(device={self.device!r})'
+
+        moment_str = _list_repr_with_indented_item_lines(self.moments)
```
maffoo, line 214: nit: call this `moment_repr`

cirq/circuits/circuit_test.py
```diff
@@ -77,43 +83,51 @@ def test_setitem():
     ])
 
 
-def test_equality():
+@pytest.mark.parametrize('frozen', [False, True])
+def test_equality(frozen):
```
maffoo, line 87: Might be more clear to parametrize over the circuit class directly:
```python
@pytest.mark.parametrize('circuit_cls', [cirq.Circuit, cirq.FrozenCircuit])
def test_equality(circuit_cls):
    ...
    eq.add_equality_group(
        circuit_cls(),
        circuit_cls(device=cirq.UNCONSTRAINED_DEVICE),
        ...
```

cirq/circuits/circuit_test.py
```diff
@@ -277,23 +310,26 @@ def test_add_iadd_equivalence():
...
+def test_bool(frozen):
+    assert not MaybeFrozenCircuit(frozen)
+    assert MaybeFrozenCircuit(frozen, cirq.X(cirq.NamedQubit('a')))
 
 
-def test_repr():
-    assert repr(cirq.Circuit()) == 'cirq.Circuit()'
+@pytest.mark.parametrize('frozen', [False, True])
+def test_repr(frozen):
+    type_name = 'FrozenCircuit' if frozen else 'Circuit'
```
maffoo, line 321: Here, could parametrize over the circuit class and name:
```python
@pytest.mark.parametrize('circuit_cls, type_name',
    [(cirq.Circuit, 'Circuit'), (cirq.FrozenCircuit, 'FrozenCircuit')])
def test_repr(circuit_cls, type_name):
    ...
```

maffoo (commented):
This is looking great! Thanks for splitting out the method reordering; it's much easier to review now.

cirq/circuits/circuit.py
95-martin-orion: Is there a reason to prefer this structure over `def __getitem__(...) -> 'cirq.AbstractCircuit'`, which we've used for `Circuit` elsewhere in this file?

cirq/circuits/circuit_test.py
95-martin-orion: Opting to use `circuit_cls.__name__` instead.

cirq/circuits/circuit.py
maffoo: I think it's nice to preserve that type information when slicing otherwise simple things may fail to type check like:
```python
c = cirq.Circuit(...)
c = c[:-1]  # error! cannot assign `AbstractCircuit` to variable of type `Circuit`
```
I totally made up that error message, but you get the idea :-)

cirq/circuits/circuit.py
95-martin-orion: I was originally worried about having `unfreeze` return a copy while `freeze` returns the original, but given that `FrozenCircuit` is immutable I don't think there's any actual issue there. Will update tests to match.

cirq/circuits/circuit.py
95-martin-orion: Ah, I see the difference now. Fix applied.

cirq/circuits/circuit.py
```diff
@@ -68,74 +74,159 @@ class Circuit:
...
+            return self._with_sliced_moments(sliced_moments)
+        if hasattr(key, '__index__'):
+            return self.moments[key]
+        if isinstance(key, tuple):
+            if len(key) != 2:
+                raise ValueError('If key is tuple, it must be a pair.')
+            moment_idx, qubit_idx = key
+            # moment_idx - int or slice; qubit_idx - Qid or Iterable[Qid].
+            selected_moments = self.moments[moment_idx]
+            if isinstance(selected_moments, Sequence):
```
maffoo, line 186: nit: I'd suggest reordering the logic here to return early in the simple case:
```python
selected_moments = ...
if isinstance(selected_moments, ops.Moment):
    return selected_moments[qubit_idx]
if isinstance(qubit_idx, ops.Qid):
    qubit_idx = [qubit_idx]
sliced_moments = [...]
return self._with_sliced_moments(sliced_momens)
```

maffoo (approved):
One minor comment, otherwise LGTM!

cirq/circuits/frozen_circuit_test.py
```diff
@@ -0,0 +1,56 @@
+# Copyright 2018 The Cirq Developers
```
balopat, line 1:
```suggestion
# Copyright 2020 The Cirq Developers
```

balopat (approved):
LGTM with a tiny nit, thank you both @95-martin-orion  and @maffoo !!!!

95-martin-orion:
Thank you for the thorough review!

95-martin-orion:
For bookkeeping: this is part of #3235.

Issue #3323, opened 2020-09-15
Engine integration tests


balopat:
It would be useful to have integration tests against the Engine API to make sure that things work end to end.
Now with the cirq-infra GCP project (#3317) we could have a simple job as part of our CI suite that could use a service account key to connect to the qsim "processor". These will be slow, so we are not aiming for high coverage, just for key scenarios.

balopat:
Also, if we create a good integration testing environment, we can run the Google notebooks against them as well.
The problem with the qsim "processor" is that for example reservations are not supported for it. 
Maybe we could have a fake processor implementation for integration tests? @wcourtney what do you think?

wcourtney:
> Also, if we create a good integration testing environment, we can run the Google notebooks against them as well.
> The problem with the qsim "processor" is that for example reservations are not supported for it.
> Maybe we could have a fake processor implementation for integration tests? @wcourtney what do you think?

I had the same initial thought when you framed the problem. We could create a test processor that happens to route requests to a qsim backend and is visible only to test infrastructure. A couple of things about reservations: 
1. I don't expect cirq to be the primary flow for reservations and the go/quantum-reserve notebook is deprecated.I'd treat this case as a lower priority.
2. Reservation testing could conflict if there are multiple tests running simultaneously, so we'd want to be careful about designing for that case.

dstrain115:
@wcourtney Should we still do this?  I am skeptical that this is feasible.

wcourtney:
This hasn't received any traction and if it were done, should be done outside of the the (public) Cirq repo due to the limited access nature of the service. Closing as obsolete/won't fix.

PR #3484, opened 2020-11-06
Fix notebook authentication

balopat:
Makes gcloud authentication in Google notebooks more instructive for users.  

Currently, if you are not authenticated locally, and try to use the colab authentication in Jupyter notebooks, they fail, as colab's auth is expected to run in a container, e.g. [stores the credentials](https://github.com/googlecolab/colabtools/blob/master/google/colab/auth.py#L111) to `/` or wherever `$DATA_LAB_ROOT` is pointing to. 

- Keeps the colab based auth only in case colab is the execution environment. Otherwise it assumes gcloud based authentication ("advanced usage"). 

- Also, I added empty outputs for each of the Google notebooks so that the devsite pipeline doesn't break on trying to executing them. After https://github.com/quantumlib/Cirq/issues/3323 we will be able to run these notebooks against a test environment automatically. 

Previews in colab:

https://colab.sandbox.google.com/github/balopat/Cirq/blob/fix_notebook_authentication/docs/tutorials/google/start.ipynb
https://colab.sandbox.google.com/github/balopat/Cirq/blob/fix_notebook_authentication/docs/tutorials/google/reservations.ipynb
https://colab.sandbox.google.com/github/balopat/Cirq/blob/fix_notebook_authentication/docs/tutorials/google/colab.ipynb


dstrain115 (approved):
Only comment is that there seems to be a lot of repetition.  I wonder if we should put a standard "authenticate_for_notebook" type function in our code base somewhere.  This could be done later though.

balopat:
> Only comment is that there seems to be a lot of repetition. I wonder if we should put a standard "authenticate_for_notebook" type function in our code base somewhere. This could be done later though.

I agree! It could be part of a `cirq.google.ipython` package. It would be useful for users as well as if the number of google colabs grow, but for this three we should be good for now. It would also require to switch over to cirq-unstable until we release the functionality.

PR #3485, opened 2020-11-07
Remove todos from protocols documentation

### LINKED ISSUE #2973: Document protocols, how they work, and how they interoperate
Brought up in #1815 we don't have a good location for looking at all the protocols and their signatures (and #2939 means the classes for the protocols are essentially undocumented).  We should make documentation for the protocols since they are less discoverable.

balopat:
Removes the TODOs from the protocol documentation.

Added a bit of a description of what a protocol actually is + a table of notable protocols + a description of the channel / mixture / unitary operator representation protocols.

Fixes #2973.


docs/protocols.ipynb
```diff
@@ -181,16 +145,112 @@
...
    "source": [
-    "## What is a Protocol\n",
+    "## What is a protocol? \n",
+    "\n",
+    "A protocol is a combination of the following two items: \n",
+    "- a `SupportsXYZ` class, which defines and documents all the magic functions that need to be implemented in order to support that given protocol \n",
+    "- the entrypoint function(s), which are exposed to the main cirq namespace as `cirq.xyz()`\n",
     "\n",
-    "TODO: describe the class SupportsXYZ, private methods, the entry-point cirq.xyz and the fallback strategies.\n",
+    "Note: While the protcol is technically both of these things, we refer to the public utility functions interchangeably as protocols. See the list of them below.\n",
```
dstrain115, line 154: protcol -> protocol.

docs/protocols.ipynb
```diff
@@ -181,16 +145,112 @@
...
     "\n",
     "\n",
     "## Cirq's protocols\n",
     "\n",
-    "TODO: draw a nice diagram of what areas the current protocols belong to and how they are interrelated\n",
+    "Cirq's protocols can be useful in multiple feature areas of quantum computing. Here's a list of all of them. \n",
+    "\n",
+    "| Protocol | Areas | \n",
+    "|----------|-------|\n",
+    "|`cirq.act_on`| simulation |\n",
```
dstrain115, line 163: Should we add a short description of each protocol?  

Also, these areas are a bit fuzzy and awkward.  I use cirq.unitary for analysis, for instance.  I guess it is analysis of a simulation, but that means that "simulation" covers a broad ground.  Also, this is referred to as "Quantum operator representation" in the next section, which is not really simulation.

I wonder if it would just be better to have a short description of the function and people can use them how they want.

docs/protocols.ipynb
```diff
@@ -181,16 +145,112 @@
...
+    "\n",
+    "The following family of protocols is an important and frequently used set of features of Cirq and it is worthwhile mentioning them and and how they interact with each other. They are, in the order of generality (starting with the least generic):\n",
+    "\n",
+    "* `*unitary`\n",
+    "* `*mixture`\n",
+    "* `*channel`\n",
+    "\n",
+    "All these protocols make it easier to work with different representations of quantum operators: finding that representation (`unitary`, `mixture`, `channel`), determining whether the operator has that representation (`has_*`) and applying them (`apply_*`) on a state vector. \n",
+    "\n",
+    "The `unitary` representation is the most special case, only applies when the quantum operator is unitary, and in this case represents it via the unitary matrix in the computational basis. We saw an example of the unitary protocol above.\n",
```
dstrain115, line 205: This sentence is very long and the phrase "only applies when the quantum operator is unitary," seems out of place.  You might want to consider splitting it into multiple sentences.

docs/protocols.ipynb
```diff
@@ -181,16 +145,112 @@
...
+    "\n",
+    "The `mixture` representation is a probabilistic representation of a quantum operator, where each possible unitary effect $U_i$ occurs with a certain probability $p_i$, where $\\sum p_i = 1$ In case the operator does not implement `SupportsMixture`, but does implement `SupportsUnitary`, `*mixture` protocols fall back to the `*unitary` methods, based on the simple correspondence: the $U$ is a mixture of a single unitary with probability $p=1$. \n"
```
dstrain115, line 207: It might be worth explaining what each of these return (numpy array for unitary, for instance), since we are going into these methods in some depth.

docs/protocols.ipynb
```diff
@@ -181,16 +145,112 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "id": "5f9ec1e69ba6"
+   },
+   "outputs": [],
+   "source": [
+    "# cirq.Y has a unitary effect but does not implement SupportsMixture\n",
+    "print(cirq.mixture(cirq.Y))\n",
+    "print(cirq.has_mixture(cirq.Y))"
```
dstrain115, line 220: It might be worth showing an operator that actually has a mixture as a more illustrative example.  Same with channel below.

docs/protocols.ipynb
```diff
@@ -181,16 +145,300 @@
...
     "## Cirq's protocols\n",
     "\n",
-    "TODO: draw a nice diagram of what areas the current protocols belong to and how they are interrelated\n",
+    "For a complete list of Cirq protocols, refer to the `cirq.protocols` package. \n",
+    "Here we provide a list of frequently used protocols for debugging, simulation and testing.\n",
+    "\n",
+    "\n",
+    "| Protocol | Description | \n",
+    "|----------|-------|\n",
+    "|`cirq.apply_channel`| High performance evolution under a channel evolution. |\n",
```
mpharrigan, line 165: what's the plan for keeping this up to date?

balopat: Well...I did generate the full set and then trimmed it, so it could be generated very easily. 

```
for o in dir(cirq.protocols):
    obj = eval(f"cirq.protocols.{o}")
    if callable(obj) and not isinstance(obj, type):
        docs = obj.__doc__
        if docs is not None:
            first_line = docs.split('\n')[0]
            print(f"|`cirq.{o}`| {first_line} |")
        else:
            raise ValueError(f"{o} has no public docs and it's a protocol function!")
```

But I would think that the "notable" category for users is not going to change that much, and if we want to show case a new protocol we can include it here. 

Of course a better version would be to have an annotation `@showcase` or similar, that would then be fed to the generator... I'll open an issue for this, but I think for now this can go in this way. WDYT?

balopat: https://github.com/quantumlib/Cirq/issues/3512

mpharrigan: Yes, it's probably true that they won't change enough to merit anything fancy. Just have to make sure it doesn't change *so* infrequently, we forgot that we documented and it becomes stale like 2+ years from now with us forgetting about it :)

balopat: Agreed, added #3512

docs/protocols.ipynb
```diff
@@ -181,16 +145,300 @@
...
+    "|`cirq.trace_distance_from_angle_list`| Given a list of arguments of the eigenvalues of a unitary matrix, |\n",
+    "|`cirq.unitary`| Returns a unitary matrix describing the given value. |\n",
+    "|`cirq.validate_mixture`| Validates that the mixture's tuple are valid probabilities. |\n",
+    "\n",
+    "\n",
+    "### Quantum operator representation protocols\n",
+    "\n",
+    "The following family of protocols is an important and frequently used set of features of Cirq and it is worthwhile mentioning them and and how they interact with each other. They are, in the order of increasing generality:\n",
+    "\n",
+    "* `*unitary`\n",
```
dstrain115, line 208: How come the protocols are denoted "*unitary" as opposed to just "unitary" or "cirq.unitary"?

dstrain115 (approved):
LGTM!

docs/protocols.ipynb
balopat: Well, with the definition of a protocol above, being a single class defining the magic methods to implement and the helper methods around it, technically has_unitary, apply_unitary and unitary are three separate protocols: SupportsExplicitHasUnitary, SupportsConsistentApplyUnitary, SupportsUnitary. The same applies to the others.

PR #3489, opened 2020-11-10
Remove "python" symbolic link

kevinsung:
It appears this was added by mistake in #3458 (right?)


vtomole (approved):

balopat:
Yupp, thank you for noticing it - the oneliner totally skipped my eye during review!

PR #3495, opened 2020-11-11
Add measurement key remapping protocol.

95-martin-orion:
Related to #3235. This is necessary for encapsulating FrozenCircuits in Gates; if a FrozenCircuit contains a measurement, any Gate that contains that FrozenCircuit cannot be reused in the same Circuit without remapping its measurement keys.

### DIFF
--- a/cirq/circuits/circuit.py
+++ b/cirq/circuits/circuit.py
@@ -1392,6 +1392,14 @@ class Circuit(AbstractCircuit):
 
     __hash__ = None  # type: ignore
 
+    def _with_measurement_key_mapping_(self, key_map: Dict[str, str]):
+        remapped_ops = [
+            protocols.with_measurement_key_mapping(op, key_map)
+            if protocols.is_measurement(op) else op
+            for op in self.all_operations()
+        ]
+        return Circuit(remapped_ops, device=self.device)
+
     def with_device(
             self,
             new_device: 'cirq.Device',
--- a/cirq/circuits/circuit.py
+++ b/cirq/circuits/circuit.py
@@ -815,6 +815,14 @@ class AbstractCircuit(abc.ABC):
     def all_measurement_keys(self) -> AbstractSet[str]:
         return protocols.measurement_keys(self)
 
+    def _with_measurement_key_mapping_(self, key_map: Dict[str, str]):
+        remapped_ops = [
+            protocols.with_measurement_key_mapping(op, key_map)
+            if protocols.is_measurement(op) else op
+            for op in self.all_operations()
+        ]
+        return type(self)(remapped_ops, device=self.device)
+
     def _qid_shape_(self) -> Tuple[int, ...]:
         return self.qid_shape()
 
@@ -1391,14 +1399,6 @@ class Circuit(AbstractCircuit):
         return cirq.Circuit(inv_moments, device=self._device)
 
     __hash__ = None  # type: ignore
-
-    def _with_measurement_key_mapping_(self, key_map: Dict[str, str]):
-        remapped_ops = [
-            protocols.with_measurement_key_mapping(op, key_map)
-            if protocols.is_measurement(op) else op
-            for op in self.all_operations()
-        ]
-        return Circuit(remapped_ops, device=self.device)
 
     def with_device(
             self,
--- a/cirq/circuits/circuit.py
+++ b/cirq/circuits/circuit.py
@@ -815,14 +815,6 @@ class AbstractCircuit(abc.ABC):
     def all_measurement_keys(self) -> AbstractSet[str]:
         return protocols.measurement_keys(self)
 
-    def _with_measurement_key_mapping_(self, key_map: Dict[str, str]):
-        remapped_ops = [
-            protocols.with_measurement_key_mapping(op, key_map)
-            if protocols.is_measurement(op) else op
-            for op in self.all_operations()
-        ]
-        return type(self)(remapped_ops, device=self.device)
-
     def _qid_shape_(self) -> Tuple[int, ...]:
         return self.qid_shape()
 
@@ -1399,6 +1391,14 @@ class Circuit(AbstractCircuit):
         return cirq.Circuit(inv_moments, device=self._device)
 
     __hash__ = None  # type: ignore
+
+    def _with_measurement_key_mapping_(self, key_map: Dict[str, str]):
+        remapped_ops = [
+            protocols.with_measurement_key_mapping(op, key_map)
+            if protocols.is_measurement(op) else op
+            for op in self.all_operations()
+        ]
+        return Circuit(remapped_ops, device=self.device)
 
     def with_device(
             self,
--- a/cirq/circuits/circuit.py
+++ b/cirq/circuits/circuit.py
@@ -815,6 +815,12 @@ class AbstractCircuit(abc.ABC):
     def all_measurement_keys(self) -> AbstractSet[str]:
         return protocols.measurement_keys(self)
 
+    def _with_measurement_key_mapping_(self, key_map: Dict[str, str]):
+        return self._with_sliced_moments([
+            protocols.with_measurement_key_mapping(moment, key_map)
+            for moment in self.moments
+        ])
+
     def _qid_shape_(self) -> Tuple[int, ...]:
         return self.qid_shape()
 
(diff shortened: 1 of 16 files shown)


cirq/protocols/measurement_key_protocol.py
```diff
@@ -143,3 +150,8 @@ def is_measurement(val: Any) -> bool:
     a non-empty result for them.
     """
     return bool(measurement_keys(val))
+
+
+def with_measurement_key_mapping(val: Any, key_map: Dict[str, str]):
```
balopat, line 155: nit: Add docstrings to public method exposed to the cirq namespace

cirq/circuits/circuit.py
```diff
@@ -1392,6 +1392,14 @@ def __pow__(self, exponent: int) -> 'Circuit':
 
     __hash__ = None  # type: ignore
 
+    def _with_measurement_key_mapping_(self, key_map: Dict[str, str]):
```
balopat, line 1395: it feels like this  and the FrozenCircuit version could be deduped into `AbstractCircuit` and use type(self) to instantiate?

balopat (approved):
LGTM with a nit and a question

cirq/circuits/circuit.py
95-martin-orion: Done.

cirq/protocols/measurement_key_protocol.py
95-martin-orion: Done.

95-martin-orion:
Had to update the protocol methods to handle NotImplemented passthrough properly - PTAL.

cirq/circuits/circuit.py
95-martin-orion: Discovered why we can't dedupe these: `AbstractCircuit` has no `__init__` method, so mypy doesn't like it.

balopat: OK. Alternative solutions:
1. We could create an @abstractmethod __init__ - not sure I like that
2. We could have a `with_operations` abstract method that would be implemented in the subclasses as `Circuit(ops, device=self.device`.
WDYT?

cirq/ops/gate_operation.py
```diff
@@ -63,6 +63,14 @@ def with_gate(self, new_gate: 'cirq.Gate') -> 'cirq.Operation':
             return self
         return new_gate.on(*self.qubits)
 
+    def _with_measurement_key_mapping_(self, key_map: Dict[str, str]):
+        new_gate = protocols.with_measurement_key_mapping(self.gate, key_map)
+        if new_gate is NotImplemented:
+            return NotImplemented
+        if new_gate is self.gate:
```
balopat, line 70: I think this should be `==` ?

cirq/ops/measurement_gate.py
```diff
@@ -79,6 +79,11 @@ def with_key(self, key: str) -> 'MeasurementGate':
                                invert_mask=self.invert_mask,
                                qid_shape=self._qid_shape)
 
+    def _with_measurement_key_mapping_(self, key_map: Dict[str, str]):
+        if self.key not in key_map:
+            return self
```
balopat, line 84: I'm not entirely sure that this is the right way to implement this. 
It is kind of surprising to me that the object returns itself as a result of a `with_*` method and makes it harder to reason about code and mutability.

cirq/ops/gate_operation.py
balopat: Oh scratch that - I see what you are doing now - if the mapping is not applicable, the measurement gate is going to be the same instance, the operation as well. I'll resolve this and let's continue the discussion on the other comment about whether a `with_*` method should return `self` ever or not

cirq/ops/measurement_gate.py
95-martin-orion: I was conflicted on this as well. What threw me off is the `GateOperation.with_gate` method, which returns `self` if nothing changed...but that's the exception, and I've never actually seen that method called. The only other appearance of `with_gate` in the code is for `ParallelGateOperation`, and that version does not return self. Maybe `GateOperation.with_gate` should be changed?

I'll go ahead and convert `_with_measurement_key_mapping_` to return a new object for now.

balopat: > Maybe `GateOperation.with_gate` should be changed?

I think so!!!

cirq/circuits/circuit.py
95-martin-orion: Option (2) seems right. I'm also realizing that `Circuit(circuit.ops)` might not preserve the moment structure of the original circuit, so we may want to reuse the existing `_with_sliced_moments` method for this.

cirq/ops/measurement_gate.py
95-martin-orion: Discussion from #3499: `GateOperation` can safely return `self` from `with`-functions, as it is immutable. I'll document this more clearly and continue returning `self` in `_with_measurement_key_mapping_`.

cirq/circuits/circuit.py
balopat: That is a great point - this whole PR is missing remapping a Moment!!! That's the definitely the level where we should operate.

95-martin-orion: Protocol method is now provided for moments, and tests are added to ensure that Circuits correctly preserve the moment structure during remapping.

balopat:
Good to go!!!

maffoo:
It would be nice to be able to include the same FrozenCircuit _instance_ in multiple places, so that you could still get the benefits of more compact serialization even while remapping keys (also, internally we use identity of these instances to indicate where waveforms can be reused; if remapping measurement keys creates a new instance we lose this signal). Did you think about whether it would be possible to do this? I think it would require making the simulators aware of circuit operations, because they'd need to keep track of the remapping as they step through the circuit so they can record measurements in the correct place when they encounter a measurement gate inside a circuit operation.

95-martin-orion:
> It would be nice to be able to include the same FrozenCircuit _instance_ in multiple places {...}

That aligns with my intent. The eventual plan is for CircuitOperations to capture measurement key mappings for their constituent FrozenCircuit without immediately applying them; those mappings can later be applied via this protocol during decomposition (e.g. for simulators).

A mock example:
```python
q1, q2 = cirq.LineQubit.range(2)
fc = FrozenCircuit(cirq.H(q1), cirq.measure(q1, key='m1'))

c = cirq.Circuit(
  cirq.CircuitOperation(fc).on(q1),
  cirq.CircuitOperation(fc).on(q2).with_key_map('m1': 'm2'),
)

assert str(c) == """\
q1 --CircuitOperation(fc)---------------------------

q2 --CircuitOperation(fc).with_key_map('m1': 'm2')--
"""

assert cirq.decompose(c) == [
  cirq.H(q1), cirq.measure(q1, key='m1'),
  cirq.H(q2), cirq.measure(q2, key='m2'),
]
```

daxfohl:
@95-martin-orion Would it be possible instead to allow `CircuitOperation` itself to have a key?

```python
q1, q2 = cirq.LineQubit.range(2)
fc = FrozenCircuit(cirq.H(q1), cirq.measure(q1, key='m'))

c = cirq.Circuit(
  cirq.CircuitOperation(fc).on(q1).key('x'),
  cirq.CircuitOperation(fc).on(q2).key('y'),
)

assert str(c) == """\
q1 --CircuitOperation(fc).key('x')--

q2 --CircuitOperation(fc).key('y')--
"""

assert cirq.decompose(c) == [
  cirq.H(q1), cirq.measure(q1, key='x.m'),
  cirq.H(q2), cirq.measure(q2, key='y.m'),
]
```

I think having some good way to scope measurement keys is going to be important when starting on flow control etc. This would be an easy way to prevent conflicts anywhere in the whole circuit, which may be difficult if you're using third-party subcircuits or something.

95-martin-orion:
> @95-martin-orion Would it be possible instead to allow `CircuitOperation` itself to have a key?

Key-remapping has already been submitted (#3580), and is a superset of the behavior you describe. If you see a use case for it, you could implement `CircuitOperation.key()` entirely within the existing behavior:
```
def key(self, k: str):
  key_map = {
    base_key: f'{k}.{base_key}'
    for base_key in cirq.measurement_keys(self)
  }
  return self.with_measurement_key_mapping(key_map)
```

PR #3499, opened 2020-11-12
Enforce copying in with_gate

95-martin-orion:
See also: PR #3495. This is a breaking change.

`with`-functions are generally expected to return a copy of the object, even if the copy is identical. Returning the original invites the possibility of mutation errors.

### DIFF
--- a/cirq/ops/gate_operation.py
+++ b/cirq/ops/gate_operation.py
@@ -59,8 +59,6 @@ class GateOperation(raw_types.Operation):
         return cast(TSelf, self.gate.on(*new_qubits))
 
     def with_gate(self, new_gate: 'cirq.Gate') -> 'cirq.Operation':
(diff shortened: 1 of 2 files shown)


95-martin-orion:
Note however that `GateOperation` is functionally immutable, though we don't state this explicitly in the code.

maffoo (commented):
Can we keep the current behavior and just document that GateOperation should be immutable? It could be very expensive when resolving parameters in circuits to have to re-allocate every operation, even when the gates haven't changed.

95-martin-orion:
> Can we keep the current behavior and just document that GateOperation should be immutable? It could be very expensive when resolving parameters in circuits to have to re-allocate every operation, even when the gates haven't changed.

Sounds good to me. I'll do this in #3495, which prompted this discussion.

PR #3503, opened 2020-11-13
Perform CliffordState measurements via method dispatch

smitsanghavi:
And deprecate the old `perform_measurement` method that called the `_measure` method directly.

Also deprecate all uses of `CliffordTableau` inside the `CliffordState`. Full clean up will follow.

#2948 #2423


cirq/sim/clifford/clifford_simulator.py
```diff
@@ -354,6 +372,7 @@ def apply_unitary(self, op: 'cirq.Operation'):
         rewrite=lambda args, kwargs: (args, {('collapse_state_vector' if k ==
                                               'collapse_wavefunction' else k): v
                                              for k, v in kwargs.items()}))
+    @deprecated(deadline='v0.10.0', fix='Use the apply_measurement instead')
```
smitsanghavi, line 375: Is the `CliffordState` directly used by end users? Do they call the `perform_measurement` function? If not, I can just delete it.

cirq/sim/clifford/clifford_simulator.py
```diff
@@ -346,6 +344,26 @@ def apply_unitary(self, op: 'cirq.Operation'):
...
+        qids = [self.qubit_map[i] for i in op.qubits]
+
+        args = clifford.ActOnCliffordTableauArgs(state.tableau, qids, prng,
+                                                 measurements)
+        act_on(op, args)
+
+        for i, qid in enumerate(qids):
+            state.ch_form.project_Z(qid, measurements[key][i])
+
     @deprecated_parameter(
```
smitsanghavi, line 367: Given that the whole function is deprecated now, should I remove the argument deprecation notice?

smitsanghavi:
Friendly Ping!

smitsanghavi:
Holding off on this PR for now

cirq/sim/clifford/clifford_simulator.py
```diff
@@ -312,11 +312,13 @@ def __str__(self) -> str:
     def to_numpy(self) -> np.ndarray:
         return self.ch_form.to_state_vector()
 
+    @deprecated(deadline='v0.10.0', fix='use CliffordTableau instead')
```
balopat, line 315: This should be deprecated in 0.11.0 at least - we are in 0.10.0.dev on master!

cirq/sim/clifford/clifford_simulator.py
```diff
@@ -312,11 +312,13 @@ def __str__(self) -> str:
...
     def stabilizers(self) -> List[DensePauliString]:
         """Returns the stabilizer generators of the state. These
         are n operators {S_1,S_2,...,S_n} such that S_i |psi> = |psi>"""
         return self.tableau.stabilizers()
 
+    @deprecated(deadline='v0.10.0', fix='use CliffordTableau instead')
```
balopat, line 321:
```suggestion
    @deprecated(deadline='v0.11.0', fix='use CliffordTableau instead')
```

cirq/sim/clifford/clifford_simulator.py
```diff
@@ -359,6 +374,7 @@ def apply_unitary(self, op: 'cirq.Operation'):
             },
         ),
     )
+    @deprecated(deadline='v0.10.0', fix='Use the apply_measurement instead')
```
balopat, line 377:
```suggestion
    @deprecated(deadline='v0.11.0', fix='Use the apply_measurement instead')
```

cirq/sim/clifford/clifford_simulator.py
balopat: Let's just be on the safe side, it's a public function. We don't have usage numbers!

smitsanghavi:
Ping!

cirq/sim/clifford/clifford_simulator.py
```diff
@@ -331,21 +333,34 @@ def wave_function(self):
...
         return
 
+    def apply_measurement(
+        self,
+        op: 'cirq.Operation',
+        measurements: Dict[str, List[np.ndarray]],
+        prng: np.random.RandomState,
+        collapse_state_vector=True,
+    ):
+        if collapse_state_vector:
```
balopat, line 354: as this is a public method, I think we should ensure that the op passed in has a MeasurementGate as op.gate. WDYT?

smitsanghavi: Makes sense. Added the validation and some tests.

balopat (approved):
LGTM, thank you!

PR #3505, opened 2020-11-16
Add PhasedFSimGate

viathor:


cirq/ops/fsim_gate.py
```diff
@@ -164,3 +165,268 @@ def __repr__(self) -> str:
...
+    """General excitation-preserving two-qubit gate.
+
+    The unitary matrix of this gate is:
+
+        [[1,           0,             0,           0],
+         [0,    f1 cos(θ), -i f2 sin(θ),           0],
+         [0, -i f3 sin(θ),    f4 cos(θ),           0],
+         [0,           0,             0, f5 exp(-iφ)]]
+
+    where the phase factors are
```
obriente, line 182: Minor comment: Maybe add some mention that the phase factors aren't independent and can't be individually set in the gate?

obriente (approved):
@mrwojtek raised a point separately that I fully agree with - we should make sure to be consistent in our nomenclature. Otherwise LGTM. Thanks! This is a useful generalization to have.

cirq/ops/fsim_gate.py
```diff
@@ -164,3 +165,268 @@ def __repr__(self) -> str:
...
+    the fact that the gate is equivalent to the following circuit:
+
+        0: ───Rz(b0)───FSim(theta, phi)───Rz(a0)───
+                       │
+        1: ───Rz(b1)───FSim(theta, phi)───Rz(a1)───
+
+    where b0 and b1 are phase angles to be applied before the core FSimGate,
+    a0 and a1 are phase angles to be applied after FSimGate and theta and
+    phi specify the core FSimGate. Use the static factory function
+    PhasedFSimGate.from_phase_angles_and_fsim to instantiate the gate
```
maffoo, line 211: Internally we also have dataclasses to store just the parameters in these different representations with functions to canonicalize them (since the various phase parameters are periodic) and to convert between them. It might be nice to include that code here, though it might be awkward to support symbolic values so would have to try it out. Note that internally we call the canonical parametrization above `FsimPhase` while this alternate parametrization with per-qubit z rotations is called `FsimRz`. If we stick with that nomenclature then the factory function would better be called `PhasedFSimGate.from_fsim_rz` or something like that, and the properties for getting the rotation angles before and after might also best be renamed as `rz_angles_before` and `rz_angles_after`.

cirq/ops/fsim_gate.py
```diff
@@ -164,3 +165,268 @@ def __repr__(self) -> str:
...
+        return out
+
+    def _decompose_(self, qubits) -> 'cirq.OP_TREE':
+        q0, q1 = qubits
+        q0_z_exponent_before = self.phase_angles_before[0] / np.pi
+        q1_z_exponent_before = self.phase_angles_before[1] / np.pi
+        q0_z_exponent_after = self.phase_angles_after[0] / np.pi
+        q1_z_exponent_after = self.phase_angles_after[1] / np.pi
+        yield cirq.Z(q0)**q0_z_exponent_before
+        yield cirq.Z(q1)**q1_z_exponent_before
```
maffoo, line 404: I'd suggest writing this in terms of rotations, and can also use names that correspond with the docstring:
```python
b0, b1 = self.phase_angles_before
a0, a1 = self.phase_angles_after
yield cirq.rz(q0, b0)
yield cirq.rz(q1, b1)
yield FSimGate(self.theta, self.phi).on(q0, q1)
yield cirq.rz(q0, a0)
yield cirq.rz(a1, a1)
```

viathor: This is nearly identical to my original implementation. Unfortunately, protocol consistency checks are sensitive to global phase, so we have to give up `__unitary__()[0, 0] == 1` or we have to yield `cirq.Z` instead of `cirq.rz` here. I chose the latter so that `__unitary__` agrees with eq (43) from [QS supplement](https://arxiv.org/abs/1910.11333).

(Per other comments, I'll be updating this PR to agree with eq (18) in [FH paper](https://arxiv.org/abs/2010.07965), but this merely renames existing "delta" parameters, so I think I'll keep the constraint that `__unitary__()[0, 0] == 1`.)

viathor: (We could also yield a global phase gate, but that seems even uglier.)

maffoo: I see. In that case, perhaps include a comment why this differs from the docstring, which shows `Rz` gates. In addition, would be good to refactor to avoid valuating each angle property twice, and you might also consider following the `rz` [implementation](https://github.com/quantumlib/Cirq/blob/master/cirq/ops/common_gates.py#L1332) which uses `sympy.pi` where appropriate to preserve symbolic values.

cirq/ops/fsim_gate.py
viathor: Done.

cirq/ops/fsim_gate.py
mrwojtek: +1, it would be nice to follow up on this because it will be one of the building blocks of the some calibration features.

mrwojtek (approved):

cirq/ops/fsim_gate.py
viathor: Renamed the factory method and the two properties as suggested. As discussed on a different PR, I'd like to defer dataclasses (and possibly re-use existing code that @mrwojtek is introducing elsewhere).

cirq/ops/fsim_gate.py
viathor: Good suggestions. All done.

cirq/ops/fsim_gate.py
```diff
@@ -164,3 +165,252 @@ def __repr__(self) -> str:
 
     def _json_dict_(self) -> Dict[str, Any]:
         return protocols.obj_to_dict_helper(self, ['theta', 'phi'])
+
+
+@value.value_equality(approximate=True)
+class PhasedFSimGate(gate_features.TwoQubitGate,
+                     gate_features.InterchangeableQubitsGate):
```
mrwojtek, line 172: I just started to wonder about this: what does it mean that this gate implements InterchangeableQubitsGate? Does it mean that operation is equal after qubits are swapped? That's not true, right?

viathor: Thanks for catching this! Sent #3524.

PR #3506, opened 2020-11-16
New landing page for tutorials

rmlarose:
Cosmetic changes to the tutorial landing page, keeping a similar structure (beginner / intermediate / advanced) to the old `index.md`. 

One could argue some advanced tutorials should be under the intermediate section (e.g., Rabi oscillation seems more intermediate to me) and vice versa, but these are easy changes once the structure is in place. 

Tutorials hosted on GitHub are marked with a GitHub icon.


95-martin-orion:
Everything looks like it was translated over properly. My one concern is regarding the link in the `basics` notebook: on the devsite it should be fine, but on the Cirq repo it links to the raw file since Github doesn't render YAML. Are we okay with this?

rmlarose:
Thanks Orion - it's a good point and I wasn't sure what's best to do here. Two ideas:

1. Include an absolute link to the devsite (if possible).
    - This would fix the viewing a raw yaml file problem, but would take people away from RTD, which could be confusing.

2. Keep a separate `index.md` for RTD and link to this from `basics.ipynb`.
    - Also fixes the viewing a raw yaml file problem, but then there would be two separate index files, both of which would have to be updated.

95-martin-orion:
> 1. Include an absolute link to the devsite (if possible).
>    
>    * This would fix the viewing a raw yaml file problem, but would take people away from RTD, which could be confusing.

IIRC the long-term plan is to migrate off of RTD anyways, so this sounds fine. I think the right course of action here is to keep the index.md link until the devsite goes public, and change to an absolute link at that time.

balopat (approved):

PR #3507, opened 2020-11-17
Fixes run_batch requiring param list

### LINKED ISSUE #3285: "Number of circuits and sweeps must match" -- run_batch
I have a list of non-parameterized, non-sweepy circuits and thought I could just not pass `params_list` but it insists that it be of equal length to the number of circuits. 

1. the default value shouldn't be `None` if that doesn't actually work (?)
2. can I pass a list of `None`s? can we make this happen automatically?

dabacon:
In #3285 it points out that there is a requirement that there always be param lists specified for a batch program on `Engine.run_batch`. This changes this to the sensible default, which is that if the params are None that empty param sweeps are sent to match the programs.  This also required changes to the similar calls in `EngineProgram`.  Here I've also added `batch_size` for batch programs, that uses the cached value of the program, or potentially fetches a new program.

While this technically changes the api of this method, it only makes it possible to now use the default or specify None, so don't think this needs to got through deprecation.

Fixed #3285


cirq/google/engine/engine.py
```diff
@@ -347,7 +349,9 @@ def run_batch(
             for a circuit are listed in the order imposed by the associated
             parameter sweep.
         """
-        if not params_list or len(programs) != len(params_list):
+        if params_list is None:
+            params_list = [dict()] * len(programs)
+        if len(programs) != len(params_list):
```
95-martin-orion, line 354: The line above ensures `params_list` and `programs` are of equal length, so this `if` can be an `elif`.

95-martin-orion (approved):
Agreed that deprecation is unnecessary here. One nit and this is ready to merge.

cirq/google/engine/engine.py
```diff
@@ -347,7 +349,9 @@ def run_batch(
             for a circuit are listed in the order imposed by the associated
             parameter sweep.
         """
-        if not params_list or len(programs) != len(params_list):
+        if params_list is None:
+            params_list = [dict()] * len(programs)
```
maffoo, line 353: Can do `[None] * len(programs)` since `cirq.Sweepable` includes `None` (via [`ParamResolverOrSimilarType`](https://github.com/quantumlib/Cirq/blob/master/cirq/study/resolver.py#L33)).

cirq/google/engine/engine_program.py
```diff
@@ -149,16 +151,20 @@ def run_batch(
...
+
+        Raises:
+            ValueError: if the program was not a batch program or no processors
+                were supplied.
         """
         import cirq.google.engine.engine as engine_base
         if self.result_type != ResultType.Batch:
             raise ValueError('Can only use run_batch() in batch mode.')
+        if params_list is None:
+            params_list = [dict()] * self.batch_size()
```
maffoo, line 163: Here, too, can do `[None] * self.batch_size()`.

dabacon: done

cirq/google/engine/engine.py
dabacon: done

cirq/google/engine/engine.py
dabacon: done

dabacon (commented):
PTAL addressed comments

maffoo (approved):
LGTM

Issue #3292, opened 2020-09-02
Better formatting for cirq code


maffoo:
I find the autoformatting that we currently use in cirq to be less than ideal. There are three main issues:
- yapf formatting is often bad, particular with things like nested tuples but also due to mixing hanging versus 4-space indentation styles (here's an example of [mixed indentation](https://github.com/quantumlib/Cirq/blob/master/cirq/google/op_deserializer_test.py#L93))
- 80 character lines seem to be too short, particularly with heavy use of type annotations as we have in our code.
- The incremental formatting is line-based, so we have lots of old code that has never been formatted.

**Describe the solution you'd like**

- Switch formatter from yapf to [black](https://github.com/psf/black). Black's formatting algorithm is simpler and the resulting formatting is more predictable and regular than yapf's, IME.
- Increase the line length to something like 100 characters. I think this would cut down on a lot of unnecessary line breaking and improve readability of the code.
- Switch the incremental autoformatting to be file-based instead of line-based. As part of doing this, we would have a one-time change to format the entire code base, which would be a massive diff but entirely automated. There's also a bit of work to update PRs that are in-flight at the time of the big formatting change, but this too is manageable.

**Describe alternatives/workarounds you've considered**

- Could increase line length without switching from yapf to black.
- Could tweak various yapf knobs to possibly improve the formatting (I prefer black's approach of having fewer knobs in the first place so there's less to argue about, but YMMV).
- Could switch to file-based incremental formatting and pay the one-time cost of formatting all old code without switching to black.

**What is the urgency from your perspective for this issue? Is it blocking important work?**

P3 - I'm not really blocked by it, it is an idea I'd like to discuss / suggestion based on principle

mpharrigan:
I absolutely hate the yapf formatter and it often makes my code less readable. I also think a line length of 100 strikes a better balance between screen real-estate and superfluous wrapping. 

I personally like the pycharm formatter. It leaves things alone that don't break any of the rules. While that means there isn't a "canonical" way to format things, it lets the code author use line breaks and indentation to indicate intent, e.g. when building a nested list of stuff. Of course, it's much more difficult to install and run the pycharm formatter than a command-line utility. 

File-based incremental sounds good. A massive formatting change scares me, but may be our best bet. Maybe we can use that as motivation to get the number of open PRs down :)

balopat:
[...]

PR #3508, opened 2020-11-18
Add JSON serialization for Calibration metrics

dstrain115:
- Add json_dict and from_json_dict for serializiation
of Calibration objects
- Add a valid repr for Calibration objects.


cirq/google/engine/calibration.py
```diff
@@ -50,9 +50,15 @@ class Calibration(abc.Mapping):
             the epoch.
     """
 
-    def __init__(self, calibration: v2.metrics_pb2.MetricsSnapshot) -> None:
+    def __init__(self,
+                 calibration: v2.metrics_pb2.MetricsSnapshot = v2.metrics_pb2.
+                 MetricsSnapshot(),
```
mpharrigan, line 55: this is a weird line wrap. cc: that bug about how we need a better formatter

cirq/google/engine/calibration.py
```diff
@@ -103,8 +109,39 @@ def __len__(self) -> int:
...
+                metric_dict[metric['name']][key] = metric['values'][idx]
+
+        return cls(metric_dict=metric_dict)
+
+    def _json_dict_(self):
+        rtn = []
+        count = 0
+        for key in self._metric_dict:
+            metric_dict = {}
+            rtn.append(metric_dict)
```
mpharrigan, line 134: why not put this at the end of the loop body so it reads more idiomatically? i.e.: make empty dict, fill it in, add it to list.

cirq/google/engine/calibration.py
```diff
@@ -103,8 +109,39 @@ def __len__(self) -> int:
...
+            metric_dict['name'] = key
+            cur_dict = self._metric_dict[key]
```
mpharrigan, line 136: Is there an opportunity for better variable names to make this more readable? It's strange that `metric_dict` is an entry in `self._metric_dict`

cirq/google/engine/calibration.py
```diff
@@ -103,8 +109,39 @@ def __len__(self) -> int:
...
+            metric_dict['targets'] = [
+                [v2.qubit_to_proto_id(q) for q in k] for k in cur_dict
```
mpharrigan, line 138: Might be more readable with

```python
[q_to_id(q) for q in qubits] for qubits in cur_dict.keys()
```

cirq/google/engine/calibration.py
```diff
@@ -103,8 +109,39 @@ def __len__(self) -> int:
...
+            ]
+            metric_dict['values'] = [
+                [val for val in cur_dict[k]] for k in cur_dict
```
mpharrigan, line 141: I know dictionaries are ordered now in Python, but it still makes me nervous. What about?

```python
targets = []
values = []
for target, value in cur_dict.items():
  targets.append([q_to_id(q) for q in target])
  values.append(value)
metric_dict['targets'] = targets
metric_dict['values'] = values
```

cirq/google/engine/calibration.py
```diff
@@ -103,8 +109,39 @@ def __len__(self) -> int:
...
+            ]
+            count += 1
```
mpharrigan, line 143: what is count for?

cirq/google/engine/calibration.py
```diff
@@ -103,8 +109,39 @@ def __len__(self) -> int:
...
+        return {'cirq_type': 'Calibration', 'metrics': rtn}
```
mpharrigan, line 144: should we be namespacing these `cirq_type`s that don't live in the top level? Discussion beyond the scope of this PR

mpharrigan (commented):
a little convoluted for listifying a dictionary

dstrain115:
> a little convoluted for listifying a dictionary

Thanks for the review.  Do you have any better suggestions?  I struggled with trying to figure out the best way to serialize this.  I was worried about making the metric names as json keys, since they are user supplied and also used as python arguments when objects are constructed, and it felt insecure and error_prone if we have metrics or values that have odd names.

cirq/google/engine/calibration.py
dstrain115: Fixed formatting with parentheses.

cirq/google/engine/calibration.py
dstrain115: Done.

cirq/google/engine/calibration.py
dstrain115: Changed name to current_metric and removed cur_dict.

cirq/google/engine/calibration.py
dstrain115: Obsolete given next comment.

cirq/google/engine/calibration.py
dstrain115: Good idea.  Done.

cirq/google/engine/calibration.py
dstrain115: Yeah, I don't need that anymore.  Deleted.

cirq/google/engine/calibration.py
dstrain115: Uh, maybe?  Will that still work or will that break the current resolution mechanism?  I don't think I want to go diving into json serialization internals here, but I don't want to create a backwards compatibility mess later down the road.  WDYT?

maffoo:
The `Calibration` class is a thin wrapper around the `MetricsSnapshot` protobuf and protos already have a well-defined JSON serialization. Could we use that? Basically Calibration would serialize as a dict with two fields, one containing type info for cirq, the other containing the canonical json form of the metrics snapshot message.

mpharrigan:
> Do you have any better suggestions?

I think it reads much better now

mpharrigan (dismissed):

wcourtney (commented):
Are we meant to follow the google style guide? Only part way through, but I saw an approval and don't want comments to be missed.

cirq/google/engine/calibration.py
```diff
@@ -103,8 +109,37 @@ def __len__(self) -> int:
...
     def __str__(self) -> str:
+        return f'Calibration(keys={list(sorted(self.keys()))})'
 
-        return 'Calibration(keys={})'.format(list(sorted(self.keys())))
+    def __repr__(self) -> str:
+        return ('cirq.google.Calibration(metric_dict='
+                f'{repr(dict(self._metric_dict))})')
+
+    @classmethod
+    def _from_json_dict_(cls, metrics, **kwargs):
```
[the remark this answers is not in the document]
mpharrigan: This is what you name the first argument in Python for `@classmethod`s. Instead of `self`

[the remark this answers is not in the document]
mpharrigan: This is the method used by the JSON protocol when deserializing

cirq/google/engine/calibration.py
balopat: cc #3292

cirq/google/engine/calibration.py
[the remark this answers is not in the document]
dstrain115: cls is for class.  This is a pretty standard for python classmethods (similar to how you always use 'self') so I am not going to change it.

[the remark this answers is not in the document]
wcourtney: Thanks. Fine by me if this is standard; sorry for the Python noob question.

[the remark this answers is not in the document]
dstrain115: This is part of the JSON protocol, so we need it.  I will add a short comment.

cirq/google/engine/calibration.py
```diff
@@ -50,9 +50,15 @@ class Calibration(abc.Mapping):
             the epoch.
     """
 
-    def __init__(self, calibration: v2.metrics_pb2.MetricsSnapshot) -> None:
+    def __init__(self,
+                 calibration: v2.metrics_pb2.MetricsSnapshot = (
+                     v2.metrics_pb2.MetricsSnapshot()),
+                 metric_dict: Optional[
+                     Dict[str, Dict[Tuple['cirq.GridQubit', ...], Any]]] = None
```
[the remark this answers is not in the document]
dstrain115: Yeah, I will get back to this.  I need to modify this class to allow strings in the next PR, so I will fix up the typing in that PR.

cirq/google/engine/calibration.py
```diff
@@ -50,9 +50,15 @@ class Calibration(abc.Mapping):
             the epoch.
     """
 
-    def __init__(self, calibration: v2.metrics_pb2.MetricsSnapshot) -> None:
+    def __init__(self,
+                 calibration: v2.metrics_pb2.MetricsSnapshot = (
+                     v2.metrics_pb2.MetricsSnapshot()),
+                 metric_dict: Optional[
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/engine/calibration.py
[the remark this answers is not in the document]
dstrain115: Will fix typing in the next PR, since I want to modify this anyway to allow strings in addition to qubits.

cirq/google/engine/calibration.py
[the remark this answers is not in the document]
dstrain115: Added a comment for clarification that this is a magic method for JSON serialization.

cirq/google/engine/calibration.py
```diff
@@ -103,8 +109,37 @@ def __len__(self) -> int:
...
+    def __repr__(self) -> str:
+        return ('cirq.google.Calibration(metric_dict='
+                f'{repr(dict(self._metric_dict))})')
+
+    @classmethod
+    def _from_json_dict_(cls, metrics, **kwargs):
+        metric_dict = {}
+        for metric in metrics:
+            metric_dict[metric['name']] = {}
+            for idx, val in enumerate(metric['targets']):
```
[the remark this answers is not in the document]
dstrain115: Removed.

cirq/google/engine/calibration.py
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/engine/calibration.py
```diff
@@ -103,8 +109,37 @@ def __len__(self) -> int:
...
+                key = tuple(v2.grid_qubit_from_proto_id(q) for q in val)
+                metric_dict[metric['name']][key] = metric['values'][idx]
+
+        return cls(metric_dict=metric_dict)
+
+    def _json_dict_(self):
+        rtn = []
```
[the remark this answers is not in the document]
dstrain115: Removed.

dstrain115:
Changed this to output JSON based on the proto JSON serialization.

dstrain115:
Re-requested review since I changed the way the objects are serialized.  PTAL and let me know if you approve.  Thanks!

cirq/google/engine/calibration.py
```diff
@@ -103,8 +113,44 @@ def __len__(self) -> int:
...
+                current_metric.name = key
+                current_metric.targets.extend(
+                    [v2.qubit_to_proto_id(q) for q in target])
+                for value in value_list:
+                    current_value = current_metric.values.add()
+                    if isinstance(value, float):
+                        current_value.double_val = value
+                    if isinstance(value, int):
+                        current_value.int64_val = value
+                    if isinstance(value, str):
```
wcourtney, line 137: Should this cause an error if there's a bad type rather than failing silently?

cirq/protocols/json_test_data/Calibration.repr
```diff
@@ -1,13 +1,13 @@
...
             },
         't1': {
-            (cirq.GridQubit(0, 0),): [321.0],
-            (cirq.GridQubit(0, 1),): [911.0],
-            (cirq.GridQubit(1, 0),): [505.0]},
-        'globalMetric': {(): [12300]}
+            (cirq.GridQubit(0, 0),): [321],
+            (cirq.GridQubit(0, 1),): [911],
+            (cirq.GridQubit(1, 0),): [505]},
+        'globalMetric': {(): ['abcd']}
```
wcourtney, line 11: Consider adding a test for an unsupported type to assert the expected behavior (whatever it is).

wcourtney (approved):

cirq/google/engine/calibration.py
dstrain115: Done.

cirq/protocols/json_test_data/Calibration.repr
dstrain115: Done.

cirq/google/engine/calibration.py
```diff
@@ -45,14 +46,23 @@ class Calibration(abc.Mapping):
...
+    def __init__(self,
+                 calibration: v2.metrics_pb2.MetricsSnapshot = v2.metrics_pb2.
+                 MetricsSnapshot(),
+                 metrics: Optional[
+                     Dict[str, Dict[Tuple['cirq.GridQubit', ...], Any]]] = None
+                ) -> None:
         self.timestamp = calibration.timestamp_ms
-        self._metric_dict = self._compute_metric_dict(calibration.metrics)
+        self._metric_dict = (metrics or
+                             self._compute_metric_dict(calibration.metrics))
```
viathor, line 65: IIUC, the intention is to compute the metrics if `None` is passed, however this also computes the metrics when empty dict is passed. We should only have one value that triggers such behavior, so the following is better:

```python
self._metric_dict = self._compute_metric_dict(calibration.metrics) if metrics is None else metrics
```

cirq/protocols/json_test_data/Calibration.json
```diff
@@ -0,0 +1,4 @@
+{
+  "cirq_type": "Calibration",
+  "metrics": "{\n  \"metrics\": [\n    {\n      \"name\": \"xeb\",\n      \"targets\": [\n        \"0_0\",\n        \"0_1\"\n      ],\n      \"values\": [\n        {\n          \"doubleVal\": 0.9999\n        }\n      ]\n    },\n    {\n      \"name\": \"xeb\",\n      \"targets\": [\n        \"0_0\",\n        \"1_0\"\n      ],\n      \"values\": [\n        {\n          \"doubleVal\": 0.9998\n        }\n      ]\n    },\n    {\n      \"name\": \"t1\",\n      \"targets\": [\n        \"0_0\"\n      ],\n      \"values\": [\n        {\n          \"int64Val\": \"321\"\n        }\n      ]\n    },\n    {\n      \"name\": \"t1\",\n      \"targets\": [\n        \"0_1\"\n      ],\n      \"values\": [\n        {\n          \"int64Val\": \"911\"\n        }\n      ]\n    },\n    {\n      \"name\": \"t1\",\n      \"targets\": [\n        \"1_0\"\n      ],\n      \"values\": [\n        {\n          \"int64Val\": \"505\"\n        }\n      ]\n    },\n    {\n      \"name\": \"globalMetric\",\n      \"values\": [\n        {\n          \"strVal\": \"abcd\"\n        }\n      ]\n    }\n  ]\n}"
```
viathor, line 3: What is the format of the string value under the "metrics" key? It looks like JSON... If so, it's a little ugly to embed JSON as string in another JSON. Could we use a proper JSON dictionary as a value for the "metrics" key?

cirq/google/engine/calibration.py
```diff
@@ -103,8 +113,48 @@ def __len__(self) -> int:
...
+        return proto
+
+    @classmethod
+    def _from_json_dict_(cls, metrics: str, **kwargs):
+        """Magic method for the JSON serialization protocol."""
+        metric_proto = v2.metrics_pb2.MetricsSnapshot()
+        json_format.Parse(metrics, metric_proto)
+        return cls(metric_proto)
+
+    def _json_dict_(self):
```
viathor, line 152: type annotation:

```python
```

viathor (approved):
LGTM with two minor comments (conditional expression and type annotation). Serialization logic can be kept as is.

cirq/protocols/json_test_data/Calibration.json
viathor: As discussed offline: due to restrictions of existing APIs (protobufs and cirq's JSON serialization protocol) making serialization format more human readable requires jumping through some hoops that complicates the logic. Code readability is more important than serialization format readability, so feel free to keep as is.

cirq/google/engine/calibration.py
```diff
@@ -103,8 +113,48 @@ def __len__(self) -> int:
...
+        """Magic method for the JSON serialization protocol."""
+        return {
+            'cirq_type': 'Calibration',
+            'metrics': json_format.MessageToJson(self.to_proto())
```
maffoo, line 156: Can use `json_format.MessageToDict` to convert to a dict that can then be converted to string by the standard json serialization. This avoids the json-string-in-json problem. Similarly, use `json_format.ParseDict` instead of `json_format.Parse` when deserializing.

dstrain115: Ooohhh, that's just what I was looking for.  Let me try that out!

cirq/google/engine/calibration.py
viathor: Still needs type annotation :-)

viathor (approved):
Very cool! Thanks!

cirq/google/engine/calibration.py
```diff
@@ -103,8 +115,48 @@ def __len__(self) -> int:
...
+                        raise ValueError(f'Unsupported metric value {value}. '
+                                         'Must be int, float, or str to '
+                                         'convert to proto.')
+        return proto
+
+    @classmethod
+    def _from_json_dict_(cls, metrics: str, **kwargs) -> 'Calibration':
+        """Magic method for the JSON serialization protocol."""
+        metric_proto = v2.metrics_pb2.MetricsSnapshot()
+        json_format.ParseDict(metrics, metric_proto)
```
maffoo, line 151: Note that `ParseDict` returns the message, so can do this as a one-liner if desired:
```python
return cls(json_format.ParseDict(metrics, v2.metrics_pb2.MetricsSnapshot()))
```

dstrain115: Done.

cirq/google/engine/calibration.py
dstrain115: Done.

PR #3509, opened 2020-11-18
CircuitGates

95-martin-orion:
This PR adds CircuitGates, a Gate wrapper for FrozenCircuits. It also provides a stub for the CircuitOperation class, which is a specialized GateOperation for wrapping CircuitGates.

Part of #3235.

### DIFF
--- a/cirq/circuits/circuit_gate.py
+++ b/cirq/circuits/circuit_gate.py
@@ -18,7 +18,7 @@ of a larger circuit, a CircuitGate will execute all component gates in order,
 including any nested CircuitGates.
 """
 
-from typing import TYPE_CHECKING, Any, Dict, Optional, Sequence, Union
+from typing import TYPE_CHECKING, Any, Dict, Optional, Sequence, Tuple, Union
 
 from cirq import circuits, devices, ops, protocols
 from cirq.circuits.insert_strategy import InsertStrategy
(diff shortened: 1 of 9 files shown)


95-martin-orion:
I've opened a follow-up PR which fleshes out the CircuitOperation class (#3515, not ready for review!). It might be useful for understanding design choices made in this PR.

cirq/circuits/circuit_gate.py
```diff
@@ -0,0 +1,190 @@
...
+
+class CircuitGate(ops.Gate):
+    """A single gate that encapsulates a Circuit."""
+
+    def __init__(self,
+                 *contents: 'cirq.OP_TREE',
+                 strategy: 'cirq.InsertStrategy' = InsertStrategy.EARLIEST,
+                 device: 'cirq.Device' = devices.UNCONSTRAINED_DEVICE,
+                 name: Optional[str] = None,
+                 exp_modulus: Optional[int] = None) -> None:
```
maffoo, line 39: I'm not sure this is a good idea. When we ask hardware to repeat a circuit we typically want it to run that many times, we don't want things that are equivalent to identities to get optimized away (think about randomized benchmarking). Such transformations could be implemented in an optimization pass, but it seems odd to bake this in to the gate itself.

maffoo:
Could you maybe include the `CircuitOperation` changes in this PR as well? I think it makes sense to review them together since they are tightly related and I find myself looking back and forth between the two PRs to understand things anyway.

95-martin-orion:
> Could you maybe include the `CircuitOperation` changes in this PR as well? I think it makes sense to review them together since they are tightly related and I find myself looking back and forth between the two PRs to understand things anyway.

#3515 contains all of the changes added here - if you prefer, we can move the review entirely over to that PR.

cirq/circuits/circuit_gate.py
```diff
@@ -0,0 +1,190 @@
...
+        # Until Circuit supports _commutes_, this will return NotImplemented.
+        return protocols.commutes(self.circuit, other, atol=atol)
+
+    def _has_unitary_(self) -> bool:
+        return protocols.has_unitary(self.circuit)
+
+    def _decompose_(self, qubits=None) -> 'cirq.OP_TREE':
+        applied_circuit = self.circuit.unfreeze()
+        if qubits is not None and qubits != self.ordered_qubits():
+            qmap = {old: new for old, new in zip(self.ordered_qubits(), qubits)}
```
maffoo, line 149: The implicit reliance on a particular ordering here seems like it could be error prone.

maffoo (commented):
Do we actually want `CircuitGate` as a separate class from `CircuitOperation`? Given that a circuit consists of operation (already applied to qubits), it's quite natural to treat a circuit as an operation, but it's not clear what it means to treat it as a gate. And since the `CircuitOperation` class will provide a place to store the qubit and measurement maps if you need to modify a circuit when embedding, it seems like `CircuitGate` doesn't give us much.

maffoo:
@95-martin-orion, yes, let's move to the `CircuitOperation` PR.

PR #3513, opened 2020-11-19
Generalize types for basic qubit experiments

maffoo:
These should accept any cirq.Qid, not just cirq.GridQubit. Similarly, for tomography experiments we can take AbstractCircuit instead of Circuit. I've also updated type definitions to use the preferred convention of `'cirq.Foo'` for types that are exported on the main `cirq` namespace.


balopat (approved):
LGTM, nice and clean. Thank you!

viathor (approved):
Thank you!

PR #3514, opened 2020-11-19
Make WaitGate variadic and add wait helper function

maffoo:
In some cases we want a wait gate that applies to multiple qubits, to
creates a "barrier" that other gates cannot be moved past even while
a circuit is being manipulated in other ways. This makes the wait gate
variadic and adds a helper function to create WaitGate instances and
deal with qid_shape, similar to MeasurementGate/measure.

### DIFF
--- a/cirq/__init__.py
+++ b/cirq/__init__.py
@@ -274,6 +274,7 @@ from cirq.ops import (
     TwoQubitDiagonalGate,
     TwoQubitGate,
     VirtualTag,
+    wait,
     WaitGate,
     X,
     XPowGate,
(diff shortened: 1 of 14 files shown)


95-martin-orion (approved):
Looks great! `cirq.wait` is the kind of function that I'd install `cirq-unstable` to use.

dstrain115:
Sorry I didn't get to this quick enough to review.  My questions have to do with backwards compatibility.  Will this be fully backwards compatible, including proto serialization?

Can we add a (or modify an existing) test to make sure multi-qubit wait gates can serialize a multi-qubit wait gate?

PR #3515, opened 2020-11-19
CircuitOperations

95-martin-orion:
Sequel to #3509. Part of #3235.

This PR adds an initial version of the "CircuitOperation" type, which wraps a CircuitGate for use in Circuits or other CircuitGates. CircuitOperations can specify measurement keys, parameter values, and repetitions for their component CircuitGate.

Notable features that are absent in this version:
- Measurement in loops. The current measurement format (a 1:1 dict of keys to results) does not play well with repeating the same measurement multiple times, or with attempting to rename that measurement based on loop index.
- Condensed serialization. This feature requires changes outside the CircuitGate itself to allow for references inside the serialization structure.
- Condensed circuit diagrams. Similar to serialization, this requires changes to the diagram behavior for printing CircuitOperations separate from the rest of the circuit.

### DIFF
--- a/cirq/circuits/circuit_gate.py
+++ b/cirq/circuits/circuit_gate.py
@@ -139,7 +139,7 @@ class CircuitGate(ops.Gate):
         # TODO: support out-of-line subcircuit definition in string format.
         if self.name is None:
             # In the absence of a name, use a unique ID.
-            header = f'CircuitGate_{hash(self) % int(1e7):06d}:'
+            header = f'CircuitGate_{hash(self) % int(1e6):06d}:'
         else:
             header = f'{self.name}:'
         msg_lines = str(self.circuit).split('\n')
@@ -288,8 +288,8 @@ class CircuitOperation(ops.GateOperation):
         new_op._gate = new_gate
         return new_op
 
-    def with_measurement_key_mapping(
-            self, key_map: Dict[str, str]) -> 'CircuitOperation':
+    def with_measurement_key_mapping(self, key_map: Dict[str, str]
+                                    ) -> 'CircuitOperation':
         """Returns a copy of this operation with an updated key mapping.
 
         Parameter values and repetitions are preserved. The provided key_map
@@ -309,12 +309,12 @@ class CircuitOperation(ops.GateOperation):
         })
         return new_op
 
-    def _with_measurement_key_mapping_(
-            self, key_map: Dict[str, str]) -> 'CircuitOperation':
+    def _with_measurement_key_mapping_(self, key_map: Dict[str, str]
+                                      ) -> 'CircuitOperation':
         return self.with_measurement_key_mapping(key_map)
 
-    def with_params(
-            self, param_values: Dict[sympy.Symbol, Any]) -> 'CircuitOperation':
+    def with_params(self, param_values: Dict[sympy.Symbol, Any]
+                   ) -> 'CircuitOperation':
         """Returns a copy of this operation with an updated key mapping.
 
         Key mappings and repetitions are preserved. The provided param_values
--- a/cirq/circuits/circuit_gate.py
+++ b/cirq/circuits/circuit_gate.py
@@ -18,9 +18,13 @@ of a larger circuit, a CircuitGate will execute all component gates in order,
 including any nested CircuitGates.
 """
 
-from typing import TYPE_CHECKING, Any, Dict, Optional, Sequence, Tuple, Union
+from typing import (TYPE_CHECKING, Any, Dict, List, Optional, Sequence, Tuple,
+                    Union)
 
-from cirq import circuits, devices, ops, protocols
+import numpy as np
+import sympy
+
+from cirq import circuits, devices, linalg, ops, protocols
 from cirq.circuits.insert_strategy import InsertStrategy
 from cirq.type_workarounds import NotImplementedType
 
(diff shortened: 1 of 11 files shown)


95-martin-orion:
Things that are missing but should be added in this PR:
- Tag propagation: CircuitOperations should pass their tags to all sub-operations during decomposition. This is pretty simple, it just got missed in the first round of commits.
- `cirq.CircuitOperation`: requires serialization tests, which I feel compelled to make fairly thorough for CircuitOperation given its unique relationship with serialization.

cirq/circuits/circuit_gate.py
```diff
@@ -0,0 +1,384 @@
...
+        return self.with_measurement_key_mapping(key_map)
+
+    def with_params(self, param_values: Dict[sympy.Symbol, Any]
+                   ) -> 'CircuitOperation':
+        """Returns a copy of this operation with an updated key mapping.
+
+        Key mappings and repetitions are preserved. The provided param_values
+        are composed with the existing values, so calling this function with
+        param_values = {'b': 1} on a CircuitOperation with map {'a': 'b'} will
+        result in a final map of {'a': 1}.
```
maffoo, line 323: This is somewhat confusing to me. I wonder if it would be more clear to maintain two separate maps, a "symbol map" that is a bi-directional mapping between "inner" and "outer" symbols (like the measurement key map) and a separate dict of "resolved" outer parameters.

Maybe it would help to actually implement the parameter resolution protocol methods `_resolve_parameters_` and `_parameter_names_` because it's not clear how `with_params` fits in with parameter resolution.

cirq/circuits/circuit_gate.py
```diff
@@ -0,0 +1,384 @@
...
+        """
+        if not isinstance(repetitions, int):
+            raise TypeError('Only integer repetitions are allowed.')
+        if protocols.is_measurement(self.circuit):
+            raise NotImplementedError(
+                'Loops over measurements are not supported.')
+        new_op = self.copy()
+        new_op._repetitions *= repetitions
+        if self.gate.exp_modulus is not None:
+            # Map repetitions to [-exp_modulus / 2, exp_modulus / 2)
```
maffoo, line 351: If we do include something like `exp_modulus`, I'd suggest that there should be a way to specify whether or not it should be inverted. Repetitions between `[0, exp_modulus)` might make more sense in some cases. Also, for the half interval it might make more sense to be inclusive on the upper instead of lower bound, so that circuits with modulus 2 never need to be inverted, for example.

cirq/circuits/circuit_gate.py
```diff
@@ -0,0 +1,384 @@
...
+            new_op._repetitions %= self.gate.exp_modulus
+            if new_op._repetitions >= self.gate.exp_modulus / 2:
```
maffoo, line 353: Use integer division?

cirq/circuits/circuit_gate.py
95-martin-orion: This map assigns values to "inner" parameters, not "outer" ones. The example makes more sense in a nested context, e.g.:
```
Circuit(
  CircuitGate(
    X(k) ** b,
    CircuitGate(
      X(k) ** a,
    ).on(k).with_params('a': 'b')
  ).on(k).with_params('b': 1)
)
```

Its purpose is to defer applying parameter values until the `CircuitOperation` is decomposed, similar to the measurement key mapping. This helps with serialization and will open up some dynamic behavior possibilities once classical control is supported.

maffoo: The parameter map conflates two things currently:
- "rebinding" symbols so that you can embed the same circuit multiple times and vary the parameters on each one independently (e.g. embed `CircuitGate(X(k) ** a)` twice but rebind the exponent to separate symbols `a0` and `a1` in the "outer" circuit).
- assigning values to parameters (actual resolution is deferred until later so that we can reuse embedded circuit instances, as you said).

I'm wondering if we should explicitly separate those two in the code, instead of 'collapsing' them into one map as in the current implementation. It would definitely help to implement the parameter resolution special methods to make all this more clear.

One thing that is a bit odd in the current implementation is that it looks like one could "unresolve" parameters. For example:
```python
g = CircuitGate(X(k) ** a).on(k).with_params({'a': 1})
cirq.is_parameterized(g)  # => False (need to implement this)
cirq.parameter_names(g)  # => set()  (

g2 = g.with_params({'a': 'b'})
cirq.is_parameterized(g2)  # => True
```
That behavior feels a bit odd to me.

cirq/circuits/circuit_gate.py
```diff
@@ -0,0 +1,384 @@
...
+        """
+        new_op = self.copy()
+        new_op._measurement_key_map = {
+            k: (key_map[v] if v in key_map else v)
+            for k, v in self.measurement_key_map.items()
+        }
+        new_op._measurement_key_map.update({
+            key: val
+            for key, val in key_map.items()
+            if key not in self.measurement_key_map.values()
```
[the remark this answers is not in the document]
95-martin-orion: There should be a cleaner way of doing this using a for-loop rather than comprehensions.

cirq/circuits/circuit_gate.py
95-martin-orion: My current plan is to package this behavior (exp_modulus, modulus validation, and invertibility) into the `repeat` method instead. That should ensure users (or optimizers) are explicit about using those behaviors.

95-martin-orion:
Per discussion on #3509 and offline, I'll be opening a new PR without the CircuitGate construct.

cirq/circuits/circuit_gate.py
95-martin-orion: I'm currently adding support for `ParamResolver` composition in #3546 and #3538. Following those changes, `CircuitOperation` will use a `ParamResolver` with single-step resolution for its param value mapping.

PR #3516, opened 2020-11-20
Switch autoformatter to black instead of yapf

maffoo:
See discussion in #3292. This PR has three commits:
- an initial commit to fix a couple of places where black currently has trouble. These are all related to https://github.com/psf/black/issues/1629 and and places where we have long multiline expressions like `assert <multiline_expression> == <foo>` or `if <multiline_expression> == <bar>:`. In all cases, I fixed them by adding a trailing comma in the outermost parens of the multiline expression and reformatting with yapf. Then black was able to handle them. I don't expect this to come up much with using black, but if it does the fix is straightforward.
- a commit to change the `format-incremental` script to use black and reformat just the changed files from the first commit. I also tried to fix up the tests for the bash script, though it's not clear to me that these are being run anywhere because they seem to fail for other scripts that I haven't touched in this PR. (As an aside, it seems to me very brittle to test against the exact output of these scripts, and I would also suggest that the complexity in some cases is reaching a point where these should be refactored into python scripts, but I'm biased because I'm terrible at bash :-) )
- a final commit to reformat all the python code in the repo with black. For now, this is mostly for informational purposes to see what the black formatting looks like. We can decide if we want to merge with this "big bang" change, or change the formatting without this and then reformat code in chunks in subsequent PRs.


mpharrigan:
"big bang" will cause merge conflicts for all changes in-flight. Relying on incremental updates could get rid of this issues unless two people are working on changes for e.g. the same file

cirq/protocols/json_serialization.py
```diff
@@ -44,175 +46,167 @@
...
-                'sympy.Pow': lambda args: sympy.Pow(*args),
-                'sympy.Float': lambda approx: sympy.Float(approx),
-                'sympy.Integer': sympy.Integer,
-                'sympy.Rational': sympy.Rational,
-                'complex': complex,
-            }
-        return self._crd
-
-
-RESOLVER_CACHE = _ResolverCache()
```
mpharrigan, line 204: this is technically a backwards incompatible change. It may affect packages that layer on additional json-resolution paths on top of the base Cirq objects. However, I think those use cases *should* be solved by using `DEFAULT_RESOLVERS` below

mpharrigan: e.g. https://github.com/quantumlib/ReCirq/blob/master/recirq/serialization_utils.py#L225

balopat: +1

maffoo:
@mpharrigan, my plan would be to merge this without squashing, so that we have a commit in the commit history which just changes the formatting script but doesn't reformat anything. We would tag this commit as something like `fmt-black`. Then for outstanding PR's one would do:
```
git merge fmt-black
check/format-incremental --apply
# commit incremental formatting changes
git merge master  # merge with all formatting changes
```
Now the diff with master will only show the logical changes from the branch, not the formatting-related changes. We used this procedure internally when we switched formatters and it worked quite well. We'll have to help contributors go through the process for outstanding PRs, but it shouldn't be too much trouble.

mpharrigan:
Ok that makes sense. We'll have to proactively help contributors, since the above sequence requires you to be pretty clever to come up with it on your own :)

check/format-incremental
```diff
@@ -35,77 +36,66 @@ cd "$(git rev-parse --show-toplevel)"
...
-    exit 1
+if (( any_files == 0 )); then
+    echo -e "\033[32mNo files to format\033[0m."
 else
-    echo -e "\033[33mReformatted changed lines\033[0m."
+    args=()
+    if (( only_print == 1 )); then
+        args=("--check" "--diff")
+    fi
+    black --color "${args[@]}" ${format_files}
```
viathor, line 99: Shouldn't we run twice with --fast to work around https://github.com/psf/black/issues/1629?

maffoo:
Ok, all the pre-commits are now merged so this is down to just two commits: one to change the formatting script and one to actually do the formatting. Reviewers, please take a look at the tooling changes in the first commit, and of course feel free to look at the formatting changes in the second commit as well if you'd like.

check/format-incremental
balopat: It does seem like it...

balopat (approved):
I'm okay to merge this. Let's do it! I am a bit worried about https://github.com/psf/black/issues/1629 and whether we should run the formatter twice or not, can you clarify that before merging? Also you do have some merge conflicts now...:)

check/format-incremental
```diff
@@ -35,77 +36,87 @@ cd "$(git rev-parse --show-toplevel)"
...
+# black --color "${args[@]}" ${format_files}
+# exit $?
+LOGS=$(black --color "${args[@]}" ${format_files} 2>&1)
+STATUS=$?
+echo "$LOGS"
+if [[ "$STATUS" == "123" && "$LOGS" =~ ^.*"INTERNAL ERROR: Black produced different code on the second pass of the formatter.".*$ ]]; then
+  echo -e "\033[31mWarning, it seems we ran into https://github.com/psf/black/issues/1629. Falling back to calling black --fast twice. Please double check that your code is still valid!\033[0m"
+  black --fast --color "${args[@]}" ${format_files}
+  STATUS=$?
+  if (( only_print != 1 )); then
```
balopat, line 114: why the `only_print` conditional for the second run? I'm thinking about the two commands as a "single transaction" replacing `black --color`.  The logic for `only_print` is already encoded in `args`...

check/format-incremental
```diff
@@ -35,77 +36,87 @@ cd "$(git rev-parse --show-toplevel)"
...
+# Once that is fixed upstream, we can just do:
+# black --color "${args[@]}" ${format_files}
+# exit $?
+LOGS=$(black --color "${args[@]}" ${format_files} 2>&1)
+STATUS=$?
+echo "$LOGS"
+if [[ "$STATUS" == "123" && "$LOGS" =~ ^.*"INTERNAL ERROR: Black produced different code on the second pass of the formatter.".*$ ]]; then
+  echo -e "\033[31mWarning, it seems we ran into https://github.com/psf/black/issues/1629. Falling back to calling black --fast twice. Please double check that your code is still valid!\033[0m"
+  black --fast --color "${args[@]}" ${format_files}
+  STATUS=$?
```
balopat, line 113: I think we can be conservative here and exit on failure. Reason: we know that for the 1629 bug the first formatting succeeds. Thus we can be specific about it here and ask the user to file a bug instead if the first command fails.

check/format-incremental
balopat: Oh I get it - if we only_print == 1 then it doesn't make sense to output twice.

balopat (approved):
LGTM

maffoo:
If you have an in-progress PR that was created before the formatting change, you might see lots of merge conflicts because of the formatting changes. To fix this, you can merge the tag "fmt-black" into your branch to get the new formatting script. Then format your changed files before merging master. Here's the process, starting with your in-progress branch checked out:

```
$ git fetch origin
$ git merge fmt-black
$ pip install -r dev_tools/conf/pip-list-dev-tools.txt
$ check format-incremental --apply
$ git commit -am "Format"
$ git merge origin/master
```

In the last step, you may need to fix some merge conflicts, but they will be due to the actual logic changes you've made, rather than due to formatting. If you'd like to automatically accept all changes on your branch, you can try running `git merge origin/master -X ours` instead; this will accept all changes on your branch automatically, but may need some manually fixup if things have changed on master in the meantime.

dabacon:
@balopat  What is guidance for multiline string reformatting for new 100 char limit as the format didn't touch those?  Fix as we find them? 

Also it sometimes does fun things like concatenates the strings with a space, i.e.
`ValueError('foo'
'bar')`
becomes
`ValueError('foo' 'bar')

95-martin-orion:
Couple notes:
> `$ git fetch origin`

`$ git fetch upstream` if you are working off of a fork

> `$ check format-incremental --apply`

Missing slash: `$ check/format-incremental --apply`

> `$ git merge origin/master`

Again, `$ git fetch upstream` if working from a fork

Issue #3274, opened 2020-08-27
Support cirq.commutes(operation, moment)


Strilanc:
It would be handy to be able to easily check if an operation commutes with everything in a moment. For example, it makes things like https://quantumcomputing.stackexchange.com/questions/13488/reordering-commuting-gates-in-cirq-to-reduce-circuit-depth easier to implement.

Currently, checking if a moment and an operation commute raises an exception saying the case hasn't been implemented.

Here's a simple implementation, which has the downside of having quadratic overhead when comparing two moments.

```
    def _commutes_(self, other: Any, *, atol: float) -> bool:
        for op in self.operations:
            op_result = protocols.commutes(op,
                                           other,
                                           atol=atol,
                                           default=NotImplemented)
            if not op_result or op_result is NotImplemented:
                return op_result
        return True
```

balopat:
Thanks for filing this! It should be easy to avoid the quadratic overhead in the moment vs moment case by checking commutation only between operations that share at least one qubit.

akushnarov:
Esteemed, 

I've started working on this issue. Cirq is pretty new for me and I would need some guidance here, could you please advise?

Especially my question was:
Where it is better to implement "commutes" functionality:
- In the "[commutes_protocol.py](https://github.com/quantumlib/Cirq/blob/master/cirq/protocols/commutes_protocol.py)" (by extending the list of strategies, see the [diff](https://github.com/quantumlib/Cirq/compare/master...akushnarov:master))
- OR in the Moment class ("[moment.py](https://github.com/quantumlib/Cirq/blob/master/cirq/ops/moment.py)"), by implementing a method "_commutes_"?

Also I would need some support in writing tests, since I'm not that good at modelling what commutes and what doesn't. Anyone can guide me?

Thanks and I appreciate any help
Andriy

akushnarov:
1. As discussed on the https://gitter.im/cirqdev/community, I'm going to implement the solution on the `Moment` class side.
2. As for the tests I will take into account the feedback (from the same channel):

> Regarding testing, I think it should be relatively easy to cook up some testcases, just experiment a bit with cirq.commutes(op1, op2) where op1, op2 are different operations on qubits (e.g. cirq.X(cirq.NamedQubit(“a”))). Now, for a Moment, you’ll have a list of these guys, so it should be straightforward to create testcases. A Moment commutes with a given operation if all of its operations commute with it. If an operation does not share qubits with another operation, they trivially commute.

[...]

PR #3519, opened 2020-11-21
Implement cirq.commute for Moment and Operation

akushnarov:
Adds cirq.commute support to check whether an Operation commutes with everything in a Moment. 

Fixes https://github.com/quantumlib/Cirq/issues/3274.


cirq/ops/moment_test.py
```diff
@@ -472,3 +472,24 @@ def __str__(self):
 0 │ Empty
   │
     """)
+
+
+def test_commutes():
+    a = cirq.NamedQubit('a')
```
balopat, line 478: nit / tip: you can just do `a,b,c,d = cirq.LineQubit.range(4)` here.

balopat (approved):
This looks good to me, I added a tiny nit and edited your title + description as they will become the final commit messages.
Please fix that one nit and then we can merge! 

Thank you!

Issue #3479, opened 2020-11-04
RFC proposal: IonQ integration


dabacon:
This is a proposal to integrate vendor IonQ's api into Cirq:

[RFC: Cirq IonQ integration](https://docs.google.com/document/d/1ml1mZ-LyCxh1_FtRPBOqDfoxZAxvMDl95ePc-Dvq-aw/edit#heading=h.mckr847qq60k)

See RFC document for details.

dabacon:
Ooops wrong doc shared, should be correct now.

dabacon:
This work is done.

PR #3523, opened 2020-11-23
Add cancel and delete to IonQ job

dabacon:
Also:
* factors out ``IonQExceptions`
* checks that jobs are not canceled or deleted when jobs are queried for fields that no longer exists when job enteres those states.


cirq/ionq/ionq_client.py
```diff
@@ -226,10 +208,10 @@ def get_job(self, job_id: str):
         """Get the job from the IonQ API.
 
         Args:
-            job_id: The UUID of the job (returned when the job is created).
+            job_id: The UUID of the job (returned when the job was created).
 
         Returns:
-            A `cirq.ionq.IonQJob` corresponding to the job.
+            The json body of the response a dict.
```
95-martin-orion, line 214: "a dict" -> "as a dict"

cirq/ionq/ionq_client_test.py
```diff
@@ -317,3 +317,38 @@ def test_ionq_client_get_job_retry(mock_get):
                                           default_target='simulator')
     _ = client.get_job('job_id')
     assert mock_get.call_count == 2
+
+
+@mock.patch('requests.put')
+def test_ionq_client_cancel_job(mock_put):
```
95-martin-orion, line 323: Should this and `delete_job` also have failure tests, or is the coverage provided by the `get` and `create` tests sufficient?

95-martin-orion (changes requested):
Looks good overall, just a couple points to check on.

cirq/ionq/ionq_client.py
dabacon: fixed

cirq/ionq/ionq_client_test.py
dabacon: Probably but better form to test these.  I've added tests.  What's the new acronym for the opposite of dry in tests?

dabacon:
PTAL

dabacon:
Thanks for the quick review :)

cirq/ionq/ionq_client_test.py
95-martin-orion: Internet tells me it's WET (for "Write Everything Twice") 😝 

I have a feeling these could be parameterized to DRY things up, but I'm not going to block the merge on that. LGTM

95-martin-orion (approved):

dabacon:
Part of #3479

PR #3524, opened 2020-11-23
Fix symmetry of PhasedFSimGate under qubit exchange

viathor:
Fixes an issue found by @mrwojtek in #3505.

### DIFF
--- a/cirq/ops/fsim_gate.py
+++ b/cirq/ops/fsim_gate.py
@@ -34,6 +34,16 @@ from cirq._compat import proper_repr
 from cirq.ops import gate_features
 
 
+def _zero_mod_pi(param: Union[float, sympy.Basic]) -> bool:
+    """Returns True iff param, assumed to be in [-pi, pi], is 0 (mod pi)."""
+    return param in (-np.pi, 0.0, np.pi, -sympy.pi, sympy.pi)
+
+
+def _half_pi_mod_pi(param: Union[float, sympy.Basic]) -> bool:
+    """Returns True iff param, assumed to be in [-pi, pi], is pi/2 (mod pi)."""
+    return param in (-np.pi / 2, np.pi / 2, -sympy.pi / 2, sympy.pi / 2)
+
+
 @value.value_equality(approximate=True)
 class FSimGate(gate_features.TwoQubitGate,
                gate_features.InterchangeableQubitsGate):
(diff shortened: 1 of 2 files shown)


cirq/ops/fsim_gate.py
```diff
@@ -301,6 +301,15 @@ def rz_angles_after(
         a1 = (-self.gamma - self.zeta + self.chi) / 2.0
         return a0, a1
 
+    def qubit_index_to_equivalence_group_key(self, index: int) -> int:
+        """Returns a key that differs between non-interchangeable qubits."""
+        # PhasedFSimGate is symmetric with respect to qubit swap
+        # if and only if zeta and chi are integer multiples of pi.
+        r = (-np.pi, 0.0, np.pi)
+        if self.zeta in r and self.chi in r:
```
mrwojtek, line 309: It would be very nice to use isclose for this check but I guess there is a problem how to pass tolerance here. (I'm fine with the exact comparison as well.)

viathor: Yeah, exact comparisons of floating point numbers make me unhappy, too. Unfortunately, this is a recurring issue in cirq and in past discussions we settled on using such comparisons in some cases (e.g. when algebraic properties are at stake as here) on the premise that we're interested in the case when a user provides a special value such as pi exactly (as opposed to the value resulting from a floating point computation).

This is in some sense more appropriate because for any other value the gate is not strictly speaking symmetric with respect to qubit exchange (even though it may be approximately symmetric).

That said, I'm open to doing it either way if your use-case suggests we should do something else.

mrwojtek: Thanks for explanation, this is perfectly fine withe me - I personally don't intend to use the interchangeability property in anything that I can think of right now.

mrwojtek:
Could you mention in the class docstring under which circumstances this gate is interchangeable? Even knowing this, I tend to sometimes forget that I need to be careful with this order and end up debugging obscure bugs. For people who see that untiary for the first time, that is relevant information.

mrwojtek (approved):

mrwojtek:
I'll double-check the math tomorrow but this looks good from the code point of view.

mrwojtek:
I checked your math and everything seems fine. This is becoming pretty sophisticated logic now!

cirq/ops/fsim_gate.py
```diff
@@ -301,7 +306,26 @@ def rz_angles_after(
         a1 = (-self.gamma - self.zeta + self.chi) / 2.0
         return a0, a1
 
+    def _zeta_insensitive(self) -> bool:
+        return self.theta in (-np.pi / 2, np.pi / 2, -sympy.pi / 2,
+                              sympy.pi / 2)
+
+    def _chi_insensitive(self) -> bool:
+        return self.theta in (-np.pi, 0.0, np.pi, -sympy.pi, sympy.pi)
```
maffoo, line 314: Maybe we can add module-level helper functions:
```python
def _zero_mod_pi(param: Union[float, sympy.Basic]) -> bool:
    """Returns true if param == 0 (mod pi)."""
    return param in (-np.pi, 0.0, np.pi, -sympy.pi, sympy.pi)

def _half_pi_mod_pi(param: Union[float, sympy.Basic]) -> bool:
    """Returns true if param == pi / 2 (mod pi)."""
    return param in (-np.pi / 2, np.pi / 2, -sympy.pi / 2, sympy.pi / 2)

...
def _zeta_insensitive(self) -> bool:
    return _half_pi_mod_pi(self.theta)

def _chi_insensitive(self) -> bool:
    return _zero_mod_pi(self.theta)

def qubit_index_to_equivalence_group_key(self, index: int) -> int:
    if ((_zero_mod_pi(self.zeta) or self._zeta_insensitive()) and
        (_zero_mod_pi(self.chi) or self._chi_insensitive())):
         ...
```

maffoo (approved):
LGTM

cirq/ops/fsim_gate.py
viathor: Done.

PR #3525, opened 2020-11-23
codeowners file

balopat:
Adds CODEOWNERS file. 

This will enable the following behavior: 

- @cirq-maintainers, @cduck and @vtomole  are the default codeowners for the content of the repo. This has been decided based on levels of contributions in the last 6 months on Cirq Cynque, PR approvals and issue conversations. 
- @wcourtney is added as an owner to the google related code and docs
- Owners are going to be pinged on every PR
- I will turn on "owner approval required" - which means that no one can merge a PR unless an owner has approved the given PR. 

Eventually this will enable the definition of codeowners for vendors, e.g. AQT, IonQ, Pasqal and experimental modules - so that these modules can move faster without the overhead of waiting on Cirq maintainer approval

@cduck, @vtomole, let me know if you want to opt out of this role. 
See more at: https://docs.github.com/en/free-pro-team@latest/github/creating-cloning-and-archiving-repositories/about-code-owners

### DIFF
--- a/dev_tools/conf/pip-list-dev-tools.txt
+++ b/dev_tools/conf/pip-list-dev-tools.txt
@@ -39,8 +39,5 @@ ipykernel
 # For verifying rst
 rstcheck~=3.3.1
 
-# for verifying github's code owners file
-codeowners
-
 # For testing time specific logic
 freezegun~=0.3.15
--- a/dev_tools/conf/pip-list-dev-tools.txt
+++ b/dev_tools/conf/pip-list-dev-tools.txt
@@ -39,7 +39,7 @@ ipykernel
 # For verifying rst
 rstcheck~=3.3.1
 
-# for verifying codeowners
+# for verifying github's code owners file
 codeowners
 
 # For testing time specific logic
--- a/dev_tools/conf/pip-list-dev-tools.txt
+++ b/dev_tools/conf/pip-list-dev-tools.txt
@@ -39,5 +39,8 @@ ipykernel
 # For verifying rst
 rstcheck~=3.3.1
 
+# for verifying codeowners
+codeowners
+
 # For testing time specific logic
 freezegun~=0.3.15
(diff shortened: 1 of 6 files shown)


dev_tools/conf/pip-list-dev-tools.txt
```diff
@@ -39,5 +39,8 @@ ipykernel
 # For verifying rst
 rstcheck~=3.3.1
 
+# for verifying github's code owners file
+codeowners
+
```
95-martin-orion, line 44: This is enough to make the devtools themselves happy, but testing the devtools probably requires this to be added to `requirements.txt` as well.

95-martin-orion (approved):
Approval pending test fixes.

dev_tools/conf/pip-list-dev-tools.txt
balopat: No, requirements.txt should never have testing related deps. 
The issue is something weird with the codeowners lib on Mac (weirdly because it installs locally just fine for me) and Windows (which I'll test in a sec on my Win box).

balopat: It turns out Win wheels are not published at all (https://github.com/sbdchd/codeowners/issues/5) and it seems not all Mac wheels are published either. We should write our own parser eventually, but for now this works. I added a comment and turned off the new tests on win / mac.

balopat: oh I'm silly, I need to somehow pip install only on linux! I'll do that soon too.

dabacon:
That's a lot of spam for a single commit.  Because so many people get added on a single PR, it may be a problem "who is actually reviewing this"

balopat:
> That's a lot of spam for a single commit. Because so many people get added on a single PR, it may be a problem "who is actually reviewing this"

I'm open for alternatives, but there is no other way to enable code ownership with built-in tools unfortunately - that was still an issue before, a lot of times people forget to press the "request review" button and then the PR just sits there.

An alternative with external tools would be something like Prow: https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md

PR #3527, opened 2020-11-23
Switching to pre-releases instead of cirq-unstable

### LINKED ISSUE #3175: cirq-unstable updates are broken

### LINKED ISSUE #3267: Prevent installing both cirq and cirq-unstable.

If `cirq-unstable` is installed in a virtualenv and you `pip install cirq`, the virtualenv gets into a bad state. This is because they are different packages with separate metadata, but they put the python code in the same location in a `cirq` folder in site-packages. So pip thinks both are installed but in fact only the most-recently installed package is "active".

If you try to fix such a situation by uninstalling one package, the code gets removed but pip still thinks the other package is installed. The other package is now in a broken state (can't be imported, etc.). The only way to recover is to uninstall _both_ packages and then install one of them again.

I'd like `pip install cirq` to fail if `cirq-unstable` is installed, and vice-versa. I did a quick glance at setuptools docs to see if there is some way to specify packages that _must not_ be present, but I couldn't see anything. It also doesn't seem to work to add a package requirement like `cirq != *`. I'm not sure if there's any other mechanism we can use to accomplish this.

We have internal workarounds for this to try to keep people's virtualenvs in a good state if they accidentally install both versions, but it'd be great if we could prevent the problem from happening at all.

balopat:
Switch to PyPi pre-releases instead of `cirq-unstable`.

Two major reasons for this change: 

1. in preparation for modules - for which we would have to duplicate all of them if we followed this pattern. 
2. As raised in #3267, `cirq-unstable` doesn't play well with `cirq` installations

Fixes #3267.

A resulting state after a test-run for this looks like this on the [Test PyPi](https://test.pypi.org/project/cirq/#history):

image

The dev releases will still get the "this is an unstable version" notice for their description.


README.rst
```diff
@@ -23,7 +23,7 @@ Installation and Documentation
 
 Cirq documentation is available at `cirq.readthedocs.io <https://cirq.readthedocs.io>`_.
 
-Documentation for the latest **unstable** version of cirq (tracks the repository's master branch; what you get if you ``pip install cirq-unstable``), is available at `cirq.readthedocs.io/latest <https://cirq.readthedocs.io/en/latest/>`_.
+Documentation for the latest **unstable** version of cirq (tracks the repository's master branch; what you get if you ``pip install --pre cirq``), is available at `cirq.readthedocs.io/latest <https://cirq.readthedocs.io/en/latest/>`_.
```
viathor, line 26: s/unstable/pre-release/

dev_tools/packaging/verify-published-package.sh
```diff
@@ -44,17 +44,17 @@ if [ -z "${PROJECT_VERSION}" ]; then
...
-    PYPI_REPOSITORY_FLAG="--index-url=https://test.pypi.org/simple/"
+    PIP_FLAGS="--index-url=https://test.pypi.org/simple/"
     PYPI_REPO_NAME="TEST"
     PYPI_PROJECT_NAME="cirq"
 elif [ "${PROD_SWITCH}" = "--prod" ]; then
-    PYPI_REPOSITORY_FLAG=''
+    PIP_FLAGS=''
     PYPI_REPO_NAME="PROD"
     PYPI_PROJECT_NAME="cirq"
 elif [ "${PROD_SWITCH}" = "--unstable" ]; then
```
viathor, line 54: Optional: Consider renaming the flag to --pre. Alternatively, you can add a second flag to serve as an alias.

(If you choose to do this, update the top comment, too.

setup.py
```diff
@@ -28,10 +28,11 @@
 # README file as long_description.
 long_description = io.open('README.rst', encoding='utf-8').read()
 
-# If CIRQ_UNSTABLE_VERSION is set then we use cirq-unstable as the name of the package
-# and update the version to this value.
+# If CIRQ_UNSTABLE_VERSION is set then we update the version to this value.
+# As it contains "dev" it will be a pre-release version on PyPi. See
```
viathor, line 32: IIUC, this should be changed: "As it contains" -> "If it contains"

("As" implies we know it does, which I don't think we do.)

viathor (approved):
LGTM with minor comments.

balopat:
Thanks @viathor - I'm going to have to put a pause on this as I discovered Craig's comment here, that pointed out that tensorflow and tfq-nightly have the same issue as cirq and cirq-unstable: https://github.com/quantumlib/Cirq/issues/2431#issuecomment-547650606.
Also I need a bit more evidence that in a multi-module setting the pre-releases will play nicely together - i.e. whether I can setup for example `pip install --pre cirq-pasqal` should bring in the correct latest `cirq`... I'll convert this back to live if I'm convinced that this is the right way to go.

mpharrigan:
Isn't Craig's comment further justification for using pre-releases?

balopat:
> Isn't Craig's comment further justification for using pre-releases?

I interpreted it as "if this is okay with tensorflow, it should be okay for us too to live with this" based on the context that he seemed to be "pro separate package".

balopat:
Oh by the way some more updates on this: 

1. https://www.python.org/dev/peps/pep-0440/#developmental-releases: 

>Note
>
>While they may be useful for continuous integration purposes, publishing developmental releases of pre-releases to general purpose public index servers is strongly discouraged, as it makes the version identifier difficult to parse for human readers. If such a release needs to be published, it is substantially clearer to instead create a new pre-release by incrementing the numeric component.
>
>Developmental releases of post-releases are also strongly discouraged, but they may be appropriate for projects which use the post-release notation for full maintenance releases which may include code changes. 
>

2. There is no impact on dependency management whether we use the same package or separate ("unstable") packages. It is straightforward to generate requirements to the given version in the setup.py for each module. 

I'm more concerned about the 1st point and playing nice in the python community. More motivating conversations around this: 

- How PyPi does not allow for retention policy: https://discuss.python.org/t/publishing-nightly-builds-on-test-pypi-org-with-a-time-based-retention-policy/3152
- scikiy-learn [ended up on anaconda.org](https://scikit-learn.org/stable/developers/advanced_installation.html) as they have the ability to delete packages: https://docs.anaconda.com/anaconda-cloud/user-guide/tasks/work-with-packages/

The only issue I have with anaconda.org is that this is how we'd install modules: 

```
pip install --pre --extra-index https://pypi.anaconda.org/cirq-unstable/simple cirq
```

```
pip install --pre --extra-index https://pypi.anaconda.org/cirq-unstable/simple cirq-google
```

balopat:
> Oh by the way some more updates on this:
> 
> 1. https://www.python.org/dev/peps/pep-0440/#developmental-releases:
> 
> > Note
> > While they may be useful for continuous integration purposes, publishing developmental releases of pre-releases to general purpose public index servers is strongly discouraged, as it makes the version identifier difficult to parse for human readers. If such a release needs to be published, it is substantially clearer to instead create a new pre-release by incrementing the numeric component.
> > Developmental releases of post-releases are also strongly discouraged, but they may be appropriate for projects which use the post-release notation for full maintenance releases which may include code changes.
> 
> 1. There is no impact on dependency management whether we use the same package or separate ("unstable") packages. It is straightforward to generate requirements to the given version in the setup.py for each module.
> 
> I'm more concerned about the 1st point and playing nice in the python community. More motivating conversations around this:
> 
> * How PyPi does not allow for retention policy: https://discuss.python.org/t/publishing-nightly-builds-on-test-pypi-org-with-a-time-based-retention-policy/3152
> * scikiy-learn [ended up on anaconda.org](https://scikit-learn.org/stable/developers/advanced_installation.html) as they have the ability to delete packages: https://docs.anaconda.com/anaconda-cloud/user-guide/tasks/work-with-packages/
> 
> The only issue I have with anaconda.org is that this is how we'd install modules:
> 
> ```
> pip install --pre --extra-index https://pypi.anaconda.org/cirq-unstable/simple cirq
> ```
> 
> ```
> pip install --pre --extra-index https://pypi.anaconda.org/cirq-unstable/simple cirq-google
> ```

Oh hold on - the first point is irrelevant, it is just a strong recommendation around non-human-readable version numbers. Ignore that. 

I was for a second also worried about "littering" the main package with dev releases for readability - but maybe it's not that big of a deal. Also, pre-releases like `v1.0.0.rc1`, `v1.0.0.rc2` can still work, just `pip install --pre cirq` will be shadowed by the dev releases, so one will have to use the specific versions there instead explicitly.
2. In terms of space, I am actually not that worried about it. With 1.6MB per release and ~800 commits per year we are below a 1GB per year. 


So the only question remains is readability of different versions on the cirq pypi page...

setup.py
balopat: We can clarify that - but it _is_ assumed that it contains dev.

balopat: Actually it can contain whatever, but it is assumed that it contains `devN`, `aN`, `bN`, `rcN` hence the long description that changes it to the warning text below

balopat:
In the end I am ready to merge this as it will make our (less packages to deal with) and our users lives easier (no weird namespace collisions).

As I understand the per project limit is 10GB which we'll get to in 5+ years if we never clean up a release. 
Cleaning up releases is manual at this point - but there is an issue for PyPi that pushes for automated cleanups (https://github.com/pypa/warehouse/issues/8792). 

@quantumlib/cirq-maintainers this is a final chance to object :) Please comment, otherwise I'm ready to merge tomorrow!

mpharrigan:
All in all, I think this is a good idea. We did this for a previous open source project I maintained and it worked well iirc

maffoo (approved):
Minor comment about release docs, otherwise this LGTM!

release.md
```diff
@@ -6,7 +6,7 @@ themselves are created. Note that development is done on the `master`
 branch, so if you want to use a more stable version you should use one
 of the [releases](https://github.com/quantumlib/Cirq/releases) or
 install from pypi using `pip install cirq`.  The release from the
-latest commit to master can be installed with `pip install cirq-unstable`.
+latest commit to master can be installed with `pip install cirq==dev`.
```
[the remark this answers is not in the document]
balopat: Good spotting! This was a complete accident - one of those where I was search & replacing `-unstable` and tried to type something in another window - weirdly looks like a valid syntax somehow lol.

balopat:
As we have quorum, I'm going to go ahead and merge it. Thanks all for chiming in!

PR #3531, opened 2020-11-25
[doc] remove CircuitDag methods from API doc that are inherited from networkx

balopat:
Removes CircuitDag methods from API doc that are inherited from networkx.

### DIFF
--- a/dev_tools/docs/build_api_docs.py
+++ b/dev_tools/docs/build_api_docs.py
@@ -16,7 +16,6 @@
 
 import os
 import types
-from types import FunctionType
 
 import networkx
 from absl import app
@@ -24,7 +23,6 @@ from absl import flags
 from tensorflow_docs.api_generator import doc_controls
 from tensorflow_docs.api_generator import generate_lib
 from tensorflow_docs.api_generator import public_api
-from tensorflow_docs.api_generator.doc_controls import do_not_doc_inheritable
 
 import cirq
 from cirq import _doc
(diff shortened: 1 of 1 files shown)


dev_tools/docs/build_api_docs.py
```diff
@@ -15,12 +15,16 @@
...
+import types
+from types import FunctionType
 
+import networkx
 from absl import app
 from absl import flags
 from tensorflow_docs.api_generator import doc_controls
 from tensorflow_docs.api_generator import generate_lib
 from tensorflow_docs.api_generator import public_api
+from tensorflow_docs.api_generator.doc_controls import do_not_doc_inheritable
```
95-martin-orion, line 27: With this, we have both `import doc_controls` and `from {...}.doc_controls import {...}`. Is the extra import necessary?

dev_tools/docs/build_api_docs.py
```diff
@@ -15,12 +15,16 @@
 """Tool to generate external api_docs for Cirq (Shameless copy from TFQ)."""
 
 import os
+import types
+from types import FunctionType
```
95-martin-orion, line 19: Only one of these is needed.

95-martin-orion (changes requested):

dev_tools/docs/build_api_docs.py
balopat: oh, sure I can clean that up

95-martin-orion (approved):

PR #3532, opened 2020-11-25
Refactor how we generate default json resolver

maffoo:
This came out of #3516; removing two levels of indentation makes the formatting change less drastic, but IMO also makes the code more understandable by removing the singleton `_ResolverCache` class.

### DIFF
--- a/cirq/protocols/json_serialization.py
+++ b/cirq/protocols/json_serialization.py
@@ -12,11 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import dataclasses
+import functools
 import json
 import numbers
 import pathlib
 from typing import (
     Any,
+    Callable,
     cast,
     Dict,
     IO,
@@ -36,7 +38,7 @@ import sympy
 from typing_extensions import Protocol
 
 from cirq._doc import doc_private
-from cirq.ops import raw_types  # Tells mypy that the raw_types module exists
+from cirq.ops import raw_types
 from cirq.type_workarounds import NotImplementedType
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


cirq/protocols/json_serialization.py
```diff
@@ -440,8 +428,9 @@ def _cirq_object_hook(d, resolvers: Sequence[JsonResolver]):
         raise ValueError("Could not resolve type '{}' "
                          "during deserialization".format(d['cirq_type']))
 
-    if hasattr(cls, '_from_json_dict_'):
-        return cls._from_json_dict_(**d)
+    from_json_dict = getattr(cls, '_from_json_dict_', None)
+    if from_json_dict is not None:
```
mpharrigan, line 432: why the change?

cirq/protocols/json_serialization.py
```diff
@@ -36,183 +38,169 @@
...
-                'sympy.Pow': lambda args: sympy.Pow(*args),
-                'sympy.Float': lambda approx: sympy.Float(approx),
-                'sympy.Integer': sympy.Integer,
-                'sympy.Rational': sympy.Rational,
-                'complex': complex,
-            }
-        return self._crd
-
-
-RESOLVER_CACHE = _ResolverCache()
```
mpharrigan, line 204: technically a backwards-incompatible change but I think it should be fine because extensions (at least when I write them) use the `DEFAULT_RESOLVERS` global instead e.g. https://github.com/quantumlib/ReCirq/blob/master/recirq/serialization_utils.py#L225

cirq/protocols/json_serialization.py
maffoo: The old code doesn't type check because mypy doesn't understand that `cls._from_json_dict_` is valid after the `hasattr` check.

cirq/protocols/json_serialization.py
maffoo: We don't export `RESOLVER_CACHE` in the package-level `__init__.py` and it was only used in the associated tests so I wasn't considering it part of the public api. But I guess it's a bit ambiguous from the naming alone (no leading underscore).

balopat: I agree that this should be fine to remove, I added the BREAKING_CHANGE label so that we can list it in the release notes.

balopat (approved):
LGTM

PR #3533, opened 2020-11-25
Fix type error in (X/Y)PowGate._act_on_

maffoo:
There was a type error due to the fact that the common superclass `EigenGate`  does not define `_act_on_`. We can act on each gate instead of looping over them, which avoids the type error.

### DIFF
--- a/cirq/ops/common_gates.py
+++ b/cirq/ops/common_gates.py
@@ -338,20 +338,20 @@ class YPowGate(eigen_gate.EigenGate,
                 return NotImplemented
             effective_exponent = self._exponent % 2
             state = args.state
+            Z = ZPowGate()
             if effective_exponent == 0.5:
-                assert all(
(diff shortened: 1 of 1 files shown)


smitsanghavi (changes requested):
If we are changing this to avoid `type: ignore`s, might as well fix the same pattern in `YPowGate` for consistency

balopat (approved):
This looks good - though I do like the list comprehension as it's a bit more python-esque.
An alternative way to keep it is to actually use the protocol type in the type definition: 

```
                ops = [ZPowGate(), H]  # type: List[SupportsActOn]
                assert all(gate._act_on_(args) for gate in ops)
```
What do you think of this direction?

maffoo:
@balopat, I added a helper function to call `_act_on_` on a series of gates. The helper has a type annotation as you suggested, which takes care of the types.

PR #3534, opened 2020-11-25
Tweak formatting to prepare for switching to black

maffoo:
Switching the formatting to black in #3516 brought up a few issues. Black has an edge case that requires inserting trailing commas in some constructs (https://github.com/psf/black/issues/1629), which I've done here. Also removes trailing whitespace in a few places that pylint complains about when the files are reformatted (those lint errors are latent since we only lint changed files on PRs).

### DIFF
--- a/cirq/contrib/qasm_import/_parser_test.py
+++ b/cirq/contrib/qasm_import/_parser_test.py
@@ -67,12 +67,12 @@ def test_comments():
     parser = QasmParser()
 
     parsed_qasm = parser.parse("""
-    //this is the format 
+    //this is the format
     OPENQASM 2.0;
     // this is some other comment
     include "qelib1.inc";
     // and something at the end of the file
-    // multiline 
+    // multiline
     """)
 
     assert parsed_qasm.supportedFormat
--- a/cirq/interop/quirk/url_to_circuit_test.py
+++ b/cirq/interop/quirk/url_to_circuit_test.py
@@ -79,15 +79,15 @@ def test_parse_with_qubits():
 
     assert quirk_url_to_circuit(
         'http://algassert.com/quirk#circuit={"cols":[["H"],["•","X"]]}',
-        qubits=cirq.GridQubit.rect(4, 4)) == cirq.Circuit(
-            cirq.H(a),
-            cirq.X(b).controlled_by(a))
+        qubits=cirq.GridQubit.rect(4, 4),
+    ) == cirq.Circuit(cirq.H(a),
+                      cirq.X(b).controlled_by(a))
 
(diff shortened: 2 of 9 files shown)


cirq/ops/pauli_string_test.py
```diff
@@ -1676,13 +1677,18 @@ def test_circuit_diagram_info():
             1j * cirq.X(a) * cirq.Y(b),
             -1j * cirq.Y(b),
             1j**0.5 * cirq.X(a) * cirq.Y(b),
-        ), """
+        ),
+        """
```
maffoo, line 1681: The trailing comma causes yapf to move the leading quotes to their own line, which for some reason seems to confuse pylint. I could only get it to ignore this long string literal by putting the `pylint: disable` directive at the top level. No idea why.

cirq/contrib/qasm_import/_parser_test.py
```diff
@@ -67,12 +67,12 @@ def test_comments():
     parser = QasmParser()
 
     parsed_qasm = parser.parse("""
-    //this is the format 
+    //this is the format
```
viathor, line 70: nit: add whitespace between // and text for consistency

cirq/interop/quirk/url_to_circuit_test.py
```diff
@@ -79,15 +79,15 @@ def test_parse_with_qubits():
 
     assert quirk_url_to_circuit(
         'http://algassert.com/quirk#circuit={"cols":[["H"],["•","X"]]}',
-        qubits=cirq.GridQubit.rect(4, 4)) == cirq.Circuit(
-            cirq.H(a),
-            cirq.X(b).controlled_by(a))
+        qubits=cirq.GridQubit.rect(4, 4),
+    ) == cirq.Circuit(cirq.H(a),
```
viathor, line 83: It's fairly common to add commas in long literal lists/tuples/sets/etc, but I suppose people are likely to forget to do this in function argument lists. From a quick glance at https://github.com/psf/black/issues/1629, it seems that the fallout is fairly scary: black produces inequivalent code! However, they do have a workaround (running twice with --fast), so perhaps we should use that in #3516? Added a comment there.

cirq/ops/pauli_string_test.py
balopat: Hmm, the battle of the "productivity" tools...well, soon enough yapf is going to go.

balopat (approved):
LGTM

viathor (approved):

HGSilveri (commented):
LGTM

PR #3535, opened 2020-11-25
Tweak coverage to prepare for switching to black

maffoo:
These changes came out of #3516 when trying to switch to using black for formatting.
- Add qubit_order test to cover exception case
- Remove `_unitary_` method on a test class which is not covered by tests
- Ignore coverage check for performance tests which are not run in a standard pytest run
- Modify the coverage script to ignore `if TYPE_CHECKING` blocks which don't run during tests (this resulted in unrelated formatting changes on `incremental_coverage.py` which apparently had never been formatted by yapf)


balopat (approved):
LGTM

PR #3536, opened 2020-11-25
Allow strings as targets of metrics in Calibration objects.

dstrain115:
- Currently all metrics are keyed by qubits or pairs of qubits.
However, the proto allows for arbitrary strings:
https://github.com/quantumlib/Cirq/blob/master/cirq/google/api/v2/metrics.proto#L31

This PR allows the Calibration to handle arbitrary strings. This
flexibility may become more important since we use this for calibration
API as well.

Most of the changes are merely due to mypy typing issues.

It also types the value of the metric as Union[str,int,float] rather
than Any.


dstrain115:
Could I get a review on this?  It seems to be blocking some of Wojtek's testing.

cirq/google/engine/calibration.py
```diff
@@ -174,6 +181,30 @@ def timestamp_str(self,
...
+
+        Attempts to parse it as a GridQubit.  If this fails,
+        returns the string itself.
+        """
+        try:
+            return v2.grid_qubit_from_proto_id(target)
+        except:
+            return target
+
+    def key_to_qubit(self, target: METRIC_KEY) -> devices.GridQubit:
```
95-martin-orion, line 195: Similar behavior is used in `noise_models.py` for the values of metrics - would it be worthwhile to have a `value_to_float` function?

cirq/google/engine/calibration_test.py
```diff
@@ -119,6 +119,14 @@ def test_to_proto():
         invalid_value.to_proto()
 
 
+def test_calibrations_with_string_key():
+    calibration = cg.Calibration(metrics={'metric1': {('alpha',): [0.1]}})
+    assert calibration == cg.Calibration(calibration.to_proto())
```
95-martin-orion, line 124: Is this the proto generated by this?
```
metrics: [{
  name: 'metric1'
  targets: ['alpha']
  values: [{double_val: 0.1}]
}]
```
It might be useful for future readers if this is hardcoded in the test.

95-martin-orion (changes requested):

cirq/google/engine/calibration.py
```diff
@@ -125,11 +128,15 @@ def to_proto(self) -> v2.metrics_pb2.MetricsSnapshot:
         """Reconstruct the protobuf message represented by this class."""
         proto = v2.metrics_pb2.MetricsSnapshot()
         for key in self._metric_dict:
-            for target, value_list in self._metric_dict[key].items():
+            for targets, value_list in self._metric_dict[key].items():
                 current_metric = proto.metrics.add()
                 current_metric.name = key
-                current_metric.targets.extend(
-                    [v2.qubit_to_proto_id(q) for q in target])
+                for target in targets:
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/engine/calibration.py
```diff
@@ -174,6 +181,30 @@ def timestamp_str(self,
...
 
+    def str_to_key(self, target: str) -> Union[devices.GridQubit, str]:
+        """Turns a string into a calibration key.
+
+        Attempts to parse it as a GridQubit.  If this fails,
+        returns the string itself.
+        """
+        try:
+            return v2.grid_qubit_from_proto_id(target)
+        except:
```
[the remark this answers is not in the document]
dstrain115: Done.

cirq/google/engine/calibration.py
dstrain115: Wouldn't this just be float(x) though?  I am not sure I see the point in adding this.

The value is Union[str, int, float].  If it is int or float, float() will convert it.  If it is a str, float() will either convert it or raise a ValueError with a descriptive message "could not convert string to float: 'x'"

I think this is exactly what value_to_float would do.  float() is more standard and more concise.

cirq/google/engine/calibration.py
```diff
@@ -174,6 +181,30 @@ def timestamp_str(self,
...
+    def key_to_qubit(self, target: METRIC_KEY) -> devices.GridQubit:
+        """Returns a single qubit from a metric key.
+
+        If the metric key is multiple qubits, return the first one.
+
+        Raises:
+           ValueError if the metric key is a tuple of strings.
+        """
+        if (isinstance(target, tuple) and
+                isinstance(target[0], devices.GridQubit)):
```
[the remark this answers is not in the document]
dstrain115: Yes, for global metrics.  Good catch.  Fixed.

cirq/google/engine/calibration_test.py
dstrain115: Added.

cirq/google/engine/calibration.py
95-martin-orion: `METRIC_VALUE` is a `List[Union[str, int, float]]` - for a value `x`, we use `float(x[0])` to retrieve it. If you don't see a need for this, though, I'm happy to leave as-is.

95-martin-orion (approved):
See my reply on `value_to_float`, but otherwise this LGTM.

cirq/google/engine/calibration.py
dstrain115: Ok, I added this function to the class.

cirq/google/engine/calibration.py
```diff
@@ -212,13 +228,12 @@ def heatmap(self, key: str) -> vis.Heatmap:
         """
         metrics = self[key]
         assert all(len(k) == 1 for k in metrics.keys()), (
-            'Heatmaps are only supported if all the targets in a metric'
-            ' are single qubits.')
+            'Heatmaps are only supported if all the targets in a metric' ' are single qubits.'
```
95-martin-orion, line 231: Looks like an instance of #3559 here.

95-martin-orion (approved):
New function looks good.

cirq/google/engine/calibration.py
dstrain115: Fixed.

PR #3538, opened 2020-11-25
ParamResolver composition.

95-martin-orion:
This is convenient to have for subcircuits (#3235), since nested circuits must be able to resolve their parameters based on the parent's parameter mapping.

### DIFF
--- a/cirq/study/resolver.py
+++ b/cirq/study/resolver.py
@@ -20,6 +20,7 @@ import sympy
 from sympy.core import numbers as sympy_numbers
 from cirq._compat import proper_repr
 from cirq._doc import document
+from cirq import protocols
 
 if TYPE_CHECKING:
     import cirq
@@ -180,6 +181,13 @@ class ParamResolver:
         else:
             self._deep_eval_map[value] = self.value_of(v, recursive)
         return self._deep_eval_map[value]
+
+    def _resolve_parameters_(self, param_resolver: 'ParamResolver'
+                            ) -> 'ParamResolver':
+        return ParamResolver({
+            k: protocols.resolve_parameters(v, param_resolver)
+            for k, v in self.param_dict.items()
+        })
 
     def __iter__(self) -> Iterator[Union[str, sympy.Symbol]]:
         return iter(self.param_dict)
--- a/cirq/study/resolver.py
+++ b/cirq/study/resolver.py
@@ -20,7 +20,6 @@ import sympy
 from sympy.core import numbers as sympy_numbers
 from cirq._compat import proper_repr
 from cirq._doc import document
-from cirq import protocols
 
 if TYPE_CHECKING:
     import cirq
@@ -186,14 +185,12 @@ class ParamResolver:
         """Returns a resolver with all keys mapped to their final output."""
         return ParamResolver()._resolve_parameters_(self, recursive=True)
 
-    def _resolve_parameters_(self, param_resolver: 'ParamResolver',
-                             recursive: bool) -> 'ParamResolver':
+    def _resolve_parameters_(
+        self, param_resolver: 'ParamResolver', recursive: bool
+    ) -> 'ParamResolver':
         new_dict = {k: k for k in param_resolver}
         new_dict.update({k: self.value_of(k, recursive) for k in self})
-        new_dict.update({
-            k: param_resolver.value_of(v, recursive)
-            for k, v in new_dict.items()
-        })
+        new_dict.update({k: param_resolver.value_of(v, recursive) for k, v in new_dict.items()})
         if recursive and self.param_dict:
             return ParamResolver(new_dict).flatten()
         return ParamResolver(new_dict)
--- a/cirq/study/resolver.py
+++ b/cirq/study/resolver.py
@@ -181,8 +181,8 @@ class ParamResolver:
             self._deep_eval_map[value] = self.value_of(v, recursive)
         return self._deep_eval_map[value]
 
-    def flatten(self) -> 'ParamResolver':
-        """Returns a resolver with all keys mapped to their final output."""
+    def resolved_to_single_step(self) -> 'ParamResolver':
+        """Returns a copy of this resolver with all keys mapped to their final output."""
         return ParamResolver()._resolve_parameters_(self, recursive=True)
 
     def _resolve_parameters_(
@@ -192,7 +192,7 @@ class ParamResolver:
         new_dict.update({k: self.value_of(k, recursive) for k in self})
         new_dict.update({k: param_resolver.value_of(v, recursive) for k, v in new_dict.items()})
         if recursive and self.param_dict:
-            return ParamResolver(new_dict).flatten()
+            return ParamResolver(new_dict).resolved_to_single_step()
         return ParamResolver(new_dict)
 
     def __iter__(self) -> Iterator[Union[str, sympy.Symbol]]:
--- a/cirq/study/resolver.py
+++ b/cirq/study/resolver.py
@@ -181,10 +181,6 @@ class ParamResolver:
             self._deep_eval_map[value] = self.value_of(v, recursive)
         return self._deep_eval_map[value]
 
-    def resolved_to_single_step(self) -> 'ParamResolver':
-        """Returns a copy of this resolver with all keys mapped to their final output."""
-        return ParamResolver()._resolve_parameters_(self, recursive=True)
-
     def _resolve_parameters_(
         self, param_resolver: 'ParamResolver', recursive: bool
     ) -> 'ParamResolver':
@@ -192,7 +188,9 @@ class ParamResolver:
         new_dict.update({k: self.value_of(k, recursive) for k in self})
         new_dict.update({k: param_resolver.value_of(v, recursive) for k, v in new_dict.items()})
         if recursive and self.param_dict:
-            return ParamResolver(new_dict).resolved_to_single_step()
+            new_resolver = ParamResolver(new_dict)
+            # Resolve down to single-step mappings.
+            return ParamResolver()._resolve_parameters_(new_resolver, recursive=True)
         return ParamResolver(new_dict)
 
     def __iter__(self) -> Iterator[Union[str, sympy.Symbol]]:
--- a/cirq/study/resolver.py
+++ b/cirq/study/resolver.py
@@ -182,12 +182,21 @@ class ParamResolver:
             self._deep_eval_map[value] = self.value_of(v, recursive)
         return self._deep_eval_map[value]
 
-    def _resolve_parameters_(self, param_resolver: 'ParamResolver'
-                            ) -> 'ParamResolver':
-        return ParamResolver({
-            k: protocols.resolve_parameters(v, param_resolver)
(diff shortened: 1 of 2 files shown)


cirq/study/resolver.py
```diff
@@ -154,6 +155,13 @@ def value_of(self,
         # No known way to resolve this variable, return unchanged.
         return value
 
+    def _resolve_parameters_(self, param_resolver: 'ParamResolver'
+                            ) -> 'ParamResolver':
+        return ParamResolver({
+            k: protocols.resolve_parameters(v, param_resolver)
+            for k, v in self.param_dict.items()
+        })
```
viathor, line 163: In order to avoid surprises, composition should be [associative](https://en.wikipedia.org/wiki/Associative_property), but this implementation isn't. Consider

```python
pr1 = ParamResolver({a: b})
pr2 = ParamResolver({})
pr3 = ParamResolver({b: 10})
```

Since `self` (i.e. LHS) controls what keys go into the result, we see that pr2+pr3 is empty and therefore pr1+(pr2+pr3) is pr1. OTOH, pr1+pr2 is just pr1 so (pr1+pr2)+pr3 is `ParamResolver({a: 10})`.

cirq/study/resolver_test.py
```diff
@@ -166,6 +166,32 @@ def test_formulas_in_param_dict():
...
+    r23 = cirq.resolve_parameters(r2, r3)
+    assert r23.value_of('b') == 12 + d
+
+    r123 = cirq.resolve_parameters(r12, r3)
+    assert r123.value_of('a') == 12 + d
+
+    r13 = cirq.resolve_parameters(r1, r3)
+    assert r13.value_of('a') == b
+
+
```
viathor, line 194: Perhaps add a unit test for associativity?

viathor (changes requested):

cirq/study/resolver.py
95-martin-orion: I suspect that it is not feasible to achieve this without significant changes to the core behavior of `ParamResolver`. My basis for this is that `cirq.resolve_parameters` attempts to fully resolve each parameter using the given resolver, but only succeeds if the resolution proceeds in alphabetical order:
```
a, b, c, d = [sympy.Symbol(x) for x in ['a', 'b', 'c', 'd']]

# Multiple resolution steps are taken...
assert cirq.resolve_parameters(a, {a: b, b: c}) == c

# ...but only in alphabetical order.
assert cirq.resolve_parameters(c, {c: b, b: a}) == b
```
This is not a Cirq issue - it's just how `sympy.Basic.subs` behaves. From `help(sympy.Basic.subs)`:
> In order to obtain a canonical result, unordered iterables are sorted by count_op length, number of arguments and by the default_sort_key to break any ties.

Example:
```
assert c.subs({c: b, b: a}) == b
```
I tried one-step parameter resolution, but that causes issues when reassigning values:
```
r1 = cirq.ParamResolver({a: b})
r2 = cirq.ParamResolver({c: d, d: 1})
r3 = cirq.ParamResolver({d: 2})

r12 = cirq.resolve_parameters(r1, r2)    # {a: b, c: 1, d: 1}
r23 = cirq.resolve_parameters(r2, r3)    # {c: 2, d: 2}
r12_3 = cirq.resolve_parameters(r12, r3)  # {a: b, c: 1, d: 2} 
r1_23 = cirq.resolve_parameters(r1, r23)  # {a: b, c: 2, d: 2}
```
-----
This could probably be resolved by enforcing one-step parameter resolution in `ParamResolver` with `simultaneous=True` in the `subs` call, but that's a breaking change.

95-martin-orion: Opened #3544 to track this.

viathor: Ah, this is a fairly subtle problem. Thanks for identifying it and for filing the issue!

I'm still wondering whether we could change this function here in a way that would make it easy for us to make resolver composition associative if/when the resolution mechanism is made one-step.

One way to avoid the problem in your example would be to require that there be no conflicts between resolvers being composed. Specifically, we could raise an exception if the two resolvers both mapped the same key to a different value. WDYT?

95-martin-orion: There are two issues that break associativity in the current version: lack of implicit maps (k: k for all unlisted keys) and mixing of resolution depth (`self.value_of` vs. `self.param_dict.get`). In a one-step resolution world, the following should work:
```
def _resolve_parameters_(self, param_resolver):
  new_dict = {k: k for k in param_resolver}
  new_dict.update({k: self.value_of(k) for k in self})
  new_dict.update({k: param_resolver.value_of(v) for k, v in new_dict.items()})
  return ParamResolver(new_dict)
```
(This may also work for full resolution, but I'm not as certain on that point.)

In practice:
```
a, b, c, d = [sympy.Symbol(x) for x in ['a', 'b', 'c', 'd']]
r1 = cirq.ParamResolver({a: b})
r2 = cirq.ParamResolver({c: d, d: 1})
r3 = cirq.ParamResolver({d: 2})

r12 = cirq.resolve_parameters(r1, r2)    # {a: b, c: d, d: 1} - no "chain resolution" from the new resolver
r23 = cirq.resolve_parameters(r2, r3)    # {c: 2, d: 1}
r12_3 = cirq.resolve_parameters(r12, r3)  # {a: b, c: 2, d: 1} 
r1_23 = cirq.resolve_parameters(r1, r23)  # {a: b, c: 2, d: 1}
```
Despite "losing" the `d: 2` mapping from r3, I believe that this gives the correct behavior, since r2 removes any "initial-d" parameters from the circuit. The only parameters that care about the new d-mapping are those that are "currently" d when r3 is applied (e.g. "initial-c" parameters):
```
q0, q1, q2, q3 = cirq.LineQubit.range(4)
c0 = cirq.Circuit(cirq.X(q0) ** a, cirq.X(q1) ** b, cirq.X(q2) ** c, cirq.X(q3) ** d)

c1 = resolve_parameters(c0, r1)  # X(q0)^b, X(q1)^b, X(q2)^c, X(q3)^d
c2 = resolve_parameters(c1, r2)  # X(q0)^b, X(q1)^b, X(q2)^d, X(q3)^1
c3 = resolve_parameters(c2, r3)  # X(q0)^b, X(q1)^b, X(q2)^2, X(q3)^1

c123 = cirq.resolve_parameters(c0, r12_3)
assert c123 == c3
```

95-martin-orion: > (This may also work for full resolution, but I'm not as certain on that point.)

This is only partially true. While the full-resolution behavior of the two resolvers will be the same, their internal mappings may differ, as seen below:
```
r1 = cirq.ParamResolver({a: b})
r2 = cirq.ParamResolver({x: y})
r3 = cirq.ParamResolver({b: x})

r12 = cirq.resolve_parameters(r1, r2)    # {a: b, x: y}
r12_3 = cirq.resolve_parameters(r12, r3)  # {a: x, x: y, b: x} - maps all to y, but transitively
r23 = cirq.resolve_parameters(r2, r3)    # {b: x, x: y} - loss of ordering occurs here
r1_23 = cirq.resolve_parameters(r1, r23)  # {a: y, x: y, b: y}
```
The fix is to "flatten" the resulting ParamResolver post-composition. This has the added perk of simplifying the internal mapping, which speeds up future parameter resolution:
```
def flatten(self):
  # resolving the empty dict against a PR flattens that PR
  return ParamResolver()._resolve_parameters_(self, recursive=True)
```

viathor: Nice analysis. I think this should work. In particular, I think you're right that we don't need to raise an exception when there is a conflict as long as we choose consistently which resolver wins (above we choose RHS which is fine). After all, commutativity is *not* a requirement (just as it is not in case of other forms of composition).

I think this is tricky enough that we should really have a number of unit tests that exercise interesting scenarios (see other comment), each in both "once" and "recursive" modes (as long as both modes are possible; some scenarios will probably fail in the "recursive" mode, e.g. a: a+1). The cases we considered above are probably a good starting point.

95-martin-orion:
Force-pushed the current version rebased to #3546 (minus some of the more recent changes). Not ready for review, but this shows the rough suite of tests I intend to use.

cirq/study/resolver_test.py
```diff
@@ -159,11 +159,156 @@ def test_formulas_in_param_dict():
...
+    assert sympy.Eq(r123.value_of('a'), 12 + d)
+
+    r13 = cirq.resolve_parameters(r1, r3)
+    assert r13.value_of('a') == b
+
+
+@pytest.mark.parametrize('resolve_fn', [
+    cirq.resolve_parameters,
+    cirq.resolve_parameters_once,
+])
```
viathor, line 260: Test parametrization is generally used to achieve separation of test data from test logic as in [table-driven tests](https://github.com/golang/go/wiki/TableDrivenTests). This is desirable because test logic is repetitive and we often want to perform the same operation on multiple inputs. Thus, this in a sense inverts the intended use of test parametrization: it uses test parameter to select code-under-test, but hardcodes test data within the test.

I suggest something like this instead:

```python
# Top of file up above
a, b, c, d, = sympy.Symbol('a'), sympy.Symbol('b'), ...

# ...

@pytest.mark.parametrize('p1, p2, p3', (
    ({a: b}, {c: d, d: 1}, {d: 2}),
    ({a: b}, {}, {b: 10}),
    ({a: 1}, {}, {}),
    ({}, {a: 1}, {}),
    ({}, {}, {a: 1}),
    # ...
))
def test_composition_is_associative(p1, p2, p3):
    r1 = cirq.ParamResolver(p1)
    r2 = cirq.ParamResolver(p2)
    r3 = cirq.ParamResolver(p3)
    for resolve_fn in (cirq.resolve_parameters, cirq.resolve_parameters_once):
        assert resolve_fn(r1, resolve_fn(r2, r3)) == resolve_fn(resolve_fn(r1, r2), r3)
```

This makes test logic general and cleanly separates the inputs we exercise within the decorator. Also, it makes it very easy to extend the test to cover new cases, e.g. if we find bugs.

95-martin-orion: That would definitely be a lot cleaner! I think I'll parameterize on both, since that will make it easier to identify which case is failing by adding the function to the test name.

viathor: Good point. You can probably use `itertools.product` here to make sure each test case is matched with each function :-)

95-martin-orion: `pytest` has us covered: if a test has two separate `parametrize` lines, it will parameterize over their product 😃

cirq/study/resolver.py
```diff
@@ -181,6 +181,20 @@ def value_of(
             self._deep_eval_map[value] = self.value_of(v, recursive)
         return self._deep_eval_map[value]
 
+    def resolved_to_single_step(self) -> 'ParamResolver':
```
viathor, line 184: Do you have other uses for this method? If not, I'd just inline it on line 195 below.

viathor (approved):

cirq/study/resolver.py
95-martin-orion: Done.

PR #3539, opened 2020-11-25
Modular JSON serialization protocol

balopat:
Introduces a modular design for the JSON serialization protocol.

This allows for better separation during PR reviews, smaller, more readable files in general and future possibility to handle optional modules easily. 

The first two modules that are separately registered are `cirq.google` and `cirq` (though currently it's called `cirq.protocol` as the json_test_data folder lives there, but in a follow-up pr this should be moved out to the top level cirq/json_test_data to comply with the pattern).

The modules `cirq.contrib`, `cirq.pasqal`, `cirq.aqt`, `cirq.ionq` will follow - though they are not even registered yet properly!

Key improvements: 
- I pulled out most of the "test framework" into `cirq.testing.json`. 
  -  The actual tests still live in `cirq/protocols/json_serialization_test.py` 
   - When registering a new module, it has to come with the following setup:
     - add the module name to `TESTED_MODULES` variable in json_serialization_test to opt-in for central testing
     - test data: a new package to contain the test data, see for example `cirq/google/json_test_data`
     - test spec: the `json_test_data` package should export an object name `TestSpec`, which should be of type ModuleJsonTestSpec, see for example  `cirq/google/json_test_data/spec.py`
     - resolver cache: a resolver cache for the exposed types - see for example `cirq/json_resolver_cache.py` or `cirq/google/json_resolver_cache.py`
- The `ModuleJsonTestSpec` class formalizes what is required to test the json serialization properties of a module: 
```python
class ModuleJsonTestSpec:
    # for test failures, a better representation
    name: str
    # the packages that have the public classes to be checked for serialization
    packages: List[ModuleType]
    # the path for the folder containing the test files
    test_data_path: pathlib.Path
    # these public class names are exposed but not yet serializable
    not_yet_serializable: List[str]
    # these public class names are exposed but do not need to be serialized
    should_not_be_serialized: List[str]
    # points to the resolver cache's dict for this module
    resolver_cache: Dict[str, ObjectFactory]
```
- I introduced two new consistency checks as it was already getting messy: the "skip-lists" should not contain items that are simultaneously in the resolver cache
- I added a nice user friendly message in case probably a class is missing from a resolver, for a given module
- as members of the `DEFAULT_RESOLVERS` now get lazily registered, as modules initialize, there is no guarantee for Cirq modules that it will have the full list of built-in resolvers. For this reason, for 3rd party packages it is still okay to rely on `DEFAULT_RESOLVERS` but Cirq packages, like `cirq.contrib` should be registered instead into `DEFAULT_RESOLVERS`

Note for reviewers: This is a bigger PR but not monstrous: 29/42 is just trivial json data file movement to `cirq.google` and new `__init__.py` files. It would be hard to break it up.


cirq/contrib/json_test.py
```diff
@@ -12,9 +11,10 @@ def test_quantum_volume():
         heavy_set=[1, 2, 3],
         compiled_circuit=cirq.Circuit(cirq.H.on_each(qubits)),
         sampler_result=.1)
-    assert_json_roundtrip_works(qvr, resolvers=DEFAULT_CONTRIB_RESOLVERS)
+
+    assert_json_roundtrip_works(qvr)
```
Strilanc, line 15: This test should still require you to explicitly say you're including unstable serialization data.

cirq/google/__init__.py
```diff
@@ -87,3 +87,12 @@
 
 from cirq.google.serializable_gate_set import (
     SerializableGateSet,)
+
+
+def _register_resolver():
```
Strilanc, line 92: Type annotation.

cirq/protocols/json_serialization.py
```diff
@@ -225,6 +85,28 @@ def my_read_json(file_or_fn, resolvers=None):
 """
 
 
+def register_resolver(dict_factory: Callable[[], Dict[str, ObjectFactory]]):
```
Strilanc, line 88: This method needs to have "internal only" somewhere in its name, and an unscore prefix, or something else to indicate that users looking through `cirq.protocols` should not be calling it.

cirq/google/json_test_data/spec.py
```diff
@@ -0,0 +1,37 @@
...
+                              test_data_path=pathlib.Path(__file__).parent,
+                              not_yet_serializable=[
+                                  'CalibrationResult', 'CalibrationLayer',
+                                  'FSIM_GATESET', 'SYC_GATESET', 'Sycamore',
+                                  'Sycamore23', 'SerializableDevice',
+                                  'SerializableGateSet', 'SQRT_ISWAP_GATESET',
+                                  'XmonDevice', 'XMON'
+                              ],
+                              should_not_be_serialized=[
+                                  'Engine',
```
Strilanc, line 19: We could have an internal `@_NotJsonSerializable` annotation to localize this information better.

I'm a bit wary of using annotations to opt-in, because then there's no central place to see the list of what can be parsed, but for opt-out it's fine.

cirq/__init__.py
```diff
@@ -557,3 +557,12 @@
     pasqal,
     testing,
 )
+
+
+def _register_resolver():
```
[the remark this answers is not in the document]
balopat: Oh, okay, I'll add `-> None` - as this is a void.

cirq/contrib/json.py
```diff
@@ -2,20 +2,21 @@
...
         QuantumVolumeResult,
         SwapPermutationGate,
     ]
-    d = {cls.__name__: cls for cls in classes}
-    return d.get(cirq_type, None)
+    return {cls.__name__: cls for cls in classes}
 
 
-DEFAULT_CONTRIB_RESOLVERS = [contrib_class_resolver] + DEFAULT_RESOLVERS
+# TODO: contrib should be properly tested
```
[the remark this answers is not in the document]
balopat: https://github.com/quantumlib/Cirq/issues/3555

cirq/contrib/json_test.py
balopat: The new model makes this unnecessary. Contrib is registered now as part of the `DEFAULT_RESOLVERS`. 

Also, as contrib is referenced from within `cirq/__init__.py`, it will not be able to rely on `DEFAULT_RESOLVERS` in a constant at package initialization to create `DEFAULT_CONTRIB_RESOLVERS`, because we can't guarantee that `DEFAULT_RESOLVERS` is "ready" by the time contrib is imported. 

We could unregister contrib potentially and rely on recreating a `contrib-resolvers` within each test but that doesn't seem to be a useful pattern for end users, as they would have to the same in their code.

cirq/google/json_test_data/spec.py
balopat: Good idea - could be a follow-up PR just to not bloat this one further?

cirq/testing/json.py
```diff
@@ -1,8 +1,138 @@
...
+                    # singletons, for instance
+                    obj = obj.__class__
+
+                if name.startswith('_'):
+                    continue
+
+                if inspect.isclass(obj) and inspect.isabstract(obj):
+                    continue
+
+                # assert name != 'XPowGate'
```
95-martin-orion, line 79: Remove debug line?

cirq/testing/json.py
```diff
@@ -1,8 +1,138 @@
...
+def spec_for(module_name: str) -> ModuleJsonTestSpec:
+    import importlib.util
+
+    if importlib.util.find_spec(module_name) is None:
+        raise ModuleNotFoundError(f"{module_name} not found")
+
+    test_module_name = f"{module_name}.json_test_data"
+    if importlib.util.find_spec(test_module_name) is None:
+        raise ValueError(
+            f"{module_name} module is missing json_test_data " f"package, please set it up."
```
95-martin-orion, line 128: Remove string break, e.g. `json_test_data " f"package` -> `json_test_data package`

Same below.

95-martin-orion (dismissed):

95-martin-orion (commented):
No further issues from me. Leaving final approval to @Strilanc.

balopat:
> No further issues from me. Leaving final approval to @Strilanc.

Thank you, I'll dismiss your review then. @Strilanc can you please LGTM this?

cirq/contrib/json_test.py
Strilanc: I'm very uncomfortable include contrib in the default resolvers. Contrib code is unstable, it shouldn't be included in the definition of a guaranteed-to-be-stable file format.

Strilanc (approved):

cirq/contrib/json_test.py
balopat: OK, I'll think about this a bit more before merging.

cirq/google/json_test_data/spec.py
balopat: I forgot to open this issue, and forgot about the idea and then re-invented it :) I'll work on it after this. https://github.com/quantumlib/Cirq/issues/3734

cirq/contrib/json_test.py
balopat: The root of the problem was that contrib is imported from cirq proper. After moving contrib to after the resolver cache registration, the tests pass now.

PR #3540, opened 2020-11-26
JSON Serialization for CalibrationResult and CalibrationLayer

dstrain115:
- Serializes CalibrationResult dataclass
    - Uses underlying Calibration serialization and datetime utc timestamp.
- Serializes CalibrationLayer dataclass using json helper

These objects are the client objects for calibration calls in engine.


mpharrigan (approved):
looks good

cirq/protocols/json_test_data/CalibrationResult.json
```diff
@@ -0,0 +1,79 @@
+{
```
balopat, line 1: these should live under cirq.google.json_test_data

balopat (changes requested):
lgtm but the json test data is in the wrong place

balopat (approved):

PR #3546, opened 2020-11-27
Provide one-step and recursive parameter resolution

### LINKED ISSUE #3544: cirq.resolve_parameters sometimes fails to fully resolve parameters
**Description of the issue**

Previous discussion: #3538 

When calling `cirq.resolve_parameters` with a multi-step ParamResolver (e.g. `{a: b, b: c}`), whether or not the resolution completes depends on the alphabetical ordering of the symbols involved. From `sympy.Basic.subs`:
> In order to obtain a canonical result, unordered iterables are sorted by count_op length, number of arguments and by the default_sort_key to break any ties.

Given that this is the expected behavior in sympy, we either need to switch to one-step parameter resolution or hack together our own recursive multi-step resolution process.

```
import cirq, sympy
a, b, c = [sympy.Symbol(x) for x in ['a', 'b', 'c']]

# returns c
cirq.resolve_parameters(a, {a: b, b: c})

# returns b, but we want a
cirq.resolve_parameters(c, {c: b, b: a})
```

**Cirq version**

0.10.0.dev

95-martin-orion:
Fixes #3544.

### Recursive parameter resolution

Calling `ParamResolver.value_of` with `recursive=True` (the default value) will attempt to recursively resolve the given value. Intermediate evaluations are memoized to provide better performance for heavily-nested resolvers, or on subsequent calls to the same resolver.

This behavior is shared by `cirq.resolve_parameters`, although an actual `ParamResolver` (instead of a dict) must be passed to that method to receive the full benefit of memoization.

Known side effects of this change:
- Parameter resolution no longer depends on alphabetization of parameters involved (😀)
- Cirq will raise a `RecursionError` when encountering a loop in parameter resolution (🙂)
- Resolver is more zealous in its conversion of `int` to `float` (😐)

### Single-step parameter resolution

Calling `ParamResolver.value_of` with `recursive=False` will instead perform a single parameter resolution step. This is convenient for operations that would be illegal in recursive resolution, such as incrementing a parameter with `{a: a+1}` or swapping a pair of parameters with `{a: b, b: a}`. Single-step resolution does not affect the memoization used in recursive resolution.

This behavior is shared by the new `cirq.resolve_parameters_once` protocol method.


95-martin-orion:
The failing coverage checks highlighted that the `_resolve_parameters_` magic function must also be updated to allow for recursive resolution. Marking as WIP until update is in place.

95-martin-orion:
Dropping [WIP] tag as coverage tests are now passing. For ease of review, I recommend looking at the "Add recursive" commits separately from the rest - they just propagate the `_resolve_parameters_` changes.

95-martin-orion:
**Known issue:** mapping expressions to symbols is currently handled incorrectly. For example:
```
a, b, c, d = [sympy.Symbol(x) for x in 'abcd']
r = cirq.ParamResolver({a: b + c, b + c, d})
cirq.resolve_parameters(a, r)  # returns b + c, not d
```
However, this matches the previous behavior (#3550) so the fix can be captured in a separate PR.

cirq/circuits/circuit.py
```diff
@@ -2001,12 +2000,17 @@ def with_noise(self, noise: 'cirq.NOISE_MODEL_LIKE') -> 'cirq.Circuit':
...
-        resolved_operations.append(protocols.resolve_parameters(
-            op, param_resolver))
+    if recursive:
+        for op in operations:
+            resolved_operations.append(
+                protocols.resolve_parameters(op, param_resolver))
+    else:
+        for op in operations:
+            resolved_operations.append(
+                protocols.resolve_parameters_once(op, param_resolver))
```
viathor, line 2013: This conditional on `recursive` flag is present throughout this PR which suggests that client code would be simpler if we combined `resolve_parameters` and `resolve_parameters_once` protocols into one that takes the recursive flag.

cirq/ops/raw_types.py
```diff
@@ -625,10 +625,17 @@ def _parameter_names_(self) -> AbstractSet[str]:
         }
         return protocols.parameter_names(self.sub_operation) | tag_params
 
-    def _resolve_parameters_(self, resolver):
-        resolved_op = protocols.resolve_parameters(self.sub_operation, resolver)
-        resolved_tags = (
-            protocols.resolve_parameters(tag, resolver) for tag in self._tags)
+    def _resolve_parameters_(self, resolver, recursive):
```
viathor, line 628: missing type annotations

cirq/protocols/resolve_parameters.py
```diff
@@ -47,7 +47,8 @@ def _parameter_names_(self: Any) -> AbstractSet[str]:
         """
 
     @doc_private
-    def _resolve_parameters_(self: Any, param_resolver: 'cirq.ParamResolver'):
+    def _resolve_parameters_(self: Any, param_resolver: 'cirq.ParamResolver',
+                             recursive: bool):
```
viathor, line 51: missing return type annotation

cirq/protocols/resolve_parameters.py
```diff
@@ -47,7 +47,8 @@ def _parameter_names_(self: Any) -> AbstractSet[str]:
         """
 
     @doc_private
-    def _resolve_parameters_(self: Any, param_resolver: 'cirq.ParamResolver'):
+    def _resolve_parameters_(self: Any, param_resolver: 'cirq.ParamResolver',
```
viathor, line 50: It is possible to express the fact that the function takes Any type, but returns the same type using [generic functions](https://mypy.readthedocs.io/en/stable/generics.html#generic-functions).

(Also applies to `resolve_parameters_once` elsewhere.)

viathor (commented):
Thanks for doing this! I had a quick first pass and have a few suggestions. Will return to  this later. Let me know what you think.

cirq/circuits/circuit.py
95-martin-orion: Good point. I think I'll keep `resolve_parameters_once` as an alias for `resolve_parameters({...}, recursive=False)` though, since it has a nice parallel with `decompose[_once]`.

cirq/protocols/resolve_parameters.py
95-martin-orion: Unfortunately this generates mypy errors, since raw Symbols can resolve to floats while other types expect the input type back. `Union[float, BaseType]` gives errors in places that explicitly require `BaseType` back.

I don't intend to address this in this PR since I've already rabbit-holed away from subcircuits a fair amount, but please open an issue if you'd like to see this resolved.

viathor: SGTM!

cirq/circuits/circuit.py
95-martin-orion: This is now done.

95-martin-orion:
Pinging @viathor for review - this is the current head of the PR chain for subcircuits (#3235).

cirq/protocols/resolve_parameters.py
```diff
@@ -121,7 +119,9 @@ def parameter_symbols(val: Any) -> AbstractSet[sympy.Symbol]:
     return {sympy.Symbol(name) for name in parameter_names(val)}
 
 
-def resolve_parameters(val: Any, param_resolver: 'cirq.ParamResolverOrSimilarType') -> Any:
+def resolve_parameters(
+    val: Any, param_resolver: 'cirq.ParamResolverOrSimilarType', recursive: bool = True
```
viathor, line 123: need to add `recursive` to  Args section of docstring

cirq/protocols/resolve_parameters.py
```diff
@@ -134,24 +134,31 @@ def resolve_parameters(val: Any, param_resolver: 'cirq.ParamResolverOrSimilarTyp
...
-    result = NotImplemented if getter is None else getter(param_resolver)
+    result = NotImplemented if getter is None else getter(param_resolver, recursive)
 
     if result is not NotImplemented:
         return result
     else:
         return val
+
+
+def resolve_parameters_once(val: Any, param_resolver: 'cirq.ParamResolverOrSimilarType'):
```
viathor, line 163: missing docstring

cirq/protocols/resolve_parameters_test.py
```diff
@@ -94,11 +109,37 @@ def _resolve_parameters_(self, resolver: ParamResolver):
...
             return 5
 
     t = Tester()
-    assert cirq.resolve_parameters(t, {}) is t
-    assert cirq.resolve_parameters(t, {'x': 2}) == 5
+    assert resolve_fn(t, {}) is t
+    assert resolve_fn(t, {'x': 2}) == 5
+
+
+def test_resolve_loops():
```
viathor, line 129: better name: `test_recursive_resolve` (since only the last resolver has an actual loop)

cirq/study/resolver.py
```diff
@@ -42,7 +42,7 @@ class ParamResolver:
     A ParamResolver is an object that can be used to assign values for these
     variables.
 
-    ParamResolvers are hashable.
+    ParamResolvers are hashable; their param_dict must not be mutated.
```
viathor, line 45: Out of scope of this PR: I wonder if it's easy to make it private

cirq/study/resolver.py
```diff
@@ -121,33 +130,57 @@ def value_of(self, value: Union['cirq.TParamKey', float]) -> 'cirq.TParamVal':
...
-        # No known way to resolve this variable, return unchanged.
-        return value
+        # Recursive parameter resolution. We can safely assume that value is a
+        # single symbol, since combinations are handled earlier in the method.
+        try:
+            v = self._deep_eval_map[value]
+            if v is not None:
+                return v
+            raise RecursionError('Evaluation of {value} indirectly contains itself.')
+        except KeyError:
```
viathor, line 173: This isn't really an error, but one of the normal branches of the algorithm. We should handle it with a regular conditional:

```python
if value in self._deep_eval_map:
    v = self._deep_eval_map[value]
    if v is not None:
        return v
    raise RecursionError(...)
self._deep_eval_map[value] = None
```

cirq/study/resolver.py
```diff
@@ -82,9 +85,15 @@ def value_of(self, value: Union['cirq.TParamKey', float]) -> 'cirq.TParamVal':
...
             value: The parameter to try to resolve.
+            recursive: Whether to recursively evaluate formulas. If this is
+                set to True, the evaluation tree will be memoized in
+                self._deep_eval_map.
 
         Returns:
             The value of the parameter as resolved by this resolver.
+
+        Raises:
+            RecursionError if the ParamResolver detects a loop in resolution.
```
viathor, line 96: Maybe "...a loop in a recursive resolution." to make it clear this cannot happen when `recursive=False`.

cirq/study/resolver.py
```diff
@@ -82,9 +85,15 @@ def value_of(self, value: Union['cirq.TParamKey', float]) -> 'cirq.TParamVal':
 
         Args:
             value: The parameter to try to resolve.
+            recursive: Whether to recursively evaluate formulas. If this is
+                set to True, the evaluation tree will be memoized in
+                self._deep_eval_map.
```
viathor, line 90: I'd leave out this implementation detail from docstring.

viathor (approved):
LGTM with a few more minor comments. Feel free to merge once they're addressed.

cirq/study/resolver.py
95-martin-orion: I tried making it a `@property`, but that apparently doesn't stop users from mutating its contents :(

cirq/study/resolver.py
95-martin-orion: I had the impression that this usage of `try...except` was the Pythonic way of doing "if x in dict, get x". Open to changing it if there's an issue that I'm missing here.

viathor: I disagree that triggering a `KeyError` is a pythonic way to check whether a key exists in a dict. Python provides the `in` operator to do this directly and cleanly. BTW: A second pythonic way is the `dict.get(key, default)` function, but I don't think it's very helpful here.

A try-except block is suitable to handle actual errors, not normal control flow. In particular, it should be used when errors may originate at various function call depth and so local checks cannot detect error condition. Also, handling errors this way comes with costs such as the need to instantiate the exception, to do an `isinstance` check and to potentially unwind the stack. There is also the readability cost due to the clutter introduced by extra keywords and type names.

viathor: BTW: I checked whether we do this in cirq and only found four instances. I didn't count the number of times we use the `in` operator directly, but searched for `'if .* in [a-zA-Z_]*:'` as a rough proxy. Found 87 instances. Thus, even by the standard of consistency with cirq coding style, we should not be triggering `KeyError` here.

95-martin-orion: Fair enough. I wasn't able to find a source for my impression either, so I've made the change.

viathor: As @mpharrigan points out [here](https://github.com/quantumlib/Cirq/pull/3568#issuecomment-739070461), I was wrong about try-except being a non-pythonic way of doing this.

Except for the four cases which I changed in #3568, we had generally been avoiding using try-except blocks to check key presence, so I still think we should prefer the `in` operator for local consistency and also for other reasons listed above (efficiency and readability costs and, in some instances, the risk of catching exceptions raised deeper down the call stack). That said, I'm open to changing my mind if people want to do it the EAFP way going forward.

mpharrigan: I don't actually like EAFP, so works for me

Issue #3529, opened 2020-11-24
Functionality decoupling for "_strat_commutes_from_operation"


akushnarov:
Move "commutes" check from the protocols to the "Operation" class.
- Now: "_strat_commutes_from_operation" is a part of [commutes_protocol.py](https://github.com/quantumlib/Cirq/blob/master/cirq/protocols/commutes_protocol.py)
- Should be: part of "\_commutes\_" in [ops.Operation](https://github.com/quantumlib/Cirq/blob/master/cirq/ops/raw_types.py)
(This is related to the #3274) 

P3 - I'm not really blocked by it, it is an idea I'd like to discuss / suggestion based on principle

balopat:
This sounds like a good idea in principle, but maybe there was a reason that it lives in the protocol module. Would you like to try if it will work on the Operation class? It should be quick to try to make all the tests pass.

akushnarov:
Hey @balopat, I did some research and here is the following output:
- it's possible to move the logic from `_strat_commutes_from_operation` to the Operation class (with some adjustments)
- if the functionality is moved, the tests from `protocols/commutes_protocol_test.py` fail (the reason is not in the moved code itself)

The reason of test fail is in the  class "GateOperation" which implements `_commute_` with the `self.gate._commutes_on_qids_`:
`
def _commutes_(self, other: Any,
                   atol: float) -> Union[bool, NotImplementedType, None]:
        return self.gate._commutes_on_qids_(self.qubits, other, atol=atol)
`
but that method itself does nothing, but returns NotImplemented:
`
def _commutes_on_qids_(self, qids: 'Sequence[cirq.Qid]', other: Any,
                           atol: float
                          ) -> Union[bool, NotImplementedType, None]:
        return NotImplemented
`

**So my suggestion would be**:
- move the `_strat_commutes_from_operation` functionality to the Operation
- remove `_commute_` from the `GateOperation` (anyway it does nothing)
- adjust the tests

vtomole:
@akushnarov 

What is gained by this change?

akushnarov:
Hi @vtomole 

- _Reason for having this request_: Moving the Operation's `_commute_` functionality from the general "commute strategy" (which should not know anything about object types and should only orchestrate the commute logic) to the Operation class (where the commute functionality should be controlled)

[...]

PR #3548, opened 2020-11-30
Proposed fix for #3529 (Functionality decoupling for _strat_commutes_…

akushnarov:
1. Moved "commute" check from `protocols/commutes_protocol.py` to `ops/raw_types.py:Operator`
2. Adjusted the "ops/gate_operation.py" (extended `_commute_` functionality to check if the parent class commutes)


cirq/protocols/commutes_protocol_test.py
```diff
@@ -15,6 +15,7 @@
 import numpy as np
 import pytest
 import sympy
+import copy
```
vtomole, line 18: You are not using this import. Remove it.

cirq/protocols/commutes_protocol.py
```diff
@@ -185,31 +184,4 @@ def _strat_commutes_from_matrix(
...
-    # Don't create gigantic matrices.
-    if np.product(qid_shape_protocol.qid_shape(circuit12)) > 2**10:
-        return NotImplemented  # coverage: ignore
-
-    m12 = unitary_protocol.unitary(circuit12, default=None)
-    m21 = unitary_protocol.unitary(circuit21, default=None)
-    if m12 is None:
-        return NotImplemented
-    return np.allclose(m12, m21, atol=atol)
+    return linalg.matrix_commutes(v1, v2, atol=atol)
```
vtomole, line 187: Add newline at the end of this file.

cirq/ops/pauli_string.py
```diff
@@ -652,8 +652,8 @@ def zip_paulis(self, other: 'cirq.PauliString'
                   ) -> Iterator[Tuple[pauli_gates.Pauli, pauli_gates.Pauli]]:
         return (paulis for qubit, paulis in self.zip_items(other))
 
-    def _commutes_(self, other: Any,
-                   atol: float) -> Union[bool, NotImplementedType, None]:
+    def _commutes_(self, other: Any, *, atol: Union[int, float] = 1e-8
```
vtomole, line 655: Why the addition of ` 1e-8`?

vtomole (changes requested):

cirq/protocols/commutes_protocol_test.py
akushnarov: Roger that

cirq/protocols/commutes_protocol.py
akushnarov: This comment is a bit confusing since this code (lines 213:215 from the comment) was moved to ops/raw_types.py:Operation:_commutes_

cirq/ops/pauli_string.py
akushnarov: I've just added a default value. Across Cirq code there are sometimes `_commutes_` methods with the default value and sometimes without, e.g.:
- cirq/ops/raw_types.py:Gate -> `def _commutes_(self, other: Any, atol: float) -> Union[None, NotImplementedType, bool]:`
- cirq/ops/raw_types.py:TaggedOperation -> `def _commutes_(self, other: Any, *, atol: Union[int, float] = 1e-8) -> Union[bool, NotImplementedType, None]:`

What are the advantages of removing the default value?

cirq/ops/raw_types.py
```diff
@@ -507,6 +507,36 @@ def validate_args(self, qubits: Sequence['cirq.Qid']):
         """
         _validate_qid_shape(self, qubits)
 
+    def _commutes_(
+        self, other: Any, *, atol: Union[int, float] = 1e-8
+    ) -> Union[bool, NotImplementedType, None]:
+        """Determine if this Operation commutes with the object"""
+        if isinstance(other, str):
```
balopat, line 514: Why aren't we checking for `other` to be the instance of `ops.Operation`? That was the original check as well.

cirq/ops/pauli_string.py
```diff
@@ -698,7 +698,10 @@ def zip_paulis(
     ) -> Iterator[Tuple[pauli_gates.Pauli, pauli_gates.Pauli]]:
         return (paulis for qubit, paulis in self.zip_items(other))
 
-    def _commutes_(self, other: Any, atol: float) -> Union[bool, NotImplementedType, None]:
+    def _commutes_(
```
balopat, line 701: this seems unrelated - can you revert this formatting change?

cirq/ops/gate_operation.py
```diff
@@ -167,8 +167,18 @@ def _unitary_(self) -> Union[np.ndarray, NotImplementedType]:
...
-        return self.gate._commutes_on_qids_(self.qubits, other, atol=atol)
+    def _commutes_(
+        self, other: Any, atol: Union[int, float] = 1e-8
+    ) -> Union[bool, NotImplementedType, None]:
+
+        commutes = self.gate._commutes_on_qids_(self.qubits, other, atol=atol)
+        if commutes is not NotImplemented:
+            return commutes
+
+        getter = getattr(super(), '_commutes_', None)
```
balopat, line 178: I have two questions: 
1. What motivated the changes in this method? 
2. Also, why do we need to check for the existence of `_commutes_` on `super()`? We _know_ that `cirq.Operation` implements `_commutes_` now, as we just implemented it...

cirq/ops/raw_types.py
akushnarov: 1. In the prev implementation (with the strategy) we are checking that at least one of the arguments is Operation (because it might be anything when you pass to cirq.commutes(v1, v2))
2. In new implementation (a method `_commutes_` of the Operation class) we already know that `self` is an instance of the Operation class and `other` might be Operation or Gate or etc.

akushnarov:
@balopat please see my answers above, also I adjusted the code

cirq/ops/raw_types.py
balopat: The original logic was this: 

```
if not isinstance(v1, ops.Operation) or not isinstance(v2, ops.Operation):	
        return NotImplemented
```

This means that both values should be `Operation` otherwise it is not `NotImplemented`.

akushnarov: Oh, sorry, you are 100% right, my bad.  Fixed in the last commit

akushnarov:
Hi @balopat, I'm back to Cirq. Could you please check the changes?

cirq/ops/pauli_string.py
balopat: in fact there seems to be some inconsistency around this in the codebase:

image

Alright, let's keep the * in front of atol, I'll open an issue to cleanup this inconsistency.

cirq/ops/gate_operation.py
```diff
@@ -167,8 +167,14 @@ def _unitary_(self) -> Union[np.ndarray, NotImplementedType]:
             return getter()
         return NotImplemented
 
-    def _commutes_(self, other: Any, atol: float) -> Union[bool, NotImplementedType, None]:
-        return self.gate._commutes_on_qids_(self.qubits, other, atol=atol)
+    def _commutes_(
```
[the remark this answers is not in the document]
balopat: OK, I'll merge this, and then we can fix the inconsistency in #3695

balopat (approved):

PR #3551, opened 2020-11-30
Update requirements.txt

m-szalay:
Fixes compatibility issues introduced by pip 20.3 and its more aggressive package dependencies resolver.


balopat (changes requested):
Let's bump this to a pin to 3.13.0 for now to avoid surprises in CI. 
To relax the `install_requires` constraints I opened https://github.com/quantumlib/Cirq/issues/3552.

balopat (approved):

PR #3562, opened 2020-12-02
Serialize measurement keys through IonQ metadata

dabacon:
IonQ API v0.1 does not have measurement gates.  To handle measurement gates we store their keys in a `metadata` field which the API preserves on a job. We expect this to be fixed in future API version, so this is written in a way that will make that upgrade faster (i.e. serialized a placeholder measurement).


dabacon:
Part of #3479

cirq/ionq/job.py
```diff
@@ -154,15 +148,25 @@ def repetitions(self) -> Optional[int]:
...
             return int(self._job['metadata']['shots'])
         return None
 
+    def measurement_dict(self) -> Dict[str, Sequence[int]]:
+        """Returns a dictionary of measurement keys to target qubit index."""
+        measurement_dict: Dict[str, Sequence[int]] = {}
+        if 'metadata' in self._job:
+            for key, value in self._job['metadata'].items():
+                # If we had to pass shots, ignore, it cannot be a measurement key.
+                if key == 'shots':
```
gmauricio, line 165: what if we use a specific key for measurement keys, instead of them being in the root, so it would look something like this:
`{ "shots": 100, "measurements": { ... } }`  Just to avoid this "special" handling in case we extend metadata with more fields.

cirq/ionq/serializer.py
```diff
@@ -50,46 +65,84 @@ def __init__(self, atol: float = 1e-8):
...
+                'IonQ API must use LineQubits from 0 to number of qubits - 1. Instead found line '
+                f'qubits with indices {all_qubits}.'
+            )
         num_qubits = cast(line_qubit.LineQubit, max(all_qubits)).x + 1
-        return {'qubits': num_qubits, 'circuit': self._serialize_circuit(circuit, num_qubits)}
+        return num_qubits
 
-    def _serialize_circuit(self, circuit: 'cirq.Circuit', num_qubits: int) -> list:
+    def _validate_metadata(self, metadata: dict):
+        if len(metadata) > 9:
```
gmauricio, line 113: I guess this also depends on other fields added to the metadata dict. Another reason to put measurement keys under specific metadata dict key, so we only check size on that one.

cirq/ionq/ionq_client.py
```diff
@@ -95,8 +93,8 @@ def _target(self, target: Optional[str]) -> str:
             AssertionError: if both `target` and `default_target` are not set.
         """
         assert target is not None or self.default_target is not None, (
-            'One must specify a target on this call, or a default_target on '
-            'the service/client, but neither were set.'
+            'One must specify a target on this call, or a default_target on he service/client, '
```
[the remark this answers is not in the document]
dabacon: fixed

cirq/ionq/ionq_client.py
```diff
@@ -155,50 +152,51 @@ def _make_request(self, request: Callable[[], requests.Response]) -> requests.Re
...
-                be set if the target is `qpu`. If not specified and target is
-                `qpu`
-            target: If supplied the target to run on. Supports one of `qpu` or
-                `simulator`. If not set, uses `default_target`.
-            name: An optional name of the job. Different than the `job_id` of
-                the job.
+            serialized_program: The `cirq.ionq.SerializedProgram` containing the serialized
+                information about the circuit to run.
+            repetitions: The number of times to repeat the circuit. Only can be set if the target
+                is `qpu`. If not specified and target is `qpu`
```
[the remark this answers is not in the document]
dabacon: Yeah incomplete.  Completed with IonQ default, which is 100, I belive.

cirq/ionq/job.py
dabacon: That's a good idea.  The issue mostly is that really limits measurement key size plus number of qubits and this starts to get very close to cases that people might want to do with a larger qc.  Instead I will make multiple measurementX keys and then just a giant serialized string.

dabacon: I should also mention that doing it this way does add a restriction on the key, since it needs to be in some delimitated form.

cirq/ionq/serializer.py
dabacon: Serialized across potentially multiple keys, so made that check here.

cirq/ionq/job.py
```diff
@@ -154,15 +148,30 @@ def repetitions(self) -> Optional[int]:
...
         self._check_if_unsuccessful()
         if 'metadata' in self._job and 'shots' in self._job['metadata']:
             return int(self._job['metadata']['shots'])
         return None
 
+    def measurement_dict(self) -> Dict[str, Sequence[int]]:
+        """Returns a dictionary of measurement keys to target qubit index."""
+        measurement_dict: Dict[str, Sequence[int]] = {}
+        if 'metadata' in self._job:
+            full_str = ''.join(
```
gmauricio, line 163: Ah I see, I was suggesting a nested map, but I now see that we're restricting metadata to be key,value pairs of strings. 
This works then.

gmauricio (approved):
LGTM

cirq/ionq/job.py
dabacon: I went back and forth.  It increases complexity, but increases size of circuits that the API works for.

PR #3564, opened 2020-12-02
Measure using StabilizerStateChForm

smitsanghavi:
Going through the [paper](https://arxiv.org/abs/1808.00128), there **is** a way to simulate measurements without using the tableau. 

Previous discussion at https://github.com/quantumlib/Cirq/issues/2423#issuecomment-549584353.


smitsanghavi:
The `CliffordTableau` is still not being removed from the `CliffordSimulator` because `direct_fidelity_estimation` still relies on the availability of `stabilizers` from `CliffordState`. We can move that to use the tableau directly

tonybruguier:
Hi,
I opened:
https://github.com/quantumlib/Cirq/issues/3567

balopat (commented):
Looks simple enough - I do have to dig in myself and understand it before I can approve it. 
In the meantime, do you mind adding a test for this?

cirq/sim/clifford/stabilizer_state_ch_form.py
```diff
@@ -243,6 +243,19 @@ def to_state_vector(self) -> np.ndarray:
...
+        """Measures the q'th qubit.
+
+        Reference: Section 4.1 "Simulating measurements"
+
+        Returns: Computational basis measurement as 0 or 1.
+        """
+        w = self.s.copy()
+        for i, v_i in enumerate(self.v):
+            if v_i == 1:
+                w[i] = bool(prng.randint(2))
```
balopat, line 256: I am not convinced yet that this gives the same probability distribution as described in the paper. 
There we sample `w` once and then use it as a dot product with `G` to get the final bitstring `x`. Here, you are using the primitive _per qubit_...which means that per qubit you are using a different `w` value.

smitsanghavi: That's a valid concern. I was also worried that it might not work but it seems to work because the prescription in the paper can be generalized to per qubit (or any qubit subset). This is because we **always project to the measured value after measuring**. The `project_Z` and `update_sum` that it calls ensure that `v` and `s` get updated in a way that is consistent with the performed measurement. For e.g. any fully entangled qubits will get projected to a definite value post measurement. So, when we later perform this procedure for those other qubits, `v_i` will not be `1` for them since there is no degree of freedom in that qubit post-projection.

I will add tests that ensure that the measurements are not treated independently if the above explanation sounds reasonable to you.

One other thing I was considering is to have the `project_Z` be part of the `_measure` method itself. This will ensure that it's safe to call `_measure` and expect it to work as intended without needed to call another method. WDYT?

balopat: I just don't get the motivation why we need to do this instead of following the paper's convention? 
Could we not just have in `perform_measurement` something like this? 

```python
...
        w = self.s.copy()
        for i, v_i in enumerate(self.v):
            if v_i == 1:
                w[i] = bool(prng.randint(2))

        x = self.G @ w

        for i, qubit in enumerate(qubits):
            state.ch_form.project_Z(self.qubit_map[qubit], x[i])
            results.append(x[i])

        return results        
```

smitsanghavi: But that would be measuring all the qubits at once. The paper assumes that all measurements are terminal (which is a reasonable assumption for theoretical purpose since any Clifford circuit can be represented as such, with an equivalent density matrix). But that is not the design or the requirement for our use case. I don't think we ever enforce a terminal-measurement restriction for the `CliffordSimulator`. 

We iterate on all operations in the circuit, moment-by-moment, and perform measurement if we encounter a `MeasurementGate`. But the measurement is only performed on the subset of qubits for which this particular `MeasurementGate` is defined. It's not expected that we *measure* the rest, only *project*. The main loop in `CliffordSimulator`:

        state = CliffordState(qubit_map, initial_state=initial_state)

        for moment in circuit:
            measurements: Dict[str, List[np.ndarray]] = collections.defaultdict(list)

            for op in moment:
                    key = protocols.measurement_key(op)
                    measurements[key].extend(state.perform_measurement(op.qubits, self._prng))
                elif protocols.has_unitary(op):
                    state.apply_unitary(op)
                else:
                    raise NotImplementedError(f"Unrecognized operation: {op!r}")

Our simulator treats the measurements as sequential, so our underlying state also needs to support it. This shouldn't be a problem as long as we update everything properly after each measurement.

smitsanghavi:
> Looks simple enough - I do have to dig in myself and understand it before I can approve it.
> In the meantime, do you mind adding a test for this?

I have added a test (not really a unit test since it's pretty error prone to initialize the detailed internal state before measurement by hand). Let me know if it looks good. My plan after this PR:

- Update `MeasurementGate` to use this `_measure` method and update the test added here to use `act_on` instead of going through `CliffordSimulator`
- Resume #3503 but this time with `StabilizerStateChForm` instead of `CliffordTableau`
- Remove `CliffordTableau` from `CliffordSimulator`

cirq/sim/clifford/stabilizer_state_ch_form.py
balopat: > But that would be measuring all the qubits at once.

I don't think so, it's still just projecting to individual qubits enumerated from the `op.qubits`. The key difference is that the same `w` is used in case we do a multi-measurement gate... 

My biggest concern was using a different `w` for each qubit, but you are right in a sense that if the user has multiple measurement gates then it will become independent anyway, so we'll have to figure that partial measurement out. 

I'm not worried about the projection behavior. I think that looks good. 

I was a bit worried about whether probability distribution changes by applying uniformly random samples multiple times. I think I convinced myself that this is not a big issue - as sampling multiple times from the same distribution should follow the same distribution trivially :)

balopat (approved):

Issue #3341, opened 2020-09-18
`cirq.num_qubits` protocol on Gate classes


mpharrigan:
I would like to use the `cirq.num_qubits` protocol on gate classes

```python
cirq.num_qubits(cirq.X)
>>>1

cirq.num_qubits(cirq.XPowGate)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-50-ac8f0eb43a74> in <module>
----> 1 cirq.num_qubits(cirq.XPowGate)

~/cirq/cirq/cirq/protocols/qid_shape_protocol.py in num_qubits(val, default)
    158     """
    159     num_getter = getattr(val, '_num_qubits_', None)
--> 160     num_qubits = NotImplemented if num_getter is None else num_getter()
    161     if num_qubits is not NotImplemented:
    162         return num_qubits

TypeError: _num_qubits_() missing 1 required positional argument: 'self'

```

cc @cduck 

**Cirq version**
HEAD

mpharrigan:
Thing being Python, you can do the following underhanded thing:

```python
cirq.XPowGate._num_qubits_(None)
>>> 1
```

cduck:
What would the protocol do for gates like MeasurementGate that take a variable number of qubits to `__init__`?

If you decorate `_num_qubits_` with `@staticmethod` or `@classmethod` it might just work.

mpharrigan:
For variable qubit gates, then the `default` argument will come into play and by default default raise a TypeError.

I agree with your diagnosis that those decorators could do the trick

satw1knandala:
I can work on this if no one else is on it already. Let me know @balopat @cduck @mpharrigan

dstrain115:
This actually works now.  Closing this issue.

PR #3565, opened 2020-12-02
Enable `cirq.num_qubits` protocol with `Gate` classes

BillGatesNephew:
Enables usage of the `cirq.num_qubits` protocol with `Gate` classes per requested in #3341.


mpharrigan:
This PR would be extremely compelling if you add a test that would have failed without the fix!

mpharrigan:
In the linked issue, we were discussing using @staticmethod decorators. Is there a reason you didn't use that route?

BillGatesNephew:
@mpharrigan I added a test that checks a few different `Gate` classes with the updated functionality. Also, I initially thought about going that route, but I think it would've required changing the declaration of every `_num_qubits_`  from `_num_qubits_(self)` to `_num_qubits_(cls)`. I also wasn't sure how well the addition of `@staticmethod` would play with the decorators on the `_num_qubits_` function in `Gate`: 
```python
@value.alternative(requires='num_qubits', implementation=_num_qubits_proto_from_num_qubits)
@value.alternative(requires='_qid_shape_', implementation=_num_qubits_from_shape)
```
because both of the implementation parameters are instance methods. Because of this, I figured the route mentioned in the original issue would've ended up touching a lot of files. I just figured that it would be better to contain the solution to one function than make changes to every file where `_num_qubits_` was defined.

mpharrigan (changes requested):
Only certain classes know their number of qubits statically. For example: this code breaks

```python
cirq.num_qubits(cirq.MeasurementGate)
```

BillGatesNephew:
@mpharrigan I added a potential change to try calling the function if it is defined statically, and catch a raised exception to return `NotImplemented`. I also added additional assertions to check this case raises the appropriate exception. Generally, I try to avoid `try: ... except: ...` blocks, but to keep this change contained to a single file I added one. I'm not sure how you guys feel about them, and if they're frowned upon, then I can look into the potential `@staticmethod`solution.

mpharrigan:
There's nothing wrong with try/except blocks; but there's also nothing wrong with a change touching multiple files :) 

I think the staticmethod decorators would be more idiomatic. As a side benefit, you could call the method directly e.g. `cirq.XPowGate._num_qubits_()`

cirq/protocols/qid_shape_protocol.py
```diff
@@ -160,14 +162,27 @@ def num_qubits(
...
+    def call_getter(opt_getter_func):
+        if opt_getter_func is None:
+            return NotImplemented
+
+        if not isclass(val):
+            return opt_getter_func()
+
+        try:
+            return opt_getter_func(None)
+        except AttributeError:
```
viathor, line 175: I'm a little leery about using a try-except block here since `AttributeError` could be raised for reasons other than the one we have in mind here. It'd be safer to directly check the presence of the attribute using `hasattr`.

(This is similar to the EAFP idiom for checking key presence in a dict, so it's certainly idiomatic. It seems unnecessarily risky, though.)

mpharrigan:
@BillGatesNephew are you still interested in working on this PR?

mpharrigan:
closing as stale, feel free to re-open

PR #3568, opened 2020-12-03
Avoid triggering KeyError to check key presence in dicts

viathor:


HGSilveri (approved):
No issues on my side.

pschindler (approved):
No issues from my side

95-martin-orion (approved):

mpharrigan:
what motivated this change? As I understand it, idiomatic python likes try/excepts under the "EAFP" https://docs.python.org/3/glossary.html approach. Here's another thing I found: https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl

Note that I personally disagree with Python on this, so I'm fine with this change :) was just wondering if other corner cases motivated it or if it's just a style thing

95-martin-orion:
> what motivated this change?

I had an impression that this was idiomatic Python when writing #3546, but couldn't find the docs to back it up - AFAICT @viathor wrote this in response to the discussion there. (Coming from a C++ background, I'm also wary of the Python idiom here.)

viathor:
Yeah, I was being wrong [here](https://github.com/quantumlib/Cirq/pull/3546#discussion_r534532837). Sorry about that, @95-martin-orion. See my response there.

PR #3569, opened 2020-12-03
Polish noise docs

rmlarose:
Revamped as per discussions with @karlunho.

Let me know what you think! cc @balopat @lamberta as this may be time-sensitive


docs/noise.ipynb
```diff
@@ -62,196 +62,673 @@
...
-    "### `cirq.channel` and `def _channel_`\n",
-    "\n",
-    "To represent an operator sum evolution, a `Gate` should implement the\n",
-    "`SupportsChannel` protocol. To do this, the `Gate` should implement the\n",
-    "`_channel_(self) -> Sequence[np.ndarray]:` method. \n",
-    "\n",
-    "This method should return the sequence of `numpy` matrices corresponding to the Kraus operators. The basis in which this matrix is expressed is always implicit with respect to the object being called. \n",
-    "\n",
-    "For example, in `GateOperations`, these matrices must be ordered with respect to the list of qubits that the channel is applied to. The qubit-to-amplitude order mapping matches the ordering of `numpy.kron(A, B)`, where `A` is a qubit earlier in the list than the qubit `B`.\n",
+    "Cirq defines many commonly used quantum channels in [`ops/common_channels.py`](https://github.com/quantumlib/Cirq/blob/master/cirq/ops/common_channels.py). For example, we can get a single-qubit bit-flip channel defined as\n",
```
lamberta, line 130: Google dev docs style prefers "you" instead of "we": https://developers.google.com/style/highlights
But I would just remove it, if possible, or clarify who you are referring to. This could be "For example, a single-qubit bit-flip ..."

Here and throughout.

docs/noise.ipynb
```diff
@@ -62,196 +62,673 @@
...
+    "    print(f\"With probability {prob}, apply\\n\", kraus, end=\"\\n\\n\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "72486a155131"
+   },
+   "source": [
+    "> *Note:* Since `_mixture_` is defined, we can also use the `cirq.channel` protocol."
```
lamberta, line 471:
```suggestion
    "Note: Since `_mixture_` is defined, we can also use the `cirq.channel` protocol."
```

"Note:" is a style element used as a callout box on the webpage

lamberta (changes requested):
Thanks! A few nits

rmlarose:
Thanks @lamberta! Just addressed these.

docs/noise.ipynb
```diff
@@ -62,196 +62,673 @@
...
    "metadata": {
     "id": "7d3daa944d6b"
    },
    "source": [
-    "For simulation, it is useful to have `Gate` objects that enact noisy quantum evolution.\n",
-    "\n",
-    "Cirq supports modeling noise via *operator sum* representations of\n",
-    "noise (these evolutions are also known as quantum operations, quantum\n",
-    "dynamical maps, or superoperators).  \n",
+    "For simulation, it is useful to have `Gate` objects that enact noisy quantum evolution. Cirq supports modeling noise via *operator sum* representations of noise (these evolutions are also known as quantum operations, quantum dynamical maps, or superoperators). \n",
```
mpharrigan, line 98: as far as I understand it, the operator sum representation is *in contrast* to superoperator representation (?). If we're doing "also known as"-es then I'd mention Kraus operators (I see they're mentioned below)

docs/noise.ipynb
```diff
@@ -62,196 +62,673 @@
...
+    "This formalism models evolution of the density matrix $\\rho$ via\n",
     "\n",
-    "$$\\rho \\rightarrow \\sum_k A_k \\rho A_k^\\dagger$$\n",
+    "$$\n",
+    "\\rho \\rightarrow \\sum_{k = 1}^{m} A_k \\rho A_k^\\dagger\n",
+    "$$\n",
     "\n",
-    "Where A<sub>k</sub> are *Kraus* operators. These operators are not\n",
-    "necessarily unitary and must satisfy the trace-preserving property:\n",
+    "where $A_k$ are known as *Kraus operators*. These operators are not necessarily unitary and must satisfy the trace-preserving property\n",
```
mpharrigan, line 106: and -> but?

docs/noise.ipynb
```diff
@@ -62,196 +62,673 @@
...
     "\n",
-    "$$\\sum_k A_k^\\dagger A_k = I$$\n",
+    "$$\n",
+    "\\sum_k A_k^\\dagger A_k = I .\n",
+    "$$\n",
     "\n",
-    "As a noisy channel, Kraus operators are not unique. For more details of these operators, see [John Preskill's notes](http://www.theory.caltech.edu/people/preskill/ph219/chap3_15.pdf).\n"
+    "A channel with $m = 1$ unitary Kraus operator is called *coherent*, otherwise the channel is called *incoherent*. For a given noisy channel, Kraus operators are not necessarily unique. For more details on these operators, see [John Preskill's lecture notes](http://theory.caltech.edu/~preskill/ph219/chap3_15.pdf)."
```
mpharrigan, line 112: and is equivalent to a unitary gate operation?

docs/noise.ipynb
```diff
@@ -62,196 +62,673 @@
...
-    "dissipation to a surrounding environment as well as dephasing that\n",
-    "does not exchange energy. The amplitude damping channel only models dissipation of energy to a surrounding environment. \n",
+    "To summarize:\n",
+    "\n",
+    "* Every channel supports the `cirq.channel` protocol. (Note this includes gates like `cirq.X` and `cirq.CNOT`.)\n",
+    "  - If magic method `_channel_` is not defined, `cirq.channel` looks for `_mixture_` then for `_unitary_`.\n",
+    "* A subset of channels which support `cirq.channel` also support the `cirq.mixture` protocol.\n",
+    "  - If magic method `_mixture_` is not defined, `cirq.mixture` looks for `_unitary_`.\n",
+    "* A subset of channels which support `cirq.mixture` also support the `cirq.unitary` protocol.\n",
+    "  - If `_unitary_` is not defined for an object, the object is probably not meant for a circuit!"
```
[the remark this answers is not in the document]
obriente: Doubly confused now, because we use an amplitude_damping channel below in a circuit. So now I'm not really sure what this sentence means.

docs/noise.ipynb
```diff
@@ -62,196 +62,673 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "id": "55240879394e"
+   },
+   "outputs": [],
+   "source": [
+    "\"\"\"Minimal example of defining a custom channel.\"\"\"\n",
```
obriente, line 415: Would it be possible to include examples of what one should provide if one wanted to define _unitary_(self) or _channel_(self)?

docs/noise.ipynb
```diff
@@ -62,196 +62,673 @@
...
+    "Note that the density matrix simulator supports the `run` method which only gives access to measurements as well as the `simulate` method (used above) which gives access to the full density matrix."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "6f3caf9cea92"
+   },
+   "source": [
+    "Circuits in which all channels support the `cirq.mixture` protocol can also be simulated with the `cirq.Simulator`. When simulating such a channel, a single Kraus operator is randomly sampled from the mixture (according to the probability distribution) and applied to the wavefunction. This method is known as \"Monte Carlo (wavefunction) simulation\" or \"quantum trajectories.\""
```
balopat, line 560: We do support monte carlo for arbitrary channels not just mixtures: https://github.com/quantumlib/Cirq/pull/3043/files

docs/noise.ipynb
```diff
@@ -62,196 +62,673 @@
...
+    "    cirq.qft(*qreg),\n",
+    "    bit_phase_flip.on_each(qreg[1::2]),\n",
+    "    cirq.qft(*qreg, inverse=True),\n",
+    "    cirq.reset(qreg[1]),\n",
+    "    cirq.measure(*qreg),\n",
+    "    cirq.bit_flip(p=0.07).controlled(1).on(*qreg[2:]),\n",
+    ")\n",
+    "\n",
+    "print(\"Circuit with multiple channels:\\n\")\n",
+    "print(circ)"
```
balopat, line 731: one more method could be mentioned here: `Gate.with_probability(p)` creates a probabilistic mixture out of a gate.

balopat (approved):
This is looking really good - I added two more nits

docs/noise.ipynb
rmlarose: This was a remark from the original docs I was also confused by but left in. In the context of quantum channels I think the answer to your question is yes - operator sum representations and superoperators are treated as two separate concepts. 

I guess purely mathematically, Kraus operators act on density matrices (operators) and so are technically superoperators - maybe this was the original meaning? I find the operator, superoperator, super-duper-operator hierarchy unnecessarily confusing.

Anyhow I think the remark is vague and should be clarified, but will heed to better judgement.

docs/noise.ipynb
```diff
@@ -62,196 +62,673 @@
...
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "99ac77a2c896"
+   },
+   "source": [
+    "Other than these general methods, channels can be added to circuits at any moment just as gates are. The channels can be different, be correlated, act on a subset of qubits, be custom defined, etc."
```
obriente, line 706: If I remember rightly, it should be possible to generate a 'noisy_moment' function that takes in a moment and adds noise to the entire chip. This could be in principle useful (e.g. if one wants to simulate crosstalk by adding unwanted rotations to nearby gates). Can we include mention + perhaps an example of this?

obriente (commented):
Looks good on the whole, but it would be great if we could get more examples of how to build a noisy channel, currently I cant see this. But, very much like that we mention how to extract Kraus operators up-front - I didn't know this was possible.

Cheers,

Tom.

docs/noise.ipynb
rmlarose: Yeah I think that sounds better. I guess it was "and" originally because they are two independent criteria, but "but" sounds better with the qualifier "not necessarily unitary." English is hard.

docs/noise.ipynb
rmlarose: Ahah! To check: Monte Carlo with only channels (not mixtures) will be slower than Monte Carlo with all mixtures because the probability to apply a Kraus depends on the state - is this an accurate remark to include?

rmlarose: I didn't know they were both supported though, that's really cool

docs/noise.ipynb
```diff
@@ -62,196 +62,673 @@
...
    "source": [
-    "### `cirq.mixture` and `def _mixture_`\n",
-    "\n",
-    "Some channels can be interpreted as probabilistically selecting between\n",
-    "different unitary evolutions:\n",
-    "\n",
-    "$$\\rho \\rightarrow \\sum_k p_k U_k \\rho U_k^\\dagger {\\rm ~where~} \\sum_k p_k =1 {\\rm ~and~ U_k U_k^\\dagger= I}$$\n",
-    "\n",
-    "In this case, it is possible to perform **Monte Carlo simulations** of these gates using a wave function based simulator (and not a density matrix based simulator). An example of this can be seen below: \n"
+    "To see the Kraus operators of a channel, the `cirq.channel` protocol can be used."
```
balopat, line 157: we could also link to the Protocols page from somewhere - maybe here or in a new sentence for users who don't understand what a protocol is.

docs/noise.ipynb
balopat: Yes your understanding is correct!

docs/noise.ipynb
rmlarose: Added.

docs/noise.ipynb
rmlarose: Nice! Added a note on this but eariler in the tutorial - after the `BitAndPhaseFlip` custom definition as an alternative way to define this.

docs/noise.ipynb
rmlarose: Sounds good. Is there a standard protocol (ha ha ha) to link to docs from docs? Should it be a devsite path like /cirq/protocols or link to GitHub/elsewhere?

docs/noise.ipynb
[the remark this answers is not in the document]
rmlarose: The inclusions should be reversed - something in a circuit must have a channel, not necessarily a mixture or a unitary, though. Some channels have mixtures (e.g. cirq.bit_flip), and some channels have unitaries (e.g. cirq.X). 

So, {channels with unitaries} ⊂ {channels with mixtures} ⊂ {channels}.

Thanks for pointing this out, I'm sure others who read it will have the same confusion. I'll add an example to hopefully clarify.

docs/noise.ipynb
lamberta: If you use the full API symbol in backticks it will autolink to the reference page on the site.

For narrative docs, if in the same project/repo you can relative link to the file (w/ extension), like `[foo guide](./foo_guide.ipynb)`.

But if linking to a doc in another project/repo, need the absolute site path. This keeps the links working in GitHub and Colab.

docs/noise.ipynb
rmlarose: Great point. Added a note on defining `_channel_` - let me know what you think. The case for defining `_unitary_` is documented in [custom gates](https://github.com/quantumlib/Cirq/blob/master/docs/custom_gates.ipynb).

docs/noise.ipynb
rmlarose: Do you mean the `Circuit.with_noise` method (example in previous cell) or something else?

docs/noise.ipynb
rmlarose: For now I removed the reference to superoperators.

balopat:
Given the time pressure on this, I'm going to go ahead and merge - @obriente - if you have follow-up ideas those will be super welcome as an issue and we'll fix it up soon.

docs/noise.ipynb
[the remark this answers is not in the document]
obriente: Alright, perhaps remove the last line here though?  "If `_unitary_` is not defined for an object, the object is probably not meant for a circuit!" sounds like I can't add objects that aren't unitary to a circuit, which is confusing.

docs/noise.ipynb
obriente: Looks good, thanks

docs/noise.ipynb
obriente: I mean the NoiseModel.noisy_moment and NoiseModel.noisy_moments methods. These can be quite useful in some cases (e.g. the example I gave).

obriente:
I left some comments in the review, is that sufficient or should I start
another issue?

Cheers,

Tom.

rmlarose:
Thanks @obriente! No need for another issue, I will open a follow-up PR to address all remaining comments here and ping you to take another look.

PR #3570, opened 2020-12-03
Add get current calibration to IonQ service

dabacon:
* Adds get current calibration and creates not `ionq.Calibration` object for result.
* Future PR will add list calibrations.

Note: requires google approval as I fixed documentation in an Engine method. @wcourtney


cirq/ionq/ionq_client_test.py
```diff
@@ -427,3 +427,68 @@ def test_ionq_client_delete_job_retry(mock_put):
...
+        'http://example.com/v0.1/calibrations/current', headers=expected_headers
+    )
+
+
+@mock.patch('requests.get')
+def test_ionq_client_get_current_calibration_unauthorized(mock_get):
+    mock_get.return_value.ok = False
+    mock_get.return_value.status_code = requests.codes.unauthorized
+
+    client = ionq.ionq_client._IonQClient(
```
nakardo, line 451: Does `default_target` argument affect the outcome of the test or can be omitted?

nakardo (approved):
LGTM. I left a question.

cirq/ionq/calibration.py
```diff
@@ -0,0 +1,61 @@
...
+        return self._calibration_dict['target']
+
+    def calibration_time(self, tz: Optional[datetime.tzinfo] = None) -> datetime.datetime:
+        """Return a python datetime object for the calibration time.
+
+        Args:
+            tz: The timezone for the string. If None, the method uses the platform's local timezone.
+
+        Returns:
+            A `datetime` class with the time.
```
wcourtney, line 40: Nit: it returns an object, not a class.

cirq/ionq/calibration.py
```diff
@@ -0,0 +1,61 @@
...
+        """
+        # Python datetime only like microseconds, not milliseconds, and does not like 'Z'.
+        first, second = self._calibration_dict['date'].split('.')
+        modified_date = f'{first}.{second[:3]}'
+        dt = datetime.datetime.strptime(modified_date, '%Y-%m-%dT%H:%M:%S.%f')
+        return dt.replace(tzinfo=datetime.timezone.utc).astimezone(tz=tz)
+
+    def fidelities(self) -> dict:
+        """"Returns the metrics (fidelities)."""
```
wcourtney, line 49: Extra "

cirq/ionq/calibration.py
```diff
@@ -0,0 +1,61 @@
...
+    def timings(self) -> dict:
+        """Returns the gate, measurement, and reseting timings."""
+        return self._calibration_dict['timing']
+
+    def connectivity(self) -> Tuple[Tuple[int, int], ...]:
+        """Returns which qubits and can interact with which.
+
+        Note that this only returns one of (q0, q1) and (q1, q0) if q0 and q1 can interact.
+        """
+        return tuple((int(x), int(y)) for x, y in self._calibration_dict['connectivity'])
```
wcourtney, line 61: Optional: consider using a set to more clearly represent the unordered nature of the tuple.

cirq/ionq/calibration_test.py
```diff
@@ -0,0 +1,66 @@
...
+        'timing': {
+            '1q': 1.1e-05,
+            '2q': 0.00021,
+            'readout': 0.000175,
+            'reset': 3.5e-05,
+        },
+        'date': '2020-08-07T12:47:22.337Z',
+    }
+    cal = ionq.Calibration(calibration_dict=calibration_dict)
+    assert cal.connectivity() == ((0, 1), (0, 2), (1, 2))
```
wcourtney, line 38: >  Note that this only returns one of (q0, q1) and (q1, q0) if q0 and q1 can interact.

Consider adding a test for this since it was asserted in the documentation.

wcourtney (approved):
LGTM for the google/engine updates. Thanks again for the minor doc cleanup!

cirq/ionq/calibration_test.py
dabacon: I think this is a contract of the api's returned values, so testing it just means mocking this property and then asserting it.  I don't think this really tests anything, does it?

cirq/ionq/calibration.py
dabacon: Even better made it a set of all tuples with both orders. This is likely the more useful representation.

cirq/ionq/ionq_client_test.py
dabacon: Yep, good catch.  Removed.  Though I will ask you later how there can be calibrations for different targets in the long run :)

dabacon (commented):
Address comments.

dabacon:
Part of #3479

Issue #3567, opened 2020-12-02
Remove CliffordTableau


tonybruguier:
**Is your feature request related to a use case or problem? Please describe.**
The CliffordTableau is still not being removed from the CliffordSimulator because direct_fidelity_estimation still relies on the availability of stabilizers from CliffordState. 

**Describe the solution you'd like**
We can move to use the tableau directly

**[optional] Describe alternatives/workarounds you've considered**

**[optional] Additional context (e.g. screenshots)**
https://github.com/quantumlib/Cirq/pull/3564

**What is the urgency from your perspective for this issue? Is it blocking important work?**
P1

tonybruguier:
Happy to take this on if you want me to.

smitsanghavi:
Thanks for creating the issue, Tony. I can take it up. 

By a quick glance at the `direct_fidelity_estimation` implementation (for Cliffords), it seems to just be iterating over the circuit and using `CliffordSimulator`'s`apply_unitary` on all the ops. After that it just needs the `stabilizers` from the final state to proceed. It seems easy to just switch to using `CliffordTableau` instead of `CliffordState` and `act_on` protocol instead of `apply_unitary`. It will also add the ability to handle any measurements the circuit might have. Let me know if that sounds good or if I'm missing some caveat.

tonybruguier:
I sent you:
https://github.com/quantumlib/Cirq/pull/3571

PR #3571, opened 2020-12-03
Directly use Clifford tableaux for direct fidelity estimation

tonybruguier:
For issue #3567


examples/direct_fidelity_estimation.py
```diff
@@ -173,19 +175,19 @@ class PauliTrace:
 
 
 def _estimate_pauli_traces_clifford(
-    n_qubits: int, clifford_state: cirq.CliffordState, n_measured_operators: Optional[int]
+    n_qubits: int, clifford_tableau: cirq.CliffordTableau, n_measured_operators: Optional[int]
```
smitsanghavi, line 178: Do we need to pass in the entire tableau? The implementations seems to only need the stabilizers. Can we just pass that so that this method remains agnostic to the representation used to get the stabilizers?

examples/direct_fidelity_estimation.py
```diff
@@ -173,19 +175,19 @@ class PauliTrace:
...
     and all the other traces have probability 0. In addition, there is a fast
     way to compute find out what the traces are. See the documentation of
-    cirq.CliffordState for more detail. This function uses the speedup to sample
+    cirq.CliffordTableau for more detail. This function uses the speedup to sample
     the Pauli states with non-zero probability.
 
     Args:
         n_qubits: An integer that is the number of qubits.
-        clifford_state: The basis of the Pauli states with non-zero probability.
+        clifford_tableau: The basis of the Pauli states with non-zero probability.
```
smitsanghavi, line 190: This description does not seem accurate for CliffordTableau

examples/direct_fidelity_estimation.py
```diff
@@ -379,21 +381,25 @@ def direct_fidelity_estimation(
     n_qubits = len(qubits)
 
     clifford_circuit = True
-    clifford_state: Optional[cirq.CliffordState] = None
+    clifford_tableau: Optional[cirq.CliffordTableau] = None
```
smitsanghavi, line 384: The tableau can be instantiated here directly, removing the need to wrap it in Optional. Any reason to instantiate with `None` first?

examples/direct_fidelity_estimation.py
```diff
@@ -379,21 +381,25 @@ def direct_fidelity_estimation(
...
     try:
-        clifford_state = cirq.CliffordState(qubit_map={qubits[i]: i for i in range(len(qubits))})
+        qubit_map = {qubits[i]: i for i in range(n_qubits)}
+        clifford_tableau = cirq.CliffordTableau(n_qubits)
```
smitsanghavi, line 387: `qubit_map` also does not need to be inside the `try` block. If there is a `TypeError` here, it should loudly fail.

examples/direct_fidelity_estimation_test.py
```diff
@@ -2,6 +2,8 @@
 import pytest
 import cirq
 import cirq.google as cg
+from cirq.protocols import act_on
```
smitsanghavi, line 5: You should be able to use `cirq.act_on` directly. Is that causing circular dependencies?

examples/direct_fidelity_estimation_test.py
```diff
@@ -108,13 +110,17 @@ def test_same_pauli_traces_clifford():
...
 
-    clifford_state = cirq.CliffordState(qubit_map={qubits[i]: i for i in range(len(qubits))})
+    qubit_map = {qubits[i]: i for i in range(n_qubits)}
+    clifford_tableau = cirq.CliffordTableau(n_qubits)
     for gate in circuit.all_operations():
-        clifford_state.apply_unitary(gate)
+        tableau_args = clifford.ActOnCliffordTableauArgs(
+            clifford_tableau, [qubit_map[i] for i in gate.qubits], np.random.RandomState(), {}
+        )
+        act_on(gate, tableau_args)
```
smitsanghavi, line 119: The above lines in the test will always need to be kept in sync with the implementation exactly. This is usually a bad sign. Can we somehow test the equivalency of the two`_estimate_pauli_traces_` methods without duplicating the implementation here?

Maybe stick a no-op non-clifford gate at the start for the `general` one so that it doesn't pick up the clifford path? Not sure if such changes would affect this particular case. Let me know what you think.

smitsanghavi (changes requested):

smitsanghavi:
Thanks for making the changes!

examples/direct_fidelity_estimation.py
tonybruguier: Done.

examples/direct_fidelity_estimation.py
tonybruguier: Indeed, but maybe with the change you requested above, the comment is actually accurate now (the variable is stabilizer_basis).

examples/direct_fidelity_estimation.py
tonybruguier: Done.

examples/direct_fidelity_estimation.py
tonybruguier: Done.

examples/direct_fidelity_estimation_test.py
tonybruguier: No circular dependencies, just my incompetence. Fixed.

examples/direct_fidelity_estimation_test.py
tonybruguier: Yes, I can do that. However, I am not sure how to create a no-op gate that's not Clifford. Would you have some code pointers and then I'll do it.

tonybruguier:
Thanks for the quick turn-around. I tried to address the comments, but some are left unresolved, with some questions on my part. I do plan to resolve them eventually of course.

PTAL.

95-martin-orion:
Failing test is due to a state change in our `pandas` requirement - you can sync past #3572 to resolve the issue.

examples/direct_fidelity_estimation_test.py
smitsanghavi: A `CCX` (Toffoli) gate at the beginning of the second circuit would be a no-op if the starting state is |000> (default). But I'm not very familiar with the direct fidelity estimation to know if it is sensitive to the initial state or tries to be agnostic to the initial state (if it's the latter it will not be a no-op).

examples/direct_fidelity_estimation_test.py
smitsanghavi: Don't worry about it! It's valid either way. Just trying to follow the style guide at https://cirq.readthedocs.io/en/stable/docs/dev/style.html#import-statements

It's still isn't showing up as updated for me.

examples/direct_fidelity_estimation.py
```diff
@@ -218,7 +215,7 @@ def _estimate_pauli_traces_clifford(
     pauli_traces: List[PauliTrace] = []
     for dense_pauli_string in dense_pauli_strings:
         # The code below is equivalent to calling
-        # clifford_state.state_vector() and then calling
+        # clifford_tableau.state_vector() and then calling
```
smitsanghavi, line 218: Just realized that CliffordTableau doesn't have a state_vector() method (yet). This will need some rephrasing.

smitsanghavi (changes requested):
Looks good overall! Just some minor nits.

examples/direct_fidelity_estimation.py
smitsanghavi: Looks good 👍

tonybruguier: Ack.

examples/direct_fidelity_estimation_test.py
tonybruguier: Done. Thanks for the tip.

examples/direct_fidelity_estimation.py
tonybruguier: Done.

tonybruguier:
Thanks. PTAL.

smitsanghavi (approved):

smitsanghavi:
LGTM! But you'll need approvals from one of the maintainers to submit.

examples/direct_fidelity_estimation.py
```diff
@@ -379,21 +376,29 @@ def direct_fidelity_estimation(
     n_qubits = len(qubits)
 
     clifford_circuit = True
-    clifford_state: Optional[cirq.CliffordState] = None
+    qubit_map = {qubits[i]: i for i in range(n_qubits)}
+    clifford_tableau: cirq.CliffordTableau = cirq.CliffordTableau(n_qubits)
```
mpharrigan, line 380: unnecessary type annotation?

tonybruguier: Removed

tonybruguier:
Thanks. PTAL.

mpharrigan (approved):
Thanks for reviewing @smitsanghavi 

LGTM as well

PR #3572, opened 2020-12-03
Fix pandas dependency

95-martin-orion:
Error highlighted by @akushnarov: `pandas requires NumPy >= 1.6.1 due to datetime64 dependency`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ dataclasses; python_version < '3.7'
 google-api-core[grpc] >= 1.14.0, < 2.0.0dev
 matplotlib~=3.0
 networkx~=2.4
-numpy~=1.16.1
+numpy~=1.17
 pandas
 protobuf~=3.13.0
 requests~=2.18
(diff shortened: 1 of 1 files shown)


95-martin-orion:
`numpy.random.default_rng` was added in v1.17, so that's our actual minimum version. The reason that `numpy~=1.16` still passed our test suite is that `~=` gets the latest compatible version (i.e. v1.*); however, it only guarantees that this version will be at least v1.16.

95-martin-orion:
Okay, this is getting strange: `numpy~=1.17` still gives the `pandas requires NumPy >= 1.6.1` error.

I found [this SO question](https://stackoverflow.com/questions/12436979/how-to-fix-python-numpy-pandas-installation) which suggests that MacOS has a default numpy version that may be causing this. Recommended fix is `pip install numpy --upgrade` or manually deleting the default `numpy` install.

I can only assume that this didn't appear until now because of some change to `pandas` which increased its minimum numpy requirement.

95-martin-orion:
Based on the explosion of package-version downloads in the logs and an apparent change from pip-20.2.4 to 20.3, I suspect this is actually caused by https://github.com/pypa/pip/issues/9215. Closing in lieu of a more appropriate fix.

95-martin-orion:
> Closing in lieu of a more appropriate fix.

See #3574.

mpharrigan:
Can we delete this branch?

PR #3576, opened 2020-12-03
Fix formatter on deleted or moved files

### LINKED ISSUE #3573: Formatter script fails on moved file
**Description of the issue**

After moving or deleting a python file, the formatter script fails

```
 check/format-incremental --apply                                                                                                                                                    
```

```
Comparing against revision 'upstream/master'.
Usage: black [OPTIONS] [SRC]...
Try 'black -h' for help.

Error: Invalid value for '[SRC]...': Path 'cirq/google/optimizers/two_qubit_gates/__init__.py' does not exist.
```

1. Delete a .py file 
2. run check/format-incremental 

**Cirq version**

master

balopat:
Fix formatter on deleted or moved files. 

Fixes #3573.

### DIFF
--- a/check/format-incremental
+++ b/check/format-incremental
@@ -77,8 +77,17 @@ if (( only_changed == 1 )); then
         rev="${base}"
     fi
 
-    # Get the _test version of changed python files.
-    format_files=$(git diff --name-only ${rev} -- | grep "\.py$" | grep -v "_pb2\.py$")
+    # Get the modified and added python files. Lines in git diff --name-status look like this:
+    #M       cirq/__init__.py
+    #A       cirq/added.py
+    modified_files=$(git diff --name-status ${rev} -- | grep '\.py$' | grep -v '_pb2\.py$' | grep '^[MA].*$' | awk '{print $2}')
+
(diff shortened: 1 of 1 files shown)


check/format-incremental
```diff
@@ -77,8 +77,17 @@ if (( only_changed == 1 )); then
         rev="${base}"
     fi
 
-    # Get the _test version of changed python files.
-    format_files=$(git diff --name-only ${rev} -- | grep "\.py$" | grep -v "_pb2\.py$")
+    # Get the modified and added python files. Lines in git diff --name-status look like this:
+    #M       cirq/__init__.py
+    #A       cirq/added.py
```
maffoo, line 82: nit: add space between `#` and `M` or `A`.

maffoo (approved):
LGTM

I would suggest that this script is complicated enough that it might be better to rewrite it in python. But I'm biased because my bashfu is weak.

balopat:
+1 on moving to python for complicated scripts: https://github.com/quantumlib/Cirq/issues/3577

mpharrigan:
my rule is if you ever start writing a for loop in bash, stop what you're doing immediately and convert to python :)

PR #3580, opened 2020-12-03
CircuitOperation

95-martin-orion:
Rewrite of #3515 to remove CircuitGates, plus a few other cleanup items. Part of #3235.

Note again the following features are absent from this PR:
* Measurement in loops. The current measurement format (a 1:1 dict of keys to results) does not play well with repeating the same measurement multiple times, or with attempting to rename that measurement based on loop index.
* Condensed serialization. This feature requires changes outside the CircuitGate itself to allow for references inside the serialization structure.
* Condensed circuit diagrams. Similar to serialization, this requires changes to the diagram behavior for printing CircuitOperations separate from the rest of the circuit.


95-martin-orion:
Test failure appears to be a flake; filed #3584 to track. This PR is ready for review.

95-martin-orion:
Caught an issue in this PR while looking into serialization: the `name` field of FrozenCircuit is not preserved across serializations. The tests miss this because `AbstractCircuit.__eq__` does not compare the `name` field. Working on a fix.

95-martin-orion:
Opted to remove the `name` field for now, as it's kind of a non sequitur in this PR. Added a TODO to add it back in the condensed-serialization change.

cirq/circuits/circuit.py
```diff
@@ -762,12 +757,42 @@ def are_all_matches_terminal(self, predicate: Callable[['cirq.Operation'], bool]
...
+                return True
+            if isinstance(op, ops.TaggedOperation):
+                return is_circuit_op(op.sub_operation)
+            return False
+
+        if not all(
             self.next_moment_operating_on(op.qubits, i + 1) is None
             for (i, op) in self.findall_operations(predicate)
-        )
+            # TODO: this misbehaves with tagged CircuitOperations.
```
balopat, line 776: add issue for todo

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,333 @@
...
+                '.with_measurement_key_mapping(' + f'{dict_repr(self.measurement_key_map)})'
+            )
+        if self.param_resolver != base_op.param_resolver:
+            base_repr += f'.with_params({proper_repr(self.param_resolver)})'
+        if self.repetitions != 1:
+            base_repr += f'.repeat({self.repetitions})'
+        return base_repr
+
+    def __str__(self):
+        # TODO: support out-of-line subcircuit definition in string format.
```
balopat, line 181: Add Github issue for TODO

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,333 @@
...
+
+    def _with_measurement_key_mapping_(self, key_map: Dict[str, str]) -> 'CircuitOperation':
+        return self.with_measurement_key_mapping(key_map)
+
+    def with_params(
+        self, param_values: study.ParamResolverOrSimilarType, recursive: bool = False
+    ) -> 'CircuitOperation':
+        """Returns a copy of this operation with updated param values.
+
+        The provided param_values are composed with the existing values, so
```
balopat, line 319: nit: format docsstring for public method: Args are missing, Return is missing

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,333 @@
...
+        return new_op
+
+    def with_qubits(self, *new_qubits: 'cirq.Qid'):
+        if len(new_qubits) != protocols.num_qubits(self.circuit):
+            expected = protocols.num_qubits(self.circuit)
+            raise ValueError(f'Expected {expected} qubits, got {len(new_qubits)}.')
+        return self.with_qubit_mapping(dict(zip(self.qubits, new_qubits)))
+
+    def with_measurement_key_mapping(self, key_map: Dict[str, str]) -> 'CircuitOperation':
+        """Returns a copy of this operation with an updated key mapping.
```
balopat, line 299: nit: format docsstring for public method: Args are missing, Return is missing

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,333 @@
...
+        return new_op
+
+    def repeat(
+        self,
+        repetitions: int,
+        modulus: int = 0,
+        allow_invert: bool = False,
+        validate_modulus: bool = False,
+    ) -> 'CircuitOperation':
+        """Returns a copy of this operation repeated 'repetitions' times.
```
balopat, line 253: nit: format docsstring for public method: Args are missing, Return is missing

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,333 @@
...
+        if repetitions != 1:
+            return result.repeat(repetitions)
+        return result
+
+    # Methods for constructing a similar object with one field modified.
+
+    def with_circuit(self, new_circuit: 'cirq.AbstractCircuit') -> 'CircuitOperation':
+        """Returns a copy of this operation with the provided circuit.
+
+        Key mappings, parameter values, and repetitions are preserved.
```
balopat, line 238: nit: format docsstring for public method: Args are missing, Return is missing

balopat (commented):
This is so exciting, we are very close! I will have to play with this tomorrow, but it's looking really good from a quick skim!!!

cirq/circuits/circuit_operation.py
95-martin-orion: Opened #3605.

cirq/circuits/circuit.py
95-martin-orion: This TODO is done - removed.

cirq/circuits/circuit_operation.py
95-martin-orion: Done.

cirq/circuits/circuit_operation.py
95-martin-orion: Done.

cirq/circuits/circuit_operation.py
95-martin-orion: Done.

cirq/circuits/circuit_operation.py
95-martin-orion: Done.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,384 @@
...
+                raise ValueError('Raising the circuit to "modulus" must produce the identity.')
+            new_op._repetitions %= modulus
+            if allow_invert and new_op._repetitions > modulus // 2:
+                new_op._repetitions -= modulus
+        return new_op
+
+    def __pow__(self, power: int) -> 'CircuitOperation':
+        return self.repeat(power)
+
+    def with_qubit_mapping(self, qubit_map: Dict['cirq.Qid', 'cirq.Qid']) -> 'CircuitOperation':
```
maffoo, line 306: `Circuit` has a similar method `transform_qubits` which takes a `Callable[[Qid], Qid]` instead of a dict. I think it'd be nice to harmonize these APIs. Ideally `transform_qubits` would accept `Mapping[Qid, Qid]` in addition to callable, since many callsite wrap a lambda around a dict anyway. Could also accept both callable and dict here. It'd be great if we could name the functions similarly as well; I like `with` since it produces a new instance, so maybe we could add a `with_qubit_mapping` on circuit too?

cirq/circuits/circuit_operation_test.py
```diff
@@ -0,0 +1,508 @@
...
+def test_properties():
+    a, b, c = cirq.LineQubit.range(3)
+    circuit = cirq.FrozenCircuit(
+        cirq.X(a),
+        cirq.Y(b),
+        cirq.H(c),
+        cirq.CX(a, b) ** sympy.Symbol('exp'),
+        cirq.measure(a, b, c, key='m'),
+    )
+    op = cirq.CircuitOperation(circuit)
```
maffoo, line 29: Maybe add a `to_op` method on frozen circuit so can do `op = circuit.freeze().to_op()`

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,384 @@
...
+            base_repr += f'.with_qubit_mapping({dict_repr(self.qubit_map)})'
+        if self.measurement_key_map != base_op.measurement_key_map:
+            base_repr += (
+                '.with_measurement_key_mapping(' + f'{dict_repr(self.measurement_key_map)})'
+            )
+        if self.param_resolver != base_op.param_resolver:
+            base_repr += f'.with_params({proper_repr(self.param_resolver)})'
+        if self.repetitions != 1:
+            base_repr += f'.repeat({self.repetitions})'
+        return base_repr
```
maffoo, line 178: This repr with a bunch of method calls seems a bit odd to me. What if we change the constructor to accept optional args for qubit_mapping, params, etc. and then can have one constructor call for the repr `cirq.CircuitOperation(..., qubit_mapping={...}, ...)`.

I think this might also simplify other code where were modify various mappings to just construct new instances instead of copy + mutate.

maffoo (commented):
This is looking good!

I'd suggest that the CircuitOperation constructor could accept qubit_mapping, params, etc as keyword args (could even implement as a dataclass so you wouldn't have to write out all the boilerplate). Similarly, the copy method can take those things as optional args and construct a new instance with given attributes overridden. I think this would clean up some code that currently copies and then mutates instances.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,384 @@
...
+                protocols.unitary(self.circuit * modulus), np.eye(np.prod(self.circuit.qid_shape()))
+            ):
+                raise ValueError('Raising the circuit to "modulus" must produce the identity.')
+            new_op._repetitions %= modulus
+            if allow_invert and new_op._repetitions > modulus // 2:
+                new_op._repetitions -= modulus
+        return new_op
+
+    def __pow__(self, power: int) -> 'CircuitOperation':
+        return self.repeat(power)
```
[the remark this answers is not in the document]
95-martin-orion: I don't think we can support consistent behavior for this in the long term due to other planned features. One such feature is a "loop iterator" param that assumes user-specified values for each repetition; there isn't a clear way to handle a non-integer power in such cases.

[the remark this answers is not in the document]
maffoo: To be clear, I'm not suggesting non-integer powers in general, only if you want to raise an already-iterated circuit to a non-integer power such that the end result is still an integer number of repetitions. This seems like it would still be compatible with loop iterators as you suggest, but I don't know exactly what you have in mind.

cirq/circuits/circuit_operation_test.py
95-martin-orion: I'd prefer to keep the `CircuitOperation` creation separate for a couple reasons:

1. `FrozenCircuits` are meant to be reusable to improve serialization; `circuit.freeze().to_op()` lends to a 1:1 relation between `FrozenCircuit` and `CircuitOperation`
2. Creating `CircuitOperation` in `FrozenCircuit` adds import-ordering confusion, which is already a little weird due to defining `CircuitOperation` in a hierarchy that prevents `ops` from importing `circuits`.

cirq/circuits/circuit_operation.py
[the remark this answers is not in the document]
95-martin-orion: In a pre-PR version of this change, loop iterators looked something like this:
```python
def repeat(self, {...}, loop_param: sympy.Symbol, loop_values: List[float]):
  # when decomposed, "loop_param" will be assigned each value in "loop_values" in turn
```
This would cause issues (or at least have unclear behavior) with something like:
```python
circuit4 = circuit.repeat(4, loop_param=sympy.Symbol('x'), loop_values: [1, 3, 6, 10])
circuit2 = circuit4.repeat(0.5)  # what values should "loop_param" be assigned?
```

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,384 @@
...
+            .with_measurement_key_mapping(measurement_key_map)
+            .with_params(param_resolver)
+        )
+        if repetitions != 1:
+            return result.repeat(repetitions)
+        return result
+
+    # Methods for constructing a similar object with one field modified.
+
+    def with_circuit(self, new_circuit: 'cirq.AbstractCircuit') -> 'CircuitOperation':
```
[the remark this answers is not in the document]
95-martin-orion: This was intended as an analogue to `GateOperation.with_gate`, but on further inspection that method is used exclusively in other `GateOperation` methods for which the analogy breaks down - e.g. `_resolve_parameters_` in `GateOperation` uses `with_gate`, but since `CircuitOperation` defers parameter resolution it doesn't need to call `with_circuit`.

I'd be happy to remove this if you don't see any other use case for it.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,382 @@
...
+            args.append(f'qubit_map={dict_str(self.qubit_map)}')
+        if self.measurement_key_map != base_op.measurement_key_map:
+            args.append(f'key_map={dict_str(self.measurement_key_map)}')
+        if self.param_resolver != base_op.param_resolver:
+            args.append(f'params={self.param_resolver.param_dict}')
+        if self.repetitions != 1:
+            args.append(f'loops={self.repetitions}')
+        if not args:
+            return f'{header}\n{circuit_msg}'
+        return f'{header}\n{circuit_msg}({", ".join(args)})'
```
balopat, line 203: Another diagram nit:
It's great to have the qubit map there. However, we could make it a bit cleaner.

Code:

```
import cirq
from cirq import *

a, b, c, d = GridQubit.rect(2,2)
q = [NamedQubit(i) for i in ["a","b","c"]]
op = CircuitOperation(FrozenCircuit(cirq.X(q[0]), cirq.Y(q[1]), cirq.Z(q[2]), cirq.CX(q[0], q[1]))).with_tags(["tag"])
print(Circuit(X(a), Y(b), Z(c), CX(a, b), H(c), op.with_qubits(a,b,d), [H(i) for i in [a,b,c,d]]))
```

Current output:

```
                   ┌──────────────────────────────────────────────────────────────────────────────────┐
                    Circuit_0x740eb7df41d7eca6:
                    [ a: ───X───@───          ]
(0, 0): ───X───@────[           │             ]───────────────────────────────────────────────────────────H───
               │    [ b: ───Y───X───          ]
               │    [                         ]
               │    [ c: ───Z───────          ](qubit_map={a: (0, 0), b: (0, 1), c: (1, 1)})[['tag']]
               │    │
(0, 1): ───Y───X────#2────────────────────────────────────────────────────────────────────────────────────H───
                    │
(1, 0): ───Z───H────┼────────────────────────────────────────────────────────────────────────────────H────────
                    │
(1, 1): ────────────#3────────────────────────────────────────────────────────────────────────────────────H───
                   └──────────────────────────────────────────────────────────────────────────────────┘
```

Improved: 
1. mapping in-place for qubits outer -> inner
2. tag(s) to the header instead 

```
                   ┌─────────────────────────────────────┐        
                    Circuit_0x740eb7df41d7eca6[['tag']]:          
                    [ (0,0) -> a: ───X───@───]                    
(0, 0): ───X───@────[                    │   ]──────────────H───  
               │    [ (0,1) -> b: ───Y───X───]                    
               │    [                        ]                    
               │    [ (1,1) -> c: ───Z───────]                    
               │    │                                             
(0, 1): ───Y───X────#2───────────────────────────────────────H─── 
                    │                                             
(1, 0): ───Z───H────┼───────────────────────────────────H──────── 
                    │                                             
(1, 1): ────────────#3───────────────────────────────────────H─── 
                   └─────────────────────────────────────┘        
```

balopat (changes requested):
Reviewed the circuit diagram part - I added some ideas, mostly stylistic and space saving improvements, feel free to add them to an issue, but some might be easy wins.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,382 @@
...
+            result += f'.repeat({self.repetitions})'
+        return result
+
+    def __str__(self):
+        # TODO: support out-of-line subcircuit definition in string format.
+        header = self.circuit.serialization_key() + ':'
+        msg_lines = str(self.circuit).split('\n')
+        msg_width = max([len(header) - 4] + [len(line) for line in msg_lines])
+        circuit_msg = '\n'.join(
+            ['[ {line:<{width}} ]'.format(line=line, width=msg_width) for line in msg_lines]
```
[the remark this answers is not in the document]
95-martin-orion: RE circuit diagrams: I intentionally avoided spending too much effort on constructing these since the intent is to replace them with out-of-line diagrams ASAP. Unless you see this as an urgent readability issue, I think we should postpone any diagram-related fixes until the out-of-line update rolls out.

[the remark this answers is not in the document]
95-martin-orion: For reference, the need for out-of-line diagrams is captured in #3605.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,382 @@
...
+
+class CircuitOperation(ops.Operation):
+    """An operation that encapsulates a circuit.
+
+    This class captures modifications to the contained circuit, such as tags
+    and loops, to support more condensed serialization. Similar to
+    GateOperation, this type is immutable.
+    """
+
+    def __init__(self, circuit: 'cirq.FrozenCircuit') -> None:
```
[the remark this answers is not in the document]
95-martin-orion: My intent here was to emphasize reusability of the `FrozenCircuit` since that's a key feature of `CircuitOperation`, but I can see how accepting `OP_TREE` would be more user-friendly. I'll look at this alongside @maffoo's recommendations for the constructor.

cirq/circuits/frozen_circuit.py
```diff
@@ -88,6 +88,10 @@ def device(self) -> devices.Device:
     def __hash__(self):
         return hash((self.moments, self.device))
 
+    def serialization_key(self):
+        # TODO: use this key in serialization and support user-specified keys.
+        return f'Circuit_{hash(self) % int(1e6):06d}:'
```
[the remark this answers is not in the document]
95-martin-orion: Needed some finagling to handle negative hash values, but done.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,382 @@
...
+        modulus: INT_TYPE = 0,
+        allow_invert: bool = False,
+        validate_modulus: bool = False,
+    ) -> 'CircuitOperation':
+        """Returns a copy of this operation repeated 'repetitions' times.
+
+        Args:
+            repetitions: Number of times this operation should repeat. This
+                is multiplied with any pre-existing repetitions.
+            modulus: If this is specified, the final repetition count will be
```
balopat, line 265: Why do we need `modulus` + `allow_invert`? Do you have a motivating use case?

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,382 @@
...
+                ignoring any existing qubit map.
+
+        Returns:
+            A copy of this operation targeting `new_qubits`.
+
+        Raises:
+            ValueError: `new_qubits` has a different number of qubits than
+                this operation.
+        """
+        if len(new_qubits) != protocols.num_qubits(self.circuit):
```
balopat, line 334: nit: you're running `num_qubits` twice on the error path... we could save that repetition by pulling `expected` out in front of the if

cirq/circuits/circuit_operation.py
95-martin-orion: Adam highlights this as a useful feature in the [RFC](https://tinyurl.com/cirq-circuitoperation-rfc) - feel free to check with him for specific use cases.

cirq/circuits/circuit_operation_test.py
balopat: Funny, I had the same idea as @maffoo - a `FrozenCircuit.on(qubits)` could return a `CircuitOperation`. Maybe we don't need `on(qubits)` just `to_op()`...

1. not sure I follow - `op = c.to_op(); [c.to_op(),c.to_op(), op, op]` will have the same exact `FrozenCircuit` instance all four times, just with 3 different `CircuitOperation` instances.
2. local import can workaround that issue 
This thing works quite well: 
```
    def to_op(self):
         from cirq.circuits import CircuitOperation
         return CircuitOperation(self)
```
This is not urgent though, we can follow-up after this PR.

cirq/circuits/circuit_operation.py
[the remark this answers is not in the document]
balopat: Sounds good. If I had to choose between the comments on the PR, this would be one to add.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,384 @@
...
+        Args:
+            circuit: The FrozenCircuit encapsulated by this operation.
+        """
+        self._circuit: 'cirq.FrozenCircuit' = circuit
+        self._repetitions: int = 1
+        self._qubit_map: Dict['cirq.Qid', 'cirq.Qid'] = {q: q for q in circuit.all_qubits()}
+        self._measurement_key_map: Dict[str, str] = {k: k for k in circuit.all_measurement_keys()}
+        self._param_resolver = study.ParamResolver(
+            {p: p for p in protocols.parameter_symbols(circuit)}
+        )
```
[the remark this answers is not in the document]
95-martin-orion: This is probably possible, but the change is non-trivial since `with_qubit_mapping` etc. would need to handle implicit mappings. IIRC, this has a surprising number of edge cases to consider - I'll see if I can revive it.

cirq/circuits/circuit_operation.py
balopat: Ah I see, cool, inviting @viathor to comment.

viathor: I imagined it a little differently: at the instantiation time of a CircuitOperation the user could provide an optional modulus for exponentiation as a hint (and we could validate it in `__init__` the way it is done in this method).

Regarding a use-case: I don't have one that I need at the moment. The rationale for this was that in many situations one wants to scan over a range of exponents. This range is often bounded, but it may not be clear what the bound for a particular circuit is. So if the user wants to scan the full range, they have to figure out what "full" means. Of course, they also have to do this in order to provide the hint, but the code that prepares a CircuitOperation and the code that consumes it may be different and this is easier to do for the producer of CircuitOperation.

That said, at the time I wrote the comment I was trying to anticipate a need and wanted to see whether others find it useful. If there is currently no explicit need for it, we can drop it and add it when users ask for it.

cirq/circuits/circuit_operation.py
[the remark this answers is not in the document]
95-martin-orion: OK, implicit maps turned out less complicated than I remembered. It shifts the performance cost from `base_operation` to the `with_*` operations, but seeing as those operations call `base_operation` I think it's a net improvement.

cirq/circuits/circuit_operation.py
balopat: I see, thank you! @95-martin-orion what do you think? I think we can keep it, but moving modulus with validation to the constructor makes more sense to me, as then the exponentiation code could make use of the modulus being "stored" as well.

95-martin-orion: Comment from @maffoo on #3509 (a prior version of this change) regarding modulus in the constructor:

> I'm not sure this is a good idea. When we ask hardware to repeat a circuit we typically want it to run that many times, we don't want things that are equivalent to identities to get optimized away (think about randomized benchmarking). Such transformations could be implemented in an optimization pass, but it seems odd to bake this in to the gate itself.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,382 @@
...
+            return '{' + ', '.join(pairs) + '}'
+
+        if self.qubit_map != base_op.qubit_map:
+            args.append(f'qubit_map={dict_str(self.qubit_map)}')
+        if self.measurement_key_map != base_op.measurement_key_map:
+            args.append(f'key_map={dict_str(self.measurement_key_map)}')
+        if self.param_resolver != base_op.param_resolver:
+            args.append(f'params={self.param_resolver.param_dict}')
+        if self.repetitions != 1:
+            args.append(f'loops={self.repetitions}')
```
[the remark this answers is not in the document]
95-martin-orion: If you'd like to see this reflected in the final subcircuit diagram format, please copy this comment to #3605.

cirq/circuits/circuit_operation.py
95-martin-orion: If you'd like to see this reflected in the final subcircuit diagram format, please copy this comment to #3605.

95-martin-orion:
> I'd suggest that the CircuitOperation constructor could accept qubit_mapping, params, etc as keyword args (could even implement as a dataclass so you wouldn't have to write out all the boilerplate).

Converted `CircuitOperation` to a dataclass and cleaned up the repr. @balopat, I'll see if I can capture `OP_TREE` + `InsertionStrategy` + device as `InitVar` alternatives to `circuit` in the new structure.

cirq/circuits/circuit_operation.py
95-martin-orion: This sounds reasonable to me, but `Operation` also has a `transform_qubits` method. Let's open a separate PR for this rather than addressing it here. For now, I'll update `with_qubit_mapping` to accept functions.

One other thought on this: `with_qubit_mapping` does not directly modify the circuit, unlike `transform_qubits`. I think having separate names to differentiate between these behaviors is worthwhile.

95-martin-orion: Opened #3611 to track the `Circuit` and `Operation` change.

cirq/circuits/circuit_operation.py
balopat: Well, for randomized benchmarking we wouldn't set the modulus. Having to keep repeating the modulus puts the extra effort on the algorithm developer's side: in case the algo developer wants to use a "modulus optimized subcircuit" across a range of different exponents, they have to ensure that for all the exponents the validation will happen - alternatively, run the validation once and then use it - but they have to ensure then to use the same modulus across all calls, i.e. they'll have to design to store the modulus somewhere. With the constructor based method, they can choose and validate once and reuse and exponentiate the subcircuit freely with built-in modulus optimization.
@maffoo what do you think?

cirq/circuits/circuit_operation.py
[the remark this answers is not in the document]
95-martin-orion: This is now supported along with the original `circuit` parameter.

cirq/circuits/circuit_operation_test.py
95-martin-orion: Done.

95-martin-orion:
All review comments are either addressed in code or require further discussion before implementing. PTAL

balopat (approved):
LGTM - The only thing bugging me is the modulus, but it is something we can change later based on feedback as well.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,407 @@
...
+            param_values: A map or ParamResolver able to convert old param
+                values to new param values. This map will be composed with any
+                existing ParamResolver.
+
+        Returns:
+            A copy of this operation with its ParamResolver updated as specified
+                by param_values.
+        """
+        new_resolver = protocols.resolve_parameters(self.param_resolver, param_values, recursive)
+        return self.updated_copy(param_resolver=new_resolver)
```
maffoo, line 402: The way parameter resolution works on resolvers, we get some behavior here that is a bit odd. For example, suppose I have a circuit with parameters `a` and `b` and a map `a -> A` and `b -> B`, and I want to now map `A -> C`. I would expect to end up with `a -> C` and `b -> B` but the actual param resolver has an extra entry:
```
In [4]: cirq.resolve_parameters(cirq.ParamResolver({'a': 'A', 'b': 'B'}), {'A': 'C'})
Out[4]: cirq.ParamResolver({'A': sympy.Symbol('C'), 'a': sympy.Symbol('C'), 'b': sympy.Symbol('B')})
```
That extra `A -> C` entry is not needed here because `A` is not a parameter of the circuit. Similarly, if we compose with a map `A -> a` we get this extra entry and the identity remains in the map:
```
In [5]: cirq.resolve_parameters(cirq.ParamResolver({'a': 'A', 'b': 'B'}), {'A': 'a'})
Out[5]: cirq.ParamResolver({'A': sympy.Symbol('a'), 'a': sympy.Symbol('a'), 'b': sympy.Symbol('B')})
```
We could modify `ParamResolver._resolve_parameters_` to remove these identity elements, but in general param resolvers can't know the "full" key set they should apply to so removing the unneeded `A ->` entries that don't exist in the circuit is something we just need to do here, e.g.:
```python
new_params = {}
for k in protocols.parameter_symbols(self.circuit):
    v = self.param_resolver.value_of(v, recursive)
    v = protocols.resolve_parameters(v, param_values, recursive)
    if v != k:
        new_params[v] = k
return self.updated_copy(param_resolver=cirq.ParamResolver(new_params))
```

cirq/circuits/circuit.py
```diff
@@ -763,12 +758,41 @@ def are_all_matches_terminal(self, predicate: Callable[['cirq.Operation'], bool]
...
-        return all(
+        from cirq.circuits import CircuitOperation
+
+        # TaggedOperations can wrap CircuitOperations.
+        def is_circuit_op(op):
+            if isinstance(op, CircuitOperation):
+                return True
+            if isinstance(op, ops.TaggedOperation):
+                return is_circuit_op(op.sub_operation)
+            return False
```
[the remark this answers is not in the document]
maffoo: Also, I should say that we should document here that we only apply the given predicate to operations that are not `CircuitOperations` and that we instead transparently recurse into `CircuitOperations` looking for matches. If the use tried to call `circuit.are_all_matches_terminal(lambda op: isinstance(op, cirq.CircuitOperation))` that wouldn't work as expected with with implementation (it would never match anything).

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,407 @@
...
+
+    circuit: 'cirq.FrozenCircuit'
+    repetitions: int
+    qubit_map: Dict['cirq.Qid', 'cirq.Qid']
+    measurement_key_map: Dict[str, str]
+    param_resolver: study.ParamResolver
+
+    def __init__(
+        self,
+        *contents: 'cirq.OP_TREE',
```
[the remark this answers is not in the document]
balopat: That was the original version - I pushed for the OP_TREE because:

1.) OP_TREE contains Circuit, FrozenCircuit as well  
2.) when manually creating CircuitOperation - you'll have to add a bunch of boilerplate unnecessarily

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,407 @@
...
+            A copy of this operation targeting `new_qubits`.
+
+        Raises:
+            ValueError: `new_qubits` has a different number of qubits than
+                this operation.
+        """
+        expected = protocols.num_qubits(self.circuit)
+        if len(new_qubits) != expected:
+            raise ValueError(f'Expected {expected} qubits, got {len(new_qubits)}.')
+        return self.with_qubit_mapping(dict(zip(self.qubits, new_qubits)))
```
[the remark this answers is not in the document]
95-martin-orion: This assertion succeeds in the current version. Note that the ordering enforced by `.qubits` is dependent on the ordering of the original circuit's qubits, not the post-mapping qubits.

cirq/circuits/circuit_operation.py
95-martin-orion: The code snippet above will behave incorrectly for `recursive=True`:
```python
x, y, z, X, Y, Z = [sympy.Symbol(x) for x in 'xyzXYZ']
op = cirq.CircuitOperation(cirq.CX(a, b) ** x, cirq.CZ(a, b) ** z)
op = op.with_params({x: X, z: Z}, recursive=True)
op2 = op.with_params({X: Y}, recursive=True)
op3 = op2.with_params({Z: X}, recursive=True)

# Removing mappings whose keys are not in the circuit:
assert op2.param_resolver.value_of(X) == X
assert op3.param_resolver.value_of(Z) == X

# Using ParamResolver composition:
assert op2.param_resolver.value_of(X) == Y
# {x: Y, X: Y, z: X, Z: X} flattens to {x: Y, X: Y, z: Y, Z: Y} when applied recursively
assert op3.param_resolver.value_of(Z) == Y
```
I'm okay with taking this as a sign that we should not support recursive parameter resolution in `CircuitOperation` (I'm frankly not fond of recursive resolution in general) but we'll need to be explicit about that.

cirq/circuits/circuit_operation.py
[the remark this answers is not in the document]
95-martin-orion: Discussed out-of-band: `OP_TREE` support and `to_op` will be reverted.

cirq/circuits/circuit_operation.py
95-martin-orion: Discussed out-of-band: removing `modulus` from this PR and opening an issue to track it as a separate feature.

cirq/circuits/circuit_operation.py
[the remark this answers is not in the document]
95-martin-orion: This change has been applied.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,407 @@
...
+    This class captures modifications to the contained circuit, such as tags
+    and loops, to support more condensed serialization. Similar to
+    GateOperation, this type is immutable.
+    """
+
+    circuit: 'cirq.FrozenCircuit'
+    repetitions: int
+    qubit_map: Dict['cirq.Qid', 'cirq.Qid']
+    measurement_key_map: Dict[str, str]
+    param_resolver: study.ParamResolver
```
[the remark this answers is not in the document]
95-martin-orion: Done (was present in pre-OP_TREE state, now restored)

cirq/circuits/circuit_operation.py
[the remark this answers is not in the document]
95-martin-orion: Opting to leave fractional repeats for future work.

cirq/circuits/circuit_operation.py
95-martin-orion: This change has been applied.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,407 @@
...
+from cirq.circuits.insert_strategy import InsertStrategy
+
+if TYPE_CHECKING:
+    import cirq
+
+
+INT_TYPE = Union[int, np.integer]
+
+
+@dataclasses.dataclass(frozen=True, init=False)
```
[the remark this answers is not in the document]
95-martin-orion: Resolved - without `OP_TREE`, we once again use the generated `__init__`.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,407 @@
...
+            )
+        }
+
+    def _decompose_(self) -> 'cirq.OP_TREE':
+        result = self.circuit.unfreeze()
+        result = result.transform_qubits(lambda q: self.qubit_map.get(q, q))
+        if self.repetitions < 0:
+            result = result ** -1
+        result = protocols.with_measurement_key_mapping(result, self.measurement_key_map)
+        result = protocols.resolve_parameters(result, self.param_resolver)
```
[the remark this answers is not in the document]
95-martin-orion: I've used `cirq.decompose_once` for this in the tests, but I agree that a direct method may be more convenient. Leaving for potential future work.

balopat:
There's some weirdness going on with hashing: 


<details>

```
from cirq import *
from sympy import *

# internal, level 2
q = LineQubit.range(3)
# internal level 1
k = LineQubit.range(3)
# external level 0
s = LineQubit.range(3)

level2_circ = FrozenCircuit(X(q[0]) ** Symbol('a'))

# "renames" internal 'a' to external 'a1'
op1_level2 = CircuitOperation(level2_circ).with_params({S('a'):S('a1')})
# "renames" internal 'a' to external 'a2'
op2_level2 = CircuitOperation(level2_circ).with_params({S('a'):S('a2')})

level1_circ = FrozenCircuit(X(k[0]) ** S('x'),
                      op1_level2.with_qubits(k[0]),
                      op2_level2.with_qubits(k[0]))
# renames level 1 params x, a1, a2 to external o1, o2, o3 
op_level1 = CircuitOperation(level1_circ).with_params({S('x'): S('o1'), S('a1'): S('o2'), S('a2'): S('o3')})

# add an external-only o0 parameter
top_circ = Circuit(Y(s[0]) ** S('o0'), op_level1.with_qubits(s[0]))
print(top_circ)
```

</details>

Throws an error: 

<details>

```
  File "scratch_10.py", line 26, in <module>
    print(top_circ)
  File "/Users/balintp/dev/proj/Cirq/cirq/circuits/circuit.py", line 215, in __str__
    return self.to_text_diagram()
  File "/Users/balintp/dev/proj/Cirq/cirq/circuits/circuit.py", line 1066, in to_text_diagram
    transpose=transpose,
  File "/Users/balintp/dev/proj/Cirq/cirq/circuits/circuit.py", line 1128, in to_text_diagram_drawer
    include_tags,
  File "/Users/balintp/dev/proj/Cirq/cirq/circuits/circuit.py", line 2085, in _draw_moment_in_diagram
    info = get_circuit_diagram_info(op, args)
  File "/Users/balintp/dev/proj/Cirq/cirq/protocols/circuit_diagram_info_protocol.py", line 168, in _op_info_with_fallback
    name = str(op.untagged)
  File "/Users/balintp/dev/proj/Cirq/cirq/circuits/circuit_operation.py", line 152, in __str__
    header = self.circuit.serialization_key() + ':'
  File "/Users/balintp/dev/proj/Cirq/cirq/circuits/frozen_circuit.py", line 93, in serialization_key
    key = hash(self)
  File "/Users/balintp/dev/proj/Cirq/cirq/circuits/frozen_circuit.py", line 89, in __hash__
    return hash((self.moments, self.device))
  File "/Users/balintp/dev/proj/Cirq/cirq/ops/moment.py", line 251, in __hash__
    return hash((Moment, tuple(sorted(self.operations, key=lambda op: op.qubits))))
  File "<string>", line 3, in __hash__
TypeError: unhashable type: 'dict'
```

</details>

cirq/circuits/circuit.py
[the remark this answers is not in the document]
95-martin-orion: Done and done. Added a test for that `are_all_matches_terminal` case - seems like documenting the behavior is a reasonable "solution" for now.

cirq/circuits/frozen_circuit.py
```diff
@@ -88,6 +88,13 @@ def device(self) -> devices.Device:
     def __hash__(self):
         return hash((self.moments, self.device))
 
+    def serialization_key(self):
+        # TODO: use this key in serialization and support user-specified keys.
+        key = hash(self)
```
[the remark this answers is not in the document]
95-martin-orion: Done.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,407 @@
...
+        )
+
+    def base_operation(self) -> 'CircuitOperation':
+        """Returns a copy of this operation with only the wrapped circuit.
+
+        Key and qubit mappings, parameter values, and repetitions are not copied.
+        """
+        return CircuitOperation(self.circuit)
+
+    def updated_copy(
```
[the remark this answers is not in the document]
95-martin-orion: Renamed to `replace` and converted to a thin wrapper around `dataclasses.replace`. Works beautifully.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,407 @@
...
+        args = [f'circuit={self.circuit!r}']
+        if self.repetitions != 1:
+            args.append(f'repetitions={self.repetitions}')
+        if self.qubit_map:
+            args.append(f'qubit_map={proper_repr(self.qubit_map)}')
+        if self.measurement_key_map:
+            args.append(f'measurement_key_map={proper_repr(self.measurement_key_map)}')
+        if self.param_resolver:
+            args.append(f'param_resolver={proper_repr(self.param_resolver)}')
+        argstr = ",\n".join(args)  # no backslashes allowed in fstring braces
```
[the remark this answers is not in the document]
95-martin-orion: Done. Also did some tidying: the args to `CircuitOperation` are now indented. Haven't tested this with nested ops yet - will update here when I do.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,407 @@
...
+                new_reps -= modulus
+        return self.updated_copy(repetitions=new_reps)
+
+    def __pow__(self, power: int) -> 'CircuitOperation':
+        return self.repeat(power)
+
+    def with_qubit_mapping(
+        self,
+        qubit_map: Optional[Dict['cirq.Qid', 'cirq.Qid']] = None,
+        transform: Optional[Callable[['cirq.Qid'], 'cirq.Qid']] = None,
```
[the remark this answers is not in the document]
95-martin-orion: Done, although I wound up using `callable(qubit_map)` because mypy didn't like `isinstance(qubit_map, Callable)`.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,407 @@
...
+        """
+        if transform is None:
+            if qubit_map is None:
+                raise ValueError('with_qubit_mapping requires one of {qubit_map, transform}.')
+            else:
+                transform = lambda q: qubit_map.get(q, q)  # type: ignore
+        elif qubit_map is not None:
+            raise ValueError('with_qubit_mapping received both of {qubit_map, transform}.')
+        base_map = {q: q for q in self.circuit.all_qubits()}
+        base_map.update(self.qubit_map)
```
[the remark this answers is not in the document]
95-martin-orion: Done.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,407 @@
...
+            A copy of this operation with measurement keys updated as specified
+                by key_map.
+
+        Raises:
+            ValueError: The new operation has a different number of measurement
+                keys than this operation.
+        """
+        base_map = {m: m for m in self.circuit.all_measurement_keys()}
+        base_map.update(self.measurement_key_map)
+        new_map = {k: key_map.get(v, v) for k, v in base_map.items() if k != key_map.get(v, v)}
```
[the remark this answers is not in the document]
95-martin-orion: Done.

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,407 @@
...
+        cls, circuit, repetitions, qubit_map, measurement_key_map, param_resolver, **kwargs
+    ):
+        result = (
+            cls(circuit)
+            .with_qubit_mapping(dict(qubit_map))
+            .with_measurement_key_mapping(measurement_key_map)
+            .with_params(param_resolver)
+        )
+        if repetitions != 1:
+            return result.repeat(repetitions)
```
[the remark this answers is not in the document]
95-martin-orion: Done. Added a note to parallel the notes in GateOperation: this is only legal because `CircuitOperation` is (intended to be) immutable. Otherwise we'd need to adhere to the "returns a copy" policy more rigorously.

95-martin-orion:
> There's some weirdness going on with hashing: {...}

Hashing weirdness should be resolved now.

cirq/circuits/circuit_operation.py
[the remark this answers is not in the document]
95-martin-orion: Tested and everything works as expected.

cirq/circuits/circuit_operation.py
95-martin-orion: Single-step resolution is now enforced, with irrelevant parameters discarded.

maffoo (approved):
LGTM!

We can revisit parameter resolution to see if we can work out a way to deal with recursive resolution; for now this looks good!

cirq/circuits/circuit_operation.py
```diff
@@ -0,0 +1,361 @@
...
+        indented_args = args.replace('\n', '\n    ')
+        return f'cirq.CircuitOperation({indented_args[:-4]})'
+
+    def __str__(self):
+        # TODO: support out-of-line subcircuit definition in string format.
+        header = self.circuit.serialization_key() + ':'
+        msg_lines = str(self.circuit).split('\n')
+        msg_width = max([len(header) - 4] + [len(line) for line in msg_lines])
+        circuit_msg = '\n'.join(
+            ['[ {line:<{width}} ]'.format(line=line, width=msg_width) for line in msg_lines]
```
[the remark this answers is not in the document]
95-martin-orion: Done.

cirq/circuits/circuit_operation.py
95-martin-orion: I've added a TODO here for graceful handling of recursive resolution, and opened #3619 to track it.

95-martin-orion:
Waiting on final review from @balopat. Balint, if I'm not around but everything looks good, you're welcome to merge this PR without further input from me.

balopat (approved):

PR #3583, opened 2020-12-04
Remove bad delete char from docstring

lamberta:
Can't see in the diff, but there was an errant delete character in the docstring causing an unicode issue when rendering the API reference page.


mpharrigan (approved):
fun

PR #3585, opened 2020-12-04
[docs] Fixes in install.md, rabi tutorial, variational algo and reservations tutorial

balopat:
More docs nits. 

- python version in install.md
- rabi oscillation link issue
- variational algo title to match nav 
- remove reservations notebook from nav

### DIFF
--- a/docs/_book.yaml
+++ b/docs/_book.yaml
@@ -147,8 +147,6 @@ upper_tabs:
       - heading: "Google hardware"
       - title: "Getting started with QCS"
         path: /cirq/tutorials/google/start
-      - title: "Reservations"
-        path: /cirq/tutorials/google/reservations
 
       ####  AQT HARDWARE  ####
       - heading: "AQT hardware"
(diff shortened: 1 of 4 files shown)


balopat:
cc @wcourtney - now that we have the reservations UI live, we can remove the notebook from the nav but I'll keep it in the repo for reference, wdyt?

viathor (approved):

docs/_book.yaml
```diff
@@ -147,8 +147,6 @@ upper_tabs:
       - heading: "Google hardware"
       - title: "Getting started with QCS"
         path: /cirq/tutorials/google/start
-      - title: "Reservations"
```
wcourtney, line 150: Is this because we can use the console rather than the colab? SGTM, but we should probably include new docs for this...that's on me :-)

wcourtney (approved):

docs/_book.yaml
balopat: Yes, exactly because of that! Agreed - we probably should :)

PR #3590, opened 2020-12-06
Circuits can now be multiplied by numpy ints

dstrain115:
- np.array([2])[0] * circuit would not work before.
- Note that __array_priority__ is needed so that numpy doesn't try to
call its own __mul__ function for its types and instead defers to
Circuit __rmul__.


cirq/circuits/circuit.py
```diff
@@ -1370,21 +1371,25 @@ def __radd__(self, other):
...
+    def __mul__(self, repetitions: INT_TYPE):
+        if not isinstance(repetitions, (int, np.int32, np.int64)):
             return NotImplemented
-        return Circuit(self._moments * repetitions, device=self._device)
+        return Circuit(self._moments * int(repetitions), device=self._device)
 
-    def __rmul__(self, repetitions: int):
-        if not isinstance(repetitions, int):
+    def __rmul__(self, repetitions: INT_TYPE):
+        print('Hi!!!!')
```
viathor, line 1389: Hi! :-)

viathor (commented):
What is the use-case for this? I worry that this could potentially conflict with a potential future multiplication to change global phase(*).

(*) This could be useful now that we have subcircuits. Global phase of U matters when you construct controlled-U.

dstrain115:
> What is the use-case for this? I worry that this could conflict with a potential future multiplication to change global phase(*).

This was requested by Jonathan Gross, internal bug number 174799117.  I don't think this should conflict with global phase any more than it already does.  2 * circuit already doubles the circuit.  This just fixes it so that the np.int64 of 2 also does the same thing.  

If we used multiplication of a circuit to effect global phase, this would conflict with both (at least for int values of phase).

I'll fix the debug statement.  I was double-checking the precedence for multiplication and forgot to take it out.  How embarrassing.

cirq/circuits/circuit.py
dstrain115: Removed.  Sigh.

cirq/circuits/circuit_test.py
```diff
@@ -631,6 +631,8 @@ def test_multiply(circuit_cls):
     assert c * 0 == circuit_cls()
     assert d * 0 == circuit_cls()
     assert d * 2 == circuit_cls([cirq.Moment([cirq.X(a)]), cirq.Moment([cirq.X(a)])])
+    assert d * np.array([2])[0] == circuit_cls([cirq.Moment([cirq.X(a)]), cirq.Moment([cirq.X(a)])])
```
balopat, line 634: can you add some cases with explicit np.dtypes? E.g. I'm not sure we're covering np.int32, though you did add it to the implementation.

balopat (approved):
LGTM with a nit around a missing test case

cirq/circuits/circuit.py
```diff
@@ -1370,21 +1371,24 @@ def __radd__(self, other):
...
+        if not isinstance(repetitions, (int, np.int32, np.int64)):
             return NotImplemented
-        self._moments *= repetitions
+        self._moments *= int(repetitions)
         return self
 
-    def __mul__(self, repetitions: int):
-        if not isinstance(repetitions, int):
+    def __mul__(self, repetitions: INT_TYPE):
+        if not isinstance(repetitions, (int, np.int32, np.int64)):
```
viathor, line 1384: Consider including other numpy integer types to avoid surprises (e.g. int8, int16).

cirq/circuits/circuit_test.py
viathor: +1

viathor (approved):
LGTM. Thanks for the explanation.

cirq/circuits/circuit.py
mpharrigan: relevant: https://stackoverflow.com/a/37727662

dstrain115: Thanks for the link.  That's super useful.  I have added a bunch of different types to the test and changed the check to np.integer.

Issue #3278, opened 2020-08-28
Refactor cirq.Moment to support efficiently finding operations by qubit


Strilanc:
A common access pattern, when optimizing or analyzing a quantum circuit, is to check whether there is an operation at a particular place at a particular time. For example, when inserting into a circuit, finding the first moment that has no operation on certain qubits is necessary. cirq.Circuit supports efficient access by time, but not by place. This is fundamentally due to cirq.Moment being a list of operations.

I suggest that we refactor cirq.Moment (preserving backwards compatibility) to always have, or lazily create on demand, an internal dictionary mapping qubits to operations. This dictionary can then be used in other operations on moment, to give them better asymptotic guarantees.

balopat:
Are you referring to roughly making this method cache its results? 

https://github.com/quantumlib/Cirq/blob/64939e53cfbbcf4984a254d3dc68690da9dd27de/cirq/ops/moment.py#L287-L295

```
cirq.Circuit(cirq.X.on(a), cirq.Y.on(b), cirq.X.on(a))[0][a]
```
works nicely: 

```
cirq.X(cirq.LineQubit(0))
```

Also xref #3145.

wingers:
Hi @balopat, I'd like to work on this.

balopat:
You got it @wingers! Thanks for taking it on!

PR #3592, opened 2020-12-08
Refactor cirq.Moment to support efficiently finding operations by qubit

wingers:
Fixes https://github.com/quantumlib/Cirq/issues/3278.

### DIFF
--- a/cirq/ops/moment.py
+++ b/cirq/ops/moment.py
@@ -87,19 +87,16 @@ class Moment:
 
         self._operations = tuple(op_tree.flatten_to_ops(contents))
 
-        affected_qubits = []
         # An internal dictionary to support efficient operation access by qubit.
         self._qubit_to_op = {}
         for op in self.operations:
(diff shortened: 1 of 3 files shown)


cirq/ops/moment.py
```diff
@@ -87,9 +87,17 @@ def __init__(self, *contents: 'cirq.OP_TREE') -> None:
 
         self._operations = tuple(op_tree.flatten_to_ops(contents))
 
-        # Check that operations don't overlap.
-        affected_qubits = [q for op in self.operations for q in op.qubits]
+        affected_qubits = []
```
mpharrigan, line 90: You can check that operations don't overlap by using the keys of `self._qubit_to_op` instead of constructing an additional list, `affected_qubits`

wingers: Done, much cleaner now!

mpharrigan (approved):
LGTM!

PR #3593, opened 2020-12-08
Remove stale rtd_docs/docs_coverage_test.py references from check/* scripts

tanujkhattar:
rtd_docs/docs_coverage_test.py file was deleted in PR3587 but some references to the file were left in check/* scripts because of which the scripts now fail. This PR removes those stale references.


dabacon:
This seems to be blocking submits?

mpharrigan (approved):

balopat:
Thanks for the quick fix - sorry for the trouble!

PR #3595, opened 2020-12-08
Remove unused dependency

dabacon:
Amusing story.  `pylint` has a flag `-j` which allows for parallel linting.  When I ran it in this mode it entered an infinite loops stuck on this.  Normal `pylint` did not seem to uncover it.  ¯\\\_(ツ)\_/¯


viathor (approved):
I wonder what it was up to in that infinite loop...

Issue #3409, opened 2020-10-10
Simulate factorizable circuit


mpharrigan:
**Is your feature request related to a use case or problem? Please describe.**

If you're running many "independent" circuits in parallel on a given set of qubits (where independent means there are no entangling gates between them; i.e. you can factorize the circuit (and therefore wavefunction)) you can simulate it much more efficiently than working in the full 2^(all n qubits) space.

**Describe the solution you'd like**

I think it should be possible to transform a circuit into an interaction graph, use a connected components algorithm to figure out how to split it into circuits operating on a subsets of qubits, and return a dictionary of state vectors keyed by the set of qubits.

**What is the urgency from your perspective for this issue? Is it blocking important work?**

Any sort of characterization that you run in parallel and requires simulating the circuit (read: XEB) would benefit from this! There's a version of this hacked into cirq.experiments.parallel_two_qubit_xeb

Ashalynd:
Is someone already looking at this, or can I give it a shot?

balopat:
Let's chat about it on Cirq Cynq today @Ashalynd - thanks for volunteering!!

Ashalynd:
Sure, sounds good! (I'll be there today ~15 minutes later, schedule mangled due to the winter time).

balopat:
Discussed on Cirq Cynque: 

Use case: mostly characterization, example: two qubit parallel XEB 

Initial design ideas: 

 1. `factorize_circuit`: Factor the circuit into smaller circuits if possible  - simulators could use this to attempt to factorize circuits in general 
 2. the reverse: create a function to run smaller circuits "in parallel" on the device/simulator and then post-processing: return the measurement results in a nicely sliceable form

Other thoughts: when SubCircuits are implemented keeping the factorized structure would be nice.

Ashalynd:
We just discussed this with Alan, who suggested that the reverse might be more interesting (and physically relevant) problem (with the reverse being understood as: given a number of circuits and the specific device (or a simulator), and the information about two-qubit error rates, figure out how many of them could be reliably run together on this device.

Will open another issue for that.

Ashalynd:
Thinking further about this.

Matthew @mpharrigan : a few clarifying questions :) 

1. Is it correct to assume that the circuits in two examples below (taken from tutorial) both have two factors (both could be split into [0] and [1, 2] in terms of the participating qubits): 

```
(0, 0): ───────H───────

(1, 0): ───@───@───H───
           │   │
[...]

PR #759, opened 2018-07-25
Add CircuitDag class

cduck:
Resolves #547.

Add networkx as a dependency.

Issue #547, opened 2018-06-30
Consider adding CircuitToDag and DagToCircuit methods


Strilanc:
This is not a priority at the moment, but something that people might find useful. Sometimes it's easier to reason about a graph.

PR #3599, opened 2020-12-09
proof-of-concept for factorization with CirquitDag (see #3409)

Ashalynd:
Related to #3409  - this is to demonstrate that CircuitDag seems to be able to do circuit factorization already.


95-martin-orion (commented):
Adding @mpharrigan to get his thoughts, but from my perspective there's a pretty clear path from this to a solution for #3409.

cirq/circuits/circuit_dag.py
```diff
@@ -209,3 +209,24 @@ def findall_nodes_until_blocked(
...
+            if node in input_set:
+                yield node.val
+
+    def factorize(self) -> Iterator['cirq.CircuitDag']:
+        """Tries to factorize the underlying graph, using connected components algorithm.
+        If no factorization is possible, returns a sequence with a single element (itself).
+        """
+        for c in networkx.weakly_connected_components(self):
+            yield CircuitDag.from_ops(
+                self.findall_nodes_from_set(c), can_reorder=self.can_reorder, device=self.device
```
mpharrigan, line 231: What impact does `can_reorder` and `device` have on the output of `factorize`?

mpharrigan:
So I thought CircuitDag wouldn't work for this problem, but clearly I was mistaken. I'm still confused exactly how it works. Can you give me an explanation for what the CircuitDag is and why the connected components of it work to factorize the circuit?

cirq/circuits/circuit_dag.py
mpharrigan: How do you get a normal `Circuit` back? Should we return `Circuit`s instead of `CircuitDag`s in this function?

cirq/circuits/circuit_dag.py
```diff
@@ -209,3 +209,24 @@ def findall_nodes_until_blocked(
...
+        """Finds all nodes from the given set.
+
+        Args:
+            input: The input set of Operation nodes.
+        """
+        for node in self.ordered_nodes():
+            if node in input_set:
+                yield node.val
+
+    def factorize(self) -> Iterator['cirq.CircuitDag']:
```
mpharrigan, line 225: How do you get a normal `Circuit` back? Should we return `Circuit`s instead of `CircuitDag`s in this function?

cirq/circuits/circuit_dag.py
```diff
@@ -209,3 +209,24 @@ def findall_nodes_until_blocked(
...
+        """Tries to factorize the underlying graph, using connected components algorithm.
+        If no factorization is possible, returns a sequence with a single element (itself).
+        """
+        for c in networkx.weakly_connected_components(self):
+            yield CircuitDag.from_ops(
```
mpharrigan, line 230: I think this will destroy the moment structure of the input circuit, right? Maybe this isn't necessary if we're using this functionality inside a simulator but it could be important if e.g. trying to factorize a circuit prior to running on a real device when we want to have more precise control over the timing of operations

maffoo: +1 to this. It'd be nice if we could factorize regular circuits (not just `CircuitDag`) in a way that preserves moments, so those circuits could be zipped together to recover the original circuit:
```python
cirq.Circuit.zip(*circuit.factorize()) == circuit
```

cirq/circuits/circuit_dag.py
Ashalynd: can_reorder is a method to decide whether two nodes (e.g. operations) are mutually independent. The default implementation checks that the qubits which participate in these operations don't intersect, but it can be possible to supply a different method at CircuitDag construction.

The device is there to properly reconstruct the original circuit, it isn't used within CircuitDag itself.

Ashalynd:
> So I thought CircuitDag wouldn't work for this problem, but clearly I was mistaken. I'm still confused exactly how it works. Can you give me an explanation for what the CircuitDag is and why the connected components of it work to factorize the circuit?

The way I understand it, CircuitDag builds a forest from circuit's operations (which is a DAG). Given that, weakly connected components algorithm seems natural to apply there: if two qubits ever participate in a common operation, their 'trees' will merge and they'll form one component. 

The qubits to which no operations have been applied, though, would be lost when CircuitDag is created (which could be worked around if necessary).

cirq/circuits/circuit_dag.py
Ashalynd: Fair point. That was my question a while ago (considering one of the examples [here](https://github.com/quantumlib/Cirq/issues/3409#issuecomment-732647238). One way to do it could be by keeping track of the moments corresponding to operations. I'll try to get something working in that direction.

cirq/circuits/circuit_dag.py
Ashalynd: There is a `to_circuit` method which generates a normal `Circuit`.

cirq/circuits/circuit_dag.py
Ashalynd: I have addressed this, but ended up making a spin-off from the existing CircuitDag (because the existing class is already used in a few places). Could you have another look at the PR and let me know your thoughts?

Ashalynd:
The latest changes introduce a separate class, SparseCircuitDag, to orchestrate factorization. I decided to separate it from CircuitDag, because it can be made a bit simpler. If this approach is acceptable, then it is trivial to add Circuit.factorize() method which would utilize SparseCircuitDag (see the tests in sparse_circuit_dag_test which already demonstrate this behavior).

cirq/circuits/sparse_circuit_dag.py
```diff
@@ -0,0 +1,132 @@
...
+        can_reorder: Callable[
+            [Tuple[int, 'cirq.Operation'], Tuple[int, 'cirq.Operation']], bool
+        ] = _disjoint_qubits,
+        incoming_graph_data: Any = None,
+        device: devices.Device = devices.UNCONSTRAINED_DEVICE,
+    ) -> None:
+        """Initializes a CircuitDag.
+
+        Args:
+        can_reorder: A predicate that determines if two operations may be
```
mpharrigan, line 55: I think you're supposed to indent the Args docstrings (according to the google docstrings convention)

cirq/circuits/sparse_circuit_dag.py
```diff
@@ -0,0 +1,132 @@
...
+            reordered.  Graph edges are created for pairs of operations
+            where this returns False.
+
```
mpharrigan, line 58: nit: I also don't think you're supposed to have blank lines in the args docstrings (happy to be corrected though)

cirq/circuits/sparse_circuit_dag.py
```diff
@@ -0,0 +1,132 @@
...
+        circuit_ops_by_moments = (
+            (index, op) for index, moment in enumerate(circuit.moments) for op in moment.operations
+        )
+        return SparseCircuitDag.from_ops_by_moments(
+            circuit_ops_by_moments, can_reorder=can_reorder, device=circuit.device
+        )
+
+    @staticmethod
+    def from_ops_by_moments(
+        operations_by_moment: Iterable[Tuple[int, 'cirq.Operation']],
```
mpharrigan, line 86: this function needs a docstring. It especially needs documentation for `operations_by_moment` which is a somewhat non-standard Cirq thing to be passing around. I assume the `int` in the type signature is the moment index? The current name sortof implies a `Dict[Moment, List[Operations]]`

mpharrigan (commented):
This all seems rather complex. I think the docstrings need to really motivate and explain what `SparseCircuitDag` is and how it compares and contrasts with `CircuitDag`. There's a lot of code that seems like it's duplicated but is subtly different, and that has a high change of confusing future readers of the codebase so any additional factorization of the code would be welcome.

I'm curious if you want to make a case for `SparseCircuitDag` being useful beyond its use in factorizing. If not, then I think the approach can be simplified drastically. For example, have a look at `cirq.contrib.routing.utils.get_circuit_connectivity`. Granted, this does not capture the full time-structure dependence of operations but it quickly and simply makes a graph of what qubits are ever entangled. Calling (strongly-)connected components on the result would give you sets of qubits; then the original circuit could be sliced along those axes. [You can't use that function verbatim: 1) it's in contrib 2) you'd need to do something like `g.add_nodes(circuit.all_qubits())` to avoid the same problem of non-entangling qubits being dropped completely]

cirq/circuits/sparse_circuit_dag.py
```diff
@@ -0,0 +1,132 @@
...
+from cirq import ops, devices
+from cirq.circuits import circuit
+from cirq.ops.moment import Moment
+
+if TYPE_CHECKING:
+    import cirq
+
+
+def _disjoint_qubits(op1: Tuple[int, 'cirq.Operation'], op2: Tuple[int, 'cirq.Operation']) -> bool:
+    """Returns true only if the operations have qubits in common."""
```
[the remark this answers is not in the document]
Ashalynd: Thanks, updated.

cirq/circuits/sparse_circuit_dag.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Edges of the graph are tuples of nodes.  Each edge specifies a required
+    application order between two operations.  The first must be applied before
+    the second.
+
+    Unlike CircuitDag, this graph is not maximalist (the given operation is directly connected
+    only to its immediate predecessors).
+    """
+
+    disjoint_qubits = staticmethod(_disjoint_qubits)
```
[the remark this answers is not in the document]
Ashalynd: Oops, removed.

cirq/circuits/sparse_circuit_dag.py
```diff
@@ -0,0 +1,132 @@
...
+from cirq.circuits import circuit
+from cirq.ops.moment import Moment
+
+if TYPE_CHECKING:
+    import cirq
+
+
+def _disjoint_qubits(op1: Tuple[int, 'cirq.Operation'], op2: Tuple[int, 'cirq.Operation']) -> bool:
+    """Returns true only if the operations have qubits in common."""
+    return not set(op1[1].qubits) & set(op2[1].qubits)
```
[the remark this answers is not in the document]
Ashalynd: Yeah you are right. Changed.

cirq/circuits/sparse_circuit_dag.py
Ashalynd: Done.

cirq/circuits/sparse_circuit_dag.py
Ashalynd: No strong opinion here, removed the line.

cirq/circuits/sparse_circuit_dag.py
Ashalynd: Done!

Ashalynd:
> This all seems rather complex. I think the docstrings need to really motivate and explain what `SparseCircuitDag` is and how it compares and contrasts with `CircuitDag`. There's a lot of code that seems like it's duplicated but is subtly different, and that has a high change of confusing future readers of the codebase so any additional factorization of the code would be welcome.
> 
> I'm curious if you want to make a case for `SparseCircuitDag` being useful beyond its use in factorizing. If not, then I think the approach can be simplified drastically. For example, have a look at `cirq.contrib.routing.utils.get_circuit_connectivity`. Granted, this does not capture the full time-structure dependence of operations but it quickly and simply makes a graph of what qubits are ever entangled. Calling (strongly-)connected components on the result would give you sets of qubits; then the original circuit could be sliced along those axes. [You can't use that function verbatim: 1) it's in contrib 2) you'd need to do something like `g.add_nodes(circuit.all_qubits())` to avoid the same problem of non-entangling qubits being dropped completely]

I have added comments to explain what happens there.
Regarding the fact that it overlaps greatly with CircuitDag - that's correct. The reason is that I think that CircuitDag can be replaced by this version, and my first attempt was to modify CircuitDag in-place. But it seems easier to do this modification in two steps.

Ashalynd:
Gentle ping :)

mpharrigan (changes requested):
Hi @Ashalynd , sorry for the delay!

The latest version looks like cool functionality to dag-ify a circuit, and clearly solves the factorization problem while preserving moment structure. I'm still concerned about two elements:

1. The code duplication and subtle distinctions between `CircuitDag` and `SparseCircuitDag`. I did some digging, since I've never personally used CircuitDag. I see it was added here https://github.com/quantumlib/Cirq/pull/759 in response to this feature request: https://github.com/quantumlib/Cirq/issues/547. Searching for usages turns up some utility in `cirq.contrib.routing`. Otherwise, my reading of these issues and some personal intuition is that this class was originally a "solution looking for a problem". Even in the case with this PR, we found out that CircuitDag was close but not what we wanted. I've opened the following issue about this broader problem: https://github.com/quantumlib/Cirq/issues/3816. In any event, I would like to see a concrete plan for how to replace CircuitDag with this version. I think we'd need to have that all in one PR so we don't have this new class that we know is going to be consolidated (and removed, which would technically be a breaking change). 
2. The nodes being tuples of (operation, moment_index) seems overfit to the current problem of factoring circuits by qubit. You've already seen how this is ergonomically unpleasant (lots of tuple unpacking, a special _disjoint_qubits function instead of using the previous one, ...). Like concern (1) above, I have to ask whether we think this functionality will be useful outside of splitting circuits by qubit. If no, we can get away with something more bespoke and clear rather than trying to be fully general for an as-yet-unseen use case.

With regard to the above concerns, I think it would be valuable to explore the alternative design I sketched out based on `cirq.contrib.routing.utils.get_circuit_connectivity`. I suspect we may be able to get away with a much smaller amount of code focused on the exact problem we're trying to solve. I must admit, this is a very fun way of casting it as a graph problem and you've made me learn a lot about both the history of CircuitDag and what transitive closure is :)

cirq/circuits/sparse_circuit_dag.py
```diff
@@ -0,0 +1,247 @@
...
+                next_frontier = set()
+                all_qubits = set()
+                cur_index = index
+            print("index {}, cur_index {}".format(index, cur_index))
+            new_node = (index, op)
+            dag.add_node(new_node)
+            next_frontier.add(new_node)
+            all_qubits |= set(op.qubits)
+            for prev_node in frontier:
+                if not dag.can_reorder(prev_node, new_node):
```
daxfohl, line 226: This is quadratic on number of qubits. If the default implementation is only add edges for overlapping qubit sets, it should be possible to maintain a mapping from qubit to all the frontier ops that contain it, and run more efficiently. IDK if anyone would plan to simulate an algo with thousands of qubits though.

Ashalynd: Yeah, well noted. It's a small improvement with regard to CircuitDag, that loops over all nodes: https://github.com/quantumlib/Cirq/blob/a61e51b53612735e93b3bb8a7605030c499cd6c7/cirq/circuits/circuit_dag.py#L121 - but I agree that this can be made better.

cirq/circuits/sparse_circuit_dag.py
```diff
@@ -0,0 +1,247 @@
...
+        # Iterate over (op, moment_num) pairs
+        # (the assumption is that the moments are listed in order).
+        # Keep track of all qubits which are, directly or indirectly,
+        # affected by these operations.
+        for index, op in operations_by_moment_index:
+            # Assumpiton: next index corresponds to the next moment
+            # in the circuit
+            if index != cur_index:
+                # Expectation: the sequence is ordered by moment indices.
+                print("! index {}, cur_index {}".format(index, cur_index))
```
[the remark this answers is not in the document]
Ashalynd: Oh, that's indeed a debug leftover. Thanks for noticing!

cirq/circuits/sparse_circuit_dag.py
```diff
@@ -0,0 +1,247 @@
...
+        - Maintain a "frontier" that keeps track of most recent operations
+        on all qubits before the current moment.
+        - Iterate over the (moment_index, operation) pairs.
+        - For every operation, add edges to the operations in the frontier
+        which have qubits in common.
+        - Add every such operation to the next version of the frontier.
+        - When the moment index changes (it is expected to never decrease),
+        add to the next frontier all the operations from the current frontier
+        if they were defined on qubits that didn't participate in the current
+        moment.
```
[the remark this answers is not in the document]
Ashalynd: In theory it should. I could try writing a test for this.

cirq/circuits/sparse_circuit_dag.py
```diff
@@ -0,0 +1,247 @@
...
+        ),
+        (
+        (2, cirq.CZ(cirq.LineQubit(0), cirq.LineQubit(1))),
+        (1, cirq.H(cirq.LineQubit(2)))
+        ),
+        (
+        (3, cirq.H(cirq.LineQubit(0))),
+        (2, cirq.CZ(cirq.LineQubit(0), cirq.LineQubit(1))),
+        (1, cirq.H(cirq.LineQubit(2)))
+        )
```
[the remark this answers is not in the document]
Ashalynd: I tried to follow the style in some other comments. The main difference between CircuitDag and SparseCircuitDag is that in the former, the given node is linked to all previous nodes, as long as they share qubits, whereas SparceCircuitDag only keeps links with the direct ancestors. (Drawing this under alphanumeric constraints might be a challenge :) )

Ashalynd:
> Hi @Ashalynd , sorry for the delay!
No problems at all, thanks for the responding! I am also guilty by not being able to carve as much time as I'd like for this.
> 
> The latest version looks like cool functionality to dag-ify a circuit, and clearly solves the factorization problem while preserving moment structure. I'm still concerned about two elements:
> 
> 1. The code duplication and subtle distinctions between `CircuitDag` and `SparseCircuitDag`. I did some digging, since I've never personally used CircuitDag. I see it was added here #759 in response to this feature request: #547. Searching for usages turns up some utility in `cirq.contrib.routing`. Otherwise, my reading of these issues and some personal intuition is that this class was originally a "solution looking for a problem". Even in the case with this PR, we found out that CircuitDag was close but not what we wanted. I've opened the following issue about this broader problem: #3816. In any event, I would like to see a concrete plan for how to replace CircuitDag with this version. I think we'd need to have that all in one PR so we don't have this new class that we know is going to be consolidated (and removed, which would technically be a breaking change).
Valid point. I could give it a try to do everything in one PR. Main motivation here is that the new class is slightly more efficient that the other due to being less dense, while being also more useful due to preserving moments. On the other side, if it appears that no one actually needs CircuitDag or other versions, I would be just as happy to remove the cruft.

> 2. The nodes being tuples of (operation, moment_index) seems overfit to the current problem of factoring circuits by qubit. You've already seen how this is ergonomically unpleasant (lots of tuple unpacking, a special _disjoint_qubits function instead of using the previous one, ...). Like concern (1) above, I have to ask whether we think this functionality will be useful outside of splitting circuits by qubit. If no, we can get away with something more bespoke and clear rather than trying to be fully general for an as-yet-unseen use case.
I totally agree that this looks ugly. I saw similar construct in other places in Cirq codebase, really wonder whether this doesn't happen to make up for the fact that the Operation doesn't know to which Moment it belongs. (If it was the case, it would have been easy to distinguish between the operations from the different moments, but could also make it trickier to verify equivalence of operations, and probably break quite a few other things). 
> 
> With regard to the above concerns, I think it would be valuable to explore the alternative design I sketched out based on `cirq.contrib.routing.utils.get_circuit_connectivity`. I suspect we may be able to get away with a much smaller amount of code focused on the exact problem we're trying to solve. I must admit, this is a very fun way of casting it as a graph problem and you've made me learn a lot about both the history of CircuitDag and what transitive closure is :)
Disclosure: until I started working on this, I didn't know about transitive closure either :)
If the problem at hand is just to separate sets of qubits, it's a kind of a unionfind problem. At the end, we just want to have one or more sets of non-overlapping qubits. I'll give it a try - it should be less complex than making a graph.

tanujkhattar (changes requested):
This is definitely an extremely complex code to solve the simple (undirected) graph connectivity problem. Unless we have a strong case that this kind of dag representation is useful for something, let's replace this with the easier version. I imagine the union-find implementation would look something like:

```python
def decompose_circuit(circuit):
    uf = nx.utils.UnionFind()
        uf.union(*op.qubits)
    ret = {}
        c_id = uf[op.qubits[0]]
        if c_id in ret:
            ret[c_id].append(op)
        else:
            ret[c_id] = cirq.Circuit(op)
    return list(ret.values())
```

mpharrigan:
closing in favor of https://github.com/quantumlib/Cirq/pull/3873

Issue #3438, opened 2020-10-21
Cirq Circuit Efficient Serialization


mpharrigan:
Cirq supports JSON serialization of all its objects. This is highly portable and extensible but can be somewhat verbose. As part of the XEB refactor I'm working on (https://github.com/quantumlib/ReCirq/issues/85); we need to generate a whole suite of large (wide and deep) and save them to persistent storage. Using the vanilla JSON serialization is both prohibitively slow and disk-space-taking. To see why this may be a problem, the simple circuit of:

```
circuit = cirq.Circuit(
    cirq.H(q0),
    cirq.CNOT(q0, q1)
)
```

turns into [expand]

<details>

```
{
  "cirq_type": "Circuit",
  "moments": [
    {
      "cirq_type": "Moment",
      "operations": [
        {
          "cirq_type": "GateOperation",
          "gate": {
            "cirq_type": "HPowGate",
            "exponent": 1.0,
            "global_shift": 0.0
          },
          "qubits": [
            {
              "cirq_type": "LineQubit",
              "x": 0
            }
          ]
        }
      ]
    },
    {
      "cirq_type": "Moment",
      "operations": [
        {
          "cirq_type": "GateOperation",
          "gate": {
            "cirq_type": "CXPowGate",
            "exponent": 1.0,
            "global_shift": 0.0
          },
          "qubits": [
            {
              "cirq_type": "LineQubit",
              "x": 0
            },
            {
              "cirq_type": "LineQubit",
              "x": 1
            }
          ]
        }
      ]
    }
  ],
  "device": {
    "cirq_type": "_UnconstrainedDevice"
  }
}
```

</details>

For example: ten circuits on nine qubits of depth 100 takes approximately 28MiB on disk. Using `xz`, it compresses down to 240kiB (gzip yields 417kiB). 

## Possible designs

1. Bespoke circuit text representation -- something like qasm where we define a full grammar for circuits that is less redundant

2. Like (1), but just use qasm

3. Wrapper or flag for json serialization that lets the Circuit give itself as compressed JSON. The specifics would have to be worked out, but you can imagine having `"cirq_type": "CompressedCircuit"` and a `data` field which is a hexencoded string of the gzipped bytes of the normal cirq JSON

4. Have a compression option in `to_json` and `read_json` that will read/write the entire file gzipped

5. Same as (4) but implement in ReCirq; i.e. punt on this issue in Cirq and let dependent libraries figure out their own solution.

[...]

PR #3601, opened 2020-12-10
Add concise serialization tools.

95-martin-orion:
Part of #3235. Also contributes to #3438.

This PR adds a `SerializableByKey` interface which any Cirq object can implement to enable more concise JSON serialization of repeated objects. Enabling this behavior for various types is left to future PRs.

For comparison, consider MatrixGate:
```
a, b, c = cirq.LineQubit.range(3)
op = cirq.MatrixGate(some_big_matrix)
circuit = cirq.Circuit(op.on(a), op.on(b), op.on(c))
cirq.to_json(circuit)
```

Without concise serialization for MatrixGate:
<details>

```
{
  "cirq_type": "Circuit",
  "moments": [
    {
      "cirq_type": "GateOperation",
      "gate": {
        "cirq_type": "MatrixGate",
        "matrix": some_big_matrix
      }
      "qubits": a
    }
    {
      "cirq_type": "GateOperation",
      "gate": {
        "cirq_type": "MatrixGate",
        "matrix": some_big_matrix
      }
      "qubits": b
    }
    {
      "cirq_type": "GateOperation",
      "gate": {
        "cirq_type": "MatrixGate",
        "matrix": some_big_matrix
      }
      "qubits": c
    }
  ]
}
```
</details>

With concise serialization for MatrixGate:

<details>

```
# If MatrixGate implements SerializableByKey, this becomes:
{
  "cirq_type": "_ContextualSerialization",
  "context_list": [
    {
      "cirq_type": "_SerializedContext",
      "key": op._serialization_key_()
      "value": {
        "cirq_type": "MatrixGate",
        "matrix": some_big_matrix
      },
    },
    {
      "cirq_type": "Circuit",
      "moments": [
        {
          "cirq_type": "GateOperation",
          "gate": {
            "cirq_type": "_SerializedKey",
            "key": op._serialization_key_()
          },
          "qubits": a
        },
        {
          "cirq_type": "GateOperation",
          "gate": {
            "cirq_type": "_SerializedKey",
            "key": op._serialization_key_()
          },
          "qubits": b
        },
        {
          "cirq_type": "GateOperation",
          "gate": {
            "cirq_type": "_SerializedKey",
            "key": op._serialization_key_()
          },
          "qubits": c
        },
      ]
    }
  ]
}
```
</details>

Note that `some_big_matrix` appears three times in the current serialization, but only once with concise serialization.

### DIFF
--- a/cirq/protocols/json_serialization.py
+++ b/cirq/protocols/json_serialization.py
@@ -170,6 +170,10 @@ def _cirq_class_resolver_dictionary() -> Dict[str, ObjectFactory]:
         'YYPowGate': cirq.YYPowGate,
         'ZPowGate': cirq.ZPowGate,
         'ZZPowGate': cirq.ZZPowGate,
+        # internal contextual-serialization types
+        '_ContextualSerialization': _ContextualSerialization,
+        '_SerializedContext': _SerializedContext,
+        '_SerializedKey': _SerializedKey,
         # not a cirq class, but treated as one:
         'pandas.DataFrame': pd.DataFrame,
         'pandas.Index': pd.Index,
@@ -408,7 +412,7 @@ class CirqEncoder(json.JSONEncoder):
         return super().default(o)  # coverage: ignore
 
 
-def _cirq_object_hook(d, resolvers: Sequence[JsonResolver]):
+def _cirq_object_hook(d, resolvers: Sequence[JsonResolver], context_map: Dict[str, Any]):
     if 'cirq_type' not in d:
         return d
 
@@ -421,12 +425,147 @@ def _cirq_object_hook(d, resolvers: Sequence[JsonResolver]):
             "Could not resolve type '{}' during deserialization".format(d['cirq_type'])
         )
 
+    if cls == _SerializedKey:
+        read_from_context = getattr(cls, 'read_from_context', lambda x, y: None)
+        return read_from_context(context_map, **d)
+
+    if cls == _SerializedContext:
+        update_context = getattr(cls, 'update_context', lambda x, y: None)
+        update_context(context_map, **d)
+        return None
+
+    if cls == _ContextualSerialization:
+        deserialize_with_context = getattr(cls, 'deserialize_with_context', lambda x: None)
+        return deserialize_with_context(**d)
+
     from_json_dict = getattr(cls, '_from_json_dict_', None)
     if from_json_dict is not None:
         return from_json_dict(**d)
 
     del d['cirq_type']
     return cls(**d)
+
+
+class SerializableByKey(SupportsJSON):
+    """Interface for objects that can be serialized to a key + context."""
+
+    @doc_private
+    def _serialization_key_(self):
+        pass
+
+
+class _SerializedKey(SupportsJSON):
+    """Internal object for holding a SerializableByKey key.
+
+    This is a private type used in contextual serialization. Its deserialization
+    is context-dependent, and is not expected to match the original; in other
+    words, `cls._from_json_dict_(obj._json_dict_())` does not return
+    the original `obj` for this type.
+    """
+
+    def __init__(self, obj: SerializableByKey):
+        self.key = obj._serialization_key_()
+
+    def _json_dict_(self):
+        return obj_to_dict_helper(self, ['key'])
+
+    @classmethod
+    def _from_json_dict_(cls, **kwargs):
+        raise TypeError(f'Internal error: {cls} should never deserialize with _from_json_dict_.')
+
+    @classmethod
+    def read_from_context(cls, context_map, key, **kwargs):
+        return context_map[key]
+
+
+class _SerializedContext(SupportsJSON):
+    """Internal object for a single SerializableByKey key-to-object mapping.
+
+    This is a private type used in contextual serialization. Its deserialization
+    is context-dependent, and is not expected to match the original; in other
+    words, `cls._from_json_dict_(obj._json_dict_())` does not return
+    the original `obj` for this type.
+    """
+
+    def __init__(self, obj: SerializableByKey):
+        self.key = obj._serialization_key_()
+        self.obj = obj
+
+    def _json_dict_(self):
+        return obj_to_dict_helper(self, ['key', 'obj'])
+
+    @classmethod
+    def _from_json_dict_(cls, **kwargs):
+        raise TypeError(f'Internal error: {cls} should never deserialize with _from_json_dict_.')
+
+    @classmethod
+    def update_context(cls, context_map, key, obj, **kwargs):
+        context_map.update({key: obj})
+
+
+class _ContextualSerialization(SupportsJSON):
+    """Internal object for serializing an object with its context.
+
+    This is a private type used in contextual serialization. Its deserialization
+    is context-dependent, and is not expected to match the original; in other
+    words, `cls._from_json_dict_(obj._json_dict_())` does not return
+    the original `obj` for this type.
+    """
+
+    def __init__(self, obj: Any):
+        self.context_list = []
+        context_keys = set()
+        for sbk in get_serializable_by_keys(obj):
+            new_sc = _SerializedContext(sbk)
+            if new_sc.key not in context_keys:
+                self.context_list.append(new_sc)
+                context_keys.add(new_sc.key)
+        self.context_list += [obj]
+
+    def _json_dict_(self):
+        return obj_to_dict_helper(self, ['context_list'])
+
+    @classmethod
+    def _from_json_dict_(cls, **kwargs):
+        raise TypeError(f'Internal error: {cls} should never deserialize with _from_json_dict_.')
+
+    @classmethod
+    def deserialize_with_context(cls, context_list, **kwargs):
+        return context_list[-1]
+
+
+def has_serializable_by_keys(obj: Any) -> bool:
+    """Returns true if obj contains one or more SerializableByKey objects."""
+    if hasattr(obj, '_serialization_key_'):
+        return True
+    json_dict = getattr(obj, '_json_dict_', lambda: None)()
+    if isinstance(json_dict, Dict):
+        return any(has_serializable_by_keys(v) for v in json_dict.values())
+    return False
+
+
+def get_serializable_by_keys(obj: Any) -> List[SerializableByKey]:
+    """Returns all SerializableByKeys contained by obj.
+
+    Objects are ordered such that nested objects appear before the object they
+    are nested inside. This is required for hooks in json.load().
+    """
+    result = []
+    if hasattr(obj, '_serialization_key_'):
+        result.append(obj)
+    json_dict = getattr(obj, '_json_dict_', lambda: None)()
+    if isinstance(json_dict, Dict):
+        for v in json_dict.values():
+            result = get_serializable_by_keys(v) + result
+    if result:
+        return result
+
+    # Handle primitive container types.
+    if isinstance(obj, Dict):
+        return [sbk for pair in obj.items() for sbk in get_serializable_by_keys(pair)]
+    if hasattr(obj, '__iter__') and not isinstance(obj, str):
+        return [sbk for v in obj for sbk in get_serializable_by_keys(v)]
+    return []
 
 
 # pylint: disable=function-redefined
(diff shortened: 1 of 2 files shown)


cirq/protocols/json_serialization.py
```diff
@@ -429,6 +446,128 @@ def _cirq_object_hook(d, resolvers: Sequence[JsonResolver]):
...
+
+
+def get_serializable_by_keys(obj: Any) -> List[SerializableByKey]:
+    """Returns all SerializableByKeys contained by obj.
+
+    Objects are ordered such that nested objects appear before the object they
+    are nested inside. This is required for hooks in json.load().
+    """
+    result = []
+    if hasattr(obj, '_serialization_key_'):
```
balopat, line 554: While this works as a single round of simplification, and especially when SBK objects are nested under non-SBK objects. However, it won't really work when SBK objects are nested? Not sure how to work this out, or how frequent this will be but that case the context will contain the potential structural repetitions - e.g.:


<details>

```
{
  "cirq_type": "_ContextualSerialization",
  "context_list": [
    {
      "cirq_type": "_SerializedContext",
      "key": "a",
      "value": {
        "cirq_type": "CircuitOperation",
        "ops": [
          {"cirq_type":  "X"},
          {
            "cirq_type": "CircuitOperation",
            "ops": ["big list of operations"]
          }
        ]
      }
    },
    {
      "cirq_type": "_SerializedContext",
      "key": "b",
      "value": {
        "cirq_type": "CircuitOperation",
        "ops": [
          {"cirq_type":  "Y"},
          {
            "cirq_type": "CircuitOperation",
            "ops": ["big list of operations"]
          }
        ]
      }
    },
    {
      "cirq_type": "_SerializedContext",
      "key": "c",
      "value": {
        "cirq_type": "CircuitOperation",
        "ops": [
          {"cirq_type":  "Z"},
          {
            "cirq_type": "CircuitOperation",
            "ops": ["big list of operations"]
          }
        ]
      }
    },
    {
      "cirq_type": "Circuit",
      "moments": [
        {
          "cirq_type": "GateOperation",
          "gate": {
            "cirq_type": "_SerializedKey",
            "key": "a",
          },
          "qubits": a
        },
        {
          "cirq_type": "GateOperation",
          "gate": {
            "cirq_type": "_SerializedKey",
            "key": "b",
          },
          "qubits": b
        },
        {
          "cirq_type": "GateOperation",
          "gate": {
            "cirq_type": "_SerializedKey",
            "key": "c",
          },
          "qubits": c
        }
      ]
    }
  ]
}
```

</details>

balopat (changes requested):
Thanks for this! Added some comments!

cirq/protocols/json_serialization_test.py
```diff
@@ -340,13 +340,23 @@ def test_mutually_exclusive_blacklist():
...
+
+
 def _find_classes_that_should_serialize() -> Set[Tuple[str, Optional[type]]]:
     result: Set[Tuple[str, Optional[type]]] = set()
     result.update(_get_all_public_classes(cirq))
     result.update(_get_all_public_classes(cirq.google))
     result.update(_get_all_public_classes(cirq.work))
 
     for k, v in json_serialization._cirq_class_resolver_dictionary().items():
+        if k in INTERNAL_SERIALIZATION_TYPES:
```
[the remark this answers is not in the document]
95-martin-orion: IIRC, I tried that and it broke `test_shouldnt_be_serialized_no_superfluous`.

cirq/protocols/json_serialization.py
```diff
@@ -468,6 +607,32 @@ def to_json(
             party classes, prefer adding the _json_dict_ magic method
             to your classes rather than overriding this default.
     """
+    if cls == CirqEncoder and has_serializable_by_keys(obj):
```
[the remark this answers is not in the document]
95-martin-orion: It will in theory, but mypy complained about using `cls` as a base type for a class. I can try a bit more wrangling here if we're interested in getting this feature for other encoders (although what other encoders do we use?)

cirq/protocols/json_serialization.py
95-martin-orion: Note that we do not immediately return `obj`. If the SBK has a `_json_dict_` method - and it must, since SBK is a `SupportsJSON` type - then all nested SBKs will be populated into the context as well.

This behavior is tested in `test_context_serialization`.

cirq/protocols/json_serialization.py
```diff
@@ -429,6 +446,128 @@ def _cirq_object_hook(d, resolvers: Sequence[JsonResolver]):
...
+
+    def __init__(self, obj: SerializableByKey):
+        self.key = obj._serialization_key_()
+
+    def _json_dict_(self):
+        return obj_to_dict_helper(self, ['key'])
+
+    @classmethod
+    def _from_json_dict_(cls, **kwargs):
+        raise TypeError(f'Internal error: {cls} should never deserialize with _from_json_dict_.')
```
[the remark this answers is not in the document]
95-martin-orion: From the `SupportsJSON` docstring:
```
    The magic method _json_dict_ must return a trivially json-serializable
    type or other objects that support the SupportsJSON protocol.
```
As `_SerializedKey` is used to replace other objects contained within existing `SupportsJSON` types, it must also support that protocol. (Note that `_from_json_dict_` is **not** part of the `SupportsJSON` protocol - it's just one way to specify how a class should deserialize.)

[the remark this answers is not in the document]
balopat: Got it, thanks for the explanation. As I looked into it, `_from_json_dict_` is part of the whole json protocol, though not the `SupportsJSON` python protocol per se :) If a class has it, it overrides the default behavior (`cls(**d)`) to convert kv pairs to an object. Given that this type should never be deserialized directly, I think this is basically an assert statement for Cirq developers and contributors - maybe users as well. And the `TypeError` is correct as well for this situation - which is developer error.

cirq/protocols/json_serialization.py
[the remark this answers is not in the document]
balopat: Well, people could extend the CirqEncoder to their liking and pass it in. At least that's what the interface is saying. Do we depend on the CirqEncoder anywhere in the SerializableKey design? I don't think so - in which case introducing business logic just to pacify mypy seems wrong - we should just add `# type: ignore` in case we can't get around it.

cirq/protocols/json_serialization_test.py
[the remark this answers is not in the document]
balopat: Hmmm...after looking into this I think you really don't need to add these three classes to the resolver dictionary. 
We could just directly use the names in the object hook where they are used and already have an if statement anyway checking for the class types! Then you can get rid of these as well.

cirq/protocols/json_serialization.py
```diff
@@ -421,6 +425,19 @@ def _cirq_object_hook(d, resolvers: Sequence[JsonResolver]):
             "Could not resolve type '{}' during deserialization".format(d['cirq_type'])
         )
 
+    if cls == _SerializedKey:
+        read_from_context = getattr(cls, 'read_from_context', lambda x, y: None)
```
balopat, line 429: these could be simplified to `return _SerializedKey.read_from_context(context_map, **d)`, right? Why are we not sure that the `_SerializedKey` privte class will have a method?

cirq/protocols/json_serialization.py
```diff
@@ -421,6 +425,19 @@ def _cirq_object_hook(d, resolvers: Sequence[JsonResolver]):
...
+        return read_from_context(context_map, **d)
+
+    if cls == _SerializedContext:
+        update_context = getattr(cls, 'update_context', lambda x, y: None)
```
balopat, line 433: same here, can use the method directly

cirq/protocols/json_serialization.py
```diff
@@ -421,6 +425,19 @@ def _cirq_object_hook(d, resolvers: Sequence[JsonResolver]):
...
+        update_context(context_map, **d)
+        return None
+
+    if cls == _ContextualSerialization:
+        deserialize_with_context = getattr(cls, 'deserialize_with_context', lambda x: None)
```
balopat, line 438: same here, can use the method directly

cirq/protocols/json_serialization.py
```diff
@@ -429,6 +446,133 @@ def _cirq_object_hook(d, resolvers: Sequence[JsonResolver]):
...
+
+    def __init__(self, obj: Any):
+        self.context_list = []
+        context_keys = set()
+        for sbk in get_serializable_by_keys(obj):
+            new_sc = _SerializedContext(sbk)
+            if new_sc.key not in context_keys:
+                self.context_list.append(new_sc)
+                context_keys.add(new_sc.key)
+        self.context_list += [obj]
```
balopat, line 528: Some musings on the naming: I find using the last slot of the context_list confusing a bit for the final object. I looked at a printed json and my mental model after all these names was "context + object" - but then we use _context_list_ that _contains_ the object. So I had to look again, and "aha" I remembered then...

Another mental model could be the "Object Definition DAG" which points towards an ordered list of dependency graph one could be to call it `object_dag` or `objects`. 

Naming it `context` makes it something "meta" instead of the actual objects themselves.

cirq/protocols/json_serialization.py
```diff
@@ -421,6 +425,19 @@ def _cirq_object_hook(d, resolvers: Sequence[JsonResolver]):
...
+
+    if cls == _SerializedContext:
+        update_context = getattr(cls, 'update_context', lambda x, y: None)
+        update_context(context_map, **d)
+        return None
+
+    if cls == _ContextualSerialization:
+        deserialize_with_context = getattr(cls, 'deserialize_with_context', lambda x: None)
+        return deserialize_with_context(**d)
+
```
balopat, line 440: What if we actually instead of making `_from_json_dict_` throw an error, we would just make it do `update_context`, `deserialize_with_context`, etc.? Why not?

balopat: If we use `_from_json_dict` then it makes sense to keep these classes in the resolvercache..

cirq/protocols/json_serialization_test.py
```diff
@@ -433,6 +443,109 @@ def test_sympy():
...
+    sbki_list = SBKImpl('sbki_list', data_list=[sbki_empty, sbki_empty])
+    assert_json_roundtrip_works(sbki_list, resolvers=test_resolvers)
+
+    sbki_tuple = SBKImpl('sbki_tuple', data_tuple=(sbki_list, sbki_list))
+    assert_json_roundtrip_works(sbki_tuple, resolvers=test_resolvers)
+
+    sbki_dict = SBKImpl('sbki_dict', data_dict={'a': sbki_tuple, 'b': sbki_tuple})
+    assert_json_roundtrip_works(sbki_dict, resolvers=test_resolvers)
+
+    sbki_json = str(cirq.to_json(sbki_dict))
```
balopat, line 507: I just tested with `cirq.to_json([sbki_dict])` - i.e. wrapping an SBK obj in an array - and the whole context based approach disappears - i.e. you'll get a full blown out version of SBKImpl objects. Is that intended?

cirq/protocols/json_serialization.py
balopat: Got it - I did misread result.append as return lol :D

cirq/protocols/json_serialization.py
95-martin-orion: This is another mypy-wrangling line: `cls == _SerializedKey` is apparently insufficient for mypy to understand that `cls.read_from_context` is a valid function.

The alternative is to call `_SerializedKey.read_from_context` directly, which is what I'm opting for along with removing these types from the default resolver.

cirq/protocols/json_serialization.py
95-martin-orion: I'll change this name to be more accurate.

For clarity: the reason that this _must_ be a list (rather than separate `context` and `object` fields) is to ensure that `read_json` parses context information before the final object. Separate fields could be arbitrarily reordered by JSON; thus the list-plus-object format.

cirq/protocols/json_serialization.py
95-martin-orion: My personal sense is that implementing this behavior in `_from_json_dict_` violates an implicit contract of the method - specifically, that the following code returns the original object:
```
def json_roundtrip(obj):
    cls = type(obj)
    return cls._from_json_dict_(object._json_dict_())
```
While we probably could get away with this since these are purely internal types, I prefer making the distinction explicit so as not to confuse future work in this area.

cirq/protocols/json_serialization_test.py
95-martin-orion: Good catch! I fixed get_SBKs but failed to update has_SBKs along with it. Will add a fix + tests.

cirq/protocols/json_serialization_test.py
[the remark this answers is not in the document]
95-martin-orion: Done.

cirq/protocols/json_serialization.py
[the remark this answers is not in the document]
95-martin-orion: Opted for `# type: ignore`. A brief glance at mypy Github issues seemed to suggest that this is a known issue.

cirq/protocols/json_serialization.py
95-martin-orion: Chose `object_dag` and added a clarifying comment.

balopat (approved):

cirq/protocols/json_serialization.py
balopat: Sounds good - I see your point. Let's leave them then for clarity.

PR #3604, opened 2020-12-11
Fix quantum volume notebooks

### LINKED ISSUE #3501: Fix quantum volume notebooks
They are broken in multiple ways. The API seem to have changed under them.

mpharrigan:
fixes #3501


balopat (approved):
Thanks for the fix!!! Can you remove the skip line from dev_tools/notebook_tests.py as well? 
Here https://github.com/quantumlib/Cirq/blob/905083cfe74a71e8cbcf5238ce4156c65f4db56d/dev_tools/notebook_test.py#L30-L32

mpharrigan:
Done; I don't understand the test failure

mpharrigan:
Now it's a different error.

 1. The notebook test failures are impossible to read. The bash color codes don't get rendered right and the rest of it is in JSON / html
 2. Why can't it import cirq? Presumably we install that before running the notebook and/or the notebook has a pip install cell

balopat:
> Now it's a different error.
> 
> 1. The notebook test failures are impossible to read. The bash color codes don't get rendered right and the rest of it is in JSON / html

Yes, agreed, to be improved. I think we should be able to output this json into a notebook that then could be previewed within github actions, or be downloadable, or even better if we can make the error more readable and colored within the log. Opened https://github.com/quantumlib/Cirq/issues/3620. 

> 2. Why can't it import cirq? Presumably we install that before running the notebook and/or the notebook has a pip install cell

This is exactly the kind of error the devsite pipelines caught before and were not an issue if you ran a notebook in a virtualenv with cirq installed. One the path for cirq installation (which is taken here), you'll need to run an `import cirq` before the first usage of it. Which in this case it seems like from the logs (which I agree is hard to find) is this line:

```
"qubits = cirq.GridQubit.rect(3,2)\n",
```

PR #3610, opened 2020-12-15
Add measurement support to `ActOnStabilizerChFormArgs` and use it in `MeasurementGate`

smitsanghavi:
#2423

### DIFF
--- a/cirq/sim/clifford/act_on_stabilizer_ch_form_args.py
+++ b/cirq/sim/clifford/act_on_stabilizer_ch_form_args.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Any, Iterable, TYPE_CHECKING
+from typing import Any, Dict, Iterable, TYPE_CHECKING
 
 import numpy as np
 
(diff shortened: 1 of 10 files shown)


smitsanghavi:
Friendly Ping!

cirq/sim/clifford/act_on_stabilizer_ch_form_args.py
```diff
@@ -34,16 +34,41 @@ class ActOnStabilizerCHFormArgs:
     Measurements are currently not supported on this object.
```
balopat, line 34: Measurements are supported from this point onwards! :)

cirq/sim/clifford/act_on_stabilizer_ch_form_args.py
```diff
@@ -34,16 +34,41 @@ class ActOnStabilizerCHFormArgs:
...
     """
 
-    def __init__(self, state: StabilizerStateChForm, axes: Iterable[int]):
+    def __init__(
+        self,
+        state: StabilizerStateChForm,
+        axes: Iterable[int],
+        prng: np.random.RandomState,
```
balopat, line 41: This could be default `np.random.RandomState()` - why wouldn't we set it?

balopat (approved):
Thanks Smit! LGTM - with two nits

cirq/sim/clifford/act_on_stabilizer_ch_form_args.py
smitsanghavi: So true!

cirq/sim/clifford/act_on_stabilizer_ch_form_args.py
smitsanghavi: Honestly, I was just following the example set in `ActOnStateVectorArgs`. I think the reasoning is that, except for tests, the simulator that is using this object should be supplying the prng. So making this arg optional may lead to easier mistake where the seeded prng of the simulator is not being passed through.

balopat: Sounds good!

balopat (approved):

balopat:
Notebook failure is due to https://github.com/quantumlib/Cirq/issues/3625

PR #3616, opened 2020-12-17
Fix superflous pip install in simulation.ipynb

mpharrigan:
It's done above with an import guard


95-martin-orion:
Pretty sure this has already been removed on the `master` branch.

vtomole:
@95-martin-orion Yup.

mpharrigan:
oh i guess i had a stale branch

PR #3622, opened 2020-12-20
Replace isinstance usages with op.gate

dwillmer:
- First attempt at fixing #3556

### DIFF
--- a/cirq/circuits/circuit_test.py
+++ b/cirq/circuits/circuit_test.py
@@ -1252,7 +1252,7 @@ def test_findall_operations(circuit_cls):
     zb = cirq.Z.on(b)
 
     def is_x(op: cirq.Operation) -> bool:
-        return isinstance(op, cirq.GateOperation) and isinstance(op.gate, cirq.XPowGate)
+        return op.gate is not None and isinstance(op.gate, cirq.XPowGate)
 
     c = circuit_cls()
     assert list(c.findall_operations(is_x)) == []
--- a/cirq/circuits/qasm_output.py
+++ b/cirq/circuits/qasm_output.py
@@ -311,7 +311,7 @@ class QasmOutput:
             should_annotate = decomposed != [main_op]
             if should_annotate:
                 output_line_gap(1)
-                if isinstance(main_op, ops.GateOperation):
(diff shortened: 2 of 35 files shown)


cirq/circuits/qasm_output.py
```diff
@@ -311,7 +311,7 @@ def on_stuck(bad_op):
             should_annotate = decomposed != [main_op]
             if should_annotate:
                 output_line_gap(1)
-                if isinstance(main_op, ops.GateOperation):
+                if main_op.gate is not None:
```
wcourtney, line 314: Might this be a problem if a new Operation is added that happens to have a gate subfield but is not a subtype of GateOperation? I would expect the default handling to pick up in this case.

wcourtney, line 314: This changes behavior if `main_op` is `None` since `isinstance(None, SomeClass)` is `false` but dereferencing `None` will cause an `AttributeError`.

Consider adding a test to ensure/document that this is handled as expected.

cirq/circuits/circuit_test.py
```diff
@@ -1252,7 +1252,7 @@ def test_findall_operations(circuit_cls):
     zb = cirq.Z.on(b)
 
     def is_x(op: cirq.Operation) -> bool:
-        return isinstance(op, cirq.GateOperation) and isinstance(op.gate, cirq.XPowGate)
+        return op.gate is not None and isinstance(op.gate, cirq.XPowGate)
```
maffoo, line 1255: Can remove the first check since `None` is not an instance of `cirq.XPowGate`:
```suggestion
        return isinstance(op.gate, cirq.XPowGate)
```
This applies generally throughout this PR; if we're just checking whether `op.gate` is an instance of some class, there's no need to first check whether it is `None`.

vtomole:
Ping @dwillmer

dabacon:
Hi @dwillmer 

Do you have time to address issues raised in this code review?  It looks like this PR has been open for a year and it would be good to get it in.

dabacon:
Closing for now as this appears to be stale.  Please reopen if you want to revisit.

PR #3630, opened 2020-12-28
An MPS (approximate) simulator

tonybruguier:
This is based on this algo:
https://arxiv.org/abs/2002.07730


tonybruguier:
Hello,

I read more of the paper, and I think I understand how to do the 2D version (famous last words).

In the meantime, is there any comment on this PR? It would be nice to have some feedback on 1D before I attempted 2D.

cirq/sim/mps_simulator.py
```diff
@@ -0,0 +1,350 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""An MPS simulator.
+
+This is based on this paper:
+https://arxiv.org/abs/2002.07730
+
+TODO(tonybruguier): Currently, only linear circuits are handled, while the paper
+handles more general topologies.Currently
```
95-martin-orion, line 20: Extra "Currently"

cirq/sim/mps_simulator_test.py
```diff
@@ -0,0 +1,196 @@
...
+    ref_simulator = cirq.Simulator()
+
+    actual = mps_simulator.simulate(circuit, qubit_order=qubit_order, initial_state=initial_state)
+    expected = ref_simulator.simulate(circuit, qubit_order=qubit_order, initial_state=initial_state)
+    np.testing.assert_almost_equal(actual.final_state.to_numpy(), expected.final_state_vector)
+    assert len(actual.measurements) == 0
+
+
+def test_various_gates():
+    gate_cls = [cirq.I, cirq.H, cirq.X, cirq.Y, cirq.Z]
```
95-martin-orion, line 21: Please add `cirq.T` to this set to make it universal.

cirq/sim/mps_simulator_test.py
```diff
@@ -0,0 +1,196 @@
...
+    result_string = ''.join(map(lambda x: str(int(x[0])), measured))
+    assert result_string == '01011001110111011011'
+
+
+def test_run_no_repetitions():
+    q0 = cirq.LineQubit(0)
+    simulator = cirq.MPSSimulator()
+    circuit = cirq.Circuit(cirq.H(q0), cirq.measure(q0))
+    result = simulator.run(circuit, repetitions=0)
+    assert sum(result.measurements['0']) == 0
```
95-martin-orion, line 188: Could we instead test `len(...) == 0`? Using `sum` seems like it could hide strange behavior or cause flakes - for example, if `repetitions=0` actually ran one repetition, but the measurement result was a zero.

cirq/sim/mps_simulator.py
```diff
@@ -0,0 +1,350 @@
...
+
+    def perform_measurement(
+        self, qubits: Sequence[ops.Qid], prng: np.random.RandomState, collapse_state_vector=True
+    ) -> List[int]:
+        results: List[int] = []
+
+        if collapse_state_vector:
+            state = self
+        else:
+            state = self.copy()
```
95-martin-orion, line 320: To confirm: this copy is required for non-collapsing measurement because `state` must be mutated to capture how each qubit measurement affects the rest (e.g. measuring `0` on the first qubit of state `|00) + |11)` forces a `0` on the second). Is this correct?

cirq/sim/mps_simulator.py
```diff
@@ -0,0 +1,350 @@
...
+            state: A MPSState
+            measurements: A dictionary from measurement gate key to measurement
+                results, ordered by the qubits that the measurement operates on.
+            qubit_map: A map from the Qubits in the Circuit to the the index
+                of this qubit for a canonical ordering. This canonical ordering
+                is used to define the state vector (see the state_vector()
+                method).
+        """
+        self.measurements = measurements
+        self.state = state
```
95-martin-orion, line 195: If this result represents an intermediate state of a circuit, could mutating the `state` of this object affect subsequent steps in the simulation? Compare PR #3109 and linked issues.

95-martin-orion (changes requested):
CC'ing @rmlarose for review / comparison with [external simulator](https://github.com/grmlarose/mpsim).

Thanks for this addition! A couple of changes and clarifications needed, but overall the test coverage looks promising.

cirq/sim/mps_simulator.py
tonybruguier: Removed.

cirq/sim/mps_simulator_test.py
tonybruguier: Done.

cirq/sim/mps_simulator_test.py
tonybruguier: Done.

cirq/sim/mps_simulator.py
tonybruguier: I am not sure, actually. I was just inspiring myself from this existing code:
https://sourcegraph.com/github.com/quantumlib/Cirq@9bad962c32a236ba72d829e5091d3bbc7711814c/-/blob/cirq/sim/clifford/clifford_simulator.py#L360

Which of course, could be wrong, or (more likely) I don't understand what I'm doing.

Leaving unresolved.

cirq/sim/mps_simulator.py
tonybruguier: I think so. There's test_simulate_moment_steps_sample that makes sure that this is the intended behavior and I was inspiring myself from this:
https://sourcegraph.com/github.com/quantumlib/Cirq@9bad962c32a236ba72d829e5091d3bbc7711814c/-/blob/cirq/sim/clifford/clifford_simulator.py#L360

Leaving unresolved.

95-martin-orion: Sounds good to me. If I understand it correctly, the MPS simulator has much smaller "state" objects (similar to the CliffordSimulator) so this is less of a concern.

cirq/sim/mps_simulator.py
95-martin-orion: I think the best path to resolving this is to add a test similar to the one in #3109, where you attempt to mutate an intermediate state and confirm that the final state isn't also mutated. If you find that it is mutated, `CliffordSimulator` should be updated as well.

Given that this touches code outside the scope of this PR, I'd also be fine with this being addressed in a follow-up PR.

tonybruguier: There was indeed a problem. I think that the Clifford simulator also had the same issue so instead of a TODO, I decided to send you another PR: #3664

cirq/sim/mps_simulator.py
tonybruguier: I think the size of the state here depends on the degree of entanglement, but if it's too big, then it probably doesn't make sense to use the proposed simulator in the first place.

tonybruguier:
Thanks. PTAL.

95-martin-orion (approved):

95-martin-orion:
@tonybruguier I'm ready to merge this PR if you're happy with it - let me know.

tonybruguier:
Yes, please merge if you think it's fine to do so. Thank you for your help!

rmlarose:
Thanks Orion for the ping! I still haven't had time to look closely but will offer a few cursory comments if they're useful.

I think the simulator [`threshold`](https://github.com/quantumlib/Cirq/pull/3630/files#diff-bb4cae8ea7b8ba334cee70c5a41f070700e7e27848e11acb29095b45a0cb56bdR254) is set internally at `1e-3`? If so, it could be good to let this value be a simulator option, e.g. `MPSSimulator(threshold=...)`. This allows for exact simulation (threshold = 0 <==> keep all singular values) in addition to "noisy" simulation (threshold > 0 <==> small singular values are thrown away).

I'm not sure in which order gates are applied, but it could be good to sweep from left to right (or right to left) to keep some canonicalization. This may not be necessary at present but certain operations are faster with canonical MPS.

Also not relevant now but maybe to future proof: this should support the "SimulatesExpectationValues" protocol when that is finalized since MPS are great for computing expectation values.

PR #3637, opened 2021-01-04
Fix for issue #1622: Some QASM annotations missing

### LINKED ISSUE #1622: For QASM of decomposed circuits add comment
I noticed that if I export CCZ into qasm, I get the h, ccx, h operations with no leading comment saying that they correspond to a CCZ. I'm pretty sure this used to happen.

daxfohl:
Fixes #1622

This PR updates QASM generator to annotate any operation that results in multiple QASM lines.

The other options to fix this bug include: update the QASM generator for each operation independently, or to update the decompositions themselves. The first option would have required each QASM generator to have its own annotation string, so we'd lose the guaranteed consistency with the generated annotations. The second option seemed nice, but ultimately QASM representation and decomposition are two different things: ccx can be represented in QASM but decomposes further.


cirq/circuits/qasm_output.py
```diff
@@ -308,7 +308,11 @@ def on_stuck(bad_op):
                 main_op, keep=keep, fallback_decomposer=fallback, on_stuck_raise=on_stuck
             )
 
-            should_annotate = decomposed != [main_op]
+            qasms = list(
```
mpharrigan, line 311: idiomatic python rarely uses `list(map(lambda`. Consider writing as a list comprehension

daxfohl: Fixed in latest commit

balopat (approved):
LGTM, thanks for the fix!!!

PR #3644, opened 2021-01-06
Simplify recursive sweepable cast

### LINKED ISSUE #3297: Sweepable is actually a recursive type
The current type signature of Sweepable is
```
Sweepable = Union[Dict[str, float], ParamResolver, Sweep, Iterable[
    Union[Dict[str, float], ParamResolver, Sweep]], None]
```
But there is no logical reason to only allow one level of iteration over Sweep. And in fact, `cirq.to_sweeps` is a recursive function, so it already assumes that Sweepable is a recursive type. So as it stands, Sweepable should be defined as a recursive type.

But I would like to ask whether Sweepable really needs to be this recursive type, rather than just something that can be converted to a single Sweep. Do people actually use lists of sweeps, rather than just plain sweeps? It seems like a rather confusing thing to attempt, given that one would need to keep track of how results correspond to the individual parameter resolvers.

daxfohl:
Fixes #3297

Simplifies the type of the recursive `to_sweep` call.


cirq/study/sweepable.py
```diff
@@ -16,17 +16,23 @@
...
 
 from cirq._doc import document
 from cirq.study.resolver import ParamResolver, ParamResolverOrSimilarType
 from cirq.study.sweeps import ListSweep, Points, Sweep, UnitSweep, Zip, dict_to_product_sweep
 
-SweepLike = Union[ParamResolverOrSimilarType, Sweep]
-document(SweepLike, """An object similar to an iterable of parameter resolvers.""")  # type: ignore
 
-Sweepable = Union[SweepLike, Iterable[SweepLike]]
+class SweepLike(Protocol):
```
kevinsung, line 26: We want to make `Sweepable` recursive, not `SweepLike`. So the definition of `SweepLike` should stay as it was, and this class should be called `Sweepable`. Then we should rename the actual type to `SWEEPABLE` so the name doesn't conflict. Basically, you can copy the the `OP_TREE` template with the following substitutions:
```
Operation -> SweepLike
OP_TREE -> SWEEPABLE
```

kevinsung (changes requested):

cirq/study/sweepable.py
```diff
@@ -16,17 +16,27 @@
...
 
 SweepLike = Union[ParamResolverOrSimilarType, Sweep]
-document(SweepLike, """An object similar to an iterable of parameter resolvers.""")  # type: ignore
+document(SweepLike, """An object similar to an iterable of parameter resolvers.""")
 
-Sweepable = Union[SweepLike, Iterable[SweepLike]]
+
+class SweepLikeRecursive(Protocol):
+    """An intermediate class allowing for recursive definition of Sweepable,
+    since recursive union definitions are not yet supported in mypy."""
```
kevinsung, line 31: I agree with not changing the name of `Sweepable`. In that case I think this class should just be named `_Sweepable` to make its purpose clearer.

daxfohl: Sounds good, I should have that later today.

daxfohl: This is done in latest commit.

balopat (approved):
LGTM, nice. Thanks!

PR #3647, opened 2021-01-07
Observable Measurement - 4 - Sampling Loop

mpharrigan:
- Full prototype #2781
  - Settings #3371 
  - Grouping #3372 
  - Data #3374 
  - Correction #3649 
  - Sampling **this** (split)
    - Pre-reqs #3792 
    - Loop #3855 
    - Stopping #3969 
    - Readout #4323 
    - Checkpointing #4352 
    - High-level API #4392 
  - Top level imports (todo)
  - Examples (todo)

Follows #3374 

Part of #2781

PR #3649, opened 2021-01-08
Observable Measurement - 3.5 - Readout Correction

mpharrigan:
This is part 3.5 or 4.5 of #2781 

It comes either before or after #3647 

This extends the vanilla `BitstringAccumulator` to handle readout calibrations which can be used to correct observables and mitigate the effect of readout error. However, users should probably never use this functionality directly; instead passing the appropriate configuration options to the functions defined in #3647. It would make git dependencies if this is merged first.

More sophisticated readout mitigation schemes would be implemented by subclassing `BitstringAccumulator`. Subclasses would (1) modify the constructor to accept the relevant calibration info, if different than this simple scheme and (2) modify the `mean()` and `variance()` functions to correct those quantities using the new scheme.

### DIFF
--- a/cirq/work/observable_measurement_data.py
+++ b/cirq/work/observable_measurement_data.py
@@ -18,13 +18,14 @@ from typing import Dict, List, Tuple, TYPE_CHECKING
 
 import numpy as np
 
-from cirq import protocols
+from cirq import protocols, ops
 from cirq._compat import proper_repr
 from cirq.work.observable_settings import (
     InitObsSetting,
     _max_weight_observable,
     _max_weight_state,
     _MeasurementSpec,
+    zeros_state,
 )
 
 if TYPE_CHECKING:
@@ -132,6 +133,14 @@ class ObservableMeasuredResult:
         return np.sqrt(self.variance)
 
 
+def _setting_to_z_observable(setting: InitObsSetting):
+    qubits = setting.observable.qubits
+    return InitObsSetting(
+        init_state=zeros_state(qubits),
+        observable=ops.PauliString(qubit_pauli_map={q: ops.Z for q in qubits}),
+    )
+
+
 class BitstringAccumulator:
     """A mutable container of bitstrings and associated metadata populated
     during a `measure_observables` run.
@@ -170,7 +179,15 @@ class BitstringAccumulator:
         readout_calibration:
             The result of `calibrate_readout_error`. When requesting
             means and variances, if this is not None, we will use the
-            calibrated value to correct the requested quantity.
+            calibrated value to correct the requested quantity. This is a
+            BitstringAccumulator containing the results of measuring Z
+            observables with readout symmetrization enabled. This class
+            does *not* validate that both this parameter and the
+            BitstringAccumulator under construction contain measurements taken
+            with readout symmetrization turned on; and we strongly
+            encourage you to avoid initializing this directly. Instead, use
+            `measure_observables` and `calibrate_readout_error`.
+
     """
 
     def __init__(
@@ -187,6 +204,12 @@ class BitstringAccumulator:
         self._simul_settings = simul_settings
         self._qubit_to_index = qubit_to_index
         self._readout_calibration = readout_calibration
+
+        # # Extract and cache <Z> observables from `readout_calibration`.
+        # self._readout_cal_dict = {
+        #     q: readout_calibration.mean(InitObsSetting(cirq.KET_ZERO(q), cirq.Z(q)))
+        #     for q in readout_calibration.max_setting.observable.qubits
+        # }
 
         if bitstrings is None:
             n_bits = len(qubit_to_index)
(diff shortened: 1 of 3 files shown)


mpharrigan:
@viathor @balopat PTAL!

cirq/work/observable_measurement_data.py
```diff
@@ -167,6 +176,18 @@ class BitstringAccumulator:
...
+            The result of `calibrate_readout_error`. When requesting
+            means and variances, if this is not None, we will use the
+            calibrated value to correct the requested quantity. This is a
+            BitstringAccumulator containing the results of measuring Z
+            observables with readout symmetrization enabled. This class
+            does *not* validate that both this parameter and the
+            BitstringAccumulator under construction contain measurements taken
+            with readout symmetrization turned on; and we strongly
+            encourage you to avoid initializing this directly. Instead, use
+            `measure_observables` and `calibrate_readout_error`.
```
balopat, line 189: So, these don't exist yet, right? Is that the dependency on  #3647?

cirq/work/observable_measurement_data.py
```diff
@@ -423,13 +453,28 @@ def variance(self, setting: InitObsSetting, *, atol: float = 1e-8):
...
             qubit_to_index=self._qubit_to_index,
             observable=setting.observable,
             atol=atol,
         )
 
+        if self._readout_calibration is not None:
+            ro_setting = _setting_to_z_observable(setting)
+            a = mean
+            if np.isclose(a, 0):
+                return np.inf  # coverage: ignore
```
balopat, line 467: Do you mind covering/testing the exception cases?

cirq/work/observable_measurement_data.py
```diff
@@ -423,13 +453,28 @@ def variance(self, setting: InitObsSetting, *, atol: float = 1e-8):
...
-        _, var = _stats_from_measurements(
+        mean, var = _stats_from_measurements(
             bitstrings=self.bitstrings,
             qubit_to_index=self._qubit_to_index,
             observable=setting.observable,
             atol=atol,
         )
 
+        if self._readout_calibration is not None:
+            ro_setting = _setting_to_z_observable(setting)
```
balopat, line 464: nit: move this line a bit below, closer to the first use of `ro_setting`

balopat (commented):
I think this looks OK. Added some nits. Also I think we could merge this but not refer to the non-existent methods yet in the docstrings...

cirq/work/observable_measurement_data.py
```diff
@@ -177,10 +198,18 @@ def __init__(
...
         chunksizes: np.ndarray = None,
         timestamps: np.ndarray = None,
+        readout_calibration: 'BitstringAccumulator' = None,
     ):
         self._meas_spec = meas_spec
         self._simul_settings = simul_settings
         self._qubit_to_index = qubit_to_index
+        self._readout_calibration = readout_calibration
+
+        # # Extract and cache <Z> observables from `readout_calibration`.
```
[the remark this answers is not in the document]
mpharrigan: good spot!

cirq/work/observable_measurement_data.py
```diff
@@ -167,6 +176,18 @@ class BitstringAccumulator:
             arise. The total number of repetitions is the sum of this 1d array.
         timestamps: We record a timestamp for each request/chunk. This
             1d array will have the same length as `chunksizes`.
+        readout_calibration:
+            The result of `calibrate_readout_error`. When requesting
+            means and variances, if this is not None, we will use the
```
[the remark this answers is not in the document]
mpharrigan: changed manually

cirq/work/observable_measurement_data.py
```diff
@@ -167,6 +176,18 @@ class BitstringAccumulator:
...
+            calibrated value to correct the requested quantity. This is a
+            BitstringAccumulator containing the results of measuring Z
```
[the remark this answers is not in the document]
mpharrigan: added manually

cirq/work/observable_measurement_data.py
```diff
@@ -167,6 +176,18 @@ class BitstringAccumulator:
...
+            observables with readout symmetrization enabled. This class
+            does *not* validate that both this parameter and the
+            BitstringAccumulator under construction contain measurements taken
```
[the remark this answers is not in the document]
mpharrigan: added manually

cirq/work/observable_measurement_data.py
mpharrigan: removed; will add back in in #3647

cirq/work/observable_measurement_data.py
mpharrigan: added

cirq/work/observable_measurement_data.py
mpharrigan: done

cirq/work/observable_measurement_data.py
```diff
@@ -423,13 +445,28 @@ def variance(self, setting: InitObsSetting, *, atol: float = 1e-8):
...
+            var_a = var
+            ro_setting = _setting_to_z_observable(setting)
+            b = self._readout_calibration.mean(ro_setting)
+            if np.isclose(b, 0, atol=atol):
+                return np.inf
+            var_b = self._readout_calibration.variance(ro_setting)
+            f = a / b
+
+            # assume cov(a,b) = 0, otherwise there would be another term.
+            var = f ** 2 * (var_a / (a ** 2) + var_b / (b ** 2))
```
balopat, line 468: can you add a reference or derivation for this formula?

balopat (approved):
LGTM with a nit for reference

cirq/work/observable_measurement_data.py
mpharrigan: added

PR #3650, opened 2021-01-08
Push redundant simulator iterator code into base classes

### LINKED ISSUE #2178: Reduce required boilerplate for implementing external simulators and samplers
I wrote a custom simulator recently and I had to copy quite a bit of boilerplate code out of cirq/sim/sparse_simulator.py.  Below is my code with the boilerplate pointed out.  It would be nice if the code in SimulatesSamples and SimulatesIntermediateState parent classes were rearranged to reduce this.  @Strilanc

```python
class MySimulator(cirq.SimulatesSamples,
                  cirq.SimulatesIntermediateState):
    # Boilerplate
    def _run(self, circuit, param_resolver, repetitions):
        param_resolver = param_resolver or cirq.ParamResolver({})
        resolved_circuit = cirq.resolve_parameters(circuit, param_resolver)
        assert not cirq.is_parameterized(resolved_circuit)
        return self._run_sweep_repeat(resolved_circuit, repetitions)

    # Boilerplate
    def _run_sweep_repeat(self, circuit, repetitions):
        measurements = defaultdict(list)
        for _ in range(repetitions):
            all_step_results = self._base_iterator(
                circuit,
                qubit_order=cirq.QubitOrder.DEFAULT,
                initial_state=0)
                for k, v in step_result.measurements.items():
                    measurements[k].append(np.array(v, dtype=np.uint8))
        return {k: np.array(v, dtype=np.uint8) for k, v in measurements.items()}

    # Boilerplate
    def _simulator_iterator(self, circuit, param_resolver, qubit_order,
                            initial_state):
        param_resolver = param_resolver or cirq.ParamResolver({})
        resolved_circuit = cirq.resolve_parameters(circuit, param_resolver)
        assert not cirq.is_parameterized(resolved_circuit)
        actual_initial_state = 0 if initial_state is None else initial_state
        return self._base_iterator(resolved_circuit, qubit_order,
                                   actual_initial_state,
                                   perform_measurements=True)

    def _base_iterator(self, circuit, qubit_order, initial_state,
                       perform_measurements=True):
        # Boilerplate
        qubits = cirq.QubitOrder.as_qubit_order(qubit_order).order_for(
            circuit.all_qubits())
        num_qubits = len(qubits)
        qid_shape = cirq.qid_shape(qubits)
        qubit_map = {q: i for i, q in enumerate(qubits)}
        if len(circuit) == 0:
            yield MySimulatorStep(<default state>, {}, qubit_map)
        def on_stuck(bad_op):
            return TypeError(
                "Can't simulate unknown operations that don't specify a"
                "unitary or a decomposition. {!r}".format(bad_op))
        def keep(potential_op):
            return (cirq.has_unitary(potential_op) or
                    cirq.has_mixture(potential_op) or
                    cirq.is_measurement(potential_op) or
                    cirq.op_gate_isinstance(potential_op, cirq.ResetChannel))

        simulator_state = ...  # New code
        for moment in circuit:
            measurements = defaultdict(list)
            known_ops = cirq.decompose(moment, keep=keep,
                                       on_stuck_raise=on_stuck)
            for op in known_ops:
                self.simulate_op(op, simulator_state)  # New code
            yield MySimulatorStep(simulator_state, measurements, qubit_map)

    def simulate_op(op, state):
        # Actual simulation code here
        ...


class MySimulatorStep(cirq.StepResult):
    # Boilerplate
    def __init__(self, state, measurements, qubit_map):
        super().__init__(measurements)
        self.state = state
        self.qubit_map = qubit_map

    # Boilerplate
    def _simulator_state(self):
        return self.state

    def sample(self, qubits, repetitions=1):
        # Actual sampling code here
        return ...
```

daxfohl:
First step for #2178 

This PR pushes redundant simulator code into base classes, making it easier to create simulators. The scope of the PR is currently only the iterator methods. The run methods have subsequently diverged among subclasses, such that the first two lines of each are all that is repetitive, and the overhead of consolidating those is not worth it.

This is still something that could be visited and refined later on. There may be some other higher-level changes we could make to consolidate additional logic. That would be substantial redesign though and is beyond the scope of this PR.


cirq/sim/simulator.py
```diff
@@ -357,6 +356,19 @@ def _simulator_iterator(
         Yields:
             StepResults from simulating a Moment of the Circuit.
         """
+        param_resolver = param_resolver or study.ParamResolver({})
```
95-martin-orion, line 359: The docstring for this method should be expanded to include the details in the previous implementations.

cirq/sim/simulator.py
```diff
@@ -357,6 +356,19 @@ def _simulator_iterator(
...
+        resolved_circuit = protocols.resolve_parameters(circuit, param_resolver)
+        _check_all_resolved(resolved_circuit)
+        actual_initial_state = 0 if initial_state is None else initial_state
+        return self._base_iterator(resolved_circuit, qubit_order, actual_initial_state)
+
+    @abc.abstractmethod
+    def _base_iterator(
```
95-martin-orion, line 366: The docstring for this class should be updated to reflect the new requirement:
```
Implementors of this interface should implement the _base_iterator method.
```

cirq/sim/simulator.py
```diff
@@ -593,3 +605,13 @@ def _verify_unique_measurement_keys(circuit: circuits.Circuit):
         duplicates = [k for k, v in result.most_common() if v > 1]
         if duplicates:
             raise ValueError('Measurement key {} repeated'.format(",".join(duplicates)))
+
+
+def _check_all_resolved(circuit):
```
95-martin-orion, line 610: This method isn't private anymore, so we can drop the leading underscore.

95-martin-orion (changes requested):

cirq/sim/simulator.py
daxfohl: done

cirq/sim/simulator.py
daxfohl: done

cirq/sim/simulator.py
daxfohl: Done, but another question is, would it make sense to inline `_simulator_iterator` into `simulate_moment_steps` now? That is the only caller, and it is a thin wrapper. Doing so would also involve updating the mocks in two unit tests.

95-martin-orion: That would be a breaking change, as (I assume) any external implementations of `SimulatesIntermediateStates` implement the `_simulator_iterator` method. For now it's fine to leave this where it is.

daxfohl:
TBH I'm only 50/50 on this PR now. We reduce a bit of boilerplate for future implementors, but now have two functions that need to be maintained for backwards compatibility instead of one. IDK if that tradeoff is worth it. Thoughts?

cirq/sim/simulator.py
daxfohl: Should we mark it deprecated?

95-martin-orion: That would be good - it would empower us to make the breaking change once the Cirq version advances.

95-martin-orion:
> TBH I'm only 50/50 on this PR now. We reduce a bit of boilerplate for future implementors, but now have two functions that need to be maintained for backwards compatibility instead of one. IDK if that tradeoff is worth it. Thoughts?

If we deprecate `_simulator_iterator`, we'll be able to get rid of it after the next Cirq release. Also, in my experience this code is fairly stable; maintenance costs should be low. I support this PR.

cirq/sim/simulator.py
daxfohl: Added the deprecated attribute. I marked it as v0.11 since that was the highest existing deadline in the project.

daxfohl:
@95-martin-orion I added the deprecated attribute. Note as-is the PR description will close the issue. If you want to leave it open for further improvements then I can change that.

95-martin-orion (approved):
This PR looks good to go from my perspective. Let me know when you're ready and I'll trigger automerge, which will automatically perform any necessary updates to this branch before merging.

95-martin-orion:
> @95-martin-orion I added the deprecated attribute. Note as-is the PR description will close the issue. If you want to leave it open for further improvements then I can change that.

I'd say either leave the current issue open, or close it and open a fresh issue tracking the removal of the deprecated method.

daxfohl:
@95-martin-orion Updated the comment so the issue will not be closed. This can be set to automerge.

PR #3351, opened 2020-09-22
Extend cirq.vis.Heatmap to visualize two-qubit interactions

crystalzhaizhai:
issue #3088
plot a two-qubit interaction between qubits on a grid

PR #3652, opened 2021-01-08
[WIP] Extend cirq.vis.Heatmap to visualize two-qubit interactions (#3351)

### LINKED ISSUE #3088: Extend cirq.vis.Heatmap to visualize two-qubit interactions
Currently cirq.vis.Heatmap supports plotting of single-qubit to float maps.

It would be very useful to be able to plot a two-qubit interaction between qubits on a grid, for example like the plot on Fig. S40 a) (page 43) of the supremacy supplementary paper: https://arxiv.org/abs/1910.11333

Direct application of this could be a visualization of cirq.google.Calibration data (each parameter could just have a plot() method for example).

balopat:
Adds two-qubit interaction heatmap. 
We need to potentially cleanup and merge the #3351 PR that was merged in a branch.

Fixes #3088.


cirq/vis/heatmap.py
```diff
@@ -28,24 +27,28 @@
...
+QubitCoordinate = Union[Tuple[float, float], grid_qubit.GridQubit]
 
-# The value map is qubit coordinate -> a type that supports float conversion.
-ValueMap = Union[Dict[grid_qubit.GridQubit, SupportsFloat], Dict[Tuple[int, int], SupportsFloat]]
+QubitPair = Union[
+    Tuple[grid_qubit.GridQubit, grid_qubit.GridQubit],
+    Tuple[Tuple[float, float], Tuple[float, float]],
+]
 
+Hashable = Union[QubitCoordinate, QubitPair]
```
balopat, line 37: not sure why we need this type, it's not used anywhere

cirq/vis/heatmap.py
```diff
@@ -28,24 +27,28 @@
 
 from cirq.devices import grid_qubit
 
-QubitCoordinate = Union[Tuple[int, int], grid_qubit.GridQubit]
+QubitCoordinate = Union[Tuple[float, float], grid_qubit.GridQubit]
```
balopat, line 30: why did we change this to float, float? The qubits are themselves still on an integer grid!

cirq/vis/heatmap.py
```diff
@@ -28,24 +27,28 @@
...
 
-# The value map is qubit coordinate -> a type that supports float conversion.
-ValueMap = Union[Dict[grid_qubit.GridQubit, SupportsFloat], Dict[Tuple[int, int], SupportsFloat]]
+QubitPair = Union[
```
balopat, line 32: can this just be `Tuple[QubitCoordinate, QubitCoordinate]`?

balopat: I guess keeping the types consistent is expressed here, similarly to `ValueMap` where we keep the key types consistent.

cirq/vis/heatmap.py
```diff
@@ -57,82 +60,81 @@ def relative_luminance(color: np.ndarray) -> float:
...
 
-class Heatmap:
+def _get_qubit_row_col(qubit: QubitCoordinate) -> Tuple[float, float]:
+    if isinstance(qubit, grid_qubit.GridQubit):
+        return qubit.row, qubit.col
+    else:
+        return qubit[0], qubit[1]
+
+
+def _ave_qubit_row_col(qubitpair: QubitPair) -> Tuple[float, float]:
```
balopat, line 70: 1. nit avg is more typical for average
```suggestion
def _avg_qubit_row_col(qubitpair: QubitPair) -> Tuple[float, float]:
```
2.  I'd actually call this is more like "midpoint" - as that's the intention to calculate the mid point between them.

cirq/vis/heatmap_test.py
```diff
@@ -243,3 +243,27 @@ def test_colorbar(ax):
...
+    colormap_name = 'Greys'
+    test_heatmap.set_colormap(0.001, 0.01, colormap_name)
+
+    annot_map = {(3.5, 2): '.3e', (4, 1.5): '.2f'}
+    test_heatmap.unset_annotation()
+    test_heatmap.set_annotation_map(annot_map)
+    test_heatmap.set_annotation_format('.3e')
+
+    test_heatmap.plot()
+    test_heatmap1.plot()
```
balopat, line 269: this test is not testing anything

examples/heatmap_example.py
```diff
@@ -0,0 +1,63 @@
...
+    value_map = {
+        (cirq.GridQubit(3, 2), cirq.GridQubit(4, 2)): 0.004619111460557768,
+        (cirq.GridQubit(4, 1), cirq.GridQubit(4, 2)): 0.0076079162393482835,
+        (cirq.GridQubit(4, 1), cirq.GridQubit(5, 1)): 0.010323903068646778,
+        (cirq.GridQubit(4, 2), cirq.GridQubit(4, 3)): 0.00729037246947839,
+        (cirq.GridQubit(4, 2), cirq.GridQubit(5, 2)): 0.008226663382640803,
+        (cirq.GridQubit(4, 3), cirq.GridQubit(5, 3)): 0.01504682356081491,
+        (cirq.GridQubit(5, 0), cirq.GridQubit(5, 1)): 0.00673880216745637,
+        (cirq.GridQubit(5, 1), cirq.GridQubit(5, 2)): 0.01020380985719993,
+        (cirq.GridQubit(5, 1), cirq.GridQubit(6, 1)): 0.005713058677283056,
```
balopat, line 29: we should use a randomly generated data instead with a seed and then use the image generated based on that

examples/heatmap_example.py
```diff
@@ -0,0 +1,63 @@
+"""The script that produces bristlecone_heatmap_example.png."""
```
balopat, line 1: This is not just the bristlecone example anymore.

balopat: I think we should just have a separate example for interaction map.

balopat:
@crystalzhaizhai, please see my changes in https://github.com/quantumlib/Cirq/pull/3652/commits/75dd309f92659bc90bc579d5b965f27446e5bc02 based on our discussion + a couple of hours of refactoring. 

Can you please have a look, review and address comments if they are still valid? Thanks!

cirq/protocols/json_serialization_test.py
```diff
@@ -291,6 +291,7 @@ def test_mutually_exclusive_blacklist():
     'ExpressionMap',
     'FSIM_GATESET',
     'Heatmap',
+    'InterHeatmap',
```
balopat, line 294: I'd like to rename this to `InteractionHeatmap`

cirq/vis/heatmap.py
```diff
@@ -184,86 +239,31 @@ def _plot_colorbar(
...
+            row, col = self._target_to_coordinate(target)
             annotation = self.annot_map.get((row, col), '')
             if not annotation:
                 continue
-            text_color = (
-                'black' if self.value_map[(row, col)][0] > (self.vmax + self.vmin) / 2 else 'white'
+            medium_brightness = (
+                (self.vmax + self.vmin) / 2
+                if (self.vmax is not None and self.vmin is not None)
+                else 0.5
```
crystalzhaizhai, line 251: It's reasonable to assume the value falls between 0 and 1 when self.vmax and self.vmin are not give, but what if it falls outside the range?

cirq/vis/heatmap_test.py
crystalzhaizhai: Yeah that's a problems. What way do you suggest to test when result is a graph?

examples/heatmap_example.py
crystalzhaizhai: I will separate this file into two.

cirq/protocols/json_serialization_test.py
crystalzhaizhai: No problem!

examples/heatmap_example.py
crystalzhaizhai: I am okay with using a randomly generated data. Any benefits as opposed to a fixed example?

balopat: It's just more concise! The end result is the same - if you provide a seed.

balopat:
As we don't typically use feature branches, I'm going to close this and switch over to my fork, against which I'll open a PR shortly.

PR #3655, opened 2021-01-08
[WIP] Use concise JSON serialization for FrozenCircuit.

### LINKED ISSUE #3633: Efficient JSON serialization of FrozenCircuit

Part of #3235.
Concise serialization (#3601) is available for Cirq objects, but not yet supported for FrozenCircuit.

FrozenCircuit needs the `_serialization_key_` magic method to enable efficient serialization. This should be a fairly small code change, but may require nontrivial updates to its JSON serialization test.

95-martin-orion:
Fixes #3633. This PR applies the following changes:
- `FrozenCircuit` uses the efficient JSON serialization format.
- Users can optionally assign names to `Circuit` and `FrozenCircuit` objects. These have no special behavior for `Circuit`s, but `FrozenCircuit`s use their name and hash value as their serialization key.
- The JSON serialization test is now agnostic to hash values (`0x` + 16 hex characters) in serialization keys. This allows hardcoded JSON to match the JSON generated from code despite hash values not being consistent between Python sessions.

Note that since names are preserved across the `Circuit`/`FrozenCircuit` boundary, it is possible to e.g. call `freeze()` twice on the same `Circuit`, producing two `FrozenCircuit`s with identical names. This is the desired behavior, as shown below:
- If the `Circuit` has not been mutated, the `FrozenCircuit`s will have matching serialization keys and will be deduped in serialization.
- If the `Circuit` _has_ been mutated, the `FrozenCircuit`s' serialization keys will differ, and they will appear separately in serialization.

### DIFF
--- a/cirq/circuits/circuit.py
+++ b/cirq/circuits/circuit.py
@@ -110,6 +110,11 @@ class AbstractCircuit(abc.ABC):
     def device(self) -> devices.Device:
         pass
 
+    @property
+    @abc.abstractmethod
+    def name(self) -> Optional[str]:
+        pass
+
     def freeze(self) -> 'cirq.FrozenCircuit':
         """Creates a FrozenCircuit from this circuit.
 
@@ -119,7 +124,9 @@ class AbstractCircuit(abc.ABC):
 
         if isinstance(self, FrozenCircuit):
             return self
-        return FrozenCircuit(self, strategy=InsertStrategy.EARLIEST, device=self.device)
+        return FrozenCircuit(
+            self, strategy=InsertStrategy.EARLIEST, device=self.device, name=self.name
+        )
 
     def unfreeze(self) -> 'cirq.Circuit':
         """Creates a Circuit from this circuit.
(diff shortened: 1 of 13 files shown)


95-martin-orion:
@Strilanc has pointed out that hashes should not be used for serialization. I'm leaving this PR up for reference, but if we continue with this serialization format I'll need to replace the hash with something else.

95-martin-orion:
Experiments with a deeply nested circuit, defined as:
```
cs = [cirq.FrozenCircuit(cirq.H(a), cirq.CX(a, b))]
for i in range(1, 12):
    cs.append(cirq.FrozenCircuit(cs[i-1].to_op(), cs[i-1].to_op()))
test_circuit = cs[11]
```
show a distinct difference between compression times and sizes for `gzip` vs. this PR.

In the results below (collected from a single run), "nested" uses the circuit as described above, while "flattened" uses a decomposition of that circuit re-composed into a single-level circuit.
value | this PR | gzip
-|-|-
nested circuits | 0.435255354 s | 7.016707049 s
serialized size (bytes) | 17500 | 299304
flattened circuit | 2.14050902 s | 1.278540381 s
serialized size (bytes) | 4948515 | 20911

gzip code and benchmark can be seen in #3662; the benchmark for this PR is identical but with `to_json`/`read_json` in place of `to_gzip`/`read_gzip`.

95-martin-orion:
Closing in favor of #3815.

PR #3659, opened 2021-01-10
Fix SWAP and ResetChannel not working with CliffordSimulator

Strilanc:
- Just call `act_on` instead of doing type inspection in CliffordSimulator's _base_iterator
- Add ActOnStabilizerCHFormArgs case to ResetChannel
- Add ActOnStabilizerCHFormArgs case to SwapPowGate
- Remove unit test asserting swap gate doesn't work (!)


viathor (approved):

smitsanghavi:
#3656, FYI for removing the other type check based eligibility test in CliffordSimulator.

PR #3660, opened 2021-01-10
Fix TaggedOperation not delegating _act_on_ to its sub_operation

Strilanc:


cirq/ops/raw_types.py
```diff
@@ -630,6 +630,12 @@ def _is_parameterized_(self) -> bool:
             protocols.is_parameterized(tag) for tag in self.tags
         )
 
+    def _act_on_(self, args) -> bool:
```
viathor, line 633: missing type annotation for `args`

(admittedly, this should simply be `args: Any`, but I'd still add it for consistency with other `_act_on_` definitions)

cirq/ops/raw_types.py
```diff
@@ -630,6 +630,12 @@ def _is_parameterized_(self) -> bool:
...
+        sub = getattr(self.sub_operation, "_act_on_", None)
+        if sub is not None:
+            return sub(args)
+        return NotImplemented
```
viathor, line 637: Optional: I think we could simply say

```python
def _act_on_(self, args: Any) -> bool:
    return protocols.act_on(self._sub_operation, args)
```

which is simpler and consistent with how `TaggedOperation` delegates other methods.

BTW: This will probably fix the test coverage error.

viathor (approved):

cirq/ops/raw_types.py
Strilanc: That would actually do things like repeat calls to decompose.

The coverage error was due to defining a Gate instead of an Operation for the nothing case (GateOperation implements it).

viathor: Yup. I agree.

cirq/ops/raw_types.py
Strilanc: Done.

PR #3661, opened 2021-01-10
Support even powers of SWAP gate in CliffordSimulator

smitsanghavi:
#3659 adds support for odd powers


smitsanghavi:
Part of #2423

cirq/ops/swap_gates.py
```diff
@@ -85,16 +85,32 @@ def _trace_distance_bound_(self) -> Optional[float]:
...
 
-        if isinstance(args, sim.ActOnStabilizerCHFormArgs) and self._exponent % 2 == 1:
-            args.state.omega *= 1j ** (2 * self.global_shift * self._exponent)
-            protocols.act_on(ops.CNOT, args)
-            args.axes = args.axes[::-1]
-            protocols.act_on(ops.CNOT, args)
-            args.axes = args.axes[::-1]
-            protocols.act_on(ops.CNOT, args)
+        if isinstance(args, (sim.ActOnStabilizerCHFormArgs, sim.ActOnCliffordTableauArgs)):
+            if not protocols.has_stabilizer_effect(self):
```
Strilanc, line 97: Specialize this instead of going through the general method.

cirq/ops/swap_gates.py
```diff
@@ -85,16 +85,32 @@ def _trace_distance_bound_(self) -> Optional[float]:
...
+                return NotImplemented
+            if isinstance(args, sim.ActOnStabilizerCHFormArgs):
```
Strilanc, line 99: We're already in a block that has done this check.

cirq/ops/swap_gates.py
```diff
@@ -85,16 +85,32 @@ def _trace_distance_bound_(self) -> Optional[float]:
...
+                protocols.act_on(ops.CNOT, args)
+                args.axes = args.axes[::-1]
+                protocols.act_on(ops.CNOT, args)
+                args.axes = args.axes[::-1]
+                protocols.act_on(ops.CNOT, args)
+            else:
+                # Any fractional exponent SWAP gate should have been rejected
+                # already. Any even exponent does not change anything except
+                # the global phase above.
+                assert self._exponent % 2 == 0
```
Strilanc, line 112: I'd drop the assert. Especially after the has stabilizer effect method is inlined.

cirq/ops/swap_gates.py
smitsanghavi: When you say specialize, do you mean to call the private method instead of protocol or inlining as you mentioned in the below comment?

cirq/ops/swap_gates.py
smitsanghavi: It could be ActOnCliffordTableauArgs

cirq/ops/swap_gates.py
smitsanghavi: Done

smitsanghavi:
Friendly ping!

Strilanc (approved):

PR #3664, opened 2021-01-12
Copy states in Clifford simulator

tonybruguier:
During the (ongoing) review of #3630 95-martin-orion@ found a bug in my draft code. He said that I should copy state like in #3109. I changed my code and added a test in that PR.

However, I had inspired myself from the Clifford code which could have the same issue (I think), and so I am fixing the code in the present PR.

### DIFF
--- a/cirq/sim/clifford/clifford_simulator.py
+++ b/cirq/sim/clifford/clifford_simulator.py
@@ -225,7 +225,7 @@ class CliffordSimulatorStepResult(simulator.StepResult):
                 method).
         """
         self.measurements = measurements
-        self.state = state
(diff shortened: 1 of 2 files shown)


tonybruguier:
By the way, I'm not pushing for it. It's just a response to the comment, and it could very well be based on me not understanding the suggestion.

95-martin-orion (approved):
Thanks for the fix! I've added the BREAKING CHANGE label since this does modify the simulator behavior - if anyone relied on the old (incorrect) ability to mutate the simulator state from a step result, their workflow would be broken by this.

PR #3670, opened 2021-01-13
MPS simulator can handle any topology (still required to have 1- or 2-qubit gates)

tonybruguier:


tonybruguier:
Hi,
I tried to be careful, but the size of some tensors is quite big (up to 8-dimensions!) and so it's possible that I made a mistake somewhere. I tried to have good test coverage, but I'm more than happy to add more tests, more comments, more assert statements, etc...

Thanks a lot in advance and this is not particularly urgent.

mpharrigan:
density simulator -> mps simulator?

For this, I think you should really look into quimb (and the `cirq.contrib.quimb` simulator) to get a handle on the indices. The nicety afforded by quimb (or tensor network libraries in general) is you get to provide names for the indices instead of relying on their relative ordering or `np.einsum` letters or `tensor[:, :, 1, :, :]`

tonybruguier:
I followed your advice and used Quimb, and the code is much nice indeed.

The unit tests do pass on my machine, but not on Github because it cannot import quimb. I looked around and it seems other files in Cirq can import it, so I am not sure what is wrong. Do you have a suggestion?

Thanks so much for teaching me about Quimb! I'm a bit concerned that Cirq is just a subset of Quimb :)

PTAL

vtomole:
Quimb is only installed in [contrib](https://github.com/quantumlib/Cirq/blob/e9c09d356e20eb2b408b9fab6c6879c8d54db826/cirq/contrib/contrib-requirements.txt#L7). You need to move that line to [requirements.txt](https://github.com/quantumlib/Cirq/blob/master/requirements.txt)

95-martin-orion:
> Quimb is only installed in [contrib](https://github.com/quantumlib/Cirq/blob/e9c09d356e20eb2b408b9fab6c6879c8d54db826/cirq/contrib/contrib-requirements.txt#L7). You need to move that line to [requirements.txt](https://github.com/quantumlib/Cirq/blob/master/requirements.txt)

Where possible, we're trying to avoid "promoting" contrib dependencies to Cirq-wide dependencies. We should confirm that this change is OK at the Cirq sync before moving forward.

vtomole:
@95-martin-orion Agreed.

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
         return np.array(measurements, dtype=int)
 
 
+def _sum_reduce(M, ind):
+    # TODO(tonybruguier): Use M.sum_reduce(ind, inplace=True) once
```
mpharrigan, line 239: interesting that this doesn't exist (yet, apparently). How is this different than the partial trace? Can you just manipulate the indices' names so they will be contracted away and do the sum you want? Or insert an identity matrix with appropriately named indices so contraction results in a sum-reduction

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
     """A state of the MPS simulation."""
 
-    def __init__(self, qubit_map, initial_state=0):
+    def __init__(self, qubit_map, rsum2_cutoff, initial_state=0):
         self.qubit_map = qubit_map
         self.M = []
-        for qubit in qubit_map.keys():
+
+        # The order of the qubits matters, because the state |01> is different from |10>. Since
+        # Qimb uses strings to name tensor indices, we want to be able to sort them too. If we are
```
mpharrigan, line 261: Quimb

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
+        # working with, say, 123 qubits then we want qubit 3 to come before qubit 100, but then
+        # we want write the string '003' which comes before '100' in lexicographic order. The code
+        # below is just simple string formatting.
+        max_num_digits = len('%d' % (max(qubit_map.values())))
+        self.format_i = 'i_%%.%dd' % (max_num_digits)
+        self.format_mu = 'mu_%%.%dd_%%.%dd' % (max_num_digits, max_num_digits)
```
mpharrigan, line 267: can you convert these to using `"{}".format()` style string formatting? the `%`-style has been discouraged for a while

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
-                    d,
-                )
-            )
-            x[0, 0, (initial_state % d)] = 1.0
-            self.M.append(x)
+            x = np.zeros(d)
+            x[initial_state % d] = 1.0
+
+            i = qubit_map[qubit]
+            self.M.append(qtn.Tensor(x, inds=(self.i_str(i),)))
```
mpharrigan, line 275: Note: it's nice that you are actually dealing with `initial_state`. You (or an interested party) could factor this code out and fix e.g. https://github.com/quantumlib/Cirq/blob/master/cirq/contrib/quimb/state_vector.py#L70 in follow-on work

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
-        self.threshold = 1e-3
+        self.rsum2_cutoff = rsum2_cutoff
+        self.num_svd_splits = 0
+
+    def i_str(self, i):
+        return self.format_i % (i)
+
+    def mu_str(self, i, j):
+        # By convention, the lower index is always the first.
+        i, j = min(i, j), max(i, j)
```
mpharrigan, line 286: could split into two lines for more readability

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
+        return self.format_mu % (i, j)
 
     def __str__(self) -> str:
-        return str(self.M)
+        return str([x.data for x in self.M])
```
mpharrigan, line 290: this could be big, no? Why not just use the `str` of the `TensorNetwork`?

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
 
     def copy(self) -> 'MPSState':
-        state = MPSState(self.qubit_map)
+        state = MPSState(self.qubit_map, self.rsum2_cutoff)
         state.M = [x.copy() for x in self.M]
-        state.threshold = self.threshold
+        state.num_svd_splits = self.num_svd_splits
         return state
 
+    def _get_qubit_queue(self):
```
mpharrigan, line 301: can you add a docstring that explains what this helper function does. I know it's private, but it's rather involved

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
+        # We can aggregate the qubits in any order we want, theoretically. However, it's quite
```
mpharrigan, line 302: aggregate = contract? 

Is this algorithm from the paper? Have you tried using quimb's built-in contraction ordering finder?

cirq/contrib/quimb/mps_simulator.py
mpharrigan: type annotations for return value.

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
+
+                return delta_row + delta_col
+            else:
+                # For non-2D, we just use the naive ordering.
+                return self.qubit_map[qubit1] - self.qubit_map[qubit2]
+
+        qubit_queue.sort(key=functools.cmp_to_key(qubit_ordering_fn))
+        return qubit_queue
+
+    def _sum_up(self, skip_tracing_out_for_qubits):
```
mpharrigan, line 338: type annotations, short docstring. maybe a more descriptive method name?

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
+                    pop_idx = idx
+
+            # Pop the element to aggregate
+            qubit = qubit_queue.pop(pop_idx)
+
+            i = self.qubit_map[qubit]
+            Mi = self.M[i]
+            if _trace_out(i):
+                Mi = _sum_reduce(Mi, self.i_str(i))
+            M = _make_tensor_if_scalar(M @ Mi)
```
mpharrigan, line 376: is this in place of quimb's `TensorNetwork` contraction facilities? why are you dotting "by hand"?

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
+        return M
+
+    def partial_state_vector(self, skip_tracing_out_for_qubits):
```
mpharrigan, line 379: type annotations, docstring

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
+        M = self._sum_up(
+            skip_tracing_out_for_qubits={
+                self.qubit_map[qubit] for qubit in skip_tracing_out_for_qubits
+            }
+        )
+        # Here, we rely on the formatting of the indices, and the fact that we have enough
+        # leading zeros so that 003 comes before 100.
+        sorted_ind = tuple(sorted(M.inds))
```
mpharrigan, line 387: instead of relying on sortable indices could you keep a parallel mapping of e.g. qubit to string-index and do all "logic" on the qubits themselves and only translate to string-indices when calling a quimb api?

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
+        return M.fuse({'i': sorted_ind}).data
+
     def state_vector(self):
-        M = np.ones((1, 1))
-        for i in range(len(self.M)):
-            M = np.einsum('ni,npj->pij', M, self.M[i])
-            M = M.reshape(M.shape[0], -1)
-        assert M.shape[0] == 1
-        return M[0, :]
+        return self.partial_state_vector(skip_tracing_out_for_qubits={})
```
mpharrigan, line 391: Usually "tracing out" qubits means removing them. I.e., you trace out all qubits from a simulation you don't get the state vector, you get the number 1. Is this rather something like `skip_contracting_qubits={}`? or is it `trace_out_qubits={}`?

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
-            self.M[n] = np.einsum('mis,sn->mni', X, S)
-            self.M[p] = np.einsum('ns,spj->npj', S, Y)
+        if len(op.qubits) == 1:
+            n = self.qubit_map[op.qubits[0]]
+
+            old_n = self.i_str(n)
+            new_n = 'new_' + old_n
+
+            U = qtn.Tensor(U, inds=(new_n, old_n))
+            self.M[n] = (U @ self.M[n]).reindex({new_n: old_n})
```
mpharrigan, line 406: so you're actually contracting in the one qubit gates here explicitly? I think usually with quimb you can just throw everything into its own `Tensor` and contract in one fell swoop at the end, but maybe to schedule the SVD's how you want, you have to do it this way

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
+            n, p = [self.qubit_map[qubit] for qubit in op.qubits]
+
+            old_n = self.i_str(n)
+            old_p = self.i_str(p)
+            new_n = 'new_' + old_n
+            new_p = 'new_' + old_p
+
+            U = qtn.Tensor(U, inds=(new_n, new_p, old_n, old_p))
+
+            # This is the index on which we do the joining. We need to add it iff it's the first
```
mpharrigan, line 419: joining = contraction?

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
+                cutoff_mode='rsum2',
+                get='tensors',
+                absorb='both',
+                bond_ind=mu_ind,
+            )
+
+            self.M[n] = X.reindex({new_n: old_n})
+            self.M[p] = Y.reindex({new_p: old_p})
         else:
+            # NOTE(tonybruguier): There could be a way to handle higher orders. I think this could
```
mpharrigan, line 442: I think `T.split` will work for higher-order tensors.

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
+            # involve HOSVDs:
+            # https://en.wikipedia.org/wiki/Higher-order_singular_value_decomposition
+            #
+            # TODO(tonybruguier): Evaluate whether it's even useful to implement and learn more
+            # about HOSVDs.
             raise ValueError('Can only handle 1 and 2 qubit operations')
 
+    def estimation_stats(self):
```
mpharrigan, line 450: docstring

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -231,92 +235,243 @@ def sample(
...
+
+        # The computation below is done for numerical stability, instead of directly using the
+        # formula:
+        # estimated_fidelity = (1 - self.rsum2_cutoff) ** self.num_svd_splits
+        estimated_fidelity = 1.0 + np.expm1(np.log1p(-self.rsum2_cutoff) * self.num_svd_splits)
+        estimated_fidelity = round(estimated_fidelity, ndigits=3)
+
+        return {
+            "num_coefs_used": num_coefs_used,
+            "memory_bytes": memory_bytes,
```
mpharrigan, line 462: check out https://github.com/quantumlib/Cirq/blob/master/cirq/contrib/quimb/state_vector.py#L171 where we just let quimb do the contraction and it can tell us the max RAM throughout the contraction ordering

mpharrigan (commented):
This is cool.

You've exploited some of the niceties of quimb's named indices, but it still looks like you're doing the contraction "by hand". This is obviously functional and I think we should work to get this merged with the existing implementation but I'm also curious whether you could use some of the higher-order quimb functionality to just whop a whole circuit into a `TensorNetwork` and then use the contraction ordering functionality to get it to do what you want.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: This has been added with this commit:
https://github.com/jcmgray/quimb/commit/7876b4fd7daf6a3b3a94e70d8370f55515b1d21d

But that was on 2020-08-20 and the current release is from 2020-02-18:
https://quimb.readthedocs.io/en/latest/changelog.html

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done. I also decided to add a unit test that verifies the formatting directly.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: I am a bit confused by your comment, but I did take a second look at the code:
https://sourcegraph.com/github.com/jcmgray/quimb@88cc3d88c2a43f08e95b32528594e3c2239d2184/-/blob/quimb/tensor/tensor_core.py#L611

Let's say that we have a 3-qubit gate, shouldn't we split along three groups of indices? Like instead of 'left_inds' and 'right_inds', we would have 'group_1_inds', 'group_2_inds', and 'group_3_inds'.

Leaving unresolved, because I suspect I am missing something.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Added a TODO in my name.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: How would you do this? Here's what I tried:

```
tensor_network
(Tensor(shape=(2,), inds=('i_0',), tags=set()), Tensor(shape=(2,), inds=('i_1',), tags=set()))

U
Tensor(shape=(2, 2), inds=('new_i_0', 'i_0'), tags=set())

new_tensor = tensor_network @ U
Tensor(shape=(2, 2), inds=('i_1', 'new_i_0'), tags=set())

new_network = tensor_network & U
(Tensor(shape=(2,), inds=('i_0',), tags=set()), Tensor(shape=(2,), inds=('i_1',), tags=set()), Tensor(shape=(2, 2), inds=('new_i_0', 'i_0'), tags=set()))

```
So if I use the `@` operator I go from a network with 2 tensors to a single tensor, and if I use the `&` operator, then I get non-collapsed tensors.

Leaving unresolved because I suspect I am missing something.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Nice! That deleted a lot of code, including some that you had comments one. Thank you for showing me this! Done.

tonybruguier:
Thanks for the insightful comments, they simplified the code greatly and I learned a lot.

There are two comments that I left unresolved, because I suspect more work needs to be done.

PTAL.

cirq/contrib/quimb/mps_simulator.py
mpharrigan: The key point is that it contracts any indices that share the same name. In your example, there is still a tensor-factorized structure between the first and second tensor in `tensor_network` and `new_network`. 

File_000

cirq/contrib/quimb/mps_simulator.py
mpharrigan: This is outside of the scope of the current PR, and I'm not sure I understand 100% the algorithm; but you can still split it either asymmetrically (and just accept that entanglement will grow, albeit slower than if you kept it as 3 qubits). or split twice (i.e. SVD into (1qubit)(singularvalues)(2qubit) --> (1qubit)(singularvalues1)(1qubit)(singularvalues2)(1qubit))

mpharrigan:
Wow! That's a lot of removed code :) isn't quimb (and tensor networks) great!?

cirq/contrib/quimb/mps_simulator.py
mpharrigan: (optional for current PR)

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -235,83 +238,155 @@ def sample(
...
+        state.num_svd_splits = self.num_svd_splits
         return state
 
+    def _sum_up(self, skip_tracing_out_for_qubits):
+        end_bras = []
+        if skip_tracing_out_for_qubits:
+            for qubit, i in self.qubit_map.items():
+                if i not in skip_tracing_out_for_qubits:
+                    d = qubit.dimension
+                    data = [math.sqrt(d)] * d
```
mpharrigan, line 295: Why is this sqrt(d)? That means `end_bra` isn't normalized! Even if it was something like [1/sqrt(2), 1/sqrt(2)] I don't know why we'd want to "cap" the circuit with |+> state end bras. Maybe this is described in the paper?

cirq/contrib/quimb/mps_simulator.py
mpharrigan: the tensor network contraction simplification means I can now understand what's being controlled by this argument, although I have big questions about why you're using the bras you're using (see above). This isn't tracing out. It's "projecting out" or "selecting" the output states for these qubits

cirq/contrib/quimb/mps_simulator.py
mpharrigan: optional for current PR, since I think it would be a whole exercise to try to re-write this function to wholely use quimb's contraction engine. I'm pleased you got it working in `_sum_up`.

mpharrigan (changes requested):
Looking really good. Main outstanding thing is docstrings and answering some of my questions. I explicitly marked some of my other comments as "optional", since this PR is functional and readable at present and don't want to block merging it on those ideas

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Ack.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: That's because I was wrong. I was assuming the state vector for the measured qubit was pure, which is not always the case. I was just summing out the other indices with a |+> that I had rescaled so that probs would add up to 1.0. However, I had only checked that the probs add up to 1.0 for one case, but it wasn't always the case.

I never caught the issue because all my testing had either pure states, or states |0> or |1>. In addition, I was always force-renormalizing to sum of probs to 1.0, even if they were very far from 1.0.

Instead, what I should have done is compute the partial trace, which is what I do now. I added a test that makes sure the partial trace is the same, and I tried to have a more coverage. In addition, I now have a tolerance for the sum of the probs, making sure they are close to 1.0 (that test was failing before my new code).

Thanks for catching this and sorry about this mistake.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Moot now for this function, but added doctrings and annotations throughout.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Now using Quimb's built-in contraction ordering.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Moot, but added a new function called 'partial_trace' instead.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Moot now.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: You had the good intuition, and it was indeed very wrong. See complete comment above.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Ack. Added a TODO.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Added a TODO.

tonybruguier:
Thanks. PTAL.

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -263,60 +336,98 @@ def __init__(self, qubit_map, rsum2_cutoff, initial_state=0):
             initial_state = initial_state // d
         self.M = self.M[::-1]
         self.rsum2_cutoff = rsum2_cutoff
+        self.sum_prob_atol = sum_prob_atol
         self.num_svd_splits = 0
 
-    def i_str(self, i):
+    def i_str(self, i: int) -> str:
+        # Returns the indice name for the i'th qid.
```
mpharrigan, line 343: index

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -263,60 +336,98 @@ def __init__(self, qubit_map, rsum2_cutoff, initial_state=0):
...
         return self.format_i.format(i)
 
-    def mu_str(self, i, j):
-        # By convention, the lower index is always the first.
+    def mu_str(self, i: int, j: int) -> str:
+        # Returns the indice name for the pair of the i'th and j'th qids. Note
```
mpharrigan, line 347: index

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -263,60 +336,98 @@ def __init__(self, qubit_map, rsum2_cutoff, initial_state=0):
...
+
+        Returns:
+            An array that contains the partial trace.
+        """
+
+        contracted_inds = set(
+            [self.i_str(i) for qubit, i in self.qubit_map.items() if qubit not in keep_qubits]
+        )
+
+        CONJ_PFX = "conj_"
```
mpharrigan, line 397: lowercase for local variables or promote to an actual global variable

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -263,60 +336,98 @@ def __init__(self, qubit_map, rsum2_cutoff, initial_state=0):
...
+
         # Here, we rely on the formatting of the indices, and the fact that we have enough
         # leading zeros so that 003 comes before 100.
-        sorted_ind = tuple(sorted(M.inds))
-        return M.fuse({'i': sorted_ind}).data
+        sorted_ind = tuple(sorted(state_vector.inds))
+        return state_vector.fuse({'i': sorted_ind}).data
+
+    def partial_trace(self, keep_qubits: Set[ops.Qid]) -> np.ndarray:
+        """Returns the partial trace only for the keep_qubits specified.
```
mpharrigan, line 381: "Trace out all qubits except `keep_qubits`"

(as written, it's confusing whether keep_qubits are traced out or kept)

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -263,60 +336,98 @@ def __init__(self, qubit_map, rsum2_cutoff, initial_state=0):
...
+        tensor_network = qtn.TensorNetwork(self.M)
+
+        # Rename the internal indices to avoid collisions. Also rename the qubit
+        # indices that are kept. We do not rename the qubit indices that are
+        # traced out.
+        conj_tensor_network = tensor_network.conj()
+        for M in conj_tensor_network.tensors:
+            for ind in M.inds:
+                if ind not in contracted_inds:
+                    conj_tensor_network.reindex({ind: CONJ_PFX + ind}, inplace=True)
```
mpharrigan, line 408: I think how `reindex` is defined, you should create the reindexing dictionary once and then call `reindex` once

```python
mapping = {}
for m:
  for ind:
    if ind:
      mapping[ind] = conj + ind
tn.reindex(mapping)
```

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -263,60 +336,98 @@ def __init__(self, qubit_map, rsum2_cutoff, initial_state=0):
...
+        partial_trace = conj_tensor_network @ tensor_network
+
+        old_inds = [self.i_str(self.qubit_map[keep_qubit]) for keep_qubit in keep_qubits]
+        new_inds = [CONJ_PFX + old_ind for old_ind in old_inds]
+        return partial_trace.to_dense(old_inds, new_inds)
```
mpharrigan, line 413: "old" and "new" are bad names for this. How about `orig_inds, conj_inds`. or `forward_inds, backward_inds`. I sortof leaned into the "forwards backwards" thing in the quimb density simulator as that's how I like to think about it: you keep two copies of the circuit and one runs "forwards" i.e. normal and the other applies the conj transpose of each operation, i.e. it evolves "backwards" in time

mpharrigan (approved):
sweet

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done.

tonybruguier:
Thanks for all the help. I think I've addressed all the comments, but happy to do more if required.

mpharrigan (approved):

PR #3671, opened 2021-01-13
Fix assumption about order in MPSSimulator

tonybruguier:
My unit testing didn't test the case when the order of qubits is inverted. I added a test and this causes an error that this PRs attempts to fix.

The error isn't too bad I think, because I had added some asserts (so no bad results), but it should be fixed nonetheless. I apologize for this mistake.

### DIFF
--- a/cirq/sim/mps_simulator.py
+++ b/cirq/sim/mps_simulator.py
@@ -284,10 +284,15 @@ class MPSState:
             n = idx[0]
             self.M[n] = np.einsum('ij,mnj->mni', U, self.M[n])
         elif len(idx) == 2:
-            n = idx[0]
-            p = idx[1]
-            if abs(n - p) != 1:
+            if abs(idx[0] - idx[1]) != 1:
                 raise ValueError('Can only handle continguous qubits')
+            elif idx[0] < idx[1]:
+                n = idx[0]
+                p = idx[1]
+            else:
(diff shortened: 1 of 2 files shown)


cirq/sim/mps_simulator.py
```diff
@@ -284,10 +284,15 @@ def apply_unitary(self, op: 'cirq.Operation'):
             n = idx[0]
             self.M[n] = np.einsum('ij,mnj->mni', U, self.M[n])
         elif len(idx) == 2:
-            n = idx[0]
-            p = idx[1]
-            if abs(n - p) != 1:
+            if abs(idx[0] - idx[1]) != 1:
                 raise ValueError('Can only handle continguous qubits')
+            elif idx[0] < idx[1]:
+                n = idx[0]
```
mpharrigan, line 290: more readable with tuple unpacking?

```python
n, p = idx
```

tonybruguier: Done.

tonybruguier:
Thanks. PTAL

tonybruguier:
Hi,
I've integrated the changes in this PR (including your comment), in the PR #3670 which adds the 2D version. Happy to work on both PRs, or abandon this one and work exclusively on #3670

mpharrigan (approved):

PR #3673, opened 2021-01-13
Enforce unique serialization keys

95-martin-orion:
This PR modifies the `SerializableByKey` serialization process to assign a unique suffix to each serialized object. This way keys are guaranteed to be unique within a serialized object - we no longer rely on `_serialization_key_` to provide unique values.

Keys generated this way are consistent between serializations of the same top-level object, provided that each component object has a consistent serialization. However, the key of an object may change between serializations if the top-level object changes. For example:
```
# Pseudocode - each letter represents a SerializableByKey object.
a = [x, y]  # suffix of key(x) is "_1"
b = [y, x]  # suffix of key(x) is "_2"
```

Prerequisite to #3655, which will require updates to remove hashes from serialization keys. Technically a breaking change, but only for unreleased behavior.


95-martin-orion:
Ping! (In related news, gzipped-JSON serialization has been submitted: #3662)

cirq/protocols/json_serialization.py
```diff
@@ -447,12 +447,11 @@ class SerializableByKey(SupportsJSON):
...
-    def _serialization_key_(self) -> str:
-        """Returns a unique string identifier for this object.
+    def _serialization_name_(self) -> str:
+        """Returns a string identifier for this object.
 
-        This should only return the same value for two objects if they are
-        equal; otherwise, an error will occur if both are serialized into the
-        same JSON string.
+        This identifier need not be globally unique; a unique suffix will be
+        dynamically generated to create the serialization key for this object.
```
balopat, line 454: Mention recommended readability for the name

cirq/protocols/json_serialization.py
```diff
@@ -518,12 +517,12 @@ def __init__(self, obj: Any):
         # Context information and the wrapped object are stored together in
         # `object_dag` to ensure consistent serialization ordering.
         self.object_dag = []
-        context_keys = set()
+        context = []
```
balopat, line 520: can you double check whether this could be a set?

cirq/protocols/json_serialization.py
```diff
@@ -623,23 +622,23 @@ def to_json(
...
-                    raise ValueError(
-                        'Found different objects with the same serialization key:'
-                        f'\n{ContextualEncoder.context_map[skey]}\n{o}'
-                    )
-                if skey is not None:
-                    ContextualEncoder.context_map[skey] = o._json_dict_()
-                return super().default(o)
+                if not hasattr(o, '_serialization_name_'):
+                    return super().default(o)
+                try:
```
balopat, line 634: we could convert this to a for loop for linear search - that would be a bit more readable / less fragile than a try/except

cirq/protocols/json_serialization.py
```diff
@@ -623,23 +622,23 @@ def to_json(
...
+                    idx = ContextualEncoder.context.index(o)
+                    return _SerializedKey(ContextualEncoder.keys[idx])._json_dict_()
+                except ValueError:
+                    ContextualEncoder.context.append(o)
+                    key = f'{o._serialization_name_()}_{len(ContextualEncoder.context)}'
+                    ContextualEncoder.keys.append(key)
+                    return super().default(o)
```
balopat, line 641: here's our hacked version of the Encoder:

```
        obj = _ContextualSerialization(obj)

        class ContextualEncoder(cls):  # type: ignore
            """An encoder with a context map for concise serialization."""

            # These lists populate gradually during serialization. An object
            # with components defined in 'context' will represent those
            # components using their keys instead of inline definition.
            seen: Set[str] = set()

            def default(self, o):
                if not hasattr(o, '_serialization_name_'):
                    return super().default(o)
                for candidate in obj.object_dag[:-1]:
                    if candidate.obj == o:
                        if not candidate.key in ContextualEncoder.seen:
                            ContextualEncoder.seen.add(candidate.key)
                            return candidate.obj._json_dict_()
                        else:
                            return _SerializedKey(candidate.key)._json_dict_()
                raise ValueError("this should never happen")
```

balopat: Namely, this removes the duplication of the suffix calculation and context building

balopat (changes requested):
Discussed it offline/(online?), please make the necessary changes then we can merge it! Thanks!

cirq/protocols/json_serialization.py
95-martin-orion: Making this a set requires SBKs to be hashable. In principle I agree with this, but I don't think SBK is the right level to apply the restriction; since we support serialization of non-hashable objects, SBKs should also be allowed to be non-hashable.

cirq/protocols/json_serialization.py
95-martin-orion: Done.

cirq/protocols/json_serialization.py
95-martin-orion: Included in "hacked version".

cirq/protocols/json_serialization.py
95-martin-orion: Added. Thanks for your help here!

balopat (approved):

PR #3645, opened 2021-01-07
Remove CliffordTableau from CliffordState

smitsanghavi:
#2423

PR #3679, opened 2021-01-15
Wrap input json_dict entries in np.array to avoid wrong type deduction

smitsanghavi:
Not doing this caused an issue where the JSON was parsed without issue but when using the object, certain methods would break since they expected an np.array and got a list.

Example: https://github.com/quantumlib/Cirq/pull/3645#discussion_r557795288

Dynamic typing strikes again!

### DIFF
--- a/cirq/sim/clifford/stabilizer_state_ch_form.py
+++ b/cirq/sim/clifford/stabilizer_state_ch_form.py
@@ -73,18 +73,18 @@ class StabilizerStateChForm:
     def _from_json_dict_(cls, n, G, F, M, gamma, v, s, omega, **kwargs):
         copy = StabilizerStateChForm(n)
 
-        copy.G = G.copy()
-        copy.F = F.copy()
-        copy.M = M.copy()
-        copy.gamma = gamma.copy()
-        copy.v = v.copy()
-        copy.s = s.copy()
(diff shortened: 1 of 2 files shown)


balopat (approved):

cirq/sim/clifford/stabilizer_state_ch_form.py
```diff
@@ -73,18 +73,18 @@ def _json_dict_(self) -> Dict[str, Any]:
     def _from_json_dict_(cls, n, G, F, M, gamma, v, s, omega, **kwargs):
         copy = StabilizerStateChForm(n)
 
-        copy.G = G.copy()
-        copy.F = F.copy()
-        copy.M = M.copy()
-        copy.gamma = gamma.copy()
-        copy.v = v.copy()
-        copy.s = s.copy()
+        copy.G = np.array(G.copy())
```
mpharrigan, line 76: You can remove the `copy`. This can be just `np.array(G)` which copies by default. [otherwise, you should be using `np.asarray`, which doesn't copy]

smitsanghavi: Thanks, I'll send out a follow up PR

PR #3682, opened 2021-01-16
[POC] Factor a circuit into entanglement sets

### LINKED ISSUE #3675: Canonical Subcircuit Factorization

I think given [TEBD](https://en.wikipedia.org/wiki/Time-evolving_block_decimation), it allows for the ability to create a canonical subcircuit factorization of any circuit. The canonical subcircuits would be determined at any particular moment by the disjoint sets of entangled qubits, and for each of those would extend on the t-axis until the moment where it is entangled with another set, or one of its qubits is measured.

Combined with #3674, caching the matrix representations of each of these subcircuits could speed up simulation time. Additionally, many of these subcircuits may be quite common across different projects when factored in this way, increasing the cache hit rate even when working across projects. This representation also allows a smart simulator to run the dependency graph of subcircuits in parallel.

An optimizer that factorizes the circuit in the above fashion. Eventually a simulator that makes use of caching and/or parallelism to simulate these in an efficient manner.

daxfohl:
Closes #3675

Optimizer that factors a circuit into subcircuits such that each subcircuit represents a set of qubits that are all entangled with each other, for the duration of that entanglement, and such that there is no entanglement across subcircuits.

Completed: Join into new circuit when entanglement is detected, split into multiple circuits on measurement.

Some things not done: Correctly identify entangling gates (Id(2) would be considered "entangling", just noticed Reset gate is not handled -- maybe there should be a protocol where a gate definition specifies what it entangles/detangles, rather than explicitly placing that logic here).

### DIFF
--- a/cirq/optimizers/factor_test.py
+++ b/cirq/optimizers/factor_test.py
@@ -12,21 +12,74 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from cirq import NamedQubit, Circuit, X, Y, Moment, measure, CNOT
+from cirq import NamedQubit, Circuit, X, Y, Moment, measure, CNOT, CircuitOperation, FrozenCircuit
 from cirq.optimizers.factor import factor_circuit
 
 
 def assert_optimizes(before, after):
     factored = factor_circuit(before)
-    print()
-    print()
-    print()
-    print(factored)
-    # assert factored == after
+    assert factored == after
 
 
-def test_align_left():
-    print()
+def test_factor_simple():
+    q1 = NamedQubit('q1')
+    assert_optimizes(
+        before=Circuit(
+            [
+                Moment([X(q1)]),
+            ]
+        ),
+        after=FrozenCircuit(
+            [
+                Moment([CircuitOperation(FrozenCircuit([X(q1)]))]),
+            ]
+        ),
+    )
+
+
+def test_factor_medium():
+    q1 = NamedQubit('q1')
+    q2 = NamedQubit('q2')
+    assert_optimizes(
+        before=Circuit(
+            [
+                Moment([X(q1)]),
+                Moment([X(q1), X(q2)]),
+                Moment([Y(q1)]),
+            ]
+        ),
+        after=FrozenCircuit(
+            [
+                Moment(
+                    [
+                        CircuitOperation(
+                            FrozenCircuit(
+                                [
+                                    Moment([X(q1)]),
+                                    Moment([X(q1)]),
+                                    Moment([Y(q1)]),
+                                ]
+                            )
+                        ),
+                        CircuitOperation(
+                            FrozenCircuit(
+                                [
+                                    Moment(),
+                                    Moment([X(q2)]),
+                                    Moment(),
+                                ]
+                            )
+                        ),
+                    ]
+                ),
+                Moment(),
+                Moment(),
+            ]
+        ),
+    )
+
+
+def test_factor_complex():
     q1 = NamedQubit('q1')
     q2 = NamedQubit('q2')
     assert_optimizes(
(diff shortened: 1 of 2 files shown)


daxfohl:
Converts
```
q1: ───X───X───Y───@───X───X───Y───M('a')───X───X───Y───
                   │
q2: ───────X───────X───────X────────────────────X───────
```
to
```
                                             Circuit_0xd84bc11ac40e4bf5:
q1: ───Circuit_0x37d6a43e0981b777:───────────[ q1: ───@───X───X───Y───M('a')─── ]───────────────────Circuit_0x37d6a43e0981b777:───────────
       [ q1: ───X───X───Y───     ]           [        │                         ]                   [ q1: ───X───X───Y───     ]
                                             [ q2: ───X───────X──────────────── ]
                                             │
q2: ───Circuit_0xea3d9b4fd937d2aa:───────────#2─────────────────────────────────────────────────────Circuit_0xea3d9b4fd937d2aa:───────────
       [ q2: ───────X───────     ]                                                                  [ q2: ───────X───────     ]
```

daxfohl:
@95-martin-orion BTW note in the above diagram, the circuits at the start and the end appear to be given the same ID. They are definitely two different objects created at different times in the code even though they have the same structure. Is that the expected behavior?

cirq/optimizers/factor_test.py
```diff
@@ -0,0 +1,172 @@
...
+                                [
+                                    Moment(),
+                                    Moment([X(q2)]),
+                                    Moment(),
+                                ]
+                            )
+                        ),
+                    ]
+                ),
+                Moment(),
```
95-martin-orion, line 75: These extra empty moments shouldn't be necessary - if you can remove them easily, I'd recommend doing so.

daxfohl: Yeah, those came about from my misunderstanding that subcircuit moments were synchronous with the primary circuit's moments, which you clarified on the other thread. I'll figure out how to fix that up.

95-martin-orion:
> @95-martin-orion BTW note in the above diagram, the circuits at the start and the end appear to be given the same ID. They are definitely two different objects created at different times in the code even though they have the same structure. Is that the expected behavior?

This is due to the current naming behavior of `FrozenCircuit`: the ID is based on the circuit hash, which is identical for identical circuits. Sharing an ID is fine in this case.

cirq/optimizers/factor_test.py
daxfohl: Done in latest push

daxfohl:
@95-martin-orion do you see any use for something like this? The main thing I was thinking is that simulators could run each circuit independently and maybe gain some efficiency there. But if that's not how the simulators work, or if real-world circuits are mostly fully entangled, then probably this feature doesn't add value? I'd rather close it and pick up something else if that's the case. Can always reopen later if we find a use for it.

balopat (changes requested):
Hi @daxfohl, 

1. I think it is in general a hard problem to figure out what amount of entanglement a gate produces, as it depends on the state that it is applied to. E.g. a CNOT can produce maximum entanglement between two qubits after an H gate on the control qubit (i.e. CNOT(|+0>)), however CNOT(|00>) will not give you any entanglement. Also, if you apply another CNOT later on, it will reduce the entanglement! What about noise? Thus I think a factorization like this needs to be based on something else than just purely checking number of qubits in a gate, e.g. the Schmidt decomposition or something along those lines that would be able to determine product states vs not in subsystems and keep track of that. But I'm not sure it's going to be easy or something that would speed up simulation - given that you'd have to calculate the state vector at intermediate steps - i.e. do a simulation for the whole thing. I might be wrong about this, happy to be proven wrong. 
2. There might be some benefit of this kind of factorization - 1 qubit vs multiqubit sets - but I am not aware of one. 

If you are motivated to see this to the end, it would be beneficial to dig a bit deeper and see whether there is a way our built-in simulator could make use of this and prove it to yourself (and us) that this is a valid avenue. Unfortunately doing the research itself would take some time on our end which is currently not likely that we'll invest. If you bring this up on a Cirq Cynque, you can give it a bit more attention and maybe someone from the maintainers will know more about this area.

mpharrigan:
Can you compare and contrast this with https://github.com/quantumlib/Cirq/pull/3599 and https://github.com/quantumlib/Cirq/issues/3409

daxfohl:
@balopat Ah, interesting. It occurred to me that like a swap gate wouldn't entangle anything, but I didn't realize there would be gates that may conditionally entangle depending on input state. So to compute the subcircuits, it sounds like basically you have to do the simulation, which, kind of defeats the purpose. I believe this would still factor into the smallest unentangled subcircuits given arbitrary unentangled input states to the main circuit, but that doesn't seem useful.

Okay, I think I'll drop this and start looking at something more immediately useful. Maybe I'll keep noodling with it in my spare time if I come up with a potential use case.

@mpharrigan My understanding was that that one splits the circuit horizontally all the way through, for qid sets that never entangle through the whole circuit. (Maybe I'm wrong). This one creates rectangles of subcircuits that are not entangled for a certain duration in time.

mpharrigan:
Check out the quimb simulator (https://github.com/quantumlib/Cirq/blob/master/cirq/contrib/quimb/Contract-a-Grid-Circuit.ipynb) especially look into the call `tn.rank_simplify`. this fuses in all gates. I'm sure you could get quimb to partially-contract circuits into bigger blobs

Issue #3686, opened 2021-01-19
Decouple simulation state model from circuit traversal


daxfohl:
**Is your feature request related to a use case or problem? Please describe.**

I think the core of #3684 and #3677 is that simulator defies the single responsibility pattern. It is responsible for maintaining and transitioning state space as operations are applied, traversing the circuit, maintaining measurements, among other things. I think we fix this by decoupling some of those aspects of simulation, so I'm breaking this out as a separate issue. This will also be the next level of solving #2178.

**Describe the solution you'd like**

I'd propose breaking this into at least two separate things: a Simulation Algorithm and a Simulation Engine. The simulation algorithm would only be concerned with creating initial state from qubit sets, and modifying that state based off the current state and an operation. The simulation engine would be responsible for traversing a circuit and feeding operations into the simulation algorithm.

* Simulation Algorithm: Init state given qubits, update state given operations
* Simulation Engine: traverse circuit using simulation algorithm

This creates a simpler interactive mode, where you only need qubits, a simulation algorithm, and some gates, solving #3684. It also solves #3677 since the simulation algorithm would have the functionality to create a state space from a list of qubits.

Then this allows you to do optimizations and tweaks around the simulations themselves independently of the simulation algorithm. For example, caching, aggregating waveforms, splitting into unitary/non-unitary parts, parallelization, logging, etc. Perhaps there's an opportunity for even further decoupling of simulation engine pieces in the future, but this seems like the first and most fundamental issue to handle.

Need to investigate other simulators and determine whether this type of decoupling is possible.

**What is the urgency from your perspective for this issue? Is it blocking important work?**

P3 - I'm not really blocked by it, it is an idea I'd like to discuss / suggestion based on principle

95-martin-orion:
Could you provide more detail on the intended behavior here? In particular, many simulators (qsim, qFlex, and I believe quimb) achieve performance improvements over the Cirq simulator by analyzing the circuit as a whole. How does this mesh with feeding in operations with a Simulation Engine?

From my perspective, this model is oriented towards simplifying a future "interactive mode" for circuit execution, but I don't think that's our primary use case. Cirq is first and foremost a library for constructing circuits; outside of tests, the execution of those circuits should primarily take place on quantum hardware or faster external simulators, which largely prefer the "circuit in, results out" interface we currently use.

daxfohl:
[...]

PR #3688, opened 2021-01-20
[POC] Decoupling state from simulator

daxfohl:
Working toward #3686 and continuing #2178

Here's a POC of what the decoupled structure would look like for sparse simulator. It's in line with the suggestion from @Strilanc here (maybe). https://github.com/quantumlib/Cirq/issues/2178#issuecomment-535221018. To create a new state handler, you implement the AbstractState interface. Then you can pass it into SparseSimulator and it'll work. If we're lucky, this interface is enough to let us implement lots of different state management algorithms and not have to deal with tying them to circuit traversal.

Will look through some other simulators to see whether this pattern can be reused as nicely as I'm hoping, or whether the simulators are all coupled strongly to the circuit traversal.

cc @cduck @95-martin-orion


daxfohl:
(To note, this doesn't break any existing simulators or interfaces. It just adds new possibilities for simulators that go with this approach.)

mpharrigan:
Do we have multiple simulators that step through op-by-op? I guess this could factor out some shared code between the statevector `Simulator` and the `DensityMatrixSimulator`. Would be nice to see more code removed than added in a refactoring PR like this

daxfohl:
Okay, thanks for the feedback. It sounds like this abstraction doesn't add a whole lot of real world value. I'll go ahead and close it for now.

mpharrigan:
Well I was just asking if anything other than the current code can be factored out. I don't know if this abstraction would be useful

PR #3689, opened 2021-01-20
Add more test for heatmap colorbar

### LINKED ISSUE #2969: Add to heatmap visualization tests
In the `test_colorbar` test there is a comment about testing that the position size and pad arguments are respected.

ybc1991:
Adds more testing logic for colorbar position, size and padding in heatmaps. 

Fixes #2969.


ybc1991:
@googlebot I signed it!

cirq/vis/heatmap_test.py
```diff
@@ -226,20 +226,70 @@ def test_urls(ax, test_GridQubit):
...
-    # TODO: Make this is a more thorough test, e.g., we should test that the
-    # position, size and pad arguments are respected.
-    # Github issue: https://github.com/quantumlib/Cirq/issues/2969
+    fig_pos = fig2.get_axes()[0].get_position()
+    colorbar_pos = fig2.get_axes()[1].get_position()
+    size_perc, pad_perc = int(size.replace("%", "")) / 100, int(pad.replace("%", "")) / 100
+
+    # By default, the canvas size for plot is [[0.125, 0.125], [0.9, 0.88]]
+    canvas_size = 0.775 if position in ["left", "right"] else 0.755
+    expected_pad = canvas_size * pad_perc / (1 + size_perc + pad_perc)
```
balopat, line 274: There are a couple of issues here: 
1.) the canvas size is magically inserted here...in tests it would be better to control these external defaults, instead putting them in comments.
2.) however, we don't need to know the canvas size!  Based on https://matplotlib.org/3.3.3/api/_as_gen/matplotlib.pyplot.colorbar.html - the pad property sets the `Fraction of original axes between colorbar and new image axes.`. The same applies for colorbar `size`! Thus, I think you'd get better results if you'd calculate the actual percentage of padding based on the _size of the original axes_.

balopat (changes requested):
Thanks for your first contribution! I added my comments.

cirq/vis/heatmap_test.py
balopat: I did some experiments and confirmed that both the `size` and `pad` arguments are percentages of the original axes width (left right) or height (bottom, top)

ybc1991: Thanks for the suggestion! Yes, using the size of original axes makes the test much simplifier and accurate.

cirq/vis/heatmap.py
```diff
@@ -37,8 +37,13 @@
 def _get_qubit_row_col(qubit: QubitCoordinate) -> Tuple[int, int]:
     if isinstance(qubit, grid_qubit.GridQubit):
         return qubit.row, qubit.col
-    elif isinstance(qubit, tuple):
-        return qubit[0], qubit[1]
+    elif isinstance(qubit, tuple) and len(qubit) == 2:
+        return int(qubit[0]), int(qubit[1])
+    else:
```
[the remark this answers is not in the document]
ybc1991: I agree it is odd to add raise ValueError here, so removed in the new commit.

However, I have different opinions on this:
1. I add this is not for internal coding but users. Although this is a private function, it directly applied to the user input.
2. It is unlikely but the following usage does not throw the error clearly:
image

So I kept the int() and len() check for early failure.

ybc1991:
@balopat Thanks for the review! May I ask if there is some way to log into the testing environment to debug?
I ask this because Axes.get_position() doesn't work properly. According to the test logs, the position is still the default position even fig.canvas.draw() is called explicitly. Or do you have other ideas to fix it (related to the non-interactive shell)?

balopat:
> @balopat Thanks for the review! May I ask if there is some way to log into the testing environment to debug?
> I ask this because Axes.get_position() doesn't work properly under the testing environment. According to the test logs, the position is still the default position even after fig.canvas.draw() is called explicitly.

Weird! I couldn't reproduce on either gLinux or Mac the issues! Maybe it has to do with `fig2.canvas.draw()` being ignored in a non-tty/non-UI environment? I did an experiment and pushed instead of `draw()`, I am trying `fig2.savefig('/tmp/bla.png')`...

> Or do you have other ideas to fix it (related to the non-interactive shell)?

It is very rare that we have to do this. The env is an ephemeral sandbox generated by Github Actions (probably using Docker under the hood). There are three options as I see, I typically rely on Option 0, as you can see from my push to your branch :) For the sake of completeness here they are:

1. Option 0: speed up the debug loop and focus on your tests on Github Actions. For this I'd dumb down .github/workflows/ci.yml to be just the single job running changed_files. From the tests i'd use print statements, export images and upload them as [artifacts](https://docs.github.com/en/actions/guides/storing-workflow-data-as-artifacts). 
1. Option 1: Spin up an SSH session within the job with a time limit using TMate: https://github.com/marketplace/actions/debugging-with-tmate to spin up an ssh terminal against the given job (feel free to just push a dumbed down version of the ci.yml file to your fork which has the installation script and the changed_files test job. 
1. Option 2: The other is using locally with Docker the [Act tool](https://github.com/nektos/act) - however it seems to be finicky for some reason on our setup, I'm getting a reference error:

```
act pull_request -j changed_files
WARN[0000] unable to get git repo: section "remote \"origin\"" does not exist 
[Continuous Integration/Changed files test] 🚀  Start image=node:12.6-stretch-slim
WARN[0000] unable to get git repo: section "remote \"origin\"" does not exist 
[Continuous Integration/Changed files test]   🐳  docker run image=node:12.6-stretch-slim entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[Continuous Integration/Changed files test]   🐳  docker cp src=/usr/local/google/home/balintp/dev/proj/Cirq/. dst=/github/workspace
WARN[0001] unable to get git repo: section "remote \"origin\"" does not exist 
WARN[0001] unable to get git repo: section "remote \"origin\"" does not exist 
[Continuous Integration/Changed files test] ⭐  Run actions/checkout@v2
WARN[0001] unable to get git repo: section "remote \"origin\"" does not exist 
[Continuous Integration/Changed files test]   ✅  Success - actions/checkout@v2
WARN[0001] unable to get git repo: section "remote \"origin\"" does not exist 
WARN[0002] unable to get git repo: section "remote \"origin\"" does not exist 
[Continuous Integration/Changed files test] ⭐  Run actions/setup-python@v1
[Continuous Integration/Changed files test]   ☁  git clone 'https://github.com/actions/setup-python' # ref=v1
[Continuous Integration/Changed files test] Unable to resolve v1: reference not found
[Continuous Integration/Changed files test]   ❌  Failure - actions/setup-python@v1
Error: reference not found
```

balopat:
Now you seem to have only typing failures! You can test those with `check/mypy`.

balopat:
The other weird thing is that `Changed files test` passes and the others didn't. That might also give us a hint what the difference is - as they both run the same test - supposedly on similar environments (which we can now conclude that are significantly different from this test's perspective).

balopat:
It seems `savefig()` did the trick so no need for trying to login to the envs, thankfully!

ybc1991:
Thanks for the help! Thankfully the `savefig()` did work. Although I no longer need to dig deep into the test environment, it is still super useful of your listed options for Github Actions debugging. (Hopefully, we rarely need that in the future though. Lol)

I think all tests passed now. Please take a look again. Thanks.

balopat (approved):
LGTM, thank you!

balopat:
I updated the description - it would be nice for your next PRs to have these details:  
- add a bit to describe the actual change - the content of the PR description comment becomes the body of the commit message with our automerge bot
- added "Fixes #xyz" - that sentence gets picked up by Github automatically and closes the issue when the PR is merged

PR #3690, opened 2021-01-20
Add a convenience interface that allows to issue Floquet calibration requests on cirq.google.Engine

mrwojtek:
This change adds a containers that represent Floquet calibration requests and responses. Having this abstraction is useful in multiple ways:
 - it will allow for seamless updates of the calibration protocol, without need for changes in the users' code.
 - it allows to unify interface of simulators, compilers and other tools that will potentially grow on top of this feature.

This change contains all the logic necessary to encode, decode and issue the calibration API requests.


cirq/google/calibration/phased_fsim.py
```diff
@@ -130,3 +140,217 @@ def override_by(self, other: 'PhasedFSimCharacterization') -> 'PhasedFSimCharact
             other that are not None). Otherwise the current values are used.
         """
         return other.merge_with(self)
+
+
+# TODO: Add start and end calibration timestamp
+# TODO: Add export to Panda's data frame
```
mpharrigan, line 146: open issue(s) and link here?

cirq/google/calibration/phased_fsim.py
```diff
@@ -130,3 +140,217 @@ def override_by(self, other: 'PhasedFSimCharacterization') -> 'PhasedFSimCharact
...
+class PhasedFSimCalibrationResult:
+    """The PhasedFSimGate characterization result.
+
+    Attributes:
+        parameters: Map from qubit pair to characterization result. For each pair of characterized
+            quibts a and b either only (a, b) or only (b, a) is present.
+        gate: Characterized gate for each qubit pair.
+    """
+
+    # TODO: Add validation that only either (a, b) or (b, a) is present.
```
mpharrigan, line 157: do or open and link issue

cirq/google/calibration/phased_fsim.py
```diff
@@ -130,3 +140,217 @@ def override_by(self, other: 'PhasedFSimCharacterization') -> 'PhasedFSimCharact
...
+
+# TODO: Add start and end calibration timestamp
+# TODO: Add export to Panda's data frame
+@dataclasses.dataclass(frozen=True)
+class PhasedFSimCalibrationResult:
+    """The PhasedFSimGate characterization result.
+
+    Attributes:
+        parameters: Map from qubit pair to characterization result. For each pair of characterized
+            quibts a and b either only (a, b) or only (b, a) is present.
```
mpharrigan, line 153: should we just always sort the key, i.e. canonicalize the key?

edit: I see below, that you have to call `parameters_for_qubits_swapped` to get the other ordering. This doesn't obviate the utility of storing things with a canonical key

cirq/google/calibration/phased_fsim.py
```diff
@@ -130,3 +140,217 @@ def override_by(self, other: 'PhasedFSimCharacterization') -> 'PhasedFSimCharact
...
+        gate: Characterized gate for each qubit pair.
+    """
+
+    # TODO: Add validation that only either (a, b) or (b, a) is present.
+    parameters: Dict[Tuple[Qid, Qid], PhasedFSimCharacterization]
+    gate: Gate
```
mpharrigan, line 159: why do we need `gate` as well? Isn't it a `cirq.PhasedFSimGate` always? otherwise the parameters type signature wouldn't make sense

cirq/google/calibration/phased_fsim.py
```diff
@@ -130,3 +140,217 @@ def override_by(self, other: 'PhasedFSimCharacterization') -> 'PhasedFSimCharact
...
+    @classmethod
+    def _create_parameters_dict(
+        cls,
+        parameters: List[Tuple[Qid, Qid, PhasedFSimCharacterization]],
+    ) -> Dict[Tuple[Qid, Qid], PhasedFSimCharacterization]:
+        """Utility function to create parameters from JSON.
+
+        Can be used from child classes to instantiate classes in a _from_json_dict_
+        method."""
+        return {(entry[0], entry[1]): entry[2] for entry in parameters}
```
mpharrigan, line 179: use tuple unpacking to make this more readable

```python
return {(a, b): char for a, b, char in parameters}
```

cirq/google/calibration/phased_fsim.py
```diff
@@ -130,3 +140,217 @@ def override_by(self, other: 'PhasedFSimCharacterization') -> 'PhasedFSimCharact
...
+    """Description of the request to characterize PhasedFSimGate.
+
+    Attributes:
+        pairs: Set of qubit pairs to characterize. A single qubit can appear on at most one pair in
+            the set.
+        gate: Gate to characterize for each qubit pair from pairs. This must be a supported gate
+            which can be described cirq.PhasedFSim gate.
+    """
+
+    # TODO: Validate that each pair is unique and non-overlaping with any other pair.
```
mpharrigan, line 215: do or canonicalize or open an issue

cirq/google/calibration/phased_fsim.py
```diff
@@ -130,3 +140,217 @@ def override_by(self, other: 'PhasedFSimCharacterization') -> 'PhasedFSimCharact
...
+        cls,
+        parameters: List[Tuple[Qid, Qid, PhasedFSimCharacterization]],
+        gate: Gate,
+        **kwargs,
+    ) -> 'PhasedFSimCalibrationResult':
+        """Magic method for the JSON serialization protocol.
+
+        Converts serialized dictionary into a dict suitable for
+        class instantiation."""
+        return cls(cls._create_parameters_dict(parameters), gate, kwargs['options'])
```
mpharrigan, line 271: can't you change the superclass's implementation of `_from_json_dict_` to pass all extra `kwargs` to the `cls()` constructor call? Then you wouldn't have to override it here.

One thing to note: You probably have to pop `'cirq_type'` from `kwargs` first.

cirq/google/calibration/phased_fsim.py
```diff
@@ -130,3 +140,217 @@ def override_by(self, other: 'PhasedFSimCharacterization') -> 'PhasedFSimCharact
...
+
+@dataclasses.dataclass(frozen=True)
+class FloquetPhasedFSimCalibrationResult(PhasedFSimCalibrationResult):
+    """PhasedFSim characterization result specific to Floquet calibration.
+
+    Attributes:
+        options: Options of the characterization from the request.
+    """
+
+    options: FloquetPhasedFSimCalibrationOptions
```
mpharrigan, line 258: Can you explain why we're saving `options` in the `...Result` as well? I understand why this is essential for the `...Request`. Is this just to simplify recordkeeping?

mpharrigan: If the result type can be more general for a multiple calibration requests, it could make swapping between different calibration routines easier. Instead of class hierarchy, consider using something like an `Optional[Any]` `metadata` field if you really want to tag results with extra info

cirq/google/calibration/phased_fsim.py
```diff
@@ -130,3 +140,217 @@ def override_by(self, other: 'PhasedFSimCharacterization') -> 'PhasedFSimCharact
...
+                'est_theta': self.options.characterize_theta,
+                'est_zeta': self.options.characterize_zeta,
+                'est_chi': self.options.characterize_chi,
+                'est_gamma': self.options.characterize_gamma,
+                'est_phi': self.options.characterize_phi,
+                'readout_corrections': True,
+            },
+        )
+
+    # TODO: Handle unsuccessful calibrations and throw appropriate exceptions.
```
mpharrigan, line 306: do, or open an issue and link here.

cirq/google/calibration/phased_fsim.py
```diff
@@ -130,3 +140,217 @@ def override_by(self, other: 'PhasedFSimCharacterization') -> 'PhasedFSimCharact
...
+        return CalibrationLayer(
+            calibration_type=_FLOQUET_PHASED_FSIM_HANDLER_NAME,
+            program=circuit,
+            args={
+                'est_theta': self.options.characterize_theta,
+                'est_zeta': self.options.characterize_zeta,
+                'est_chi': self.options.characterize_chi,
+                'est_gamma': self.options.characterize_gamma,
+                'est_phi': self.options.characterize_phi,
+                'readout_corrections': True,
```
mpharrigan, line 302: any reason why this isn't part of `self.options`? Can we change either the fields of the `..Options` or the API arg-names so this can be something like `args=asdict(self.options)`?

cirq/google/calibration/phased_fsim.py
```diff
@@ -130,3 +140,217 @@ def override_by(self, other: 'PhasedFSimCharacterization') -> 'PhasedFSimCharact
...
+        gate: Gate,
+        pairs: List[Tuple[Qid, Qid]],
+        options: FloquetPhasedFSimCalibrationOptions,
+        **kwargs,
+    ) -> 'PhasedFSimCalibrationRequest':
+        """Magic method for the JSON serialization protocol.
+
+        Converts serialized dictionary into a dict suitable for
+        class instantiation."""
+        instantiation_pairs = tuple((entry[0], entry[1]) for entry in pairs)
```
mpharrigan, line 346: why not `tuple(pairs)`? is `entry` larger than two entries? maybe tuple unpacking could make this explicit (if it's true)

```python
tuple((a, b) for a, b, _ in pairs)
```

cirq/google/calibration/workflow.py
```diff
@@ -0,0 +1,49 @@
...
+from cirq.google.engine import Engine
+from cirq.google.serializable_gate_set import SerializableGateSet
+
+
+def run_characterizations(
+    calibrations: List[PhasedFSimCalibrationRequest],
+    engine: Engine,
+    processor_id: str,
+    gate_set: SerializableGateSet,
+) -> List[PhasedFSimCalibrationResult]:
```
mpharrigan, line 29: Do we have an object that contains `(engine, processor_id, gate_set)` that we can use instead of having the user have to specify all of these? `QuantumEngineSampler` solves this for running programs but I don't know if it has a `run_calibration` method. We should add this method and then accept an `EngineSampler`.

In general, I think we should be encouraging the use of the `cirq.google.get_engine_sampler` family of functions. For sampling, it lets you sub in a simulator (which is nice) but even for things like this where there isn't a comparable simulator method (side note: maybe there should be a simulator path for `run_calibration` for users to test that it's working) it is useful to group these three objects together. Furthermore, the `get_engine_xx` functions can grab the `project_id` from environment variables, making it easier to version control and distribute demos without a hardcodded project id

mpharrigan (commented):
Some nits. Additional thoughts

 - I'm most concerned about the `run_characterizations` function signature, specifically letting you do it without a "bare `Engine`" and potentially letting you do it in a way that a simulator could be swapped in to test your code before getting engine access.
 - The class hierarchy makes my eyes glaze over a bit, but maybe it's irreducible complexity inherent in the task. We'll likely need some further convenience functions for making `FlorquetPhasedFSimCalibrationRequest`s more easily

cirq/google/calibration/phased_fsim.py
mrwojtek: I put Gate becuase it can be Union[ISwapPowGate, PhasedFSimGate, FSimGate, PhasedISwapPowGate], anything which can serialize to a compatible gate.

I guess we could make a decision to support only PhasedFSimGate here but I guess this is a broader question on what is the strategy when using gates in Cirq?

cirq/google/calibration/phased_fsim.py
mrwojtek: Moved all the TODOs to a separate doc for now.

cirq/google/calibration/phased_fsim.py
mrwojtek: Canonalization is not a requirement for QE Floquet calibration so I didn't enforce it here. This way it is more flexible if people would create requests on their own. I'm not particularly attached to that decision tough.

cirq/google/calibration/phased_fsim.py
mrwojtek: Moved all the TODOs to a separate doc for now.

cirq/google/calibration/phased_fsim.py
mrwojtek: Done.

cirq/google/calibration/phased_fsim.py
mrwojtek: Moved to separate doc with TODO

cirq/google/calibration/phased_fsim.py
mrwojtek: Moved to a separate doc

cirq/google/calibration/phased_fsim.py
mrwojtek: Done.

cirq/google/calibration/phased_fsim.py
mrwojtek: This is experimental feature which is not ready to be played with yet.

cirq/google/calibration/phased_fsim.py
mrwojtek: Done

cirq/google/calibration/phased_fsim.py
mrwojtek: Currently options are almost trivial but in the future they'll potentially grow with gauges to manipulate characterization accuracy as well as  server-based calibrations. My intention was to keep all the the necessary context that describes the data with data itself.

As for class hierarchy: my idea was to share everything which is common to all the characterization techniques in a base class. In that way features which rely only on this generic data (like single-qubit phased calibration presented in a follow-up PR) will work out of the box for all the calibrations. On the other hand more sophisticated features (for example engine simulator presented in a follow-up) might use the specialized "metadata" from the derived class. In case of XEB, it would be nice to put the estimated fidelity there. I was thinking about putting more measures related to the estimation accuracy as well (I'm not sure how much it can be shared between different techniques).

Although I'm not a particular fan of "metadata" because of its lack of structure, I guess it's fine to do it like this. We can re-introduce the class structure any time in the future if it would become necessary

Let me know what you think about this?

cirq/google/calibration/workflow.py
mrwojtek: I agree with all what you just said but can that be done in a follow-up after we have a basic features working?

As for simulator for run_calibrations I have already introduced it as one of the follow-up PRs. It is a necessary thing to be able to do unit-testing of the calibration functionality. It might also have other uses in the future.

mrwojtek:
> Some nits. Additional thoughts
> 
> * I'm most concerned about the `run_characterizations` function signature, specifically letting you do it without a "bare `Engine`" and potentially letting you do it in a way that a simulator could be swapped in to test your code before getting engine access.
Does https://github.com/mrwojtek/Cirq/pull/4/files#diff-dffa1a181ce775202a7bb1952aed024378cab37e35610f3dbf6ff167589e3e9eR220
solves your problems?

> * The class hierarchy makes my eyes glaze over a bit, but maybe it's irreducible complexity inherent in the task. We'll likely need some further convenience functions for making `FlorquetPhasedFSimCalibrationRequest`s more easily
I found it the most reasonable thing to do. I guess it could be done with protocols as well but I think that would grow this feature a lot. I'm not a python expert...

mrwojtek:
> Some nits. Additional thoughts
> 
> * I'm most concerned about the `run_characterizations` function signature, specifically letting you do it without a "bare `Engine`" and potentially letting you do it in a way that a simulator could be swapped in to test your code before getting engine access.
> * The class hierarchy makes my eyes glaze over a bit, but maybe it's irreducible complexity inherent in the task. We'll likely need some further convenience functions for making `FlorquetPhasedFSimCalibrationRequest`s more easily

Btw, as for the second point, this is what is in the next PR:
https://github.com/mrwojtek/Cirq/pull/3/files#diff-dffa1a181ce775202a7bb1952aed024378cab37e35610f3dbf6ff167589e3e9eR107

cirq/google/calibration/phased_fsim.py
mpharrigan: ack

cirq/google/calibration/phased_fsim.py
mpharrigan: But what is it actually used for? Is it possible to have `PhasedFSimCharacterization` with 5 values but `gate` is ISwapPowGate? Also this is a continuation of the comment below where I think `gate` in the `..Result` object is just to preserve the input context? It doesn't actually change the results, right?

cirq/google/calibration/phased_fsim.py
```diff
@@ -130,3 +140,200 @@ def override_by(self, other: 'PhasedFSimCharacterization') -> 'PhasedFSimCharact
...
+
+    Attributes:
+        pairs: Set of qubit pairs to characterize. A single qubit can appear on at most one pair in
+            the set.
+        gate: Gate to characterize for each qubit pair from pairs. This must be a supported gate
+            which can be described cirq.PhasedFSim gate.
+    """
+
+    pairs: Tuple[Tuple[Qid, Qid], ...]
+    gate: Gate  # Any gate which can be described by cirq.PhasedFSim
```
mpharrigan, line 213: What if `gate` is `ISwapPowGate` and you set `characterize_zeta` to `True`? How does this `gate` actually change the behavior of the algorithm?

cirq/google/calibration/phased_fsim.py
mpharrigan: If the intent is to keep the necessary context with the data, that's fine. We'll have to see how that works in practice, though. Should the parent result class have an abstract property `options`, and derived classes have to fill it in with their specific `Options` type? It would make reading the class hierarchy easier, at the least.

cirq/google/calibration/phased_fsim.py
mpharrigan: ok. It would be nice for future experimental features if the `args` was "forwards compatible" so one doesn't have to sneak in this option under-the-hood

cirq/google/calibration/workflow.py
mpharrigan: sounds good

mpharrigan (approved):
mainly still baffled by the `gate` argument, but lgtm

cirq/google/calibration/phased_fsim.py
mrwojtek: Now I see what you mean. `Gate` is anything which is compatible and handled by the `SerializableGateSet` used for the engine execution. Since the `SerializableGateSet` accept many forms of a gate, I copied this logic here. This is reminiscence of the logic adopted in Cirq where we don't really have a notion of a "hardware gate", instead we say for example `criq.CZ` and this means: run the approximate `cirq.CZ` gate on this qubit pair. If we have a notion of hardware gate which is not fixed to any specific parameters but is just an instruction to do, that might be maybe less confusing. However, it is also questionable.

Answering your questions:
 - It is perfectly fine to have all 5 parameters in `PhasedFSimCharacterization` for `ISwapPowGate` because `ISwapPowGate` is just an approximation to the perfect gate, and `PhasedFSimCharacterization` tells you distance to this approximation in this particular context.
 -  `gate` in `...Result` is just for context, right.

At some point options will not only have requests to characterize but also to calibrate, in which case some of the angles will be automatically fixed to perfect values, and in the end `PhasedFSimCharacterization` will be a perfect gate that was requested for.

cirq/google/calibration/phased_fsim.py
mrwojtek: `gate` is just a selector of which hardware gate should be used for characterization. Our processor have a limited choice of those gates.

cirq/google/calibration/phased_fsim.py
mrwojtek: It might not be so easy to introduce this property with dataclass and inheritance...

Maybe something like this:
 - I'll define PhasedFSimCalibrationOptions abstract class, empty
 - I'll remove FloquetPhasedFSimCalibrationResult so there will be just PhasedFSimCalibrationResult with option field of type PhasedFSimCalibrationOptions, Each calibration will have its own specialization of Options?
 
This way a type of calibration that was performed will be encoded in the options field, not in the class instance type.

cirq/google/calibration/phased_fsim.py
mrwojtek: I take back one statement above: in the PR that I'm doing right now I'll be using gate parameter to verify that the circuit being compiled is compatible with characterization.

cirq/google/calibration/phased_fsim.py
mpharrigan: What do you think? It's not important enough for me to block this PR; although your suggestion sounds very reasonable

cirq/google/calibration/phased_fsim.py
mpharrigan: So this can basically be `cirq.ISWAP**0.5` or `cg.SYC`? Maybe `cirq.ISWAP` as well? And it is used in construction of the moment calibration circuits. Does it also seed the algorithm with the correct "ideal" values for the angles?

cirq/google/calibration/phased_fsim.py
mpharrigan: Good point about the lack of the "hardware gate" notion. I would have thought that the `PhasedFSim...` family of classes could *implicitly* specify the hardware gate (specifically "PhasedFSimGate"), but I guess there's a couple of hardware gates (sqrtiswap, syc, iswap, ...) that all fit under that umbrella so we need to request a particular "ideal" gate.

cirq/google/calibration/phased_fsim.py
mrwojtek: Done (as per offline discussion).

mrwojtek:
Hi all, I'll merge this soon if there are no more comments from anyone.

mpharrigan (approved):

cirq/google/calibration/phased_fsim.py
mrwojtek: Also `cirq.CZ`. It is used in the construction for the moments and seeds the algorithm with the initial guess, that's correct.

cirq/google/calibration/phased_fsim.py
mrwojtek: Even `FSimGate` covers all the hardware gates we support. The problem is that depending how people create the circuits, they can be in many forms. We don't have a notion of optimizer/compiler which could normalize the gate used so I need to keep all the options in this API.

Issue #3691, opened 2021-01-20
ProductState multiplication mutates input state


rmlarose:
**Description of the issue**

In the expression `a ⊗ b` using [`cirq.ProductState`](https://github.com/quantumlib/Cirq/blob/master/cirq/value/product_state.py#L51), `a` gets modified to `a ⊗ b`.

**How to reproduce the issue**

Shown for `cirq.KET_PLUS`, but the behavior is the same for any of `cirq.PAULI_STATES`.

```python
import cirq

q0, q1 = cirq.LineQubit.range(2)

plus0 = cirq.KET_PLUS(q0)
plus1 = cirq.KET_PLUS(q1)

product = plus0 * plus1

print(plus0)  # Not expected behavior
# +X(0) * +X(1)

print(plus1)  # Expected behavior
# +X(1)

print(product)  # Expected behavior
# +X(0) * +X(1)
```

**Cirq version**

0.10.0.dev

A potential fix is copying on [this line](https://github.com/quantumlib/Cirq/blob/master/cirq/value/product_state.py#L83) instead of assigning.

balopat:
Thanks @rmlarose - good catch! I think that this is a real bug and the recommended fix seems to make sense to me!

vtomole:
@rmlarose It seems like i can't assign you as reviewer. Please have a look at https://github.com/quantumlib/Cirq/pull/3692.

rmlarose:
> @rmlarose It seems like I can't assign you as reviewer. Please have a look at #3692.

Thanks @vtomole for the quick fix! +1 on Matt's suggestion for a test, then I offer an unauthorized LGTM :-)

PR #3692, opened 2021-01-20
Copy instead of assign input state for ProductState multiplication

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/3691


mpharrigan:
Can we add a test that would have failed with the previous behavior?

vtomole:
@mpharrigan Done.

balopat (approved):

balopat:
Thanks for the fix!!!

PR #3694, opened 2021-01-20
Add tags to MeasureInfo

dstrain115:
- When we call cirq.google.api.v2.find_measurements, we need a way
to retrieve tags that were originally on the circuit.
- This adds tags to the NamedTuple that is returned by this function
that includes tags.

This does change the output of this function, but, since it is an
additive change, it should be minimally breaking.

### DIFF
--- a/cirq/google/api/v2/results.py
+++ b/cirq/google/api/v2/results.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 from typing import (
     cast,
     Dict,
@@ -19,13 +18,12 @@ from typing import (
     Iterable,
     Iterator,
     List,
-    NamedTuple,
     Optional,
     Set,
     TYPE_CHECKING,
 )
-
 from collections import OrderedDict
+import dataclasses
 import numpy as np
 
 from cirq.google.api import v2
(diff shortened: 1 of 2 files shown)


balopat (approved):
LGTM

balopat:
> since it is an additive change, it should be minimally breaking.
I think so too, especially that `NamedTuple`s cannot be unpacked (e.g. `a,b,c,d = v2.MeasurementInfo` is invalid), users have to rely on field, thus old code will work with the new MeasurementInfo.

maffoo:
@balopat, `NamedTuple` actually _can_ be unpacked (it's just a typed wrapper around [`collections.namedtuple`](https://docs.python.org/3/library/collections.html#collections.namedtuple). I'd actually suggest changing this to be a `dataclasses.dataclass`.

mpharrigan:
+1 use dataclasses wherever you're inclined to use namedtuples

dstrain115:
SGTM, I don't like NamedTuples anyway.  Changed to a dataclass.
This should only be as breaking as the original PR as long as people aren't doing any inspection of the internals of the class, it should be fine.  Also, it is in cirq.google.

balopat:
> @balopat, `NamedTuple` actually _can_ be unpacked (it's just a typed wrapper around [`collections.namedtuple`](https://docs.python.org/3/library/collections.html#collections.namedtuple)). I'd actually suggest changing this to be a `dataclasses.dataclass`.

Huh, they really can, I must have overlooked something and made a mistake when I tried it out. TIL!

balopat:
I added the BREAKING_CHANGE label for warning in the release notes.

PR #3697, opened 2021-01-22
Add Circuit.are_any_matches_terminal()

95-martin-orion:
Exactly what it says in the title. This is the "any" analog to `Circuit.are_all_matches_terminal()`; it will be used for the expectation value API ([RFC](https://tinyurl.com/cirq-expectation-value-api)).

### DIFF
--- a/cirq/circuits/circuit.py
+++ b/cirq/circuits/circuit.py
@@ -744,11 +744,11 @@ class AbstractCircuit(abc.ABC):
     def has_measurements(self):
         return any(self.findall_operations(protocols.is_measurement))
 
-    def are_all_measurements_terminal(self):
+    def are_all_measurements_terminal(self) -> bool:
         """Whether all measurement gates are at the end of the circuit."""
         return self.are_all_matches_terminal(protocols.is_measurement)
 
-    def are_all_matches_terminal(self, predicate: Callable[['cirq.Operation'], bool]):
+    def are_all_matches_terminal(self, predicate: Callable[['cirq.Operation'], bool]) -> bool:
         """Check whether all of the ops that satisfy a predicate are terminal.
 
         This method will transparently descend into any CircuitOperations this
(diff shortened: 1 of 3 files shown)


cirq/circuits/circuit.py
```diff
@@ -792,6 +792,54 @@ def get_op_circuit(op: ops.Operation) -> Optional['cirq.FrozenCircuit']:
                     return False
         return True
 
+    def are_any_measurements_terminal(self):
+        """Whether any measurement gates are at the end of the circuit."""
+        return self.are_any_matches_terminal(protocols.is_measurement)
+
+    def are_any_matches_terminal(self, predicate: Callable[['cirq.Operation'], bool]):
```
balopat, line 799: nit: return type

cirq/circuits/circuit.py
```diff
@@ -792,6 +792,54 @@ def get_op_circuit(op: ops.Operation) -> Optional['cirq.FrozenCircuit']:
                     return False
         return True
 
+    def are_any_measurements_terminal(self):
```
balopat, line 795: nit: return type

cirq/circuits/circuit.py
```diff
@@ -792,6 +792,54 @@ def get_op_circuit(op: ops.Operation) -> Optional['cirq.FrozenCircuit']:
...
+
+        Returns:
+            Whether or not any `Operation` s in a circuit that satisfy the
+            given predicate are terminal. Also checks within any CircuitGates
+            the circuit may contain.
+        """
+        from cirq.circuits import CircuitOperation
+
+        # TaggedOperations can wrap CircuitOperations.
+        def get_op_circuit(op: ops.Operation) -> Optional['cirq.FrozenCircuit']:
```
balopat, line 817: I think we can do better in terms of deduping this and the `are_all_matches_terminal`, e.g. `get_op_circuit` could be pulled out to a shared private method. 
Not sure about the iteration over the moments is worth to dedupe but maybe

balopat (approved):
LGTM with nits

cirq/circuits/circuit.py
95-martin-orion: Done (+`are_all_*`).

cirq/circuits/circuit.py
95-martin-orion: Done (+`are_all_*`).

cirq/circuits/circuit.py
95-martin-orion: Extracted `get_op_circuit`. Moment iteration probably isn't worth it, especially since the `any` version runs faster by iterating over the moments in reverse. I've reverse the `any` iterator in this commit.

Issue #3492, opened 2020-11-10
Expectation value API for simulators


95-martin-orion:
**Is your feature request related to a use case or problem? Please describe.**

Currently, finding expectation values from a simulator in Cirq requires implementing the `SimulatesFinalState` interface and using the resulting state vector to get the expectation values. If simulators provide a more efficient way of producing expectation values, we should prefer that instead.

**Describe the solution you'd like**

In addition to `SimulatesFinalState`, `SimulatesAmplitudes`, etc., Cirq should offer a `SimulatesExpectationValues` interface for simulators that generate expectation values.

**What is the urgency from your perspective for this issue? Is it blocking important work?**

P2 - we should do it in the next couple of quarters. The current workaround is to continue using `SimulatesFinalState`, but as noted above this may be inefficient.

balopat:
Discussed on Cirq Cynq: this is a good idea, needed and `PauliSum`s are the way to go.

95-martin-orion:
Further discussion on the qsim weekly sync identified that this is a proper prerequisite for trajectory-based Monte Carlo simulations; using `SimulatesFinalState` and processing each result in turn would reduce performance to the point of being infeasible.

95-martin-orion:
Moving forward with the format discussed in the RFC: https://tinyurl.com/cirq-expectation-value-api

PR #3698, opened 2021-01-22
SimulatesExpectationValues

95-martin-orion:
Stage 1 of the Expectation Value API (#3492).

This PR introduces the `simulates_expectation_values` interface and implements it for the `sparse_simulator`. This interface unblocks integration with the qsim trajectory simulator.

Next step on the Cirq side is to add the `samples_expectation_values` interface in Sampler.


cirq/sim/simulator.py
```diff
@@ -173,6 +173,103 @@ def compute_amplitudes_sweep(
...
+            observables: A map of observable names (strings) to observables.
+            param_resolver: Parameters to run with the program.
+            qubit_order: Determines the canonical ordering of the qubits. This
+                is often used in specifying the initial state, i.e. the
+                ordering of the computational basis states.
+            initial_state: The initial state for the simulation. The  form of
+                this state depends on the simulation implementation.  See
+                documentation of the implementing class for details.
+            permit_terminal_measurements: If the provided circuit ends in a
+                measurement, this method will generate an error unless this
```
balopat, line 213: nit: optional plural? 
```suggestion
                measurement(s), this method will generate an error unless this
```

cirq/sim/simulator.py
```diff
@@ -173,6 +173,103 @@ def compute_amplitudes_sweep(
...
+        expectation values by sampling multiple times.
+
+        Args:
+            program: The circuit to simulate.
+            observables: A map of observable names (strings) to observables.
+            param_resolver: Parameters to run with the program.
+            qubit_order: Determines the canonical ordering of the qubits. This
+                is often used in specifying the initial state, i.e. the
+                ordering of the computational basis states.
+            initial_state: The initial state for the simulation. The  form of
```
balopat, line 209: nit: extra space
```suggestion
            initial_state: The initial state for the simulation. The form of
```

cirq/sim/simulator.py
```diff
@@ -173,6 +173,103 @@ def compute_amplitudes_sweep(
...
+                this state depends on the simulation implementation.  See
```
balopat, line 210:
```suggestion
                this state depends on the simulation implementation. See
```

cirq/sim/simulator.py
```diff
@@ -173,6 +173,103 @@ def compute_amplitudes_sweep(
...
+    expectation values for each observable at the end of the circuit.
+
+    Implementors of this interface should implement the
+    simulate_expectation_values_sweep method.
+    """
+
+    def simulate_expectation_values(
+        self,
+        program: 'cirq.Circuit',
+        observables: Dict[str, 'cirq.PauliSum'],
```
balopat, line 189: huh, I didn't catch this in the design doc - why this vs just a list and the same ordering in the returned value which would be a list?

cirq/sim/sparse_simulator.py
```diff
@@ -232,6 +251,39 @@ def _base_iterator(
...
+        params: 'study.Sweepable',
+        qubit_order: ops.QubitOrderOrList = ops.QubitOrder.DEFAULT,
+        initial_state: Any = None,
+        permit_terminal_measurements: bool = False,
+    ) -> List[Dict[str, float]]:
+        if not permit_terminal_measurements and program.are_any_measurements_terminal():
+            raise ValueError(
+                'Provided circuit has terminal measurements, which may '
+                'skew expectation values. If this is intentional, set '
+                f'permit_terminal_measurements=True.\n\nCircuit:\n{program}'
```
balopat, line 267: Circuits can be very big. I'd recommend just printing the terminal measurements (moment, qubit). 
```
terminal measurements: 
 - Moment: 3,  qubits: "(0,1)", "(0,2)"
 - Moment: 5, qubits: "(5,6)"
```

balopat (changes requested):
Looking good - did a first skim, found some nits!

cirq/sim/sparse_simulator.py
95-martin-orion: This information isn't returned by `are_any_matches_terminal`. If it isn't strictly necessary, I'd prefer to just drop the circuit from the error message.

cirq/sim/simulator.py
95-martin-orion: I updated the doc after some discussion with Matt - in short, this is intended to make finding a result for a given observable less confusing, particularly in the case of `simulate_expectation_values_sweep`.

cirq/sim/sparse_simulator.py
balopat: SGTM

cirq/sim/simulator.py
balopat: But now the user has to come up with a naming scheme for a potentially large list of PauliSums. Why does the sweep version matter? Because it would be a list of lists?

95-martin-orion: That was the point of concern, yes. Pinging @mpharrigan for his take on this.

mpharrigan: I raised a concern about list-of-lists because it's not clear which should be nested first and reading the code or type signature wouldn't help very much. The issue--of course--is that sweeps and the collections of observables both serve the same role: "Hey take this circuit [ansatz] and monkey with it". With sweeps, you can explicitly outer-product two together. Here, it's implicit. Lots of words to say I'm not the biggest fan of list-of-lists.

That being said, it would be a huge nightmare to have to come up with keys for this. On the design doc comment, there was some initial discussion about having the return value be a pandas dataframe where the sweep values and paulisum observables could coexist as columns. I don't know if there's much appetite for that. It might behoove us to think about user stories for doing a sweep of parameters on many observables. When I was making the examples for the nisq-observable-estimation routines a dataframe felt very natural but I'm a big pandas person

cirq/sim/simulator.py
```diff
@@ -173,6 +173,103 @@ def compute_amplitudes_sweep(
...
+            qubit_order,
+            initial_state,
+            permit_terminal_measurements,
+        )[0]
+
+    @abc.abstractmethod
+    def simulate_expectation_values_sweep(
+        self,
+        program: 'cirq.Circuit',
+        observables: Dict[str, 'cirq.PauliSum'],
```
mpharrigan, line 238: can we expand the set of types we accept as observables. Assuming we nix the dict thing, You should be able to seamlessly pass

 - list of paulisum
 - one paulisum -> wrap in list of length 1
 - list of paulistring -> wrap in `n` paulisums
 - one paulistring -> wrap in 1 paulisium

95-martin-orion: Should be easy enough with `List[PauliSumLike]` and `PauliSum.wrap`. Waiting on the dict/list decision to make this change.

cirq/sim/simulator.py
95-martin-orion: One note: I had suggested `pandas.DataFrame` for the Sampler API, since other Sampler methods also return `pandas`. It could work for simulator as well, but there are (currently) no simulator methods using `pandas`.

Rough list of options (likely incomplete):
1. Keep dict -> list-of-dicts.
    a. Clarifies nesting, but adds the difficulty of key creation. (looks like `results[1]['ZXZ']`)
1. Use dict -> `pandas.DataFrame`.
     a. Adds pandas convenience methods, but contrasts with simulator methods. (looks like...?)
1. Revert to list -> list-of-lists.
     a. Easier to create input, less clarity in output. (looks like `results[1][3]`)
1. Use type aliases, e.g. `ExpectationValues = List[float]`
     a. Clarifies function signature, but not return-value usage (would still look like `result[1][3]`)
1. Use custom return types, e.g.

    ```
    @dataclass
    class ExpectationValues:
        evs: List[float]
    ```
    a. Clarifies return-value usage, but may add clutter (looks like `result[1].evs[3]`)

Thoughts? If the type signature is the main concern, option 4 is the cheapest solution.

mpharrigan: I'd say (3) or (2). The others add a lot of bookkeeping without really helping all that much

95-martin-orion: SGTM. To avoid the key-generation nightmare, I'll go ahead with (3). Opened #3708 to discuss potentially shifting simulators to `pandas`.

cirq/sim/simulator.py
95-martin-orion: Done.

cirq/sim/sparse_simulator.py
95-martin-orion: Done.

balopat (approved):
LGTM

cirq/sim/simulator.py
balopat: I'm comfortable with (3) as it's simple - but I understand the ambiguity that it introduces, and the mental bookkeeping the user has to do - they'll have to remember that sweeps are the outer index. 

I like the idea of (2) - adds more power, more structure and consistency. 

My OO java background would go for (5) to avoid [primitive obsession](https://dzone.com/articles/code-quality-fighting-primitive-obsession-code-sme-1) - it would force the user to write something like: `result.sweeps[0].evs[1]` but I'm not sure that that's "pythonic" :)

Issue #2623, opened 2019-12-05
Support sample-based expectation in PauliSum


jaeyoo:
It is very important to support sample-based expectation in PauliSum for real & simulating QPU devices. Given `num_shots` (or we can name it as `repetitions`. which one do you prefer?), the result of cirq.Sampler().run() is a list of qubit measurements. Based on the result, a helper to calculate (estimated) expectation value of given PauliSum is really nice to have.

I suggest `cirq.PauliSum.expectation_from_samples()` and the usage is identical to that of `cirq.PauliSum.expectation_from_wavefunction()` and `cirq.PauliSum.expectation_from_density_matrix()` except input signature has measurements rather than final_state or final_density_matrix. It also output float expectation value. 

It is assumed that the given measurements have the shape `[num_shots, num_qubits]` for example. so the function `expectation_from_samples()` doesn't need to have `num_shots` as an additional input.

jaeyoo:
I want to work on it.

trevormccrt:
@mpharrigan

trevormccrt:
This is more complicated than it looks. I'd say someone with some experimental quantum computing experience should do this.

mpharrigan:
I'm putting the finishing touches on a design doc for a full solution to executing circuits to estimate a bunch of observables. That may be more than what you're asking for here. Specifically

> I suggest cirq.PauliSum.expectation_from_samples() and the usage is identical to that of cirq.PauliSum.expectation_from_wavefunction() and cirq.PauliSum.expectation_from_density_matrix() except input signature has measurements rather than final_state or final_density_matrix. It also output float expectation value.

This is a rather modest request. The difficulty is that the input measurements need to somehow be guaranteed to have come from modified circuits that include pre-measurement single-qubit change-of-basis rotations. This is generally the challenging part.

peterse:
Gonna leave a few comments/notes that came up from a cirq meeting months ago:
 - minimizing the number of terms required for a given pauli-sum expectation is [NP hard](https://arxiv.org/pdf/1907.07859.pdf)
- Previous work:
https://arxiv.org/abs/1907.13623)
 https://arxiv.org/abs/1908.05628 (includes table of similar work)
 - This seems like something that should be done with respect to gate sets (e.g. readout circuits in the first link are composed from CNOT, which is may not be a good choice for some hardware)

jaeyoo:
[...]

PR #3700, opened 2021-01-25
Add Sampler API for expectation values

95-martin-orion:
Stage 2 of the Expectation Value API (#3492).

This PR introduces the sample_expectation_values method and its default implementation. Hardware and software samplers can use this method as-is or implement their own version for custom behavior (e.g. server-side calculation of expectation values).

Question for reviewers: would a `pandas.DataFrame` be more convenient as output from this method?


95-martin-orion:
TODO: update to reflect discussion from #3698.

95-martin-orion:
> TODO: update to reflect discussion from #3698.

Done. Returned value is now a list-of-lists. `result[x][y]` contains the `y`th observable on the `x`th sweep.

cirq/work/sampler.py
```diff
@@ -133,6 +134,55 @@ def sample(
...
+            # Aggregate results from all samples for the same parameter values
+            for __ in range(num_samples):
+                curr_row = next(row_iter)
+                for key in mto:
+                    state_vector_map[key][getattr(curr_row, key)] += 1
+            # Compute estimated expectation values from aggregated results
+            sweep_result = []
+            for key, observables in mto.items():
+                norm = np.linalg.norm(state_vector_map[key])
+                state_vector_map[key] /= norm
```
balopat, line 178: Hmm...I don't understand this calculation. You are trying to construct a full statevector based on potentially partial measurements...

The problems that I see here is
1. nothing guarantees that the partial measurement of the selected qubits is going to add up to 1 probability 
2. even if we assume that the chosen subset of amplitudes fill up the whole state vector, we'd have to do an inverse Born rule kind of thing: 
```
# sum of observed frequencies for each bitstring
norm = state_vector_map[key].sum()
# relative probabilities 
state_vector_map[key] /= norm
# to amplitudes 
state_vector_map[key] = np.sqrt(state_vector_map[key])
```

However, I think we should go about this differently than reusing the `obs.expectation_from_state_vector` as we don't have a statevector/density matrix here. Is there a way to tie this together with the Observable Measurements functionality in cirq.work? Because that uses a Sampler and inserts the right type of rotations before measurements...

cc @mpharrigan

balopat (changes requested):

cirq/work/sampler.py
```diff
@@ -133,6 +134,55 @@ def sample(
...
         return pd.concat(results)
 
+    def sample_expectation_values(
+        self,
+        program: 'cirq.Circuit',
+        measurement_to_observables: Dict[
+            str, Union['cirq.PauliSumLike', List['cirq.PauliSumLike']]
+        ],
+        *,
+        num_samples: int,
```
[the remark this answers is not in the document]
95-martin-orion: I would argue that these arguments are conceptually distinct: 
- `repetitions` controls the **number of results** produced by `run` or `sample`.
- `num_samples` controls the **accuracy** of the expectation value estimate.

cirq/work/sampler.py
95-martin-orion: You're right about the "inverse Born rule" step vs using the norm - I guess I missed this because all of my tests have uniform amplitudes. I'll commit a fix for it shortly.

The intent here is to construct a frequentist approximation of the state vector for the subset of qubits measured at each measurement event. Phase information is lost, but that's fine since we don't need it for estimating expectation values.

`state_vector_map[key]` may not be a "true" state vector, but it approximates one. Since `expectation_from_state_vector` does exactly what we want, I don't see why we wouldn't use it here unless there's an optimization possible for the pre-normalization state histogram.

95-martin-orion: Relevant: #2623

cirq/work/sampler_test.py
```diff
@@ -132,6 +132,118 @@ def test_sampler_sample_inconsistent_keys():
...
+def test_sampler_simple_sample_expectation_values():
+    a = cirq.LineQubit(0)
+    sampler = cirq.Simulator()
+    circuit = cirq.Circuit(cirq.H(a), cirq.measure(a, key='out'))
+    obs = cirq.X(a)
+    results = sampler.sample_expectation_values(circuit, {'out': obs}, num_samples=10000)
+
+    # Precision of non-Z-basis expectation values is dependent on num_samples.
+    # To keep tests short, low precision is used here.
+    assert np.allclose(results, [[1]], atol=0.01)
```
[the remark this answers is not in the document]
95-martin-orion: The level of error was partially due to my flawed EV calculation scheme, but your point stands. Swapped in a deterministic-but-imperfect "sampler" for this test.

cirq/work/sampler.py
95-martin-orion: Added a new test with more imbalanced amplitudes and verified that it fails with 2-norm and passes with inverse Born rule.

cirq/work/sampler.py
```diff
@@ -133,6 +134,55 @@ def sample(
...
+        params: 'cirq.Sweepable' = None,
+    ) -> List[List[float]]:
```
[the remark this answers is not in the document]
95-martin-orion: Done.

cirq/work/sampler.py
balopat: Here's my reasoning why this won't work - the key problem is that you lose out on relative phases for the amplitudes... 

```
import cirq
import numpy as np

a = cirq.NamedQubit("a")

alpha = 0.3 + 0.1j
beta = 1j * np.sqrt((1 - np.abs(alpha) ** 2))

# the true state vector 
psi = np.array([alpha, beta])
print(psi)
print(np.linalg.norm(psi))
print("Expectation value1: ", cirq.X(a).expectation_from_state_vector(psi, qubit_map={a: 0}))

# this is the resulting "state vector" from the "inverse born rule"
psi2 = np.array([np.abs(alpha), np.abs(beta)], dtype=np.complex64)

print(psi2)
print(np.linalg.norm(psi2))
print("Expectation value2: ", cirq.X(a).expectation_from_state_vector(psi2, qubit_map={a: 0}))

```

cirq/work/sampler.py
```diff
@@ -133,6 +134,78 @@ def sample(
...
+            key: [ops.PauliSum.wrap(o) for o in obs]
+            if isinstance(obs, List)
+            else [ops.PauliSum.wrap(obs)]
+            for key, obs in measurement_to_observables.items()
+        }
+        # Loop over sweeps
+        for _ in range(num_param_values):
+            state_vector_map = {}
+            for key in mto:
+                state_vector_map[key] = np.zeros(2 ** num_qubits, dtype=np.complex64)
```
mpharrigan, line 188: do you really want to pre-allocate one state vector per measurement??

cirq/work/sampler.py
```diff
@@ -133,6 +134,78 @@ def sample(
...
+    ) -> List[List[float]]:
+        """Calculates estimated expectation values from samples of a circuit.
+
+        This method can be run on any device or simulator that supports circuit
+        sampling. Compare with `simulate_expectation_values` in simulator.py,
+        which is limited to simulators but provides exact results.
+
+        Args:
+            program: The circuit to simulate.
+            measurement_to_observables: A map of measurement keys to the
```
mpharrigan, line 155: so `program` has measure instructions? I think this is the wrong way to do this API. We either have to have "assume measurements happen at the end" or finish the "generalized measurement operations" roadmap item so you can just add cirq.Measure(qubits, basis=X) to the circuit

mpharrigan (commented):
I think this is taking the wrong road towards sampling expectation values. Just like `simulator.sample` behind-the-scenes creates a wavefunction and draws samples from it, we should do the same thing here. I think the problem is trying to make this work with MeasurementGate.

95-martin-orion:
@balopat:
> Here's my reasoning why this won't work - the key problem is that you lose out on relative phases for the amplitudes...

@mpharrigan:
> I think this is taking the wrong road towards sampling expectation values. Just like `simulator.sample` behind-the-scenes creates a wavefunction and draws samples from it, we should do the same thing here. I think the problem is trying to make this work with MeasurementGate.

Confirmed with Matt offline that I'm taking the wrong approach here - essentially, in order to get expectation values for a basis other than Z we need to inject gates before the measurements. This precludes having a `measurements_to_observables` API, since anything other than a Z-observable will require material changes to the circuit.

I'll retreat to reconsider how to construct this, but most likely it'll involve gate injection and be restricted to terminal measurements.

PR #3701, opened 2021-01-25
Add hash values to tags in cirq.google

dstrain115:
- By mypy definitions, tags should be hashable.
- Add __hash__ fucntions so they actually are hashable.

### DIFF
--- a/cirq/google/ops/physical_z_tag.py
+++ b/cirq/google/ops/physical_z_tag.py
@@ -42,3 +42,6 @@ class PhysicalZTag:
 
     def __eq__(self, other) -> bool:
         return isinstance(other, PhysicalZTag)
+
+    def __hash__(self) -> int:
+        return 123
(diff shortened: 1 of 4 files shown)


mpharrigan (approved):
what do you mean "according to mypy"? Is there a test that can catch these issues automatically?

cirq/google/ops/physical_z_tag.py
```diff
@@ -42,3 +42,6 @@ def _json_dict_(self) -> Dict[str, Any]:
 
     def __eq__(self, other) -> bool:
         return isinstance(other, PhysicalZTag)
+
+    def __hash__(self) -> int:
+        return 123
```
vtomole, line 47: Why `123`?

dstrain115:
> what do you mean "according to mypy"? Is there a test that can catch these issues automatically?

We don't have a sub-class for Tags, we just call them Hashable in the with_tags.  I am not sure if there is a way to test this in general or not.
https://github.com/quantumlib/Cirq/blob/master/cirq/ops/raw_types.py#L428

I feel like we have used with_tags(cirq.google.PhysicalZ()) elsewhere and mypy didn't complain that it wasn't Hashable though.

cirq/google/ops/physical_z_tag.py
dstrain115: Just an arbitrary value so that all instances hash to the same value.

maffoo (approved):
LGTM

PR #3669, opened 2021-01-12
Cross entropy benchmarking spruce up

mpharrigan:
1. Add an explicit function for generating two-qubit fragments. These are important building blocks for doing "parallel XEB" e.g. for characterization/calibration. The new signature accepts any type of Qid and doesn't require a `pattern` argument.
2. I changed some of the internals of the random circuit generation to use the relevant Cirq type, like `Moment` instead of `Dict[qubit, op]`. 
3. Refactored the docstrings to associate specific elements of the circuit generation to the argument docs that control it. I personally think this is more organized than the wall-o-paragraph approach where every corner case of the function is documented up top
4. Add a notebook that expands upon the mathy docstring of `cirq.experiments.fidelity_estimation.least_squares_xeb_fidelity_from_expectations` with nice diagrams and plots. Let me know if this should be in `docs/` or `docs/xeb/` or `docs/characterization/`.

@XiaoMiQC, who I can't add as reviewer

PR #3706, opened 2021-01-26
Cross Entropy Benchmarking sample and simulate two qubits

mpharrigan:
Following #3669.

This wraps up the sampling and simulation functions teased in the previous notebook. It includes utilities for batching sampler requests and multiprocessing simulation. The notebook walks through how coherent error can cause fidelity to decay. In the next PR, we'll add an optimization loop to characterize the mystery angle.


cirq/experiments/fidelity_estimation.py
```diff
@@ -372,3 +394,225 @@ def least_squares_xeb_fidelity_from_probabilities(
     return least_squares_xeb_fidelity_from_expectations(
         measured_expectations, exact_expectations, uniform_expectations
     )
+
+
+@dataclass(frozen=True)
+class _Sample2qXEBTask:
+    """Helper contained for grouping a circuit to be sampled."""
```
mrwojtek, line 401: Helper container?

cirq/experiments/fidelity_estimation.py
```diff
@@ -372,3 +394,225 @@ def least_squares_xeb_fidelity_from_probabilities(
...
+
+    n_tasks = len(tasks)
+    b_tasks = [tasks[i : i + batch_size] for i in range(0, n_tasks, batch_size)]
+
+    run_batch = _SampleInBatches(sampler=sampler, repetitions=repetitions)
+    with ThreadPoolExecutor(max_workers=2) as pool:
+        futures = [pool.submit(run_batch, task) for task in b_tasks]
+
+        records = []
+        with tqdm.tqdm(total=n_tasks) as progress:
```
mrwojtek, line 495: Have you checked that this work in all circuimstances, when this is executed as a script or test?

I always assumed the pattern of selecting the progress framework to the end user of a library. This allows this functionality to be composed in a better way. What might be a good compromise to have a "report_progress" argument for this function.

cirq/experiments/fidelity_estimation.py
```diff
@@ -372,3 +394,225 @@ def least_squares_xeb_fidelity_from_probabilities(
...
+
+    Args:
+        circuits: A library of two-qubit circuits generated from
+            `random_rotations_between_two_qubit_circuit` of sufficient length
+            for `cycle_depths`.
+        cycle_depths: A sequence of cycle depths at which we will truncate each
+            of the `circuits` to simulate.
+        param_resolver: If circuits contain parameters, resolve according
+            to this ParamResolver prior to simulation
+        pool: If provided, execute the simulations in parallel.
```
mrwojtek, line 546: Is this formatting good? Looks narrow.

cirq/experiments/fidelity_estimation.py
```diff
@@ -372,3 +394,225 @@ def least_squares_xeb_fidelity_from_probabilities(
...
+
+    if pool is not None:
+        records = pool.map(_simulate_2q_xeb_circuit, tasks, chunksize=4)
+    else:
+        records = [_simulate_2q_xeb_circuit(record) for record in tasks]
+
+    return pd.DataFrame(records).set_index(['circuit_i', 'cycle_depth'])
+
+
+def simulate_2q_xeb_fidelities(
```
mrwojtek, line 572: Is simulate a correct verb here? benchmark_2q_xeb_fidelities?

cirq/experiments/fidelity_estimation_test.py
```diff
@@ -240,3 +248,94 @@ def test_least_squares_xeb_fidelity_from_probabilities():
...
+    ]
+    cycle_depths = np.arange(3, 50, 9)
+
+    sampled_df = sample_2q_xeb_circuits(
+        sampler=cirq.Simulator(seed=53), circuits=circuits, cycle_depths=cycle_depths
+    )
+    fid_df = simulate_2q_xeb_fidelities(sampled_df, circuits, cycle_depths)
+    print()
+    print(fid_df)
+    print()
```
mrwojtek, line 337: Is it purposeful?

cirq/experiments/fidelity_estimation_test.py
```diff
@@ -240,3 +248,94 @@ def test_least_squares_xeb_fidelity_from_probabilities():
     np.testing.assert_allclose(np.sum(np.array(r_log_math) ** 2), 0.0, atol=1e-2)
 
     np.random.set_state(prng_state)
+
+
+def test_sample_2q_xeb_circuits():
```
mrwojtek, line 253: Are circuits of depth 50 necessary to test this functionality? Is it expensive to run those tests? I'm asking because Cirq tests are not becoming any shorter these days...

cirq/experiments/random_quantum_circuit_generation.py
```diff
@@ -154,6 +154,69 @@ def __repr__(self) -> str:
 )
 
 
+def random_rotations_between_two_qubit_circuit(
+    q0: 'cirq.Qid',
+    q1: 'cirq.Qid',
+    depth: int,
+    two_qubit_op_factory: Callable[
+        ['cirq.Qid', 'cirq.Qid', 'np.random.RandomState'], 'cirq.OP_TREE'
+    ] = lambda a, b, _: google.SYC(a, b),
```
mrwojtek, line 163: Btw, If you want it to be more readable, you could extract the default arguments as constants and give them a nice names.

cirq/experiments/random_quantum_circuit_generation.py
```diff
@@ -154,6 +154,69 @@ def __repr__(self) -> str:
 )
 
 
+def random_rotations_between_two_qubit_circuit(
```
mrwojtek, line 157: Is all below changes from the previous PR?

mrwojtek (changes requested):

cirq/experiments/random_quantum_circuit_generation.py
mpharrigan: Now that the previous is merged (and I merged master into this) these diffs should be all gone!

cirq/experiments/fidelity_estimation.py
mpharrigan: thanks!

cirq/experiments/fidelity_estimation.py
```diff
@@ -372,3 +394,225 @@ def least_squares_xeb_fidelity_from_probabilities(
...
+            (circuit, cycle_depth) tasks to be run in each batch
+            is given by this number.
+
+    Returns:
+        A pandas dataframe with index given by ['circuit_i', 'cycle_depth'] and
+        column "sampled_probs".
+    """
+    import tqdm
+
+    all_all_qubits_s: Set['cirq.Qid'] = set()
```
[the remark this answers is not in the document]
mpharrigan: half of these variable names are because mypy doesn't let you re-declare a variable with a different type which is very annoying. I'll try to make them more readable

cirq/experiments/fidelity_estimation.py
```diff
@@ -372,3 +394,225 @@ def least_squares_xeb_fidelity_from_probabilities(
...
+        self.repetitions = repetitions
+
+    def __call__(self, tasks: List[_Sample2qXEBTask]):
+        prepared_circuits = [task.prepared_circuit for task in tasks]
+        results = self.sampler.run_batch(prepared_circuits, repetitions=self.repetitions)
+        assert len(results) == len(tasks)
+        records = []
+        for task, n_result in zip(tasks, results):
+            assert len(n_result) == 1, len(n_result)
+            result = n_result[0]  # sweep nesting
```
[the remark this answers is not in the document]
mpharrigan: done. also changed `n_result` to `nested_result`

cirq/experiments/fidelity_estimation.py
```diff
@@ -372,3 +394,225 @@ def least_squares_xeb_fidelity_from_probabilities(
...
+            ]
+        return records
+
+
+def sample_2q_xeb_circuits(
+    sampler: 'cirq.Sampler',
+    circuits: Sequence['cirq.Circuit'],
+    cycle_depths: Sequence[int],
+    repetitions: int = 10_000,
+    batch_size: int = 9,
```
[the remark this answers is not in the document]
mpharrigan: All the samplers support batching because of the fallback implementation of just running things in a for loop. I think the default should definitely be _with_ batching for these. You can always set it to `1`. And if there's a bug where it errors out, that's a failure of a sampler following the `Sampler` interface

cirq/experiments/fidelity_estimation.py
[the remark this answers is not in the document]
mpharrigan: Went through and gave full names. Also added a docstring to `_Sample2qXEBTask` to describe what is meant by `prepared_circuit`.

cirq/experiments/fidelity_estimation.py
mpharrigan: First, yes it works as a script, test, or notebook. 

But! Point taken. Yes, usually there should be some sort of optional progress callback. I think for this function in particular it would be nice to have it to default to having a nice progress bar. This is one of the issues with having a function that rides the line between "library" code and "application" code all in Cirq. Curious what @balopat or @maffoo would think about putting a progress bar here

cirq/experiments/fidelity_estimation.py
mpharrigan: simulate is not the correct verb. Benchmark is better. I guess I was going for the idea that the performance of this function is limited by simulating the circuits.

`simulate_in_order_to_benchmark_2q_xeb_circuits` is too long of a name :) Changing to `benchmark_2q_xeb_fidelities`.

cirq/experiments/fidelity_estimation_test.py
mpharrigan: It "only" takes 948 ms. I'll pare it down. `cycle_depths = np.arange(3, 20, 6)` is 306 ms

cirq/experiments/fidelity_estimation_test.py
mpharrigan: no... how do I always miss these

cirq/experiments/fidelity_estimation.py
mpharrigan: I thought we stuck to 80 character width for docstrings still. Or are new docstrings supposed to be wide? @balopat

cirq/experiments/fidelity_estimation.py
mpharrigan: pushed a commit with more-optional tqdm, but tqdm by default

mpharrigan:
PTAL!

cirq/experiments/fidelity_estimation.py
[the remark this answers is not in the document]
mrwojtek: Alright.

cirq/experiments/fidelity_estimation.py
```diff
@@ -372,3 +395,254 @@ def least_squares_xeb_fidelity_from_probabilities(
...
+        pass
+
+
+def sample_2q_xeb_circuits(
+    sampler: 'cirq.Sampler',
+    circuits: Sequence['cirq.Circuit'],
+    cycle_depths: Sequence[int],
+    repetitions: int = 10_000,
+    batch_size: int = 9,
+    progress_bar=tqdm.tqdm,
```
mrwojtek, line 476: Missing type information

cirq/experiments/fidelity_estimation.py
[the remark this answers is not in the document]
mrwojtek: Much better

cirq/experiments/fidelity_estimation.py
mpharrigan: I'm loath to figure out what the annotation is for "thing that satisfies the context manager protocol"

mpharrigan: I can change it to `Optional[Type]` but that doesn't really clarify it

maffoo: `typing.ContextManager[Foo]`, where `Foo` the type returned by `__enter__` (or yielded if using `contextlib.contextmanager` decorator).

mpharrigan: `Optional[Callable[..., ContextManager]]` works, but again I'm not sure if that helps anything

mpharrigan: I don't know why `Optional[Callable[[int], ContextManager]]` doesn't work:

```
cirq/experiments/fidelity_estimation.py:526: error: Unexpected keyword argument "total"
```

which is the line where I'm calling the callable with an `int` argument .....

mpharrigan:
anything else on this?

mrwojtek (approved):

cirq/experiments/fidelity_estimation.py
balopat: docstrings can be up to 100 too

docs/characterization/xeb_coherent_noise.ipynb
```diff
@@ -0,0 +1,283 @@
+{
```
balopat, line 1: <p>we're missing standard notebook headers</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/quantumlib/Cirq/pull/3706/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

balopat, line 1: <p>This notebook will eventually end up on the site. Please annotate the types with packages, e.g <code>cirq.PhasedFSimGate</code>  - this will create automated links to the names on the reference docs. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/quantumlib/Cirq/pull/3706/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

cirq/experiments/fidelity_estimation.py
mpharrigan: re-wrapped

docs/characterization/xeb_coherent_noise.ipynb
mpharrigan: added

mpharrigan: cool, done

mpharrigan:
@balopat back at you

requirements.txt
```diff
@@ -14,3 +14,4 @@ sortedcontainers~=2.0
 scipy
 sympy
 typing_extensions
+tqdm
```
balopat, line 17: New dependency alert! 

In lieu of keeping cirq light, I'm hesitant to add tqdm as a dependency.

1. We should check if this is going to break g3
2. Can we make this somehow optional so that we reduce the dependency requirements for Cirq? (e.g. `cirq[ui]` or `cirq[visualization]` or via making the default value progress bar NoProgress) 
3. do we need to pin this? 
4. can we abstract it out better in our code instead of directly depending on it? That would help with testing progress logic as well as well as would

balopat (commented):
I would still have to look at the code - though I trust @mrwojtek's review. 
I have questions around adding tqdm.

requirements.txt
mpharrigan: If we're going to package our experiments/algorithms code in Cirq, we need to figure this out. For now, I can change it to default to no-progress but I think this hurts usability of this application code

balopat: Fair point about if we package this together we'll have to deal with this - and that is the plan. 
I'll make a mental note of this. I checked 1.) and we should be fine (though I haven't run tests on that version).

balopat (approved):

mpharrigan:
Very strange. All the commits look like they're authored by googlers (?)

balopat:
I've seen this error before when cirqbot gets confused...we should have a look for cases when lack of "cla/google" is blamed for no reason and fix it in CirqBot.

PR #3707, opened 2021-01-26
Add ability to construct service via environment variables

dabacon:
This adds the ability to construction the ionq.Service from environment variables.  By setting `IONQ_REMOTE_HOST` and `IONQ_API_KEY` one can avoid storing the API key in code (and potentially checking it into a public version control system).


gmauricio (approved):

dabacon:
Part of #3479

PR #3709, opened 2021-01-26
Join DensityMatrixSimulator and SparseSimulator in common base class

daxfohl:
Creates a common base class `OpByOpSimulator` from which the two above simulators extend. The `_run` and `_base_iterator` functions have been combined in the base class to suit both cases. The only differences between the two classes now are in their state types and their StepResult/FinalResult types. The only function remaining in each of the two classes is `__init__`, which injects the corresponding state type manager and their StepResult type factory into the base class. This should make it much easier to create new simulators in the future.

Additionally this PR improves the generic type declarations, and I believe it allows SparseSimulator to handle noise models (not sure how to check).

This same base class should work well for MPS and Clifford simulators. It may be a good idea to confirm this before merging.


cirq/sim/density_matrix_simulator.py
```diff
@@ -13,202 +13,121 @@
...
+    def log_of_measurement_results(self):
+        return self._log_of_measurement_results
+
+    def record_measurement_result(self, key: str, value: Any):
+        if key not in self._log_of_measurement_results:
+            self._log_of_measurement_results[key] = []
+        self._log_of_measurement_results[key].extend(value)
 
-class DensityMatrixSimulator(simulator.SimulatesSamples, simulator.SimulatesIntermediateState):
-    """A simulator for density matrices and noisy quantum circuits.
```
mpharrigan, line 39: where did this lovely docstring go

cirq/sim/abstract_state.py
```diff
@@ -0,0 +1,32 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import abc
+from typing import Any
+
+
+class AbstractState(metaclass=abc.ABCMeta):
```
95-martin-orion, line 19: Docstring for class?

95-martin-orion, line 19: The [nomenclature docs](https://github.com/quantumlib/Cirq/blob/master/docs/dev/nomenclature.md) caution against using `state` as a type name, since it's a heavily overloaded term. That being said, whether this is a density matrix or state vector depends on the implementation; `state_rep` seems like the closest analog in the doc. Maybe `AbstractStateRepresentation` for this?

cirq/sim/density_matrix_simulator.py
```diff
@@ -13,202 +13,121 @@
...
+from typing import Any, Dict, List, TYPE_CHECKING, Tuple, Type, Union
 
 import numpy as np
 
-from cirq import circuits, ops, protocols, qis, study, value, devices
+from cirq import ops, protocols, qis, study, value
 from cirq.sim import density_matrix_utils, simulator
-from cirq.sim.simulator import check_all_resolved
+from cirq.sim.abstract_state import AbstractState
+from cirq.sim.op_by_op_simulator import StateFactory, SimulationResultFactory, OpByOpSimulator
```
95-martin-orion, line 23: Prefer importing the full module, rather than individual objects / methods: [style guide](https://github.com/quantumlib/Cirq/blob/master/docs/dev/style.md).

Similarly for other files in this PR.

cirq/sim/density_matrix_simulator_test.py
```diff
@@ -327,7 +327,7 @@ def test_run_measurement_not_terminal_no_repetitions(dtype):
                     result.measurements, {'0': np.empty([0, 1]), '1': np.empty([0, 1])}
                 )
                 assert result.repetitions == 0
-        assert mock_sim.call_count == 0
+        assert mock_sim.call_count == 4
```
95-martin-orion, line 330: Why did this increase? For `repetitions=0`, I wouldn't expect `mock_sim` to be called.

Similarly for the other changes in this file.

cirq/sim/op_by_op_simulator.py
```diff
@@ -0,0 +1,289 @@
...
+)
+
+if TYPE_CHECKING:
+    import cirq
+
+
+TState = TypeVar('TState', bound=AbstractState)
+
+
+class StateFactory(Generic[TState], metaclass=abc.ABCMeta):
```
95-martin-orion, line 51: Please add class and method docstrings in this file to clarify how to implement these classes.

95-martin-orion (changes requested):
In addition to the comments below, #3698 added a new method to the simulator interface which may complicate this PR.

balopat (changes requested):
Hi @daxfohl, 

In general, before refactoring design suggestions, it would be really helpful to provide  provide motivating example scenarios that clearly show the benefit of the new design. 

In this case there is a worry that - while there is some repetition between the two classes - this refactoring would end up coupling two separate code paths (state vector vs density matrix based simulations) that now won't be able to diverge easily. DRY is not necessarily the optimal solution here. 

There are several ideas in this PR (typing related as well as structural) - maybe you could split this up to smaller ideas?

daxfohl:
@balopat Thanks for the feedback. Both are good points. There are some smaller, less controversial chunks I can probably break out of this first. Per the merging of the classes, I did have the same concern going into it, but ended up really liking the result, and hope I can convince others the same: when creating a new simulator now, all you have to do is create a class that manages your state transitions, and inherit from OpByOpSimulator, and done. No `_run`, no `_base_iterator`. Plus you get the optimizations like split into unitary and non-unitary, and functionality like noise application for free. I'll write up a doc with some more examples.

daxfohl:
@balopat nvm I no longer think this is a good idea :)

Once the `act_on` is merged into density matrix simulator, there will be less shared code between the different simulators. I may revisit later as we start thinking about flow control and how to enable it in all simulators without a bunch of repetition, but for now, closing this PR.

PR #3710, opened 2021-01-27
Fix bug of in IonQ result conversion

dabacon:
`cirq.Result` is supposed to be supplied measurement results as a dict from string to a multidimensional numpy array.  However it is possible to instead supply the value as a list of lists.  In this case `cirq.Result` equality will still work, but it will fail to when converted to a string (since `__str__` uses numpy slicing).  

I'm not quite sure why mypy didn't yell about the incorrect type here.  I guess the value in a dict is covariant? I thought it was invariant.


balopat:
> `cirq.Result` is supposed to be supplied measurement results as a dict from string to a multidimensional numpy array. However it is possible to instead supply the value as a list of lists. In this case `cirq.Result` equality will still work, but it will fail to when converted to a string (since `__str__` uses numpy slicing).
> 
> I'm not quite sure why mypy didn't yell about the incorrect type here. I guess the value in a dict is covariant? I thought it was invariant.

mypy doesn't recognize numpy types.
All of this is accepted by default
```
Result(params=ParamResolver({}), measurements={"a":[1,2,3]}) # should fail but it won't 
Result(params=ParamResolver({}), measurements={"a":"b"})  # should fail but it won't 
Result(params=ParamResolver({}), measurements={"a":[[1,0],[0,1]]})  # should fail but it won't 

Result(params=ParamResolver({}), measurements={"a":np.array([1,2,3])})  # OK 
Result(params=ParamResolver({}), measurements={"a":np.array([[1,0],[0,1]])}) # OK 
```

I tried with `pip install data-science-types` - that project creates type stubs for numpy, matplotlib, pandas.
With this, mypy catches the first three:  

```
cirq/bla.py:5: error: Dict entry 0 has incompatible type "str": "List[int]"; expected "str": "ndarray[Any]"
cirq/bla.py:6: error: Dict entry 0 has incompatible type "str": "str"; expected "str": "ndarray[Any]"
cirq/bla.py:7: error: Dict entry 0 has incompatible type "str": "List[List[int]]"; expected "str": "ndarray[Any]"
``` 

However, it generates a lot of other noise: 

```
Found 1337 errors in 154 files (checked 763 source files)
```

That includes errors due to missing type stubs in the package e.g.: 

```
cirq/contrib/acquaintance/gates_test.py:20: error: Module 'numpy.random' has no attribute 'poisson'
cirq/contrib/acquaintance/executor_test.py:93: error: Module has no attribute "allclose"
cirq/contrib/acquaintance/executor_test.py:94: error: Module has no attribute "absolute"
cirq/contrib/acquaintance/executor_test.py:116: error: Module has no attribute "random"; maybe "rand"?
cirq/contrib/acquaintance/executor_test.py:177: error: Unexpected keyword argument "verbose" for "assert_allclose"
```

So I think we'll just have to live with this not being in mypy. Maybe we could / should do runtime type validation in the constructor though for the measurement argument!

ColemanCollins (approved):
Can't speak to the mypy conversation but this looks good to me!

dabacon:
Part of #3479

PR #3711, opened 2021-01-27
Add utilities that allow to characterize circuits using Floquet calibration

mrwojtek:
This adds two major methods:
 - make_floquet_request_for_circuit that generates a set of characterization requests that later can be used to trigger characterization, and
 - run_floquet_characterization_for_circuit that uses make_floquet_request_for_circuit and executes the characterization against cirq.google.Engine.

The logic is simple and supports ISWAP ** -0.5 gates only at this moment.

There is a notion of "moments mapping" introduced which is used to connect characterization requests/results to the moments in the circuit. Alternatively this could be automated but it's not a trivial problem from the algorithmic point of view.


cirq/google/calibration/workflow.py
```diff
@@ -34,16 +269,99 @@ def run_characterizations(
...
+        merge_subsets: Whether to merge moments that can be characterized at the same time
+            together.
+        max_layers_per_request: Maximum number of calibration requests issued to cirq.Engine at a
+            single time. Defaults to 1.
+        progress_func: Optional callback function that might be used to report the calibration
+            progress. The callback is called with two integers, the first one being a number of
+            layers already calibrated and the second one the total number of layers to calibrate.
+
+    Returns:
+        Tuple of:
```
mrwojtek, line 346: See also https://github.com/mrwojtek/Cirq/pull/3#discussion_r564104095

cirq/google/calibration/phased_fsim.py
```diff
@@ -33,6 +46,14 @@
     from cirq.protocols import json_serializable_dataclass
 
 
+T = TypeVar('T')
+
+
+# Workaround for: https://github.com/python/mypy/issues/5858
+def lru_cache_typesafe(func: Callable[..., T]) -> T:
```
mpharrigan, line 53: Do we have an issue where we can record all the places where mypy makes things worse? @balopat

mpharrigan (approved):
no blockers from me, since I reviewed the "fork" version.

cirq/google/calibration/workflow.py
mrwojtek: Updated based on that conversation

mrwojtek:
PTAL. Now make_floquet_request_for_circuit and run_floquet_characterization_for_circuit return dataclasses instead of tuples.

mrwojtek:
I'm slightly blocked on this now. I'll submit it - if there are any further comments I'll create a follow-up PRs to fix them.

Issue #3603, opened 2020-12-11
Notebook test slow tests


dabacon:
Probably worth trying to speed up that notebook test as it is long poll in the CI.  Maybe there is a pattern to add something that can "fix" the notebooks to smaller parameters to speed them up?
```
63.27s call     dev_tools/notebook_test.py::test_notebooks[/Users/bacon/code/github/Cirq/cirq/contrib/quimb/Cirq-to-Tensor-Networks.ipynb]
42.70s call     dev_tools/notebook_test.py::test_notebooks[/Users/bacon/code/github/Cirq/docs/tutorials/quantum_walks.ipynb]
33.26s call     dev_tools/notebook_test.py::test_notebooks[/Users/bacon/code/github/Cirq/docs/tutorials/educators/qaoa_ising.ipynb]
28.50s call     dev_tools/notebook_test.py::test_notebooks[/Users/bacon/code/github/Cirq/docs/tutorials/hidden_linear_function.ipynb]
27.31s call     dev_tools/notebook_test.py::test_notebooks[/Users/bacon/code/github/Cirq/docs/tutorials/educators/chemistry.ipynb]
24.82s call     dev_tools/notebook_test.py::test_notebooks[/Users/bacon/code/github/Cirq/docs/tutorials/qaoa.ipynb]
21.05s call     dev_tools/notebook_test.py::test_notebooks[/Users/bacon/code/github/Cirq/docs/tutorials/variational_algorithm.ipynb]
14.91s call     dev_tools/notebook_test.py::test_notebooks[/Users/bacon/code/github/Cirq/cirq/contrib/quimb/Contract-a-Grid-Circuit.ipynb]
11.20s call     dev_tools/notebook_test.py::test_notebooks[/Users/bacon/code/github/Cirq/docs/tutorials/rabi_oscillations.ipynb]
10.77s call     dev_tools/notebook_test.py::test_notebooks[/Users/bacon/code/github/Cirq/docs/interop.ipynb]
9.48s call     dev_tools/notebook_test.py::test_notebooks[/Users/bacon/code/github/Cirq/docs/tutorials/educators/intro.ipynb]
9.44s call     dev_tools/notebook_test.py::test_notebooks[/Users/bacon/code/github/Cirq/docs/tutorials/educators/textbook_algorithms.ipynb]
7.76s call     dev_tools/notebook_test.py::test_notebooks[/Users/bacon/code/github/Cirq/docs/circuits.ipynb]
[...]

PR #3713, opened 2021-01-27
only test notebooks that are checked into git

balopat:
This helps with local running of the notebook tests - it will only pick up notebooks that are checked into git.
Related to https://github.com/quantumlib/Cirq/issues/3603.


mpharrigan:
Awesome! but why are we using a custom `shell_tools` instead of `subprocess.call`? It seems to be causing problems on different platforms

balopat:
> Awesome! but why are we using a custom `shell_tools` instead of `subprocess.call`? It seems to be causing problems on different platforms

I've been using shell_tools everywhere as it was convenient - even in this test - it was never executed on non-Linux platforms - however the test collection phase is executed, so I will use subprocess then.

mpharrigan (approved):

PR #3715, opened 2021-01-28
Fix a bug in Depol channel with multi qubits

### LINKED ISSUE #3685: DepolarizingChannel with more than one qubit is buggy
**Description of the issue**

The `cirq.DepolarizingChannel` class is buggy for `n_qubits` larger than 1.

```python
import cirq
cirq.DepolarizingChannel(p=0.1, n_qubits=2)(*cirq.LineQubit.range(2))
```

Gives:
```
ValueError: Wrong number of qubits for <cirq.depolarize(p=0.1,n_qubits=2)>. Expected 1 qubits but got <[cirq.LineQubit(0), cirq.LineQubit(1)]>.
```

Full log:
<details>
ValueError                                Traceback (most recent call last)
<ipython-input-6-5bc7460d03b5> in <module>
      1 import cirq
      2 
----> 3 cirq.DepolarizingChannel(p=0.1, n_qubits=2)(*cirq.LineQubit.range(2))

~/anaconda3/lib/python3.7/site-packages/cirq/ops/raw_types.py in __call__(self, *args, **kwargs)
    268 
    269     def __call__(self, *args, **kwargs):
--> 270         return self.on(*args, **kwargs)
    271 
    272     def with_probability(self, probability: 'cirq.TParamVal') -> 'cirq.Gate':

~/anaconda3/lib/python3.7/site-packages/cirq/ops/raw_types.py in on(self, *qubits)
    210         from cirq.ops import gate_operation
    211 
--> 212         return gate_operation.GateOperation(self, list(qubits))
    213 
    214     def wrap_in_linear_combination(

~/anaconda3/lib/python3.7/site-packages/cirq/ops/gate_operation.py in __init__(self, gate, qubits)
     58             qubits: The qubits to operate on.
     59         """
---> 60         gate.validate_args(qubits)
     61         self._gate = gate
     62         self._qubits = tuple(qubits)

~/anaconda3/lib/python3.7/site-packages/cirq/ops/raw_types.py in validate_args(self, qubits)
    199             ValueError: The gate can't be applied to the qubits.
    200         """
--> 201         _validate_qid_shape(self, qubits)
    202 
    203     def on(self, *qubits: Qid) -> 'Operation':

~/anaconda3/lib/python3.7/site-packages/cirq/ops/raw_types.py in _validate_qid_shape(val, qubits)
    728         raise ValueError(
    729             'Wrong number of qubits for <{!r}>. '
--> 730             'Expected {} qubits but got <{!r}>.'.format(val, len(qid_shape), qubits)
    731         )
    732     if any(qid.dimension != dimension for qid, dimension in zip(qubits, qid_shape)):

ValueError: Wrong number of qubits for <cirq.depolarize(p=0.1,n_qubits=2)>. Expected 1 qubits but got <[cirq.LineQubit(0), cirq.LineQubit(1)]>.

</details>

**Cirq version**
`'0.10.0.dev'`

**Temporary workaround**
For any users interested in multi-qubit depolarizing noise, this is a possible workaround to avoid the problem:

```python
import itertools
from cirq import AsymmetricDepolarizingChannel

def my_depolarizing_channel(p: float, n_qubits: int):
    """Builds a depolarizing channel from cirq.AsymmetricDepolarizingChannel."""

    error_probabilities = {}
    p_depol = p / (4**n_qubits - 1)
    p_identity = 1.0 - p
    for pauli_tuple in itertools.product(['I', 'X', 'Y', 'Z'], repeat=n_qubits):
        pauli_string = ''.join(pauli_tuple)
        if pauli_string == 'I' * n_qubits:
            error_probabilities[pauli_string] = p_identity
        else:
            error_probabilities[pauli_string] = p_depol
    return AsymmetricDepolarizingChannel(error_probabilities=error_probabilities)
```

tonybruguier:
Fixes #3685


cirq/ops/common_channels.py
```diff
@@ -231,7 +231,7 @@ def asymmetric_depolarize(
 
 @value.value_equality
 class DepolarizingChannel(gate_features.SingleQubitGate):
```
balopat, line 233: Why is this still a SingleQubitGate?

balopat (approved):
LGTM - we can merge after the removal of the SingleQubitGate parent.

cirq/ops/common_channels.py
tonybruguier: Changed to raw_types.Gate

tonybruguier:
Thanks. I changed to a generic gate, but it made a test fail because it was relying on `on_each`. I considered three ways to fix:
1) Add a specialized `on_each` function to `DepolarizingChannel` but that seemed strange to have it only for this channel and not the others
2) Add the function to the base class (taking a sequence of tuples) but that seemed like quite a big undertaking and there could be a reason I don't understand why it wasn't done in the first place.
3) Fix the test directly, thus leaving the code in a way that could break outside code.

I'm happy to reconsider the choice (or consider other options) of course. Thanks for your guidance!

balopat:
> Thanks. I changed to a generic gate, but it made a test fail because it was relying on `on_each`. I considered three ways to fix:
> 
> 1. Add a specialized `on_each` function to `DepolarizingChannel` but that seemed strange to have it only for this channel and not the others
> 2. Add the function to the base class (taking a sequence of tuples) but that seemed like quite a big undertaking and there could be a reason I don't understand why it wasn't done in the first place.
> 3. Fix the test directly, thus leaving the code in a way that could break outside code.
> 
> I'm happy to reconsider the choice (or consider other options) of course. Thanks for your guidance!

Oh, we're lucky the tests caught this! Good thinking around the options!
Yes, the right thing to do is to switch the parent class to `Gate`, instead of the `SingleQubitGate` parent class. 
That class only introduces the `on_each` method, nothing else. However, we do want to be nice about it and not break people that depended on it, so let's introduce a deprecated `on_each` method with a warning to `DepolarizingChannel`. v0.11 is a good target to deprecate this.

mpharrigan:
You should have an `on_each` method that checks if the gate is indeed a single-qubit gate, at which point it can happily do an "on each". Otherwise, it is an error

tonybruguier:
Like this? I wasn't sure whether to mark it as deprecated.

Also the test failures are unrelated, right?

mpharrigan:
I think @balopat wants it to be deprecated. I think (but can be convinced otherwise) that it should exist, since a common case is to use DepolarizingChannel as a single-qubit gate. In fact --- that used to be the only way!

We could factor the function out into a mixin called like `SometimesOneQubitGate` if we wanted to get fancy (and then add it to things like `MeasurementGate` and `IdentityGate`)

tonybruguier:
Thanks for the advice! I created a new class `SometimesOneQubitGate` but not as a mixin, but rather as an in-between class (it seemed to make more sense, but I don't feel super strongly about this). I also modified the `IdentityGate` because might as well not do an incomplete refactoring.

There was a MyPy isssue in `pauli_string_test.py` with `test_pass_operations_over_double` that I solved with a cast. It seems that it's just an exisiting failure that was triggered by the change. It seems a similar issue had been solved already the very same way in `test_pass_operations_over_single` just above.

Finally, there are still some lint failures, but I was able to reproduce them at head, and so I sent a separate PR #3718 to fix these.

PTAL.

mpharrigan:
I don't know the technical definition of "mixin" but this is exactly what I had in mind. What do you think @balopat ? Too crazy? I think it's kinda nice

cirq/ops/gate_features.py
```diff
@@ -49,6 +46,8 @@ def on_each(self, *targets: Union[raw_types.Qid, Iterable[Any]]) -> List[raw_typ
         Raises:
             ValueError if targets are not instances of Qid or List[Qid].
```
balopat, line 47: We should change the raises docstring to cater for the "times" when it's not a single qubit gate

balopat:
> I don't know the technical definition of "mixin" but this is exactly what I had in mind. What do you think @balopat ? Too crazy? I think it's kinda nice

I think it fell in nicely in the end as the n > 1 case is the generalization of the n == 1 case, and hence it is the parent class. However that is true for the `Gate` class as well, it _is_ the ultimate n-qubit gate. Because of this I don't think I like the `SometimesOne` name - maybe `SupportsOnEach` would be a bit more what it _does_ - not what it _is_? Also added a nit around the docstrings. Otherwise I like the inheritance structure, and we don't need to deprecate the function this way. Good stuff!

cirq/ops/gate_features.py
tonybruguier: Done.

tonybruguier:
Thanks. PTAL.

mpharrigan:
Good, yeah I think `SupportsOnEach` is better

cirq/ops/gate_features.py
```diff
@@ -31,11 +31,8 @@ def qubit_index_to_equivalence_group_key(self, index: int) -> int:
         return 0
 
 
-class SingleQubitGate(raw_types.Gate, metaclass=abc.ABCMeta):
-    """A gate that must be applied to exactly one qubit."""
-
-    def _num_qubits_(self) -> int:
-        return 1
+class SometimesOneQubitGate(raw_types.Gate, metaclass=abc.ABCMeta):
```
balopat, line 34: can you rename this to `SupportsOnEach`?  And then we can merge it!

tonybruguier: Sorry, I had forgotten to git push. Like this?

cirq/ops/identity.py
```diff
@@ -64,35 +64,6 @@ def _qid_shape_(self) -> Tuple[int, ...]:
...
-        Returns:
-            Operations applying this gate to the target qubits.
-
-        Raises:
-            ValueError if targets are not instances of Qid or List[Qid] or
-            the gate from which this is applied is not a single qubit identity
-            gate.
-        """
-        if len(self._qid_shape) != 1:
-            raise ValueError('IdentityGate only supports on_each when it is a one qubit gate.')
```
mpharrigan, line 83: heyo!

balopat: Elaborate! Is that a positive exclamation that "heyo, nice find, look, this works here as well and the code is simpler" or "heyo don't do this"? :)

balopat (approved):

cirq/ops/identity.py
mpharrigan: > heyo, nice find, look, this works here as well and the code is simpler

This one!

tonybruguier: Ack. Thanks!

balopat (approved):

PR #3719, opened 2021-01-29
Backwards-compatibility behavior for resolve_parameters.

### LINKED ISSUE #3714: _resolve_parameters_ changes are not backwards-compatible
**Description of the issue**

External code that defined a `_resolve_parameters_` magic method for Cirq 0.9 is broken by #3546, which changed the function signature for that method in a non-backwards-compatible way.

Construct a parameterized circuit in Cirq v0.10.0-dev, then attempt to run it in the latest version of qsim.

**Cirq version**

Occurs when upgrading from v0.9.* to v0.10.0-dev

95-martin-orion:
Fixes #3714.


cirq/protocols/resolve_parameters.py
```diff
@@ -154,7 +156,17 @@ def resolve_parameters(
         return type(val)(resolve_parameters(e, param_resolver, recursive) for e in val)
 
     getter = getattr(val, '_resolve_parameters_', None)
-    result = NotImplemented if getter is None else getter(param_resolver, recursive)
+    try:
+        result = NotImplemented if getter is None else getter(param_resolver, recursive)
+    except TypeError:
```
balopat, line 161: I would be very explicit about the exact error message here. Otherwise other type of implementation errors resulting in `TypeError` will be handled by this logic, which will be confusing as the real `TypeError` will be then swallowed. 

In particular I would narrow the handling of this branch in case we get this specific error string: `_resolve_parameters_() takes 2 positional arguments but 3 were given`

cirq/protocols/resolve_parameters_test.py
```diff
@@ -143,3 +143,22 @@ def test_recursive_resolve():
     assert cirq.resolve_parameters_once(a, resolver) == b
     with pytest.raises(RecursionError):
         _ = cirq.resolve_parameters(a, resolver)
+
+
+# TODO: remove in Cirq v0.11
+def test_backwards_compatible():
```
balopat, line 149: In accordance with my other comment I'd add another test that would cover the case when the `TypeError` is not due to the specific incompatibility that we're trying to fix here.

balopat (changes requested):
Looks good, I'd be more specific about the error type though!

cirq/protocols/resolve_parameters.py
dstrain115, line 161: Is this the only way we can get this exception?  I am not sure this is the best way to catch the function not having recursive.  How about something like this instead?

```
if 'recursive' not in getter.__code__.co_varnames:
   if not recursive:
            raise ValueError(
                f'Object type {type(val)} does not support non-recursive parameter resolution.'
                ' This must be updated before Cirq v0.11.'
            )
     result = NotImplemented if getter is None else getter(param_resolver)
```

I am not sure how version compatible this is, but I think something like this would work.

95-martin-orion: I'll try this first, seeing as it sidesteps the more fragile `try/except` logic.

95-martin-orion: Opted to avoid `try/except` altogether - see Doug's comment.

cirq/protocols/resolve_parameters_test.py
95-martin-orion: Opted to avoid `try/except` altogether - see Doug's comment.

95-martin-orion:
A qsim bug prevents this from working at `{cirq==0.10.0.dev, qsimcirq==0.7.1}`, but it has been fixed in the unreleased version of qsim. Known working versions with this PR:
```
{cirq==0.9.1, qsimcirq==0.7.1}
{cirq==0.9.1, qsimcirq==0.7.2}  # qsimcirq 0.7.2 is unreleased
{cirq==0.10.0.dev, qsimcirq==0.7.2}
```
The qsim 0.7.2 release should roll out later this week.

balopat (approved):
LGTM

PR #3720, opened 2021-01-29
Add from_moment to PhasedFSimCalibrationRequest

dstrain115:
- Allows the user to specify a calibration request from a Moment instead a
bunch of tuples of qubits.


cirq/google/calibration/phased_fsim.py
```diff
@@ -220,6 +220,22 @@ class PhasedFSimCalibrationRequest(abc.ABC):
     pairs: Tuple[Tuple[Qid, Qid], ...]
     gate: Gate  # Any gate which can be described by cirq.PhasedFSim
 
+    @staticmethod
+    def _create_pairs_from_moment(moment: Moment):
+        gate = None
```
mpharrigan, line 225: short (one line) docstring?

cirq/google/calibration/phased_fsim.py
```diff
@@ -261,6 +277,11 @@ class FloquetPhasedFSimCalibrationRequest(PhasedFSimCalibrationRequest):
 
     options: FloquetPhasedFSimCalibrationOptions
 
+    @classmethod
+    def from_moment(cls, moment: Moment, options: FloquetPhasedFSimCalibrationOptions):
```
mpharrigan, line 281: docstring, especially since there are restrictions on what `moment` can be, namely all the same two-qubit gate

cirq/google/calibration/phased_fsim.py
```diff
@@ -220,6 +220,22 @@ class PhasedFSimCalibrationRequest(abc.ABC):
     pairs: Tuple[Tuple[Qid, Qid], ...]
     gate: Gate  # Any gate which can be described by cirq.PhasedFSim
 
+    @staticmethod
```
mpharrigan, line 223: why not have this as a private free function?

cirq/google/calibration/phased_fsim.py
mrwojtek: +1, especially for explanation on what moments are allowed

cirq/google/calibration/phased_fsim.py
```diff
@@ -220,6 +220,22 @@ class PhasedFSimCalibrationRequest(abc.ABC):
...
+    def _create_pairs_from_moment(moment: Moment):
+        gate = None
+        pairs = []
+        for op in moment:
+            if gate is None:
+                gate = op.gate
+            elif gate != op.gate:
```
mrwojtek, line 230: Shouldn't that be gated against a check that operation is a GateOperation? Not all operations have gate?

cirq/google/calibration/phased_fsim.py
```diff
@@ -220,6 +220,22 @@ class PhasedFSimCalibrationRequest(abc.ABC):
     pairs: Tuple[Tuple[Qid, Qid], ...]
     gate: Gate  # Any gate which can be described by cirq.PhasedFSim
 
+    @staticmethod
+    def _create_pairs_from_moment(moment: Moment):
+        gate = None
+        pairs = []
+        for op in moment:
+            if gate is None:
+                gate = op.gate
```
mrwojtek, line 229: op.gate is optional, isn't it possible that the first operation will not have a gate and will sneak through this code?

cirq/google/calibration/phased_fsim.py
balopat: +1

cirq/google/calibration/phased_fsim.py
dstrain115: Done.

cirq/google/calibration/phased_fsim.py
dstrain115: Done.

cirq/google/calibration/phased_fsim.py
dstrain115: Done.

cirq/google/calibration/phased_fsim.py
dstrain115: Changed to check if op.gate is not None, since it could be a TaggedOperation or some other class instead.

cirq/google/calibration/phased_fsim.py
dstrain115: Done.

dstrain115:
Thanks for the reviews!  PTAL!

mrwojtek:
Thanks!

mrwojtek (approved):

mpharrigan (approved):

PR #3724, opened 2021-02-01
Fix: tutorial index headings do not match tutorial content page

daxfohl:
Index does not match content on tutorials page. This fixes that. I haven't worked the examples myself, so whether this is the correct fix or whether the content should be changed to match the index, is beyond my ability to judge.


balopat (approved):
Thanks for the fix!!!

PR #3725, opened 2021-02-01
Add PhasedFSimEngineSimulator which allows to simulate calibration requests and unitary shifts of the PhasedFSimGate

mrwojtek:
One limitation is that it doesn't support multi-moment circuits (circuits where a single PhasedFSimGate appears in multiple moment configurations). It's still extremely useful without that support.


cirq/google/calibration/engine_simulator.py
```diff
@@ -0,0 +1,405 @@
...
+                zeta=sample_value(mean.zeta, sigma.zeta),
+                chi=sample_value(mean.chi, sigma.chi),
+                gamma=sample_value(mean.gamma, sigma.gamma),
+                phi=sample_value(mean.phi, sigma.phi),
+            )
+
+        if mean.any_none():
+            raise ValueError(f'All mean values must be provided, got mean of {mean}')
+
+        rand = parse_random_state(random_or_seed)
```
mpharrigan, line 169: I get weirded out when variables used in closures are defined after the closure, although you've taught me it works fine. I still think it would read better if the `parse_random_state` and mean check goes above the definition of `sample_gate`

cirq/google/calibration/engine_simulator.py
```diff
@@ -0,0 +1,405 @@
...
+        Returns:
+            New PhasedFSimEngineSimulator instance.
+        """
+
+        def sample_gate(_1: Qid, _2: Qid, gate: FSimGate) -> PhasedFSimCharacterization:
+            assert np.isclose(gate.theta, np.pi / 4) and np.isclose(
+                gate.phi, 0.0
+            ), f'Expected ISWAP ** -0.5 like gate, got {gate}'
+
+            def sample_value(
```
mpharrigan, line 150: this doesn't need to be double-closured. It might read a little better if it's defined outside `sample_gate`? I don't know, up to you

cirq/google/calibration/engine_simulator.py
```diff
@@ -0,0 +1,405 @@
...
+    internal state of a simulator. Circuits which are run on this simulator are modified to account
+    for the changes in the unitary parameters as described by the calibration results.
+    """
+
+    def __init__(
+        self,
+        simulator: Simulator,
+        *,
+        drift_generator: ParametersDriftGenerator,
+        gates_translator: Callable[[Gate], Optional[FSimGate]] = try_convert_sqrt_iswap_to_fsim,
```
mpharrigan, line 70: why are the gates all `FSimGate` instead of `PhasedFSimGate`?

cirq/google/calibration/engine_simulator.py
```diff
@@ -0,0 +1,405 @@
...
+            parameters: Parameters to use for each gate. All keys must be stored in canonical order,
+                when the first qubit is not greater than the second one.
+            simulator: Simulator object to use. When None, a new instance of cirq.Simulator() will
+                be created.
+            ideal_when_missing_gate: When set and parameters for some gate for a given pair of
+                qubits are not specified in the parameters dictionary then the
+                FSimGate(theta=π/4, phi=0) gate parameters will be used. When not set and this
+                situation occurs, ValueError is thrown during simulation.
+            ideal_when_missing_parameter: When set and some parameter for some gate for a given pair
+                of qubits are is specified then the matching parameter of FSimGate(theta=π/4, phi=0)
```
mpharrigan, line 198: "are is" -> is

cirq/google/calibration/engine_simulator.py
```diff
@@ -0,0 +1,405 @@
...
+        if (a, b, gate) in self._drifted_parameters:
+            parameters = self._drifted_parameters[(a, b, gate)]
+        elif (b, a, gate) in self._drifted_parameters:
+            parameters = self._drifted_parameters[(b, a, gate)].parameters_for_qubits_swapped()
+        else:
+            parameters = self._drift_generator(a, b, gate)
+            self._drifted_parameters[(a, b, gate)] = parameters
+        return PhasedFSimGate(**parameters.asdict())
+
+    class _PhasedFSimConverter(PointOptimizer):
```
mpharrigan, line 381: I don't think this should be inside the class definition

cirq/google/calibration/engine_simulator.py
```diff
@@ -0,0 +1,405 @@
...
+    def _base_iterator(
+        self,
+        circuit: Circuit,
+        qubit_order: QubitOrderOrList,
+        initial_state: Any,
+    ) -> Iterator[StepResult]:
+        converted = self._convert_to_circuit_with_drift(circuit)
+        return self._simulator._base_iterator(converted, qubit_order, initial_state)
+
+    def _convert_to_circuit_with_drift(self, circuit: Circuit) -> Circuit:
```
mpharrigan, line 365: this can be a free function

cirq/google/calibration/workflow.py
```diff
@@ -269,19 +268,23 @@ def _merge_into_calibrations(
 
 def run_characterizations(
     calibrations: Sequence[PhasedFSimCalibrationRequest],
-    engine: Engine,
-    processor_id: str,
-    gate_set: SerializableGateSet,
+    engine: Union[Engine, PhasedFSimEngineSimulator],
```
mpharrigan, line 271: I really really think this should take a `QuantumEngineSampler` object (and then get rid of `processor_id` and `gate_set`. cc @dstrain115 we may need to add a `run_calibrations` method to q.e.sampler

cirq/google/calibration/workflow.py
```diff
@@ -300,24 +303,36 @@ def run_characterizations(
...
+            request_results = job.calibration_results()
+            results += [
+                calibration.parse_result(result)
+                for calibration, result in zip(calibrations, request_results)
+            ]
+            if progress_func:
+                progress_func(len(results), len(calibrations))
+
+    elif isinstance(engine, PhasedFSimEngineSimulator):
+        results = engine.get_calibrations(calibrations)
```
mpharrigan, line 333: this massive if/else around the whole function body with different argument types is a code smell. Can we add a `run_calibration` mock to `PhasedFSimEngineSimulator`?

mpharrigan (changes requested):
Some grammar things, questions, and nits;

One major thing: this doesn't actually mock out the engine API :)

We also discussed offline how in the future we should extend `cirq.Simulator` to take an arbitrary coherent noise noisemodel and then this `cirq.google` class can wrap that functionality

cirq/google/calibration/engine_simulator.py
```diff
@@ -0,0 +1,405 @@
...
+            gates_translator: Function that translates a gate to a supported FSimGate which will
+                undergo characterization.
+        """
+        self._simulator = simulator
+        self._drift_generator = drift_generator
+        self._gates_translator = gates_translator
+        self._drifted_parameters: Dict[Tuple[Qid, Qid, FSimGate], PhasedFSimCharacterization] = {}
+
+    @staticmethod
+    def create_with_ideal_sqrt_iswap(
```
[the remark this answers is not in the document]
mrwojtek: Done

cirq/google/calibration/engine_simulator.py
```diff
@@ -0,0 +1,405 @@
...
+        Attributes:
+            simulator: Simulator object to use. When None, a new instance of cirq.Simulator() will
+                be created.
+
+        Returns:
+            New PhasedFSimEngineSimulator instance.
+        """
+
+        def sample_gate(_1: Qid, _2: Qid, gate: FSimGate) -> PhasedFSimCharacterization:
+            assert np.isclose(gate.theta, np.pi / 4) and np.isclose(
```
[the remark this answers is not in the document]
mrwojtek: Done

cirq/google/calibration/engine_simulator.py
```diff
@@ -0,0 +1,405 @@
...
+        if simulator is None:
+            simulator = Simulator()
+
+        return PhasedFSimEngineSimulator(
+            simulator, drift_generator=sample_gate, gates_translator=try_convert_sqrt_iswap_to_fsim
+        )
+
+    @staticmethod
+    def create_with_random_gaussian_sqrt_iswap(
+        mean: PhasedFSimCharacterization,
```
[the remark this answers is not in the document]
mrwojtek: The method is labeled as _sqrt_iswap so the default arguments are just SQRT_ISWAP parameters.

cirq/google/calibration/engine_simulator.py
```diff
@@ -0,0 +1,405 @@
...
+        ),
+        random_or_seed: RANDOM_STATE_OR_SEED_LIKE = None,
+    ) -> 'PhasedFSimEngineSimulator':
+        """Creates a PhasedFSimEngineSimulator that introduces a random deviation from the mean.
+
+        The random deviations are described by a Gaussian distribution of a given mean and sigma,
+        for each angle respectively.
+
+        Each gate for each pair of qubits retains the sampled values for the entire simulation, even
+        weh used multiple times within a circuit.
```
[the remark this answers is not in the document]
mrwojtek: Done.

cirq/google/calibration/engine_simulator.py
mrwojtek: Moved to top level. Not that there is a tight coupling between this class and the simulator; that's why I made in an inner class.

cirq/google/calibration/engine_simulator.py
mrwojtek: Done.

cirq/google/calibration/engine_simulator.py
mrwojtek: Done.

cirq/google/calibration/engine_simulator.py
mrwojtek: Heh, done.

cirq/google/calibration/engine_simulator.py
```diff
@@ -0,0 +1,405 @@
...
+
+        Attributes:
+            mean: The mean value for each unitary angle. All parameters must be provided.
+            simulator: Simulator object to use. When None, a new instance of cirq.Simulator() will
+                be created.
+            sigma: The standard deviation for each unitary angle. When some sigma parameter is None,
+                the mean value will be always used for that parameter.
```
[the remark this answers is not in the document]
mrwojtek: Improved.

cirq/google/calibration/engine_simulator.py
```diff
@@ -0,0 +1,405 @@
...
+        for each angle respectively.
+
+        Each gate for each pair of qubits retains the sampled values for the entire simulation, even
+        weh used multiple times within a circuit.
+
+        Attributes:
+            mean: The mean value for each unitary angle. All parameters must be provided.
+            simulator: Simulator object to use. When None, a new instance of cirq.Simulator() will
+                be created.
+            sigma: The standard deviation for each unitary angle. When some sigma parameter is None,
```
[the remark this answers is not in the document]
mrwojtek: Improved.

cirq/google/calibration/engine_simulator.py
mrwojtek: That's cleaner, you're right.

cirq/google/calibration/engine_simulator.py
```diff
@@ -0,0 +1,405 @@
...
+
+    This simulator introduces get_calibrations which allows to simulate
+    cirq.google.run_characterizations requests. The returned calibration results represent the
+    internal state of a simulator. Circuits which are run on this simulator are modified to account
+    for the changes in the unitary parameters as described by the calibration results.
+    """
+
+    def __init__(
+        self,
+        simulator: Simulator,
```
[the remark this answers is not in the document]
mrwojtek: If I set the default argument then I either have to make some dummy drift_generator as well, or I move simulator as a second argument following the drift_generator. I wouldn't want to do that because I want to keep the simulator non-keyword argument. My intention was that this initializer is actually private, and creation happens inside.

cirq/google/calibration/workflow.py
mrwojtek: There is a big difference between `cg.Engine.run_calibrations` and this method: `cg.Engine.run_calibrations` works on a proto/message level and this method, as well as `PhasedFSimEngineSimulator`, work on the dataclass/interface level.

If we move `run_calibrations` to q.e.sampler and stick to the proto/message level, than we need to do two more things:
1. Add support for proto-level simulation to the `PhasedFSimEngineSimulator`
2. Somehow hack the XEB implementation so it could work both in the client-side mode as well as the server-side mode.

If we move this method, which works on high-level interface than those problems are gone but we're promoting this functionality to be more primary. I wanted to do it in a second phase of this work after we evaluate its usefulness.

One option would be to keep the `PhasedFSimEngineSimulator` internal to this module for now, and use it only for testing purposes but we're loosing some potential features there.

cirq/google/calibration/workflow.py
mrwojtek: I feel that adding proto/message level support to `PhasedFSimEngineSimulator` is the wrong way around. We should rather have the `q.e.Sampler` that supports the higher level interface instead. I wouldn't like to do it just yet though.

mrwojtek:
PTAL

cirq/google/calibration/engine_simulator.py
mrwojtek: I guess it was not really important to me. I wanted to connect it more to the hardware gates we plan to support: it's very unlikely that we'd like to have `PhasedFSimGate` supported on the engine API level so `FSimGate` describes better what's applicable. I can change it, although it will trigger changes beyond this PR.

cirq/google/calibration/engine_simulator.py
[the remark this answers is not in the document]
mpharrigan: works for me. I see most people will use the factory functions

cirq/google/calibration/workflow.py
mpharrigan: Offline discussion with @mrwojtek that this would solve my issues, but we'll do it in a follow-up PR

mpharrigan (approved):

Issue #3552, opened 2020-11-30
Relax install_requires constraints


balopat:
Based on https://packaging.python.org/discussions/install-requires-vs-requirements/ "It is not considered best practice to use install_requires to pin dependencies to specific versions". 

We need to loosen up `install_requires` while keeping the CI stable. We don't have good, documented reasons for the lower/upper boundaries. Maybe we could just pin to major versions and let the minors fly free? And then, for CI we could setup a bot that checks for updates weekly and raises a PR with an updated minor version for each requirement (with pur: https://pypi.org/project/pur/) to make sure we're testing against the latest available versions. 

Context: 
 - https://github.com/quantumlib/Cirq/pull/2648
 - https://github.com/quantumlib/Cirq/pull/3551
 - https://github.com/quantumlib/OpenFermion-Cirq/pull/352#issuecomment-513044359

maffoo:
Thanks for raising this issue. For some things I think it makes more sense to test against the oldest supported version (e.g. want to make sure we don't require new features) while for other things it's nice to test against newer versions (e.g. want to test that we take advantage of new features if available). Since it's infeasible to run tests against the entire matrix of possible versions, one possibility would be to run tests twice, once against the oldest supported versions of all deps, and once against the newest supported versions. Clearly, this argues for keeping the list of requirements as small as possible; the packaging effort should help with that.

mpharrigan:
@maffoo , your suggestion is a good one; but that speaks to what we should put in `requirements.txt`; i.e. how we set up the test environment(s). 

I think @balopat is asking about what to put in `install_requires`. I'd naively prefer using only `>=` to set minimum versions of things and blacklist versions we know don't work. In my (perhaps naive!) view of the world, one of our small number of stable dependencies will only rarely break something in cirq. Then we'd have to push out a quick point release to exclude the problematic version. Pinning to major versions would produce more-or-less the same effect. Unless one of our dependencies is very strict about semantic versioning and releases major versions (that are largely compatible, or the api surface area used by cirq is compatible) *or* one of our dependencies is highly stable but refuses to move to version 1.0 like scipy of yore. Then we might cause users to have incompatible environments and we may be less responsive to pushing out a quick point release to certify that yes, the new version is fine.

mpharrigan:
[...]

PR #3726, opened 2021-02-01
make three qubit decomposition tests scipy version dependent

balopat:
This is a fix for older environments. The code already does check whether scipy has the `cossin` method. This PR just makes this explicit in the tests as well - by skipping those tests that would require `cossin` to exist. I also added a test for when scipy is <1.5.0 - in which case we except an `ImportError`. This latter test is not currently tested in our CI (see https://github.com/quantumlib/Cirq/issues/3552 for suggestion about having maybe an upper / lower limit testing for dependencies).

cc @dstrain115


cirq/optimizers/three_qubit_decomposition_test.py
```diff
@@ -29,6 +29,20 @@
 )
 
 
+def _skip_if_scipy(*, has_cossin: bool) -> Callable[[Callable], Callable]:
+    def decorator(func):
+        try:
+            # the cossin function was introduced in 1.5
+            from scipy.linalg import cossin
```
viathor, line 36: Need to tell pylint to ignore unused import.

cirq/optimizers/three_qubit_decomposition_test.py
```diff
@@ -54,6 +68,7 @@ def test_three_qubit_matrix_to_operations(u):
     assert num_two_qubit_gates <= 20, f"expected at most 20 CZ/CNOTs got {num_two_qubit_gates}"
 
 
+@_skip_if_scipy(has_cossin=False)
```
viathor, line 71: Optional: The code in the except block in the 3-qubit decomposition has test coverage check disabled, so skipping the test on l.81 below should not cause coverage errors. However, if we skip this test, then I think coverage error will be reported (since we don't disable coverage on the code using cossin). This is not a problem as long as CI has scipy >= 1.5.0, which it should.

However, I'd add a comment here in case this turns out not to be the case and some poor soul ends up debugging what is going on.

viathor (approved):

cirq/optimizers/three_qubit_decomposition_test.py
balopat: Thanks! I added a comment!

PR #3731, opened 2021-02-02
Add functionality that compensate (calibrate out) for a single-qubit phases of the circuit's PhasedFSimGates

mrwojtek:
This is the single-qubit Z phases compiler. This PR also converges on cleaner return values of all the methods.


cirq/google/calibration/workflow.py
```diff
@@ -14,24 +14,50 @@
...
     WITHOUT_CHI_FLOQUET_PHASED_FSIM_CHARACTERIZATION,
+    ZETA_GAMMA_FLOQUET_PHASED_FSIM_CHARACTERIZATION,
     try_convert_sqrt_iswap_to_fsim,
 )
 from cirq.google.engine import Engine
 from cirq.google.serializable_gate_set import SerializableGateSet
 
 
+@dataclasses.dataclass(frozen=True)
+class CircuitCalibration:
```
mrwojtek, line 48: I could use some better name suggestions for this

cirq/google/calibration/workflow.py
```diff
@@ -14,24 +14,50 @@
...
+    """Circuit with calibration data annotations.
+
+    Attributes:
+        circuit: Circuit instance.
+        moment_allocations: Maps each moment within a circuit to an index of a characterization
+            request or response. None means that there is no characterization data for that moment.
+    """
+
+    circuit: Circuit
+    moment_allocations: List[Optional[int]]
```
mrwojtek, line 58: and this

mrwojtek:
This is now ready for review.

cirq/google/calibration/phased_fsim.py
```diff
@@ -304,6 +304,16 @@ class FloquetPhasedFSimCalibrationOptions(PhasedFSimCalibrationOptions):
 )
 
 
+"""PhasedFSimCalibrationOptions with zeta and gamma angles characterization requests set to True."""
+ZETA_GAMMA_FLOQUET_PHASED_FSIM_CHARACTERIZATION = FloquetPhasedFSimCalibrationOptions(
```
mpharrigan, line 308: add documentation (or even reflect in the name of this object!) that these options are for situations that can be remedied with single-qubit Z rotations.

cirq/google/calibration/workflow.py
mpharrigan: Do you expect users to be constructing this object? Maybe it should be private (i.e. starting with an underscore). As for names: it seems important that we're mapping from the contained circuit to these characterization records, so maybe something like `CircuitRequestResponseMapper` or `CircuitToRecord`. I made up the term "record" meaning "either a request or response". Maybe there's a better word than that

cirq/google/calibration/workflow.py
mpharrigan: yes, allocation seems like the wrong word here. `moment_to_index`? `moment_to_record_index`? `moment_to_req[_index]`?

cirq/google/calibration/workflow.py
```diff
@@ -148,7 +159,10 @@ def make_floquet_request_for_circuit(
             across many circuits.
 
     Returns:
-        Instance of CircuitFloquetPhasedFSimCalibrationRequests.
+        Tuple of:
+          - Circuit together with its calibration metadata connecting to the characterized requests,
```
mpharrigan, line 163: "Circuit and its mapping from moments to indices into the list of calibration requests (the second returned value)"

There's no worry of being too obvious here :)

cirq/google/calibration/workflow.py
```diff
@@ -337,19 +351,150 @@ def run_characterizations(
...
+
+    Args:
+        circuit_calibration: Description of the circuit together with its calibration metadata that
+            is compatible with the characterizations argument.
+        characterizations: List of characterization results.
+        gates_translator: Function that translates a gate to a supported FSimGate which will undergo
+            characterization. Defaults to sqrt_iswap_gates_translator.
+
+    Returns:
+        Calibrated circuit together with its calibration metadata in CircuitCalibration object. The
```
mpharrigan, line 375: Why are we returning the CircuitCalibration instead of just the circuit? Is it to (indirectly) keep track of the mapping from the input circuit to the output circuit, namely after the new Z moments have been added?

cirq/google/calibration/workflow.py
```diff
@@ -337,19 +351,150 @@ def run_characterizations(
...
+    """Compensates circuit against errors in zeta, chi and gamma angles.
+
+    This method creates a new circuit with a single-qubit Z gates added in a such way so that
+    zeta, chi and gamma angles discovered by characterizations are cancelled-out and set to 0.
+
+    This function preserves a moment structure of the circuit. All single qubit gates appear on new
+    moments in the final circuit.
+
+    Args:
+        circuit_calibration: Description of the circuit together with its calibration metadata that
```
mpharrigan, line 368: "A `CircuitRequestMapper` (likely returned from `run_calibrations`) whose mapping argument corresponds to the results in the `characterizations` argument"

^^ suggested wording, but you should flesh out, make sure it's accurate, and replace the names with whatever you settle on. I find it helpful to spell out explicitly exactly how these parts fit together.

cirq/google/calibration/workflow.py
```diff
@@ -337,19 +351,150 @@ def run_characterizations(
...
+            is compatible with the characterizations argument.
+        characterizations: List of characterization results.
```
mpharrigan, line 370: "List of characterization results (likely returned from `run_calibrations`). This should correspond to the circuit and mapping in the `circuit_request_mapper` argument".

Being redundant helps drive home the point

cirq/google/calibration/workflow.py
```diff
@@ -337,19 +351,150 @@ def run_characterizations(
...
+        other = []
+        new_moment_mapping = None
+        for op in moment:
+            if not isinstance(op, GateOperation):
+                raise IncompatibleMomentError(
+                    'Moment contains operation different than GateOperation'
+                )
+
+            if isinstance(op.gate, (MeasurementGate, SingleQubitGate)):
+                other.append(op)
```
mpharrigan, line 403: readability trick, if you like:

```python
if isisnglequbit:
  other.append(op)
  continue
a, b = op.qubits
look_ma_no_indent()
...
```

cirq/google/calibration/workflow.py
```diff
@@ -337,19 +351,150 @@ def run_characterizations(
...
+                    raise ValueError(f'Missing characterization data for moment {moment}')
+
+                pair_parameters = parameters.get_parameters(a, b)
+                if pair_parameters is None:
+                    raise ValueError(
+                        f'Missing characterization data for pair {(a, b)} in {parameters}'
+                    )
+                pair_parameters = pair_parameters.merge_with(default_phases)
+
+                corrected = PhaseCorrectedFSimOperations(
```
mpharrigan, line 423: see comment below, but I think the variable name should be `corrections`

cirq/google/calibration/workflow.py
```diff
@@ -337,19 +351,150 @@ def run_characterizations(
...
+    compensated_allocations = []
+    for moment, characterization_index in zip(
+        circuit_calibration.circuit, circuit_calibration.moment_allocations
+    ):
+        parameters = None
+        if characterization_index is not None:
+            parameters = characterizations[characterization_index]
+
+        decompositions = []
+        other = []
```
mpharrigan, line 394: type annotations could be helpful here for `decompositions` `other` and `new_moment_mapping`

cirq/google/calibration/workflow.py
```diff
@@ -337,19 +351,150 @@ def run_characterizations(
...
+        calibrated circuit has single-qubit Z gates added which compensates for the true gates
+        imperfections.
+    """
+    if len(circuit_calibration.circuit) != len(circuit_calibration.moment_allocations):
+        raise ValueError('Moment allocations does not match circuit length')
+
+    default_phases = PhasedFSimCharacterization(zeta=0.0, chi=0.0, gamma=0.0)
+
+    compensated = Circuit()
+    compensated_allocations = []
```
mpharrigan, line 385: type annotation could be helpful here. I'm confused by the word "allocation"

cirq/google/calibration/workflow.py
```diff
@@ -337,19 +351,150 @@ def run_characterizations(
...
+                        new_moment_mapping == corrected.moment_allocations
+                    ), f'Inconsistent decompositions with a moment {moment}'
+
+        if other and decompositions:
+            raise IncompatibleMomentError(f'Moment {moment} contains mixed operations')
+        elif other:
+            compensated += Moment(other)
+            compensated_allocations.append(characterization_index)
+        elif decompositions:
+            for operations in zip_longest(*decompositions, fillvalue=()):
```
mpharrigan, line 441: why are you doing this zip(*x)? This "transposes" a nested list, no? I'm confused why it's needed here.

cirq/google/calibration/workflow.py
```diff
@@ -337,19 +351,150 @@ def run_characterizations(
...
-            does not require characterization.
+        operations: Tuple of tuple of operations that describe the gate. The first index iterates
+            over moments of the composed operation.
+        moment_allocations: List of indices pointing to the characterizations for each moment in the
+            composed operation.
     """
 
-    results: List[PhasedFSimCalibrationResult]
-    moment_allocations: List[Optional[int]]
+    def __init__(
```
mpharrigan, line 459: my knee-jerk reaction is that this should be a free function that returns a "dumb" dataclass container, but maybe that's excessive. 

As a piece of evidence: when I was reading through the above function I couldn't intuit what PhaseCorrectedFsimOperations was since I expected it to be a data container but it's actually doing the "work" of figuring out the corrections (in the constructor, no less)

cirq/google/calibration/workflow.py
```diff
@@ -337,19 +351,150 @@ def run_characterizations(
...
+        elif decompositions:
+            for operations in zip_longest(*decompositions, fillvalue=()):
+                compensated += Moment(operations)
+            assert new_moment_mapping is not None  # Required for mypy
+            compensated_allocations += new_moment_mapping
+
+    return CircuitCalibration(compensated, compensated_allocations)
+
+
+class PhaseCorrectedFSimOperations:
```
mpharrigan, line 449: Maybe name `FSimPhaseCorrections`?

cirq/google/calibration/workflow.py
```diff
@@ -390,21 +535,83 @@ def run_floquet_characterization_for_circuit(
...
+          - List of characterizations that were triggered in order to calibrate the circuit.
+    """
+    circuit_calibration, requests = make_floquet_request_for_circuit(
+        circuit, options, gates_translator, merge_subsets=merge_subsets
+    )
+    characterizations = run_characterizations(
+        requests,
+        engine,
+        processor_id,
+        gate_set,
```
mpharrigan, line 610: since the formatter is putting these all on their own lines, you could make them all kwargs to 1) drill home the point that you're just forwarding the values 2) prevent accidental reordering either here or changing the `run_characterizations` signature

mpharrigan (commented):
This design looks fine and makes sense to me.

Per usual, I've left a bunch of comments on naming and readability which were basically made as I was going through trying to understand the code myself, with some suggestions on what would have helped me get up-to-speed faster :)

cirq/google/calibration/workflow.py
mrwojtek: This should be actually CircuitWithCharacterizationIndex or CircuitWithCalibrationIndex in the future. I made it to short CircuitWithCalibration for now. I want to schedule meeting with some of you to discuss all the naming issues soon.

cirq/google/calibration/workflow.py
mrwojtek: I made it moment_to_calibration for consistency.

cirq/google/calibration/phased_fsim.py
mrwojtek: Done

cirq/google/calibration/workflow.py
mrwojtek: Done.

cirq/google/calibration/workflow.py
mrwojtek: Added a comment: this is updated CircuitWithCalibration; the idea is that calibration can be composed. Will see how it ends.

cirq/google/calibration/workflow.py
mrwojtek: Done.

cirq/google/calibration/workflow.py
mrwojtek: Done.

cirq/google/calibration/workflow.py
mrwojtek: Done

cirq/google/calibration/workflow.py
mrwojtek: Done

cirq/google/calibration/workflow.py
mrwojtek: Done

cirq/google/calibration/workflow.py
mrwojtek: The assumption is that the circuit provided here is a generic user circuit. This transposition allows me to split all the decompositions moment by moment and add those moments one by one into the circuit.

cirq/google/calibration/workflow.py
mrwojtek: You're right, this is a much better design.

cirq/google/calibration/workflow.py
mrwojtek: Done.

cirq/google/calibration/workflow.py
mrwojtek: Done.

mrwojtek:
PTAL

cirq/google/calibration/workflow.py
mrwojtek: As for user construction, this object is a public interface of a method and shouldn't be public. I can imagine situations where advanced users would create it.

cirq/google/calibration/workflow.py
mpharrigan: got it

mpharrigan (approved):
sweet

PR #3733, opened 2021-02-02
import os before using os

### LINKED ISSUE #3730: [tutorials/google/reservations] name 'os' is not defined
If you edit the `project_id` field (like it says to do if running in Jupyter) the `import os` import never runs, but the else clause uses `os`; thus an import error occurs

viathor:
Fixes #3730.


mpharrigan (approved):
is there a way to test for this? probably not, right?

viathor:
I think we have [tests for notebooks](https://github.com/quantumlib/Cirq/issues?q=is%3Aopen+is%3Aissue+label%3Aarea%2Fnotebook-testing), so there should be a way. Not sure the same is true for checking test coverage...

balopat:
> I think we have [tests for notebooks](https://github.com/quantumlib/Cirq/issues?q=is%3Aopen+is%3Aissue+label%3Aarea%2Fnotebook-testing), so there should be a way. Not sure the same is true for checking test coverage...

Yeah, this notebook is excluded as it uses the engine API :/ 
There could be a way to make notebooks somehow mockable + even do test coverage possibly but currently it's not possible to test this.

PR #3735, opened 2021-02-03
Add the ability to do hybrid MPS/dense/inbetween simulations by having qubit grouping

tonybruguier:


cirq/contrib/quimb/mps_simulator.py
```diff
@@ -306,10 +318,14 @@ def __init__(
             sum_prob_atol: Because the computation is approximate, the sum of
                 the probabilities is not 1.0. This parameter is the absolute
                 deviation from 1.0 that is allowed.
+            grouping: How to group qubits together, if None all are individual.
             initial_state: An integer representing the initial state.
         """
         self.qubit_map = qubit_map
+        self.grouping = qubit_map if grouping is None else grouping
```
95-martin-orion, line 325: Should we verify that all `qubit_map` qubits appear in `grouping`?

cirq/contrib/quimb/mps_simulator_test.py
```diff
@@ -248,9 +278,15 @@ def test_empty_step_result():
 
 def test_state_equal():
     q0, q1 = cirq.LineQubit.range(2)
-    state0 = ccq.mps_simulator.MPSState(qubit_map={q0: 0}, rsum2_cutoff=1e-3, sum_prob_atol=1e-3)
-    state1a = ccq.mps_simulator.MPSState(qubit_map={q1: 0}, rsum2_cutoff=1e-3, sum_prob_atol=1e-3)
-    state1b = ccq.mps_simulator.MPSState(qubit_map={q1: 0}, rsum2_cutoff=1729.0, sum_prob_atol=1e-3)
+    state0 = ccq.mps_simulator.MPSState(
+        qubit_map={q0: 0}, rsum2_cutoff=1e-3, sum_prob_atol=1e-3, grouping=None
```
95-martin-orion, line 282: Is explicit assignment of `grouping` here a style choice to clarify test behavior? If not, it's fine to remove them and rely on the default.

Same elsewhere in this file.

95-martin-orion (changes requested):
Some minor nits, but this is looking good.

For my own benefit: IIUC, the extra `grouping` info lets the MPS simulator know which qubits are already entangled, allowing it to be used on an output or intermediate state of another simulator. Is that correct?

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done. Also added a test.

cirq/contrib/quimb/mps_simulator_test.py
tonybruguier: Removed here and elsewhere.

tonybruguier:
I was trying to implement section VI.A of the paper. The way I see it, is that we currently have two extremes:
- We do a dense simulation, where we have a single tensor of dimension 2**n that corresponds to all the qubits
- We do a (non-grouped) MPS simulation, where we have a n tensors of dimensions (2, mu_1_2, mu_1_3...) where the mu indices are the entanglement.

We could do something in between. For example, we could group split the qubits in two groups. We thus would have 2 tensors of dimensions (2**(n/2), mu) where there is a single mu index for the entanglement between the two groups. We could also group arbitrarily, which is the code intends to allow.

Thanks for the review. I tried to address your comments, but happy to do more if required. PTAL.

95-martin-orion (approved):

Issue #3737, opened 2021-02-03
Remove references to cirq.google in cirq core / contrib


balopat:
In a bunch of our tests, there are a bunch of cirq.google references that need to be cleaned up in order to be able to extract cirq.google as a separate module.

```
git grep cirq.google | grep -v "cirq/google" | grep -v "json_test" | grep -v "docs" | grep -v "examples" | grep -v "json_serialization" | grep -v "setup"
```

These need to be replaced with cirq.google independent logic. 

```
cirq/circuits/circuit_dag_test.py:    circuit = cirq.Circuit(cirq.X(q0), cirq.Y(q0), device=cirq.google.Bristlecone)
cirq/circuits/circuit_dag_test.py:    q0, q1, q2, q3 = cirq.google.Bristlecone.col(5)[:4]
cirq/circuits/circuit_dag_test.py:        device=cirq.google.Bristlecone,
cirq/circuits/circuit_test.py:import cirq.google as cg
cirq/circuits/circuit_test.py:    c = circuit_cls(device=cirq.google.Bristlecone)
cirq/circuits/circuit_test.py:    assert c2.device is cirq.google.Bristlecone
cirq/circuits/circuit_test.py:    c = circuit_cls(device=cirq.google.Bristlecone)
cirq/circuits/circuit_test.py:    assert repr(c) == f'cirq.{circuit_cls.__name__}(device=cirq.google.Foxtail)'
cirq/circuits/circuit_test.py:], device=cirq.google.Foxtail)"""
cirq/circuits/frozen_circuit_test.py:        c.device = cirq.google.devices.Foxtail
cirq/experiments/random_quantum_circuit_generation_test.py:    yield cirq.google.SYC(a, b)
cirq/experiments/random_quantum_circuit_generation_test.py:            lambda a, b, _: cirq.google.SYC(a, b),
cirq/experiments/random_quantum_circuit_generation_test.py:            lambda a, b, _: cirq.google.SYC(a, b),
cirq/experiments/random_quantum_circuit_generation_test.py:            lambda a, b, _: cirq.google.SYC(a, b),
cirq/experiments/random_quantum_circuit_generation_test.py:            lambda a, b, _: cirq.google.SYC(a, b),
cirq/ops/three_qubit_gates_test.py:    ops = tuple(cirq.flatten_op_tree(cirq.google.ConvertToXmonGates().convert(op)))
[...]

PR #3738, opened 2021-02-03
Remove cirq.google references from circuit_test and circuit_dag_test

balopat:
Remove cirq.google references from circuit_test and circuit_dag_test.

Related to #3737.


cirq/circuits/circuit_test.py
```diff
@@ -581,44 +579,126 @@ def test_concatenate_with_device():
     assert len(cone) == 0
 
 
+class FakeXmonDevice(cirq.Device):
+    """A Fake XMon device that was created to ensure certain Device validation features are
```
95-martin-orion, line 583: I think it's important to note (either here or elsewhere in this file) that what we're testing is *not* the specific behavior of Foxtail and Bristlecone, but rather a set of common behaviors for which these processors are reasonable examples.

cirq/circuits/circuit_test.py
```diff
@@ -581,44 +579,126 @@ def test_concatenate_with_device():
...
+        ops.XPowGate,
+        ops.YPowGate,
+        ops.ZPowGate,
+    ),
+    qubits={
+        cirq.GridQubit(0, 0),
+        cirq.GridQubit(1, 0),
+        cirq.GridQubit(0, 1),
+        cirq.GridQubit(1, 1),
+        cirq.GridQubit(0, 6),
```
95-martin-orion, line 648: This may generate confusion - I'd recommend `cirq.GridQubit.rect(2,7)` instead.

cirq/circuits/circuit_test.py
```diff
@@ -3375,25 +3455,25 @@ def test_insert_operations_errors():
...
+    c = cirq.Circuit(device=foxy)
 
     with pytest.raises(ValueError, match='Unsupported qubit type'):
         # Wrong type of qubit.
         c.append(cirq.Z(cirq.NamedQubit('q')))
     with pytest.raises(ValueError, match='Qubit not on device'):
         # A qubit that's not on the device.
         c[:] = [cirq.Moment([cirq.Z(cirq.GridQubit(-5, 100))])]
-    c.append(cirq.Z(cirq.GridQubit(0, 0)))
+    c.append(cirq.Z(cirq.GridQubit(0, 1)))
```
95-martin-orion, line 3466: Why this change?

cirq/circuits/circuit_test.py
```diff
@@ -336,19 +334,19 @@ def test_repr(circuit_cls):
 ])"""
     )
 
-    c = circuit_cls(device=cg.Foxtail)
+    c = circuit_cls(device=foxy)
     cirq.testing.assert_equivalent_repr(c)
-    assert repr(c) == f'cirq.{circuit_cls.__name__}(device=cirq.google.Foxtail)'
+    assert repr(c) == f'cirq.{circuit_cls.__name__}(device=cirq.circuits.circuit_test.foxy)'
```
95-martin-orion, line 339: Can this just be `device=foxy` instead? Same below.

95-martin-orion (changes requested):

cirq/circuits/circuit_dag_test.py
```diff
@@ -109,9 +109,14 @@ def test_from_circuit():
     assert sorted(circuit.all_qubits()) == sorted(dag.all_qubits())
 
 
+class TestDevice(cirq.Device):
+    def validate_operation(self, operation: cirq.Operation) -> None:
```
dstrain115, line 113: Can't you use cirq.UNCONSTRAINED_DEVICE?

cirq/circuits/circuit_test.py
dstrain115: Yes, I would call this something more specific, like CZOnlyTestDevice or something like that.

cirq/circuits/circuit_test.py
```diff
@@ -581,44 +579,126 @@ def test_concatenate_with_device():
...
+        return not (
+            isinstance(operation.gate, cirq.CZPowGate)
+            and any(isinstance(op.gate, cirq.CZPowGate) for op in moment.operations)
+        )
+
+    def __repr__(self):
+        return self._repr
+
+
+foxy = FakeXmonDevice(
```
dstrain115, line 635: Should this be capitalized?

cirq/circuits/circuit_test.py
```diff
@@ -581,44 +579,126 @@ def test_concatenate_with_device():
...
+        cirq.GridQubit(0, 0),
+        cirq.GridQubit(1, 0),
+        cirq.GridQubit(0, 1),
+        cirq.GridQubit(1, 1),
+        cirq.GridQubit(0, 6),
+    },
+    name='cirq.circuits.circuit_test.foxy',
+)
+
+bcone = FakeXmonDevice(
```
dstrain115, line 653: I would capitalize this.

cirq/circuits/circuit_test.py
balopat: yes!

cirq/circuits/circuit_dag_test.py
balopat: Ah, nice, let me try that

cirq/circuits/circuit_test.py
balopat: Right, okay, I will remove the reference to anything Foxtail / Bristlecone and rename the variable names as well. I had originally "ValidatingTestDevice" but that was too generic, maybe I'll go back to that.

cirq/circuits/circuit_test.py
balopat: Otherwise there was an overlapping moment exception. I added the logic now to the testclass in `can_add_operation_into_moment` to call `super()` like Foxtail does, so now this change can be reverted.

cirq/circuits/circuit_test.py
balopat: Oh, actually, it cannot, as this is then used above in the assert_equivalent_repr - so the testclass have to generate repr that actually works.

cirq/circuits/circuit_test.py
balopat: done.

balopat:
@95-martin-orion @dstrain115 ptal

95-martin-orion (approved):

PR #3739, opened 2021-02-03
[XEB] Optimize two qubit circuits

mpharrigan:
Implement an outer-loop optimizer to fit parameterized circuits to maximize XEB fidelity.


cirq/experiments/fidelity_estimation.py
```diff
@@ -32,17 +33,21 @@
 
 import numpy as np
 import pandas as pd
+import scipy.optimize
+import sympy
 import tqdm
 
 from cirq import ops, protocols, sim
 from cirq.circuits import Circuit
+from cirq.google.calibration.phased_fsim import PhasedFSimCalibrationOptions
```
balopat, line 42: I am trying to remove cirq.google dependencies from cirq core. Is this really necessary here? Should we pull up PhasedFSimCalibrationOptions to cirq core? PhasedFSimGate itself lives in cirq.google. Alternatively we can think about introducing cirq/google/experiments maybe?

mrwojtek: There is also another alternative to consider:
Make cirq.experiments.fidelity_estimation independent from the PhasedFSimCalibrationOptions and move the XEB options to the cirq.google.calibration. The XEB in cirq.google.calibration would then be a very thin wrapper on complicated functionality in the cirq.experiments.fidelity_estimation. It would only translate between options and estimation logic. I think this would be a good separation because in the end we'll have:
 1) XEB estimation in the cirq.google.experiments
 2) XEB estimation in the QE
 3) Floquet estimation in the QE
The interfaces of 1) and 2) are different and they would be unified in cirq.google.calibration.

There is not much effort do it like that actually: the variables from lines 757-761 would just need to be a method arguments and characterize_phased_fsim_parameters_with_xeb instead of options would need to take a simplex. I think it would be even cleaner that way.

All the options and dataclasses can be a subject of a follow-up PR. Hopefully, it would not clutter the notebook too much.

mrwojtek (commented):
The notebook looks great.

cirq/experiments/fidelity_estimation.py
mpharrigan: Yeah, the inheritance from `PhasedFSimCalibrationOptions` isn't necessary for the `cirq.experiments` code or this PR. I wanted to foreshadow how it would be integrated into @mrwojtek 's cirq.google utilities, which has in some sense worked because the above comment goes through a good way for this to be done :)

cirq/experiments/fidelity_estimation.py
```diff
@@ -642,3 +647,195 @@ def per_cycle_depth(df):
...
+    fatol: float = 1e-3,
+    verbose: bool = True,
+    pool: Optional['multiprocessing.pool.Pool'] = None,
+):
+    """Run a classical optimization to fit phased fsim parameters to experimental data, and
+    thereby characterize PhasedFSim-like gates.
+
+    Args:
+        sampled_df: The DataFrame of sampled two-qubit probability distributions returned
+            from `sample_2q_xeb_circuits`.
```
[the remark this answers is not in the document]
mpharrigan: Right now, you'd have to save the dataframe. I plan on refactoring the sampling function to return dataclass objects (per row) with all the data and metadata (bitstrings, timestamps, etc.) with a callback in the sampling function to save-as-you-go

cirq/experiments/fidelity_estimation.py
balopat: Great!

mpharrigan: removed

cirq/experiments/fidelity_estimation.py
```diff
@@ -642,3 +647,195 @@ def per_cycle_depth(df):
...
+        fatol: The `fatol` argument for Nelder-Mead. This is the absolute error for convergence
+            in the function evaluation.
+        pool: An optional multiprocessing pool to execute circuit simulations in parallel.
+    """
+    initial_simplex, names = phased_fsim_options.get_initial_simplex_and_names(
+        initial_simplex_step_size=initial_simplex_step_size
+    )
+    x0 = initial_simplex[0]
+
+    def _f(x):
```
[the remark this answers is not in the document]
mpharrigan: done

cirq/experiments/fidelity_estimation.py
```diff
@@ -642,3 +647,195 @@ def per_cycle_depth(df):
...
+
+    As such, the default for theta is changed to -pi/4 and the parameterization
+    predicate seeks out sqrt(ISWAP) gates.
+    """
+
+    theta_default: float = -np.pi / 4
+
+    @staticmethod
+    def should_parameterize(op: 'cirq.Operation') -> bool:
+        return op.gate == SQRT_ISWAP
```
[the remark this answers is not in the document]
mpharrigan: Yes. In light of the `cirq.google` discussion above, I'm going to punt on this until I open the PR for wrapping this in cirq.google calibration api functionality

cirq/experiments/fidelity_estimation.py
```diff
@@ -642,3 +647,195 @@ def per_cycle_depth(df):
...
+        if self.parameterize_phi:
+            x0 += [self.phi_default]
+            names += [PHI.name]
+
+        x0 = np.asarray(x0)
+        n_param = len(x0)
+        initial_simplex = [x0]
+        for i in range(n_param):
+            basis_vec = np.zeros(n_param)
+            basis_vec[i] = 1
```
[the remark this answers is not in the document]
mpharrigan: nice trick

mpharrigan:
PTAL @mrwojtek and @balopat

cirq/experiments/fidelity_estimation.py
```diff
@@ -642,3 +646,195 @@ def per_cycle_depth(df):
...
     return df.reset_index().groupby('cycle_depth').apply(per_cycle_depth).reset_index()
+
+
+# mypy issue: https://github.com/python/mypy/issues/5374
+@dataclass(frozen=True)  # type: ignore
+class XEBPhasedFSimCalibrationOptions:
+    """Options for calibrating a PhasedFSim-like gate using XEB.
+
+    You may want to use more specific subclasses like `SqrtISwapXEBOptions`
+    which have sensible defaults.
```
mrwojtek, line 657: Instead of creating subclasses for defaults, I would crate a creation method that creates instance with correct values. I would also assume that all the parameters must be provided explicitly. The defaults here are meaningless and impractical, right?

cirq/experiments/fidelity_estimation.py
```diff
@@ -642,3 +646,195 @@ def per_cycle_depth(df):
...
+
+    Attributes:
+        parameterize_theta: Whether to characterize θ angle.
```
mrwojtek, line 660: Is there a reason to call this parametrize_? In the beginning I was using estimate_ but based on feedback from some people we settled down on characterize_

cirq/experiments/fidelity_estimation.py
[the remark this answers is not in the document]
mrwojtek: Alright, sounds reasonable.

cirq/experiments/fidelity_estimation.py
```diff
@@ -642,3 +646,195 @@ def per_cycle_depth(df):
...
+                        theta=theta,
+                        zeta=zeta,
+                        chi=chi,
+                        gamma=gamma,
+                        phi=phi,
+                    ).on(*op.qubits)
+                ]
+            else:
+                new_ops += [op]
+        new_moments += [ops.Moment(new_ops)]
```
mrwojtek, line 777: I would be tempted to rewrite this as a nested list-generator expression. Up to you.

cirq/experiments/fidelity_estimation.py
```diff
@@ -43,6 +46,7 @@
     import cirq
     import multiprocessing
 
+THETA, ZETA, CHI, GAMMA, PHI = sympy.symbols('theta zeta chi gamma phi')
```
mrwojtek, line 49: THETA_SYMBOL, ZETA_SYMBOL or THETA_SYM, ZETA_SYM?

cirq/experiments/fidelity_estimation.py
```diff
@@ -642,3 +646,195 @@ def per_cycle_depth(df):
         return pd.Series({'fidelity': fid_lsq})
 
     return df.reset_index().groupby('cycle_depth').apply(per_cycle_depth).reset_index()
+
+
+# mypy issue: https://github.com/python/mypy/issues/5374
+@dataclass(frozen=True)  # type: ignore
+class XEBPhasedFSimCalibrationOptions:
```
mrwojtek, line 653: In this context it fits better as XEBPhasedFSimCharacterizationOptions or XEBPhasedFSimParametrizationOptions

mpharrigan: done (characterization)

cirq/experiments/fidelity_estimation.py
mpharrigan: done

cirq/experiments/fidelity_estimation.py
mpharrigan: Some people don't like gnarly list comprehensions. I think it could look nicer if I were allowed to format it:

```python
    circuit = Circuit([
        ops.Moment([
            ops.PhasedFSimGate(theta=theta, zeta=zeta, chi=chi, gamma=gamma, phi=phi).on(*op.qubits)
            if options.should_parameterize(op)
            else op for op in moment.operations
        ]) for moment in circuit.moments
    ])
```

Under the black formatter:

```python
    circuit = Circuit(
        [
            ops.Moment(
                [
                    ops.PhasedFSimGate(theta=theta, zeta=zeta, chi=chi, gamma=gamma, phi=phi).on(
                        *op.qubits
                    )
                    if options.should_parameterize(op)
                    else op
                    for op in moment.operations
                ]
            )
            for moment in circuit.moments
        ]
    )

```


Without comprehensions:
```python
    new_moments = []
    for moment in circuit.moments:
        new_ops = []
        for op in moment.operations:
            if options.should_parameterize(op):
                new_ops += [
                    ops.PhasedFSimGate(
                        theta=theta,
                        zeta=zeta,
                        chi=chi,
                        gamma=gamma,
                        phi=phi,
                    ).on(*op.qubits)
                ]
            else:
                new_ops += [op]
        new_moments += [ops.Moment(new_ops)]
    circuit = Circuit(new_moments)
```

@balopat or others: do you have a preference for style here?

cirq/experiments/fidelity_estimation.py
mpharrigan: changed to characterize

cirq/experiments/fidelity_estimation.py
mpharrigan: So I'm going through and re-writing this without subclassing but I'm worried I'm just inventing the notion of subclasses. The rough sketch is 

 - move `should_parameterize` to a callable class attribute
 -  write a factory function that fills in the correct values

The factory function--however--should still accept all the characterize_xxx bools *and also* most of the `xxx_default` floats and override the defaults inherent in the factory function. The reason why we still need to override is because we still want people to choose whether or not they're doing certain angles and we may have additional information about a good starting point for optimizing the angles (for example: theory that phi is a small non-zero number or results from a previous characterization)

At this point my factory function is encoding a lot of the behaviors of subclassing; and passing a callable around is similar to overriding an abstract method.

cirq/experiments/fidelity_estimation.py
95-martin-orion: This doesn't answer the style question, but both Circuit and Moment accept raw comprehensions (i.e. no brackets):
```python
circuit = Circuit(
    ops.Moment(
        ops.PhasedFSimGate(theta=theta, zeta=zeta, chi=chi, gamma=gamma, phi=phi).on(
            *op.qubits
        )
        if options.should_parameterize(op)
        else op
        for op in moment.operations
    )
    for moment in circuit.moments
)
```
Doing this makes the `black` formatting a little less egregious.

maffoo: +1 from me for raw comprehensions. In this particular case, you could also clean it up by reusing the gate:
```python
fsim_gate = ops.PhaseFSimGate(theta=theta, zeta=zeta, chi=chi, gamma=gamma, phi=phi)

circuit = Circuit(
    ops.Moment(
        fsim_gate.on(*op.qubits) if options.should_parameterize(op) else op
        for op in moment.operations
    )
    for moment in circuit.moments
)
```
It might also be possible to wrap up that conditional expression into a method on the `XEBPhasedFSimCharacterizationOptions` class:
```python
circuit = Circuit(
    ops.Moment(
        options.parameterize(op) for op in moment.operations
    )
    for moment in circuit.moments
)
```
Finally, as a general comment, this pattern is common enough that we might consider adding a `map_op` method on `cirq.Circuit` that takes a function from op -> op.

balopat: Yupp, +1 for list comprehension, it's more concise and for me more readable. I also really like @maffoo's comment on moving the responsibility of parametrization to the option. 

Re: `map_op` - it feels like then we're entering "compiler" territory with the circuit. While it may seem like a convenient syntactical sugar, it would be a bit surprising to me, as it is not the circuit's responsibility to transform itself.

mpharrigan: tell that to `Circuit.transform_qubits`

mpharrigan: Thanks all for the advice!

cirq/experiments/fidelity_estimation.py
mrwojtek: I missed the existence of `should_parametrize`, this is in fact a strong pointer for subclassing.

As for defaults, I'm still not really convinced but I'm not pushing for this as well. If there are defaults which you think are reasonable than it would be good to have a create method with name that describes what those defaults are. For example, the current defaults would be something like `create_identity_gate_with_all_angles_options`. Again, this is something which will become more clear later I guess

mrwojtek (approved):

95-martin-orion (approved):
CODEOWNERS approval on behalf of @mrwojtek.

cirq/experiments/fidelity_estimation.py
mpharrigan: The defaults are used to seed the classical optimizer. *In theory* you should be able to find the true angles starting from any initial guess, it might just take longer. In practice it could send you into a different local optimum. The defaults correspond to what gate you're replacing/parameterizing; which is why setting the default-defaults currently goes in the subclass (with the `should_parameterize` override)

mpharrigan:
Test failure is unrelated (?)

```
_____________________________ test_api_retry_times _____________________________

client_constructor = <MagicMock name='QuantumEngineServiceClient' spec='QuantumEngineServiceClient' id='5062465808'>
mock_time = <MagicMock name='sleep' id='5064730512'>

    @mock.patch('time.sleep', return_value=None)
    @mock.patch.object(quantum, 'QuantumEngineServiceClient', autospec=True)
    def test_api_retry_times(client_constructor, mock_time):
        grpc_client = setup_mock_(client_constructor)
        grpc_client.get_quantum_program.side_effect = exceptions.ServiceUnavailable(
            'internal error')
    
        client = EngineClient(max_retry_delay_seconds=0.3)
        with pytest.raises(TimeoutError,
                           match='Reached max retry attempts.*internal error'):
            client.get_program('proj', 'prog', False)
        assert grpc_client.get_quantum_program.call_count == 3
    
>       assert len(mock_time.call_args_list) == 2
E       AssertionError: assert 52 == 2
E        +  where 52 = len([call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1)... call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.2)])
E        +    where [call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1)... call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.1),\n call(0.2)] = <MagicMock name='sleep' id='5064730512'>.call_args_list

cirq/google/engine/engine_client_test.py:884: AssertionError
```

PR #3741, opened 2021-02-03
Repetition IDs for CircuitOperation

### LINKED ISSUE #3663: Repeated CircuitOperation with measurements
Currently, CircuitOperation does not allow repeating the sub-circuit if there is a measurement in it. This feature is essential for flow-control (#3234) through repeated/conditional substructures.

Use repetition identifiers as measurement key prefixes to uniquely identify each measurement
RFC: https://tinyurl.com/measurement-repeated-circuitop

P2 - we should do it in the next couple of quarters

smitsanghavi:
Useful for keying individual measurements in repeated `CircuitOperation`s

Closes #3663


cirq/circuits/circuit_operation.py
```diff
@@ -188,53 +262,97 @@ def _json_dict_(self):
...
     ) -> 'CircuitOperation':
         """Returns a copy of this operation repeated 'repetitions' times.
+         Each repetition instance will be identified by a single repetition_id.
 
         Args:
             repetitions: Number of times this operation should repeat. This
                 is multiplied with any pre-existing repetitions.
+            repetition_ids: List of IDs, one for each repetition. The length
+                should be equal to either `repetitions` or
+                `repetitions*self.repetitions`. In the latter case, these
```
95-martin-orion, line 300: What is the use case for the `repetitions * self.repetitions` / override-all-IDs behavior? It seems like the `repetitions` / prefixing case should be sufficient.

Separately, we shouldn't rely on the size of this argument to determine its purpose. If we need both behaviors, we should have a separate argument (or method) for each.

cirq/circuits/circuit_operation_test.py
```diff
@@ -480,6 +545,33 @@ def test_decompose_loops():
...
+        cirq.CX(b, a),
+        cirq.measure(b, a, key='1-m'),
+        cirq.H(b),
+        cirq.CX(b, a),
+        cirq.measure(b, a, key='2-m'),
+    )
+    assert cirq.Circuit(cirq.decompose_once(op)) == expected_circuit
+
+    with pytest.raises(ValueError, match='circuit is not invertible'):
+        _ = base_op.repeat(-2)
```
95-martin-orion, line 572: Is this required here? There's another test that covers this, IIRC.

cirq/circuits/circuit_operation_test.py
```diff
@@ -542,6 +634,48 @@ def test_decompose_nested():
     assert cirq.Circuit(cirq.decompose(final_op)) == expected_circuit
 
 
+def test_decompose_repeated_nested_measurements():
```
95-martin-orion, line 637: [No action required] Cool to see this realized in test form!

cirq/protocols/json_test_data/CircuitOperation.json
```diff
@@ -176,7 +177,8 @@
     "param_resolver": {
       "cirq_type": "ParamResolver",
       "param_dict": []
-    }
+    },
+    "repetition_ids": ["0", "1"]
```
95-martin-orion, line 181: We should assign explicit repetition IDs for this case (or for a new case) to verify that it behaves as expected.

cirq/circuits/circuit_operation.py
```diff
@@ -188,53 +262,97 @@ def _json_dict_(self):
...
-            .repeat(repetitions)
+            .repeat(repetitions, repetition_ids)
         )
 
     # Methods for constructing a similar object with one field modified.
 
     def repeat(
-        self,
-        repetitions: INT_TYPE,
+        self, repetitions: INT_TYPE, repetition_ids: Optional[List[str]] = None
```
95-martin-orion, line 290: What do you think about making both args optional? I think it'd be nice for `repeat(['a', 'b', 'c'])` to work without specifying the number of IDs. (Requires addressing `repetitions` vs `repetitions * self.repetitions` first)

95-martin-orion (changes requested):

cirq/circuits/circuit_operation.py
smitsanghavi: Since we assign default `repetition_ids` even when the circuit is non-measurement, without override-all-IDs the following will be true
```
circuit_op.repeat(2).repeat(3) != circuit_op.repeat(6)
circuit_op.repeat(2).repetition_ids == ['0-0', '1-0']
circuit_op.repeat(2, ['a', 'b']).repetition_ids == ['a-0', 'b-0']
```

The above can be handled by special casing based on if the base or input repetition_ids are "defaults"/None. But if we are overriding in some cases, might as well make that behavior available to the caller. For e.g. if `circuit_op.repeat(2).repeat(3) == circuit_op.repeat(6)`, it would be nice to be able to specify all 6 `repetition_ids` in the left-side construction.

But I agree that size-based might not be the best way to go about this. How about I add an `override_all_ids: Optional[bool]` arg? If specified, the `repetition_ids` size will be validated based on that. If unspecified, we choose the default value of `override_all_ids` based on `repetition_ids` and `self.repetition_ids`.

cirq/circuits/circuit_operation.py
smitsanghavi: That was what I had originally implemented but because of multiple optionals, the default-ing logic was becoming too complicated and I felt it was making the API messier. But if we are making the override-all-IDs case more explicit, I should be able to add this in as well.

cirq/circuits/circuit_operation.py
```diff
@@ -93,6 +124,9 @@ def qubits(self) -> Tuple['cirq.Qid', ...]:
         ordered_qubits = ops.QubitOrder.DEFAULT.order_for(self.circuit.all_qubits())
         return tuple(self.qubit_map.get(q, q) for q in ordered_qubits)
 
+    def _default_repetition_ids_(self) -> List[str]:
```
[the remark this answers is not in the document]
smitsanghavi: Done

cirq/circuits/circuit_operation.py
```diff
@@ -117,8 +151,40 @@ def _decompose_(self) -> 'cirq.OP_TREE':
             result = result ** -1
         result = protocols.with_measurement_key_mapping(result, self.measurement_key_map)
         result = protocols.resolve_parameters(result, self.param_resolver, recursive=False)
-
-        return list(result.all_operations()) * abs(self.repetitions)
+        # repetition_ids don't need to be taken into account if the circuit has no measurements
+        # or if it has repetition of size 1 with default repetition_ids.
+        if not protocols.is_measurement(result) or self.repetition_ids == default_repetition_ids(1):
```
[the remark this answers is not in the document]
smitsanghavi: Done

cirq/circuits/circuit_operation_test.py
smitsanghavi: Right. Was trying to contrast with the non-measurement test above but it's not really needed.

cirq/circuits/circuit_operation_test.py
smitsanghavi: 😄

cirq/protocols/json_test_data/CircuitOperation.json
smitsanghavi: Done, assigned repetition_ids here itself.

cirq/circuits/circuit_operation.py
smitsanghavi: Made the above modifications. Let me know if this looks better.

cirq/circuits/circuit_operation.py
smitsanghavi: Made `repetitions` `Optional` as well. The actual conditions of what defaults to what are a little difficult to describe but added lot of docstrings to compensate. Also, even though it's verbose to explain, the actual behavior seems pretty intuitive to me.

cirq/circuits/circuit_operation.py
95-martin-orion: In my mind, `op.repeat(2).repeat(3) != op.repeat(6)` is the correct behavior, since they represent different things. Conceptually, it's similar to how `[[0, 1], [2, 3]] != [0, 1, 2, 3]`.

Regarding `circuit_op.repeat(2).repetition_ids`: if the original `circuit_op` has an explicitly-defined ID of `0` for its single repetition, I think `['0-0', '1-0']` is correct. However, if no ID has been defined, we want `['0', '1']` instead. This behavior is similar to a raw value vs. a list with one element:
```
a = 0
a2 = [a, a]  # [0, 0]
a2[0] == a2[1] == 0

b = [0]
b2 = [b, b]  # [[0], [0]]
b2[0][0] == b2[1][0] == 0
```
With this in mind, our only edge case is the single-repetition, no-ID CircuitOperation; everything else follows the "prefix" case. By extension, we can drop `override_all_ids` and the extra complexity associated with it.

smitsanghavi: Sure, I was more going for the `ControlledGate` type behavior we have. Chained controls with different number of qubits are automatically "flattened" and the final `ControlledGate` is equivalent to if all of them were applied at once.
 
But I guess `pow` semantics (repeat) don't need to be similar to that `addition` type operation.

Made the above change and simplified a couple of other things.

cirq/circuits/circuit_operation.py
smitsanghavi: Now it's even more intuitive :)

cirq/circuits/circuit_operation.py
```diff
@@ -152,36 +160,35 @@ def _decompose_(self) -> 'cirq.OP_TREE':
...
-                    if isinstance(op, CircuitOperation):
-                        # For a measurement CircuitOperation, prefix the current repetition_id to
-                        # the children repetition_ids.
-                        ops.append(
-                            op.with_repetition_ids(
-                                cartesian_product_of_string_lists([parent_id], op.repetition_ids)
-                            )
+                if isinstance(op, CircuitOperation):
+                    # For a measurement CircuitOperation, prefix the current repetition_id to
+                    # the children repetition_ids.
```
95-martin-orion, line 173: Update comment?

cirq/circuits/circuit_operation.py
```diff
@@ -188,53 +269,95 @@ def _json_dict_(self):
...
-            raise NotImplementedError('Loops over measurements are not supported.')
-        return self.replace(repetitions=self.repetitions * repetitions)
+
+        expected_repetition_id_length = abs(repetitions)
+        # The eventual number of repetitions of the returned CircuitOperation.
+        final_repetitions = self.repetitions * repetitions
+
+        if repetition_ids is None:
+            repetition_ids = default_repetition_ids(repetitions)
+        elif len(repetition_ids) != abs(expected_repetition_id_length):
```
95-martin-orion, line 344: `abs` not needed here? (already called when initializing `expected_repetition_id_length`)

cirq/circuits/circuit_operation_test.py
```diff
@@ -167,31 +196,70 @@ def test_with_params():
...
 
 
-def test_repetition():
-    a, b = cirq.LineQubit.range(2)
-    # This circuit has a modulus of 8.
-    circuit = cirq.FrozenCircuit(cirq.H(a), cirq.CX(a, b))
-    op_base = cirq.CircuitOperation(circuit)
-    assert op_base.repeat(1) is op_base
+def cartesian_product_of_string_lists(list1: List[str], list2: List[str]):
+    return [f'{first}-{second}' for first in list1 for second in list2]
```
95-martin-orion, line 200: Are you not able to import this from the file directly? i.e.
`from cirq.circuits.circuit_operation import cartesian_product_of_string_lists`

If not, this should be updated to match that one.

cirq/circuits/circuit_operation_test.py
```diff
@@ -167,31 +196,70 @@ def test_with_params():
...
+    assert op_base.repeat(1, ['0']) == op_base.repeat(repetition_ids=['0'])
+    assert op_base.repeat(1, ['0']) == op_base.with_repetition_ids(['0'])
+
+    initial_repetitions = -3
+    if add_measurements:
+        with pytest.raises(ValueError, match='circuit is not invertible'):
+            _ = op_base.repeat(initial_repetitions)
+        initial_repetitions = abs(initial_repetitions)
+
+    final_repetitions = 2 * initial_repetitions
```
95-martin-orion, line 222: This can be moved closer to where it's used, for readability.

95-martin-orion (changes requested):
A few relatively minor items and this should be good to go.

cirq/circuits/circuit_operation_test.py
smitsanghavi: Done. Thanks for catching this!

cirq/circuits/circuit_operation.py
smitsanghavi: Done

cirq/circuits/circuit_operation.py
smitsanghavi: Done

cirq/circuits/circuit_operation_test.py
smitsanghavi: Done

95-martin-orion (approved):

PR #3743, opened 2021-02-03
Add ISWAP and its inverse to FSIM_GATESET

dstrain115:
- Adding ISWAP to FSIM_GATESET as an allowed gate
for experimental usage.

This allows serialization of ISWAP and ISWAP ** -1 using
this gateset.  Note that this gate may not be available on
all processors and may not have good/any performance.


mpharrigan (approved):
looks pretty simple to me

balopat (approved):

PR #3745, opened 2021-02-03
Fix density matrix simulator when subcircuits exist

### LINKED ISSUE #3723: DensityMatrixSimulator not working with subcircuits
**Description of the issue**

It appears density matrix simulator does not work with subcircuits? After looking through the noise simulation code I was trying to check whether noise propagated through subcircuits but I'm unable to even get a measurement. Entirely possible I'm using this incorrectly, or that this is expected behavior.

Note: if changed to the sparse simulator, the test passes.

```python
    def test_simulate_noise_with_subcircuits():
        q = cirq.LineQubit(0)
        circuit1 = cirq.Circuit(cirq.measure(q))
        circuit2 = cirq.Circuit(cirq.CircuitOperation(cirq.Circuit(cirq.measure(q)).freeze()))

        simulator = cirq.DensityMatrixSimulator(noise=cirq.X)
        result1 = simulator.run(circuit1, repetitions=10)
        result2 = simulator.run(circuit2, repetitions=10)

>       assert result1 == result2
E       AssertionError: assert cirq.Result(params=cirq.ParamResolver({}), measurements={'0': np.array([[0], [0], [0], [0], [0], [0], [0], [0], [0], [0]], dtype=np.int8)}) == cirq.Result(params=cirq.ParamResolver({}), measurements={})
E         +cirq.Result(params=cirq.ParamResolver({}), measurements={'0': np.array([[0], [0], [0], [0], [0], [0], [0], [0], [0], [0]], dtype=np.int8)})
E         -cirq.Result(params=cirq.ParamResolver({}), measurements={})

```

**Cirq version** 0.10.0.dev (latest master)

daxfohl:
Fixes #3723 by forcing `run_sweep_repeat` in density matrix simulator if subcircuits are detected in the circuit. `run_sweep_sample` does not work in this case because there's no way to sample a subcircuit.


95-martin-orion (approved):
This is reasonable as a workaround for #3723, but `run_sweep_repeat` takes O(repetitions) times longer to complete than `run_sweep_sample`. In the long term, we'll want to switch back to "sample" where possible. The same applies to `sparse_simulator`, which will mark an entire subcircuit as non-unitary even if it only has terminal measurements.

One cheap-ish way to do this would be to split the subcircuit and its measurements:
```
# pseudocode
Circuit(CircuitOp(op1, op2, op3, measure))
--> Circuit(CircuitOp(op1, op2, op3), measure)
```
That step needs to be recursive, but it would be much cheaper to implement "subcircuit splitting" than full subcircuit sampling.

PR #3748, opened 2021-02-04
Forrelation

### LINKED ISSUE #3687: Add Fourier checking as a tutorial

As described in https://arxiv.org/abs/0910.4698 Fourier Checking is a simple quantum algorithm that can decide whether two 2^n random bitstrings are "forrelated" or not. I would like to showcase with this problem oracle query complexity and oracle synthesis using diagonal gates. 

We need the following: 
- a docs/tutorials/fourier_checking.ipynb that explains 
   - Fourier Checking and why it's interesting
   - create classically two distributions based on n qubits that are within the promised boundaries to be forrelated 
   - explains query complexity
~- why simple oracles based on truth tables are just simple diagonal gates - and how you can decompose those into 1 and 2 qubit gates in Cirq~
~- we will need an extension of the ThreeQubitDiagonalGate class to multiple qubits. An initial starting point around this would be this paper: https://iopscience.iop.org/article/10.1088/1367-2630/16/3/033040/meta - but if a contributor finds a more efficient algorithm, that would be even better.~

**Additional context (e.g. screenshots)**

I did some hacking on this colab, that could be used as a starting point: https://colab.research.google.com/drive/1iT-PO6GSstreo5edvxdvVQ2Xj4fxHRXo?usp=sharing. 

ybc1991:
Adds notebook tutorial for Fourier checking problem

Fixes #3687.


docs/tutorials/fourier_checking.ipynb
```diff
@@ -0,0 +1,1187 @@
...
+   "metadata": {
+    "id": "ucEcJmSQ26vg"
+   },
+   "outputs": [],
+   "source": [
+    "def bitwise_dot(x, y):\n",
+    "    \"\"\"Compute the dot product of two integers bitwise.\"\"\"\n",
+    "\n",
+    "    # fast bit parity\n",
+    "    def parallel_swar(i):\n",
```
balopat, line 432: This is from the blogpost http://p-nand-q.com/python/algorithms/math/bit-parity.html, forgot to mention that. Now that I'm thinking about it I'd prefer rewriting it to a maybe slower but our own method. If we end up sticking with this version, we'll need to provide attribution, run a round with legal to see how we should do that.

docs/tutorials/fourier_checking.ipynb
```diff
@@ -0,0 +1,1187 @@
...
+    "print(f\"f: {list(f)} f_hat: {list(f_hat)}\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "02GAaXsAsYOW"
+   },
+   "source": [
+    "## Forrelated\n",
```
balopat, line 505:
```suggestion
    "## Forrelation\n",
```

docs/tutorials/fourier_checking.ipynb
```diff
@@ -0,0 +1,1187 @@
...
+    "\n",
+    "The concept of *forrelated* is the key concepts in Fourier checking problem, which is a combination word coming from \"Fourier Transform\" and \"Correlation\".\n",
```
balopat, line 507:
```suggestion
    "The concept of *forrelation* is the key concept in the Fourier checking problem, which is a combination of the words \"Fourier Transform\" and \"Correlation\".\n",
```

docs/tutorials/fourier_checking.ipynb
```diff
@@ -0,0 +1,1187 @@
...
+    "    print(\"fs and gs is forrelated!\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "IHLLADQvT-lK"
+   },
+   "source": [
+    "Last, we can randomly draw the functions from either $\\mathcal{U}$ or $\\mathcal{F}$ set to evaluate the confusion matrix of the fourier checking algorithm. With the confusion matrix, you should be confident that quantum algorithm instead can solve the fourier checking in $O(1)$ time -- even though we need to measure 100 or 1000 times to evaluate the probability, it is irrelevant to the number of states or qubits."
```
balopat, line 1034:
```suggestion
    "Last, we can randomly draw the functions from either $\\mathcal{U}$ or $\\mathcal{F}$ set to evaluate the confusion matrix of the Fourier Checking algorithm. With the confusion matrix, you should be confident that quantum algorithm instead can solve the Fourier Checking in $O(1)$ time -- even though we need to measure 100 or 1000 times to evaluate the probability, it is irrelevant to the number of states or qubits."
```

balopat (changes requested):
This looks great, added my nitpicking. I looked through it offline before the PR, so I'd like someone from the @quantumlib/cirq-maintainers (@Strilanc?) to confirm theoretical correctness as well.

docs/tutorials/fourier_checking.ipynb
ybc1991: Agree. Change it into ```bin(i).count("1")```, which is much easier to read.

docs/tutorials/fourier_checking.ipynb
```diff
@@ -0,0 +1,1187 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "view-in-github"
+   },
+   "source": [
+    "<a href=\"https://colab.research.google.com/github/ybc1991/Cirq/blob/forrelation/docs/tutorials/fourier_checking.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
```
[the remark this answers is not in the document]
ybc1991: Done. Thanks for the notebook guide, which is really helpful! 

Just one comment: what I felt is not clear is what to determine the notebook has external dependency or not. 
For example, depending on networkx and pandas does not count as having external dependency because it is in the request in cirq. Can I understand this external dependency means any packages excluding cirq + its dependency?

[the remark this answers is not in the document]
balopat: Yeah, I should use a different word there. I mean external API calls. Which this notebook does not have, so we should be able to test it. Added https://github.com/quantumlib/Cirq/issues/3803 to clarify this in the guide

docs/tutorials/fourier_checking.ipynb
balopat: Great, I would still call it bit parity instead of count_1s_in_z2...

ybc1991: Done

balopat (approved):

PR #3750, opened 2021-02-04
Add support for WaitGate and inverse of sqrt(iSWAP) to characterization and compilation

mrwojtek:
This adds support for cirq.WaitGate, which is just ignored.

The support for iSWAP ** -0.5 is achieved by introducing a phase exponent that is a rotation in chi angle of PhasedFSimGate which coincides with sign change in iSWAP ** 0.5 for rotation -2 i p, where p = 0.5 is the phase exponent.


cirq/google/calibration/phased_fsim.py
```diff
@@ -434,15 +434,37 @@ class IncompatibleMomentError(Exception):
     """Error that occurs when a moment is not supported by a calibration routine."""
 
 
-def try_convert_sqrt_iswap_to_fsim(gate: Gate) -> Optional[FSimGate]:
+@dataclasses.dataclass(frozen=True)
+class FSimGateCalibration:
```
dstrain115, line 438: I don't think this is a good name.  I would use AdjustedFSimGate or CalibratedFSimGate.  We should also consider making this implement Gate, since it is basically a gate anyway.

If we make it a gate, it should probably go in cirq.google.ops.

cirq/google/calibration/phased_fsim.py
```diff
@@ -434,15 +434,37 @@ class IncompatibleMomentError(Exception):
...
 
     Args:
         gate: Gate to verify.
 
     Returns:
-        FSimGate(theta=π/4, phi=0) if provided gate either  FSimGate, ISWapPowGate, PhasedFSimGate
-        or PhasedISwapPowGate that is equivalent to FSimGate(theta=π/4, phi=0). None otherwise.
+        FSimGateCalibration with engine_gate FSimGate(theta=π/4, phi=0) if the provided gate is
+        either FSimGate, ISWapPowGate, PhasedFSimGate or PhasedISwapPowGate that is equivalent to
+        FSimGate(theta=±π/4, phi=0). None otherwise.
```
dstrain115, line 467: This function does not respect periodicity.

cirq/google/calibration/phased_fsim.py
mrwojtek: This is definitely not a gate. This is association between a gate and hardware gate to characterize.

It reflects the state in two realms: the `engine_gate` is a gate which is supposed to be characterized. The pair (`engine_gate`, `phase_exponent`) is a gate that is supposed to be used during the compilation after merging with characterization obtained for engine_gate. I do think it belongs to this package rather than `cirq.ops`. Maybe `EngineFSimGate` would be a better name though?

cirq/google/calibration/phased_fsim.py
mrwojtek: Added for theta.

cirq/google/calibration/engine_simulator.py
```diff
@@ -364,25 +367,35 @@ def get_calibrations(
...
             parameters = self._drifted_parameters[(b, a, gate)].parameters_for_qubits_swapped()
         else:
             parameters = self._drift_generator(a, b, gate)
             self._drifted_parameters[(a, b, gate)] = parameters
-        return PhasedFSimGate(**parameters.asdict())
+
+        return PhasedFSimGate(
+            theta=parameters.theta,
+            zeta=parameters.zeta,
+            chi=parameters.chi - 2 * np.pi * gate_calibration.phase_exponent,
```
mpharrigan, line 395: why do you need `phase_exponent` to tag alongside `gate`? Why not use `gate.exponent`?

cirq/google/calibration/phased_fsim.py
mpharrigan: Why can't the phase_exponent just be `engine_gate.exponent`? (or `.theta` I guess)

mpharrigan (commented):
I'm unconvinced that we need a `Tuple[Gate, float]` instead of just setting an appropriate attribute on the gate

cirq/google/calibration/phased_fsim.py
mrwojtek: Because engine_gate wouldn't be serializable to FSIM/SQRT_ISWAP gateset? This is not theta, this is shift of -2 * pi * phase_exponent  in chi of PhasedFSimGate.  Maybe we should discuss it in person next week...

mrwojtek:
I thought about this a little bit more and did some refactoring in the end. I moved all the phase corrections logic to this new class, in that way there is a very clear contract on what it does and how it is supposed to be used. The phase_exponent could be internal right now (it's not because it's not easy to do that with a frozen dataclass).

This PR  should be more clear now, PTAL. We can have a VC to discuss it if you like.

cirq/google/calibration/phased_fsim.py
mpharrigan: Will we need one of these classes for each type of correction we want to implement?

mrwojtek:
> I'm unconvinced that we need a `Tuple[Gate, float]` instead of just setting an appropriate attribute on the gate

The Gate in the tuple is fixed, you can't modify/change it at all - this must be gate as is exactly passed to circuit serialization and reflects the hardware gate which needs to run for calibration.

On the other hand due to the phase rotations the set of gates which can easily be decomposed to to this particular gate is larger. This is what I meant to represent here.

I don't know how it should be done in the future and what's the best way to tackle this. This require a holistic look at the gate sets/compilers in Cirq which I didn't think about. What I'm doing here is the easiest way to support the crucial functionality which is missing from the current compiler: inverse of sqrt(iSWAP) gate which is commonly used.

cirq/google/calibration/phased_fsim.py
mrwojtek: My hope is that this will be solved in a better way until this becomes necessary.

mpharrigan (approved):
We talked about this briefly offline today (offline in a very abstract sense; it was obviously over a video call). The missing magic from Cirq is the notion of a "hardware gate" which has a definite pulse sequence but not a definite unitary. To do this the "right" way would require careful design of this abstraction in Cirq proper. 

I think the re-work (putting methods on the dataclass) has actually clarified things a lot. I don't think `FSimGateCalibration` is the best name, but I'll give approval anyways. Maybe `CalibratedFSimGate`? `PhaseCalibratedFSimGate`? `FSimWithPhase`?

cirq/google/calibration/phased_fsim.py
```diff
@@ -434,15 +443,92 @@ class IncompatibleMomentError(Exception):
     """Error that occurs when a moment is not supported by a calibration routine."""
 
 
-def try_convert_sqrt_iswap_to_fsim(gate: Gate) -> Optional[FSimGate]:
+@dataclasses.dataclass(frozen=True)
+class FSimGateCalibration:
+    """Association of a user gate with gate to calibrate.
+
+    This association stores an information regarding rotation of the calibrated FSim gate by
```
mpharrigan, line 450:
```suggestion
    This association stores information regarding rotation of the calibrated FSim gate by
```

mrwojtek:
I like `PhaseCalibratedFSimGate` and I used this. The refactoring allowed me to write better tests for this functionality which actually revealed a bug! The bug was in the sign of phase exponent in `with_zeta_chi_gamma_compensated` method: the method is cancelling out errors which means I should actually subtract the phase so it will be added back by cancellation. (This bug doesn't matter for any functionality until now because iSWAP ** -0.5 and iSWAP ** 0.5 are phased by 2pi which doesn't matter if you add or subtract; it can matter in the future where intermediate phases would be used). 

I'll merge this PR once @mpharrigan solves the issues with Google CLA.

PR #3752, opened 2021-02-05
Rename calibration methods

mrwojtek:
After going through the reviews I propose those changes in names, they are more accurate in terms of what they do and they'll also not conflict with the future changes that are planned.


dstrain115 (approved):
Names are fine with me.  Will let other chime in as well.

balopat (approved):
LGTM!

cirq/google/__init__.py
```diff
@@ -32,12 +32,12 @@
     PhasedFSimEngineSimulator,
     SQRT_ISWAP_PARAMETERS,
     THETA_ZETA_GAMMA_FLOQUET_PHASED_FSIM_CHARACTERIZATION,
-    make_floquet_request_for_circuit,
-    make_floquet_request_for_moment,
-    zeta_chi_gamma_calibration_for_moments,
-    run_characterizations,
+    prepare_floquet_characterization_for_circuit,
+    prepare_floquet_characterization_for_moment,
+    zeta_chi_gamma_compensation_for_moments,
```
mpharrigan, line 37: everything else has a verb

mpharrigan (approved):

cirq/google/__init__.py
mrwojtek: That is true, I think make is the most appropriate (add could also work but only for this particular method, in future there might be more substantial circuit modifications).

Issue #3681, opened 2021-01-15
Need more visualizations for qubit picking from calibration metrics


karlunho:
When using the quantum computer, I would like to pick the best qubits and the best 2-qubit gates for my experiment based on the calibration metrics.

Currently, the calibration.heatmap('two_qubit_parallel_sqrt_iswap_gate_xeb_average_error_per_cycle') allows me to see the 1 qubit metrics, but doesn't allow me to see the 2-qubit metrics

**Describe the solution you'd like**

In an ideal world, there will be a heatmap showing the 2-qubit fidelities of all the two-qubit metrics ([link](https://quantumai.google/cirq/google/calibration)).  In addition to seeing all the two-qubit grids, it would be nice to have CDFs plots as well (similar to Google's internal reports). 

I'd also like to see a grid showing which qubits have high coherent error, which is equal to (two_qubit_parallel_sqrt_iswap_gate_xeb_average_error_per_cycle -  two_qubit_parallel_sqrt_iswap_gate_xeb_incoherent_error_per_cycle), in order to avoid qubits with high coherent error for my experiment.

Lastly, a tutorial would be helpful for visualization

**[optional] Describe alternatives/workarounds you've considered**

**[optional] Additional context (e.g. screenshots)**

The following [bug](https://github.com/quantumlib/Cirq/issues/3088) would help quite a bit. 

**What is the urgency from your perspective for this issue? Is it blocking important work?**

P1 - I need this no later than the next release (end of quarter)

tanujkhattar:
xref #3642

PR #3756, opened 2021-02-06
Sanitize value / annotation keys in heatmap

balopat:
Sanitize the keys for value, annotation and url maps in heatmap.

Part of #3681.

### DIFF
--- a/cirq/vis/heatmap.py
+++ b/cirq/vis/heatmap.py
@@ -72,15 +72,6 @@ class Heatmap:
             f"Unrecognized key type, expected cirq.GridQubit or Tuple[int,int], " f"got {repr(key)}"
         )
 
-    def _sanitize_annotation_key(self, key: QubitCoordinate) -> Tuple[int, int]:
-        if isinstance(key, grid_qubit.GridQubit):
-            return key.row, key.col
-        if isinstance(key, tuple) and tuple(map(type, key)) == (int, int):
-            return key
-        raise ValueError(
-            f"Unrecognized key type, expected cirq.GridQubit or Tuple[int,int], " f"got {repr(key)}"
-        )
-
     def set_value_map(self, value_map: ValueMap) -> 'Heatmap':
         """Sets the values for each qubit.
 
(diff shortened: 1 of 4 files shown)


cirq/vis/heatmap.py
```diff
@@ -63,8 +56,7 @@ class Heatmap:
     def __init__(self, value_map: ValueMap) -> None:
         self.set_value_map(value_map)
         self.annot_map = {  # Default annotation.
-            _get_qubit_row_col(qubit): format(float(value), '.2g')
-            for qubit, value in value_map.items()
+            (*qubit,): format(float(value), '.2g') for qubit, value in value_map.items()
```
maffoo, line 59: I think `tuple(qubit)` is more clear. Though a method to explicitly convert to tuple would be better since it could be typed as `Tuple[int, int]` whereas this would give `Tuple[int, ...]` with an unknown number of args.

cirq/vis/heatmap.py
```diff
@@ -204,7 +195,7 @@ def plot(
             np.nan, index=range(min_row, max_row + 1), columns=range(min_col, max_col + 1)
         )
         for qubit, (float_value, _) in self.value_map.items():
-            row, col = _get_qubit_row_col(qubit)
+            row, col = (*qubit,)
```
maffoo, line 198: can just do `row, col = qubit`.

maffoo (commented):
What if we were to add a method instead (or in addition), e.g. `astuple`? This could have return type `Tuple[int, int]` which means mypy can check that you are unpacking as the correct number of items, or converting to a tuple with the correct number of items.

95-martin-orion:
> What if we were to add a method instead (or in addition), e.g. `astuple`?

+1 to this suggestion. I don't have any concrete examples, but it feels like defining `__iter__` for GridQubits would lead to weird, hard-to-debug issues in the future.

balopat:
Thanks for your quick inputs! I'll regroup and will try to solve the type ambiguity in heatmap's valuemap keys differently.

balopat:
I'm going to close this as @tanujkhattar will rework the Heatmap design anyway.

PR #3760, opened 2021-02-06
[XEB] Support parallel execution

mpharrigan:
Support parallel execution of XEB circuits. 

 - Instead of generating a set of circuits for each pair, we generate one library of two-qubit circuits which are mixed-and-matched during execution. This speeds up circuit generation, circuit serialization (for historical record), and speeds up simulation (when not optimizing ... when you're optimizing each pair presumably gets its own set of angles and needs to be simulated separately)
 - As such, a new function is added to the random_circuit_generation module to create a combinations object that describes how to mix-and-match these two-qubit circuits
 - The sample function is modified to accept this combination description to on-the-fly zip the circuits, sample from them, and split out the results. This allows maximal code re-use between isolated and parallel XEB! Analysis is mostly shared! Huge caveat below
 - If you run the `characterize_phased_fsim_parameters_with_xeb` on the code in this PR, it will find only one set of angles that works for the average of all pairs. I've split up the modifications to that half of the code-base into a subsequent PR.

This can be reviewed independently of #3753


karlunho:
The tutorials look good, and I suggest that we create entries into the book.yaml so they show up on the documentation site.

cirq/experiments/fidelity_estimation.py
```diff
@@ -644,10 +773,29 @@ def _summary_stats(row):
...
+            """If all the values for a key `k` are the same in this group, we can keep it."""
+            if k not in df.columns:
+                return  # coverage: ignore
+            vals = df[k].unique()
+            if len(vals) == 1:
+                ret[k] = vals[0]
+
+        _try_keep('q0')
+        _try_keep('q1')
+        _try_keep('pair_name')
```
mrwojtek, line 790: Shouldn't that be rather a failure? (When they're not the same)

cirq/experiments/fidelity_estimation.py
```diff
@@ -496,32 +556,97 @@ def sample_2q_xeb_circuits(
...
+            the circuits in `circuits` into wide, parallel-XEB-style circuits for execution.
 
     Returns:
-        A pandas dataframe with index given by ['circuit_i', 'cycle_depth'] and
-        column "sampled_probs".
+        A pandas dataframe with index given by ['circuit_i', 'cycle_depth'].
+        Columns always include "sampled_probs". If `combinations_by_layer` is
+        not `None` and you are doing parallel XEB, additional metadata columns
+        will be attached to the returned DataFrame.
     """
```
mrwojtek, line 570: This method became pretty long, it would be good to refactor it and split to smaller sections

cirq/experiments/fidelity_estimation.py
```diff
@@ -496,32 +556,97 @@ def sample_2q_xeb_circuits(
...
     tasks = []
     for cycle_depth in cycle_depths:
-        for circuit_i, circuit in enumerate(circuits):
+        for zipped_circuit in zipped_circuits:
             circuit_depth = cycle_depth * 2 + 1
-            assert circuit_depth <= len(circuit)
-            truncated_circuit = circuit[:circuit_depth]
-            prepared_circuit = truncated_circuit + ops.measure(q0, q1)
+            assert circuit_depth <= len(zipped_circuit.wide_circuit)
+            # Slicing creates a copy, although this isn't documented
```
mrwojtek, line 629: Is this something you want to rely for future compatiblity?

cirq/experiments/fidelity_estimation.py
```diff
@@ -470,17 +505,42 @@ def __enter__(
...
 
+@dataclass(frozen=True)
+class _ZippedCircuit:
+    """A fully-wide circuit made by zipping together a bunch of two-qubit circuits
+    and its provenance data.
+
+    Args:
+        wide_circuit: The zipped circuit on all pairs
+        pairs: The pairs of qubits operated on in the wide circuit.
+        layer_i: The index of the GridInteractionLayer to which the pairs correspond.
```
mrwojtek, line 520: I'm having trouble understanding why this code is so tightly coupled with those interaction layers? The layers are static but in the end this code will need to run with completely dynamic, moment-dependent configurations. What is relation between those two?

cirq/experiments/fidelity_estimation.py
```diff
@@ -470,17 +505,42 @@ def __enter__(
     def __exit__(self, exc_type, exc_val, exc_tb):
         pass
 
-    def update(self, increment: int):
+    def update(self, n: int = 1):
```
mrwojtek, line 508: Why obscure this?

cirq/experiments/fidelity_estimation_test.py
```diff
@@ -306,6 +307,95 @@ def test_sample_2q_xeb_circuits_no_progress(capsys):
...
+    with pytest.raises(ValueError, match='.*invalid indices.*'):
+        _ = sample_2q_xeb_circuits(
+            sampler=cirq.Simulator(),
+            circuits=circuits,
+            cycle_depths=cycle_depths,
+            combinations_by_layer=combs,
+        )
+
+
+def test_sample_2q_parallel_xeb_circuits_error_2():
```
mrwojtek, line 373: please be more specific what kind of errors is that (as well in the above)

cirq/experiments/random_quantum_circuit_generation.py
```diff
@@ -217,6 +236,114 @@ def random_rotations_between_two_qubit_circuit(
...
+        n_library_circuits: The number of circuits to generate.
+        two_qubit_gate: The two qubit gate to use in the circuits.
+        max_cycle_depth: The maximum cycle_depth in the circuits to generate. If you are using XEB,
+            this must be greater than or equal to the maximum value in `cycle_depths`.
+        q0: The first qubit to use when constructing the circuits.
+        q1: The second qubit to use when constructing the circuits
+        random_state: A random state or seed used to deterministically sample the random circuits.
+    """
+    rs = value.parse_random_state(random_state)
+    exponents = np.linspace(0, 7 / 4, 8)
```
mrwojtek, line 260: I would make this an optional argument with defaults set to your value. Is that corect? Or rather the entire single_qubit_gates array, and just fix the defaults to constants?

cirq/experiments/fidelity_estimation.py
```diff
@@ -470,17 +505,42 @@ def __enter__(
...
+        combination_i: The row index of the combinations matrix that identifies this
+            particular combination of component narrow circuits.
+        combination: The row of the combinations matrix. Each entry is an index
+            into the (narrow) `circuits` library. Each entry indexes the
+            narrow circuit operating on the corresponding pair in `pairs`.
+    """
+
+    wide_circuit: 'cirq.Circuit'
+    pairs: List[Tuple['cirq.Qid', 'cirq.Qid']]
+    layer_i: int
```
mrwojtek, line 530: I guess it's fine that layer itself might be None?

cirq/experiments/random_quantum_circuit_generation.py
```diff
@@ -217,6 +236,114 @@ def random_rotations_between_two_qubit_circuit(
...
+    """Extract pairs of qubits from a device graph and a GridInteractionLayer."""
+    for edge in graph.edges:
+        if edge in grid_layer:
+            yield edge
+
+
+@dataclasses.dataclass(frozen=True)
+class CircuitLibraryCombination:
+    """For a given layer (which defines a set of pairs of qubits), `combinations` is a 2d array
+    of shape (n_combinations, len(pairs)) where each row represents a combination (with replacement)
```
mrwojtek, line 288: n_combinations is number of "experiments" which are used for fidelity estimation? The more of them the better accuracy?

docs/characterization/parallel_xeb.ipynb
```diff
@@ -0,0 +1,267 @@
+{
```
mrwojtek, line 1: <p>typo: deign</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/quantumlib/Cirq/pull/3760/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

mrwojtek, line 1: <p>If you name the variable circuits_library, this notebook will be much more clear.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/quantumlib/Cirq/pull/3760/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

mrwojtek, line 1: <p>It would be good reference circuits/circuits_library directly here</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/quantumlib/Cirq/pull/3760/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

mrwojtek, line 1: <p>This is beyond this PR but if instead of simulating circuits inside of this method, you would move it to the very beginning where circuits are actually generated then in the future you could serialize circuits and simulation results together, potentially making the fidelity estimation very fast since simulation results can be done ahead of execution on the hardware.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/quantumlib/Cirq/pull/3760/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

cirq/experiments/fidelity_estimation.py
mpharrigan: this matches the tqdm signature if you wanted to use keyword arguments

cirq/experiments/fidelity_estimation.py
mpharrigan: This might become clear in a following PR, but let me try to explain and then I'll think about how to make it clearer in code.

The primary idea is that you have a number of different moments that define the spacial layout of simultaneous two-qubit gates. The essential essence of each of those moments/layers is just the list of pairs, here as `pairs`. Indeed, `pairs` (and the circuit) is really the only essential part of this dataclass. The other attributes are metadata to maintain the provenance of the particular set of pairs. 

Previous XEB implementations relied heavily on these `GridInteractionLayer` objects to semi-implicitly define the set of pairs by making several assumptions (like you're operating on a nice grid, all the couplings on the grid work, you'll be using the whole device and/or the calling function will downsample only the pairs that actually exist). So in traditional XEB there are always four layers (one for each of the N, S, E, W set of pairs) and they're represented by these objects.

Of course, for the calibration API we re-purpose `cirq.Moment` as a "layer". What you'll see in a following PR is that these fields will either contain a `GridInteractionLayer`, a `Moment`, or even `None` since it's just metadata that doesn't really matter as long as you fill in `pairs`. 

`layer_i` and `combination_i` are just propagated through to be part of the result object. Let me see if I can clarify that

cirq/experiments/fidelity_estimation.py
mpharrigan: yes, see longer description above. The `layer` might be `None`. `layer_i` is just a contiguous 0-based index into which spacial configuration is being characterized. Imagine using the calibration API where this is the moment index of the circuit, where each moment typifies a configuration of 2q gates.

cirq/experiments/fidelity_estimation.py
```diff
@@ -470,17 +505,42 @@ def __enter__(
...
+    combination_i: int
```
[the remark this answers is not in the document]
mpharrigan: Yes, for each layer we run many (~10) combinations [i.e. different random circuits on a given pair]

cirq/experiments/fidelity_estimation.py
mpharrigan: I'm changing the docstrings to be more clear about what `layer_i` and `combination_i` are. I've added a note to both that they don't modify any behavior, but are just propagated to the output. I'm also re-ordering the fields in this dataclass to put them last, because they are less important. I've added a note to `combination` that it's important to be able to "unzip" the circuit.

cirq/experiments/fidelity_estimation.py
mpharrigan: yes, it's huge. Not helped by the `black` formatter putting everything on its own line! 

Factored out
 - `_get_combinations_by_layer_for_isolated_xeb`
 - `_zip_circuits`
 - `_generate_sample_2q_xeb_tasks`
 - `_execute_sample_2q_xeb_tasks_in_batches`

cirq/experiments/fidelity_estimation.py
mpharrigan: yes, although I would hope if anyone ever tries to change this in cirq it would cause a lot of tests to fail, be recognized as a breaking change, and be rejected. What do you think?

cirq/experiments/fidelity_estimation.py
mpharrigan: yes, I think you're right. Maybe this leniency was a holdover from previous design? changing to error. edit: ah yes, I was going to just run it on every column instead of hand-writing the ones to keep. But changed my mind

cirq/experiments/fidelity_estimation_test.py
mpharrigan: changed both

cirq/experiments/random_quantum_circuit_generation.py
mpharrigan: The fully-generic version would be to write your own list comprehension around `random_rotations_between_two_qubit_circuit`. This is a slightly-more abstracted entry point where these things are nailed down:

 - the single qubit rotations
 - the fact that the two qubit gate is constant
 - you want a list of `n_library_circuits` of them

and these things given more opinionated defaults:
 - depth
 - qubit indices (for parallel xeb, they will get re-mapped anyways)

If any of these don't apply it's very straightforward to drop down to `random_rotations_between_two_qubit_circuit`

cirq/experiments/random_quantum_circuit_generation.py
mpharrigan: technically more "precise". But yes, it's the number of random circuits. I've added a note to `get_random_combinations_for_device` that increasing `n_combinations` increases precision of the estimate but increases sampling runtime

mpharrigan (commented):
address some of the comments

docs/characterization/parallel_xeb.ipynb
mpharrigan: <p>this is a word! I'll change it to something less archaic </p>

mpharrigan: <p>changed</p>

mpharrigan: <p>good idea, added!</p>

mpharrigan: <p>The simulation is rather fast (&lt; 1 second). The fact that we use one circuit library instead of a new random circuit for each pair makes this scale much better for large devices.</p><p> </p><p>For benchmarking, this sort of thing could be worth it (where you pre-generate one library of circuits and you pre-simulate all of them). But for characterization, this simulation needs to be done in a hot loop (i.e. once per optimization step with the new parameters) so you can't amortize the simulation cost. I do want to see how hard it would be to throw these two-qubit circuits into a symbolic math library and get the final probabilities in terms of the parameters. The other option is to slice into the qsim API which should be very fast for these two qubit circuits. </p>

cirq/experiments/fidelity_estimation.py
balopat: One more reason to wrap tqdm api in our own api? We could then even swap it for alternative implementations....

mpharrigan: I'm not sure using `n` instead of `increment` adds all that much impetus to (re)wrap tqdm, a very popular progress bar library :)

As it exists, you can swap in alternative implementations. One such example is given here!

cirq/experiments/fidelity_estimation.py
mrwojtek: I guess that is what would happen, it's fine I think ;)

docs/characterization/parallel_xeb.ipynb
mrwojtek: Sorry, never heard! ;)

mrwojtek:
Matthew, are you still working on the changes or this is ready for next round of review?

mpharrigan:
Ready for review! I think I've addressed everything

cirq/experiments/fidelity_estimation.py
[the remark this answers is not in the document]
mrwojtek: I see, so combination_i doesn't really point to the index on combination field, the combination field is already selected from the larger structure.

cirq/experiments/fidelity_estimation.py
```diff
@@ -496,42 +674,51 @@ def sample_2q_xeb_circuits(
...
-            for future in concurrent.futures.as_completed(futures):
-                records += future.result()
-                progress.update(batch_size)
+    # Shim isolated-XEB as a special case of combination-style parallel XEB.
+    if combinations_by_layer is None:
+        combinations_by_layer, circuits = _get_combinations_by_layer_for_isolated_xeb(circuits)
+        one_pair = True
+    else:
+        _verify_two_line_qubits_from_circuits(circuits)
+        one_pair = False
```
mrwojtek, line 699: instead of one_pair you can create drop_columns = [...]

cirq/experiments/fidelity_estimation.py
```diff
@@ -470,17 +505,160 @@ def __enter__(
...
+    """Helper function used in `sample_2q_xeb_circuits` to prepare circuits in sampling tasks."""
+    tasks: List[_Sample2qXEBTask] = []
+    for cycle_depth in cycle_depths:
+        for zipped_circuit in zipped_circuits:
+            circuit_depth = cycle_depth * 2 + 1
+            assert circuit_depth <= len(zipped_circuit.wide_circuit)
+            # Slicing creates a copy, although this isn't documented
+            prepared_circuit = zipped_circuit.wide_circuit[:circuit_depth]
+            for pair_i, pair in enumerate(zipped_circuit.pairs):
+                prepared_circuit += ops.measure(*pair, key=str(pair_i))
```
mrwojtek, line 614: Have you checked if this aligns all the measurements in one moment?

cirq/experiments/fidelity_estimation.py
```diff
@@ -470,17 +505,160 @@ def __enter__(
...
+    zipped_circuits: List[_ZippedCircuit] = []
+    for layer_i, layer_combinations in enumerate(combinations_by_layer):
+        for combination_i, combination in enumerate(layer_combinations.combinations):
+            wide_circuit = Circuit.zip(
+                *(
+                    circuits[i].transform_qubits(lambda q: pair[q.x])
+                    for i, pair in zip(combination, layer_combinations.pairs)
+                )
+            )
+            zipped_circuits.append(
```
mrwojtek, line 590: This and _generate_sample_2q_xeb_tasks could potentialy be generators with `yield`

cirq/experiments/fidelity_estimation.py
mrwojtek: Hmm, so what's the final outcome on this?

cirq/experiments/random_quantum_circuit_generation.py
mrwojtek: In that case could you at least document  in docstring what kind of signle-qubit rotations are considered and used?

cirq/experiments/random_quantum_circuit_generation.py
mrwojtek: Sounds good, I just started to wonder if n_combinations is a good name. Combinations is not the main concern in the end (they could have been generated on the fly as well), it's number of circuit sampled what really matters.

cirq/experiments/fidelity_estimation.py
mpharrigan: I actually think it reads more intuitively later on when you might be wondering why we're dropping some of the columns: because it's one pair! Also, if there are any other behaviors that need to be slightly modified (other than column dropping) this name would still fit.

cirq/experiments/fidelity_estimation.py
mpharrigan: for the current code, it should just because these are dense circuits by construction; but I've changed it to add the measure operations in one moment. Good spot!

cirq/experiments/fidelity_estimation.py
mpharrigan: `zipped_circuits` (in `sample_2q_xeb_circuits`) needs to be a list as it's re-iterated-through for each cycle depth. `tasks` needs to be a list as we index into it in order to nest into batches. Therefore, if we change these helper functions to generators an extra `list(...)` would be needed at the call site, so I think it's a wash.

cirq/experiments/fidelity_estimation.py
mpharrigan: it will preserve the keys listed, but a) doesn't care if they're not present as they are metadata b) makes sure the grouped values in a given column are all compatible.

I think it's a good idea to keep this around if and when we add additional metadata

cirq/experiments/random_quantum_circuit_generation.py
mpharrigan: added. (also reached out to sergio to see how strong of a theoretical basis there is for this choice. I stole it from the CRNG work)

mpharrigan:
back at you @mrwojtek !

mrwojtek (approved):

balopat (approved):

Issue #3233, opened 2020-08-19
Expand measurement results


balopat:
**Problem**: Cirq’s implementation of measurements is a model in which binary outcomes come from measurements in the computational basis.  There are a variety of cases in which measurements are more general than this.  For experimental implementation measurement results are often not just binary, but can correspond to more general floating point data, such as IQ points in superconducting qubits.  Another example is that the outcomes may be more rich, such as ternary outcomes.  Further measurements in Cirq are simple key value maps, but if measurements are to be used in feedforward or data flow, then these may need to be stored in classical registers and used later in a circuit.  Further there is inconsistent use of these measurement results in the results produced by simulators, samplers, and run methods.  

**Depends on:** Classical data (#3231). 
**Rough requirements:**

- Allow more than just binary values to be used for measurements
- Be compatible with solutions to classical data / feedforward.

95-martin-orion:
Between #4274 being resolved and #3868 being closed, I think this has been appropriately addressed. The only remaining open item is #3808, which has been marked post-1.0.

PR #3785, opened 2021-02-11
Add method to get "mapped_circuit" from CircuitOperation

### LINKED ISSUE #3783: Flatten circuits containing CircuitOperation
I want to use `CircuitOperation` to carefully construct a circuit while respecting moment structure. At some point I need to flatten/unnest it to a normal two-dimensional circuit.

@Strilanc suggests

```python
def flatten(obj: Union[cirq.Moment, cirq.Circuit, cirq.FrozenCircuit]) -> cirq.Circuit:
    if isinstance(obj, cirq.Moment):
        flattenable = [op for op in obj if isinstance(op, cirq.CircuitOperation)]
        if not flattenable:
            return cirq.Circuit(obj)
        return cirq.Circuit.zip(
            cirq.Circuit(obj - flattenable),
            *[flatten(cast(cirq.CircuitOperation, e).circuit) for e in flattenable]
        )
    return cirq.Circuit(
        [moment for e in obj for moment in flatten(e)]
    )
```

but I'm running into problems with `circuit_operation.circuit` respecting a qubit mapping.

maffoo:
The mapped circuit is equivalent to the inner circuit, but with various
maps and transformations applied (qubit map, param map, measurement map,
repetitions, repetition ids). The mapping can be applied deeply, in
which case contained CircuitOperations are expanded recursively, or
shallowly, in which case contained CircuitOperations are left unchanged.

To implement the deep mapping, we add a map_ops method on
AbstractCircuit that applies a function to each op in the circuit and
then recombines them in a way that preserves moment structure as much as
possible.

Fixes #3783


cirq/circuits/circuit_operation.py
```diff
@@ -152,6 +152,28 @@ def _parameter_names_(self) -> AbstractSet[str]:
...
+        if self.repetitions < 0:
+            circuit = circuit ** -1
+        if self.repetition_ids is None or not protocols.is_measurement(circuit):
+            return circuit * abs(self.repetitions)
+        keys = protocols.measurement_keys(circuit)
+        return cirq.Circuit(
+            protocols.with_measurement_key_mapping(circuit, {key: f'{rep}-{key}' for key in keys})
+            for rep in self.repetition_ids
+        )
+
```
smitsanghavi, line 173: Just want to preemptively chime in on the draft PR. I think for `deep == False`, you'd still want to propagate `repetition_ids` for `CircuitOperations` and only map measurement_keys for direct child measurements like in `_decompose_` below. This is to avoid the pitfalls described in [repeated measurement RFC](http://docs/document/d/1_dY92lTKumaxH4wRdyabqQz_-i_xupNVwMzbrz__1U8?resourcekey=0-bcAZJNPDvnmthf5Y8VFVZQ#heading=h.gbik77iezhj5).

Also, it might be nice to have `mapped_op_tree` method that does what this does without converting to `Circuit`. That way, we can reuse this code for `mapped_circuit` and `_decompose_` both.

smitsanghavi (changes requested):

cirq/circuits/circuit_operation.py
95-martin-orion: +1 to reviewing pitfalls for `repetition_ids`, but I don't think `mapped_op_tree` is strictly necessary. If everything works as intended, `_decompose_` should just reduce to
```
def _decompose_(self):
    return self.mapped_circuit(False).all_operations()
```
`mapped_circuit` works better as the base method since `OP_TREE` doesn't preserve moments.

maffoo:
So I reworked this to handle repetition_ids in the `deep=False` case, but TBH I'm left with the feeling that the current way we are handling measurements inside subcircuits is not great. One problem is that the `measurement_keys` protocol seems broken in that it doesn't agree with the measurements that are actually returned if you sample a repeated subcircuit:
```python
In [8]: c = cirq.Circuit(
    cirq.Circuit(cirq.measure(q, key='m')).freeze().to_op().with_measurement_key_mapping({'m': 'n'}).repeat(3)
).freeze().with_measurement_key_mapping({'n': 'p'}).repeat(2)

In [9]: print(c.mapped_circuit().to_text_diagram(transpose=True))
(0, 0)
│
Circuit_0x633d381951d049ef:
[ (0, 0): ───M('m')───    ](key_map={m: p}, repetition_ids=['0-0', '0-1', '0-2'])
│
Circuit_0x633d381951d049ef:
[ (0, 0): ───M('m')───    ](key_map={m: p}, repetition_ids=['1-0', '1-1', '1-2'])
│

In [10]: print(c.mapped_circuit(deep=True).to_text_diagram(transpose=True))
(0, 0)
│
M('0-0-p')
│
M('0-1-p')
│
M('0-2-p')
│
M('1-0-p')
│
M('1-1-p')
│
M('1-2-p')
│

In [11]: cirq.measurement_keys(c)
Out[11]: {'p'}

In [12]: cirq.measurement_keys(c.mapped_circuit())
Out[12]: {'p'}

In [13]: cirq.measurement_keys(c.mapped_circuit(deep=True))
Out[13]: {'0-0-p', '0-1-p', '0-2-p', '1-0-p', '1-1-p', '1-2-p'}

In [14]: cirq.Simulator().sample(cirq.Circuit(c), repetitions=100)
Out[14]: 
    0-0-p  0-1-p  0-2-p  1-0-p  1-1-p  1-2-p
0       0      0      0      0      0      0
1       0      0      0      0      0      0
2       0      0      0      0      0      0
3       0      0      0      0      0      0
4       0      0      0      0      0      0
..    ...    ...    ...    ...    ...    ...
95      0      0      0      0      0      0
96      0      0      0      0      0      0
97      0      0      0      0      0      0
98      0      0      0      0      0      0
99      0      0      0      0      0      0

[100 rows x 6 columns]
```
The actual measurement keys produced by sampling agree with the deep mapped circuit ('0-0-p', '0-1-p', etc.) but asking for measurement keys on the original or shallowly mapped circuits gives a different result. I think we need to reconcile these. In particular, do we think of the key `p` as fundamental and then there is extra indexing information when it occurs inside loops, or is the prefixed key `0-0-p` the important thing? I would lean toward the former where we keep indexing info and keys separate because I think not doing so makes things like this "mapping" transform quite complex (and other transforms will likely be impossible without parsing keys out into indexing and suffix parts, then transforming and reassembling). But this probably would mean we need to incorporate the possibility for different measurement shapes into the `Result` type. I guess another possibility would be to support structure `Key` objects instead of just strings, where a `Key` object would contain both a string label and index info.

95-martin-orion:
> One problem is that the measurement_keys protocol seems broken in that it doesn't agree with the measurements that are actually returned if you sample a repeated subcircuit: {...}

@smitsanghavi: I suspect that this was the issue I encountered in my early attempt at repeated measurement.
1. The `measurement_keys` protocol must return the "final" keys of an object, but...
2. For re-mapping the COp needs access to the "pre-repetition" keys.

**Item (1) is broken and must be fixed.** This fix will break item (2); we'll probably need a separate method for calculating "pre-repetition" keys of a CircuitOperation. Some care is required here to avoid infinite recursion: "final" keys require re-mapping, which requires "pre-repetition" keys.

smitsanghavi:
@95-martin-orion Maybe it's time to revisit/prioritize the tuple measurement key idea in the [RFC](https://tinyurl.com/measurement-repeated-circuitop#heading=h.s3u7e3nnef7n)? As @maffoo mentioned, having a structure that enables separating the indexing and the leaf measurement key part from the "final" keys would help simplify this quite a bit. Key remapping would simply always operate on the "end" part. This would enable us to modify `measurement_keys` to return the "final" indexed keys without breaking the remapping of the non-indexed part of the key.

That said, it is *possible* to do this efficiently already using `rsplit('-', maxsplit=1)[-1]` to get the "pre-repetition" key from an indexed key (or even an unindexed one). Maybe we can put this logic in a public function `to_unindexed_measurement_key` for all to use, thereby taking out some of the complication.

But it would be nicer and faster to be able to do something like `final_key[-1]` or `final_key.key`. Also, the string splitting/parsing approach breaks down if the original measurement key has the `-` separator in it.

95-martin-orion:
> @95-martin-orion Maybe it's time to revisit/prioritize the tuple measurement key idea in the [RFC](https://tinyurl.com/measurement-repeated-circuitop#heading=h.s3u7e3nnef7n)?

I discussed this with @dstrain115 yesterday, and we came to a similar conclusion: since this is likely to generate breaking changes across Cirq, it should be prioritized ahead of the classical data work. Combining this change with #3233 might be worthwhile, as both affect a similar cross-section of the repository.

@smitsanghavi, could you take this item? Separately, we still need to address the immediate issue in this PR. If you can see a quick fix for both issues in my previous comment, that would be best; otherwise, we likely need to roll back the repeated measurement PR until a new measurement system is in place.

95-martin-orion:
@maffoo, if we need to unblock this you can prohibit calling `mapped_circuit` on repeated-measure `CircuitOperation`. Just add a TODO to remove the restriction and use `mapped_circuit` in decompose once `measurement_keys` is fixed for the repeated-measure case.

smitsanghavi:
SG, I'll explore the structured and expanded measurement representations and write an RFC.
Quick fix for the immediate issues highlighted here coming soon.

smitsanghavi:
#3823 FYI. The [integration tests](https://github.com/quantumlib/Cirq/actions) are a few hours behind in queue and will take time to catch up.

95-martin-orion:
@maffoo: #3823 is merged, with relatively clean solutions to the issues raised above. Syncing past that PR should make this behave as expected.

maffoo:
Superseded by #4188.

PR #3786, opened 2021-02-11
Format automatically covered lines with flynt

vtomole:
[Flynt: A tool to automatically convert old string literal formatting to f-strings](https://github.com/ikamensh/flynt)

What I did was

```
$ pip install flynt
$ flynt .
```
This is the first half to https://github.com/quantumlib/Cirq/issues/3804


balopat:
Nice! I'm pro this change! The coverage misses are probably genuinely uncovered areas.

balopat:
So close: 

```
************* cirq/ops/pauli_string.py (1 uncovered)
Line  990 is new or changed but not covered:          return False

Found 1 uncovered touched lines.
```

:)

vtomole:
@balopat ahhh!!!

vtomole:
@balopat That line might be unreachable and trying to refactor it so that it passes mypy is a time sink for me at the moment.  We can merge this and i can look into that line later.

vtomole:
@balopat Ping for your thoughts.

dabacon:
This is f-ing awesome.

balopat:
> This is f-ing awesome.

lol. 

regarding my thoughts: I think we can just retry the flynt, and revert the problematic files with no coverage. Then we can open an issue to finalize the flynt conversion on those files as well and fix the coverage.

vtomole:
Okay. In that case, this is ready to go in. I've filed https://github.com/quantumlib/Cirq/issues/3804.

mpharrigan:
What's the overarching plan here? Are we going to add a CI check that we only allow fstrings? Or just run this periodically? Is this something we should check during reviews? Is there consensus that fstrings are universally better than `.format`? Are we concerned about causing merge conflicts with in-flight PRs and branches? There's also the minor issue of muddying the git-blame waters. There were similar concerns with #3516 . @maffoo carefully crafted a series of tags to alleviate merge issues

balopat:
> What's the overarching plan here? Are we going to add a CI check that we only allow fstrings? Or just run this periodically? Is this something we should check during reviews? Is there consensus that fstrings are universally better than `.format`? Are we concerned about causing merge conflicts with in-flight PRs and branches? There's also the minor issue of muddying the git-blame waters. There were similar concerns with #3516 . @maffoo carefully crafted a series of tags to alleviate merge issues

Hmmm, great points. 

1. From what I can see (stackoverflow, blogs, and internal recommendations) f-strings are more performant and definitely more readable. There is only exception is when logging is used (https://stackoverflow.com/questions/52012564/lazy-evaluation-of-strings-in-python-logging-comparing-with-format/52012660): `logging.debug('log message with some %s', 'var')` - for lazy evaluation - but I'm not even sure flynt would detect that! 
2. Thus, I'd argue we could turn on flynt as a CI check 
3. I think the merge issues are going to be signficantly smaller here - with black basically _all_ the lines were reformatted. Here we have a portion of the files and very sparsely spread lines that are probably easy to merge. 
4. Git blame: yeah, somewhat annoying, but I wouldn't worry about it too much - again, the sparseness of the formatted string makes this okay, similar to larger scale refactorings (e.g. when we introduced num_qubits across all the gate classes)

mpharrigan:
Ok. It might be annoying to submit this before the CI check is in place as more `.format`s will sneak in

vtomole:
@mpharrigan I can add the CI check to this PR.  Currently, only one line is left to change here; but I can't budget the fix for  it. (see https://github.com/quantumlib/Cirq/pull/3786#issuecomment-777687140)

check/flynt
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
mpharrigan, line 1: can this be a part of check/format-incremental? Otherwise it's yet another thing you need to run before submitting a PR

vtomole: Done.

balopat: +1 to that suggestion

vtomole: Done.

vtomole:
@balopat  @dabacon  @mpharrigan  This is ready for a second look.

check/format-incremental
```diff
@@ -1,5 +1,21 @@
 #!/usr/bin/env bash
 
+################################################################################
+# Runs flynt on the repository
+#
+# Usage:
```
balopat, line 6: well, it is not check/flynt anymore - it is a bit confusing how this repeats the exact same lines of working directory as well. I think we can do better in merging the scripts!

balopat (changes requested):
LGTM after the nit + resolved conflict

check/format-incremental
vtomole: Done.

balopat (approved):
LGTM

mpharrigan:
`check/format-incremental` now takes ~8 seconds instead of ~1 second

It also puts a little promotional message each time

> _-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_.
Please run your tests before committing. Did flynt get a perfect conversion? give it a star at: 
~ https://github.com/ikamensh/flynt ~
Thank you for using flynt. Upgrade more projects and recommend it to your colleagues!

mpharrigan:
opened https://github.com/quantumlib/Cirq/issues/3901

PR #3789, opened 2021-02-11
Implement automatic inference of qid shape for quantum states

kevinsung:
Follow up to #3419 . Addresses #2904 .


cirq/qis/states.py
```diff
@@ -321,6 +313,231 @@ def _infer_qid_shape_from_dimension(dim: int) -> Tuple[int, ...]:
...
+    """Infer the qid shape of a set of states.
+
+    This is a heuristic that is guaranteed to return a qid shape compatible
+    with all of the given states, but may fail (raising an error) even
+    if there is a unique compatible qid shape. If, after applying the heuristic,
+    the qid shape is ambiguous, then an error is raised unless the qid shape is
+    to be inferred from the shape of a state vector or density matrix. In that
+    case, if the dimension of the state space is a power of 2, then the state
+    space is assumed to be composed of qubits; otherwise, it is assumed to be
+    composed of a single qudit.
```
mpharrigan, line 326: the syntax of this docstring is hard to parse. It's written like .. backwards .. for computers instead of for humans.

"We use a heuristic to determine a qid shape compatible with all of the given shapes. The heuristic [insert brief description here]. If the dimension of state vector or density matrix (not tensors!) is a power of 2, then we assume qubits, otherwise it is assumed to be composed of a single qudit. If the qid shape is still ambiguous, an error is raised."

cirq/qis/states.py
```diff
@@ -321,6 +313,231 @@ def _infer_qid_shape_from_dimension(dim: int) -> Tuple[int, ...]:
     return (dim,)
 
 
+def infer_qid_shape(*states: 'cirq.QUANTUM_STATE_LIKE') -> Tuple[int, ...]:
+    """Infer the qid shape of a set of states.
+
+    This is a heuristic that is guaranteed to return a qid shape compatible
```
mpharrigan, line 319: I think we have different definitions of guarantee. 60% of the time it works every time

cirq/qis/states.py
```diff
@@ -321,6 +313,231 @@ def _infer_qid_shape_from_dimension(dim: int) -> Tuple[int, ...]:
...
+    non_integer_states = []
+    for state in states:
+        if isinstance(state, int):
+            integer_states.append(state)
+        else:
+            non_integer_states.append(state)
+
+    # if all states are specified as integers, the shape is ambiguous
+    if not non_integer_states:
+        raise ValueError('The qid shape of the given states is ambiguous.')
```
mpharrigan, line 348: explain why in this error message (not just the comment)

cirq/qis/states.py
```diff
@@ -321,6 +313,231 @@ def _infer_qid_shape_from_dimension(dim: int) -> Tuple[int, ...]:
...
+
+    # attempt to infer qid shape from non-integer states
+    potential_shapes = _potential_qid_shapes(non_integer_states[0])
+    for state in non_integer_states[1:]:
+        potential_shapes = potential_shapes.intersection_subset(_potential_qid_shapes(state))
+    if potential_shapes.is_ambiguous():
+        raise ValueError('The qid shape of the given states is ambiguous.')
```
mpharrigan, line 355: change `is_ambiguous` to `assert_ambiguous` and give more helpful errors why it's ambiguous, or change `is_ambiguous` to return a `reason` string, which you can report here

cirq/qis/states.py
```diff
@@ -321,6 +313,231 @@ def _infer_qid_shape_from_dimension(dim: int) -> Tuple[int, ...]:
...
+        return _QidShapeSet(
+            explicit_qid_shapes=explicit_qid_shapes,
+            unfactorized_total_dimension=unfactorized_total_dimension,
+            min_qudit_dimensions=min_qudit_dimensions,
+        )
+
+    def is_ambiguous(self) -> bool:
+        """Whether the qid shape is ambiguous and cannot be inferred."""
+        if self.min_qudit_dimensions is not None:
+            return True
```
mpharrigan, line 491: return reasons instead of just true/false so people can figure out what's going on

cirq/qis/states.py
```diff
@@ -321,6 +313,231 @@ def _infer_qid_shape_from_dimension(dim: int) -> Tuple[int, ...]:
...
+
+    # attempt to infer qid shape from non-integer states
+    potential_shapes = _potential_qid_shapes(non_integer_states[0])
+    for state in non_integer_states[1:]:
+        potential_shapes = potential_shapes.intersection_subset(_potential_qid_shapes(state))
+    if potential_shapes.is_ambiguous():
+        raise ValueError('The qid shape of the given states is ambiguous.')
+    qid_shape = potential_shapes.infer_qid_shape()
+    if qid_shape is None:
+        raise ValueError('Failed to infer the qid shape of the given states.')
```
mpharrigan, line 358: more descriptive error telling you why. there's also a redundant check, since `infer_qid_shape` calls `is_ambiguous`

cirq/qis/states.py
```diff
@@ -321,6 +313,231 @@ def _infer_qid_shape_from_dimension(dim: int) -> Tuple[int, ...]:
...
+
+    # check if the shape is compatible with the states specified as integers
+    if integer_states and np.prod(qid_shape, dtype=int) <= max(integer_states):
+        raise ValueError('Failed to infer the qid shape of the given states.')
```
mpharrigan, line 362: describe the failure!

cirq/qis/states.py
```diff
@@ -321,6 +313,231 @@ def _infer_qid_shape_from_dimension(dim: int) -> Tuple[int, ...]:
...
+
+    Returns:
+        The inferred qid shape.
+
+    Raises:
+        ValueError: The qid shape of the given states is ambiguous.
+        ValueError: Failed to infer the qid shape of the given states.
+    """
+    integer_states = []
+    non_integer_states = []
```
mpharrigan, line 339: type annotations could be helpful here

cirq/qis/states.py
```diff
@@ -321,6 +313,231 @@ def _infer_qid_shape_from_dimension(dim: int) -> Tuple[int, ...]:
...
+    np.ndarray,
+    Sequence[Union[int, float, complex]],
+    # Product state object
+    'cirq.ProductState',
+    # Quantum state object
+    'cirq.QuantumState',
+]
+
+
+def _potential_qid_shapes(state: _NON_INT_STATE_LIKE) -> '_QidShapeSet':
```
mpharrigan, line 380: short docstring would be helpful here

cirq/qis/states.py
```diff
@@ -321,6 +313,231 @@ def _infer_qid_shape_from_dimension(dim: int) -> Tuple[int, ...]:
...
+
+    if isinstance(state, Sequence):
+        state = np.array(state)
+    if state.ndim == 1:
+        (dim,) = state.shape
+        if state.dtype.kind == 'c':
+            # definitely state vector amplitudes
+            return _QidShapeSet(unfactorized_total_dimension=dim)
+        # could also be per-qudit computational basis values
+        min_qudit_dimensions = tuple(state.astype(int, copy=False) + 1)
```
mpharrigan, line 394: what if `state.dtype.kind` is float (i.e. it just happens to not have complex elements)

cirq/qis/states.py
```diff
@@ -321,6 +313,231 @@ def _infer_qid_shape_from_dimension(dim: int) -> Tuple[int, ...]:
...
+            explicit_qid_shapes |= _intersection_explicit_with_unfactorized_qid_shapes(
+                other.explicit_qid_shapes, self.unfactorized_total_dimension
+            )
+
+        if self.unfactorized_total_dimension == other.unfactorized_total_dimension:
+            unfactorized_total_dimension = self.unfactorized_total_dimension
+
+        # don't attempt to intersect unfactorized total dimension with
+        # min qudit dimensions
+
```
mpharrigan, line 468: extra newline?

cirq/qis/states.py
```diff
@@ -321,6 +313,231 @@ def _infer_qid_shape_from_dimension(dim: int) -> Tuple[int, ...]:
...
+        if self.min_qudit_dimensions is not None and other.explicit_qid_shapes:
+            explicit_qid_shapes |= _intersection_explicit_with_min_qudit_dims_qid_shapes(
+                other.explicit_qid_shapes, self.min_qudit_dimensions
+            )
+
+        # don't attempt to intersect min qudit dimensions with
+        # unfactorized total dimension
+
```
mpharrigan, line 476: extra newline?

mpharrigan (commented):
cool

cirq/qis/states.py
kevinsung: I've rewritten it following your suggestion.

cirq/qis/states.py
kevinsung: Done.

cirq/qis/states.py
kevinsung: Done.

cirq/qis/states.py
kevinsung: Done.

cirq/qis/states.py
kevinsung: Done.

cirq/qis/states.py
kevinsung: Done.

cirq/qis/states.py
kevinsung: Done.

cirq/qis/states.py
kevinsung: Done.

cirq/qis/states.py
kevinsung: I've changed the code to check the type before converting to int.

cirq/qis/states.py
kevinsung: Those were deliberate, but I've changed these comments to be more clear.

cirq/qis/states.py
kevinsung: Those were deliberate, but I've changed these comments to be more clear.

kevinsung:
@mpharrigan I've addressed your comments, please have another look.

cirq/qis/measures.py
```diff
@@ -132,10 +133,7 @@ def fidelity(
 
     # Use QuantumState machinery for the general case
     if qid_shape is None:
-        # coverage: ignore
-        raise NotImplementedError(
-            'Qid shape inference not yet implemented. Please specify the qid shape explicitly.'
-        )
+        qid_shape = infer_qid_shape(state1, state2)
```
balopat, line 136: I think a message to specify the qid shape explicitly is still useful here to help the user. Otherwise only the inference errors will be shown which are a bit more obscure.

cirq/__init__.py
```diff
@@ -331,6 +331,7 @@
     dirac_notation,
     eye_tensor,
     fidelity,
+    infer_qid_shape,
```
balopat, line 334: I'm not convinced yet that this needs to be exposed to the cirq namespace. 
It feels like this would be useful mostly for other functions in the qis package.

balopat:
A couple of comments: 
- There is also `cirq.to_valid_state_vector` that should make use of this functionality - `states._state_like_to_state_tensor` seems to do a bunch of qid shape inference itself...

After playing around with the code, it felt a bit fragile and hard to understand the rules just by discovery...it might be that I'm not smart enough for it or I'm misusing the tool :) 

- I wonder if we want to adopt a "use minimal dimensions where possible" - it would allow for `infer_qid_shape(1)` to be `(2,1)`
- I wonder if we want to adopt automated casting - e.g. `infer_qid_shape([1,0])` fails now but  `infer_qid_shape([1,1j])` doesn't...what do we think about automated casting to complex?

mpharrigan:
> I wonder if we want to adopt automated casting - e.g. infer_qid_shape([1,0]) fails now but infer_qid_shape([1,1j]) doesn't...what do we think about automated casting to complex?

I think this is by design to allow integer specification of basis states vs complex amplitudes

cirq/qis/states.py
```diff
@@ -343,39 +368,36 @@ def infer_qid_shape(*states: 'cirq.QUANTUM_STATE_LIKE') -> Tuple[int, ...]:
...
-    if potential_shapes.is_ambiguous():
-        raise ValueError('The qid shape of the given states is ambiguous.')
     qid_shape = potential_shapes.infer_qid_shape()
     if qid_shape is None:
-        raise ValueError('Failed to infer the qid shape of the given states.')
+        raise ValueError(
+            'Failed to infer the qid shape of the given states. '
+            'Either there is no compatible qid shape for the given states, or there is one '
+            'but the heuristic failed to find it.'
+        )
```
mpharrigan, line 386: The rest of the new error messages look great. This one reads really funny, however. "Either there is none or there is one" is a bad excuse for a program to make :) :) If this is truly an edge case, I'd recommend admitting as much

> Failed to infer the qid shape of the given states. If you think there is an unambiguous result for this case, please raise a GitHub issue.

Otherwise, parcel out the conditionals that can lead here: e.g. 

> Could not determine the total (unfactorized) Hilbert space dimension and could not find any explicit Qid shapes.

cirq/__init__.py
mpharrigan: We've never really held back from exposing every public function to the global namespace before. We've never had a good packaging strategy before either. But in any event, we should make the criteria clear for global namespace instead of picking and choosing per PR

mpharrigan (approved):
My comments have been addressed. Please look at @balopat 's tho

balopat:
> > I wonder if we want to adopt automated casting - e.g. infer_qid_shape([1,0]) fails now but infer_qid_shape([1,1j]) doesn't...what do we think about automated casting to complex?
> 
> I think this is by design to allow integer specification of basis states vs complex amplitudes

I'm confused - list of ints wouldn't be integer specification of basis states as far as I understand, `infer_qid_shape([1,0])` is the `|0>` state, it's just not complex...

balopat:
> > > I wonder if we want to adopt automated casting - e.g. infer_qid_shape([1,0]) fails now but infer_qid_shape([1,1j]) doesn't...what do we think about automated casting to complex?
> > 
> > 
> > I think this is by design to allow integer specification of basis states vs complex amplitudes
> 
> I'm confused - list of ints wouldn't be integer specification of basis states as far as I understand, `infer_qid_shape([1,0])` is the `|0>` state, it's just not complex...

Oh, I got it - it could be looked at as a multiqubit integer state `|10>`...ahhh...

cirq/__init__.py
mpharrigan: I'm always in favor of making things package-private and expanding as necessary

cirq/qis/measures.py
kevinsung: Done.

cirq/__init__.py
kevinsung: I've removed the top-level exposure.

cirq/qis/states.py
```diff
@@ -311,17 +323,30 @@ def _infer_qid_shape_from_dimension(dim: int) -> Tuple[int, ...]:
...
-    space is assumed to be composed of qubits; otherwise, it is assumed to be
-    composed of a single qudit.
+    This is a heuristic to determine a qid shape compatible with all of the
+    given states. It works by attempting to find the intersection of the sets
+    of potential qid shapes for each given state. It may fail (raising an
+    error) even if there is a unique compatible qid shape. If the dimension of
+    a state vector or density matrix (but not state tensor) is a power of 2,
+    then the state space is assumed to be composed of qubits; otherwise, it is
+    assumed to be composed of a single qudit. If the qid shape is ambiguous,
+    an error is raised.
```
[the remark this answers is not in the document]
kevinsung: I agree, thanks for the suggestion!

cirq/qis/states.py
kevinsung: I took your first suggestion; I do believe that this case would be reached only in unusual circumstances.

kevinsung:
> A couple of comments:
> 
>     * There is also `cirq.to_valid_state_vector` that should make use of this functionality - `states._state_like_to_state_tensor` seems to do a bunch of qid shape inference itself...
> 
> 
> After playing around with the code, it felt a bit fragile and hard to understand the rules just by discovery...it might be that I'm not smart enough for it or I'm misusing the tool :)
> 
>     * I wonder if we want to adopt a "use minimal dimensions where possible" - it would allow for `infer_qid_shape(1)` to be `(2,1)` --- python world precedence: `bin(3) -> 0b11'
> 
>     * I wonder if we want to adopt automated casting - e.g. `infer_qid_shape([1,0])` fails now but  `infer_qid_shape([1,1j])` doesn't...what do we think about automated casting to complex?

I'll edit `states._state_like_to_state_tensor` to use this, and I'll also implement your suggestion to use minimum dimensions for integer states.

kevinsung:
@balopat I've addressed your comments, please have another look.

mpharrigan:
@balopat did Kevin address your comments?

balopat (approved):
LGTM, thank you!

mpharrigan:
@kevinsung can you sign the CLA?

kevinsung:
@mpharrigan I signed one years ago; do you have any idea why the check would stop going through all of a sudden?

balopat:
All is good on the CLA front, I think either Cirqbot might have a bug here or there was a CLA bot flake? I suspect the former.

PR #3791, opened 2021-02-12
Don't serialize GridQubit hash when pickling

### LINKED ISSUE #3777: q1 == q1 but hash(q1) != hash(q2)
I'm having trouble reproducing this, but I'm in a situation where the thing in the title is happening with `GridQubit`. I've narrowed it down to the fact that the cached `cirq.GridQubit._hash` values are different and recomputing them with `cirq.Qid.__hash__(q1)` gives correct, identical hash functions.

My leading theory is that I'm loading a previously-pickled gridqubit, a copy is being made somewhere and the hash is recomputed. The hash value of `Qid` (the class) is part of the hash input, which could differ over python instantiations

maffoo:
When hashing Qid objects we were including the `Qid` class, but class hashes are not stable between restarts of the interpreter. This caused #3777 where the hash of unpickled `GridQubit` objects disagrees with other `GridQubits` to which they compare equal. Here we fix this by hashing based on the class _name_ rather than the class object itself.

Fixes #3777


mpharrigan:
Are string hashes stable :)

maffoo:
@mpharrigan, that is a very good question :-)

I also noticed the tests were failing because we specifically _do not_ want the hashes to depend on the specific Qid class, since, e.g. `GridQubit(0, 0)` and `GridQid(0, 0, dimension=2)` compare equal so should hash to the same thing. I've switched instead to just implementing the `__getstate__` and `__setstate__` functions used when pickling to skip storing the hash and recompute it when unpickling. Also added a test that this works as intended with pickle.

dabacon:
Looks like there are also other places where the hash of the class is used, like Moment?

maffoo:
I think hash of a class is fine; we just have to avoid serializing hashes when e.g. pickling, because the hashes are not stable from one run of the python interpreter to the next. I'll change the issue title to reflect this.

mpharrigan (approved):
awesome! Thanks for doing the fix

Strilanc (approved):

PR #3792, opened 2021-02-12
[obs] 4.1 - Pre-requisites

mpharrigan:
This splits up #3647 into even more parts!

Coming up first: some private functions for adding parameterized rotations before and after a circuit, as well as mappings to initial states and final observable values. This also has a private class that previews our strategy for symmeterizing readout: In this procedure, we do half of our repetitions normally and half with an X gate immediately prior measurement. Since we're already monkeying with pre-measurement post-rotations, that X gate can get folded into the rotation used to measure a given observable! 

Preview of the next PR: https://github.com/mpharrigan/Cirq/pull/10


mpharrigan:
@balopat @viathor can you take a look?

cirq/work/observable_measurement.py
```diff
@@ -0,0 +1,210 @@
...
+
+if TYPE_CHECKING:
+    import cirq
+    from cirq.value.product_state import _NamedOneQubitState
+
+
+def _with_parameterized_layers(
+    circuit: 'cirq.Circuit',
+    qubits: Sequence['cirq.Qid'],
+    no_initialization: bool,
```
balopat, line 36: The lack of symmetry between initial and final rang some warning bells for me here, though I understand that the final layer will be always required (even in case of ZZZZZZ...) to implement the flips. 
However, we could just have the inital layer always defaulting to KET_ZERO...and as long as the compilers compile away negligible gates, then we don't have to have this extra arg that's carried across all the function calls.

cirq/work/observable_measurement.py
```diff
@@ -0,0 +1,210 @@
...
+    if no_initialization:
+        total_circuit = circuit.copy()
+    else:
+        total_circuit = circuits.Circuit([x_beg_mom, y_beg_mom])
+        total_circuit += circuit.copy()
+    total_circuit.append([x_end_mom, y_end_mom, meas_mom])
+    return total_circuit
+
+
+_PAULI_TO_PARAM_VAL: Dict[Tuple['cirq.Pauli', bool], Tuple[float, float]] = {
```
balopat, line 64: Maybe 
```suggestion
_OBS_TO_PARAM_VAL: Dict[Tuple['cirq.Pauli', bool], Tuple[float, float]] = {
```
? 
It took me a second to figure out what this is. Maybe add a bit of comment here.

cirq/work/observable_measurement.py
```diff
@@ -0,0 +1,210 @@
...
+        # for that qubit. We mandate that by the time you get to this stage,
+        # each _max_setting has
+        # weight(in_state) == weight(out_operator) == len(qubits)
+        # See _pad_setting
+        pauli = setting.observable[qubit]
+        xf_param, yf_param = _PAULI_TO_PARAM_VAL[pauli, flip]
+        params[f'{qubit}-Xf'] = xf_param
+        params[f'{qubit}-Yf'] = yf_param
+
+        if not no_initialization:
```
balopat, line 116: Ohhh double negative. If we keep this - maybe consider renaming to `prepare_state`?

But also, if the init state is well defined anyway, we could just do this always.

cirq/work/observable_measurement.py
```diff
@@ -0,0 +1,210 @@
...
+    qubits: List['cirq.Qid'],
+    pad_init_state_with=value.KET_ZERO,
+    pad_obs_with: 'cirq.Gate' = ops.Z,
+) -> InitObsSetting:
+    """Pad `max_setting`'s `init_state` and `observable` with `pad_xx_with` operations
+    (defaults:  |0> and Z) so each max_setting has the same qubits. We need this
+    to be the case so we can fill in all the parameters, see `_get_params_for_setting`.
+    """
+    obs = max_setting.observable
+    assert obs.coefficient == 1, "Only max_setting's should be padded."
```
balopat, line 136:
```suggestion
    assert obs.coefficient == 1, "Only the max_setting should be padded."
```

cirq/work/observable_measurement_test.py
```diff
@@ -0,0 +1,165 @@
...
+    _with_parameterized_layers,
+    _get_params_for_setting,
+    _pad_setting,
+    _subdivide_meas_specs,
+)
+
+
+def test_with_parameterized_layers():
+    qs = cirq.LineQubit.range(3)
+    circuit = cirq.Circuit(
```
balopat, line 30: I'd assert the actual content of the parameterized layers as well...this test can be passed with just adding a couple of empty moments or moments containing identity gates.

balopat (changes requested):

cirq/work/observable_measurement.py
```diff
@@ -0,0 +1,210 @@
...
+    import cirq
+    from cirq.value.product_state import _NamedOneQubitState
+
+
+def _with_parameterized_layers(
+    circuit: 'cirq.Circuit',
+    qubits: Sequence['cirq.Qid'],
+    no_initialization: bool,
+) -> 'cirq.Circuit':
+    """Return a copy of the input circuit with a parameterized single-qubit
```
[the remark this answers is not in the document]
mpharrigan: done (by hand, with re-wrapping)

cirq/work/observable_measurement.py
mpharrigan: The theory here is that the user will have already compiled their circuit and we only add single-qubit gates that we presume to be in the hardware-native gateset, so there's no room for a negligibility compiler. Or at least we wouldn't want to do that in case it disturbs how the user supplied `circuit`. 

It's true that with these utility functions this boolean has to be passed around a lot. I think this is just the reality of the situation where you sometimes have to have rotations at the beginning. We don't want to add them indiscriminately, as it could have performance implications. Of course, these are all private functions and the appropriate value for `no_initialization` will be deduced from user-supplied input in the next PR and handled transparently :)

cirq/work/observable_measurement.py
mpharrigan: changed the name, added a docstring, added a short docstring to the next mapping too

cirq/work/observable_measurement.py
mpharrigan: done offline

cirq/work/observable_measurement.py
mpharrigan: I think it complains if you try to set parameters that don't exist in the circuit

mpharrigan: changed the name of this parameter to `needs_init_layer` and inverted the logic

cirq/work/observable_measurement_test.py
mpharrigan: added gate-isinstance checks and checked the sympy symbol names in the parameterized layers here and for the needs_init_layer=True case

mpharrigan:
@balopat back at you

balopat (approved):
LGTM

PR #3793, opened 2021-02-12
Adding two options for the MPS simulator: max_bond and method

tonybruguier:
The ultimate goal is to truncate the SVD in a way that is closer to the paper. I think what they do is fix the maximum number of singular values.

Doing so means that the estimation of the fidelity reduction must be done at every step (and could be more accurate). However, this requires an option that is not yet part of the Quimb release, so left as a TODO.

Finally, since the options were getting numerous, I decided to factor out the options in a separate class.


cirq/contrib/quimb/mps_simulator.py
```diff
@@ -29,35 +29,60 @@
 from cirq.sim import simulator
 
 
+class MPSOptions:
```
95-martin-orion, line 32: Consider using `@dataclasses.dataclass` for this.

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -29,35 +29,60 @@
...
+            grouping: How to group qubits together, if None all are individual.
+        """
+        self.method = method
+        self.max_bond = max_bond
+        self.cutoff_mode = cutoff_mode
+        self.cutoff = cutoff
+        self.sum_prob_atol = sum_prob_atol
+
+    cutoff: float
+    sum_prob_atol: float
```
95-martin-orion, line 64: Are these necessary?

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -354,9 +368,8 @@ def __init__(
             n = self.grouping[qubit]
             self.M[n] @= qtn.Tensor(x, inds=(self.i_str(i),))
             initial_state = initial_state // d
-        self.rsum2_cutoff = rsum2_cutoff
-        self.sum_prob_atol = sum_prob_atol
-        self.num_svd_splits = 0
+        self.simulation_options = simulation_options
+        self.e_n_s: List[float] = []
```
95-martin-orion, line 372: Is there a more descriptive name we could use here (e.g. expand the acronym)? Otherwise a comment describing what this is would be helpful.

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -29,35 +29,60 @@
...
+        method: str = 'svds',
+        max_bond: Optional[int] = None,
+        cutoff_mode: str = 'rsum2',
+        cutoff: float = 1e-6,
+        sum_prob_atol: float = 1e-3,
+    ):
+        """Numerical options for the simulation.
+
+        Args:
+            method: How to split the tensor. Refer to the Quimb documentation
```
95-martin-orion, line 44: Recommend adding a link here to the quimb docs.

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -505,14 +531,13 @@ def estimation_stats(self):
 
         # The computation below is done for numerical stability, instead of directly using the
         # formula:
-        # estimated_fidelity = (1 - self.rsum2_cutoff) ** self.num_svd_splits
-        estimated_fidelity = 1.0 + np.expm1(np.log1p(-self.rsum2_cutoff) * self.num_svd_splits)
+        # estimated_fidelity = \prod_i (1 - e_n_s_i)
+        estimated_fidelity = 1.0 + np.expm1(sum([np.log1p(-x) for x in self.e_n_s]))
```
95-martin-orion, line 535: `sum` can operate directly on list comprehensions, e.g. `sum(x for x in range(5))`

95-martin-orion (changes requested):
In addition to the comments below, could you add tests for the new fields introduced in `MPSOptions`?

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Named in `estimated_gate_error_list`.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: No. Sorry. Removed.

tonybruguier:
Thanks. PTAL.

cirq/contrib/quimb/mps_simulator.py
95-martin-orion: `@dataclasses.dataclass` automatically creates the `__init__` method (and several others) from variable definitions - see the [docs](https://docs.python.org/3/library/dataclasses.html) for details. With that in mind, it should be possible to condense this to:
```
@dataclasses.dataclass
class MPSOptions:
    """Put the __init__ docstring here"""
    # ...Quimb docs link...
    method: str = 'svds',
    max_bond: Optional[int] = None,
    cutoff_mode: str = 'rsum2',
    cutoff: float = 1e-6,
    sum_prob_atol: float = 1e-3,
```

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -52,17 +54,17 @@ def __init__(
...
                 the probabilities is not 1.0. This parameter is the absolute
                 deviation from 1.0 that is allowed.
-            grouping: How to group qubits together, if None all are individual.
         """
+
+        # Quimb documentation:
+        # https://quimb.readthedocs.io/en/latest/_autosummary/ \
+        #       quimb.tensor.tensor_core.html#quimb.tensor.tensor_core.tensor_split
         self.method = method
         self.max_bond = max_bond
```
95-martin-orion: Please add tests for these new fields.

95-martin-orion (changes requested):

cirq/contrib/quimb/mps_simulator.py
mpharrigan: also use `frozen=True`

tonybruguier: Neat. Done and Done.

cirq/contrib/quimb/mps_simulator.py
tonybruguier: Done.

tonybruguier:
Thanks. PTAL.

95-martin-orion (approved):
All looks good!

Issue #3775, opened 2021-02-10
[XEB] Plan to deprecate duplicate functionality


mpharrigan:
## `cirq.experiments.random_quantum_circuit_generation` ✅ 

This has fully-general utilities for making random quantum circuits on grids. Everything should use this.

## `cirq.experiments.fidelity_estimation` 🟡 

The first set of functions are nice enough, and are appropriate for computing XEB fidelity like in Arute 2019, although there are no examples on how to use this in practice. We should add some!

The final set of functions (`least_squares_xxx`) are appropriate for computing XEB fidelity for narrow (i.e. two-qubit) circuits. This math is used in the forthcoming suite of xeb characterization utilities (https://github.com/quantumlib/Cirq/issues?q=label%3Aarea%2Fxeb+) however, the data structures used as input and output are very confusing and not suitable for fast, numpy-vectorized-style application to real results. I think these should be deprecated in their current form

## `cirq.experiments.cross_entropy_benchmarking` 🔴 

And `examples/cross_entropy_benchmarking_example`.

This has a self-contained end-to-end implementation of cross entropy benchmarking. Therefore, it has its own fidelity estimation code and random circuit generation code. At the very least, it should be refactored to use the functionality in the more generalized modules described above and the example should use a noisy simulator. However, I recommend these modules be deprecated/removed completely: the example is not particularly instructive and the tight coupling between circuit construction, sampling, simulation, and analysis makes it hard to adapt and extend. The example is redundant with the more rich documentation provided as part of the [xeb prs](https://github.com/quantumlib/Cirq/issues?q=label%3Aarea%2Fxeb+)

## `cirq.experiments.google_v2_supremacy_circuit` 🔴 

This uses terminology we're trying to move away from. It also has its own, third implementation of random quantum circuit generation. These are also not the random circuits used in the flagship arute 2019 nature paper but rather the 2018 boixo theory paper. As far as I can tell, this is supposed to act as a supplemental information for that paper. This is not a good use of the Cirq repository. This should be re-written to use `random_quantum_circuit_generation`, renamed, and moved to a different repository.

## `grid_parallel_two_qubit_xeb` 🟡 

This module uses the "data collection idioms" from ReCirq to run a full parallel-2q-XEB experiment and save the results in a structured way. Unfortunately, this file is over 500 lines long and contains a lot of "business logic". This should be re-written to use the new XEB utilities and moved to ReCirq.

[...]

PR #3795, opened 2021-02-12
[XEB] Optimize/characterize by pair

mpharrigan:
Add a new method to perform optimization of angles (aka characterization) by pair. This is essential for "parallel" XEB. At it's heart, this is simply a "groupby" operation on the dataframe before calling the characterize function. There are additional code changes:

 - Lots of boilerplate for making a "closure class" so we can optimize by pair in parallel. This is an embarrassingly parallel operation, and it's important when doing very "wide" (i.e. lots of pairs) calibrations.
 - The least-squares estimation of individual (one pair, one cycle depth) fidelities involved using pandas `apply`, which turned out to be very slow! I profiled the `benchmark` function on a test workflow. It was taking 7.87 seconds, of which 5.3s were cumulative inside the apply function. The rest of it involved simulating the many circuits. `benchmark_...` is called during the optimization's objective function so it's performance critical. With the change, the call was a total of 2.5s and the least-squares estimation no longer appeared in the top many cumulative-time profiling calls.
 - Bug fix if you tried to do parallel XEB on a device that couldn't do all four layers. With added test.
 - The result of the `characterize_..` function has been beefed up to be a dataclass with more fields. Not only do you get the optimization results, but a nicer dictionary of angles and a dataframe of the refit fidelities. This simplifies one of the older notebooks. This data is also per-pair so you can do parallel xeb effectively
 - An exponential decay fitting function, taken from `experiments.cross_entropy_benchmarking`, which will be deprecated, #3775 
 - Some helpful pandas magic wrapped with nice function names to simplify plotting


cirq/experiments/xeb_fitting.py
```diff
@@ -289,10 +306,209 @@ def _mean_infidelity(angles):
...
+        fid_dfs.append(result.fidelities_df)
+
+    return XEBCharacterizationResult(
+        optimization_results=optimization_results,
+        final_params=all_final_params,
+        fidelities_df=pd.concat(fid_dfs),
+    )
+
+
+def exponential_decay(cycle_depths: np.ndarray, A: float, layer_fid: float) -> np.ndarray:
```
mrwojtek, line 422: A?

mpharrigan: this is my go-to for the multiplicative constant in exponential functions: `y = A e^{b x}`. I don't know if there's a descriptive name that can be truly given. Theoretically, it should be representative of SPAM error. Do you have a suggestion for a better name?

mpharrigan: See, for example: https://en.wikipedia.org/wiki/Exponential_function

They don't capitalize it. 

In the context of fitting: https://mathworld.wolfram.com/LeastSquaresFittingExponential.html they do capitalize it

mrwojtek: It was very strange to see this variable name here, especially since it lights up as a constant because of Python syntax coloring, Maybe it could be named just scale_constant, scale_factor?

Btw, this is not my only comment on this PR, I need a little bit more time to read into this more though. I'll do it by Monday morning PT.

cirq/experiments/xeb_fitting.py
```diff
@@ -289,10 +306,209 @@ def _mean_infidelity(angles):
...
+          column names suffixed by "_c".
+    """
+    fit_decay_df_0 = fit_exponential_decays(fidelities_df_0)
+    fit_decay_df_c = fit_exponential_decays(characterization_result.fidelities_df)
+
+    joined_df = fit_decay_df_0.join(fit_decay_df_c, how='outer', lsuffix='_0', rsuffix='_c')
+    joined_df['characterized_angles'] = [
+        characterization_result.final_params[pair] for pair, _, _ in joined_df.index
+    ]
+    angle_names = list(list(characterization_result.final_params.values())[0].keys())
```
mrwojtek, line 509: Why double list()?

cirq/experiments/xeb_fitting_test.py
```diff
@@ -83,6 +94,49 @@ def test_benchmark_2q_xeb_fidelities_parallel():
     n_pairs = sum(len(c.pairs) for c in combs)
     assert len(fid_df) == len(cycle_depths) * n_pairs
 
+    fit_df = fit_exponential_decays(fid_df)
+    for _, row in fit_df.iterrows():
+        assert list(row['cycle_depths']) == list(cycle_depths)
+        assert len(row['fidelities']) == len(cycle_depths)
+
+
+def test_benchmark_2q_xeb_fidelities_vectorized():
```
mrwojtek, line 103: Would it be possible to have a test with hard-coded expected values of all the intermediate variables instead of realying on two different implementations of the same logic?

mrwojtek (commented):
One thing which is beyond this PR but might be a concern is that it's very hard to review if the code operating on DataFrames is correct. The dataframes have a lot of context which is almost impossible to learn from the code itself.

One possibility to ease this might be having more verbose/explicit tests with hard-coded input and output dataframes.

docs/characterization/isolated_xeb.ipynb
```diff
@@ -208,6 +208,103 @@
     "plt.xlabel('Cycle Depth $d$')\n",
```
mrwojtek, line 208: <p>The same comment as in parallel version (describe options better)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/quantumlib/Cirq/pull/3795/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

docs/characterization/parallel_xeb.ipynb
```diff
@@ -77,6 +77,7 @@
     "circuit_library = rqcg.generate_library_of_2q_circuits(\n",
```
mrwojtek, line 77: <p>Make a comment what the default options are (or make all the keyword arguments explicit in this example).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/quantumlib/Cirq/pull/3795/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

mrwojtek:
I consider the comments minor and I'll LGTM it now.

mrwojtek (approved):

mpharrigan:
Re dataframes and testing experimental/workflow code in general: I think the notebooks (which thanks to @balopat are now run against cirq master on each PR!) are super helpful for making sure the whole flow works as expected. You can already see in the "unit" tests that I have to basically mock out a whole experiment to test a given part. This is expected from the more "workflow-type" functionality in this suite of PRs. The unit behavior is simple and tested elsewhere (generating random circuits, simulating them, executing them ... this is all covered well in cirq!)

cirq/experiments/xeb_fitting.py
```diff
@@ -289,10 +306,209 @@ def _mean_infidelity(angles):
...
+        initial_simplex_step_size: float = 0.1,
+        xatol: float = 1e-3,
+        fatol: float = 1e-3,
+    ):
+        self.parameterized_circuits = parameterized_circuits
+        self.cycle_depths = cycle_depths
+        self.phased_fsim_options = phased_fsim_options
+        self.initial_simplex_step_size = initial_simplex_step_size
+        self.xatol = xatol
+        self.fatol = fatol
```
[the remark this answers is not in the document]
mpharrigan: good call!

cirq/experiments/xeb_fitting.py
```diff
@@ -289,10 +306,209 @@ def _mean_infidelity(angles):
...
+    )
+
+
+def exponential_decay(cycle_depths: np.ndarray, A: float, layer_fid: float) -> np.ndarray:
+    """An exponential decay for fitting."""
+    return A * layer_fid ** cycle_depths
+
+
+def _fit_exponential_decay(cycle_depths: np.ndarray, fidelities: np.ndarray) -> Tuple[float, float]:
+    """Fit an exponential model fidelities = A * layer_fid**x using nonlinear least squares."""
```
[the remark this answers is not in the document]
mpharrigan: added. They reference the equation above

cirq/experiments/xeb_fitting.py
mpharrigan: would you be happier with a lowercase `a`? scale_factor could refer to scaling along the x-axis as well. I don't think we should try to prescribe too much meaning to a fit parameter that doesn't have a clear definition. See e.g. https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.curve_fit.html for more examples of people using a, b, c as fit parameters

cirq/experiments/xeb_fitting.py
```diff
@@ -289,10 +306,209 @@ def _mean_infidelity(angles):
...
+
+    return XEBCharacterizationResult(
+        optimization_results=optimization_results,
+        final_params=all_final_params,
+        fidelities_df=pd.concat(fid_dfs),
+    )
+
+
+def exponential_decay(cycle_depths: np.ndarray, A: float, layer_fid: float) -> np.ndarray:
+    """An exponential decay for fitting."""
```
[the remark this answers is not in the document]
mpharrigan: added

cirq/experiments/xeb_fitting.py
mpharrigan: I suppose the outer list isn't necessary. The inner one is because I want the first value of `characterization_result.final_params` and dictionaries return iterators which don't allow indexing.

I've re-worked this using fancy unpacking to avoid *any* `list` calls. How about that!

cirq/experiments/xeb_fitting_test.py
mpharrigan: that seems more fragile, no? having two independent implementations is the best possible unit test!

docs/characterization/parallel_xeb.ipynb
mpharrigan: <p>good point. Might make sense to add factory functions / globals with common, better-named defaults like in floquet. I'll write out all the kwargs for now. </p>

docs/characterization/isolated_xeb.ipynb
mpharrigan: <p>done</p>

mpharrigan:
PTAL (I know you already approved, but I'll give you a chance to respond to my responses anyways)

cirq/experiments/xeb_fitting_test.py
mrwojtek: How do you test if the other implementation doesn't have a bug then?

mrwojtek (approved):
As for the variable name the lower case a is better but I still feel the code which implements some algorithm should be structured differently than the mathematical description of that algorithm, for readability reasons.

cirq/experiments/xeb_fitting_test.py
mpharrigan: there are other tests that test the whole function, which the previous and current one is passing. This is an additional test to make sure the two implementations are identical.

95-martin-orion (approved):
CODEOWNERS approval for `docs/` files, for which @mrwojtek is not an owner.

PR #3574, opened 2020-12-03
Drop python 3.6 support

balopat:
Drop support (CI tests) for python 3.6.

Following #3347 we can safely drop python 3.6 support. This is also motivated by current breakages in Mac 3.6 and RTD #3554. This will not finalize #3347, we'll need to add some more documentation around the python support to close it fully.

Fixes #3554. See the successful build from my fork: https://readthedocs.org/projects/balopat-cirq-rtd/builds/12481463/

PR #3797, opened 2021-02-12
back to python 3.6

balopat:
Due to colab having 3.6 we'll have to relax the 3.7>= requirement.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,7 +62,7 @@ setup(
     url='http://github.com/quantumlib/cirq',
     author='The Cirq Developers',
     author_email='cirq@googlegroups.com',
-    python_requires=('>=3.7.0'),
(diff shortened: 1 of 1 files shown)


maffoo (commented):
It'd be good to also add 3.6 to the test matrix in github actions. There may be some 3.7-isms that have crept into the code.

95-martin-orion:
> It'd be good to also add 3.6 to the test matrix in github actions. There may be some 3.7-isms that have crept into the code.

+1 to this - I remember that we moved away from 3.6 partially due to errors in the 3.6 tests. If we can do a "rollback" of the original 3.6-deprecation PR, that might be best.

balopat:
yeah, I was just thinking that. I'll close this and just revert https://github.com/quantumlib/Cirq/pull/3574

Issue #3772, opened 2021-02-09
python 3.9 support


balopat:
As per our - not yet documented (https://github.com/quantumlib/Cirq/issues/3347) - python policy we should support 3 versions of python at a time. In Oct 2020, the stable version of Python 3.9.0 was released, we should support it. 
This means adding 3.9 to the test matrix in .github/workflows/ci.yml `python-version: [ '3.7', '3.8' ]` for Ubuntu, Windows and MacOS builds, and working out all the dependency kinks. 

@kevinsung started to pave the way with the [issues](https://github.com/quantumlib/Cirq/issues?q=is%3Aissue+is%3Aopen+label%3Apython3.9) he opened, someone just needs to work through them.

kevinsung:
Due to https://github.com/quantumlib/Cirq/issues/3769 it looks like this depends on numba releasing version 0.53.0.

PR #3798, opened 2021-02-12
Revert "Drop python 3.6 support"

balopat:
Reverts quantumlib/Cirq#3574.

Rationale: we want to ensure that Cirq works on colab which is stuck at the moment at Python 3.6.


mpharrigan:
What's the ETA on colab getting a more modern Python version? Python 3.9 is out now; we should be testing against it. And now our test matrix is getting quite large

mpharrigan:
We should follow https://numpy.org/neps/nep-0029-deprecation_policy.html 

According to that, we should be Python 3.7+ and numpy 1.17+

95-martin-orion:
> What's the ETA on colab getting a more modern Python version? Python 3.9 is out now; we should be testing against it. And now our test matrix is getting quite large

We have a `python 3.9` label for this - umbrella issue is #3772.

balopat:
Opened https://github.com/quantumlib/Cirq/issues/3800 to track colab's progress.

.github/workflows/ci.yml
```diff
@@ -189,7 +193,7 @@ jobs:
     name: Pytest MacOS
     strategy:
       matrix:
-        python-version: [ '3.7', '3.8' ]
+        python-version: ['3.7', '3.8' ]
```
maffoo, line 196: turbo nit: remove extra whitespace before final bracket.

maffoo (approved):
LGTM

PR #3799, opened 2021-02-12
Add noise-checker for circuits.

95-martin-orion:
Primary motivator is helping qsim determine which simulator mode (noiseless or noisy) to use for a given circuit. Possibly useful for other simulators with optimizations depending on circuit noise.

Question: is there a better name for this? I imagine channels / mixtures don't always represent noise - that's just the main case that we need this for on the qsim end.

### DIFF
--- a/cirq/circuits/circuit.py
+++ b/cirq/circuits/circuit.py
@@ -90,6 +90,7 @@ class AbstractCircuit(abc.ABC):
     *   are_all_matches_terminal
     *   are_all_measurements_terminal
     *   unitary
+    *   is_noisy
     *   final_state_vector
     *   to_text_diagram
     *   to_text_diagram_drawer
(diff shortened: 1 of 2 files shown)


cirq/circuits/circuit.py
```diff
@@ -964,6 +965,23 @@ def unitary(
         result = _apply_unitary_circuit(self, state, qs, dtype)
         return result.reshape((side_len, side_len))
 
+    def is_noisy(self) -> bool:
```
balopat, line 968: Quick question: why doesn't `has_unitary` work for you? Can we change that to cater for your requirements? What are those?

95-martin-orion: Measurements and parameters cause `has_unitary` to return false. This is the correct behavior; a circuit with a measurement or an unresolved parameter _cannot_ produce a unitary without ignoring or resolving those elements first.

However, such elements are not noisy - measurement is a permissible non-unitary in noiseless circuits, and parameterized objects with a noiseless base type will resolve to a noiseless object.

balopat: Even circuits with mid-circuit measurements are considered noiseless?

95-martin-orion: For the purposes of qsim, yes. The base qsim simulator can handle mid-circuit measurements; it can't handle a `Channel`.

More generally, while both measurement and noise can be represented as channels with some set of operators, only measurement reveals which operator was applied. For noise, this is only possible in simulation; based on this distinction, I don't think measurement should be considered "noise" even though it makes a circuit non-unitary.

95-martin-orion:
Discussed offline: not enough evidence that this is a useful unique predicate to define for Circuit.

PR #3801, opened 2021-02-13
Fix Result constructing a pandas dataframe to compute repetitions

Strilanc:
- This is really really expensive compared to looking up the value more directly like this

### DIFF
--- a/cirq/study/result.py
+++ b/cirq/study/result.py
@@ -150,6 +150,7 @@ class Result:
 
     @property
     def repetitions(self) -> int:
+        # Get the length quickly from one of the keyed results.
         return len(next(iter(self.measurements.values())))
 
     # Reason for 'type: ignore': https://github.com/python/mypy/issues/5273
--- a/cirq/study/result.py
+++ b/cirq/study/result.py
@@ -150,7 +150,7 @@ class Result:
 
     @property
     def repetitions(self) -> int:
-        return self.data.shape[0]
+        return len(next(iter(self.measurements.values())))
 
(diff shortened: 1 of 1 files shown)


cirq/study/result.py
```diff
@@ -150,7 +150,7 @@ def measurements(self) -> Dict[str, np.ndarray]:
 
     @property
     def repetitions(self) -> int:
-        return self.data.shape[0]
+        return len(next(iter(self.measurements.values())))
```
viathor, line 153: Optional: I'd add a comment explaining why we do it this way, just to be sure that no one "cleans it up" later.

viathor (approved):

cirq/study/result.py
maffoo: nit: I'd suggest avoiding `next` like this, because it the iterator happens to be empty it raises `StopIteration` which is a special `BaseException` that can have unintended consequences if you let it propagate up the stack (imagine `repetitions` gets called inside a list comprehension somewhere up the stack). I'll grant that it'd be uncommon to have an empty `Result` object, but I've had to deal with stray `StopIteration`s enough that I'm a bit of a stickler about this :-) Internally we have defined an `iterators.first` function to do what is intended here but raise `ValueError` on empty iterable instead; we should consider doing that in cirq to remove the temptation to reach for `next`.

Strilanc: Ran into this almost immediately, despite most simulators refusing to produce an empty result. https://github.com/quantumlib/Cirq/issues/3852

Issue #3246, opened 2020-08-19
Circuit construction


balopat:
**Problem:** Insertion or appending operations or moments into circuits come with a rich number of options (insert, insert_into_range, etc) as well as different behaviors in the form of insertion strategies.  For advanced users this power can be useful, but for many users, it results in them not being able to transform their thinking into what they expect to happen (in part because the effects are non-local and so hard to reason about as the size of the “action” grows).  This also has a direct impact on optimizers, which are mutating circuits, and where these insertion patterns can interact in strange ways if not done carefully.  The goal of solving this problem is to either modify existing or define new methods to aid in the construction of circuits and lessen the complexity of these notions.  Alternatively there could be other structures which aid in common construction patterns, Circuit builders.  This could be aided by the solution to Subcircuits (#3235).

**Rough requirements:**

- Augment or modify existing creation APIs to simplify the construction, or create helpers that make this easier.
- Standardize examples and tutorial code to simplest methods
- Be able to construct appropriate groups as defined in Subcircuits

balopat:
TODO: scope this and figure out whether we want to keep it or merge it with Transformers API items

vcatherine:
@tanujkhattar I recommend closing this in favour of more specific issues

dstrain115:
Closing this since the roadmap item is obsolete.  If there are more specific action items recommended, please open them.

Issue #2775, opened 2020-02-18
Circuit as simple container


viathor:
Presently, the Circuit class provides fairly sophisticated method for inserting operations which is useful for writing optimizers (e.g. insert earliest). I think this is a misplaced responsibility. Circuit should be a simple(*) container for storing and organizing gate operations but without circuit optimization logic. In particular, all insertions should require the client to say where the insertion is to occur. If some gates need to be moved earlier to reduce circuit duration by filling empty moments, then this counts as a circuit optimization and should therefore live in optimizers, not the Circuit class.

One reason for this separation is pure esthetics of the Circuit-as-a-simple-container approach. A more practical reason is that Circuit is a core concept in cirq and as such it should be simple, reliable and very stable over future releases. These properties are easier to obtain when complicated logic like circuit optimizations lives elsewhere. This is inspired by some reflection on #2743.

Concretely, this issue proposes to remove `InsertStrategy` from Circuit and instead introduce the functionality as an optimizer. This would be a breaking change that would need careful and gradual introduction. It's one of the things we'd want to complete before v1.0.

Note that another welcome effect of this change would be simplification of Circuit API.

I'm curious what y'all think.

---
(*) For the purposes of this issue, containers such as priority queues count as complex containers.

mpharrigan:
I know we're not technically at v1.0 but I'd consider circuit construction as a feature for which we already offer a high degree of backwards compatibility and for which there is already likely lots of code in the while depending on it. This sounds like a reasonable request for v2.0 several years down the line :)

dabacon:
I agree the insert methods are confusing and hard to use (I usually fail the first time).  One part of this is that this is because they behave differently depending on the content of the container, and that's not something that we typically encounter.

I disagree a bit that this functionality should be put in optimizers, there is something particularly nice about using EARLIEST insertion strategy when building some circuits.  Build, then compress, feels odd.  Maybe what we are missing here is a the proper language for defining these behaviors.  I have a similar feeling when I use PointOptimizer.  

As @mpharrigan says these might be widely used now.  I do wonder how many people use each insert strategy or the frontier or batch insert methods.

mpharrigan:
What do we think? Is there a maintenance burden to continuing to support the methods we already have? Should we start the deprecation process?

viathor:
I think the arguments above that Circuit should be simple, robust and slowly changing still stand, but before we decide, we should address the convenience issue raised above by @dabacon.

[...]

PR #3805, opened 2021-02-13
Add Circuit.tetris_concat and FrozenCircuit.tetris_concat

Strilanc:
Concatenation with overlap.


cirq/circuits/circuit.py
```diff
@@ -1258,6 +1258,60 @@ def _json_dict_(self):
     def _from_json_dict_(cls, moments, device, **kwargs):
         return cls(moments, strategy=InsertStrategy.EARLIEST, device=device)
 
+    def raggedy_add(*circuits: 'cirq.Circuit') -> 'cirq.Circuit':
```
daxfohl, line 1261: Should this be a module-level method since it doesn't contain a self or cls reference?

cirq/circuits/circuit.py
```diff
@@ -1258,6 +1258,60 @@ def _json_dict_(self):
...
+            circuits: The circuits to concatenate.
+
+        Returns:
+            The concatenated and overlapped circuit.
+        """
+        if len(circuits) == 0:
+            return Circuit()
+        acc = Circuit(circuits[0])
+        for k in range(1, len(circuits)):
+            acc = _raggedy_add_helper(acc, circuits[k])
```
daxfohl, line 1290: Looks like this is O(n^2). Could it be O(n) if the helper mutated acc directly? No biggie if this is just intended for small `n`.

cirq/circuits/circuit.py
daxfohl: Also would it work with `AbstractCircuit` input?

cirq/circuits/circuit.py
```diff
@@ -1258,6 +1258,60 @@ def _json_dict_(self):
...
         return cls(moments, strategy=InsertStrategy.EARLIEST, device=device)
 
+    def raggedy_add(*circuits: 'cirq.Circuit') -> 'cirq.Circuit':
+        """Concatenates circuits while overlapping them if possible.
+
+        Starts with the first circuit (index 0), then iterates over the other
+        circuits while folding them in. To fold two circuits together, they
+        are placed one after the other and then the second one is moved backward
+        until either just before its operations would collide with operations
+        in the first one or else their ends are equal.
```
daxfohl, line 1268: Is it important that their ends stay equal if there's no collision? Otherwise you could keep moving the second circuit backward until there's actually a collision, which, if you're adding multiple circuits together, could end up more compact overall. It would also preserve associativity this way.

cirq/circuits/circuit.py
Strilanc: This way of defining the method makes it usable as both an instance method and a class method, while preserving the ability to pass in any number of arguments. `c1.raggedy_add(c2)` works, `cirq.Circuit.raggedy_add(c1, c2)` works, `cirq.Circuit.raggedy_add()` works, `cirq.Circuit.raggedy_add(c1, c2, c3)` works, `c1.raggedy_add(c2, c3, c4)` works, etc.

cirq/circuits/circuit.py
Strilanc: Hmm, I definitely conceived of this method as trying to line up the starts or the ends as opposed to shoving it as far in as possible. Maybe it should be a parameter.

cirq/circuits/circuit.py
daxfohl: The reason I mention it is there's a roadmap item to simplify circuit construction at #3246. Some of the discussion in linked issue #2775 is that perhaps Circuit itself should only contain the core insert/append, whereas the more specialized functions could live as module-level methods, maybe in a separate builder module eventually, so that Circuit class itself remains lightweight. Once something is added to Circuit proper, it's hard to remove.

cirq/circuits/circuit.py
daxfohl: I'd pick whichever fits your use case and yagni the parameter. Mainly I came to this while thinking about the associativity. But ultimately associativity is not important.

cirq/circuits/circuit.py
daxfohl: (And my second comment was in reference to the signature: should the parameter instead be `*circuits: 'cirq.AbstractCircuit'`)?

Strilanc: Done. Parameter type loosened, and added to FrozenCircuit also.

cirq/circuits/circuit.py
Strilanc: Fixed. The runtime is now O(n). In fact it is O(min(n, q * t)) where q is the number of qubits in the system and t is the amount the circuits were overlapped.

cirq/circuits/circuit.py
Strilanc: Continuing to move it backward doesn't restore associativity. I updated the example to reflect this.

The method now has `stop_at_first_alignment = False` as a parameter. By default it stops at the last alignment. There's no option for continuing past that point when the circuits act on disjoint qubits, because I havea hard time picturing why you would ever want the second one to end up in front of the first one.

daxfohl: Ah, I didn't consider those examples. Yeah I think that parameter makes sense then.

cirq/circuits/circuit.py
daxfohl: I'm still seeing square time. In the helper function it constructs the return value from scratch. So if this is called to add, say, N identical single-moment circuits, it would call that helper N times, and constructing that return value is an O(N) operation itself. I may be missing something though.

cirq/circuits/circuit.py
```diff
@@ -1258,6 +1258,100 @@ def _json_dict_(self):
...
+
+def _overlap_collision_time(
+    c1: 'cirq.AbstractCircuit', c2: 'cirq.AbstractCircuit', stop_at_first_alignment: bool
+) -> int:
+    # Tracks the first used moment index for each qubit in c2.
+    # Tracks the complementary last used moment index for each qubit in c1.
+    seen_times: Dict['cirq.Qid', int] = {}
+
+    # Start scanning from end of first and start of second.
+    upper_bound = (min if stop_at_first_alignment else max)(len(c1), len(c2))
```
daxfohl, line 1314: I was thinking this would be `len(c1)` if not stop_at_first_alignment, not the max of the two. Otherwise if c1 and c2 have zero overlaps and c2 is longer, it pushes c2 start to before c1.

Strilanc: Pushing c2 to before c1 would correspond to the upper bound being the sum of the two lengths, not their max.

The bound has to be symmetric in the circuit lengths in order to correctly handle the case where c1 is much smaller and also the case where c2 is much smaller.

cirq/circuits/circuit.py
Strilanc: Oh, I see the quadratic time you were talking about now. I thought you were referring to the overlap collision computation (which did used to be quadratic).

The whole thing is now linear time.

cirq/circuits/circuit.py
daxfohl: An example here would be

```
c1 = cirq.Circuit(
    cirq.X(cirq.LineQubit(0)),
)
c2 = cirq.Circuit(
    cirq.H(cirq.LineQubit(1)),
    cirq.H(cirq.LineQubit(1)),
)
c3 = raggedy_add(c1, c2)
```

This results in the `X` from `c1` getting pushed forward to moment 1, which seems odd to me: I'd not expect `c1`'s positions to change. I'd expect `c2` never to get pushed back further than `c1`'s start, even if it's longer.
```
0: ───────X───
1: ───H───H───
```

Strilanc: That's the intended behavior. If you want it to stop earlier, you set `stop_at_first_alignment=True`.

Strilanc: The reasoning I'm using is that this operation should be symmetric in the sense that `f(a, b) == reverse(f(reverse(b), reverse(a))`.

Strilanc: I could rename it to something besides `raggedy_add` to possibly make this clearer. I can see that `a.****_add(b)` prepending things into `a` could  be surprising.

daxfohl: I see. I was thinking more the parameter would be something like `force_stop_when` and the values would be an enum either `rights_align` or `lefts_align`. 

One benefit of having `lefts_align` behavior would allow like if you have

circuit1a: k1 ops on q1
circuit1b: N-k1 ops on q1
circuit 2a: k2 ops on q2
circuit 2b: N-k2 ops on q2
circuit 3a: ...

and called `raggedy_add` on all of them, the resulting circuit would be a block of length N. I think the existing implementation would always end up staggered. IDK if that's a big deal or not.

Maybe call it `tetris_add`.

Strilanc: Renamed to `tetris_concat`

mpharrigan: I like the new name

balopat: @daxfohl - the block of length N behavior is what actually is happening in the current implementation with `stop_at_first_alignment=False`. 

```
from cirq import *

a,b,c = LineQubit.range(3)
c1 = Circuit(H(a), H(a), H(a))
c2 = Circuit(X(a), X(a))
c3 = Circuit(Y(b))
c4 = Circuit(Z(b),Z(b),Z(b),Z(b))

print(c1.tetris_concat(c2,c3,c4))
```

```
0: ───H───H───H───X───X───

1: ───Y───Z───Z───Z───Z───
```

Am I missing something? 

I like the name too! :)

cirq/circuits/circuit.py
```diff
@@ -1258,6 +1258,109 @@ def _json_dict_(self):
...
+            >>> f = cirq.Circuit.tetris_concat
+            >>> f(f(A, B), A) == f(A, f(B, A))
+            False
+            >>> len(f(f(f(A, B), A), B)) == len(f(f(A, f(B, A)), B))
+            False
+
+        Args:
+            circuits: The circuits to concatenate.
+            stop_at_first_alignment: Defaults to false. When true, the circuits
+                are never overlapped more than needed to align their starts or
```
balopat, line 1289:
```suggestion
                are never overlapped more than needed to align their starts (in case the left circuit is smaller) or
```

cirq/circuits/circuit.py
```diff
@@ -1258,6 +1258,109 @@ def _json_dict_(self):
...
+                to align their ends. When false, the smaller circuit can be
```
balopat, line 1290:
```suggestion
                to align their ends (in case the right circuit is smaller). When false, the smaller circuit can be
```

cirq/circuits/circuit.py
balopat: This "ends (or starts) align" was hard to understand for me I tried to add a bit more explanation when these cases happen.

cirq/circuits/circuit.py
```diff
@@ -1258,6 +1258,109 @@ def _json_dict_(self):
...
+    t = 0
+    while t < upper_bound:
+        if t < len(c2):
+            for op in c2[t]:
+                for q in op.qubits:
+                    # Record time but check if qubit already seen on other side.
+                    k2 = seen_times.setdefault(q, t)
+                    if k2 < 0:
+                        # Use this qubit collision to bound the collision time.
+                        upper_bound = min(upper_bound, t + ~k2)
```
balopat, line 1337: Using int inversion is a neat trick, but could use a bit of clarification comment maybe above when you explain the "complementary last used index in c1".

balopat (approved):
LGTM with some nits - this is a fun addition, thank you!
Thanks @daxfohl for the great review, it was super helpful!!!

cirq/circuits/circuit_test.py
```diff
@@ -4432,3 +4432,166 @@ def _circuit_diagram_info_(self, args):
...
+    f = cirq.Circuit.tetris_concat
+    assert len(f(space, ha)) == 10
+    assert len(f(space, ha, ha, ha)) == 10
+    assert len(f(space, f(ha, ha, ha))) == 10
+    assert len(f(space, ha, stop_at_first_alignment=True)) == 10
+    assert len(f(space, ha, ha, ha, stop_at_first_alignment=True)) == 12
+    assert len(f(space, f(ha, ha, ha, stop_at_first_alignment=True))) == 10
+
+    # L shape overlap (vary c1).
+    assert 7 == len(
```
balopat, line 4484: These almost look like parametrized tests...which is a better pattern than multiple assertions, if there are failures, it is easier to understand what broke compared to whack-a-mole with the multiple assertions. Would you mind rewriting it to parameterized tests?

cirq/circuits/circuit.py
daxfohl: @balopat it becomes jagged if you add in a different order, such as all first halves then all second halves. Neither option preserves the rectangle here:

```
from cirq import *

a, b, c = LineQubit.range(3)
c1a = Circuit(H(a), H(a), H(a))
c1b = Circuit(X(a), X(a), X(a))
c2a = Circuit(Y(b))
c2b = Circuit(Z(b), Z(b), Z(b), Z(b), Z(b))
c3a = Circuit(X(c), X(c), X(c), X(c), X(c))
c3b = Circuit(Y(c))

print(Circuit.tetris_concat(c1a, c2a, c3a, c1b, c2b, c3b, stop_at_first_alignment=False))
>> 0: ───────────H───H───H───X───X───X───
>>
>> 1: ───────────Y───Z───Z───Z───Z───Z───
>>
>> 2: ───X───X───X───X───X───Y───────────


print(Circuit.tetris_concat(c1a, c2a, c3a, c1b, c2b, c3b, stop_at_first_alignment=True))
>> 0: ───H───H───H───X───X───X───────────
>> 
>> 1: ───────────Y───Z───Z───Z───Z───Z───
>> 
>> 2: ───X───X───X───X───X───────────Y───
```

But if you change the code to initialize `upper_bound=len(c1)` the result is rectangular:
```
>> 0: ───H───H───H───X───X───X───
>> 
>> 1: ───Y───Z───Z───Z───Z───Z───
>> 
>> 2: ───X───X───X───X───X───Y───
```
This would mimic the "tetris" behavior. Though as Craig mentioned it's not symmetric. (A tetris block falling into a valley would fall all the way in, but if you did the same operation upside down, the arch (upside-down valley) would stop when its edges hit the ground, before its middle hits the block).

cirq/circuits/circuit.py
```diff
@@ -1258,6 +1258,109 @@ def _json_dict_(self):
...
+        Returns:
+            The concatenated and overlapped circuit.
+        """
+        if len(circuits) == 0:
+            return Circuit()
+        n_acc = len(circuits[0])
+
+        # Allocate a buffer large enough to append and prepend all the circuits.
+        pad_len = sum(len(c) for c in circuits) - n_acc
+        buffer = np.zeros(shape=pad_len * 2 + n_acc, dtype=np.object)
```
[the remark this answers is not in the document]
Strilanc: The numpy array is intentional, for performance. This is both because a) it has better constant factors and b) because adding together lists can have linear complexity per addition whereas by using the buffer I know for sure I'm not introducing unexpected quadratic behavior.

The constructor takes an iterable of anything that's moments or operations, and I pass it a slice of something that's moments or operations. I agree that when numpy type annotations become possible there will need to be an explicit cast.

cirq/circuits/circuit.py
Strilanc: That comment is at the start of the method.

cirq/circuits/circuit_test.py
Strilanc: My rule of thumb writing unit tests is that they should be very literal; minimize abstractions. Abstractions are chances to make mistakes and to obscure what is being tested. This does tend to give my tests a repetitive nature, and of course I break the rule all the time. This case is particularly relevant, because the reason I am being so explicit is because I am checking for off by one errors in min/max logic. If I write the test in a parameterized way, I will have to do the assertion using min/max logic prone to off-by-one errors. This correlates the test being wrong with the code being wrong, which is very bad.

Generally I parameterize when I have a large number of cases, or a variable number of cases (e.g. "all gates"), or the property being tested is decoupled from the values (e.g. "addition commutes"). Here the number of cases is small and fixed (5), and the cases are very tightly coupled with what is being tested (does the pole hit the backstop or not).

cirq/circuits/circuit.py
Strilanc: I added a comment noting the complement.

Strilanc:
This is me strongly agreeing with @balopat about @daxfohl doing a great review on this PR.

cirq/circuits/frozen_circuit.py
```diff
@@ -176,6 +176,14 @@ def _resolve_parameters_(
     ) -> 'FrozenCircuit':
         return self.unfreeze()._resolve_parameters_(param_resolver, recursive).freeze()
 
+    @functools.wraps(Circuit.tetris_concat)  # Inherit doc string.
```
maffoo, line 179: I think using `functools.wraps` overwrites the full signature, so mypy might think this returns `cirq.Circuit` instead of `cirq.FrozenCircuit`. Certainly, that's what happens at runtime:
```python
>>> inspect.signature(cirq.FrozenCircuit.tetris_concat)
<Signature (*circuits: 'cirq.AbstractCircuit', stop_at_first_alignment: bool = False) -> 'cirq.Circuit'>
```
If you just want the docstring, it'd be better to do:
```python
def tetris_concat(...):
    ...

tetris_concat.__doc__ = Circuit.tetris_concat.__doc__
```

maffoo: Or, write an explicit docstring that refers to `Circuit.tetris_concat`.

cirq/circuits/circuit.py
```diff
@@ -1258,6 +1258,111 @@ def _json_dict_(self):
     def _from_json_dict_(cls, moments, device, **kwargs):
         return cls(moments, strategy=InsertStrategy.EARLIEST, device=device)
 
+    def tetris_concat(
+        *circuits: 'cirq.AbstractCircuit', stop_at_first_alignment: bool = False
```
maffoo, line 1262: This `stop_at_first_alignment` argument strikes me as very odd. It would seem more natural to me to specify whether the overall alignment should be to push circuits as late or as early as possible, e.g. `align='start'` vs `align='end'`, or something like that. It would seem to me this makes it easier to predict the result without knowing which of two circuits happens to be longer.

PR #3810, opened 2021-02-16
Improve naming and docstrings of characterization methods

mrwojtek:
This change does one more rename of the characterization methods to name them more accurately. It also adds more docstrings with clarifications.


mrwojtek:
@balopat is Lint check broken?

cirq/google/calibration/workflow.py
```diff
@@ -501,7 +510,7 @@ def as_circuit(self) -> Circuit:
         return Circuit(self.operations)
 
 
-def run_floquet_characterization_for_circuit(
+def run_floquet_characterization_for_moments(
     circuit: Circuit,
```
mpharrigan, line 514: it's a little incongruous that this expects a `Circuit` but its name is "for moments"

mpharrigan (approved):

cirq/google/calibration/workflow.py
mrwojtek: It is, I have a feeling this is not the last change of those names...

PR #3811, opened 2021-02-16
pylint print version + verbose mode

balopat:
Enabling pylint version print in CI + verbose mode for more debugging information when pylint flakes occur.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,8 +68,10 @@ jobs:
           architecture: 'x64'
       - name: Install pylint
         run: cat dev_tools/conf/pip-list-dev-tools.txt | grep "pylint" | grep -v "#" | xargs pip install
+      - name: Display version
+        run: check/pylint --version
       - name: Lint
-        run: check/pylint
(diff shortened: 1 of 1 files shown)


95-martin-orion (approved):

.github/workflows/ci.yml
```diff
@@ -68,8 +68,10 @@ jobs:
           architecture: 'x64'
       - name: Install pylint
         run: cat dev_tools/conf/pip-list-dev-tools.txt | grep "pylint" | grep -v "#" | xargs pip install
+      - name: Display version
+        run: check/pylint --version
```
vtomole, line 72: What's wrong with pinning an exact version of pylint instead?

95-martin-orion: This is a fair point - the `pylint~=2.6.0` in `pip-list-dev-tools.txt` can get any version of the form `2.6.*`.

That being said, I've definitely seen discrepancies between pylint runs with the same pylint version. I don't think pinning it will resolve the flake.

PR #3812, opened 2021-02-16
Add option to do per-operation compensation

mrwojtek:
This is a simplified and less accurate version of per-moment compensation. It might be useful where the per-moment compensation is too expensive to run.


mpharrigan:
I'm a little confused at how this compares and contrasts with the other method(s). Can you expand a bit? I see the input is still a `circuit`, so where does the differing behavior come into effect?

mrwojtek:
This new method doesn't care if two operations in a compensated moment were characterized together or in a different moments. This is not the case for the other method.

I'll add more docstrings on this tomorrow.

mpharrigan:
I think I get it now: You could have done calibration any number of ways (either fully-grid-parallel, according to the circuit's moment structure, or completely randomly if you're feelin' wild) and this new function will take the results no matter what and match them up by pair whereas the previous method needs the "mapping" between circuit moment and calibration result. Is this right?

mrwojtek:
That is correct, I described this in a docstring now.

mpharrigan (approved):
Thanks for the clarification

Issue #3228, opened 2020-08-19
Packaging: Define core Cirq and packaging strategy


balopat:
**Problem**: Cirq has a growing number of third party integrations that are not core to the functioning of the library.  These lead to an increasing number of dependencies for a functionality that is only necessary for a small number of users.  In addition Cirq has not defined what its core concern is, which has led to a variety of situations where code exists in Cirq that does not solve a core concern problem.

**Rough requirements:**

- Define the core concern of Cirq and delineate the boundaries of this core.
- Create new packages that can be separately installed for third party integrations
- Keep the third party integration code in sync with Cirq’s current development build
- Allow for easy simultaneous release of core and non-core packages

PR #3813, opened 2021-02-16
Add prepare_floquet_characterization_for_operations function

mrwojtek:
This function scans the circuits in order to find at most 4 different moments to calibrate on grid devices. This is not as accurate as per-moment characterization but may be preferable because of the speed for some applications.


mrwojtek:
The tests are failing because of prohibited imports. Should they really be prohibited? In that case how to proceed with this PR? @balopat

95-martin-orion:
Import order in the top-level `__init__.py` determines import permissions: module A can only import module B if A appears after B in the `__init__.py` import list [1]. As `google` appears _before_ `experiments`, it cannot import from `experiments`:
https://github.com/quantumlib/Cirq/blob/98d950a9035048dc991d8bc84c7e2f53997c5fd5/cirq/__init__.py#L51-L54
The question of whether this _should_ be prohibited is closely related to the packaging strategy (#3228).

[1] There is a loophole here: if you move the `import` inside the function where it's used, you can ignore the import rules. IIUC, this is because the import will only occur when calling the function, so the infinite-import-loop is avoided.

balopat:
I pushed a quick fix to unblock you. We have to find a place for `generate_boixo_2018_supremacy_circuits_v2_bristlecone` method (deprecate? move to cirq.google?) as it refers to cirq.google while living in cirq proper (cirq.experiments) which should not happen. I'm in the process of removing cirq.google references from cirq (https://github.com/quantumlib/Cirq/issues/3737). It should be perfectly fine to import cirq.experiment from cirq.google however!

95-martin-orion:
> I pushed a quick fix to unblock you.

Think this will break `cirq/experiments/random_quantum_circuit_generation.py`, which imports `google` for access to `SYC`.

balopat:
> > I pushed a quick fix to unblock you.
> 
> Think this will break `cirq/experiments/random_quantum_circuit_generation.py`, which imports `google` for access to `SYC`.

I only see a reference in `random_quantum_circuit_generation_test.py` - but that should be fine from the import_test perspective as it doesn't get imported from the `cirq/__init__.py` process (I mean I will remove it eventually :)). 

Your analysis was great btw, that is exactly the reason: ordering of modules in `__init_.py` - with the packaging strategy changes, cirq.google should be able to be only at the bottom of the file, and eventually we can make it optional (only if cirq_google package is installed will it be imported).

mrwojtek:
Thanks a lot everyone, seems to work now.

mpharrigan (approved):
LGTM

PR #3815, opened 2021-02-16
Use concise JSON serialization for FrozenCircuit.

### LINKED ISSUE #3633: Efficient JSON serialization of FrozenCircuit

95-martin-orion:
Fixes #3633. This is a rewrite of #3655 that avoids using hash values in serialization; this is possible due to #3673 dynamically applying unique identifiers to each `SerializableByKey` object as it appears.

This PR continues to use FrozenCircuit hashes in circuit diagrams, where hash collision is a less serious issue.

### DIFF
--- a/cirq/circuits/frozen_circuit.py
+++ b/cirq/circuits/frozen_circuit.py
@@ -49,6 +49,7 @@ class FrozenCircuit(AbstractCircuit):
         *contents: 'cirq.OP_TREE',
         strategy: 'cirq.InsertStrategy' = InsertStrategy.EARLIEST,
         device: 'cirq.Device' = devices.UNCONSTRAINED_DEVICE,
+        name: Optional[str] = None,
     ) -> None:
         """Initializes a frozen circuit.
 
@@ -62,10 +63,13 @@ class FrozenCircuit(AbstractCircuit):
                 from `contents`, this determines how the operations are packed
                 together.
             device: Hardware that the circuit should be able to run on.
+            name: A user-specified name to identify this circuit in diagrams
+                and serialization.
         """
-        base = Circuit(contents, strategy=strategy, device=device)
+        base = Circuit(contents, strategy=strategy, device=device, name=name)
         self._moments = tuple(base.moments)
         self._device = base.device
+        self._name = base.name
 
         # These variables are memoized when first requested.
         self._num_qubits: Optional[int] = None
@@ -85,13 +89,19 @@ class FrozenCircuit(AbstractCircuit):
     def device(self) -> devices.Device:
         return self._device
 
-    def __hash__(self):
-        return hash((self.moments, self.device))
+    @property
+    def name(self) -> Optional[str]:
+        return self._name
 
-    def serialization_key(self):
-        # TODO: use this key in serialization and support user-specified keys.
+    def __hash__(self):
+        return hash((self.moments, self.device, self.name))
+
+    def _serialization_name_(self):
+        return self._name or 'Circuit'
+
+    def diagram_name(self):
         key = hash(self) & 0xFFFF_FFFF_FFFF_FFFF
-        return f'Circuit_0x{key:016x}'
+        return f'{self.name or "Circuit"}_0x{key:016x}'
 
     # Memoized methods for commonly-retrieved properties.
 
--- a/cirq/protocols/json_test_data/Circuit.json
+++ b/cirq/protocols/json_test_data/Circuit.json
@@ -116,7 +116,8 @@
     ],
     "device": {
       "cirq_type": "_UnconstrainedDevice"
-    }
+    },
+    "name": "group_measure"
   },
   {
     "cirq_type": "Circuit",
@@ -170,7 +171,8 @@
     ],
     "device": {
       "cirq_type": "_UnconstrainedDevice"
-    }
+    },
+    "name": null
   },
   {
     "cirq_type": "Circuit",
@@ -200,6 +202,7 @@
     ],
     "device": {
       "cirq_type": "_UnconstrainedDevice"
-    }
+    },
+    "name": null
   }
 ]
--- a/cirq/circuits/circuit.py
+++ b/cirq/circuits/circuit.py
@@ -110,6 +110,11 @@ class AbstractCircuit(abc.ABC):
     def device(self) -> devices.Device:
         pass
 
+    @property
+    @abc.abstractmethod
+    def name(self) -> Optional[str]:
+        pass
+
     def freeze(self) -> 'cirq.FrozenCircuit':
         """Creates a FrozenCircuit from this circuit.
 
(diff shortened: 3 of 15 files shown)


cirq/circuits/frozen_circuit.py
```diff
@@ -85,13 +89,19 @@ def moments(self) -> Sequence['cirq.Moment']:
...
 
+    @property
+    def name(self) -> Optional[str]:
+        return self._name
+
     def __hash__(self):
-        return hash((self.moments, self.device))
+        return hash((self.moments, self.device, self.name))
+
+    def _serialization_name_(self):
```
wcourtney, line 99: It doesn't look like this method is used outside of tests. Since it's "private", consider removing it (or using it).

cirq/circuits/frozen_circuit.py
```diff
@@ -85,13 +89,19 @@ def moments(self) -> Sequence['cirq.Moment']:
...
+        return self._name or 'Circuit'
 
-    def serialization_key(self):
-        # TODO: use this key in serialization and support user-specified keys.
+    def diagram_name(self):
```
wcourtney, line 102: Is this method still intended to be used for serialization? If so, consider a name that indicates this or add a new method that forwards functionality here. Alternatively, maybe a name that declares what this value is rather than how it's used?

cirq/circuits/frozen_circuit.py
```diff
@@ -85,13 +89,19 @@ def moments(self) -> Sequence['cirq.Moment']:
     def device(self) -> devices.Device:
         return self._device
 
+    @property
+    def name(self) -> Optional[str]:
+        return self._name
+
     def __hash__(self):
-        return hash((self.moments, self.device))
+        return hash((self.moments, self.device, self.name))
```
wcourtney, line 97: IIUC, this poses a risk that a user could add duplicate definitions of circuits to the serialization by providing them with unique names. Maybe it's future work, but should there be a check somewhere to prevent this?

cirq/circuits/frozen_circuit.py
95-martin-orion: This is a "magic method"; defining it here is what enables concise serialization of `FrozenCircuit`s in the JSON serialization protocol.

cirq/circuits/frozen_circuit.py
95-martin-orion: This method is only used in circuit diagrams now - hence the name. Serialization instead uses `_serialization_name_`.

cirq/circuits/frozen_circuit.py
95-martin-orion: We could check during serialization if two `FrozenCircuit`s are equal apart from their names, I guess - unless someone really likes reading JSON-formatted circuits, there probably isn't a use case for identical circuits with different names.

(Aside: thanks to #3673, hash values are not required for concise serialization)

cirq/circuits/frozen_circuit.py
wcourtney: Is it possible to reference the overridden method so that we can easily access its documentation? Should this class also extend [SerializableByKey](https://github.com/quantumlib/Cirq/blob/0677d6082bad22289c8438596dd7bf815c21e1c2/cirq/protocols/json_serialization.py#L332) to be explicit about this support?

cirq/circuits/frozen_circuit.py
```diff
@@ -85,13 +89,19 @@ def moments(self) -> Sequence['cirq.Moment']:
...
+
+    def _serialization_name_(self):
+        return self._name or 'Circuit'
 
-    def serialization_key(self):
-        # TODO: use this key in serialization and support user-specified keys.
+    def diagram_name(self):
         key = hash(self) & 0xFFFF_FFFF_FFFF_FFFF
-        return f'Circuit_0x{key:016x}'
+        return f'{self.name or "Circuit"}_0x{key:016x}'
```
wcourtney, line 104: Should `self.name or "Circuit"` be `self._serialization_name_()` for consistency, or do we want to keep them independent and they just happen to match today?

wcourtney (approved):
Mostly nits/optional suggestions at this point, so approving to let you resolve as you see fit.

cirq/circuits/frozen_circuit.py
95-martin-orion: I'm not fundamentally opposed to adding a reference here, but in general magic methods in Cirq don't reference their protocols. Instead, there's an implicit assumption that magic method usage is defined in the relevant protocol; this is supported by the [protocols doc](https://github.com/quantumlib/Cirq/blob/master/docs/protocols.ipynb).

cirq/circuits/frozen_circuit.py
95-martin-orion: It's plausible that we'll want to separate these in the future (e.g. for out-of-line subcircuits in circuit diagrams) - let's keep them independent.

95-martin-orion:
Thanks Will! It looks like I still need a cirq-maintainers review before merging this.

mpharrigan (dismissed):
1. You're modifying the json test data. The rule is to make new test data files with the new schema but ensure that the old files still load successfully. ~~Unless these classes were never in a released version of Cirq? I can't keep track of the timeline~~ ok `cirq.Circuit` has definitely been around for a while.

2. Adding a name to circuits is a pretty big deal :tm: because it opens the floodgates for users to use it in previously unpredicted ways. I personally think it's a fine idea, but maybe we should shop it around at the cirq cync

cirq/protocols/json_test_data/Circuit.json
```diff
@@ -170,7 +171,8 @@
     ],
     "device": {
       "cirq_type": "_UnconstrainedDevice"
-    }
+    },
+    "name": null
```
mpharrigan, line 175: consider omitting this field in the json output when it's not set for fully-backwards-compatible output and generally nicer json

95-martin-orion: Done.

95-martin-orion:
Thanks for catching the backwards-compatibility issue! Should be tested now with the `json_inward` files.

RE circuit names: I don't see much potential for misuse here, given that `Circuit.name` is only used in the serialization of `FrozenCircuit`, but I'm happy to delay until next cynque for confirmation.

mpharrigan:
I know right _now_ it's only used in serialization :)

95-martin-orion:
Cirq sync discussion: decouple addition of `name` from this PR

95-martin-orion:
The new, name-free version of this change is #3863.

Issue #2413, opened 2019-10-25
DensityMatrixSimulator().simulate() return value type annotation


mpharrigan:
Should be `DensityMatrixTrialResult`. Due to inheritance, it is `SimulationTrialResult` leading to problems if you actually try to use it

https://github.com/quantumlib/Cirq/pull/2308#issuecomment-546538095

Ashalynd:
Shouldn't there be some consistency? E.g. all simulate() methods should return SimulationTrialResult which then can be cast to the corresponding type. (Thought it seems odd too)

daxfohl:
@balopat This has been fixed for a while and can be closed.

mpharrigan:
Fixed by #3818

Issue #554, opened 2018-07-02
Add type annotations to method overrides in child classes


cduck:
Mypy hasn't caught that the implementation doesn't match the abstract method signature because they aren't currently type annotated.

Because of this, all gate types that implement both `ParameterizableGate` and `TextDiagrammableGate` currently have the wrong type signature for `text_diagram_exponent()`: `text_diagram_exponent() -> Union[value.Symbol, float]` vs. `text_diagram_exponent() -> float`.

dabacon:
Current abstract methods that should be checked.
```
cirq/circuits/optimization_pass.py:100:    @abc.abstractmethod
cirq/value/abc_alt_test.py:25:        @abc.abstractmethod
cirq/value/abc_alt_test.py:44:        @abc.abstractmethod
cirq/value/abc_alt_test.py:98:        @abc.abstractmethod
cirq/value/abc_alt_test.py:177:        @abc.abstractmethod
cirq/value/abc_alt_test.py:181:        @abc.abstractmethod
cirq/contrib/acquaintance/executor.py:47:    @abc.abstractmethod
cirq/contrib/acquaintance/executor.py:55:    @abc.abstractmethod
cirq/contrib/acquaintance/executor.py:60:    @abc.abstractmethod
cirq/contrib/acquaintance/permutation.py:50:    @abc.abstractmethod
cirq/contrib/graph_device/graph_device.py:32:    @abc.abstractmethod
cirq/contrib/graph_device/graph_device.py:36:    @abc.abstractmethod
cirq/interop/quirk/cells/cell.py:48:    @abc.abstractmethod
cirq/interop/quirk/cells/cell.py:64:    @abc.abstractmethod
cirq/ops/dense_pauli_string.py:341:    @abc.abstractmethod
cirq/ops/raw_types.py:45:    @abc.abstractmethod
cirq/ops/raw_types.py:55:    @abc.abstractmethod
cirq/ops/raw_types.py:375:    @abc.abstractmethod
cirq/ops/raw_types.py:389:    @abc.abstractmethod
cirq/ops/pauli_string_raw_types.py:43:    @abc.abstractmethod
cirq/ops/arithmetic_operation.py:84:    @abc.abstractmethod
cirq/ops/arithmetic_operation.py:102:    @abc.abstractmethod
cirq/ops/arithmetic_operation.py:117:    @abc.abstractmethod
cirq/ops/eigen_gate.py:229:    @abc.abstractmethod
[...]

PR #3818, opened 2021-02-17
Improve type safety with generics on simulators

daxfohl:


cirq/experiments/xeb_simulation.py
```diff
@@ -120,7 +120,9 @@ def simulate_2q_xeb_circuits(
         # Need an actual object; not np.random or else multiprocessing will
         # fail to pickle the closure object:
         # https://github.com/quantumlib/Cirq/issues/3717
-        simulator = sim.Simulator(seed=np.random.RandomState())
+        simulator = cast(
+            'cirq.SimulatesIntermediateState', sim.Simulator(seed=np.random.RandomState())
+        )
```
daxfohl, line 125: mypy complains here because `simulator` is `Option[SimulatesIntermediateState]` and it doesn't realize `Simulator` matches that. If anybody knows of a better option here then lmk.

cirq/sim/simulator.py
```diff
@@ -359,7 +366,11 @@ def simulate_sweep(
         raise NotImplementedError()
 
 
-class SimulatesIntermediateState(SimulatesFinalState, metaclass=abc.ABCMeta):
+class SimulatesIntermediateStateBase(
```
95-martin-orion, line 369: Why is a `*Base` type needed here, but not for the other newly-generic types?

cirq/experiments/xeb_simulation.py
95-martin-orion: It looks like `Simulator` is a `SimulatesIntermediateStateBase`, not a `SimulatesIntermediateState`:
https://github.com/quantumlib/Cirq/blob/4e1844e58ece3f4f7aa342e64520b4cd0b20bbe2/cirq/sim/sparse_simulator.py#L47-L49
https://github.com/quantumlib/Cirq/blob/4e1844e58ece3f4f7aa342e64520b4cd0b20bbe2/cirq/sim/state_vector_simulator.py#L33-L38
Depending on how the function uses `simulator`, you either need to update the type in the function or prohibit non-`SimulatesIntermediateState` from being passed in.

95-martin-orion (changes requested):
We've had some complaints about unclear simulator result types in the past - this should go a long way towards resolving those! Just a couple comments.

cirq/sim/simulator.py
daxfohl: Because `SimulatesIntermediateState` defines a `_create_simulator_trial_result` that returns a `SimulationTrialResult`. But some child classes return some child result type.

cirq/experiments/xeb_simulation.py
daxfohl: Doh! Fixed.

cirq/sim/simulator.py
95-martin-orion: Ah - it has type-specific implementation, but the others don't. That makes sense.

Should the other types then be `{Type}Base` for consistency? Or perhaps it's cleaner to leave the "Base" versions as-is, and rename the new `SimulatesIntermediateState` to `SimulatesIntermediateStateImpl`?

daxfohl: Probably the latter. And we may even be able to delete the `Impl` afterward because I think it's only used in a couple tests. What do you think?

95-martin-orion: That sounds good to me, although to preserve test coverage I'd suggest moving the `Impl` into the tests instead of deleting it entirely.

daxfohl: done

cirq/experiments/xeb_simulation.py
```diff
@@ -46,7 +45,7 @@ class _Simulate2qXEBTask:
 class _Simulate_2q_XEB_Circuit:
     """Closure used in `simulate_2q_xeb_circuits` so it works with multiprocessing."""
 
-    def __init__(self, simulator: 'cirq.SimulatesIntermediateState'):
+    def __init__(self, simulator: 'sim.SimulatesIntermediateState'):
```
95-martin-orion, line 48: Should be able to switch back to `cirq.{...}` here and below.

95-martin-orion (changes requested):
One final item to check: how does this interact with external simulators? e.g. will `qsim.QSimSimulator` break when this change is released, or can it remain a typeless `SimulatesFinalState` without issue?

cirq/experiments/xeb_simulation.py
daxfohl: Done

daxfohl: How to I test QSimSimulator?

95-martin-orion: Make sure you have both the latest version of Cirq (including this change) and `qsimcirq`, then run any `SimulatesFinalState` test with `QSimSimulator` in place of the Cirq simulator. See the [qsimcirq tests](https://github.com/quantumlib/qsim/blob/master/qsimcirq_tests/qsimcirq_test.py) for examples.

daxfohl: I'm befuddled. Installed qsim from source. Tried changing `test_intermediate_simulator` to use `QSimSimulator`, but it can't find `QSimSimulator`. How do I get the test to find `QSimSimulator`? Is this even what you mean?

daxfohl: Nvm, got it (I think): Added the following to beginning of `qsimcirq_test` in the qsim project:
```
import sys
sys.path.insert(1, '/home/dax/cirq')
```
Then ran those unit tests, and they passed. I updated my cirq code to throw an error just to verify it was being called, and the error was thrown, so confirms it was actually calling the new code.

Conclusion: the new class type declarations work fine with the external simulators.

95-martin-orion (approved):
Thanks for going the extra mile to verify - this is good to go!

mpharrigan:
I think there are some open issues complaining about some of these things? Can you find them and see if this addresses them?

daxfohl:
@mpharrigan There was only #554 about the general problem from a while ago. Added a link there but there's still more work to be done before closing that particular issue. There are some other open type issues around gates and operations, parameters, and such but I didn't see any open issues referring specifically to simulators.

mpharrigan:
Does this PR address https://github.com/quantumlib/Cirq/issues/2413?

daxfohl:
Oh, I didn't see that one. Yes, that should be fixed now.

PR #3819, opened 2021-02-17
[Docs] Patch multiplicative group in Shor tutorial

### LINKED ISSUE #3814: Error of multiplicative_group function in Shor.ipynb tutorial
**Description of the issue**
The implementation of multiplicative_group function in [Shor.ipynb](https://quantumai.google/cirq/tutorials/shor) tutorial is wrong. 

The code in the notebook is
```
"""Function to compute the elements of Z_n."""
def multiplicative_group(n: int) -> List[int]:
    """Returns the multiplicative group modulo n.

    Args:
        n: Modulus of the multiplicative group.
    """
    assert n > 2
    group = [1, 2]
    for x in range(3, n):
        if math.gcd(x, n) == 1:
            group.append(x)
    return group
```
Here `group`  initialized as `[1, 2]`. However, element 2 should not be included always. For example, when `n = 6`, the result should be `[1,5]`. It is easy to fix it by changing the function into 
```
    ...
    group = [1]
    for x in range(2, n):
          ...
```
There is no propagation of error since this function is not used afterward. But it is not good to keep the wrong implementation.

**Cirq version**
Irrelevant.

rmlarose:
Fixes #3814. Lower priority, but the multiplicative group function (which is only used pedagogically) contained 2 sometimes when it shouldn't.

Also clears outputs so this is tested.


docs/tutorials/shor.ipynb
```diff
@@ -1201,21 +1062,11 @@
...
+   "execution_count": null,
    "metadata": {
     "id": "1RN4SqfDYAsh"
    },
-   "outputs": [
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "Factoring n = pq = 184573\n",
```
95-martin-orion, line 1213: What is the runtime of this factorization? If it's more than ~10 seconds, we may want to keep the outputs here to prevent delays in CI tests.

docs/tutorials/shor.ipynb
```diff
@@ -171,9 +171,9 @@
     "    Args:\n",
     "        n: Modulus of the multiplicative group.\n",
     "    \"\"\"\n",
-    "    assert n > 2\n",
-    "    group = [1, 2]\n",
```
95-martin-orion, line 175: [no change needed] I suspect this was built using the common assumption for Shor's algorithm that `n` is an odd number, but the new code is more generally correct. The alternative would be to add a comment explaining the Shor-based optimization being done here.

95-martin-orion (changes requested):
Top-level comment: do not merge until after Cirq v0.10.0 release.

docs/tutorials/shor.ipynb
rmlarose: It's fast! (due to classical order finder)

```
CPU times: user 2 µs, sys: 1e+03 ns, total: 3 µs
Wall time: 5.72 µs
Factoring n = pq = 184573
p = 379
q = 487
```

95-martin-orion (approved):

PR #3821, opened 2021-02-17
Add circuit Alignment and use it for zip and tetris_concat

maffoo:
This changes the behavior of `Circuit.tetris_concat` to take an `align` parameter that can be either `START` or `END`, and which does what it says when concatting circuits that do not share any qubits: aligns their starting or ending moments, regardless of which circuit is longer (previously the behavior of the `stop_at_first_alignment` flag would result in different alignment depending on whether the first or second circuit was longer). This same alignment scheme also applies when zipping circuits, so I've added it as an option there (still need to add tests for that).


cirq/circuits/circuit.py
```diff
@@ -1310,22 +1392,20 @@ def tetris_concat(
...
-    c1: Sequence['cirq.Moment'], c2: Sequence['cirq.Moment'], stop_at_first_alignment: bool
+    c1: Sequence['cirq.Moment'], c2: Sequence['cirq.Moment'], align: 'cirq.Alignment'
 ) -> int:
     # Tracks the first used moment index for each qubit in c2.
     # Tracks the complementary last used moment index for each qubit in c1.
     seen_times: Dict['cirq.Qid', int] = {}
 
     # Start scanning from end of first and start of second.
-    upper_bound = (min if stop_at_first_alignment else max)(len(c1), len(c2))
+    upper_bound = len(c1) if align == Alignment.START else len(c2)
```
daxfohl, line 1408: Note that in the case of Alignment.END, and len(c2) > len(c1), and they don't intersect, then this implementation pushes c2 to start before c1. 

If you want to avoid this behavior, we need `len(c1) if align == Alignment.START else min(len(c1), len(c2))`. This creates the behavior where:

* Alignment.START is tetris-style.
* Alignment.END is tetris-style *except* it doesn't allow the last moment of c2 to push past the last moment of c1. It ensures c2 stays at the end of the resulting circuit.

I *think* these are the two most likely required use cases.

maffoo: In that case I _would_ expect c2 to start before c1, since I specified that the ends should align. I think the code as-is handles all the cases:
```python
In [1]: import cirq

In [2]: a, b = [cirq.X(q) for q in cirq.LineQubit.range(2)]

In [3]: cirq.Circuit.tetris_concat(cirq.Circuit([a] * 5), cirq.Circuit([b] * 3), align='END')
Out[3]: 
0: ───X───X───X───X───X───

1: ───────────X───X───X───

In [4]: cirq.Circuit.tetris_concat(cirq.Circuit([a] * 5), cirq.Circuit([b] * 3), align='START')
Out[4]: 
0: ───X───X───X───X───X───

1: ───X───X───X───────────

In [5]: cirq.Circuit.tetris_concat(cirq.Circuit([a] * 3), cirq.Circuit([b] * 5), align='START')
Out[5]: 
0: ───X───X───X───────────

1: ───X───X───X───X───X───

In [6]: cirq.Circuit.tetris_concat(cirq.Circuit([a] * 3), cirq.Circuit([b] * 5), align='END')
Out[6]: 
0: ───────────X───X───X───

1: ───X───X───X───X───X───
```

balopat: +1 to @maffoo's direction here - I think having a "tetris" stop always at the start is more confusing to think about. Aligning ends or starts is just very simple.

cirq/circuits/circuit.py
```diff
@@ -66,6 +67,14 @@
 INT_TYPE = Union[int, np.integer]
 
 
+class Alignment(enum.Enum):
```
balopat, line 70: We have `cirq.AlignLeft` and `cirq.AlignRight` optimizers. 
1. Do you think it would be useful to rename these `LEFT` and `RIGHT`? (although this assumes no transposing) If not, do you think we should rename those optimizers as well? 
2. I have this inkling that in the cirq namespace, we don't know the context of this generic term "alignment" - maybe we should name this `cirq.CircuitAlignment`?

balopat (approved):
LGTM besides some naming thoughts

cirq/circuits/circuit.py
maffoo: Changed to `LEFT` and `RIGHT`. I think it makes sense to match that terminology. I think I'll leave it as `cirq.Alignment` to match `cirq.AlignLeft` and `cirq.AlignRight` which are similarly generic terms (we could even have a factory function that takes an `Alignment` and returns the appropriate optimizer if we wanted).

PR #3823, opened 2021-02-17
Fix _measurement_keys_ in repeated CircuitOperation and update the separator

smitsanghavi:
The `measurement_keys` protocol is expected to be consistent with the final measurement keys. Therefore it needs to prefix the `repetition_ids`. It also needs to remap the keys which are always in unindexed form. Added a few small helper methods and aliased the measurement key separator for cleaner handling.


95-martin-orion (changes requested):
My primary concern with this PR is that with measurement keys that already contain the separator symbol (e.g. `cirq.measure(a, key='test-measure')`) will misbehave. Thoughts on how to mitigate this?

cirq/circuits/circuit_operation.py
```diff
@@ -31,6 +31,7 @@
 
 
 INT_TYPE = Union[int, np.integer]
+measurement_key_separator = '-'
```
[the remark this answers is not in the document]
smitsanghavi: Done

smitsanghavi:
> My primary concern with this PR is that with measurement keys that already contain the separator symbol (e.g. `cirq.measure(a, key='test-measure')`) will misbehave. Thoughts on how to mitigate this?

Yes, I added a warning for now. This is unavoidable as long as we have only `str` keys. The best we can do is migrate away from this very common character `-` to something obtuse (WDYT about `|`?) that has less chance of appearing as a key name. We can also add some validation in `MeasurementGate` (any other places that take keys without going through `MeasurementGate`?) to disallow the separator character in the key.

95-martin-orion:
> We can also add some validation in `MeasurementGate` (any other places that take keys without going through `MeasurementGate`?) to disallow the separator character in the key.

In theory someone could include "test-measure" in the `CircuitOperation` mapping...this feels fragile. Since the issues only occur on `CircuitOperation`s, which are technically immutable, I propose adding validation to the `CircuitOperation` constructor that prevents it from containing measurements with keys that contain the separator.

smitsanghavi:
Sounds good. To go a step further, I'll only validate the direct child measurements of repeated `CircuitOperations`.
Should I also update the separator to some other character?

cirq/circuits/circuit_operation_test.py
```diff
@@ -81,6 +81,27 @@ def test_repetitions_and_ids_length_mismatch():
...
+    with pytest.raises(ValueError, match='invalid key: m-a'):
+        _ = cirq.CircuitOperation(cirq.FrozenCircuit(c_op), measurement_key_map={'m': 'm-a'})
+
+    # Originally invalid key
+    circuit = cirq.FrozenCircuit(cirq.measure(a, key='m-a'))
+    with pytest.raises(ValueError, match='invalid key: m-a'):
+        _ = cirq.CircuitOperation(circuit)
+
+    # Remapped to valid key
+    _ = cirq.CircuitOperation(circuit, measurement_key_map={'m-a': 'ma'})
```
95-martin-orion, line 102: [No change needed] The fact that this works is a huge relief - this means users can reuse old, "bad-format" circuits without having to modify those circuits directly.

cirq/circuits/circuit_operation.py
```diff
@@ -123,6 +125,28 @@ def __post_init__(self):
...
+        for _, op in self.circuit.findall_operations(
+            lambda op: not isinstance(op, CircuitOperation) and protocols.is_measurement(op)
+        ):
+            for key in protocols.measurement_keys(op):
+                key = self.measurement_key_map.get(key, key)
+                if MEASUREMENT_KEY_SEPARATOR in key:
+                    raise ValueError(
+                        f'Measurement {op} found to have invalid key: {key}. '
+                        f'"{MEASUREMENT_KEY_SEPARATOR}" is not allowed for measurement keys '
+                        'in a CircuitOperation'
```
95-martin-orion, line 148: Let's add a suggestion here for the user to manually remap to valid keys.

95-martin-orion (approved):
> Should I also update the separator to some other character?

Updating separators automatically is liable to cause other issues - instead, let's suggest to the user that they manually update any offending keys. (Automatically generating _suggested_ keys should be fine.)

Otherwise, this is good to go! Thanks for the fast turnaround on this.

cirq/circuits/circuit_operation.py
smitsanghavi: Done

cirq/circuits/circuit_operation_test.py
smitsanghavi: Yeah, that was my main concern given that our separator is way too commonly part of regular measurement keys :)

smitsanghavi:
> > Should I also update the separator to some other character?
> 
> Updating separators automatically is liable to cause other issues 

I wasn't suggesting automatically updating on the fly. I'm just saying that `-` might have been a bad syntactical choice originally. I should probably just change the separator we use for repeated keys to a rarely used char (`|`or `:` or wrapping in `[]` for array-like syntax) and do away with `-` altogether.

> - instead, let's suggest to the user that they manually update any offending keys. (Automatically generating _suggested_ keys should be fine.)
> 
> Otherwise, this is good to go! Thanks for the fast turnaround on this.

PR #3824, opened 2021-02-17
Move google noisemodel to cirq.google

balopat:
Moves cirq.google references out of cirq.contrib.noise_model.

- experimental Google noise models -> `cirq/google/experimental`
- other experimental noise models -> `cirq/contrib/noise_models`
- code required by both -> `cirq/devices`

Related to https://github.com/quantumlib/Cirq/issues/2832 but motivated by https://github.com/quantumlib/Cirq/issues/3737. Also related to the Packaging strategy #3228 - taking contrib apart.


cirq/devices/noise_model.py
```diff
@@ -264,3 +264,19 @@ def noisy_moment(
     `cirq.ConstantQubitNoiseModel`.
     """,
 )
+
+
+def validate_all_measurements(moment: 'cirq.Moment') -> bool:
```
mpharrigan, line 269: do we want this to be a public function?

balopat: totally up for discussion. It is used in both places (cirq.google and cirq.contrib) - we either make it public, duplicate it, or make it private but still call into it? It's not imported into the cirq namespace so it is somewhat "package private".

cirq/google/experimental/noise_models.py
```diff
@@ -0,0 +1,160 @@
+# Copyright 2019 The Cirq Developers
```
95-martin-orion, line 1: Do we expect `cirq/google/experimental` to be small enough that `noise_models` doesn't need a subdirectory?

cirq/google/json_test_data/spec.py
```diff
@@ -7,7 +7,7 @@
 
 TestSpec = ModuleJsonTestSpec(
     name="cirq.google",
-    packages=[cirq.google],
+    packages=[cirq.google, cirq.google.experimental],
```
95-martin-orion, line 10: Should whether `cirq.google` passes CI rely on experimental objects passing JSON serialization? (Not opposed, just making sure)

95-martin-orion (changes requested):
My read of this PR:
- experimental Google noise models -> `cirq/google/experimental`
- other experimental noise models -> `cirq/contrib/noise_models`
- code required by both -> `cirq/devices`

The shared code seems generally useful, and the rest is all going to experimental-type directories, so this mostly looks fine to me.

cirq/google/experimental/noise_models.py
balopat: I guess we can keep the structure, but yeah I'm not sure how big experimental is going to grow to. I'll create the folder.

cirq/google/json_test_data/spec.py
balopat: Good question, I have a slight preference for yes, make the decisions earlier rather than later. This ensures that when we "graduate" a module to the parent package, we wouldn't have to do a massive json rework.

balopat:
> My read of this PR:
> 
> * experimental Google noise models -> `cirq/google/experimental`
> * other experimental noise models -> `cirq/contrib/noise_models`
> * code required by both -> `cirq/devices`
> 
> The shared code seems generally useful, and the rest is all going to experimental-type directories, so this mostly looks fine to me.

Your read is so spot on, I'll update my PR description with it :)

95-martin-orion (approved):

PR #3827, opened 2021-02-17
Ensure only measured subcircuits cause full simulator sweeps

daxfohl:
Fixes a perf issue that *any* subcircuit would cause full sweep/repeat, even if it was unmeasured. With this change, at least only measured subcircuits will cause full sweep. (It does not however optimize for the case where such a subcircuit itself only contains terminal measurements, which is a deeper problem).

This PR also generalizes `split_into_unitary_then_general` to allow for splitting by any protocol.

@95-martin-orion


95-martin-orion (approved):
Thanks for following up on this! In a future PR, it may be possible to combine the circuit-splitting method here with `mapped_circuit` from #3785 to recursively split `CircuitOperation`s as well.

cirq/sim/density_matrix_simulator.py
```diff
@@ -168,8 +168,11 @@ def _run(
         resolved_circuit = protocols.resolve_parameters(circuit, param_resolver)
         check_all_resolved(resolved_circuit)
 
-        if circuit.are_all_measurements_terminal() and not any(
-            circuit.findall_operations(lambda op: isinstance(op, circuits.CircuitOperation))
+        unmeasured_prefix, general_suffix = split_into_matching_protocol_then_general(
+            resolved_circuit, lambda op: not protocols.is_measurement(op)
+        )
+        if len(general_suffix) <= 1 and not any(
+            general_suffix.findall_operations(lambda op: isinstance(op, circuits.CircuitOperation))
```
95-martin-orion, line 175: Almost missed this: applying the previous `circuit`-based check to `general_suffix` should work here. `general_circuit` can be longer than one moment as long as all measurements are terminal and there are no `CircuitOperation`s.

95-martin-orion (changes requested):
Sorry, caught something at the last second.

daxfohl:
Will change it when I get back, but actually don't see how it would happen,
given the "split" implementation.

cirq/sim/density_matrix_simulator.py
daxfohl: fixed

95-martin-orion (approved):

PR #3834, opened 2021-02-18
[XEB] Generate calibrations from circuits

mpharrigan:
In addition to the traditional `GridInteractionLayer`-based approach, allow doing parallel xeb where the pairs are defined 1) explicitly as a list of pairs 2) as a layer_circuit where each moment is just two qubit operations that give you the pairs.


cirq/experiments/random_quantum_circuit_generation.py
```diff
@@ -289,16 +294,68 @@ def _get_active_pairs(graph: nx.Graph, grid_layer: GridInteractionLayer):
...
     combinations: np.array
-    pairs: List[Tuple['cirq.Qid', 'cirq.Qid']]
+    pairs: List[QidPairT]
+
+
+def _get_random_combinations(
+    n_library_circuits: int,
+    n_combinations: int,
+    *,
+    pair_gen: Generator[Tuple[List[QidPairT], Any], None, None],
```
mrwojtek, line 316: Why limit yourself to Generator? Iterable is broader and applicable to your code as well.

cirq/experiments/random_quantum_circuit_generation.py
```diff
@@ -289,16 +294,68 @@ def _get_active_pairs(graph: nx.Graph, grid_layer: GridInteractionLayer):
...
+            themselves. The more combinations, the more precise of an estimate for XEB
+            fidelity estimation, but a corresponding increase in the number of circuits
+            you must sample.
+        pair_gen: A generator that yields tuples of (pairs, layer_meta) where pairs is a list
+            of qubit pairs and layer_meta is additional data describing the "layer" assigned
+            to the CircuitLibraryCombination.layer field.
+        random_state: A random-state-like object to seed the random combination generation.
+
+    Returns:
+        A list of `CircuitLibraryCombination`, each corresponding to a layer
```
mrwojtek, line 341: Not quite, you skip the layers with no pairs.

cirq/experiments/random_quantum_circuit_generation.py
```diff
@@ -339,19 +396,149 @@ def get_random_combinations_for_device(
...
+
+def _pairs_from_moment(moment: 'cirq.Moment') -> List[QidPairT]:
+    """Helper function in `get_random_combinations_for_layer_circuit` pair generator.
+
+    The moment should contain only two qubit operations, which define a list of qubit pairs.
+    """
+    pairs: List[QidPairT] = []
+    for op in moment.operations:
+        if len(op.qubits) != 2:
+            raise ValueError("Layer circuit contains >2-qubit operations.")
```
mrwojtek, line 465: It's also 1-qubit, right?

cirq/experiments/random_quantum_circuit_generation_test.py
```diff
@@ -148,6 +151,87 @@ def test_get_random_combinations_for_small_device():
...
+    with pytest.raises(ValueError, match=r'>2-qubit operation'):
+        _ = get_random_combinations_for_layer_circuit(
+            n_library_circuits=3, n_combinations=4, layer_circuit=circuit, random_state=99
+        )
+
+
+def test_get_grid_interaction_layer_circuit():
+    graph = _gridqubits_to_graph_device(cirq.GridQubit.rect(3, 3))
+    layer_circuit = get_grid_interaction_layer_circuit(graph)
+    cirq.testing.assert_has_diagram(
```
mrwojtek, line 210: Out of curiosity, isn't it better to compare against a manually constructed expected circuit?

cirq/experiments/random_quantum_circuit_generation.py
mpharrigan: Yeah good point. I moved this check to `get_random_combinations_for_device` where it is relevant (for a given device topology sometimes one of the four grid interaction layers doesn't apply, especially for small or narrow devices). I added a note to the return docstring for that function.

The other `get_random_combinations_xxx` functions should have a 1:1 correspondence. The user shouldn't be passing in empty moments or an empty list of pairs.

cirq/experiments/random_quantum_circuit_generation.py
mpharrigan: Good catch, fixed.

cirq/experiments/random_quantum_circuit_generation_test.py
mpharrigan: yeah, you're right. I put in a manually constructed one with comment annotations about why we'd expect it to be that way

mpharrigan: I also added a check that get_random_combinations_for_layer_circuit gives the same result as get_random_combinations_for_device when using get_grid_interaction_layer_circuit

mpharrigan:
Back at you @mrwojtek

mrwojtek (approved):

cirq/experiments/random_quantum_circuit_generation.py
```diff
@@ -289,16 +294,65 @@ def _get_active_pairs(graph: nx.Graph, grid_layer: GridInteractionLayer):
...
 
-    `layer` is used for record-keeping but is otherwise unnecessary. Therefore, it is optional.
+    `layer` is used for record-keeping. This is the GridInteractionLayer if using
+    `get_random_combinations_for_device`, the Moment if using
+    `get_random_combinations_for_layer_circuit` and ommitted if using
+    `get_random_combinations_for_pairs`.
     """
 
-    layer: Optional[GridInteractionLayer]
+    layer: Optional[Any]
```
dstrain115, line 307: Instead of Any, should we list the types as a Union?  It is only three different possibilities, right?

cirq/experiments/random_quantum_circuit_generation.py
```diff
@@ -289,16 +294,65 @@ def _get_active_pairs(graph: nx.Graph, grid_layer: GridInteractionLayer):
...
+    n_library_circuits: int,
+    n_combinations: int,
+    *,
+    pair_gen: Iterator[Tuple[List[QidPairT], Any]],
+    random_state: 'cirq.RANDOM_STATE_OR_SEED_LIKE' = None,
+) -> List[CircuitLibraryCombination]:
+    """For qubit pairs, prepare a set of combinations to efficiently sample
+    parallel two-qubit XEB circuits.
+
+    This helper function should be called by one of `xx`, `xx` which define
```
dstrain115, line 322: What does "one of `xx`" mean?  Should you just remove that part and make it:  "This helper function is called by methods that define the appropriate pair_gen arguments"

cirq/experiments/random_quantum_circuit_generation.py
mpharrigan: I did this suggestion. Don't know why my reply didn't post

cirq/experiments/random_quantum_circuit_generation.py
mpharrigan: Yes, I thought about this. The strongest argument for `Any` is that someone might want to write a new wrapper for `_get_random_combinations` with a custom `pair_gen` iterable. I wanted to allow the second item yielded to be any metadata that might be useful for provenance of the particular layer. It should never actually be consumed by the parallel XEB functionality. In some sense, having this broad type will make sure no one relies on this field being one of the couple of types the present functionality may result in. 

I suppose it will always be backwards compatible to broaden the scope of a type by having it be a union now and a bigger union later as applicable, and an `Any` if it comes to that. What do you think given this additional context?

cirq/experiments/random_quantum_circuit_generation.py
mpharrigan: Oh man good spot. I wrote this part of the docstring before I had finalized the wrapper methods and probably meant to go back and fix it :doh:

cirq/experiments/random_quantum_circuit_generation.py
mpharrigan: gonna go ahead and merge now so I can open the next PR. Feel free to follow-up and we can make the change there or elsewhere if desired

PR #3835, opened 2021-02-18
[Docs] Add support page for issues, requests, and questions

rmlarose:
To increase user feedback and make it clear how to report issues on the website.

cc @balopat @lamberta @karlunho 

Note: Cirq is the only repo with issue templates.


docs/_book.yaml
```diff
@@ -68,6 +68,11 @@ upper_tabs:
       - title: "Operators and observables"
         path: /cirq/operators_and_observables
 
+      ####  Support  ####
+      - heading: "Support"
+      - title: "Issues / requests / questions"
```
lamberta, line 73: Do you anticipate additional pages? Best to avoid single entry sections, if possible. Can add to the "Contributor guide" below until then

rmlarose: Good point, there probably won't be additional pages so the single-entry section may be overkill. Then again, it does make this page stick out which I think is good. What do others think?

mpharrigan: -1 new section

vcatherine: My view is that we should not add this under contributor guide, since the point of adding this is to encourage more feedback from users outside of the contributor community. If we do not want to add a section I would suggest adding this at the bottom of the "further topics" page

docs/support.md
```diff
@@ -0,0 +1,20 @@
...
+To report an issue or request a feature in our open-source projects, 
+first check if there is already an open issue for your bug report / 
+feature request at the links above. If not, file an issue by following 
+the provided template and/or contributing guidelines. Thanks for helping
+improve open-source quantum software!
+
+# Questions
+
+If you're not sure whether a certain behavior is buggy, or if you have other questions,
+please ask us in the [Cirq Gitter channel](https://gitter.im/cirqdev).
```
dstrain115, line 20: Should we link to any email lists, such as cirq-dev?  Eventually we should probably link to QCS support email lists, but maybe not yet.

rmlarose: I like this idea. I'll add a link to cirq-dev.

docs/_book.yaml
rmlarose: Moved to further topics!

balopat (approved):
LGTM

PR #3837, opened 2021-02-19
patch fixing a serialization issue with rx,ry and rz gates.

mganahl:
The current serialization of parametrized circuits misses the case where a sympy object is a numeric constant, as for exampole sympy.pi. This problem arises for parametrized rx, ry and rz gates

This patch is an attempt at fixing this.


cirq/json_resolver_cache.py
```diff
@@ -153,5 +153,8 @@ def two_qubit_matrix_gate(matrix):
         'sympy.Float': lambda approx: sympy.Float(approx),
         'sympy.Integer': sympy.Integer,
         'sympy.Rational': sympy.Rational,
+        'sympy.pi': lambda: sympy.pi,
```
dstrain115, line 156: Could we add a test (or modify one of the existing tests) to use one of these symbols?

Thanks for contributing!

mganahl: NP, will add it right away!
Looping in @jacksonwb, this PR should fix the serialization issues with parametrized circuits

mganahl: @dstrain115 added some test, let me know what you think!

mganahl:
@dstrain115 ping for review, thanks!

mganahl:
@vtomole fixed the test!

cirq/json_resolver_cache.py
vtomole: This comment has been addressed.

vtomole (approved):
LGTM. Thanks!

PR #3841, opened 2021-02-19
Density Matrix Simulator ActOn migration

### LINKED ISSUE #3825: Clean up disparity in simulator implementation styles
Sparse and Clifford both have a proper `ActOnXXXStateArgs` where they maintain simulator state, and those are explicitly checked throughout the cirq codebase with `isinstance` checks that alter behavior of different operations depending on the args. This was first added in #3019 with the `act_on` protocol.

DensityMatrix and MPS both have self-contained state (_StateAndBuffers and MPSState, respecively).

Should we change it to be consistent?

cc @Strilanc

daxfohl:
Fixes #3825

Sparse and Clifford simulators both have a proper ActOnXXXStateArgs where simulator state is maintained and acted upon by various operations. This was first added in #3019 with the act_on protocol.

DensityMatrix and MPS simulators were never updated with the new protocol, leaving the codebase inconsistent and the migration unfinished. This PR finishes the task for DensityMatrix. A separate PR for MPS simulator hopefully will follow.

Aditionally, this PR introduces an `ActOnArgs` base class to eliminate the code duplication and type checking that was required in `MeasurementGate` (it had grown to four `if isinstance` blocks, each with duplicate bitshift/logging code), and replace it all with a single function call back into the `args` object.


cirq/sim/act_on_density_matrix_args.py
```diff
@@ -0,0 +1,134 @@
...
+
+    1. Directly edit the `target_tensor` property, which is storing the density
+        matrix of the quantum system as a numpy array with one axis per qudit.
+    2. Overwrite the `available_buffer` property with the new state vector, and
+        then pass `available_buffer` into `swap_target_tensor_for`.
+    3. Call `record_measurement_result(key, val)` to log a measurement result.
+    """
+
+    def __init__(
+        self,
```
daxfohl, line 37: A good chunk of this was copied straight from state vector args. Probably worth making a base class. That would allow some deduplication in e.g. MeasurementGate too.

daxfohl: Initial thought is that would be better as a separate PR.

daxfohl: Nvm, I went ahead and did it here, it wasn't much trouble.

daxfohl:
@tonybruguier here is a template if you want to do the same for mps

cirq/sim/act_on_args.py
```diff
@@ -0,0 +1,82 @@
...
+
+from typing import Any, Iterable, Dict, Sequence
+
+import numpy as np
+
+from cirq import protocols, ops
+from cirq.protocols.decompose_protocol import _try_decompose_into_operations_and_qubits
+
+
+class ActOnArgs:
```
balopat, line 24: Why do we prefer inheritance here over composition?

daxfohl: We get to eliminate the duplicate bit shift and logging code in MeasurementGate because they are all instances of ActOnArgs

tanujkhattar:
Can you please add a detailed summary of the change in the PR description?

For someone with lesser context, either a reviewer or someone revisiting this PR later via git blame, it's easier to follow the code if there's a summary that we can go through. 

Unless the code is very trivial or there's a detailed design discussion elsewhere (eg: RFC), I think we should have detailed summaries of the change in the PR.

daxfohl:
@tanujkhattar I updated the description, can you TAL?

cirq/sim/density_matrix_simulator.py
```diff
@@ -278,74 +249,43 @@ def _base_iterator(
...
-                potential_op.gate, ops.MeasurementGate
-            )
-
         noisy_moments = self.noise.noisy_moments(circuit, sorted(circuit.all_qubits()))
-
+        measured = collections.defaultdict(bool)  # type: Dict[Tuple[cirq.Qid, ...], bool]
         for moment in noisy_moments:
-            measurements = collections.defaultdict(list)  # type: Dict[str, List[int]]
-
-            channel_ops_and_measurements = protocols.decompose(
```
balopat, line 308: Why did you remove this decomposition step and `on_stuck` and `keep`? I am surprised that this wasn't covered by tests -- but it seems that we lost the useful error message above.

E.g.: 

```
import cirq

class TestGate(cirq.Gate):
    def _num_qubits_(self) -> int:
        return 1

a, b, c = cirq.LineQubit.range(3)

circuit = cirq.Circuit(cirq.X(a), TestGate()(b), cirq.measure(c))
print(circuit)

cirq.DensityMatrixSimulator().run(circuit)
```

would fail in master with:

```
TypeError: Can't simulate operations that don't implement SupportsUnitary, SupportsConsistentApplyUnitary, SupportsMixture, SupportsChannel or is a measurement: <__main__.TestGate object at 0x7fea86010610>.on(cirq.LineQubit(1))
```

However it fails with a significantly more obscure message on this branch: 

```
TypeError: Failed to act action on state argument.
Tried both action._act_on_ and args._act_on_fallback_.

State argument type: <class 'cirq.sim.act_on_density_matrix_args.ActOnDensityMatrixArgs'>
Action type: <class 'cirq.ops.gate_operation.GateOperation'>
Action repr: <__main__.TestGate object at 0x7fc15560ce50>.on(cirq.LineQubit(1))
```

balopat (changes requested):
Thanks for this - I like this direction, the code will be significantly cleaner this way. Added a bunch of nits / comments.

cirq/sim/act_on_args.py
```diff
@@ -0,0 +1,89 @@
...
+            raise ValueError(f"Measurement already logged to key {key!r}")
+        self.log_of_measurement_results[key] = corrected
+
+    @abc.abstractmethod
+    def perform_measurement(self) -> List[int]:
+        """Child classes that perform measurements should implement this with
+        the implementation."""
+
+
+def strat_act_on_from_apply_decompose(
```
smitsanghavi, line 68: This strategy can potentially modify the internal state halfway just to eventually fail when it encounters a decomposed op that isn't supported. This is a valid flow and we should be able to move on to the next strategy when the above happens. But having modified the state partially leads to a bug and wrong outputs.

Add a way to restore the internal state to the starting value when this happens. #3874 has that implemented but I will be closing that PR in favor of this one. Feel free to pick up that logic.

smitsanghavi (changes requested):
+1 to this PR. Having the base class decreases the boilerplate a lot.

cirq/sim/act_on_args.py
daxfohl: I'd like to do that in a separate PR if possible. For now the decompose is always the last operation called, so failure would just mean the circuit can't be simulated. The problem with creating a checkpoint is that the checkpoints can be multi-GB in size, so that would be a big perf regression.

I'll look at the on_stuck thing that Balint mentioned and see if restoring that code helps here.

cirq/sim/density_matrix_simulator.py
daxfohl: I followed the pattern established in the original act_on protocol PR, where those were removed for the sparse simulator https://github.com/quantumlib/Cirq/pull/3019/files (on_stuck was removed there and still does not exist in sparse simulator code today). If we want to restore the nice error message I'll look into how to do it for the other simulators too.

cirq/sim/act_on_density_matrix_args.py
```diff
@@ -0,0 +1,98 @@
...
+            log_of_measurement_results: A mutable object that measurements are
+                being recorded into. Edit it easily by calling
+                `ActOnStateVectorArgs.record_measurement_result`.
+        """
+        super().__init__(axes, prng, log_of_measurement_results)
+        self.target_tensor = target_tensor
+        self.available_buffer = available_buffer
+        self.qid_shape = qid_shape
+
+    def _act_on_fallback_(self, action: Any, allow_decompose: bool):
```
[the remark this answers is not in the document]
daxfohl: Sure, though IDK if I like the pattern. While it avoids some duplication, it makes it harder to follow the flow of the code, doesn't give a great place to set breakpoints, and makes it harder to tell what's going on when you step through in the debugger. I'm not pushing for changing it, but that's just been my experience with the existing pattern so far.

cirq/sim/act_on_args.py
```diff
@@ -0,0 +1,89 @@
...
+            invert_mask: The invert mask for the measurement.
+        """
+        bits = self.perform_measurement()
+        corrected = [bit ^ (bit < 2 and mask) for bit, mask in zip(bits, invert_mask)]
+        if key in self.log_of_measurement_results:
+            raise ValueError(f"Measurement already logged to key {key!r}")
+        self.log_of_measurement_results[key] = corrected
+
+    @abc.abstractmethod
+    def perform_measurement(self) -> List[int]:
```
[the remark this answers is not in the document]
daxfohl: Good catch, yes this was intended to be private. Will change.

cirq/sim/act_on_args.py
```diff
@@ -0,0 +1,89 @@
...
+    val: Any,
+    args: ActOnArgs,
+) -> bool:
+    operations, qubits, _ = _try_decompose_into_operations_and_qubits(val)
+    if operations is None:
+        return NotImplemented
+    return _act_all_on_(operations, qubits, args)
+
+
+def _act_all_on_(actions: Iterable[Any], qubits: Sequence[ops.Qid], args: ActOnArgs):
```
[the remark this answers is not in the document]
daxfohl: I'm just going to inline it. This would never get called/used anywhere else.

cirq/sim/density_matrix_simulator.py
daxfohl: Looked further, this is going to take some work. The decompose method that is called in the established act_on classes returns the qubits (which we require) but does not handle a keep/on_stuck. The decompose methods that accept keep/onstuck do not return the qubits. Will figure out if this can be fixed.

daxfohl: Actually I think this goes back to the "iterate through the list of strategies". The only error message we can provide is that `_act_on_` failed. Because if we return the fact that decompose failed, well the user may not have expected to even hit the decompose strategy and the decomposition error may be irrelevant. Will have to think about this some more.

daxfohl: We don't actually need `keep` because the act_on implementation is to do this recursively: act_on will decompose a level, then act_on that level, which will decompose further if necessary. This is different from how things were implemented prior to act_on, which required a perfect decomposition before simulating each op. So, returning the error message if the `_act_on_fallback_` doesn't succeed fixes the problem.

cirq/sim/act_on_args.py
daxfohl: One option would be, instead of having each ActOnXXXArgs implement `_act_on_fallback_`, we could implement it in the ActOnArgs base class. The implementation would look like

```
self._act_on_fallback_impl
if allow_decompose:
  _strat_decompose(self, action)
```

such that `_act_on_fallback_impl` is an abstract method implemented in the child classes, and now _strat_decompose is private so we know it should always be the last strategy called before giving up, so no need to copy the whole state tensor each time it executes.

I think this change could go in a separate PR though, since currently it's not required.

cirq/sim/act_on_args.py
```diff
@@ -0,0 +1,85 @@
+# Copyright 2021 The Cirq Developers
```
daxfohl, line 1: Test class?

cirq/sim/act_on_density_matrix_args.py
```diff
@@ -0,0 +1,119 @@
+# Copyright 2021 The Cirq Developers
```
daxfohl, line 1: Test class?

cirq/sim/act_on_density_matrix_args.py
```diff
@@ -0,0 +1,119 @@
...
+            strats.append(strat_act_on_from_apply_decompose)  # type: ignore
+
+        # Try each strategy, stopping if one works.
+        for strat in strats:
+            result = strat(action, self)
+            if result is False:
+                break  # coverage: ignore
+            if result is True:
+                return True
+            assert result is NotImplemented, str(result)
```
daxfohl, line 77: It's TypeError now.

cirq/sim/act_on_state_vector_args.py
```diff
@@ -167,8 +150,22 @@ def _act_on_fallback_(self, action: Any, allow_decompose: bool):
             if result is True:
                 return True
             assert result is NotImplemented, str(result)
```
daxfohl, line 152: Needs to change: it's TypeError now.

cirq/sim/act_on_density_matrix_args.py
```diff
@@ -0,0 +1,119 @@
...
+        return TypeError(
```
balopat, line 78: this could/should be a raise - that way we don't need to change anything in the act_on_protocol itself, right?

daxfohl: Yes, that seems better. Thanks!

balopat (approved):

balopat (approved):
can you remove the two png files that were generated?

daxfohl:
> can you remove the two png files that were generated?

Done

daxfohl:
Wait, it looks like those png files were there already, but perhaps got corrupted on my machine somehow. I'll see if I can restore them to their original state.

daxfohl:
Done

balopat:
Thanks for your hard work on this @daxfohl!

daxfohl:
@tonybruguier Will you have time to migrate MPS or do you want me to take it?

tonybruguier:
Either way is fine. I can't get to it until this weekend, so if it's urgent, please go ahead

daxfohl:
No, not urgent at all. This PR contained a lot of infra plumbing so it should be a smaller change on MPS side. You should be able to more or less follow what I did here for `density_matrix_simulator` and `act_on_density_matrix_args`.

tonybruguier:
I sent #3943 for review. I tried to follow your patterns, but I wasn't sure.

cirq/sim/act_on_args.py
daxfohl: Doing this in #4979, #5065

PR #3842, opened 2021-02-20
[XEB] Enable characterizing other gates

mpharrigan:
We don't provide sensible defaults for anything other than sqrt(iswap) but with these changes, you can write your own XEBCalibrationOptions subclass to characterize whatever you want.


mrwojtek (approved):
This is good. I'm looking forward to see how it evolves.

balopat (approved):
LGTM!

PR #3845, opened 2021-02-23
circuit = circuit.freeze()

### LINKED ISSUE #3807: FrozenCircuit should be able to equal Circuit
In python, freezing an object does not make it unequal. For example:

```
assert frozenset([1, 2]) == set([1, 2])
# passes
```

Currently cirq is breaking this convention:

```
assert cirq.Circuit().freeze() == cirq.Circuit()
# fails
```

Note that this doesn't require the hash codes to be equal. Circuit should still have no hash code, because it is mutable.

95-martin-orion:
Fixes #3807.


balopat (approved):
LGTM

Issue #3763, opened 2021-02-08
Flake: density_matrix_simulator_test.py::test_simulate_compare_to_state_vector_simulator


balopat:
On an unrelated change I saw this failure on the coverage tests.

https://github.com/quantumlib/Cirq/pull/3738/checks?check_run_id=1857693314

```
FAILED cirq/sim/density_matrix_simulator_test.py::test_simulate_compare_to_state_vector_simulator[complex64-circuit9]
1 failed, 10002 passed, 81 skipped, 63 xfailed, 12577 warnings in 464.32s (0:07:44)
```

see [logs_3596.zip](https://github.com/quantumlib/Cirq/files/5946979/logs_3596.zip)

balopat:
confirmed - just a simple rerun made it pass: https://github.com/quantumlib/Cirq/pull/3738/checks?check_run_id=1858461253

mpharrigan:
anyone know if this is still an issue @MichaelBroughton ?

mpharrigan:
presumably fixed; feel free to re-open

PR #3849, opened 2021-02-24
fix grpcio-tools version

balopat:
Fixes an erroneous upgrade (https://github.com/quantumlib/Cirq/pull/3365), that should have been more conservative. 
We have to be careful with grpcio-tools upgrades as that determines the pb2 files that are generated.
Otherwise a breakage will occur when building TFQ: the `_internal_create_key` symbol doesn't exist in older versions.


balopat:
Got a flake of https://github.com/quantumlib/Cirq/issues/3763 in https://github.com/quantumlib/Cirq/pull/3849/checks?check_run_id=1967506372

dev_tools/conf/pip-list-dev-tools.txt
```diff
@@ -12,8 +12,10 @@ filelock~=3.0.12
 
 # For generating protobufs
 
-# 1.26 is the min version that has python 3.8 wheels
-grpcio-tools~=1.26
+# 1.26.0 is the min version that has python 3.8 wheels
+# however, we can't really go much higher - the protoc version needs to be <= 3.9.2
+# otherwise it will generate pb2 files that are incompatible with tensorflow quantum
+grpcio-tools~=1.26.0
```
95-martin-orion, line 18: This seems like a roundabout way of enforcing `protoc <= 3.9.2` - is there any reason we can't just add that as a rule instead? pip-install should ensure compatibility between the selected protoc and grpcio versions.

balopat: grpcio-tools is an easy way of installing protoc, but it is included into the binary (see this comment for example https://github.com/grpc/grpc/issues/15675#issuecomment-397038811) - so there is no way to enforce protoc with pip.

95-martin-orion: Seems fine to me then.

95-martin-orion (approved):
LGTM. Is #3763 also expected to be resolved by this?

balopat:
> LGTM. Is #3763 also expected to be resolved by this?

Unfortunately no :( That is just a flake that goes away when you rerun your checks.

PR #3851, opened 2021-02-24
Ensure density matrix doesn't restart at zero between repetitions

### LINKED ISSUE #3850: Perf optimization in Density Matrix Simulator `_run_sweep_repeat`

Density matrix simulator's `_run_sweep_repeat` function unnecessarily starts from zero every time.

It should instead break the circuit into two parts: a pre-first-measurement prefix, and post-first-measurement suffix, per qubit (similar to how sparse simulator does for unitary and non-unitary). The first part need only run once. From that, the density matrix can be copied and fed as the initial state of the second part, which will be repeated `repetitions` times.

Need to double-check exactly what can go into the prefix circuit. Can all channels? Can reset gates? Does noise affect anything? (Don't think so, but maybe weird cases). It would be best if there's some protocol that can be used to break into the prefix/suffix.

We could leave it. It only affects circuits that have non-terminal measurements or subcircuits (circuits with only terminal measurements are already optimized by `_run_sweep_sample` function). Those may be rare or use other simulators.

daxfohl:
Fixes #3850

Optimizes density matrix simulator such that it only performs the operations up to the first measurement once. If there are repetitions thereafter, it doesn't redo those operations. Instead it reuses the density matrix from that first step to initialize the simulation of the remaining steps only.


mpharrigan:
Do you have profiling data to show how much this helps? `dev_tools/profiling/` could be a place to put such a script

daxfohl:
> Do you have profiling data to show how much this helps? `dev_tools/profiling/` could be a place to put such a script

Oof, just tried this and it doesn't appear to have made much difference at all. Is this expected? In density matrix is it the sampling that takes the time rather than simulation?

95-martin-orion:
I would mostly expect this to help with circuits that only have noise near the end. Seems like an unusual use case, but I'm not against optimizing it.

One question, though: what's happening to `mock_sim.call_count` in the tests? My base expectation is that it would double, given that both the prefix and suffix must run, but it doesn't look like that's always the case. In particular, I'm curious about the "no repetitions" case where it jumps from 0 to 4.

daxfohl:
Benchmarks pre:
```
python dev_tools/profiling/benchmark_simulators.py --sim_type density_matrix --run_repetitions=10
num_qubits,seconds per gate
4,0.0020236950000107755
5,0.0022459623999893665
6,0.002720043199995416
7,0.003916122199996607
8,0.008510038799999165
9,0.028214186600001995
```
Benchmarks post:
```
python dev_tools/profiling/benchmark_simulators.py --sim_type density_matrix --run_repetitions=10
num_qubits,seconds per gate
4,0.0005682005999988178
5,0.0007923434000113047
6,0.0008120525999984238
7,0.0011759149999998044
8,0.0024289809000038074
9,0.008432403800004977
```

daxfohl:
> One question, though: what's happening to `mock_sim.call_count` in the tests? My base expectation is that it would double, given that both the prefix and suffix must run, but it doesn't look like that's always the case. In particular, I'm curious about the "no repetitions" case where it jumps from 0 to 4.

It's complicated, it always runs the prefix, then it runs the suffix `num_repetitions` times for sweep_repeat or once for sweep_sample. Note that these tests generally run in loops so you've got to go through and work out how many times you expect it to run. The "no repetitions" runs the prefix because everything runs the prefix, and that test loops four times so that's why the jump from 0 to 4. Maybe "no repetitions" should jump out earlier so that it stays zero.

daxfohl:
The primary use case is if there are nonterminal measurements, mostly at the end of the circuit, and is run with repetitions>1. This will cause the prefix (everything before the first measurement) to run only once.

If that would be an abnormal/degenerate circuit style to use with density matrix, I'm fine dropping this.

95-martin-orion:
> It's complicated, it always runs the prefix, then it runs the suffix `num_repetitions` times for sweep_repeat or once for sweep_sample. {...} Maybe "no repetitions" should jump out earlier so that it stays zero.

Okay, that checks out. +1 to exiting early on `repetitions=0`, if only to avoid code smell in the test.

> If [late, nonterminal measurements] would be an abnormal/degenerate circuit style to use with density matrix, I'm fine dropping this.

I'm still in favor of merging this PR (once the above change is in). I think most of the skepticism here is because the density matrix simulator is almost exclusively used for simulating noisy circuits, since the sparse simulator is faster and consumes less memory.

daxfohl:
@95-martin-orion latest commit fixes that logic and tests for all simulators and cleaned up some of redundant code in the process.

BTW @mpharrigan I wasn't sure if you wanted me to actually commit the benchmark changes like I've done, or test them locally. If the latter, I should revert before this goes through.

95-martin-orion (approved):
It seems reasonable to me to include the benchmark changes in this PR.

One final nit: please update the `SimulatesSamples._run` docstring to indicate that repetitions is validated prior to being passed to `_run`.

daxfohl:
> One final nit: please update the `SimulatesSamples._run` docstring to indicate that repetitions is validated prior to being passed to `_run`.

Done

PR #3855, opened 2021-02-25
[obs] 4.2 - Basic sampling loop

mpharrigan:
Loop through grouped settings, execute, accumulate bitstrings. 

The following PRs will add
 - More advanced control of how many samples to take per chunk and when to stop
 - Checkpointing
 - Higher-level APIs that are more convenient to use

Part of https://github.com/quantumlib/Cirq/pull/3647
Follows https://github.com/quantumlib/Cirq/pull/3792


cirq/work/observable_measurement.py
```diff
@@ -211,3 +269,148 @@ def _subdivide_meas_specs(
...
+def _get_qubits(max_settings: Iterable[InitObsSetting]):
+    """Helper function to find all the qubits in a suite of settings used
+    in `measure_grouped_settings`.
+
+    Qubits are returned in sorted order.
+    """
+    qubits = set()
+    for max_setting in max_settings:
+        max_in_st = max_setting.init_state
+        qubits |= set(max_in_st.qubits)
```
balopat, line 290: Could be done with list comprehension: 
```
return sorted({q for ms in max_settings for q in ms.qubits})
```

cirq/work/observable_measurement.py
```diff
@@ -211,3 +269,148 @@ def _subdivide_meas_specs(
...
+    return sorted(qubits)
+
+
+def _determine_needs_init_layer(init_state: value.ProductState):
+    """Helper function to determine whether we need the initial layer of
+    rotations used in `measure_grouped_settings`.
+    """
+    for _, st in init_state:
```
balopat, line 298: it would be more pythonic to just `return any(st is not value.KET_ZERO for _, st in init_state)`.

cirq/work/observable_measurement.py
```diff
@@ -211,3 +269,148 @@ def _subdivide_meas_specs(
...
+    Qubits are returned in sorted order.
+    """
+    qubits = set()
+    for max_setting in max_settings:
+        max_in_st = max_setting.init_state
+        qubits |= set(max_in_st.qubits)
+    return sorted(qubits)
+
+
+def _determine_needs_init_layer(init_state: value.ProductState):
```
balopat, line 294:
```suggestion
def _needs_init_layer(init_state: value.ProductState):
```

cirq/work/observable_measurement.py
```diff
@@ -211,3 +269,148 @@ def _subdivide_meas_specs(
...
+        circuit_sweep: Additional parameter sweeps for parameters contained
+            in `circuit`. The total sweep is the product of the circuit sweep
+            with parameter settings for the single-qubit basis-change rotations.
+    """
+
+    qubits = _get_qubits(grouped_settings.keys())
+    qubit_to_index = {q: i for i, q in enumerate(qubits)}
+
+    needs_init_layer = False
+    for max_setting in grouped_settings.keys():
```
balopat, line 345: we could have the whole (for settings as well) "needs init layer" logic extracted into a private method, not just the state part of it

cirq/work/observable_measurement.py
```diff
@@ -211,3 +269,148 @@ def _subdivide_meas_specs(
...
+    measurement_param_circuit = _with_parameterized_layers(circuit, qubits, needs_init_layer)
+    grouped_settings = {
+        _pad_setting(max_setting, qubits): settings
+        for max_setting, settings in grouped_settings.items()
+    }
+
+    if circuit_sweep is None:
+        circuit_sweep = study.UnitSweep
+    else:
+        circuit_sweep = study.to_sweep(circuit_sweep)
```
balopat, line 359: `circuit_sweep = study.UnitSweep if circuit_sweep is None else study.to_sweep(circuit_sweep)`

cirq/work/observable_measurement.py
```diff
@@ -211,3 +269,148 @@ def _subdivide_meas_specs(
...
+            readout_symmetrization=readout_symmetrization,
+        )
+
+        resolved_params = [
+            flippy_ms.param_tuples(needs_init_layer=needs_init_layer)
+            for flippy_ms in flippy_meas_specs
+        ]
+        resolved_params = _to_sweep(resolved_params)
+        total_samples = repetitions * len(resolved_params)
+        print("Total samples", total_samples)
```
balopat, line 404: don't we want to use some proper logging instead of print statements? In that case we can also write tests for the things we print with `cirq.testing.assert_logs`, but also timestamps, formatting, file specs, etc. can be added on top of this kind of output in general.

cirq/work/observable_measurement.py
```diff
@@ -211,3 +269,148 @@ def _subdivide_meas_specs(
...
+
+        results = sampler.run_sweep(
+            program=measurement_param_circuit, params=resolved_params, repetitions=repetitions
+        )
+
+        assert len(results) == len(flippy_meas_specs)
```
balopat, line 410: Add a readable error message here. In case we run into this, it will be a bug - some hints would be useful to see what expectation failed, why do we think this equality should hold.

balopat (changes requested):
Looks good, picked a bunch of nits.

cirq/work/observable_measurement.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 The Cirq developers
+# Copyright 2021 The Cirq developers
```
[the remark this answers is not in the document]
mpharrigan: reverted

cirq/work/observable_measurement.py
```diff
@@ -150,6 +152,62 @@ def _pad_setting(
...
+    of repetitions for the next chunk. For batching efficiency, we take the
+    max and issue a warning in this case."""
+    if len(next_chunk_repetitions) == 1:
+        return list(next_chunk_repetitions)[0]
+
+    reps = max(next_chunk_repetitions)
+    warnings.warn(
+        "Your stopping criteria recommended a various numbers of "
+        "repetitions to perform next. So we can submit as a single "
+        "sweep, we will be taking the maximum of {}".format(reps)
```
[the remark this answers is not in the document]
mpharrigan: takes all the character out of it!

changed to fstring and reworded.

cirq/work/observable_measurement.py
```diff
@@ -150,6 +152,62 @@ def _pad_setting(
...
+
+
+def _repetitions_to_do(accumulator: BitstringAccumulator, desired_reps: int) -> int:
+    """Stub function to chunk desired repetitions into groups of 10,000."""
+    done = accumulator.n_repetitions
+    todo = desired_reps - done
+    if todo <= 0:
+        return 0
+
+    to_do_next = min(10_000, todo)
```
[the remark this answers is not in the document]
mpharrigan: this is a shim until the next PR which has both a `RepetitionsStoppingCriteria` and a `VarianceStoppingCriteria`. The max chunk size is configurable as part of those classes.

cirq/work/observable_measurement.py
mpharrigan: I always shy away from these double comprehensions because I can never remember how they work! Must think "odometer". Changed, and removed this helper function since it's now one line (line 337).

cirq/work/observable_measurement.py
mpharrigan: done offline

cirq/work/observable_measurement.py
mpharrigan: good call. changed and inlined this call; removed helper function.

cirq/work/observable_measurement.py
mpharrigan: what do you mean? in any event, I inlined the now-one-line definition of `_[determine_]needs_init_layer`.

Whether an init layer is required only depends on the init state.

cirq/work/observable_measurement.py
mpharrigan: changed

cirq/work/observable_measurement.py
mpharrigan: Do we have an example of user-facing logging in cirq? The python `logging` module requires the user to configure something to actually get the log messages to appear and the google cloud stuff produces a lot of logging output. 

I'm removing this print statement for now. We can add better logging later.

cirq/work/observable_measurement.py
mpharrigan: this is just testing that the number of results matches the number of params requested and done defensively because of the `zip` behavior in the next line. If this equality doesn't hold we have bigger problems. Added an assertion message.

cirq/work/observable_measurement_test.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 The Cirq developers
+# Copyright 2021 The Cirq developers
```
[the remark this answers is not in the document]
mpharrigan: reverted

mpharrigan (commented):
sorry for the delay, PTAL!

mpharrigan:
CI failure looks scary but unrelated

balopat:
> CI failure looks scary but unrelated

Wow that does look scary. Restarted the jobs to see whether it's a flake. Opened https://github.com/quantumlib/Cirq/issues/3951.

cirq/work/observable_measurement.py
balopat: I meant having `_determine_needs_init_layer` separately seemed like it could have encompassed these (now) 5 lines of code. This method is very long, I was trying to chip away extractable pieces so it reads better.

mpharrigan: factored out into a helper function, thanks

mpharrigan (commented):
anything else @balopat ?

mpharrigan:
@95-martin-orion since you've been working on similar things

balopat (approved):
LGTM

Issue #3848, opened 2021-02-23
During testing, make accidentally using deprecated methods raise errors


Strilanc:
A recurring issue in Cirq is that we deprecate something, but continue to use it in other code within Cirq (e.g. https://github.com/quantumlib/Cirq/issues/3847 ). This creates false positive deprecation warnings for users of the code. We should ensure this never happens by turning such usages into errors during unit tests.

I'm not sure exactly how to go about implementing this, but e.g. the `deprecated` decorator that we use could check if it is running in "check mode" by looking for an environment flag that is set by the continuous integration system. Note that there needs to be an escape hatch that a test can use when it is explicitly checking that the deprecated function is forwarded correctly. For example, a `with cirq.testing.ExpectDeprecatedCall()` context manager that sets a flag that the `@deprecated` decorator can look for.

maffoo:
+1. We have functions to do exactly this internally and it's very helpful.

balopat:
+1 - I would do this before the release.

Issue #3847, opened 2021-02-23
Deprecated _simulator_iterator used in simulate_moment_steps


Strilanc:
We shouldn't use deprecated methods internally. Causes false warnings when using some simulators.

PR #3860, opened 2021-02-25
fail cirq's use of deprecated features in tests

balopat:
Fail cirq's use of deprecated features in tests. 

It also:
- standardizes the deadline parameter for deprecations - it must be a valid minor version. 
- it forces the deprecation tests to specify this minor version
- it fixes one v0.10 deprecation that went uncaught: `decompose_two_qubit_interaction_into_four_fsim_gates_via_b` is now removed.  
- it fixes incorrect deprecation of of `_simulator_iterator` and adds tests for it
- forces the deprecation of JSON serializable objects to be tested

Fixes https://github.com/quantumlib/Cirq/issues/3848.
Fixes https://github.com/quantumlib/Cirq/issues/3847.


cirq/testing/deprecation.py
```diff
@@ -0,0 +1,34 @@
...
+
+import os
+
+from contextlib import contextmanager
+
+ALLOW_DEPRECATION_IN_TEST = 'ALLOW_DEPRECATION_IN_TEST'
+
+
+@contextmanager
+def allow_deprecation():
```
maffoo, line 23: Would it be possible to make this work like `pytest.raises` where it will _fail_ if no deprecation warning is emitted? Also, it could take a string or regex which is matched against the deprecation warning to ensure that we don't catch things too broadly. These could help ensure that we use `allow_deprecation` in a targeted way, document in code why it's used in specific instances, and remind us to remove it if no longer needed.

balopat: Sounds good - I was thinking about combining this with the logs assertion, but in the end went with the narrower option. I will create an `assert_deprecated` method instead that combines the two and more importantly makes the version argument mandatory which will make it easier to "flush" a certain version's deprecation backlog.

cirq/testing/deprecation.py
```diff
@@ -0,0 +1,34 @@
...
+    """Allows deprecated functions, classes, decorators in tests.
+
+    It acts as a contextmanager that can be used in with statements:
+    >>> with allow_deprecation():
+    >>>     # do something deprecated
+    """
+    os.environ[ALLOW_DEPRECATION_IN_TEST] = 'True'
```
mpharrigan, line 30: Why are you doing this via environment variables? I thought originally you wanted to leave open the possibility that people running pytest could set this environment variable themselves to run things the old way, i.e. deprecation warnings are warnings. However, this function will actually unset the environment variable any time the context manager is used, so maybe that wasn't the intent. 

Have you considered using a global variable in `cirq/_compat.py` that you can set/unset instead? I think that would be a little more hygienic

balopat: I want to keep allowing the deprecation to very narrow, specific cases: when we know that there is a deprecated function and we want to test that behavior. Any other time, things should fail while running in pytest. 

Env var vs global var - I considered that, it is essentially the same, but at least there is a symmetry with the pytest method. I have no strong opinion there - why do you think the hygiene is better with a global var compared to an env var?

balopat:
@maffoo PTAL, made it much more targeted!

cirq/_compat.py
```diff
@@ -251,12 +274,11 @@ class Wrapped(ModuleType):
         def __getattr__(self, name):
             if name in deprecated_attributes:
                 deadline, fix = deprecated_attributes[name]
-                warnings.warn(
+                _validate_deadline(deadline)
```
Strilanc, line 277: Why is the deadline being validated at call time instead of at define time?

Strilanc (approved):

cirq/_compat.py
balopat: oh, that is an oversight, thanks for catching that!

cirq/testing/deprecation.py
mpharrigan: What symmetry with what pytest method? 

Hygiene -- it blows away any user-set value for `ALLOW_DEPRECATION_IN_TEST` which the user may be using for another process, in their bashrc, etc. It's the most global of all global variables :)

cirq/sim/simulator.py
```diff
@@ -457,11 +457,13 @@ def simulate_moment_steps(
...
             moment and returning a StepResult for each moment.
         """
-        return self._simulator_iterator(
-            circuit, study.ParamResolver(param_resolver), qubit_order, initial_state
-        )
+        param_resolver = study.ParamResolver(param_resolver)
+        resolved_circuit = protocols.resolve_parameters(circuit, param_resolver)
+        check_all_resolved(resolved_circuit)
+        actual_initial_state = 0 if initial_state is None else initial_state
+        return self._base_iterator(resolved_circuit, qubit_order, actual_initial_state)
```
daxfohl, line 464: @balopat  This ends up being a breaking change though for any simulator that was overriding `_simulator_iterator`. This is why I had left the call to `_simulator_iterator` there in the original PR #3650. Otherwise I'd have just inlined and deleted `_simulator_iterator` outright: it's a private method and it's unused elsewhere.

It doesn't sound like anyone has been broken by this, so maybe it's okay. (We can just delete `_simulator_iterator` in that case; it's not called by anyone). I think there's still an unmet need for cleanly deprecate abstract methods, or methods that were intended to be overridden. The base class has to continue calling them until the deadline or it will break child classes. I like the error-by-default, but maybe there's a way to silence these errors when retaining the call to a deprecated method is required, like here.

balopat: You are totally right, I rushed this one this doesn't make sense at all as it is :).
I'll add the BREAKING_CHANGE label to this PR. Also, I'll add it to the list of breaking changes in the release notes for completeness sake. 

Re abstract methods deprecation - hmm yeah we'll need to think about that. I'll open an issue for it.

mpharrigan:
We should add a page on how to deprecate stuff correctly to the contributor guide. Asking for a friend.

PR #3862, opened 2021-02-25
Add support for using physical-Z phase matching on 2-qubit gates

maffoo:
When run on hardware, two-qubit gates must be "phase-matched" to account for the detuning between the rest frames of the two qubits. By default, this phase matching will use "virtual" Z gates where possible, namely for gates like CZ and iSWAP that commute with single-qubit Z rotations (in the case of iSWAP, the commuted Zs swap to the other qubit). In some cases, however, we may want to use "physical" Z gates for matching, for example to avoid errors due to non-ideal gates (non-zero swap angle in CZ, or incomplete swap in iSWAP).

Here we add support for tagging 2-qubit gates with `cirq.google.PhysicalZTag()` to indicate that physical Z gates should be used for matching, instead of the default matching scheme. We modify the 2-qubit gate serializers to recognize these tags and pass them through when serializing/deserializing.

Note that for sqrt-iSWAP gates and others that do not commute with Z, the default behavior is already to use physical Z gates for matching. In these cases adding an explicit `PhysicalZTag()` does not change the behavior, but we allow it for consistency with other 2-qubit gates.


cirq/google/common_serializers.py
```diff
@@ -437,7 +459,7 @@ def _convert_physical_z(op: ops.Operation, proto: v2.program_pb2.Operation):
     op_serializer.GateOpSerializer(
         gate_type=ops.FSimGate,
         serialized_gate_id='fsim_pi_4',
-        args=[],
+        args=[_phase_match_arg],
```
dstrain115, line 462: This is a little confusing, since the default is to match via physical Z for sqrt iswap, correct?

dstrain115 (approved):
Perhaps we could add more to PR description and/or the commit message?

cirq/google/common_serializers.py
maffoo: That's correct. I think it's nice to have this here for consistency with other gates. I've added a note about this in the PR description.

maffoo:
Thanks, @dstrain115. I've updated the PR description for posterity, and will include that in the commit message.

Issue #3634, opened 2020-12-30
Efficient proto serialization of FrozenCircuit


95-martin-orion:
**Is your feature request related to a use case or problem? Please describe.**

Part of #3235.

PR #3480 added a lookup table for constants in the circuit proto. Extending this table to more complex objects should help reduce the amount of network traffic required when using Quantum Engine.

**Describe the solution you'd like**

FrozenCircuits are a prime candidate for storing in the constants table, as they are often fairly large and may be repeated several times within a circuit.

**What is the urgency from your perspective for this issue? Is it blocking important work?**

P2 - we should do it in the next couple of quarters

95-martin-orion:
Blocked on #3678. Workarounds are possible, but messy.

95-martin-orion:
It should be noted that proto serialization of `FrozenCircuit` - efficient or not - is currently unavailable, as the serializers have no assigned behavior for handling `CircuitOperation`. Decomposing the circuit in advance avoids this issue, but also removes all of the intended benefits of using `CircuitOperation`s in the first place.

95-martin-orion:
It may be possible to enable more general proto compression with tools provided by gRPC: [SetCompressionLevel](https://grpc.github.io/grpc/cpp/classgrpc_1_1_server_builder.html#ae10c6dc23ec295695710c0b4c14f9353)

maffoo:
I wonder if we should consider making a v3 proto format to handle subcircuits, instead of trying to force them into the current format which has various problems? If we do want to ahead with adding support in v2, then I think we can work around #3678 in this particular case fairly easily, by checking for `CircuitOperation` explicitly before we check `op.gate` to pick a serializer. This seems like a fairly natural special case since `CircuitOperation` is so different from other operations.

95-martin-orion:
> I wonder if we should consider making a v3 proto format to handle subcircuits, instead of trying to force them into the current format which has various problems?

It seems like v3 proto is still a topic under discussion, and may take some time to be fully realized. In order to make subcircuits available to QCS as soon as possible, I plan to continue work on this, but please keep me in the loop on the v3 proto plans.

95-martin-orion:
As described in #4336, this issue is now effectively resolved; however, additional proto optimizations are still possible.

95-martin-orion:
> As described in #4336, this issue is now effectively resolved; however, additional proto optimizations are still possible.

[...]

PR #3864, opened 2021-02-25
Use gRPC tools to serialize protos.

95-martin-orion:
Wraps all transport stubs with `functools.partial`, binding `compression` to `grpc.Compression.Gzip`. In theory, this should improve our proto size when sending from this client to QCS; additional changes to QCS are necessary to compress protos created by the service. Tangentially related to #3634.

An example from the grpc repository: [CompressingHelloWorldClient](https://github.com/grpc/grpc-java/blob/ae574434eec58e917a00b94756bcfd166b4989a0/examples/src/main/java/io/grpc/examples/experimental/CompressingHelloWorldClient.java#L56-L69)

### DIFF
--- a/cirq/google/engine/client/quantum_v1alpha1/gapic/quantum_engine_service_client.py
+++ b/cirq/google/engine/client/quantum_v1alpha1/gapic/quantum_engine_service_client.py
@@ -242,12 +242,14 @@ class QuantumEngineServiceClient(object):
                     to a retryable error and retry attempts failed.
             ValueError: If the parameters are invalid.
         """
-        # Wrap the transport method to add retry and timeout logic.
+        # Wrap the transport method to add retry, timeout, and compression logic.
         if 'create_quantum_program' not in self._inner_api_calls:
             self._inner_api_calls[
                 'create_quantum_program'
             ] = google.api_core.gapic_v1.method.wrap_method(
-                self.transport.create_quantum_program,
+                functools.partial(
+                    self.transport.create_quantum_program, compression=grpc.Compression.Gzip,
+                ),
                 default_retry=self._method_configs['CreateQuantumProgram'].retry,
                 default_timeout=self._method_configs['CreateQuantumProgram'].timeout,
                 client_info=self._client_info,
@@ -315,12 +317,14 @@ class QuantumEngineServiceClient(object):
                     to a retryable error and retry attempts failed.
             ValueError: If the parameters are invalid.
         """
-        # Wrap the transport method to add retry and timeout logic.
+        # Wrap the transport method to add retry, timeout, and compression logic.
         if 'get_quantum_program' not in self._inner_api_calls:
             self._inner_api_calls[
                 'get_quantum_program'
             ] = google.api_core.gapic_v1.method.wrap_method(
-                self.transport.get_quantum_program,
+                functools.partial(
+                    self.transport.get_quantum_program, compression=grpc.Compression.Gzip,
+                ),
                 default_retry=self._method_configs['GetQuantumProgram'].retry,
                 default_timeout=self._method_configs['GetQuantumProgram'].timeout,
                 client_info=self._client_info,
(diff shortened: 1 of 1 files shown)


cirq/google/engine/client/quantum_v1alpha1/gapic/quantum_engine_service_client.py
```diff
@@ -242,12 +242,15 @@ def create_quantum_program(
...
-        # Wrap the transport method to add retry and timeout logic.
+        # Wrap the transport method to add retry, timeout, and compression logic.
         if 'create_quantum_program' not in self._inner_api_calls:
             self._inner_api_calls[
                 'create_quantum_program'
             ] = google.api_core.gapic_v1.method.wrap_method(
-                self.transport.create_quantum_program,
+                functools.partial(
+                    self.transport.create_quantum_program,
+                    compression=grpc.Compression.Gzip,
```
wcourtney, line 252: The PR description points to a Java example but doesn't really reference how it maps to this. Mind documenting somewhere where this `compression` parameter is defined? IIUC, it comes from the [`Channel`](https://github.com/grpc/grpc/blob/86f93801e5d452cff40d6958d21a8eb211da0654/src/python/grpcio/grpc/_channel.py#L1345) class, but that was a real pain to track down.

cirq/google/engine/client/quantum_v1alpha1/gapic/quantum_engine_service_client.py
```diff
@@ -647,12 +662,15 @@ def create_quantum_job(
...
             ValueError: If the parameters are invalid.
         """
-        # Wrap the transport method to add retry and timeout logic.
+        # Wrap the transport method to add retry, timeout, and compression logic.
         if 'create_quantum_job' not in self._inner_api_calls:
             self._inner_api_calls[
                 'create_quantum_job'
             ] = google.api_core.gapic_v1.method.wrap_method(
-                self.transport.create_quantum_job,
+                functools.partial(
```
wcourtney, line 670: Should this be done in a helper to assist with this wrapping readability and future maintenance? It's repeated many times.

95-martin-orion:
TODO: similar to #2979, any changes to generated code must be documented internally.

cirq/google/engine/client/quantum_v1alpha1/gapic/quantum_engine_service_client.py
```diff
@@ -890,12 +914,15 @@ def delete_quantum_job(
...
         """
-        # Wrap the transport method to add retry and timeout logic.
+        # Wrap the transport method to add retry, timeout, and compression logic.
         if 'delete_quantum_job' not in self._inner_api_calls:
             self._inner_api_calls[
                 'delete_quantum_job'
             ] = google.api_core.gapic_v1.method.wrap_method(
-                self.transport.delete_quantum_job,
+                functools.partial(
+                    self.transport.delete_quantum_job,
```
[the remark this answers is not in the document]
95-martin-orion: 1. There are [hints in the gRPC docs](https://grpc.github.io/grpc/python/grpc_testing.html) that such benchmarks are possible, but I don't think our tests currently use this behavior. I'll dig deeper and see if I can't write these myself.
2. If we can identify such a threshold (and can inspect proto sizes in the client) it should be possible to selectively compress protos sent to the same method. My read of the gRPC docs is that the client can choose its compression behavior on a per-call basis, independent from any server config.

cirq/google/engine/client/quantum_v1alpha1/gapic/quantum_engine_service_client.py
95-martin-orion: Will do - with the other changes this becomes especially necessary.

95-martin-orion:
Discussed offline with Balint - our benchmarks show that using concise serialization (similar to the current JSON behavior) actually results in better compression that gzip. It's also possible to combine both compression types, but I'll be focusing on bringing concise serialization to protos before following up on this PR.

dabacon:
@95-martin-orion did we decide to do this in leu of the concise representation work?  Worry it is getting a little stale at one year old :)

95-martin-orion:
> @95-martin-orion did we decide to do this in leu of the concise representation work? Worry it is getting a little stale at one year old :)

Doug and I implemented concise serialization for protos since this was written. Based on the note about concise serialization beating gzip on size metrics, I think it's fair to close this. We can revisit if greater compression is needed.

PR #3869, opened 2021-02-27
Remove unused perform_measurements argument from sparse simulator

daxfohl:
The `perform_measurements` argument of private method `Simulator._base_iterator` defaults to True (yes, perform the measurements please), and is only set to False in one place: when iterating the unitary prefix of the circuit. However, this prefix will never have a measurement anyway by definition, so setting this argument here has no effect on the iteration of the unitary prefix. Thus this argument and the conditional around it is unnecessary and can be removed. This PR removes it.

### DIFF
--- a/cirq/sim/sparse_simulator.py
+++ b/cirq/sim/sparse_simulator.py
@@ -187,7 +187,6 @@ class Simulator(
             circuit=unitary_prefix,
             qubit_order=qubit_order,
             initial_state=0,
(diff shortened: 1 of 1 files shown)


95-martin-orion (approved):
I was going to raise a concern about this conflicting with `SimulatesIntermediateState`, but it looks like `perform_measurements` wasn't part of that API to begin with. LGTM.

PR #3873, opened 2021-03-02
Factorise circuit into parallelizable components (will fix #3409)

### LINKED ISSUE #3409: Simulate factorizable circuit

Ashalynd:
This supersedes #3599. 

Fixes #3409.


cirq/circuits/circuit.py
```diff
@@ -2200,6 +2201,31 @@ def with_noise(self, noise: 'cirq.NOISE_MODEL_LIKE') -> 'cirq.Circuit':
...
+        if len(qubit_sets) == 1:
+            return (self,)
+        else:
+
+            new_moments: List[List['cirq.Moment']] = [[] for _ in qubit_sets]
+            for m in self.moments:
+                for mi, qubits in enumerate(qubit_sets):
+                    new_moment = m[qubits]
+                    if add_empty_moments or len(new_moment) > 0:
+                        new_moments[mi].append(m[qubits])
```
mpharrigan, line 2226: why `m[qubits]` instead of `new_moment`?

cirq/circuits/circuit.py
```diff
@@ -2200,6 +2201,31 @@ def with_noise(self, noise: 'cirq.NOISE_MODEL_LIKE') -> 'cirq.Circuit':
...
+        qubit_sets = sorted(uf.to_sets(), key=lambda x: min(x))
+        if len(qubit_sets) == 1:
+            return (self,)
+        else:
+
+            new_moments: List[List['cirq.Moment']] = [[] for _ in qubit_sets]
+            for m in self.moments:
+                for mi, qubits in enumerate(qubit_sets):
+                    new_moment = m[qubits]
+                    if add_empty_moments or len(new_moment) > 0:
```
mpharrigan, line 2225: `add_empty_moments=False` might be a nice feature, but a lot of other circuit transformers make you explicitly call `cirq.DropEmptyMoments()` to get rid of them. Removing this parameter (and defaulting to the `True` case) could greatly simplify this slicing. You wouldn't need the `mi` index and checks, and could basically do this in a big list comprehension.

cirq/circuits/circuit.py
```diff
@@ -2200,6 +2201,31 @@ def with_noise(self, noise: 'cirq.NOISE_MODEL_LIKE') -> 'cirq.Circuit':
...
+        circuit (this one).
+        """
+        uf = networkx.utils.UnionFind(self.all_qubits())
+        for op in self.all_operations():
+            if len(op.qubits) > 1:
+                uf.union(*op.qubits)
+        qubit_sets = sorted(uf.to_sets(), key=lambda x: min(x))
+        if len(qubit_sets) == 1:
+            return (self,)
+        else:
```
mpharrigan, line 2219: style nit: since you're `return`ing above, drop the `else` and de-dent the rest of the function.

cirq/circuits/circuit.py
```diff
@@ -2200,6 +2201,31 @@ def with_noise(self, noise: 'cirq.NOISE_MODEL_LIKE') -> 'cirq.Circuit':
...
+
+            new_moments: List[List['cirq.Moment']] = [[] for _ in qubit_sets]
+            for m in self.moments:
+                for mi, qubits in enumerate(qubit_sets):
+                    new_moment = m[qubits]
+                    if add_empty_moments or len(new_moment) > 0:
+                        new_moments[mi].append(m[qubits])
+            return (Circuit(m) for m in new_moments)
```
mpharrigan, line 2227: We have `cirq.Circuit.zip()` Would it be worth factoring out this second part out and allow slicing on a circuit by qubit like we do for `Moment`? I guess we already use `__get_item__` for time-slicing. Maybe this can be a method like `cirq.Circuit.slice_qubits`?

cirq/circuits/circuit.py
```diff
@@ -2200,6 +2201,31 @@ def with_noise(self, noise: 'cirq.NOISE_MODEL_LIKE') -> 'cirq.Circuit':
...
+        such that there are no entangling gates between them).
+
+        If this is not possible, will return the set consisting of the single
+        circuit (this one).
+        """
+        uf = networkx.utils.UnionFind(self.all_qubits())
+        for op in self.all_operations():
+            if len(op.qubits) > 1:
+                uf.union(*op.qubits)
+        qubit_sets = sorted(uf.to_sets(), key=lambda x: min(x))
```
mpharrigan, line 2216: just say `key=min`

mpharrigan (commented):
Super succinct and readable :) I added some comments

cirq/circuits/circuit.py
Ashalynd: Ugh, good catch. It's now part of that one large comprehension anyway :)

cirq/circuits/circuit.py
Ashalynd: Done.

cirq/circuits/circuit.py
```diff
@@ -2213,18 +2213,11 @@ def decompose(self, add_empty_moments=True) -> Iterable['cirq.Circuit']:
...
-
-            new_moments: List[List['cirq.Moment']] = [[] for _ in qubit_sets]
-            for m in self.moments:
-                for mi, qubits in enumerate(qubit_sets):
-                    new_moment = m[qubits]
-                    if add_empty_moments or len(new_moment) > 0:
-                        new_moments[mi].append(m[qubits])
-            return (Circuit(m) for m in new_moments)
+        new_moments = [[m[qubits] for m in self.moments] for qubits in qubit_sets]
+        return (Circuit(m) for m in new_moments)
```
maffoo, line 2220: nit: it seems a bit odd to me to construct the moments eagerly but then only construct the circuits lazily. I'd suggest doing the whole thing in a single comprehension, and I lean slightly toward eagerness since `Circuit` is mutable so could in theory change between the time you call `decompose` and actually run the generator:
```python
return [Circuit(m[qubits] for m in self.moments) for qubits in qubit_sets]
```

maffoo (commented):
It might be nice to make this a method on `AbstractCircuit` instead, so that it will be available for `Circuit` and `FrozenCircuit`. It could be generic in the `self` type to make the typing work out, something like:
```python
def decompose(self: CIRCUIT_TYPE) -> CIRCUIT_TYPE:
    ...
    return [self._with_sliced_moments(m[qubits] for m in self.moments) for qubits in qubit_sets]
```

cirq/circuits/circuit.py
Ashalynd: Good point, done.

cirq/circuits/circuit.py
Ashalynd: Is the use case 'getting part of the circuit which contains the given qubit'?
That means that instead of all qubit sets, only the one that contains the given qubit will be used, correct? And the output will be a single circuit.
I have added qubit iterator as a parameter to this method (and renamed it to 'factorize' because 'decompose' seems to be already taken) - would it work?

cirq/circuits/circuit.py
Ashalynd: Done

cirq/circuits/circuit.py
mpharrigan: I think the new argument clutters the factorize method a bit.

I didn't have a specific use case in mind. It just seemed like nice-to-have functionality in its own right (which is why the slicing method exists on moments) and would give feature parity between circuits and moments.

cirq/circuits/circuit.py
mpharrigan: did you not push this change? I'm not seeing it

cirq/circuits/circuit.py
Ashalynd: I see. The existing version should allow this already (I don't yet see how to declutter it further though :) )

cirq/circuits/circuit.py
Ashalynd: I did - that's how line 2230 looks for me when I look at changed files in github PR:

```
return (Circuit(m[qubits] for m in self.moments) for qubits in qubit_sets)
```

Ashalynd: If the problem persists, can you let me know what is the latest commit visible to you?

cirq/circuits/circuit.py
Ashalynd: OK, I added a separate method for slice_by_qubit and refactored the common part. Moved everything to AbstractCircuit on suggestion of maffoo, please take another look.

Ashalynd:
> It might be nice to make this a method on `AbstractCircuit` instead, so that it will be available for `Circuit` and `FrozenCircuit`. It could be generic in the `self` type to make the typing work out, something like:
> 
> ```python
> def decompose(self: CIRCUIT_TYPE) -> CIRCUIT_TYPE:
>     ...
>     return [self._with_sliced_moments(m[qubits] for m in self.moments) for qubits in qubit_sets]
> ```
Done, please take another look.

cirq/circuits/circuit.py
```diff
@@ -1387,6 +1389,48 @@ def tetris_concat(
...
+        qubit_factors = self.get_qubit_factors()
+        if len(qubit_factors) == 1:
+            return (self,)
+        return (
+            self._with_sliced_moments([m[qubits] for m in self.moments]) for qubits in qubit_factors
+        )
+
+    def slice_by_qubit(
+        self: CIRCUIT_TYPE, with_qubits: Iterable['cirq.Qid']
+    ) -> Iterable[CIRCUIT_TYPE]:
```
maffoo, line 1422: Not sure I understand the purpose of this function. From the name, I would expect it to return a single circuit that contains gates on only the given qubits, rather than several circuits where we also factorize. In other words, something:
```python
return self._with_sliced_moments([m[with_qubits] for m in self.moments])
```
The resulting circuit could be factorized if desired, but it's not clear to me why the factorization should be done here. There's a question of what to do if the original circuit contains gates on qubits both inside and outside the specified set; in the snippet above the set would be expanded to include other qubits, but one could also imagine either raising an exception in that case, or just dropping any such gates. It might make sense to allow any of those behaviors, selected by an additional arg.

cirq/circuits/circuit.py
```diff
@@ -1387,6 +1389,48 @@ def tetris_concat(
...
+
+        qubit_factors = self.get_qubit_factors()
+        if len(qubit_factors) == 1:
+            return (self,)
+        return (
+            self._with_sliced_moments([m[qubits] for m in self.moments]) for qubits in qubit_factors
+        )
+
+    def slice_by_qubit(
+        self: CIRCUIT_TYPE, with_qubits: Iterable['cirq.Qid']
```
maffoo, line 1421: `Iterable['cirq.Qid']` is probably not the best type to use here, since there's no guarantee that an `Iterable` can be iterated over more than once; it could be a generator that gets exhausted the first time you iterate over it, then is empty if you try to iterate over it again. Something like `Collection['cirq.Qid']` would be better.

cirq/circuits/circuit.py
```diff
@@ -1387,6 +1389,48 @@ def tetris_concat(
 
         return cirq.Circuit(buffer[offset : offset + n_acc])
 
+    def get_qubit_factors(self) -> List[Set['cirq.Qid']]:
```
mpharrigan, line 1392: this may be an unfamiliar name. I don't know if people usually call them "factors". Maybe `get_indepdenent_qubit_sets`

cirq/circuits/circuit.py
```diff
@@ -1387,6 +1389,48 @@ def tetris_concat(
...
+        """
+        Divide circuit's qubits into qubit factors such that there are
```
mpharrigan, line 1394: Try to write docstrings so the first line is only one line long and provides a brief summary. Details can be included in the following paragraph(s). Our style guide says lines can be 100 lines long including docstrings

cirq/circuits/circuit.py
```diff
@@ -1387,6 +1389,48 @@ def tetris_concat(
...
+        circuit's qubits) will be returned.
+        """
+        uf = networkx.utils.UnionFind(self.all_qubits())
+        for op in self.all_operations():
+            if len(op.qubits) > 1:
+                uf.union(*op.qubits)
+        return sorted(uf.to_sets(), key=min)
+
+    def factorize(self: CIRCUIT_TYPE) -> Iterable[CIRCUIT_TYPE]:
+        """Factorize circuit into a sequence of circuits, if it's possible (i.e. if
```
mpharrigan, line 1406: docstring convention

cirq/circuits/circuit.py
Ashalynd: Valid point indeed. I have changed the function to return a single circuit.

The function was introduced as a spinoff to `factorize` om mpharrigan's suggestion (return only the the circuit based on factors containing the specific qubits).

cirq/circuits/circuit.py
Ashalynd: Done.

cirq/circuits/circuit.py
Ashalynd: Done.

cirq/circuits/circuit.py
Ashalynd: Done.

cirq/circuits/circuit.py
Ashalynd: Done.

Ashalynd:
Ping - please review.

cirq/circuits/circuit.py
```diff
@@ -1387,6 +1390,126 @@ def tetris_concat(
...
+        Args:
+            with_qubits: Qubits for which the 'circuit slice' should be created.
+
+        Returns:
+            The circuit, including only the qubits which are entangled with the
+            provided qubits.
+
+        """
+
+        if not frozenset(with_qubits) < self.all_qubits():
```
mpharrigan, line 1503: should this be `<=`? Instead of `not <` would it be easier to read as `>`?

cirq/circuits/circuit.py
```diff
@@ -1387,6 +1390,126 @@ def tetris_concat(
...
+            raise ValueError(f'Unknown qubits:{with_qubits}')
```
mpharrigan, line 1504: space between colon and variable

cirq/circuits/circuit.py
```diff
@@ -1387,6 +1390,126 @@ def tetris_concat(
...
+        """
+
+        qubit_factors = self.get_independent_qubit_sets()
+        if len(qubit_factors) == 1:
+            return (self,)
+        return (
+            self._with_sliced_moments([m[qubits] for m in self.moments]) for qubits in qubit_factors
+        )
+
+    def slice_by_qubits(self: CIRCUIT_TYPE, with_qubits: Collection['cirq.Qid']) -> CIRCUIT_TYPE:
```
mpharrigan, line 1469: per the other conversations, I think there's been a misunderstanding about where why and how this function should exist. As currently written, it is misleading because it does not indeed slice by qubits. It will sortof partially factor a circuit. See other comment for more. We can get rid of this function as out of scope.

cirq/circuits/circuit.py
```diff
@@ -1387,6 +1390,126 @@ def tetris_concat(
...
+            The sequence of circuits, each including only the qubits from one
+            independent qubit set.
+
+        """
+
+        qubit_factors = self.get_independent_qubit_sets()
+        if len(qubit_factors) == 1:
+            return (self,)
+        return (
+            self._with_sliced_moments([m[qubits] for m in self.moments]) for qubits in qubit_factors
```
mpharrigan, line 1466: I guess I originally had the idea that

```python
self._with_sliced_moments([m[qubits] for m in self.moments])
```

could be the implementation of `slice_by_qubits`. E.g. if you already know your circuit is composed of independent factors. It would raise an exception otherwise. The `factorize` function can use it.

However, the semantics of `Moment.__getitem__` will not raise an exception. Rather, it will return all operations on which any of the qubits operate. We should probably add a comment to this line explaining the semantics of `m[qubits]`, namely that since we've already found the independent qubit sets we're guaranteed to get moments that *only* operate on the given qubits.

mpharrigan (changes requested):

cirq/circuits/circuit.py
```diff
@@ -1387,6 +1390,126 @@ def tetris_concat(
...
+            The list of independent qubit sets.
+
+        """
+        uf = networkx.utils.UnionFind(self.all_qubits())
+        for op in self.all_operations():
+            if len(op.qubits) > 1:
+                uf.union(*op.qubits)
+        return sorted([qs for qs in uf.to_sets()], key=min)
+
+    def factorize(self: CIRCUIT_TYPE) -> Iterable[CIRCUIT_TYPE]:
```
mpharrigan, line 1425: Be general in input (argument) types, specific in return types. Suggest changing Iterable to `Tuple`

cirq/circuits/circuit.py
Ashalynd: I see. I thought that the intent was to get the part of the circuit interacting with the specified qubit. Will remove it then.

cirq/circuits/circuit.py
Ashalynd: Done.

cirq/circuits/circuit.py
Ashalynd: I see, thanks for the explanations. Will remove `slice_by_qubits` method.

cirq/circuits/circuit.py
Ashalynd: No longer applicable (method is gone).

cirq/circuits/circuit.py
Ashalynd: Yeah. Anyways, gone now :)

Ashalynd:
Thanks for the review!

cirq/circuits/circuit.py
Ashalynd: Actually, this won't work (mypy becomes uphappy). Also, it doesn't produce a Tuple, it produces a Generator.

Ashalynd:
@mpharrigan ptal

Ashalynd:
@mpharrigan or @maffoo , PTAL

cirq/circuits/circuit.py
mpharrigan: except when there's only one instance; then it does in fact return a tuple. Which is iterable so mypy is happy

mpharrigan (approved):
Lgtm

PR #3874, opened 2021-03-03
Move decompose-and-act-on strategy to `act_on` file and reuse it in all `ActOn*Args`

smitsanghavi:
This change allows new gates to just define their decompositions and expect to be supported on relevant simulators. For e.g. `CliffordSimulator` now supports `ISWAP` without explicitly having to define `_act_on_` for it since ISWAP has a decomposition to Clifford gates specified.


smitsanghavi:
#2423

daxfohl:
This is already done in #3841

smitsanghavi:
> This is already done in #3841

Thanks! I'll close this one then.

Issue #3852, opened 2021-02-24
Add empty results fallback for Result.repetitions


Strilanc:
Currently cirq.Result.repetitions raises an error when called on an empty result. It should instead return 0.

PR #3875, opened 2021-03-03
Add empty results fallback for Result.repetitions

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/3852


cirq/study/result.py
```diff
@@ -150,8 +150,11 @@ def measurements(self) -> Dict[str, np.ndarray]:
 
     @property
     def repetitions(self) -> int:
+        measurement_values = self.measurements.values()
```
balopat, line 153: can self.measurement itself be `None`?

balopat (changes requested):
Thanks for jumping in - added a nit

balopat:
I'm gonna press the update branch button to nudge Github Actions

cirq/study/result.py
vtomole: Added check for this. :+1:

cirq/study/result_test.py
```diff
@@ -158,6 +158,16 @@ def test_multi_measurement_histogram():
         }
     )
 
+    result = cirq.Result.from_single_parameter_set(
+        params=cirq.ParamResolver({}),
+        measurements={},
+    )
```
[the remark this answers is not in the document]
vtomole: Yup. I'm just following the same pattern that every assert in this `def test_multi_measurement_histogram` follows. Just modified the test.

cirq/study/result_test.py
```diff
@@ -158,6 +158,16 @@ def test_multi_measurement_histogram():
         }
     )
```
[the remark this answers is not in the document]
vtomole: Done

cirq/study/result.py
```diff
@@ -150,6 +150,8 @@ def measurements(self) -> Dict[str, np.ndarray]:
 
     @property
     def repetitions(self) -> int:
+        if not self.measurements or not self.measurements.values():
```
balopat, line 153: I think the empty dict is Falsey so we could just use this: 
```suggestion
        if not self.measurements:
```

balopat (approved):
LGTM with the last nit to simplify the if statement

cirq/study/result.py
vtomole: Ah. Thanks.

PR #3876, opened 2021-03-04
Fix release scripts

tanujkhattar:
- Updates pytest version because existing version has a bug where --ignore= doesn't work. 
- Fixed verify-published-package script to install pip dev dependencies. 
- Updated setup.py to include `cirq.google.json_test_data`

### DIFF
--- a/dev_tools/packaging/verify-published-package.sh
+++ b/dev_tools/packaging/verify-published-package.sh
@@ -83,8 +83,7 @@ for PYTHON_VERSION in python3; do
     # Install package.
     if [ "${PYPI_REPO_NAME}" == "TEST" ]; then
         echo "Pre-installing dependencies since they don't all exist in TEST pypi"
-        "${tmp_dir}/${PYTHON_VERSION}/bin/pip" install --quiet -r "${RUNTIME_DEPS_FILE}"
-        "${tmp_dir}/${PYTHON_VERSION}/bin/pip" install --quiet -r "${DEV_DEPS_FILE}"
+        "${tmp_dir}/${PYTHON_VERSION}/bin/pip" install --quiet -r "${RUNTIME_DEPS_FILE}" -r "${DEV_DEPS_FILE}"
     fi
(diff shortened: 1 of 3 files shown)


dev_tools/packaging/verify-published-package.sh
```diff
@@ -83,6 +84,7 @@ for PYTHON_VERSION in python3; do
     if [ "${PYPI_REPO_NAME}" == "TEST" ]; then
         echo "Pre-installing dependencies since they don't all exist in TEST pypi"
         "${tmp_dir}/${PYTHON_VERSION}/bin/pip" install --quiet -r "${RUNTIME_DEPS_FILE}"
+        "${tmp_dir}/${PYTHON_VERSION}/bin/pip" install --quiet -r "${DEV_DEPS_FILE}"
```
maffoo, line 87: nit: can install both together
```suggestion
        "${tmp_dir}/${PYTHON_VERSION}/bin/pip" install --quiet -r "${RUNTIME_DEPS_FILE}" -r ${DEV_DEPS_FILE}
```
This might be more efficient, since all dependencies are resolved at the same time.

setup.py
```diff
@@ -77,5 +77,6 @@
         'cirq.google.api.v1': ['*.proto', '*.pyi'],
         'cirq.google.api.v2': ['*.proto', '*.pyi'],
         'cirq.protocols.json_test_data': ['*'],
+        'cirq.google.json_test_data': ['*'],
```
maffoo, line 80: nit: sort alphabetically

maffoo (approved):
LGTM with minor nits

dev_tools/packaging/verify-published-package.sh
tanujkhattar: done

setup.py
tanujkhattar: done

balopat (approved):

PR #3877, opened 2021-03-04
Add @mrwojtek to CODEOWNERS for cirq/docs/qcvv

mpharrigan:


tanujkhattar (approved):
LGTM

balopat (approved):
You'll have to fix the codeowners test

mpharrigan:
should we be concerned that the codeowners test wasn't failing?

mpharrigan:
see https://github.com/quantumlib/Cirq/pull/3878

Issue #3886, opened 2021-03-08
common_channels don't have approximate equality.


MichaelBroughton:
Question: Is this supposed to be False ?

```python3
>>> a = cirq.depolarize(p=0.009999999776482582)
>>> b = cirq.depolarize(p=0.01)
>>> cirq.approx_eq(a, b, atol=1e-2)
False
```
and
```python3
>>> a = cirq.depolarize(p=0.009999999776482582)(cirq.GridQubit(0, 0))
>>> b = cirq.depolarize(p=0.01)(cirq.GridQubit(0, 0))
>>> cirq.approx_eq(a, b, atol=1e-2)
False
```

vtomole:
The `DepolarizingChannel` class doesn't have an `approx_eq()`. I've opened https://github.com/quantumlib/Cirq/pull/3887 as a potential fix.

MichaelBroughton:
Ahhh thx for digging into it a little more. I took a quick look in the files in your PR and it looks like none of the channels have this `approx_eq` method. Going to re-title this to encompass the larger issue. Would you mind expanding your PR to cover all the channels in `common_channels.py` ? I'd be happy to review.

PR #3887, opened 2021-03-08
Add approximate equality for all common_channels except `ResetChannel`

vtomole:
Fixes: https://github.com/quantumlib/Cirq/issues/3886


95-martin-orion:
It seems worthwhile (and fairly cheap) to support `_approx_eq_` in the other channels as well - could you add that to this PR?

Strilanc (approved):
The other channels can be fixed in other PRs. No need expand the size.

Issue #3833, opened 2021-02-18
random_quantum_circuit_generation.random_rotations_between_grid_interaction_layers_circuit should use device_graph instead of qubits


mpharrigan:
The arguments for `random_rotations_between_grid_interaction_layers_circuit` combines `qubits` with `pattern: Sequence[GridInteractionLayer]` to figure out what pairs to activate. 

This assumes that all neighboring GridQubits are able to be activated, which may not always be the case. It also prevents easy adaptability to other topologies or other qubit types. [Granted, this would be a little more work since `GridInteractionLayer` is pretty griddy, but we really only use a containment check so other types of `xxInteractionLayer`s could be introduced]

The newer parallel-xeb fidelity code uses a `device_graph` argument where nodes are qubits and edges give couplable pairs. We should modify the code to accept a device graph. This can be done in a backwards compatible way by checking if qubits were passed and using `cirq.contrib.routing.gridqubits_to_graph_device`, which should be made faster and put into regular cirq

mpharrigan:
There's also a problem if you are operating on a topology that is a subset of a grid, for example: a line. The random circuit function for cycles divisible by 4 + 1 or 2 may not add any entangling layers. 

Options:
1. have the random generation function ignore the gridinteractionlayers that don't produce interactions. This would solve the issue at hand but wouldn't produce optimal circuits if you had a line of qubits that was snaked on a diagonal or something
2. Have a special function or pattern to do random circuits on linear topologies
3. Have a be-all-end-all random circuit generation function that works on arbitrary planar topologies by doing edge coloring to figure out activation orders.
4. Enforce cycles divisible by len(pattern)

mhucka:
Closing after consultation with people. It seems this method is only used in tests and a couple of notebooks.

PR #3888, opened 2021-03-08
Remove cirq.google reference from contrib.quantum_volume and contrib.routing

balopat:
Deprecates `cirq.contrib.routing.xmon_device_to_graph` and removes references to cirq.google in `quantum_volume` and `routing` packages.

Related to #3737.

cc @mpharrigan


cirq/contrib/quantum_volume/quantum_volume.py
```diff
@@ -474,11 +484,7 @@ def calculate_quantum_volume(
     )
 
     # Get the device graph from the given qubits or device.
-    device_graph = None
-    if isinstance(device_or_qubits, list):
-        device_graph = ccr.gridqubits_to_graph_device(device_or_qubits)
-    else:
-        device_graph = ccr.xmon_device_to_graph(device_or_qubits)
+    device_graph = ccr.gridqubits_to_graph_device(device_qubits)
```
95-martin-orion, line 487: Copy-paste from #3894:

This is a breaking change. It's contrib, so I'm not deeply worried about breaking users, but having `deprecated_parameter` seems to imply backwards-compatibility that this doesn't provide.

My recommendation is to remove the deprecation warning, since we can't really make this backwards-compatible without a `cirq.google` dependency.

95-martin-orion (changes requested):
Only one item catches my eye here, otherwise looks good to go.

cirq/contrib/quantum_volume/quantum_volume.py
balopat: sounds good!

cirq/contrib/quantum_volume/quantum_volume.py
```diff
@@ -422,12 +423,22 @@ def execute_circuits(
...
+        {('device_qubits' if k == 'device_or_qubits' else k): v for k, v in kwargs.items()},
+    ),
+)
 def calculate_quantum_volume(
     *,
     num_qubits: int,
     depth: int,
     num_circuits: int,
-    device_or_qubits: Union[cirq.google.XmonDevice, List[cirq.GridQubit]],
+    device_qubits: List[cirq.GridQubit],
```
mpharrigan, line 450: can we change this to `device_graph: nx.Graph`? All of Cirq is going to come crashing down the instant we have a device with a broken coupler. An immediate benefit is supporting qubits other than gridqubit.

cirq/contrib/quantum_volume/quantum_volume.py
balopat: oh I realized that my deprecation can take care of this conversion path.

dev_tools/notebooks/isolated_notebook_test.py
```diff
@@ -36,7 +36,12 @@
 # these notebooks rely on features that are not released yet
 # after every release we should raise a PR and empty out this list
 # note that these notebooks are still tested in dev_tools/notebook_test.py
-NOTEBOOKS_DEPENDING_ON_UNRELEASED_FEATURES = []
+NOTEBOOKS_DEPENDING_ON_UNRELEASED_FEATURES = [
```
mpharrigan, line 39: is there something in the release procedure to check that that these get removed after a release?

mpharrigan (commented):
suggest using graphs instead of topology implicit in gridqubits. I think this should be an easy change.

dev_tools/notebooks/isolated_notebook_test.py
balopat: it's all manual, but the instructions are there in release.md

cirq/contrib/quantum_volume/quantum_volume.py
balopat: I'm down - I actually really like the lightweight implementation in the routing package for the graph device...

mpharrigan: I use it all the time (once you internalize the sometimes idiosyncratic naming of the functions). See here for a related issue: https://github.com/quantumlib/Cirq/issues/3833

gridqubits_to_graph_device is implemented fairly inefficiently at present.

cirq/contrib/quantum_volume/quantum_volume.py
```diff
@@ -449,8 +470,7 @@ def calculate_quantum_volume(
         depth: The number of gate layers to generate.
         num_circuits: The number of random circuits to run.
         random_state: Random state or random state seed.
-        device_or_qubits: The device or the device qubits to run the compiled
-            circuit on.
+        device_graph: The device or qubits to run the compiled circuit on.
```
mpharrigan, line 473:
```suggestion
        device_graph: A graph whose nodes are qubits and edges represent two qubit interactions to run the compiled circuit on.
```

mpharrigan (approved):

95-martin-orion (approved):

PR #3889, opened 2021-03-09
narrow qiskit dependency

balopat:
Narrows qiskit dependency for dev tools. 
We don't need the whole qiskit suite for testing, quiskit-aer (and terra transitively) is sufficient.


vtomole (approved):
LGTM

Issue #3866, opened 2021-02-26
N-qubit diagonal gate


balopat:
**Is your feature request related to a use case or problem? Please describe.**

Quantum oracles implementing functions of type f: {0,1}^n -> {1,-1} (n is the number of qubits) can be represented as diagonal gates. Currently Cirq only has a ThreeQubitDiagonalGate. For larger oracles one can use the MatrixGate, however there is no decomposition logic implemented for it. 

**Describe the solution you'd like**

1. `DiagonalGate`
```
qubits = cirq.LineQubit.range(6)
c = cirq.Circuit(cirq.DiagonalGate([1,-1,1,1,1,-1])(qubits)) 
print(c) 
```

Should provide a meaningful circuit diagram. 

2. Decomposition 
```
a,b,c = cirq.LineQubit.range(3)
cirq.decompose(cirq.DiagonalGate([1,-1,1])(a,b,c)) 
```
should provide a list of 1 and 2 qubit gates that allows for implementation of this gate on hardware. 

**[optional] Additional context (e.g. screenshots)**

This came up during implementing the Fourier checking algo as a tutorial: https://github.com/quantumlib/Cirq/issues/3687

**What is the urgency from your perspective for this issue? Is it blocking important work?**

P3 - I'm not really blocked by it, it is an idea I'd like to discuss / suggestion based on principle

vtomole:
It looks like https://github.com/quantumlib/Cirq/pull/3890 took care of this. Please re-open if there is more work to do here.

PR #3890, opened 2021-03-09
Diagonal gate

ybc1991:
Add an N-qubit DiagonalGate, which is specified in #3866. The decomposition algorithm (2^n-1 Rz gates and 2^n -2 CNOT gates) is implemented based on the paper:

> Welch, Jonathan, et al. _"Efficient quantum circuits for diagonal unitaries without ancillas."_ New Journal of Physics 16.3 (2014): 033040.  https://iopscience.iop.org/article/10.1088/1367-2630/16/3/033040/meta

For the step-by-step explanation of the implemention, please refer [this colab](https://colab.research.google.com/drive/1Id_8oSISppCQpY2Uk0XsjYJWYMSh8DC-?usp=sharing).

A current known issue is that the decomposition algorithm has a global phase shift between the diagonal unitary matrix and the decomposed circuit implementation. It can produce a surprising effect if the diagonal gates is used within a partial system, such as under the controlled gate. Hence, a GlobalPhaseOperation is added into the \_decompose\_ method. But GlobalPhaseOperation doesn't have as universal support as the single-/two-qubit gates. So adding GlobalPhaseOperation may limit the usage of DiagonalGate.


vtomole:
Call the file and classes `diagonal_gate` instead of `n_qubit_diagonal_gate`.

cirq/ops/n_qubit_diagonal_gate_test.py
```diff
@@ -0,0 +1,130 @@
...
+# limitations under the License.
+
+from typing import List
+import numpy as np
+import pytest
+import sympy
+
+import cirq
+
+_candidate_angeles: List[float] = [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53]
```
balopat, line 22:
```suggestion
_candidate_angles: List[float] = [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53]
```

cirq/ops/n_qubit_diagonal_gate.py
```diff
@@ -0,0 +1,188 @@
...
+        The algorithm is implemented according to the paper:
+            Welch, Jonathan, et al. "Efficient quantum circuits for diagonal unitaries without
+            ancillas." New Journal of Physics 16.3 (2014): 033040.
+            https://iopscience.iop.org/article/10.1088/1367-2630/16/3/033040/meta
+        """
+        n = self._num_qubits_()
+        hat_angles = _fast_walsh_hadamard_transform(self._diag_angles_radians) / (2 ** n)
+
+        # There is one global phase shift between unitary matrix of the diagonal gate and the
+        # decomposed gates. It is not a problem usually. However, if using this diagonal gate
```
balopat, line 178:
```suggestion
        # decomposed gates. On its own it is not physically observable. However, if using this diagonal gate
```

cirq/ops/n_qubit_diagonal_gate.py
```diff
@@ -0,0 +1,188 @@
...
+        # for sub-system like controlled gated, it is no longer equivalent. Hence, we add
```
balopat, line 179:
```suggestion
        # for sub-system like controlled gate, it is no longer equivalent. Hence, we add
```

cirq/ops/n_qubit_diagonal_gate.py
```diff
@@ -0,0 +1,188 @@
...
+        # global phase shift back.
```
balopat, line 180:
```suggestion
        # global phase.
```

cirq/ops/n_qubit_diagonal_gate.py
```diff
@@ -0,0 +1,188 @@
...
+        return protocols.CircuitDiagramInfo(
+            [diag_str] + ['#' + str(i) for i in range(2, self._num_qubits_() + 1)]
+        )
+
+    def __pow__(self, exponent: Any) -> 'DiagonalGate':
+        if not isinstance(exponent, (int, float, sympy.Basic)):
+            return NotImplemented
+        angles = []
+        for angle in self._diag_angles_radians:
+            mulAngle = protocols.mul(angle, exponent, NotImplemented)
```
balopat, line 136: Prefer snake case over camel case
```suggestion
            mul_angle = protocols.mul(angle, exponent, NotImplemented)
```

cirq/ops/n_qubit_diagonal_gate.py
```diff
@@ -0,0 +1,188 @@
...
+                a_[j] = x + y
+                a_[j + h] = x - y
+        h *= 2
+    return a_
+
+
+def _gen_gray_code(n: int):
+    """Generate the Gray Code from 0 to 2^n-1.
+
+    Each iteration returns two elements. The first element is the decimal representation
```
balopat, line 51:
```suggestion
    Each iteration yields a two-tuple, `(gray_code, bit_flip)`. `gray_code` is the decimal representation
```

cirq/ops/n_qubit_diagonal_gate.py
```diff
@@ -0,0 +1,188 @@
...
+    of the gray code and the second one is the position of bits flipped for next gray code.
```
balopat, line 52:
```suggestion
    of the gray code and `bit_flip` is the position of bits flipped for next gray code.
```

cirq/ops/n_qubit_diagonal_gate.py
```diff
@@ -0,0 +1,188 @@
...
+from cirq import protocols, value
+from cirq._compat import proper_repr
+from cirq.ops import common_gates, raw_types, global_phase_op
+
+if TYPE_CHECKING:
+    import cirq
+
+
+def _fast_walsh_hadamard_transform(a: Tuple[float]) -> np.array:
+    """Fast Walsh–Hadamard Transform of array a."""
```
balopat, line 34:
```suggestion
    """Fast Walsh–Hadamard Transform of an array."""
```

cirq/ops/n_qubit_diagonal_gate.py
```diff
@@ -0,0 +1,188 @@
...
+        angles = []
+        for angle in self._diag_angles_radians:
+            mulAngle = protocols.mul(angle, exponent, NotImplemented)
+            angles.append(mulAngle)
+        return DiagonalGate(angles)
+
+    def _value_equality_values_(self) -> Any:
+        return tuple(self._diag_angles_radians)
+
+    def _decompose_for_basis(self, index, bit_flip, theta, qubits):
```
balopat, line 143: please add type annotations

cirq/ops/n_qubit_diagonal_gate.py
```diff
@@ -0,0 +1,188 @@
...
+            return []
+        largest_digit = self._num_qubits_() - (len(bin(index)) - 2)
+        yield common_gates.rz(2 * theta)(qubits[largest_digit])
+        _flip_bit = self._num_qubits_() - bit_flip - 1
+        if _flip_bit < largest_digit:
+            yield common_gates.CNOT(qubits[largest_digit], qubits[_flip_bit])
+        elif _flip_bit > largest_digit:
+            yield common_gates.CNOT(qubits[_flip_bit], qubits[largest_digit])
+
+    def _decompose_(self, qubits):
```
balopat, line 154: please add type annotations

cirq/ops/n_qubit_diagonal_gate.py
```diff
@@ -0,0 +1,188 @@
...
+
+        For n qubits decomposition looks similar but with 2^n-1 Rz gates and 2^n-2 CNOT gates.
+
+        The algorithm is implemented according to the paper:
+            Welch, Jonathan, et al. "Efficient quantum circuits for diagonal unitaries without
+            ancillas." New Journal of Physics 16.3 (2014): 033040.
+            https://iopscience.iop.org/article/10.1088/1367-2630/16/3/033040/meta
+        """
+        n = self._num_qubits_()
+        hat_angles = _fast_walsh_hadamard_transform(self._diag_angles_radians) / (2 ** n)
```
balopat, line 175: I wonder how this behaves when the gate is parametrized. I think it would blow up. To avoid this we should add here: 

```
        if protocols.is_parameterized(self):
            return NotImplemented
```

Also, we'll need tests for that.

balopat (changes requested):
Looking great, I found a bunch of nits though. +1 to renaming the file to `diagonal_gate.py`.

cirq/ops/n_qubit_diagonal_gate.py
ybc1991: Done.

cirq/ops/n_qubit_diagonal_gate.py
ybc1991: Done.

cirq/ops/n_qubit_diagonal_gate.py
ybc1991: Yeah, the current implementation cannot work with the parameterized cases. (Actually, it is still possible as long as we make symbol-based Hardmard transform.) I added the NotImplemented case. One thing I felt strange is

```
cirq.decompose(diagonal_gate(*cirq.LineQubit.range(2))) == []  # <-- parameterized gate
```

Note the result is [] instead of raising the error. Is that expected?

ybc1991:
Thanks for reviewing! The file has been renamed to diagonal_gate.  About parameterized DiagonalGate comment I have another question. Please take a look again.

cirq/ops/n_qubit_diagonal_gate.py
balopat: That is not my expectation - I'd expect `[DiagonalGate(qubits)]` - i.e. the undecomposed op ... Maybe it's a bug in the decompose protocol?

95-martin-orion: The issue is that you're `return`ing in a generator. When `decompose` encounters this, it attempts to iterate through the generator, encounters the return, and stops without `yield`ing anything.

See [this SO question](https://stackoverflow.com/questions/26595895/return-and-yield-in-the-same-function) for more information. TL;DR: the value of a `return` is ignored in a generator.

balopat: Great catch, thank you! 🙇‍♂️

ybc1991: Then, what should be the expected behavior we want here?

Some options:
1. Raise the error directly?
2. Make everything return a list instead of yield
3. Support the parameterized decomposition. It is easy to support except for one piece: GlobalPhaseOperation. It does not support sympy.exp(1j*xx) as input due to the unity check. [Awkward]

balopat: Option 2 is the right solution!

cirq/ops/diagonal_gate_test.py
```diff
@@ -77,9 +77,14 @@ def test_decomposition_diagonal_exponent(n):
...
-def test_disallow_decomposition_with_parametrized():
+def test_decomposition_with_parameterization():
     diagonal_gate = cirq.DiagonalGate([2, 3, 5, sympy.Symbol('a')])
-    assert not cirq.decompose(diagonal_gate(*cirq.LineQubit.range(2)))
+    op = diagonal_gate(*cirq.LineQubit.range(2))
+
+    # We do not support the decomposition of parameterized case yet.
+    # So cirq.decompose should do nothing.
+    assert len(cirq.decompose(op)) == 1
+    assert cirq.decompose(op)[0] == op
```
balopat, line 88: you can simplify these two lines to `assert cirq.decompose(op) == [op]`

balopat (approved):
LGTM, great! One more tiny nit

cirq/ops/diagonal_gate.py
```diff
@@ -0,0 +1,196 @@
...
+        n = self._num_qubits_()
+        hat_angles = _fast_walsh_hadamard_transform(self._diag_angles_radians) / (2 ** n)
+
+        # There is one global phase shift between unitary matrix of the diagonal gate and the
+        # decomposed gates. On its own it is not physically observable. However, if using this
+        # diagonal gate for sub-system like controlled gate, it is no longer equivalent. Hence,
+        # we add global phase.
+        decomposed_circ = [global_phase_op.GlobalPhaseOperation(np.exp(1j * hat_angles[0]))]
+        for i, bit_flip in _gen_gray_code(n):
+            decomposed_circ.extend(self._decompose_for_basis(i, bit_flip, -hat_angles[i], qubits))
```
balopat, line 190:
```suggestion
            decomposed_circ.append(self._decompose_for_basis(i, bit_flip, -hat_angles[i], qubits))
```

balopat: oh, or `+=`

balopat:
One more thing: cirq.contrib.acquaintance.executor_test.py has a test DiagonalGate defined. Can you replace that with the new shiny DiagonalGate? Thanks!

balopat:
One more thing: cirq.contrib.acquaintance.executor_test.py has a test DiagonalGate defined. Can you replace that with the new shiny DiagonalGate? Thanks!

cirq/ops/diagonal_gate.py
ybc1991: Actually, question: this surely works but why prefer yy += list(xx) than yy.extend(xx)?

balopat: `extend(list(xx))` would work - `+=` is shorter...but without the `list()` it is failing - CI checks are red because of that.

cirq/ops/diagonal_gate_test.py
ybc1991: Done.

cirq/ops/diagonal_gate.py
ybc1991: I think it is the invariance list assumption in mypy. I think denote the decomposed_circ: List[Any] will solve the problem.

ybc1991:
> One more thing: cirq.contrib.acquaintance.executor_test.py has a test DiagonalGate defined. Can you replace that with the new shiny DiagonalGate? Thanks!

Sure. Done.

daxfohl:
Over in qiskit they mention Hidden Shift can be represented by a diagonal gate https://qiskit.org/documentation/stubs/qiskit.circuit.library.Diagonal.html, but I'm not grokking. We have a hidden shift example https://github.com/quantumlib/Cirq/blob/master/examples/hidden_shift_algorithm.py. Should it be updated to use this (or perhaps Hamiltonian Binary since you mention they have the same underlying math)?

ybc1991:
> Over in qiskit they mention Hidden Shift can be represented by a diagonal gate https://qiskit.org/documentation/stubs/qiskit.circuit.library.Diagonal.html, but I'm not grokking. We have a hidden shift example https://github.com/quantumlib/Cirq/blob/master/examples/hidden_shift_algorithm.py. Should it be updated to use this (or perhaps Hamiltonian Binary since you mention they have the same underlying math)?

Hi Dax,

Right, simulating any oracle problem, like `f(x) |x>`, is equivalent to a `2^n * 2^n` diagonal unitary matrix times a `2^n` vector. In this point of view, yes we can use a diagonal gate in the Hidden Shift example. However, in this example, I think it is ok to leave it as is mainly because  the oracle function consider in the example is simply
```
def make_oracle_f(qubits):
    """Implement function {f(x) = Σ_i x_(2i) x_(2i+1)}."""
    return [cirq.CZ(qubits[2 * i], qubits[2 * i + 1]) for i in range(len(qubits) // 2)]
```
Note `CZ` itself is already a diagonal gate.  This is clean and good for example illustration. The benefit of using a diagonal gate will exhibit when we want to use other more complicated functions that we know `f(x)` but don't know how to write into existing primary gates easily. Similar opinion on Hamiltonian Binary. (I.e. if we know the oracle function boolean expression.)

Qiskit Diagonal gate has more functionalities (we have dense unitary and 1, 2-qubits decomposite representation only now). We can also consider making it more generalized first.

PR #3891, opened 2021-03-09
CircuitOperation proto

95-martin-orion:
Proposed changes to the v2 API to support CircuitOperations.

I've made an effort to keep this extensible, particularly regarding measurement keys and repetitions as both are expected to change in the future. Please let me know if you see any potential pain points for future changes in this space.


95-martin-orion:
Initial prototypes for (de-)serializers using this format are promising as far as backwards-compatibility is concerned. I haven't been able to fully test them yet, but I'd like to get this solidified before committing too much time towards it.

95-martin-orion:
This PR is part of #3634.

95-martin-orion:
See [this follow-up PR](https://github.com/95-martin-orion/Cirq/pull/1) on my fork for how this looks when used in serialization.

cirq/google/api/v2/program.proto
```diff
@@ -64,6 +66,11 @@ message Moment {
   // All of the operations in the moment. Each operation must act on different
   // qubits.
   repeated Operation operations = 1;
+
+  // As 'operations', but may include CircuitOperations (i.e. operations that
+  // contain subcircuits). If 'operations' is non-empty, this field will be
+  // ignored.
+  repeated GenericOperation generic_operations = 2;
```
dstrain115, line 73: What is the purpose of adding GenericOperation?  Why not just put repeated CircuitOperations straight into Moment?  I am not sure GenericOperation really adds much here and the name is not very descriptive either.

cirq/google/api/v2/program.proto
```diff
@@ -198,3 +216,81 @@ message ArgFunction {
...
+  // this operation.
+  MeasurementKeyMapping measurement_key_map = 4;
+
+  // Map of args in the circuit to the values (or replacement args) used in
+  // this operation.
+  ArgMapping arg_map = 5;
+}
+
+// A description of the repetitions of a subcircuit.
+message RepetitionSpec {
```
dstrain115, line 242: Probably best to avoid abbreviations "RepetitionSpecification".

cirq/google/api/v2/program.proto
```diff
@@ -198,3 +216,81 @@ message ArgFunction {
...
+  // An ordered list of IDs for a sequence of repetitions.
+  message RepetitionIds {
```
dstrain115, line 244: Do we think we will eventually change this or add onto it?

cirq/google/api/v2/program.proto
```diff
@@ -198,3 +216,81 @@ message ArgFunction {
...
+  // Map of measurement keys in the circuit to the measurement keys used in
+  // this operation.
+  MeasurementKeyMapping measurement_key_map = 4;
+
+  // Map of args in the circuit to the values (or replacement args) used in
+  // this operation.
+  ArgMapping arg_map = 5;
+}
+
+// A description of the repetitions of a subcircuit.
```
dstrain115, line 241: Explaining this in more detail (maybe a bit of motivation as to why each repetition needs an id) may be a good idea.

cirq/google/api/v2/program.proto
```diff
@@ -198,3 +216,81 @@ message ArgFunction {
...
+
+// A mapping of qubits from one value to another.
+message QubitMapping {
+  // Indicates that qubit "first" should be replaced with "second".
+  message QubitPair {
+    Qubit first = 1;
+    Qubit second = 2;
+  }
+
+  // A list of qubit mappings to apply.
```
dstrain115, line 265: Is it guaranteed that qubit pairs are applied in order (in case of overlapping pairs)?  Or are overlapping pairs not allowed?

cirq/google/api/v2/program.proto
```diff
@@ -198,3 +216,81 @@ message ArgFunction {
...
+    RepetitionIds rep_ids = 1;
+
+    // An integer number of repetitions to perform.
+    int32 rep_count = 2;
+  }
+}
+
+// A mapping of qubits from one value to another.
+message QubitMapping {
+  // Indicates that qubit "first" should be replaced with "second".
```
dstrain115, line 259: I wonder if first and second should be replaced with more descriptive terms.  Perhaps, source and target?  Or initial and mapped qubits?

Same goes for below.

cirq/google/api/v2/program.proto
```diff
@@ -198,3 +216,81 @@ message ArgFunction {
...
+message CircuitOperation {
+  // The index of the circuit in the top-level constant table.
+  int32 circuit_constant_index = 1;
+
+  // Specifier for repetitions of the circuit, including number of reps and
+  // repetition IDs.
+  RepetitionSpec repetition_spec = 2;
+
+  // Map of qubits in the circuit to the qubits used in this operation.
+  QubitMapping qubit_map = 3;
```
dstrain115, line 230: Why do we need a qubit mapping and measurement key mapping in the serialized form?

Why can't we just map the qubits and keys while serializing?  Is this because we want to re-use the circuit in the index multiple times with different mappings?

cirq/google/api/v2/program.proto
```diff
@@ -198,3 +216,81 @@ message ArgFunction {
   // The arguments to the function.
   repeated Arg args = 2;
 }
+
+// An operation that describes a circuit.
```
dstrain115, line 220: I would add more description here.  Especially since this doesn't actually describe a circuit.  It references a circuit and then modifies it according to a mapping.

cirq/google/api/v2/program.proto
95-martin-orion: Using repeated CircuitOperation would either (1) imply that a moment can only contain either Operations or CircuitOperations, or (2) require some implicit ordering between Operations and CircuitOperations. Both feel like improper uses of Moment.

What I would've preferred is to call this `Operation`, with the current operation renamed to `GateOperation`...but past experience with proto-renaming tells me that would be a nightmare for backwards-compatibility.

cirq/google/api/v2/program.proto
95-martin-orion: It's possible, given that we haven't decided on an implementation for `while` loops or variable-length `for` loops.

Come to think of it, this should probably be a `oneof` with `rep_count`, since only one of them is necessary to specify repetitions (`rep_count` implies default repetition ids)

cirq/google/api/v2/program.proto
95-martin-orion: > Is this because we want to re-use the circuit in the index multiple times with different mappings?

This is correct. Consider the case of three `CircuitOperation`s {A, B, C}, each with the same underlying circuit M but with different qubit mappings. Serializing the maps allows these `CircuitOperation`s to share a single circuit.

A real-world example of this is two-qubit XEB, where we want to run a specific circuit (XEB) on a whole bunch of qubits.

cirq/google/api/v2/program.proto
95-martin-orion: Correction: overlaping pairs **are** allowed, but all pairs are applied simultaneously. {a: b, b: a} swaps a and b.

cirq/google/api/v2/program.proto
95-martin-orion: Done.

cirq/google/api/v2/program.proto
95-martin-orion: Done.

cirq/google/api/v2/program.proto
95-martin-orion: Added clearer documentation.

cirq/google/api/v2/program.proto
95-martin-orion: Done.

cirq/google/api/v2/program.proto
95-martin-orion: Done, here and elsewhere.

cirq/google/api/v2/program.proto
maffoo: I've been thinking of this as "inner" and "outer", where "inner" refers to the identifier inside the circuit operation, and "outer" refers to the identifier in the circuit in which that op is embedded.

cirq/google/api/v2/program.proto
```diff
@@ -198,3 +216,81 @@ message ArgFunction {
...
+
+    // An integer number of repetitions to perform.
+    int32 rep_count = 2;
+  }
+}
+
+// A mapping of qubits from one value to another.
+message QubitMapping {
+  // Indicates that qubit "first" should be replaced with "second".
+  message QubitPair {
```
maffoo, line 269: nit: to me the term "pair" implies symmetry between the two items. I think the "inner/outer" terminology suggested above would help indicate the directionality, but perhaps we could call this something else like "QubitRelabeling"?

cirq/google/api/v2/program.proto
95-martin-orion: > Is the idea here that the `operations` field should be deprecated and we should put everything into `generic_operations` instead {...}?

That was my original thought, yes.

> I think just have two lists of `operations` and `circuit_operations` right here in Moment might make more sense {...}

It didn't occur to me in my previous comment, but I guess this would be possible since there isn't any ordering of operations inside a Moment. I'll go ahead with this change.

cirq/google/api/v2/program.proto
95-martin-orion: To me this is implied by the outer object; these pairs will always appear in the context of a mapping (e.g. `QubitMapping.QubitPair`), so additional clarification feels redundant.

cirq/google/api/v2/program.proto
95-martin-orion: Done.

cirq/google/api/v2/program.proto
maffoo: In that case, is the `Qubit` redundant? That is, maybe `QubitMapping.Entry` or `QubitMapping.Item` (as in python `dict.items()`) would be better.

cirq/google/api/v2/program.proto
```diff
@@ -198,3 +204,90 @@ message ArgFunction {
...
+}
+
+// A mapping of qubits from one value to another. All mappings are applied
+// simultaneously and independently; for example, [(a, b), (b, a)] will swap
+// qubits a and b.
+message QubitMapping {
+  // Indicates that qubit "source" should be replaced with "target".
+  message QubitPair {
+    Qubit source = 1;
+    Qubit target = 2;
```
maffoo, line 259: Sorry to nitpick, but I still feel like "source" and "target" is ambiguous. Is this saying that the qubit id appearing in the _outer_ circuit is the source and should be replaced by the target value in the inner circuit, or the other way around? I'd really prefer to use something like "inner" and "outer" because the _containment_ relationship is unambiguous (one circuit contains the other as an operation, not the other way around).

cirq/google/api/v2/program.proto
95-martin-orion: Opted for `Entry` to avoid confusion over whether these are `map`-type proto objects in Python code.

cirq/google/api/v2/program.proto
95-martin-orion: Discussed offline: `inner` and `outer` only make sense in the context of `CircuitOperation`, where we have comments documenting the behavior of these maps.

Switched these to `key`/`value` to more closely align with the non-proto `CircuitOperation` object.

cirq/google/api/v2/program.proto
```diff
@@ -215,8 +215,8 @@ message CircuitOperation {
   // The index of the circuit in the top-level constant table.
   int32 circuit_constant_index = 1;
 
-  // Specifier for repetitions of the circuit, including number of reps and
-  // repetition IDs.
+  // Specifier for repetitions of the circuit, which contains either a number
+  // of repetitions or a list of repetition IDs.
   RepetitionSpecification repetition_specification = 2;
 
   // Map of qubits in the circuit to the qubits used in this operation.
```
maffoo, line 222: Perhaps could be a bit more clear here, something like "Map from qubits in the inner circuit (referenced by circuit_constant_index) to the outer circuit which contains this operation." I'd suggest similar wording for `measurement_key_map` and `arg_map` below.

maffoo (approved):
Minor command about docs, then LGTM.

cirq/google/api/v2/program.proto
95-martin-orion: Done.

95-martin-orion:
Holding out for approval from @dstrain115 prior to merging.

cirq/google/api/v2/program.proto
```diff
@@ -198,3 +204,97 @@ message ArgFunction {
...
+// mappings of that circuit are different.
+message CircuitOperation {
+  // The index of the circuit in the top-level constant table.
+  int32 circuit_constant_index = 1;
+
+  // Specifier for repetitions of the circuit, which contains either a number
+  // of repetitions or a list of repetition IDs.
+  RepetitionSpecification repetition_specification = 2;
+
+  // Map from qubits in the "inner" circuit (referenced by
```
dstrain115, line 222: I am not sure I understand why it is called inner and outer circuit, but which one is which is clarified in the comment, so I guess it is fine.

dstrain115 (approved):

Issue #3884, opened 2021-03-06
GlobalPhaseOperation within CircuitOperation is not applied correctly to the whole circuit


balopat:
**Is your design idea/issue related to a use case or problem? Please describe.**

With @ybc1991  we are working on the "decomposable" N-qubit diagonal gate #3866. Here, it would be nice to be able to add the global phase of the operation stemming from the first angle. `cirq.ThreeQubitDiagonalGate` gets around this by adding in the global phase to the rotation gates as a shift. If we use `cirq.GlobalPhaseOperation` the problem is that `cirq.ControlledGate(cirq.DiagonalGate))` won't work out of the box - because `cirq.ControlledOperation(cirq.GlobalPhaseOperation)` does not behave well.

**Describe your design idea/issue**

The main issue is that `GlobalPhaseOperation` is qubitless. This is technically correct...a global phase `exp(2 pi i theta)` has ambiguity in its representation: it could be represented with a scalar (current solution), or a diagonal gate with `exp(2 pi i theta)` in all diagonal elements spanning any subset of the qubits. 

The design issue here is that Operations in Cirq meant to be reused in local contexts - that's when issues arise: 

1. `cirq.ControlledOperation` should work with most operations - barring that they have a known decomposition. However, an operation that does not have qubits cannot be made local. Here are some fun examples:

```
In [16]: cirq.unitary(cirq.GlobalPhaseOperation(-1).controlled_by(cirq.LineQubit(1)))
Out[16]: 
array([[ 1,  0],
       [ 0, -1]])
```
2. `cirq.CircuitOperation` seems to ignore it? It should propagate it properly to the top level, right?

```
In [21]: cirq.Circuit(cirq.X(cirq.LineQubit(2)), cirq.CircuitOperation(cirq.Circuit(cirq.GlobalPhaseOperation(-1)).freeze()))
Out[21]: 2: ───X───
```

smitsanghavi:
example 2 looks like a bug, the global phase should be propagated from inside a subcircuit. I don't see much of an issue with example 1. When the controlling qubit is `|1>`, the controlled operation is being applied. What is the expected output?
`
array([[ -1,  0],
       [ 0, -1]])
`? Should the phase stay global even when controlled?

balopat:
1.
[...]

PR #3892, opened 2021-03-09
Propagate global phase in diagram from sub circuits

smitsanghavi:
#3884

### DIFF
--- a/cirq/circuits/circuit.py
+++ b/cirq/circuits/circuit.py
@@ -2298,10 +2298,10 @@ def _draw_moment_in_diagram(
 
 def _get_global_phase_and_tags_for_op(op):
     if isinstance(op.untagged, ops.GlobalPhaseOperation):
-        return op.untagged.coefficient, list(op.tags)
+        return op.untagged.coefficient, op.tags
     elif isinstance(op.untagged, CircuitOperation):
         op_phase, op_tags = _get_global_phase_and_tags_for_ops(op.untagged.circuit.all_operations())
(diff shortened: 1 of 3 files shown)


cirq/circuits/circuit.py
```diff
@@ -2302,6 +2296,30 @@ def _draw_moment_in_diagram(
         moment_groups.append((x0, max_x))
 
 
+def _get_global_phase_and_tags_for_op(op):
```
95-martin-orion, line 2299: Please provide type annotations for these functions.

95-martin-orion (approved):

cirq/circuits/circuit_operation_test.py
```diff
@@ -307,6 +307,17 @@ def test_string_format():
 [                         ]"""
     )
 
+    fc0_global_phase = cirq.FrozenCircuit(cirq.GlobalPhaseOperation(-1))
```
balopat, line 310: can you please add a test for the multiplication of `GlobalPhaseOperation`s? Also for nested cases.

cirq/circuits/circuit.py
smitsanghavi: Done

cirq/circuits/circuit_operation_test.py
smitsanghavi: Expanded the test to include nested as well as unnested phase multiplication

95-martin-orion (approved):

Issue #3678, opened 2021-01-14
Remove implicit assumption that Operations have Gates


95-martin-orion:
**Is your feature request related to a use case or problem? Please describe.**

The `Operation` has a `gate` property which is frequently used to identify its type. However, not all `Operation`s are `GateOperation`s, meaning that this field is not always populated.

Until recently this has been a relatively minor problem, since the vast majority of `Operation`s are `GateOperation`s. However, with the introduction of `CircuitOperation` (#3580) this issue is becoming more pronounced.

Related issue: #3235.

**Describe the solution you'd like**

Remove `gate` field in `Operation` and replace with a `get_operator` abstract method, defined in implementations. This will require deprecating the field and replacing it across the Cirq repo.

**[optional] Describe alternatives/workarounds you've considered**

- Add local fixes to sections of code that need to recognize CircuitOperation, and leave the rest unchanged.
    - Issue: eventually, CircuitOperation should be supported everywhere.
- Make Circuit a subclass of Gate, with CircuitOperations containing CircuitGates
    - Rejected during review (#3509): treating circuits as gates introduces undesirable behavior

**What is the urgency from your perspective for this issue? Is it blocking important work?**

P1 - I need this no later than the next release (end of quarter)

This confounds work on CircuitOperations, as they do not have a value in their `gate` field. Continuing work in that space without resolving this issue requires workarounds which will increase the cost of an eventual fix.

maffoo:
We don't assume that operations have gates; that's why the `gate` field has type `Optional[cirq.Gate]` as opposed to `cirq.Gate`. Why not just have return `None` for the `gate` property on `CircuitOperation`? I'll admit I'm biased here because I was the one who initially added the `gate` field (we explicitly made it optional to accomodate all operations). This change got rid of helper functions that we used previously to identify `GateOperations` and extract their gates and check the gate type, etc. and simplified a lot of code so I'd prefer to keep it.

95-martin-orion:
> Why not just have return `None` for the `gate` property on `CircuitOperation`?

This is the current behavior for `CircuitOperation`. If we keep this behavior, every line of code that relies on `op.gate` to recognize the type of an operation must separately check for `CircuitOperation` - which itself is awkward, because simply checking `isinstance(op, CircuitOperation)` fails to identify `CircuitOperation(...).with_tags(...)`, which is a `TaggedOperation`.

[...]

PR #3893, opened 2021-03-09
Replace _get_op_circuit with op.untagged

95-martin-orion:
Since `Operation` provides a canonical way to get the untagged version of itself, we should use it instead of writing a new method to do the same thing.

Using `op.untagged` in #3634 removes any residual blocking effects from #3678, although I'm still in favor of finding a proper resolution to #3678.

### DIFF
--- a/cirq/circuits/circuit.py
+++ b/cirq/circuits/circuit.py
@@ -773,10 +773,11 @@ class AbstractCircuit(abc.ABC):
             given predicate are terminal. Also checks within any CircuitGates
             the circuit may contain.
         """
+        from cirq.circuits import CircuitOperation
         if not all(
             self.next_moment_operating_on(op.qubits, i + 1) is None
             for (i, op) in self.findall_operations(predicate)
-            if getattr(op.untagged, 'circuit', None) is None
+            if not isinstance(op.untagged, CircuitOperation)
         ):
             return False
 
@@ -813,10 +814,11 @@ class AbstractCircuit(abc.ABC):
             given predicate are terminal. Also checks within any CircuitGates
             the circuit may contain.
         """
+        from cirq.circuits import CircuitOperation
         if any(
             self.next_moment_operating_on(op.qubits, i + 1) is None
             for (i, op) in self.findall_operations(predicate)
-            if getattr(op.untagged, 'circuit', None) is None
+            if not isinstance(op.untagged, CircuitOperation)
         ):
             return True
 
--- a/cirq/circuits/circuit.py
+++ b/cirq/circuits/circuit.py
@@ -774,6 +774,7 @@ class AbstractCircuit(abc.ABC):
             the circuit may contain.
         """
         from cirq.circuits import CircuitOperation
+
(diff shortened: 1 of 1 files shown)


cirq/circuits/circuit.py
```diff
@@ -2201,15 +2201,6 @@ def with_noise(self, noise: 'cirq.NOISE_MODEL_LIKE') -> 'cirq.Circuit':
         return c_noisy
 
 
-def _get_op_circuit(op: ops.Operation) -> Optional['cirq.FrozenCircuit']:
-    """Retrieves the circuit contained by an operation, if there is one."""
-    from cirq.circuits import CircuitOperation
-
-    while isinstance(op, ops.TaggedOperation):
-        op = op.sub_operation
```
balopat, line 2209: by reverting to op.untagged - don't we miss out in this recursion? `op.untagged` only goes down to a single layer of `suboperation` as far as I can tell.

95-martin-orion: You're right, but I think this represents an issue in `TaggedOperation.untagged`: it claims to give a tag-free version of the operation, but might not if a nested `TaggedOperation` was created. (`TaggedOperation` has safeguards against nesting, but they're far from perfect.)

If you'd like, I can update `TaggedOperation.untagged` in this PR as well.

95-martin-orion: I don't think there's a real use case for nested `TaggedOperations` - we explicitly guard against that behavior (though, as noted above, the protection is imperfect). My vote is in favor of either (1) only supported fully-recursive `untagged`, or (2) fully guarding against nesting (not sure if this is possible in Python).

maffoo: I don't think there should ever be a need to create nested `TaggedOperation`. If you use `op.with_tags` it will not create nested ops, and to me that seems like it should be sufficient. If we wanted to provide more assurance against people doing this accidentally we could modify `TaggedOperation.__init__` to flatten nested structures if you pass in a sub operation that is itself a `TaggedOperation`.

cirq/circuits/circuit.py
```diff
@@ -776,13 +776,13 @@ def are_all_matches_terminal(self, predicate: Callable[['cirq.Operation'], bool]
         if not all(
             self.next_moment_operating_on(op.qubits, i + 1) is None
             for (i, op) in self.findall_operations(predicate)
-            if _get_op_circuit(op) is None
+            if getattr(op.untagged, 'circuit', None) is None
```
maffoo, line 779: I think it would more clearly convey intent here to check the type instead
```suggestion
            if not isinstance(op.untagged, CircuitOperation)
```
Would have to import `CircuitOperation` in `are_all_matches_terminal` and `are_any_matches_terminal`, but that should be fine. WDYT?

cirq/circuits/circuit.py
balopat: oh nice, thanks for expanding on this - it makes sense!

PR #3899, opened 2021-03-10
decompose_preserving_structure

95-martin-orion:
Part of #3634.

In order to send `CircuitOperation`s to QCS, we must be able to decompose the contents of those `CircuitOperation`s without decomposing the `CircuitOperation` itself. `decompose_preserving_structure` provides this behavior.


cirq/protocols/decompose_protocol.py
```diff
@@ -387,3 +382,66 @@ def _try_decompose_into_operations_and_qubits(
         return result, qubits, tuple(qid_shape_dict[q] for q in qubits)
 
     return None, (), ()
+
+
+def decompose_preserving_structure(
```
balopat, line 387: Do we really need a completely new protocol for this? Given that decompose is challenging to understand/reason about already, I feel like introducing another version makes it even more challenging...we'll also need to keep the API exactly the same which is another smell...wouldn't another parameter to decompose() work?

95-martin-orion: I'm not strictly opposed to folding this into `decompose` directly. Something like this?
```
def decompose(..., preserve_structure=False):
    if preserve_structure:
        if intercepting_decomposer is not None:
            return ValueError(...)
        return _decompose_preserving_structure(...)
    {...}
```

95-martin-orion: Change applied.

95-martin-orion:
Per @balopat's comment `_decompose_preserving_structure` is now private and can be invoked with
```
decompose(..., preserve_structure=True)
```

cirq/protocols/decompose_protocol.py
```diff
@@ -182,6 +181,9 @@ def decompose(
             returns `None`, non-decomposable operations are simply silently
             kept. `on_stuck_raise` defaults to a `ValueError` describing the
             unwanted non-decomposable operation.
+        preserve_structure: Prevents structural elements (i.e. subcircuits)
```
balopat, line 184:
```suggestion
        preserve_structure: Prevents subcircuits (i.e. `CircuitOperation`s)
```

balopat (approved):
LGTM.

cirq/protocols/decompose_protocol.py
95-martin-orion: Applied change. My intent is for this to preserve other purely-structural circuit features as they emerge, but since that's only `CircuitOperation` for now I'm happy to limit the scope of the comment.

Issue #3901, opened 2021-03-10
Flynt workflow concerns


mpharrigan:
**Description of the issue**
following https://github.com/quantumlib/Cirq/pull/3786

`check/format-incremental` now takes ~8 seconds instead of ~1 second

It also puts a little promotional message each time

> _-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_._-_.
Please run your tests before committing. Did flynt get a perfect conversion? give it a star at: 
~ https://github.com/ikamensh/flynt ~
Thank you for using flynt. Upgrade more projects and recommend it to your colleagues!

I suspect many had gotten in the habit of running format-incremental regularly during development, so this slows things down.

**Cirq version**
0.10.0

mpharrigan:
I think it definitely should only change changed files if it's being called from format-incremental

PR #3903, opened 2021-03-10
Move flynt from `format-incremental` to `lint`

vtomole:
- Renames `pylint` and `pylint-changed-files` scripts to `lint` and `lint-changed-files` respectively.
Fixes: https://github.com/quantumlib/Cirq/issues/3901


balopat:
This is a bit simpler - format-incremental is already incremental: https://github.com/quantumlib/Cirq/pull/3905

balopat:
@vtomole flynt is better in the format-incremental because it can benefit from the `--apply` flag.

balopat:
Sorry @vtomole I didn't see your PRs when I pushed my solution, but in the end I prefer that approach because it's simpler in my opinion. It also handles the `--apply` and `--dry-run` behaviors.

vtomole:
@balopat 👍

PR #3906, opened 2021-03-11
Add seconds to engine program id generation

dstrain115:
- Clients were occasionally getting duplicate program ids
when running successive results.
- Root cause is unknown, but likely cause is that a sub-module
was calling random.seed() somewhere.
- This changes program id generation to include seconds in the id,
so that successive calls will generate distinct ids even if the
random number generator is seeded.


cirq/google/engine/engine.py
```diff
@@ -65,7 +65,7 @@ class ProtoVersion(enum.Enum):
 def _make_random_id(prefix: str, length: int = 16):
     random_digits = [random.choice(string.ascii_uppercase + string.digits) for _ in range(length)]
     suffix = ''.join(random_digits)
-    suffix += datetime.date.today().strftime('%y%m%d')
+    suffix += datetime.date.today().strftime('%y%m%d-%H%M%S')
```
balopat, line 68: why not use UUID?

balopat: The problem with this second based approach that it assumes that if the faulty (assumingly seeding) client program is generating more than one program per second, you'll still get collisions. With UUID, neither seeding nor speed won't be a problem - though it will be longer (36 characters).

balopat (approved):
LGTM for experimenting but I think it might still not really fix the issue for clients that generate a lot of programs fast.

mrwojtek (approved):
I'm also fine with this as a temporary fix. I like the UUID approach, this is marginal compared to the rest of request payload.

cirq/google/engine/engine.py
dstrain115: A UUID based on what?  I am not sure if that solves the problem. 
Would you hash based on the time + circuit?  If we use a random uuid, then that would be based on random() and also fail if seeded.

dstrain115:
> LGTM for experimenting but I think it might still not really fix the issue for clients that generate a lot of programs fast.

Possibly, but you would have to generate more than one program in a second _and_ also set random.seed() accidentally before generating each program.

balopat:
Let's experiment then!

cirq/google/engine/engine.py
balopat: +1

PR #3908, opened 2021-03-12
Fix naming of sqrt(iswap) gate constants

mrwojtek:
The constant named SQRT_ISWAP_PARAMETERS is really an inverse of it and this PR fixes that.

Unfortunately this was a part of a public interface but it's unlikely that it is used externally, it's not a part of any standard application so far. I'm not sure how easy it is to go through the deprecation procedure for this.


95-martin-orion:
The correct way to deprecate this is to use `cirq._compat.deprecate_attributes`:
https://github.com/quantumlib/Cirq/blob/bfb7aa316e655f928b0aeeec96ad2ac68fe7d2b1/cirq/_compat.py#L255
There is a usage example in `cirq/sim/__init__.py`:
https://github.com/quantumlib/Cirq/blob/bfb7aa316e655f928b0aeeec96ad2ac68fe7d2b1/cirq/sim/__init__.py#L89

mrwojtek:
Thanks, I applied that.

mpharrigan:
This will be helpful in alleviating confusion

cirq/google/calibration/__init__.py
```diff
@@ -43,3 +44,14 @@
...
 )
+
+# pylint: disable=wrong-import-order
+from typing import Dict, Tuple
+
+import sys as _sys
+from cirq._compat import deprecate_attributes as _deprecate_attributes
+
+deprecated_constants: Dict[str, Tuple[str, str]] = {
+    'SQRT_ISWAP_PARAMETERS': ('v0.11', 'Use cirq.google.SQRT_ISWAP_INV_PARAMETERS instead'),
```
95-martin-orion, line 55: `v0.11` should be `v0.12` - all deprecations must go through one release in the `deprecated` state before being fully removed.

Same in the file comment.

cirq/google/calibration/phased_fsim.py
```diff
@@ -198,9 +198,11 @@ def override_by(self, other: 'PhasedFSimCharacterization') -> 'PhasedFSimCharact
         return other.merge_with(self)
 
 
-SQRT_ISWAP_PARAMETERS = PhasedFSimCharacterization(
+SQRT_ISWAP_INV_PARAMETERS = PhasedFSimCharacterization(
     theta=np.pi / 4, zeta=0.0, chi=0.0, gamma=0.0, phi=0.0
 )
+# Pending deprecation in 0.11.
+SQRT_ISWAP_PARAMETERS = SQRT_ISWAP_INV_PARAMETERS
```
95-martin-orion, line 205: Please add this test to `phased_fsim_test` to check that the deprecated variable still works:
```
def test_sqrt_iswap_parameters_deprecation():
    with pytest.warns(DeprecationWarning):
        assert SQRT_ISWAP_PARAMETERS == SQRT_ISWAP_INV_PARAMETERS
```

95-martin-orion (changes requested):
Couple of nits.

cirq/google/calibration/__init__.py
mrwojtek: Done

cirq/google/calibration/phased_fsim.py
mrwojtek: I added the tests but it's failing because it complains I'm using a deprecated parameter in Cirq. Do I do something wrong here?

95-martin-orion: I think the issue is due to importing the variable by name outside of the `pytest.warns` block. Our CI tests treat any use of deprecated fields as failures unless they're captured by one of these blocks.

95-martin-orion: Sorry, I misunderstood our deprecation-testing behavior. In place of the current test, you'll want to use `assert_deprecated` instead:
https://github.com/quantumlib/Cirq/blob/69daec18bb2d2cdd7ca8ffc6158b2f1658c27dd5/cirq/testing/deprecation.py#L23

mrwojtek: I think that I fixed whatever I could easily fix. Once I modified `cirq.google.__init_-`, the following error started to appear: `ValueError: cirq.google.__spec__ is None`. Would you know where it comes from?

95-martin-orion: I'm not familiar with this error, but it looks like it's related to the JSON modularization changes.

@balopat, could you take a look at this?

mrwojtek:
I'm thinking to drop the support for deprecation. There are multiple issues with it and this change will affect at most a few people, if any at all.

mrwojtek:
This has been merged without deprecation in #4051

PR #3909, opened 2021-03-12
cleanup _assert_equivalent_op_tree

balopat:
Adds cirq.testing.assert_equivalent_op_tree.

This method was called from three other testing packages as a private method, this change moves it into its proper place: cirq.testing.


95-martin-orion (approved):
Approved in lieu of lint fixes.

PR #3910, opened 2021-03-12
Add Sampler API for expectation values

95-martin-orion:
Stage 2 of the Expectation Value API (#3492). This is a rewrite of #3700.

This PR introduces the sample_expectation_values method and its default implementation, which injects gates to ensure measurement occurs in the correct basis. Hardware and software samplers can use this method as-is or implement their own version for custom behavior (e.g. server-side calculation of expectation values).

### DIFF
--- a/cirq/work/sampler.py
+++ b/cirq/work/sampler.py
@@ -13,9 +13,10 @@
 # limitations under the License.
 """Abstract base class for things sampling quantum circuits."""
 
-from typing import List, Optional, TYPE_CHECKING, Union
+from typing import Dict, List, Optional, TYPE_CHECKING, Union
 import abc
 
+import numpy as np
 import pandas as pd
 
 from cirq import ops, study
@@ -24,9 +25,45 @@ from cirq.work.observable_measurement import (
     _get_params_for_setting,
     _with_parameterized_layers,
 )
+from cirq.work.observable_measurement_data import (
+    _stats_from_measurements,
+)
 
 if TYPE_CHECKING:
     import cirq
+
+
+def get_samples_mask_for_parameterization(
+    samples: pd.DataFrame,
+    param_resolver: 'cirq.ParamResolverOrSimilarType',
+    mask: Optional[pd.Series] = None,
+) -> pd.Series:
+    """Generates a 'mask' for a given parameterization of a sample.
+
+    Results for the given parameterization can be extracted with:
+
+        samples.loc[series, measure_key]
+
+    where 'series' is the output of this function and 'measure_key' is the
+    measurement key to extract results for.
+
+    Args:
+        samples: a pandas DataFrame containing sampling results and their
+            associated parameterizations.
+        param_resolver: the specific parameterization of 'samples' to create a
+            mask for.
+        mask: a base mask to generate from. Any indices excluded by this mask
+            will be excluded by the final mask. If this is left unspecified,
+            it defaults to a blank mask (include-all).
+            Using a previous result of this method as the 'mask' for
+            subsequent calls allows construction of a complete mask from
+            multiple partial parameterizations of the sampling results.
+    """
+    series = mask if mask is not None else np.ones(len(samples), dtype=bool)
+    pr = study.ParamResolver(param_resolver)
+    for k in pr:
+        series = series & (samples[k] == pr.value_of(k))
+    return series
 
 
 class Sampler(metaclass=abc.ABCMeta):
@@ -149,6 +186,17 @@ class Sampler(metaclass=abc.ABCMeta):
     ) -> List[List[float]]:
         """Calculates estimated expectation values from samples of a circuit.
 
+        This is a minimal implementation for measuring observables, and is best
+        reserved for simple use cases. For more complex use cases, consider
+        upgrading to `cirq/work/observable_measurement.py`. Additional features
+        provided by that toolkit include:
+            - Chunking of submissions to support more than (max_shots) from
+              Quantum Engine
+            - Checkpointing so you don't lose your work halfway through a job
+            - Measuring to a variance tolerance rather than a pre-specified
+              number of repetitions
+            - Readout error symmetrization and mitigation
+
         This method can be run on any device or simulator that supports circuit
         sampling. Compare with `simulate_expectation_values` in simulator.py,
         which is limited to simulators but provides exact results.
@@ -194,7 +242,10 @@ class Sampler(metaclass=abc.ABCMeta):
 
         obs_pstrings: List[ops.PauliString] = []
         for psum in psums:
-            pstring_aggregate = {}
+            # This dict will be the PAULI_STRING_LIKE for a new PauliString
+            # representing this PauliSum. The PauliString is necessary for
+            # compatibility with `observables_to_settings`.
+            pstring_aggregate: Dict['cirq.Qid', 'cirq.PAULI_GATE_LIKE'] = {}
             for pstring in psum:
                 pstring_aggregate.update(pstring.items())
             obs_pstrings.append(ops.PauliString(pstring_aggregate))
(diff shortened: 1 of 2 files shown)


cirq/work/sampler.py
```diff
@@ -133,6 +138,126 @@ def sample(
...
+        num_qubits = len(qubits)
+        psums = (
+            [ops.PauliSum.wrap(o) for o in observables]
+            if isinstance(observables, List)
+            else [ops.PauliSum.wrap(observables)]
+        )
+
+        obs_pstrings: List[ops.PauliString] = []
+        for psum in psums:
+            pstring_aggregate = {}
```
mpharrigan, line 197: type annotation could be helpful here

cirq/work/sampler.py
```diff
@@ -133,6 +138,126 @@ def sample(
...
+            for pstring in psum:
+                pstring_aggregate.update(pstring.items())
+            obs_pstrings.append(ops.PauliString(pstring_aggregate))
```
mpharrigan, line 200: what is this section doing? Factor out into a private helper function?

cirq/work/sampler.py
```diff
@@ -133,6 +138,126 @@ def sample(
...
+
+        # Results sampled from the modified circuit. Parameterization ensures
+        # that all 'z' results map directly to observables.
+        samples = self.sample(mod_program, repetitions=num_samples, params=all_sweeps)
+        results: List[List[float]] = [[0] * len(psums) for _ in range(num_input_param_values)]
+
+        for max_setting, grouped_settings in sampling_groups.items():
+            ev_params = sampling_params[max_setting]
+            # A set of num_samples integer-packed results for this setting.
+            series = [True] * len(samples)
```
mpharrigan, line 234: use numpy? np.ones(len(samples), dtype=bool)

cirq/work/sampler.py
```diff
@@ -133,6 +138,126 @@ def sample(
...
+                        sum_val = 0
+                        for int_result in int_results:
+                            val = 1
+                            for q in pstr.qubits:
+                                bit_idx = num_qubits - qmap[q] - 1
+                                if (int_result >> bit_idx) & 1:
+                                    val *= -1
+                            sum_val += val
+                        ev += sum_val * pstr.coefficient
+                    results[sweep_idx][obs_idx] = ev / num_samples
```
mpharrigan, line 257: this section is very difficult to read and I suspect much slower than a vectorized numpy solution would be. Maybe factoring parts into helpfully-named helper functions or comments could go a long way?

Since you're already using some of the `cirq.work` functionality, consider using `observable_measurement_data._stats_from_measurements`

mpharrigan (changes requested):
(I would have marked this "comment" but then the "see changes since last reviewed doesn't work, so you get the angry red X)

It surprised me to see this implemented this way. This is halfway towards the sophisticated `cirq.work` set of routines. I would have expected this to serve as a simple, reference implementation. Specifically, I was surprised to see grouping and parameterized circuits showing up here, although I'm delighted that some of the functionality in `cirq.work` was useful. In that vein, we should converge on how to turn the results from a sampler to expectation values. I'm not so sure about the current bit twiddling in this PR.

The reason why I would have expected a very simple "reference implementation" is that if you wanted any degree of sophistication, it would probably make sense to just wrap the high-level `cirq.work` entry functions. That is, this function body would be like two lines. Of course that high-level API isn't merged yet. You can preview it e.g. https://github.com/quantumlib/Cirq/pull/3647/files#diff-da3dfc139c4c6d4a898a2ad1b919f0e3c72880da2c2606288077bb15ae80b630R562 here. (Line 562). The upside is the signature (given the rest of the default arguments) matches the API you've proposed here.

cirq/work/sampler.py
95-martin-orion: The purpose of this is to get `InitObsSettings` for each `PauliSum`. Since `observables_to_settings` only accepts `PauliString`s, I perform this conversion for each `PauliSum` provided.

If it makes sense to you, I could fold this into `observables_to_settings` and have it accept a mix of `PauliString` and `PauliSum` as observables.

cirq/work/sampler.py
95-martin-orion: Done.

cirq/work/sampler.py
95-martin-orion: Done.

cirq/work/sampler.py
95-martin-orion: Done. Also split off a helper function, but with `_stats_from_measurements` there's not much left to split off.

95-martin-orion:
In addition to the changes noted above, I added a plug in the method description for the full observable measurements toolkit, including the selling points listed in #2781.

mpharrigan:
For Simulators that implement the Sampler interface, have you considered just using the analytic expectation values and drawing samples from a binomial distribution?

95-martin-orion:
> For Simulators that implement the Sampler interface, have you considered just using the analytic expectation values and drawing samples from a binomial distribution?

I suppose we could do this? It seems orthogonal to this PR, though - those simulators have access to this implementation for now (since `SimulatesSamples` inherits from `Sampler`) and can override it if a faster method is required.

As an aside: what is the use case for this? IIUC, such a simulator could implement `simulate_expectation_values` and return the exact EVs instead of "sampling" inexact EVs this way.

cirq/work/sampler.py
```diff
@@ -13,17 +13,59 @@
...
 
+import numpy as np
 import pandas as pd
 
-from cirq import study
+from cirq import ops, study
+from cirq.work import group_settings_greedy, observables_to_settings
+from cirq.work.observable_measurement import (
+    _get_params_for_setting,
+    _with_parameterized_layers,
```
dstrain115, line 26: This is bad practice to import private functions from other modules.  It breaks naming assumptions.  We should either change the name to indicate that it is no longer private or remove the import (or create a new method that is public in those modules).

cirq/work/sampler.py
```diff
@@ -13,17 +13,59 @@
...
 if TYPE_CHECKING:
     import cirq
 
 
+def get_samples_mask_for_parameterization(
+    samples: pd.DataFrame,
+    param_resolver: 'cirq.ParamResolverOrSimilarType',
+    mask: Optional[pd.Series] = None,
+) -> pd.Series:
+    """Generates a 'mask' for a given parameterization of a sample.
```
dstrain115, line 41: What is a mask for a parameterization?  I think this should be explained.

cirq/work/sampler.py
```diff
@@ -133,6 +175,133 @@ def sample(
...
+        *,
+        num_samples: int,
+        params: 'cirq.Sweepable' = None,
+        permit_terminal_measurements: bool = False,
+    ) -> List[List[float]]:
+        """Calculates estimated expectation values from samples of a circuit.
+
+        This is a minimal implementation for measuring observables, and is best
+        reserved for simple use cases. For more complex use cases, consider
+        upgrading to `cirq/work/observable_measurement.py`. Additional features
```
dstrain115, line 191: Change to the user visible name instead of the source code name, such as cirq.work.measure_grouped_settings

cirq/work/sampler.py
```diff
@@ -133,6 +175,133 @@ def sample(
...
+        provided by that toolkit include:
+            - Chunking of submissions to support more than (max_shots) from
+              Quantum Engine
+            - Checkpointing so you don't lose your work halfway through a job
+            - Measuring to a variance tolerance rather than a pre-specified
+              number of repetitions
+            - Readout error symmetrization and mitigation
+
+        This method can be run on any device or simulator that supports circuit
```
dstrain115, line 200: Consider reordering or clarifying this.   It's unclear which "this method" you are referring to (observable measurement or sample_expectation_values).  Maybe consider putting the pointing to observable measurement at the end.  I think that makes more sense to explain what _this_ function does before pointing out alternatives.

cirq/work/sampler.py
```diff
@@ -133,6 +175,133 @@ def sample(
...
+        if not permit_terminal_measurements and program.are_any_measurements_terminal():
+            raise ValueError(
+                'Provided circuit has terminal measurements, which may '
+                'skew expectation values. If this is intentional, set '
+                'permit_terminal_measurements=True.'
+            )
+        qubits = ops.QubitOrder.DEFAULT.order_for(program.all_qubits())
+        qmap = {q: i for i, q in enumerate(qubits)}
+        num_qubits = len(qubits)
+        psums = (
```
dstrain115, line 237: Optional: consider avoiding abbreviations and use pauli_sums to be more clear.  Optional since this does make the pauli_string_aggregate variable name kind of long.

cirq/work/sampler.py
95-martin-orion: I agree, although this behavior is consistent with the observable measurements imports (all files are under `cirq/work`):
https://github.com/quantumlib/Cirq/blob/845836ac5fa3cf2bba0258c2e01169a28eb1a2fd/cirq/work/observable_measurement_data.py#L23-L29

Should I send a "prequel" PR that fixes this before moving forward with this one?

cirq/work/sampler.py
95-martin-orion: Clarified which method this refers to. I prefer to keep this paragraph at the top of the docstring, since all but the simplest use cases can benefit from using the Observable Measurement toolkit instead.

To that end, there's an argument to be made for dropping this API altogether: it doesn't provide functionality outside what will be possible with the Observable Measurement toolkit, and potentially distracts users from the "correct" tools for getting expectation values. As it stands, it primarily serves as a simple reference implementation, as pointed out above.

95-martin-orion:
Discussed offline: putting this on hold in lieu of full Observable Measurement toolkit. To avoid distracting users from that toolkit, we may end up dropping this PR.

balopat:
ping @95-martin-orion - if the decision is to go with the OM toolkit, maybe you could team up with Matt to clean up those PRs? :)

95-martin-orion:
> if the decision is to go with the OM toolkit, maybe you could team up with Matt to clean up those PRs? :)

Can confirm that this was the decision we reached. Reading #3647 it looks like the next PR is still in development - @mpharrigan feel free to assign me to review on that when it comes out.

95-martin-orion:
Closing this PR as discontinued.

PR #3914, opened 2021-03-14
Make assert_logs levels interval based

balopat:
Renames cirq.testing.assert_logs parameter `level` -> `min_level` and adds a new parameter `max_level`. 
This is to enable more sophisticated logs assertions based on multiple levels.

===== Release note ======

- `cirq.testing.assert_logs` parameter `level` is now deprecated and will be removed in Cirq v0.12. Please use instead `min_level`.

======================

### DIFF
--- a/cirq/testing/logs.py
+++ b/cirq/testing/logs.py
@@ -16,7 +16,19 @@
 import logging
 from typing import ContextManager, List, Optional
 
+from cirq._compat import deprecated_parameter
 
+
+@deprecated_parameter(
+    deadline="v0.12",
+    fix="use min_level instead",
+    parameter_desc="level",
+    match=lambda args, kwargs: 'level' in kwargs,
+    rewrite=lambda args, kwargs: (
+        args,
+        {('min_level' if k == 'level' else k): v for k, v in kwargs.items()},
+    ),
+)
 def assert_logs(
     *matches: str,
     count: Optional[int] = 1,
@@ -58,6 +70,8 @@ def assert_logs(
         for code executed within the entered context. This ContextManager
         checks that the asserts for the logs are true on exit.
     """
+    if min_level > max_level:
+        raise ValueError("min_level should be less than or equal to max_level")
     records = []
 
     class Handler(logging.Handler):
--- a/cirq/testing/logs.py
+++ b/cirq/testing/logs.py
@@ -20,7 +20,8 @@ from typing import ContextManager, List, Optional
 def assert_logs(
     *matches: str,
     count: Optional[int] = 1,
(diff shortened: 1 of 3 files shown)


cirq/testing/logs_test.py
```diff
@@ -118,11 +118,18 @@ def test_assert_logs_log_level():
         logging.error('orange apple fruit')
         logging.debug('should not')
         logging.info('count')
-    with cirq.testing.assert_logs('apple', level=logging.INFO, count=2):
+    with cirq.testing.assert_logs('apple', min_level=logging.INFO, count=2):
         logging.error('orange apple fruit')
         logging.debug('should not')
         logging.info('count')
 
+    with cirq.testing.assert_logs('info only 1', min_level=logging.INFO, max_level=logging.INFO):
```
95-martin-orion, line 126: As written, this test will pass even if `max_level` is ignored: `info only 1` appears in a log of level `INFO` or greater, and `count` is unspecified and therefore not checked.

cirq/testing/logs.py
```diff
@@ -20,7 +20,8 @@
 def assert_logs(
     *matches: str,
     count: Optional[int] = 1,
-    level: int = logging.WARNING,
+    min_level: int = logging.WARNING,
```
95-martin-orion, line 23: Does `level` need to go through deprecation?

cirq/testing/logs.py
```diff
@@ -20,7 +20,8 @@
...
+    max_level: int = logging.CRITICAL,
```
95-martin-orion, line 24: This should raise an error if `max_level < min_level`, as otherwise we'll silently pass assertions like this:
```
with assert_logs(max_level=logging.INFO, count=0):
    # will pass regardless of logs generated here
```

95-martin-orion (changes requested):

cirq/testing/logs.py
balopat: I was on a fence about this. I guess especially for vendor modules treating `cirq.testing` as part of the library is nice.

cirq/testing/logs_test.py
balopat:
```
>       assert count is None or len(records) == count, (
            f'Expected {count} log message but ' f'got {len(records)}. Log messages: ' f'{msgs}'
        )
E       AssertionError: Expected 1 log message but got 2. Log messages: ['info only 1', 'info warning 1']
```

cirq/testing/logs.py
balopat: Done.

95-martin-orion (approved):

PR #3918, opened 2021-03-15
Pin pandas version to >= 1.1.4

tanujkhattar:
Cirq is broken for pandas version < 1.1.4


balopat (approved):
LGTM - the g3 dep is 1.1.5, so we're good from that perspective.

PR #3921, opened 2021-03-17
Allow custom definition for _sympy_pass_through

### LINKED ISSUE #3916: Add a way to mark types that should be passed-through parameter resolution.

During parameter resolution, we use the (`sympy_pass_through`)[https://github.com/quantumlib/Cirq/blob/09bb8ebca6322386316938494281b8f151803a5a/cirq/study/resolver.py#L238] function to detect values that should be considered as "atomic" for purposes of parameter resolution. The list of types that are considered "atomic" in this sense is hard-coded in the function and not extensible.

We'd like to be able to resolve to custom "value" types beyond just those included in `_sympy_pass_through` (builtin numeric types or sympy constants). One way to do this, for example, would be to define a protocol method that custom types could implement to mark themselves for pass-through. This protocol would be checked in `_sympy_pass_through` instead of just comparing against a fixed list of types and special symbolic values. It's possible there's another way to do this; just glancing at `ParamResolver.value_of` I find it quite hard to follow :-)

P2 - we should do it in the next couple of quarters

zchen088:
Fixes #3916


cirq/study/resolver.py
```diff
@@ -244,4 +244,9 @@ def _sympy_pass_through(val: Any) -> Optional[Any]:
         return val.p / val.q
     if val == sympy.pi:
         return np.pi
+
+    getter = getattr(val, '_sympy_pass_through_', None)
```
maffoo, line 248: I think it'd be helpful to change the name, because the point of defining a special method is to broaden this beyond the hard-coded sympy handling. In addition, the code above allows transforming values, not just passing them through (e.g. sympy integers become plain old integers). So maybe we could call the special method something like `_resolver_value_` and if it exists we call it to get a value to pass through, which could be different from `val` itself. The method could return `NotImplemented` to indicate that the value should _not_ be passed through. We could also change this function itself to use `NotImplemented` as the sentinel value instead of `None` if we want to allow things to resolve to `None` (the function is only called in `value_of` above, so that change wouldn't affect any other code).

maffoo (commented):
This seems reasonable to me. Once we settle on the design, we should update the `value_of` docs (and probably the `resolve_parameters` protocol docs) to describe this mechanism and how it can be used on custom types that one might want to use when resolving parameters.

cirq/study/resolver.py
zchen088: > We could also change this function itself to use NotImplemented as the sentinel value instead of None if we want to allow things to resolve to None (the function is only called in value_of above, so that change wouldn't affect any other code).

I tried this but it turns out that we're relying on `None` as a sentinel in some subtle way that I couldn't quite unravel, probably because of this line: https://github.com/quantumlib/Cirq/blob/master/cirq/study/resolver.py#L175

95-martin-orion: That `None` was me :p A valid workaround would be to replace `None` with a file-local class used only for flagging potential loops in recursive parameter evaluation.

cirq/study/resolver_test.py
```diff
@@ -232,12 +232,11 @@ def test_resolve_unknown_type():
...
     class Foo:
-        def _sympy_pass_through_(self):
-            return True
+        def _resolver_value_(self):
+            return self
 
     class Bar:
-        def _sympy_pass_through_(self):
-            return False
+        pass
```
maffoo, line 239: maybe add a case here that has `_resolver_value_` method that returns `NotImplemented`?

cirq/study/resolver.py
95-martin-orion: Hmm...I guess replacing wouldn't resolve this entirely, since `None` is checked for earlier in `value_of`.

zchen088: @95-martin-orion thanks for the hint, got all the tests to pass now, check if this is what you had in mind?

cirq/study/resolver_test.py
```diff
@@ -197,14 +197,14 @@ def test_unbound_recursion_halted():
...
-    r = cirq.ParamResolver({a: a})
-    assert r.value_of(a) == a
+    # # Non-recursive resolution ignores loops
+    # r = cirq.ParamResolver({a: b, b: a})
+    # assert r.value_of(a, recursive=False) == b
+    # assert r.value_of(r.value_of(a, recursive=False), recursive=False) == a
+    #
+    # # Self-definition is OK (this is a terminal symbol)
+    # r = cirq.ParamResolver({a: a})
+    # assert r.value_of(a) == a
```
95-martin-orion, line 207: Uncomment test?

cirq/study/resolver.py
95-martin-orion: I'd put a leading underscore on `RecursionFlag` just to make it abundantly clear that it's a non-public object, but otherwise this looks right.

95-martin-orion:
> This seems reasonable to me. Once we settle on the design, we should update the `value_of` docs (and probably the `resolve_parameters` protocol docs) to describe this mechanism {...}

Our requirement for "magic methods" (i.e. `_functions_like_this_`) is that they have an associated protocol. For `_resolve_parameters_`, this protocol is `SupportsParameterization`:
https://github.com/quantumlib/Cirq/blob/8cf782554adbafed724987de3067de7ca565fa0c/cirq/protocols/resolve_parameters.py#L28

To follow the same rule for `_resolver_value_`, we should add a protocol (in `resolve_parameters.py` is fine) for that method.

cirq/study/resolver.py
maffoo: +1 to leading underscore. I'd also suggest adding a comment that this is used to mark values that are being resolved recursively to detect loops.

maffoo:
@95-martin-orion, the problem with adding this in `SupportsParametrization` is that this new method `_resolver_value_` is for _values_ that we want the resolver to deal with, not parametrized objects themselves. We could add a new `ResolvableValue` protocol for this, but while we have top-level methods to invoke most protocols (like `cirq.resolve_parameters` for `SupportsParametrization`) I don't think it makes sense to add a top-level method in this case. It's really an internal detail of how the existing parameter resolution protocol works.

95-martin-orion:
> @95-martin-orion, the problem with adding this in `SupportsParametrization` is that this new method `_resolver_value_` is for _values_ that we want the resolver to deal with, not parametrized objects themselves. We could add a new `ResolvableValue` protocol for this, but while we have top-level methods to invoke most protocols (like `cirq.resolve_parameters` for `SupportsParametrization`) I don't think it makes sense to add a top-level method in this case. It's really an internal detail of how the existing parameter resolution protocol works.

Protocols do not require top-level methods - see `SerializableByKey` (although, due to imnplementation details, this protocol no longer has a "magic method" associated with it):
https://github.com/quantumlib/Cirq/blob/06e48928ed9a504dba0985ed9a4fba33709a783a/cirq/protocols/json_serialization.py#L330

cirq/study/resolver.py
```diff
@@ -237,6 +237,16 @@ def _from_json_dict_(cls, param_dict, **kwargs):
         return cls(dict(param_dict))
 
 
+class ResolvableValue(Protocol):
+    @doc_private
+    def _resolver_value_(self) -> Any:
```
maffoo, line 243: Maybe we should call it `_resolved_value_` to match what's in the docstring below (I think that explanation is pretty clear). Also, I'd suggest moving this protocol into `resolve_parameters.py` alongside the other protocols that are related to parameter resolution, so we can document the whole process there.

cirq/study/resolver.py
```diff
@@ -237,6 +237,16 @@ def _from_json_dict_(cls, param_dict, **kwargs):
...
+        """Returns a resolved value during parameter resolution.
+
+        Use this to mark a custom type as "resolved", instead of requiring
+        further parsing like we do with Sympy symbols.
+        """
+
+
 def _sympy_pass_through(val: Any) -> Any:
```
maffoo, line 251: Rename this `_resolve_value`?

cirq/study/resolver_test.py
```diff
@@ -230,6 +230,32 @@ def test_resolve_unknown_type():
     assert r.value_of(cirq.X) == cirq.X
 
 
+def test_custom_sympy_pass_through():
```
95-martin-orion, line 233: `test_custom_sympy_pass_through` -> `test_custom_resolved_value`

95-martin-orion (approved):
One remaining nit, then this looks good to go.

maffoo (approved):
LGTM

PR #3922, opened 2021-03-17
Change type signature of cirq.resolve_parameters to preserve types

### LINKED ISSUE #3390: Stronger typing for cirq.resolve_parameters

Currently the type of `cirq.resolve_parameters` is:
```python
def resolve_parameters(
        val: Any,
        param_resolver: 'cirq.ParamResolverOrSimilarType') -> Any:
```
Since the result is `Any`, we lose type information whenever we resolve parameters. However, for many important cirq objects (gates, operations, moments, circuits) it is typically the case that resolving parameters returns an object of the same type. To avoid losing type information it'd be nice to have a signature like:

```python
def resolve_parameters(
        val: T,
        param_resolver: 'cirq.ParamResolverOrSimilarType') -> T:
```

Not sure the best way to do this. We probably need a separate top-level method for type-preserving parameter resolution, and a separate `_special_method_` that classes can implement to indicate that their type is preserved under parameter resolution. It would seem to me desirable to make type-preserving resolution be the default since as I said that is most common for cirq types; the exceptions are low-level things like sympy symbols whose type can obviously change when resolved, but which are not cirq types and typically don't need to be resolved directly. One possibility would be to leave `cirq.resolve_parameters` as it is and add `cirq.resolve` as a type-preserving resolution function.

Can use type casts to recover type information after resolution, but this is ugly and error-prone. Can also have more specifically typed methods on particular classes that can be called instead of the generic `cirq.resolve_parameters` protocol, but this adds boilerplate and highlights the problem with the resolution protocol itself.

I think the actual amount of work needed to implement this will be modest, once we settle on a strategy.

maffoo:
Fixes #3390

Only had to modify a couple of places where mypy complained about types in `DiagonalGate` and `TwoQubitDiagonalGate`, which accepted `List` but stored a tuple internally. Also had to add a couple of casts in `resolve_parameters`, but I've tried to document what's going on there.

### DIFF
--- a/cirq/ops/diagonal_gate.py
+++ b/cirq/ops/diagonal_gate.py
@@ -65,7 +65,7 @@ def _gen_gray_code(n: int) -> Iterator[Tuple[int, int]]:
 class DiagonalGate(raw_types.Gate):
     """A gate given by a diagonal (2^n)\\times(2^n) matrix."""
 
-    def __init__(self, diag_angles_radians: List[value.TParamVal]) -> None:
+    def __init__(self, diag_angles_radians: Sequence[value.TParamVal]) -> None:
         r"""A n-qubit gate with only diagonal elements.
 
         This gate's off-diagonal elements are zero and it's on diagonal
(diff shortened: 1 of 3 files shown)


balopat (approved):
LGTM! Thanks for keeping our types tidy!

PR #3924, opened 2021-03-17
Return type annotations for `def _eigen_components()`

vtomole:


95-martin-orion (approved):
Looks good to me (modulo lint fixes)

PR #3925, opened 2021-03-17
Clean up _resolve_parameters_ method signatures

maffoo:
Follow-up to #3922. Clean up `_resolve_parameters_` method signatures by adding types where appropriate and using consistent argument names.

### DIFF
--- a/cirq/ops/eigen_gate.py
+++ b/cirq/ops/eigen_gate.py
@@ -22,6 +22,7 @@ from typing import (
     NamedTuple,
     Optional,
     Tuple,
+    TYPE_CHECKING,
     TypeVar,
     Union,
 )
@@ -36,6 +37,8 @@ from cirq import value, protocols
 from cirq.ops import raw_types
 from cirq.type_workarounds import NotImplementedType
 
+if TYPE_CHECKING:
+    import cirq
 
 TSelf = TypeVar('TSelf', bound='EigenGate')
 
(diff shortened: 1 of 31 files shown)


cirq/ops/eigen_gate.py
```diff
@@ -349,8 +352,10 @@ def _is_parameterized_(self) -> bool:
     def _parameter_names_(self) -> AbstractSet[str]:
         return protocols.parameter_names(self._exponent)
 
-    def _resolve_parameters_(self: TSelf, param_resolver, recursive: bool) -> 'EigenGate':
-        return self._with_exponent(exponent=param_resolver.value_of(self._exponent, recursive))
+    def _resolve_parameters_(
+        self: TSelf, resolver: 'cirq.ParamResolver', recursive: bool
+    ) -> 'EigenGate':
```
95-martin-orion, line 357: Should this be `TSelf` instead of `EigenGate`, to match `self`?

95-martin-orion (approved):
One nit, otherwise LGTM.

Thanks for this cleanup! I feel at least partly responsible following my "recursive parameter resolution" rampage :P

cirq/ops/eigen_gate.py
maffoo: I tried that, but it doesn't work because `self._with_exponent` does not return `TSelf`. If you change that also, there are problems with eigengate subclasses like `_PauliX` since those can return a different type when the exponent is changed (e.g. a generic `XPowGate` in stead of `_PauliX`). I think we'll have to live with the less specific types here. I did remove the `TSelf` annotation here since it wasn't doing anything.

95-martin-orion: SGTM.

PR #3926, opened 2021-03-17
Simplify type annotations in mps_simulator.py

maffoo:
These give me type errors when running mypy locally. It seems to happen on intermittently, for reasons I haven't full understood, but in any event it seems better to just simplify them since these refer to types defined in the same module.


95-martin-orion (approved):
I'm also seeing the intermittent mypy failures here - my first thought was some sort of import-order nondeterminism. Were you able to confirm whether this PR resolved those?

maffoo:
@95-martin-orion, with this change I haven't seen any mypy errors in mps simulator, so I think this fixes it.

PR #3927, opened 2021-03-17
Make SimulatesExpectationValues public

95-martin-orion:
I forgot to make the `SimulatesExpectationValues` interface visible at the top level of Cirq :(

You can still use this type in Cirq v0.10, but it must be referenced as `cirq.sim.simulator.SimulatesExpectationValues`.

### DIFF
--- a/cirq/__init__.py
+++ b/cirq/__init__.py
@@ -372,6 +372,7 @@ from cirq.sim import (
     sample_state_vector,
     sample_sweep,
     SimulatesAmplitudes,
+    SimulatesExpectationValues,
     SimulatesFinalState,
     SimulatesIntermediateState,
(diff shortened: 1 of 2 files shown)


balopat (approved):
LGTM! We could release a 0.10.1 if we want to with this + https://github.com/quantumlib/Cirq/pull/3887 (Mike was asking for it for TFQ).

95-martin-orion:
> LGTM! We could release a 0.10.1 if we want to with this + #3887 (Mike was asking for it for TFQ).

Unless it's urgent for TFQ (or it would complicate the patch-release), let's hold off a bit longer on this. I'm gradually bringing qsim up to speed with Cirq v0.10, and there's a decent chance I'll come across more little fix-ups like this.

PR #3934, opened 2021-03-18
Add docstring for cirq.QasmOutput

### LINKED ISSUE #3658: Add docstring for `cirq.QasmOutput`
The use case is documentation.

Basically something like the docstring used here: https://github.com/quantumlib/Cirq/blob/d58423a3f06dd0a40fd8acfdd006d9a876274a7c/cirq/circuits/qasm_output.py#L86

We should do it in the next couple of quarters

unaiic:
Adds docstring for `cirq.QasmOutput` and fixes formatting.

Fixes #3658.


cirq/circuits/qasm_output.py
```diff
@@ -167,6 +167,18 @@ def __init__(
         precision: int = 10,
         version: str = '2.0',
     ) -> None:
+        """Representation of a circuit in QASM format.
+
+        Args:
+            operations: An operation or tree of operations to insert.
```
vtomole, line 173: From the type annotation, we should only pass the tree of operations. Please remove `An operation or`

vtomole (changes requested):

cirq/circuits/qasm_output.py
unaiic: Okay, changes are already made.

cirq/circuits/qasm_output.py
```diff
@@ -167,6 +167,18 @@ def __init__(
         precision: int = 10,
         version: str = '2.0',
     ) -> None:
+        """Representation of a circuit in QASM format.
+
+        Args:
+            operations: Tree of operations to insert.
+            qubits: The qubits used in the operations.
+            header: A multi-line string that is placed in a comment at the top
+                of the QASM. Defaults to a cirq version specifier.
```
vtomole, line 176: > Defaults to a cirq version specifier.

What does this mean? The default is an empty string.: `header: str = '',`

vtomole (commented):
I got one question then this should be good!

cirq/circuits/qasm_output.py
unaiic: You are right, my bad. Should it be something like _Defaults to an empty string._, or would it be better to just leave it without that part?

vtomole: Let's leave it out.

unaiic: Done.

vtomole (approved):
LGTM. Thanks!

PR #3935, opened 2021-03-19
Example of stabilizer code

tonybruguier:
This is based on the thesis by Daniel Gottesman.


examples/stabilizer_code.py
```diff
@@ -0,0 +1,192 @@
...
+    def encode(self, circuit, qubits):
+        # Equation 4.8:
+        for r, x in enumerate(self.X):
+            for j in range(self.r, self.n - self.k):
+                if x[j] == 'X' or x[j] == 'Y':  # coverage: ignore
+                    circuit.append(  # coverage: ignore
+                        cirq.ControlledOperation(  # coverage: ignore
+                            [qubits[self.n - self.k + r]], cirq.X(qubits[j])  # coverage: ignore
+                        )  # coverage: ignore
+                    )  # coverage: ignore
```
95-martin-orion, line 141: This seems like an important distinct path. Could you add test coverage for it?

examples/stabilizer_code.py
```diff
@@ -0,0 +1,192 @@
...
+
+        for qid in range(self.n):
+            for op in ['X', 'Z']:
+                syndrome = [
+                    1 if self.M[r][qid] == 'I' or self.M[r][qid] == op else -1
+                    for r in range(self.r)
+                ]
+                self.syndromes_to_corrections[tuple(syndrome)] = (op, qid)
+
+    def encode(self, circuit, qubits):
```
95-martin-orion, line 132: It may be simpler to create a new `Circuit` in this method and return it, rather than passing in a circuit and mutating it. For example, this:
```
circuit = Circuit({...})
self.encode(circuit, qubits)
```
would become this:
```
circuit = Circuit({...}) + self.encode(qubits)
```

examples/stabilizer_code.py
```diff
@@ -0,0 +1,192 @@
...
+            for n in range(self.n):
+                if n == r:
+                    continue
+                if self.M[r][n] == 'I':
+                    continue
+                op = gate_dict[self.M[r][n]]
+                circuit.append(cirq.ControlledOperation([qubits[r]], op(qubits[n])))
+        # At this stage, the state vector should be equal to equations 3.17 and 3.18.
+
+    def correct(self, circuit, qubits, ancillas):
```
95-martin-orion, line 160: As with `encode`, it may be simpler to generate and return a `correction` circuit here, rather than passing in an existing circuit.

examples/stabilizer_code_test.py
```diff
@@ -0,0 +1,39 @@
...
+
+    code.correct(circuit, qubits, ancillas)
+
+    results = cirq.Simulator().simulate(
+        circuit, qubit_order=(qubits + ancillas), initial_state=(input_val * 2 ** 4)
+    )
+
+    qubit_map = {qubit: i for i, qubit in enumerate(qubits + ancillas)}
+    pauli_string = cirq.PauliString(dict(zip(qubits, code.Z[0])))
+    trace = pauli_string.expectation_from_state_vector(results.state_vector(), qubit_map)
```
95-martin-orion, line 25: IIUC, this is performing the "decode" behavior for the circuit. It might be worthwhile to package this as a function in the `StabilizerCode` class, then call that method here.

95-martin-orion (changes requested):
First round of comments - I'll need to read the paper more closely to review the details here.

examples/stabilizer_code.py
```diff
@@ -0,0 +1,192 @@
...
+            return out
+
+        self.M = _BuildByCode(M)
+        self.X = _BuildByCode(X)
+        self.Z = _BuildByCode(Z)
+        self.syndromes_to_corrections = {}
+
+        for qid in range(self.n):
+            for op in ['X', 'Z']:
+                syndrome = [
```
95-martin-orion, line 126: Tuples support list-comprehension-like behavior, so you can replace this `syndrome = [...]` with `syndrome = tuple(...)` to save a conversion step.

95-martin-orion (changes requested):
One more note from the close inspection.

Overall this looks like a nice translation of the paper! Do you have plans (maybe as a follow-up PR) to also support applying operations to the encoded states?

examples/stabilizer_code.py
```diff
@@ -0,0 +1,192 @@
...
+        M = np.zeros((n - k, 2 * n), np.int8)
+        for i, group_generator in enumerate(group_generators):
+            for j, c in enumerate(group_generator):
+                if c == 'X' or c == 'Y':
+                    M[i, j] = 1
+                elif c == 'Z' or c == 'Y':
+                    M[i, n + j] = 1
+
+        # Performing the Gaussian elimination as in section 4.1
+        def _GaussianElimination(M, min_row, max_row, min_col, max_col):
```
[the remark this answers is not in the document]
tonybruguier: Done.

examples/stabilizer_code.py
```diff
@@ -0,0 +1,192 @@
...
+                np.eye(k, dtype=np.int8),
+            ],
+            axis=1,
+        )
+
+        self.n = n
+        self.k = k
+        self.r = r
+
+        def _BuildByCode(mat):
```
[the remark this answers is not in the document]
tonybruguier: Done.

examples/stabilizer_code.py
tonybruguier: It was indeed. I added a test and it revealed several issues when I was conflating self.r with self.n - self.k

Thanks for the careful review! I am not super confident about my code (my goal was to learn, so I am far from being an expert!).

examples/stabilizer_code.py
```diff
@@ -0,0 +1,192 @@
...
+        # Build the matrix defined in section 3.4
+        M = np.zeros((n - k, 2 * n), np.int8)
+        for i, group_generator in enumerate(group_generators):
+            for j, c in enumerate(group_generator):
+                if c == 'X' or c == 'Y':
+                    M[i, j] = 1
+                elif c == 'Z' or c == 'Y':
+                    M[i, n + j] = 1
+
+        # Performing the Gaussian elimination as in section 4.1
```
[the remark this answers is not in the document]
tonybruguier: I had looked for it, but it seemed difficult because the Gaussian elimination must be done in GF(2) not in R. Another difference is that the pivot must be chosen only in the sub-matrix E of equation 4.1.

Leaving unresolved for now.

examples/stabilizer_code.py
tonybruguier: Done.

examples/stabilizer_code.py
tonybruguier: Done. Looks nicer indeed.

examples/stabilizer_code.py
tonybruguier: Done.

examples/stabilizer_code_test.py
tonybruguier: Like this?

tonybruguier:
Thanks for the review.

I tried to address your comments. For the Gaussian elimination, I had indeed considered using an existing package, but I couldn't find anything suitable. Please let me know what you think.

Also, I am quite new to that error correction theory, so this is an exercise to understand it better. Your review uncovered some bugs, so thank you for this.

PTAL when you can.

examples/stabilizer_code.py
```diff
@@ -8,8 +8,65 @@
 # https://thesis.library.caltech.edu/2900/2/THESIS.pdf
 
 
-class StabilitizerCode(object):
-    def __init__(self, group_generators):
+def _BuildByCode(mat):
```
95-martin-orion, line 11: Docstrings and type annotations for this and `_GaussianElimination` would aid readability.

examples/stabilizer_code.py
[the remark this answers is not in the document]
95-martin-orion: That seems fair to me. It might be worthwhile to note this in a docstring for `_GaussianElimination` above.

examples/stabilizer_code.py
```diff
@@ -98,47 +117,29 @@ def _GaussianElimination(M, min_row, max_row, min_col, max_col):
...
+                syndrome = tuple(
                     1 if self.M[r][qid] == 'I' or self.M[r][qid] == op else -1
-                    for r in range(self.r)
-                ]
-                self.syndromes_to_corrections[tuple(syndrome)] = (op, qid)
+                    for r in range(self.n - self.k)
+                )
+                self.syndromes_to_corrections[syndrome] = (op, qid)
+
+    def encode(self, qubits):
```
95-martin-orion, line 133: As with the external functions, docstrings and type annotations on these methods (`encode`, `correct`, and `decode`) would make this example easier to read.

examples/stabilizer_code_test.py
```diff
@@ -3,37 +3,60 @@
...
+                    encode_corrupt_correct(five_qubit_code, input_val, error_gate, error_loc)
+                    == input_val
+                )
+
+
+def test_imperfect_code():
+    # Also known as the bit-flip code.
+    seven_qubit_code = sc.StabilizerCode(group_generators=['ZZI', 'ZIZ'], allowed_errors=['X'])
+
+    for error_gate in [cirq.X]:
```
95-martin-orion, line 56: Would be interesting to also test here that `cirq.Z` errors are _not_ corrected by this code.

95-martin-orion (dismissed):
Mostly documentation requests on this pass.

balopat:
A couple of thoughts: 
- I would like to have this reviewed by someone who actually works with Stabilizer codes @viathor @Strilanc 
- What is the goal of this example?  How does it serve the community? 
      - if it is meant to be of educational value - we should add significantly more comments to explain what is happening - I would also consider making this into a notebook instead of a script (e.g. like we have Shor's algorithm or the upcoming Forrelation notebook) - also a minor thing, that we should add it to the docs/tutorials/_index.yaml file (that's the one that creates the [tutorials and examples](https://quantumai.google/cirq/tutorials) page) in the advanced section
      - I don't think it is meant to be a starting point for someone doing actual stabilizer code experiments with Cirq - as that is out of scope as we said on Cirq Cynque - in this case we should make this somehow clear...

balopat:
> A couple of thoughts:
> 
> * I would like to have this also reviewed by someone who works with stabilizer codes @viathor @Strilanc
> * What is the goal of this example?  How does it serve the community?
>   - if it is meant to be of educational value - we should add significantly more comments to explain what is happening - I would also consider making this into a notebook instead of a script (e.g. like we have Shor's algorithm or the upcoming Forrelation notebook) - also a minor thing, that we should add it to the docs/tutorials/_index.yaml file (that's the one that creates the [tutorials and examples](https://quantumai.google/cirq/tutorials) page) in the advanced section
>   - I don't think it is meant to be a starting point for someone doing actual stabilizer code experiments with Cirq - as that is out of scope as we said on Cirq Cynque - in this case we should make this somehow clear...

This is not to say that I'm not interested in having this in the repo - I just want to have the conversation about how the value of this contribution can be maximized - and also @95-martin-orion's review points are great (I haven't read the thesis yet fully), I'm just curious to hear expert input as well.

examples/stabilizer_code.py
tonybruguier: Done.

examples/stabilizer_code.py
```diff
@@ -70,8 +89,8 @@ def _GaussianElimination(M, min_row, max_row, min_col, max_col):
         # B = M[0:r, n : (n + r)]
         C1 = M[0:r, (n + r) : (2 * n - k)]
         C2 = M[0:r, (2 * n - k) : (2 * n)]
-        # D = M[r : (2 * r), n : (n + r)]
-        E = M[r : (2 * r), (2 * n - k) : (2 * n)]
+        # D = M[r : (2 * n), n : (n + r)]
+        E = M[r : (2 * n), (2 * n - k) : (2 * n)]
```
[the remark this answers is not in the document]
tonybruguier: No, it was a plain mistake on my part. Sorry about this and thanks for catching this.

examples/stabilizer_code_test.py
tonybruguier: So I am having trouble with this, because the current code does seem to correct these errors... I debugged the code:
https://github.com/tonybruguier/sandbox/blob/master/debug_ecc.txt

I printed the state vector right after calling `simulate()`, and I think the code itself behaves correctly. The first 3 / 5 qubits are the code words and the last 2/4 ancillas are the syndromes. 

We I do see that the Z errors result in an incorrect state vector only for the bit flip code, but the Pauli string measurements do not capture this. I think this is because there is a global phase. How do you think I should test?

Leaving unresolved, because I think there's more work to do.

examples/stabilizer_code.py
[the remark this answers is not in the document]
tonybruguier: Done.

tonybruguier:
Thanks for the reviews.

@95-martin-orion 
I left one of your comments unresolved, because I didn't think I fully addressed it. I think the code is correct (never sure of course) but the measurements discard a global phase?

@balopat 
Yes, as we had discussed, I don't intend this to become a part of the core Cirq functionalities. It is indeed more as an educational value (for myself at first!) and I'm happy to change its form, or even drop it (since I've learned quite a bit already).

95-martin-orion:
> @95-martin-orion
> I left one of your comments unresolved, because I didn't think I fully addressed it. I think the code is correct (never sure of course) but the measurements discard a global phase?

That 's the correct behavior - a global phase shouldn't affect measurement results. In order for the Z-error to make a difference, I guess there would need to be some encoded operations applied. Since that's not supported here, I recommend leaving a TODO to test Z-errors if and when encoded ops are supported.

tonybruguier:
Added a TODO

95-martin-orion (commented):
Once type annotations are added, this has my approval from a code-format perspective. Leaving the QEC review and final approval to @viathor / @Strilanc.

tonybruguier:
Thanks for the reviews so far. I added the type annotations (which was the last request from Orion, I think) and took the liberty of adding a link in the tutorials (I don't think it was agreed either way, but this way we won't forget).

I think it's now @viathor 's turn to take a look at this - when time permits.

ybc1991 (commented):
@tonybruguier  Thanks for this inspiring example! I left a few comments.

I also found an implementation of the stabilizer code in pyQuil [reports](https://cs269q.stanford.edu/projects2019/stabilizer_code_report_Y.pdf) and [github](https://github.com/shubhamchandak94/stabilizer_code/). Maybe this can be helpful.

examples/stabilizer_code.py
```diff
@@ -0,0 +1,279 @@
...
+                        cirq.ControlledOperation([qubits[self.n - self.k + r]], cirq.X(qubits[j]))
+                    )
+
+        gate_dict = {'X': cirq.X, 'Y': cirq.Y, 'Z': cirq.Z}
+
+        for r in range(self.r):
+            circuit.append(cirq.H(qubits[r]))
+
+            if self.M[r][r] == 'Y' or self.M[r][r] == 'Z':
+                circuit.append(cirq.S(qubits[r]))
```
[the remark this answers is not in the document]
tonybruguier: Added a comment. I was actually following the improvements of the reference you gave above.

examples/stabilizer_code.py
```diff
@@ -0,0 +1,279 @@
...
+        E = M[r : (n - k), (2 * n - k) : (2 * n)]
+
+        X = np.concatenate(
+            [
+                np.zeros((k, r), dtype=np.int8),
+                E.T,
+                np.eye(k, dtype=np.int8),
+                np.mod(E.T @ C1.T + C2.T, 2),
+                np.zeros((k, n - k - r), np.int8),
+                np.zeros((k, k), np.int8),
```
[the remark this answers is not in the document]
tonybruguier: Done.

examples/stabilizer_code_test.py
```diff
@@ -0,0 +1,80 @@
...
+
+    decoded = code.decode(qubits, ancillas, results.state_vector())
+
+    return decoded[0]
+
+
+def test_no_error():
+    # Table 3.2.
+    five_qubit_code = sc.StabilizerCode(
+        group_generators=['XZZXI', 'IXZZX', 'XIXZZ', 'ZXIXZ'], allowed_errors=['X', 'Z']
```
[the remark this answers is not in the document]
tonybruguier: Done. I also found a bug, thanks to your suggestion: When doing the Gaussian elimination, the previous test never swapped columns. However, with the Steane code, we do end up swapping columns, but if we do that, we must do the same operation in *both* halves of the matrix. Thanks for catching this bug!

examples/stabilizer_code.py
```diff
@@ -0,0 +1,275 @@
...
+        # Get matrix sub-components, as per equation 4.3:
+        # A1 = M[0:r, r : (n - k)]
+        A2 = M[0:r, (n - k) : n]
+        # B = M[0:r, n : (n + r)]
+        C1 = M[0:r, (n + r) : (2 * n - k)]
+        C2 = M[0:r, (2 * n - k) : (2 * n)]
+        # D = M[r : (n - k), n : (n + r)]
+        E = M[r : (n - k), (2 * n - k) : (2 * n)]
+
+        X = np.concatenate(
```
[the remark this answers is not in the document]
tonybruguier: Done.

examples/stabilizer_code.py
```diff
@@ -0,0 +1,275 @@
...
+                )
+                self.syndromes_to_corrections[syndrome] = (op, qid)
+
+    def encode(self, qubits):
+        """
+        Creates a circuit that encodes the qubits using the code words.
+
+        Args:
+            qubits: The list of qubits where to encode the message. This should be a vector of
+            length self.n where the last self.k qubits are the un-encoded qubits.
```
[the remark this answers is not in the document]
tonybruguier: Humm. I am not sure it's cleaner. I did it anyway, but please take a look and let me know what you think.

tonybruguier:
Thanks for the review. PTAL.

examples/stabilizer_code.py
```diff
@@ -0,0 +1,302 @@
...
+                        )
+                    )
+
+        gate_dict = {'X': cirq.X, 'Y': cirq.Y, 'Z': cirq.Z}
+
+        for r in range(self.r):
+            circuit.append(cirq.H(qubits[r]))
+
+            if self.M[r][r] == 'Y' or self.M[r][r] == 'Z':
+                circuit.append(cirq.S(qubits[r]))
```
ybc1991, line 215: After reading the Thesis, its errata, and the reports, I think I fully understand this. Let's double-check if we understand on the same page:

1. Here `if self.M[r][r] == 'Y':` should be sufficient since in standard form it can never be `Z` in first `r` rows. Also, it has to be, otherwise, the argument of encoding is wrong.
2. The reason for adding `S` gate is `Y` gate we used is the complex format (i.e. to make it Hermitian). IIUC, the logic is this. First, consider the encoding `I+M_1` on initial state. It has following four cases: (ignore the phase factor)

> `(I+X@P_2...P_k)|0...0> = |0...0> + |1>|\psi>`
> `(I+Y@P_2...P_k)|0...0> = |0...0> + i|1>|\psi>`
> `(I+I@P_2...P_k)` and `(I+Z@P_2...P_k)` is not possible in standard form

The first case means we need `[1,1]` vector and controlled gates and in the second case we need `[1, i]` vector and controlled gates. Corresponding, it is the first column of `H` and the first column of `SH` respectively. 

Next is to build `I+M_l` and the trick is that all generators of stabilizer commute. Hence all implementation `I+M_l` is equivalent to apply in the first place, i.e. considering its effect on `|0...0>` only and then move this back to place. (Original thesis make this mismake not make entire gates back to place). So the argument repeats the same.

3. This implementation is limited to the case that all of the signs of the stabilizer generators (in standard form) are +1. It is possible that "Codes with other signs can be achieved via a tensor product Pauli operator." That is stated in Proposition 10.4 in Nielsen & Chuang's book. I am not sure hows easy makes it systematically

tonybruguier: Yes, that was indeed my understanding, but I have to re-read it every time. This is where I am the most iffy in my understanding, if I may be honest.

I took the liberty of adding comments that were largely inspired from what you wrote here. I hope that's OK.

tonybruguier:
Added some comments. Also added a Jupyter notebook:
https://github.com/quantumlib/Cirq/blob/9c3e18372c8b575efd8b125aa9a75e86937e6e21/examples/stabilizer_code.ipynb

PTAL

examples/stabilizer_code.py
```diff
@@ -0,0 +1,318 @@
...
+
+        Z = np.concatenate(
+            [
+                np.zeros((k, n), dtype=np.int8),
+                A2.T,
+                np.zeros((k, n - k - r), dtype=np.int8),
+                np.eye(k, dtype=np.int8),
+            ],
+            axis=1,
+        )
```
ybc1991, line 159: I feel it is better to put it into a function like -- `transfer_to_standard_form(M, n, k) -> Tuple[standard_M, X, Z, r] `

examples/stabilizer_code.py
```diff
@@ -0,0 +1,318 @@
...
+        self.k: int = k
+        self.r: int = r
+        self.M: List[str] = _BuildByCode(M)
+        self.logical_Xs: List[str] = _BuildByCode(X)
+        self.logical_Zs: List[str] = _BuildByCode(Z)
+
+        self.syndromes_to_corrections = {}
+
+        for qid in range(self.n):
+            for op in allowed_errors:
```
ybc1991, line 171: Actually, I think this `allowed_errors` is not clear. (Like what value is permitted?) Please add comments in the top.

examples/stabilizer_code.py
tonybruguier: Done.

examples/stabilizer_code.py
```diff
@@ -0,0 +1,318 @@
...
+            # (I+X@P_2...P_k)|0...0> = |0...0> + |1>|\psi>
+            # (I+Y@P_2...P_k)|0...0> = |0...0> + i|1>|\psi>
+            # The other forms are not possible in the standard form, by construction.
+
+            # The first case means we need [1,1] vector and controlled gates and in the
+            # second case we need [1, i] vector and controlled gates. Corresponding, it is
+            # the first column of H and the first column of SH respectively.
+
+            # For the other stabilizers, the process can be repeated, as by definition they
+            # commute.
```
ybc1991, line 228: Actually, my previous comments is not correct. For the other `r-1`stabilizers, it can use the same style because, in the standard form, the first `r x r` block is `I`. Hence, the `r'` qubit doesn't have a bit-flip only until `r'` stabilizer group. This is the reason that we only need to consider the first column of `H` and `SH`.

The rest `n-k+1` can be omitted is indeed because the commutive property that it can move to the first places and 
`Z` has no effect in `|000>` so ignored.

examples/stabilizer_code.py
tonybruguier: How about correctable errors?

examples/stabilizer_code.py
ybc1991: Oh no, I was planned to send comment together.... But accidentally send it into single comments...

tonybruguier: Modified the comment.

tonybruguier:
PTAL (test errors appear not related to this change).

dev_tools/notebooks/isolated_notebook_test.py
```diff
@@ -58,6 +58,8 @@
     # skipping fidelity estimation due to
     # https://github.com/quantumlib/Cirq/issues/3502
     "examples/*fidelity*",
+    # Also skipping stabilizer code testing.
```
viathor, line 61: This should say why we are skipping.

examples/stabilizer_code_test.py
```diff
@@ -0,0 +1,90 @@
...
+                assert (
+                    encode_corrupt_correct(bit_flip_code, input_val, error_gate, error_loc)
+                    == input_val
+                )
+
+    # NOTE(tonybruguier): Even though the state vector shows that we cannot correct the error,
+    # the decoded qubit is incorrect but only because it has the wrong phase. Since the Pauli
+    # string measurement doesn't capture the phase it doesn't detect the error.
+    # TODO(tonybruguier): Have a unit test that captures the fact that this code cannot correct
+    # phase errors.
```
viathor, line 90: Is it just a matter of appending a Hadamard so that we measure in the X basis?

examples/stabilizer_code.py
```diff
@@ -0,0 +1,340 @@
...
+        ps = ''
+        for j in range(n):
+            if mat[i, j] == 0 and mat[i, j + n] == 0:
+                ps += 'I'
+            elif mat[i, j] == 1 and mat[i, j + n] == 0:
+                ps += 'X'
+            elif mat[i, j] == 0 and mat[i, j + n] == 1:
+                ps += 'Z'
+            else:
+                ps += 'Y'
```
viathor, line 34: Optional: The three conditionals can be simplified into

```python
k = 2*mat[i, j + n] + mat[i, j]
ps += "IXZY"[k]
```

examples/stabilizer_code.py
```diff
@@ -0,0 +1,340 @@
...
+# However, there were some mistakes in the thesis:
+# https://www2.perimeterinstitute.ca/personal/dgottesman/thesis-errata.html
+
+
+def _BuildByCode(mat: np.ndarray) -> List[str]:
+    """
+    Takes into input a matrix of Boolean interpreted as row-vectors, each having dimension 2 * n.
+    The matrix is converted into another matrix with as many rows, but this time the vectors
+    contain the letters I, X, Y, and Z representing Pauli operators.
+    """
```
viathor, line 21: Most (all?) cirq code follows the style in which docstring consists of succinct one-line description of the function, followed by empty line, followed by elaborated description, followed by empty line and args and return sections (where the elaborated description and args/returns sections are optional). Please look around cirq to see what this looks like in practice and try to stay consistent. Here is a silly example:

```python
def blah(size: int) -> Blah:
    """Computes blah.

    Blah is meaningless.

    Args:
        size: Size of the blah to compute.

    Returns:
        Blah of a specified size.
    """
```

examples/stabilizer_code.py
```diff
@@ -0,0 +1,340 @@
...
+        M: The input/output matrix
+        min_row: The minimum row (inclusive) where the perform the elimination.
+        max_row: The maximum row (exclusive) where the perform the elimination.
+        min_col: The minimum column (inclusive) where the perform the elimination.
+        max_col: The maximum column (exclusive) where the perform the elimination.
+
+    Returns:
+        The rank of the matrix.
+    """
+    assert M.shape[1] % 2 == 0
```
viathor, line 62: Assert should be used for conditions that are believed impossible to arise. To signal invalid arguments, prefer to raise `ValueError`.

examples/stabilizer_code.py
```diff
@@ -0,0 +1,340 @@
...
+
+        rank += 1
+
+    # Backward replacing to get identity
+    for r in reversed(range(rank)):
+        i = min_row + r
+        j = min_col + r
+
+        # Do the elimination.
+        for k in reversed(range(min_row, i)):
```
viathor, line 103: `reversed` must drain all elements from `range` and store them before it can begin returning them in the reversed order. You can avoid this by using `for k in range(i - 1, min_row - 1, -1)` (this will include `min_row` and exclude `i` the same as `reversed(range(min_row, i))`).

examples/stabilizer_code.py
```diff
@@ -0,0 +1,340 @@
...
+
+import cirq
+
+# Based on:
+# Stabilizer Codes and Quantum Error Correction
+# Daniel Gottesman
+# https://thesis.library.caltech.edu/2900/2/THESIS.pdf
+#
+# However, there were some mistakes in the thesis:
+# https://www2.perimeterinstitute.ca/personal/dgottesman/thesis-errata.html
```
viathor, line 13: Standard form of a stabilizer code is also described in section 10.5.7. on page 470 in Nielsen & Chuang. I'm not sure whether any of the mistakes are corrected there, but they probably have given the 13 years that passed from the time the thesis was written and the latest N&C edition. Moreover, N&C may be a more familiar reference for most readers.

examples/stabilizer_code.py
```diff
@@ -0,0 +1,340 @@
...
+    r: int = _GaussianElimination(M, 0, n - k, 0, n)
+    _ = _GaussianElimination(M, r, n - k, n + r, 2 * n)
+
+    # Get matrix sub-components, as per equation 4.3:
+    # A1 = M[0:r, r : (n - k)]
+    A2 = M[0:r, (n - k) : n]
+    # B = M[0:r, n : (n + r)]
+    C1 = M[0:r, (n + r) : (2 * n - k)]
+    C2 = M[0:r, (2 * n - k) : (2 * n)]
+    # D = M[r : (n - k), n : (n + r)]
```
viathor, line 138: Let's remove unused code rather than commenting it out.

examples/stabilizer_code.py
```diff
@@ -0,0 +1,340 @@
...
+        M, X, Z, r = _transfer_to_standard_form(M, n, k)
+
+        self.n: int = n
+        self.k: int = k
+        self.r: int = r
+        self.M: List[str] = _BuildByCode(M)
+        self.logical_Xs: List[str] = _BuildByCode(X)
+        self.logical_Zs: List[str] = _BuildByCode(Z)
+
+        self.syndromes_to_corrections = {}
```
viathor, line 190: I'm surprised this doesn't need a type annotation.

viathor (commented):
I had a first quick look and have a few comments. I'll go over it again later.

examples/stabilizer_code.py
ybc1991: [Add some background information] Yes, N&C book has description of standard form. But the error in the thesis is constructing encoding circuit from standard form stabilizer, which is not trivia at all. (You can tell that by Daniel himself made mistake in this. Twice.) The link Tony posted is the errata stated by Daniel, which is authentic. N&C's book doesn't describe the construction but leave it as problem 10.3 & 10.4. (Based on my own learning experience, it is not sufficient to list N&C book reference only here.) Maybe add more context in the comment is helpful. This is also good reason to introduce notebook since there are lots needed to be explained.

dev_tools/notebooks/isolated_notebook_test.py
tonybruguier: Done. I took the liberty of reusing the same issues, since the solution would likely be the same. I put a note in there.

examples/stabilizer_code_test.py
tonybruguier: I wasn't sure what you meant by this. I tried to add the Hadamard gates, but I wasn't sure how to measure.

Nevertheless, it seems that you are keen on getting it tested, and so I decided to test the state vector directly. I modified the code to return the state vector with the ancillas traced out. Then, I make sure that it's different for a Z error. I also tested that the state vector is fine for all the other correctable errors in the other tests.

I'm leaving this unresolved for now, since I am not sure that this is what you wanted.

examples/stabilizer_code.py
tonybruguier: Done.

examples/stabilizer_code.py
tonybruguier: Done.

examples/stabilizer_code.py
tonybruguier: Thanks, I didn't know about the behavior of reversed. Done.

examples/stabilizer_code.py
tonybruguier: Added a reference.

examples/stabilizer_code.py
tonybruguier: Done.

examples/stabilizer_code.py
tonybruguier: Tests seemed fine. Nevertheless, code would be clearer with an annotation so I added one.

examples/stabilizer_code.py
tonybruguier: It was my understanding of the difference between assert and ValueError indeed.

But that's what I intended to do, since I created M with: M = np.zeros((n - k, 2 * n), np.int8) and so it really should always have an even number of columns.

tonybruguier:
Thanks. PTAL.

tonybruguier:
@viathor Gentle ping (not super urgent, but it would be nice to get this in eventually -- if you still think it makes sense to have this example).

95-martin-orion (commented):
My previously-requested changes have been resolved - this good to go with approval from the theory side.

tonybruguier:
@viathor It seems both Yicheng and Martin and OK with this PR. You had a request and I added a test. Do you have any further comment on this PR?

viathor (approved):

examples/stabilizer_code.py
```diff
@@ -0,0 +1,340 @@
...
+# https://thesis.library.caltech.edu/2900/2/THESIS.pdf
+#
+# However, there were some mistakes in the thesis:
+# https://www2.perimeterinstitute.ca/personal/dgottesman/thesis-errata.html
+#
+# An alternative explanation can be found in exercises 10.3 and 10.4 of
+# "Quantum Computation and Quantum Information" by Michael Nielsen and Isaac Chuang.
+
+
+def _BuildByCode(mat: np.ndarray) -> List[str]:
```
vtomole, line 19: Curious: Why is the name of this function like this instead of `build_by_code`?

tonybruguier: Good point, I wasn't consistent with my naming convention. I think, though, that it should have an underscore first, but I otherwise renamed it. I took the liberty of also renaming GaussianElimination since it was suffering from the same inconsistency. Hope that it's OK.

tonybruguier:
Thanks. @vtomole  PTAL.

tonybruguier:
Hello,
Maybe this PR is ready to be merged? I'm happy to address comments if there are more, of course.

tonybruguier:
@vtomole 
It seems some of the existing test was flaky (kernel for notebook failed to install). I re-ran them and everything passed.

However, the automerge label got auto-removed. Could you add it back when you can?
Thanks again!

PR #3937, opened 2021-03-19
Fix a bug in module deprecator

balopat:
Adds `__dict__` to the wrapped module to ensure its member attributes stay on it.

### DIFF
--- a/cirq/_compat.py
+++ b/cirq/_compat.py
@@ -543,6 +543,8 @@ def _setup_deprecated_submodule_attribute(
     setattr(parent_module, old_child, new_module)
 
     class Wrapped(ModuleType):
+        __dict__ = parent_module.__dict__
+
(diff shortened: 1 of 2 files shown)


balopat:
ping @cduck this was a bug that causes issues when `create_attribute = True`.

cduck (approved):
LGTM.

balopat:
> LGTM. Did this only affect `dir`?

Yes, well, also things that were using it, e.g. `inspect.getmembers` - which is in turn used in the json serialization tests - which were failing after extracting cirq.google, that's how I spotted it.

Another thing to note that this was my mistake - the `deprecated_attributes` utility already does this `__dict__` redefinition, I just didn't understand where this was important, so I decided to remove it yesterday. Now I understand it better and added the test to protect it.

PR #3940, opened 2021-03-19
Adds git-blame-ignore-rev and instructions

balopat:
Adds git-blame-ignore-rev and instructions.

This sets up git blame, PyCharm and other clients that integrate with git well to ignore large scale refactoring revisions. Note that unfortunately this is not yet supported by Github (https://github.community/t/support-ignore-revs-file-in-githubs-blame-view/3256/21).

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# migrating to black
+a61e51b53612735e93b3bb8a7605030c499cd6c7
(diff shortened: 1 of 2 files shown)


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# migrating to black
+a61e51b53612735e93b3bb8a7605030c499cd6c7
```
tanujkhattar, line 2: Maybe also add the flynt commit? 

Commit id: 7a335e3ef297057b4a12f7c21d4e8dd426b44ca0
#3786

balopat: yeah, sure, let's do that!

tanujkhattar (approved):
LGTM % comments

PR #3942, opened 2021-03-19
Add integrated histogram visualizations

### LINKED ISSUE #3941: Integrated Histograms Visualization
Add the ability to draw integrated histograms in cirq. This is useful to visualize qubit error distributions.  

`cirq.integrated_histogram(data)` should plot an integrated histogram like:

image

n/a

See https://tinyurl.com/cirq-visualizations for more details.

This is part of the Cirq Roadmap

Relate issues - #3681 #3642

tanujkhattar:
This PR adds integrated histogram visualizations support to Cirq. Closes #3941

See https://tinyurl.com/cirq-visualizations for the larger roadmap item. 

**Usage:**
- `calibration.plot('single_qubit_errors')` now produces the following image: 

image

**Next steps**
- The `visualizing_calibration_metrics` tutorial will be updated in a follow up PR once this is checked in.

### DIFF
--- a/cirq/vis/histogram.py
+++ b/cirq/vis/histogram.py
@@ -1,0 +1,150 @@
+# Copyright 2021 The Cirq Developers
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Any, Mapping, Optional, Sequence, Union, SupportsFloat
+
+import numpy as np
+from matplotlib import pyplot as plt
+
+
+def integrated_histogram(
+    data: Union[Sequence[SupportsFloat], Mapping[Any, SupportsFloat]],
+    ax: Optional[plt.Axes] = None,
+    *,
+    cdf_on_x: bool = False,
+    axis_label: str = '',
+    semilog: bool = True,
+    median_line: bool = True,
+    median_label: Optional[str] = 'median',
+    mean_line: bool = False,
+    mean_label: Optional[str] = 'mean',
+    hide_zero: bool = True,
+    title: Optional[str] = None,
+    **kwargs,
+) -> plt.Axes:
+    """Plot the integrated histogram for an array of data.
+
+    Suppose the input is a list of gate fidelities. The x-axis of the plot will
+    be gate fidelity, and the y-axis will be the probability that a random gate
+    fidelity from the list is less than the x-value. It will look something like
+    this
+
+    1.0
+    |              |
+    |           ___|
+    |           |
+    |       ____|
+    |      |
+    |      |
+    |_____|_______________
+    0.0
+
+    Another way of saying this is that we assume the probability distribution
+    function (pdf) of gate fidelities is a set of equally weighted delta
+    functions at each value in the list. Then, the "integrated histogram"
+    is the cumulative distribution function (cdf) for this pdf.
+
+    Args:
+        data: Data to histogram. If the data is a mapping, we histogram the
+            values. All nans will be removed.
+        ax: The axis to plot on. If None, we generate one.
+        cdf_on_x: If True, flip the axes compared the above example.
+        axis_label: Label for x axis.
+        semilog: If True, force the x-axis to be logarithmic.
+        median_line: If True, draw a vertical line on the median value.
+        median_label: If drawing median line, optional label for it.
+        mean_line: If True, draw a vertical line on the mean value.
+        mean_label: If drawing mean line, optional label for it.
+        **plot_options: Kwargs to forward to `ax.step()`. Some examples are
+            color: Color of the line.
+            linestyle: Linestyle to use for the plot.
+            lw: linewidth for integrated histogram
+            ms: marker size for a histogram trace
+            label: An optional label which can be used in a legend.
+
+
+    Returns:
+        The axis that was plotted on.
+    """
+    show_plot = not ax
+    if ax is None:
+        fig, ax = plt.subplots(1, 1)
+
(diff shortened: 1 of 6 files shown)


cirq/vis/histogram.py
```diff
@@ -0,0 +1,150 @@
...
+    function (pdf) of gate fidelities is a set of equally weighted delta
+    functions at each value in the list. Then, the "integrated histogram"
+    is the cumulative distribution function (cdf) for this pdf.
+
+    Args:
+        data: Data to histogram. If the data is a mapping, we histogram the
+            values. All nans will be removed.
+        ax: The axis to plot on. If None, we generate one.
+        cdf_on_x: If True, flip the axes compared the above example.
+        axis_label: Label for x axis.
```
balopat, line 62: Is this the label for the y axis when `cdf_on_x=True`?

cirq/vis/histogram.py
```diff
@@ -0,0 +1,150 @@
...
+        semilog: If True, force the x-axis to be logarithmic.
+        median_line: If True, draw a vertical line on the median value.
+        median_label: If drawing median line, optional label for it.
+        mean_line: If True, draw a vertical line on the mean value.
+        mean_label: If drawing mean line, optional label for it.
+        **plot_options: Kwargs to forward to `ax.step()`. Some examples are
+            color: Color of the line.
+            linestyle: Linestyle to use for the plot.
+            lw: linewidth for integrated histogram
+            ms: marker size for a histogram trace
```
balopat, line 72: what does this mean?

cirq/vis/histogram.py
```diff
@@ -0,0 +1,150 @@
...
+        axis_label: Label for x axis.
+        semilog: If True, force the x-axis to be logarithmic.
+        median_line: If True, draw a vertical line on the median value.
+        median_label: If drawing median line, optional label for it.
+        mean_line: If True, draw a vertical line on the mean value.
+        mean_label: If drawing mean line, optional label for it.
+        **plot_options: Kwargs to forward to `ax.step()`. Some examples are
+            color: Color of the line.
+            linestyle: Linestyle to use for the plot.
+            lw: linewidth for integrated histogram
```
balopat, line 71: line ending periods are inconsistent

cirq/vis/histogram.py
```diff
@@ -0,0 +1,150 @@
...
+            ms: marker size for a histogram trace
+            label: An optional label which can be used in a legend.
+
```
balopat, line 74: hide_zero, title are missing

cirq/vis/histogram.py
```diff
@@ -0,0 +1,150 @@
...
+    |_____|_______________
+    0.0
+
+    Another way of saying this is that we assume the probability distribution
+    function (pdf) of gate fidelities is a set of equally weighted delta
+    functions at each value in the list. Then, the "integrated histogram"
+    is the cumulative distribution function (cdf) for this pdf.
+
+    Args:
+        data: Data to histogram. If the data is a mapping, we histogram the
```
balopat, line 58:
```suggestion
        data: Data to histogram. If the data is a `Mapping`, we histogram the
```

cirq/vis/histogram.py
```diff
@@ -0,0 +1,150 @@
...
+
+    if isinstance(data, Mapping):
+        data = list(data.values())
+
+    data = [d for d in data if not np.isnan(d)]
+    n = len(data)
+
+    if not hide_zero:
+        bin_values = np.linspace(0, 1, n + 1)
+        parameter_values = sorted(np.concatenate(([0], data)))
```
balopat, line 91: Can you help me understand this? It looks a bit fishy to me - we are modifying the data to show where zero is? Why not just modify the x axis min value?

cirq/vis/histogram.py
```diff
@@ -0,0 +1,150 @@
...
+        "where": 'post',
+        "color": 'b',
+        "linestyle": '-',
+        "lw": 1.0,
+        "ms": 0.0,
+    }
+    plot_options.update(kwargs)
+    if cdf_on_x:
+        ax.step(bin_values, parameter_values, **plot_options)
+        setsemilog = ax.semilogy
```
balopat, line 105: 1. can you add underscores to these variable names? 
2. also, I think inline (e.g. `set_semilog = ax.semilogy if cdf_on_x else ax.semilogx`) would be a bit less error prone in this case - instead of maintaining two lists we'll have a single one.

balopat (approved):
Thanks!!! Looks pretty good % nits to follow your notation :)

cirq/vis/histogram.py
tanujkhattar: Eg: 
- `cirq.integrated_histogram(range(10), marker='o', ms=10.0)`

image

Marker size corresponds to the size of "o"s.

cirq/vis/histogram.py
tanujkhattar: A step plot starts by plotting a horizontal line between the first (x0, y0) and (x1, y0) before going up from (x1, y0) to (x1, y1). Because of this, the first horizontal line would co-inside with x-axis (y values are generated by `np.linespace` so first `y` value would always be 0). 

If `hide_zero` is false, we move the plot up by 1 unit. For example, the left plot below is generated using `hide_zero = True` and right plot is generated using `hide_zero = False`. 
- Left: `cirq.integrated_histogram(range(1, 10), marker='o', ms=10.0, hide_zero=True)`
- Right: `cirq.integrated_histogram(range(1, 10), marker='o', ms=10.0, hide_zero=False)`
image

cirq/vis/histogram.py
tanujkhattar: yes. Updated the comment.

cirq/vis/histogram.py
tanujkhattar: fixed.

cirq/vis/histogram.py
tanujkhattar: added. Also modified `hide_zero` to `show_zero` to avoid double negation. (if not `hide_zero`: do something special)

cirq/vis/histogram.py
tanujkhattar: done.

cirq/vis/histogram.py
tanujkhattar: done.

PR #3943, opened 2021-03-20
Move MPSSimulator to use act_on

tonybruguier:
This is a follow-up for #3841


daxfohl:
Would you be able to make `MPSState` inherit from `ActOnArgs` and use that to perform measurements?

tonybruguier:
Had an offline discussion with Dax and it might be premature to do this modification, so I'll withdraw this for now.

PR #3947, opened 2021-03-21
support pytest.skip() in subprocess tests in _compat_test.py

balopat:
Adds support for pytest.skip() in subprocess tests in _compat_test.py.

Also adding a conditional skip for the g3 env, where importlib_metadata doesn't exist. 

Note: while using strings might seem wonky, pickling exceptions to get over the subprocess boundary is a bit more complicated - especially for pytest's internal types over which we don't have too much control.

### DIFF
--- a/cirq/_compat_test.py
+++ b/cirq/_compat_test.py
@@ -17,6 +17,7 @@ import multiprocessing
 import sys
 import traceback
 import types
+import warnings
 from types import ModuleType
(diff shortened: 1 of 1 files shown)


viathor (approved):
Re not pickling: Makes sense. Also, pickle is inherently [unsafe](https://docs.python.org/3/library/pickle.html).

PR #3923, opened 2021-03-17
(De-)serialization of CircuitOperations

95-martin-orion:
Follow up for #3891. Part of #3634.

This PR adds the `CircuitOp(De)serializer` classes and updates existing behavior to handle them appropriately, including:
- Addition of `Op(De)serializer` superclasses for operation (de-)serializers
- Backwards-compatible conversion of `GateOp(De)serializer` fields to non-public, with public properties
- Updates and deprecations for names made awkward by this change (e.g. `gate_type` -> `internal_type`)

Changes **not** included in this PR:
- Add `CircuitOperation` (de-)serializers to the default QCS gateset

PR #3949, opened 2021-03-21
Fix name collision in program proto

balopat:
Fix name collision in program proto around `Entry`-s (causes a name collision in the discovery API generation in g3).


viathor (approved):

cirq/google/api/v2/program.proto
```diff
@@ -276,25 +276,25 @@ message MeasurementKey {
 // swap measurement keys a and b.
 message MeasurementKeyMapping {
   // Indicates that measurement key "key" should be replaced with "value".
-  message Entry {
+  message MeasurementKeyEntry {
```
95-martin-orion, line 279: For consistency, the other two should be `QubitEntry` and `ArgEntry`.

95-martin-orion (changes requested):
TODO(95-martin-orion): Update #3923 with these changes.

cirq/google/api/v2/program.proto
balopat: Oh good point with the consistency - I would call this instead `MeasurementKeyMappingEntry` then...? Too long?

maffoo: +1 for `QubitEntry`, `ArgEntry`, `MeasurementKeyEntry`.

maffoo:
Too bad that discovery api generation doesn't respect nested messages. This would seem to me to be a common problem; is there any way to change the discovery API generation to deal with this?

95-martin-orion:
> Too bad that discovery api generation doesn't respect nested messages. This would seem to me to be a common problem; is there any way to change the discovery API generation to deal with this?

IIRC, there's a similar restriction on enums in protos: even when nested, they can't have the same name. I suspect this has more to do with the fundamental behavior of protos than our API generation tools.

cirq/google/api/v2/program.proto
95-martin-orion: Echoing @maffoo: these will always appear as `*Mapping.<name>`. We only need enough prefix to dedupe these types, so I also prefer `*Entry` instead of `*MappingEntry`.

balopat: gotcha, fixing!

95-martin-orion (approved):

PR #3950, opened 2021-03-21
Allow deprecation for freezegun

balopat:
Allows usage of deprecated functionality for test functions leveraging freezegun.

During cirq.google migration this came up - as we don't have yet any deprecated attributes it didn't bite us. 

Note: all other alternatives are significantly more complex, for example
- we could make this more precise and introduce an env var specifically for the attributes
- we could create a wrapper that removes all deprecated attributes before running these tests


cirq/google/engine/engine_processor_test.py
```diff
@@ -554,6 +554,25 @@ def test_get_schedule_filter_by_time_slot(list_time_slots):
...
+    # used elsewhere, it is specific to freezegun functionality.
+    def wrapper(*args, **kwargs):
+        import os
+        from cirq.testing.deprecation import ALLOW_DEPRECATION_IN_TEST
+
+        os.environ[ALLOW_DEPRECATION_IN_TEST] = 'True'
+        try:
+            return func(*args, **kwargs)
+        finally:
+            del os.environ[ALLOW_DEPRECATION_IN_TEST]
```
95-martin-orion, line 570: If someone runs `check/pytest` with `ALLOW_DEPRECATION_IN_TEST` enabled in their environment, wouldn't this disable it for all subsequent tests? I think we need to capture its value prior to this test and reinstate the same value afterwards.

balopat: It would, but it is not meant to be enabled for all tests, across the board, intentionally. 
We have the `@cirq.testing.assert_deprecated` decorator that allows for executing deprecated code. I do not want to make it easy to escape from running deprecated code in Cirq itself.

95-martin-orion: > It would, but it is not meant to be enabled for all tests, across the board, intentionally.

This won't stop devs from using it this way anyways 😛 . I found it useful to enable `ALLOW_DEPRECATION_IN_TEST` for the proto-serialization changes to separate deprecation failures from behavioral failures; I suspect other devs will use it this way as well.

Even if we don't want to treat this as a valid use case, I think it's worthwhile to guard against "spooky errors at a distance" like this. Unit tests should **never** fail due to the order in which they are run.

balopat: > This won't stop devs from using it this way anyways 😛 .

Those cheeky developers.

> I found it useful to enable `ALLOW_DEPRECATION_IN_TEST` for the proto-serialization changes to separate deprecation failures from behavioral failures; I suspect other devs will use it this way as well.

Fair - being able to tackle piecemeal what is yelling at you sounds like a good feature.

balopat: > Unit tests should **never** fail due to the order in which they are run.

That is a good point too. I'll revert to the previous state of it.

95-martin-orion (approved):

Issue #3041, opened 2020-05-27
study/visualize.py sould be moved to vis and updated to use plot style


dabacon:
Should also adhere to plotting.md so that we can remove the warning from the test.

balopat:
Related #3257.

tanujkhattar:
This will be done as part of #3642

PR #3953, opened 2021-03-22
Refactor state histograms

tanujkhattar:
Solves #3041


cirq/study/visualize.py
```diff
@@ -14,14 +14,21 @@
...
+    import matplotlib.pyplot as plt
 
 
-def plot_state_histogram(result: 'result.Result') -> np.ndarray:
+@deprecated(
+    deadline="v0.12",
+    fix="use cirq.vis.plot_state_histogram or cirq.vis.get_state_histogram instead",
+    name="cirq.study.visualize.plot_state_histogram",
+)
+def plot_state_histogram(result: 'result.Result', ax: Optional['plt.Axis'] = None) -> np.ndarray:
```
95-martin-orion, line 31: No need to update the args for a deprecated method - you can initialize `ax` (with this type) in the function body. If someone wants to specify `ax`, they can use the new API.

cirq/vis/state_histogram.py
```diff
@@ -0,0 +1,72 @@
...
+    return values
+
+
+def plot_state_histogram(
+    values: Union['result.Result', np.ndarray], ax: Optional['plt.Axis'] = None
+) -> 'plt.Axis':
+    """Plot the state histogram from a single result with repetitions.
+
+    States is a bitstring representation of all the qubit states in a single
+    result.
```
95-martin-orion, line 49: This comment should be updated to reflect the new accepted types for `values`.

cirq/vis/state_histogram.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Args:
+        values: The histogram values to plot. If `result.Result` is passed, the
+                values are computed by calling `get_state_histogram`.
```
95-martin-orion, line 53: Need a comment for `ax`.

cirq/vis/state_histogram_test.py
```diff
@@ -0,0 +1,95 @@
...
+        cirq.measure(*qubits),  # One multi-qubit measurement
+    )
+    r = cirq.sample(c, repetitions=5)
+    expected_values = [0, 0, 0, 0, 0, 0, 0, 5, 0, 0, 0, 0, 0, 0, 0, 0]
+    _, (ax1, ax2) = plt.subplots(1, 2)
+    state_histogram.plot_state_histogram(r, ax1)
+    state_histogram.plot_state_histogram(expected_values, ax2)
+    for r1, r2 in zip(ax1.get_children(), ax2.get_children()):
+        if isinstance(r1, mpl.patches.Rectangle) and isinstance(r2, mpl.patches.Rectangle):
+            assert str(r1) == str(r2)
```
95-martin-orion, line 80: I assume this is comparing the actual vs. expected plots, and we don't care about the other objects in `get_children`?

95-martin-orion (changes requested):

cirq/study/visualize.py
tanujkhattar: The reason I introduced new args is because it helps with testing. If I don't explicitly pass the ax variable, it'll plot the figure by default (call `fig.show()`) which results in a warning. This further results in an error when I try to catch only the deprecation warning in the test (as we get 2 warnings instead of 1).

cirq/vis/state_histogram_test.py
tanujkhattar: yes, that's correct. The `get_children` loop is used as a proxy for comparing two different plots (`ax1` and `ax2`)

cirq/study/visualize.py
tanujkhattar: removed the additional arg and modified the test.

cirq/vis/state_histogram.py
tanujkhattar: done.

cirq/vis/state_histogram.py
tanujkhattar: done

95-martin-orion (approved):

PR #3957, opened 2021-03-23
Modular cirq: cirq.google is extracted

balopat:
This PR extracts cirq.google into cirq_google. 

This is a huge change that I don't really see how could be broken up to smaller pieces. It will probably stick around for a couple of weeks until all the downstream system's testing is done. 

Major features:

- user code using cirq.google is going to get a DeprecationWarning, but otherwise will still work
- cirq will become a "metapackage" 
  - `pip install cirq` will install cirq, cirq-core and cirq-google 
  - `pip install cirq-google` will install cirq-core and cirq-google 
- all scripts, CI config, notebooks and release scripts have been reconfigured 

Testable versions: 
- https://test.pypi.org/project/cirq-google/0.11.0.dev20210323153007/
- https://test.pypi.org/project/cirq-core/0.11.0.dev20210323153007/
- https://test.pypi.org/project/cirq/0.11.0.dev20210323153007/
- `pip install cirq==0.11.0.dev20210323153007 --pre --index-url=https://test.pypi.org/simple --extra-index-url=https://pypi.org/simple` should install the cirq-google, cirq-core packages and the cirq metapackage

The plan: get consensus and sign-off from all major stakeholders and impacted downstream projects (cirq, qsim, quantum OS, TFQ, QCS) and then merge. I do have the changes somewhat scripted that makes it easier to keep the branch up to date with master - there is probably not going to be a need for a code freeze as such.

There are several commits that allow for easier review:

- in-place changes
- new files 
- cirq.google -> cirq_google rename 
- final move of the modules 

Note that this includes currently https://github.com/quantumlib/Cirq/pull/3948 which adds an extra commit at the beginning.


dev_tools/pypath
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#!/usr/bin/bash
+
+export PYTHONPATH="$(pwd)"
```
95-martin-orion, line 17: This would overwrite the user's existing PYTHONPATH, which could have unwanted side effects. I'd recommend either appending to it (as with the lines below), or a more involved "add if not present" behavior to avoid flooding the PYTHONPATH.

setup.py
```diff
@@ -43,40 +43,27 @@
...
-dev_requirements = open('dev_tools/conf/pip-list-dev-tools.txt').readlines()
-dev_requirements = [r.strip() for r in dev_requirements]
-
-cirq_packages = ['cirq'] + ['cirq.' + package for package in find_packages(where='cirq')]
-
 # Sanity check
 assert __version__, 'Version string cannot be empty'
 
+# This is a pure metapackage that installs all our packages
+requirements = [f'{p}=={__version__}' for p in ['cirq-core', 'cirq-google']]
```
95-martin-orion, line 50: This assumes that `cirq-core` and `cirq-google` will always have the same version. Is this a valid assumption to make?

cirq-google/setup.py
```diff
@@ -0,0 +1,76 @@
...
+# It is assumed that it ends with one of `.devN`, `.aN`, `.bN`, `.rcN` and hence
+# it will be a pre-release version on PyPi. See
+# https://packaging.python.org/guides/distributing-packages-using-setuptools/#pre-release-versioning
+# for more details.
+if 'CIRQ_PRE_RELEASE_VERSION' in os.environ:
+    __version__ = os.environ['CIRQ_PRE_RELEASE_VERSION']
+    long_description = (
+        "**This is a development version of Cirq-google and may be "
+        "unstable.**\n\n**For the latest stable release of Cirq "
+        "see**\n`here <https://pypi.org/project/cirq>`__.\n\n" + long_description
```
95-martin-orion, line 41: Should we link this to the `cirq-google` package instead?

cirq-google/requirements.txt
```diff
@@ -0,0 +1,2 @@
+google-api-core[grpc] >= 1.14.0, < 2.0.0dev
```
95-martin-orion, line 1: Should `cirq-core` be part of these requirements?

cirq-google/setup.py
```diff
@@ -0,0 +1,76 @@
...
+# If CIRQ_PRE_RELEASE_VERSION is set then we update the version to this value.
+# It is assumed that it ends with one of `.devN`, `.aN`, `.bN`, `.rcN` and hence
+# it will be a pre-release version on PyPi. See
+# https://packaging.python.org/guides/distributing-packages-using-setuptools/#pre-release-versioning
+# for more details.
+if 'CIRQ_PRE_RELEASE_VERSION' in os.environ:
+    __version__ = os.environ['CIRQ_PRE_RELEASE_VERSION']
+    long_description = (
+        "**This is a development version of Cirq-google and may be "
+        "unstable.**\n\n**For the latest stable release of Cirq "
```
95-martin-orion, line 40: `Cirq` -> `Cirq-google`?

cirq-core/setup.py
```diff
@@ -0,0 +1,78 @@
...
+# It is assumed that it ends with one of `.devN`, `.aN`, `.bN`, `.rcN` and hence
+# it will be a pre-release version on PyPi. See
+# https://packaging.python.org/guides/distributing-packages-using-setuptools/#pre-release-versioning
+# for more details.
+if 'CIRQ_PRE_RELEASE_VERSION' in os.environ:
+    __version__ = os.environ['CIRQ_PRE_RELEASE_VERSION']
+    long_description = (
+        "**This is a development version of Cirq and may be "
+        "unstable.**\n\n**For the latest stable release of Cirq "
+        "see**\n`here <https://pypi.org/project/cirq>`__.\n\n" + long_description
```
95-martin-orion, line 43: Should we link this to the `cirq-core` package instead?

95-martin-orion (changes requested):
Bundle of setup.py nits.

setup.py
balopat: This is stronger than an assumption - it's by design. This is how we "simulate" the monorepo structure. Unless we have good reasons to do otherwise, all modules are going to be released simultaneously and cirq vx.y.z will always pull in cirq-* x.y.z. 

This expectation is captured here and if the design changes for any of our modules, we can change the test as well.

cirq-google/setup.py
balopat: we can!

cirq-google/requirements.txt
balopat: I don't think so, requirements.txt is for development. When you're developing cirq.google, you will have cirq-core checked out as well, so you should depend on that. 

However https://github.com/quantumlib/Cirq/pull/3957/commits/57687f989535e1c82a3258e0df0a1232bdecb112#diff-75c481d943df4a45c7599762865505aa8860d35baf87e3629911224dd37fa620R50 should capture the dependency of cirq-google on cirq-core and it doesn't, so I'll add it there!

balopat: Oh ignore that - I was looking at the wrong file. 
We do capture the dependency in https://github.com/quantumlib/Cirq/commit/57687f989535e1c82a3258e0df0a1232bdecb112#diff-be82d3fb5dd590eca7a9d68afc211891571cfcb4816e05cc96355895f0b2b34aR52

dev_tools/pypath
balopat: Done!

check/pytest
```diff
@@ -19,23 +19,23 @@ cd "$(git rev-parse --show-toplevel)"
 
 PYTEST_ARGS=()
 ACTUALLY_QUIET=""
+CIRQ_ONLY=""
```
dstrain115, line 22: Can we document this command line flag in the doc string at the top?

cirq-google/cirq_google/api/v1/params.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2018 The Cirq Developers
```
dstrain115, line 1: How come none of the cirq_google files register as moves?  Can we put in git hints or however it works in order for it to register as renames instead of deletes and copies?

dstrain115 (approved):

cirq-google/cirq_google/api/v1/params.py
95-martin-orion: If you view just the final commit (`extracted cirq_google.`) these all show up properly as moves.

95-martin-orion (commented):
LGTM from me. As double-approval has been requested on this PR, I'm leaving the final "Approve" flag-flip to Doug.

check/pytest
balopat: Yes, I can add that - this flag already exists though, so I'll add it in a separate PR to keep this smaller.

95-martin-orion (approved):

PR #3961, opened 2021-03-23
Migrate MPS to act_on protocol

daxfohl:
This migrates the MPS simulator to use the act_on protocol, completing this migration for all the simulators.

@tonybruguier can you verify this change doesn't affect MPS functionality? All the tests pass, but please make sure this doesn't break anything (including performance).


daxfohl:
@tonybruguier it also looks like it's requiring you to sign the cla again specifically for this PR since I grabbed the initial commit from your fork, or something. Can you try to follow the link in the google-cla bot message above and see if it fixes the error?

tonybruguier:
@googlebot I consent.

tonybruguier:
As long as the unit tests pass, I think it should be fine in term of functionality.

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -296,10 +302,12 @@ def __init__(
 
         Args:
             qubit_map: A map from Qid to an integer that uniquely identifies it.
+            prng: A random number generator, used to simulate measurements.
```
95-martin-orion, line 305: Document `axes` and `log_of_measurement_results` here as well.

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -415,7 +430,7 @@ def to_numpy(self) -> np.ndarray:
         """An alias for the state vector."""
         return self.state_vector()
 
-    def apply_op(self, op: 'cirq.Operation', prng: np.random.RandomState):
```
95-martin-orion, line 418: This changes the API for this class - is there a use case for his method outside of `act_on`? If so, `apply_op` should be kept and `_act_on_fallback_` converted to:
```
def _act_on_fallback_(self, op: Any, allow_decompose: bool):
    return self.apply_op(op, self.prng)
```
Otherwise, since this is under `contrib` deprecation isn't necessary.

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -415,7 +430,7 @@ def to_numpy(self) -> np.ndarray:
...
+    def _act_on_fallback_(self, op: Any, allow_decompose: bool):
```
95-martin-orion, line 433: Return type annotation?

cirq/contrib/quimb/mps_simulator.py
```diff
@@ -282,12 +285,15 @@ def sample(
...
-class MPSState:
+class MPSState(ActOnArgs):
     """A state of the MPS simulation."""
 
     def __init__(
         self,
         qubit_map: Dict['cirq.Qid', int],
+        axes: Iterable[int],
+        prng: np.random.RandomState,
+        log_of_measurement_results: Dict[str, Any],
```
95-martin-orion, line 296: These can probably be made optional (with the possible exception of `prng`), with default values assigned in the constructor body.

cirq/contrib/quimb/mps_simulator_test.py
```diff
@@ -324,7 +340,12 @@ def test_supremacy_equal_more_cols():
 def test_tensor_index_names():
     qubits = cirq.LineQubit.range(12)
     qubit_map = {qubit: i for i, qubit in enumerate(qubits)}
-    state = ccq.mps_simulator.MPSState(qubit_map)
+    state = ccq.mps_simulator.MPSState(
+        qubit_map,
+        prng=value.parse_random_state(0),
+        axes=[],
+        log_of_measurement_results={},
```
95-martin-orion, line 347: Please add a test confirming that these are passed through to the base ActOnArgs object. The behavior of these fields should already be tested by that class, so this new test doesn't need to be complex.

95-martin-orion (changes requested):

cirq/sim/act_on_args.py
```diff
@@ -26,9 +26,9 @@ class ActOnArgs:
 
     def __init__(
         self,
-        axes: Iterable[int],
         prng: np.random.RandomState,
-        log_of_measurement_results: Dict[str, Any],
+        axes: Iterable[int] = None,
+        log_of_measurement_results: Dict[str, Any] = None,
```
95-martin-orion, line 31: [No change needed] IIUC, this class was added after the v0.10 release so changing its argument ordering like this is safe.

95-martin-orion (approved):
Looks good!

PR #3962, opened 2021-03-23
Type annotations for `optimization_pass`

vtomole:
Part of: https://github.com/quantumlib/Cirq/issues/554


cirq/circuits/optimization_pass_test.py
```diff
@@ -149,12 +155,17 @@ def test_point_optimizer_raises_on_gates_changing_qubits():
...
         """Changes all single qubit operations to act on LineQubit(42)"""
 
-        def optimization_at(self, circuit, index, op):
+        def optimization_at(
+            self, circuit: Circuit, index: int, op: 'cirq.Operation'
+        ) -> Optional[PointOptimizationSummary]:
             if len(op.qubits) == 1:
+                op = cast(cirq.GateOperation, op)
                 new_op = op.gate(cirq.LineQubit(42))
+                print(type(new_op))
```
95-martin-orion, line 164: Remove debug print.

cirq/circuits/optimization_pass_test.py
```diff
@@ -149,12 +155,17 @@ def test_point_optimizer_raises_on_gates_changing_qubits():
     class EverythingIs42(cirq.PointOptimizer):
         """Changes all single qubit operations to act on LineQubit(42)"""
 
-        def optimization_at(self, circuit, index, op):
+        def optimization_at(
+            self, circuit: Circuit, index: int, op: 'cirq.Operation'
+        ) -> Optional[PointOptimizationSummary]:
             if len(op.qubits) == 1:
+                op = cast(cirq.GateOperation, op)
```
95-martin-orion, line 162: Is this `cast` required? Maybe add `isinstance(op, cirq.GateOperation)` to the if condition instead, since this test doesn't appear to have behavior for non-gate operations.

95-martin-orion (changes requested):
Top-level comment: don't add cirq imports for type annotations. Instead, use `'cirq.{TYPE}'`. Other imports (like `from typing import Optional`) are fine.

Other than that, looks good to me.

cirq/circuits/optimization_pass_test.py
vtomole: Done.

cirq/circuits/optimization_pass_test.py
vtomole: Done.

vtomole:
@95-martin-orion  PTAL

95-martin-orion (approved):

PR #3968, opened 2021-03-25
Add support for Tuple[str] as measurement key

smitsanghavi:
Tuples will be used to represent [nested CircuitOperations](https://docs.google.com/document/d/1_dY92lTKumaxH4wRdyabqQz_-i_xupNVwMzbrz__1U8/edit?resourcekey=0-bcAZJNPDvnmthf5Y8VFVZQ#heading=h.s3u7e3nnef7n)


95-martin-orion (changes requested):
There's something bothering me about this PR, but it's difficult to pin down. I have a sense that it would be cleaner to replace all internal representations of measurement keys with `Tuple[str]` and only convert to/from strings in protocols and `MeasurementGate` construction, but I can't say how that would work (especially in a backwards-compatible way).

I think my question is this: after all of the deprecations are complete, what is the intended final state of this feature? In that state, where will string keys appear (if at all)?

smitsanghavi:
Converting this PR to draft since the design is still under discussion.

cirq/ops/measurement_gate.py
```diff
@@ -64,7 +64,10 @@ def __init__(
         self._qid_shape = qid_shape
         if len(self._qid_shape) != num_qubits:
             raise ValueError('len(qid_shape) != num_qubits')
-        self.key = key
+        if isinstance(key, str):
+            self.key = key
+        else:
+            raise TypeError('Wrong type for measurement key. MeasurementGate only accepts str key')
```
daxfohl, line 70: Pretty sure we'd need to allow tuples here for when you flatten a circuit that has tuples produced by subcircuits. The equivalent flattened circuit would have to measure directly into the tuple.

95-martin-orion:
Closing as discontinued. The updated design for this can be found in #4040 and the [linked RFC](https://tinyurl.com/structured-measurement-keys).

PR #3969, opened 2021-03-25
[obs] 4.3 - Stopping criteria

mpharrigan:
Extend the observable measurement "main sampling loop" to use `StoppingCriteria` instances to provide different ways of stopping. Two are built in: Variance and Number of repetitions. Users can write their own by implementing the abstract interface.

For `measure_grouped_settings`, the caller must provide a `StoppingCriteria`. Forthcoming, higher-level wrappers will provide a more user-friendly way of taking the more traditional path of `RepetitionsStoppingCriteria`. 

cc @balopat @95-martin-orion 

Part of #3647
Follows #3855


cirq/work/observable_measurement_test.py
```diff
@@ -213,31 +260,73 @@ def test_meas_specs_still_todo():
...
+        _, _ = _check_meas_specs_still_todo(
+            meas_specs=[meas_spec],
+            accumulators={meas_spec: bsa},
+            stopping_criteria=bad_stop,
+        )
+
+
+def test_meas_spec_still_todo_lots_of_params():
+    # This test is fairly slow (1.5s) because we need at least 300_001 measurement specs
+    # to trigger the exception being tested.
```
95-martin-orion, line 318: You should be able to mock out MAX_REPETITIONS_PER_JOB to speed this up ([SO example](https://stackoverflow.com/questions/38336090/mocking-a-global-variable)):
```
from unittest import mock

mock_max_reps = mock.patch.object(cw, 'MAX_REPETITIONS_PER_JOB', 1000)
with mock_max_reps:
    # ...run test...
```

cirq/work/observable_measurement.py
```diff
@@ -153,7 +211,7 @@ def _pad_setting(
 
 
 def _aggregate_n_repetitions(next_chunk_repetitions: Set[int]) -> int:
-    """In the future, we will allow each accumulator to request a different number
+    """Each stopping criteria can request a different number
```
95-martin-orion, line 214: IIUC, the rep counts passed to this function are all from the same stopping criteria, but for different measurement specs. This comment should reflect that.

cirq/work/observable_measurement.py
```diff
@@ -191,21 +238,44 @@ def _check_meas_specs_still_todo(
...
+        repetitions = MAX_REPETITIONS_PER_JOB // len(still_todo)
+
+        if repetitions < 10:
+            raise ValueError(
+                "You have requested too many parameter settings to batch your job effectively. "
+                "Consider fewer sweeps or manually splitting sweeps into multiple jobs."
+            )
+
+        # coverage: ignore
+        warnings.warn(
```
95-martin-orion, line 272: Recommend adding coverage for this warning with `pytest.warns`

cirq/work/observable_measurement.py
```diff
@@ -62,6 +66,60 @@ def _with_parameterized_layers(
...
+    def more_repetitions(self, accumulator: BitstringAccumulator) -> int:
+        if len(accumulator.bitstrings) == 0:
+            return self.repetitions_per_chunk
+
+        cov = accumulator.covariance()
+        n_terms = cov.shape[0]
+        ones = np.ones(n_terms)
+        sum_variance = ones @ cov @ ones.T
+        var_of_the_e = sum_variance / len(accumulator.bitstrings)
+        vpt = var_of_the_e / n_terms
```
95-martin-orion, line 98: To make sure I'm reading this properly:
1. get the covariance matrix as provided by the accumulator
2. sum over all elements of the matrix
3. compute the average from the sum

Is that correct? If so, I think 2-3 could be replaced with [`np.average`](https://numpy.org/doc/stable/reference/generated/numpy.average.html).

cirq/work/observable_measurement.py
```diff
@@ -31,6 +33,8 @@
     import cirq
     from cirq.value.product_state import _NamedOneQubitState
 
+MAX_REPETITIONS_PER_JOB = 3_000_000
```
95-martin-orion, line 36: Where does this value come from? Might be useful to add a comment about this.

cirq/work/observable_measurement_test.py
```diff
@@ -198,7 +199,53 @@ def test_aggregate_n_repetitions():
     assert reps == 6
```
95-martin-orion, line 199: Could you replace this test with an example using `_check_meas_specs_still_todo`? This would help clarify the purpose of `_aggregate_n_repetitions`.

95-martin-orion (changes requested):
Behavior generally looks good - comments are mostly on docs and testing.

cirq/work/observable_measurement_test.py
mpharrigan: cool; I used pytest's monkeypatch fixture which is short and cleans itself up afterwards

cirq/work/observable_measurement.py
mpharrigan: good point.

cirq/work/observable_measurement.py
mpharrigan: added, now that I know the monkeypatching trick for max_repetitoins_per_job :)

cirq/work/observable_measurement.py
mpharrigan: documented using the `document` thing

cirq/work/observable_measurement_test.py
mpharrigan: It's impossible with the current two StoppingCriteria because they both either are done or do `repetitions_per_chunk`. If there was a less-well-behaved stopping criteria it could trigger it. Maybe I'll throw one together

cirq/work/observable_measurement.py
mpharrigan: yes, I think you're correct. I took this from @ncrubin

mpharrigan: there's the extra division by len(accumulator.bitstrings) which happens "in between" the sum and the division by terms, which I'd like to keep separate for clarity of intent

cirq/work/observable_measurement_test.py
mpharrigan: did a somewhat contrived test :)

mpharrigan:
@95-martin-orion ptal!

95-martin-orion (approved):
New coverage all looks good, LGTM!

Issue #3958, opened 2021-03-23
Density matrix simulator doesn't accept array with shape=(2,)*2**len(qubits) as an initial state


Strilanc:
See https://quantumcomputing.stackexchange.com/questions/16636/

95-martin-orion:
Seems pretty self-explanatory: `cirq.qis.to_valid_density_matrix` assumes inputs are either rank-2 (and therefore density matrices) or rank-1 (and therefore state vectors, which must be converted to density matrices).
https://github.com/quantumlib/Cirq/blob/c27dcffcab6a62540bd939fb05055fabd474a1a1/cirq/qis/states.py#L788-L795

Since either `num_qubits` or `qid_shape` is required, it should always be possible to tell whether `density_matrix_rep` is a state vector or density matrix, but we'll need to break the existing assumptions about input tensor rank.

daxfohl:
There's an `is_raw_state` input to `_base_iterator`, that maybe we could expose. This also may fall into the bucket of #3872, that would allow the simulation input to be a proper `act_on_args` rather than just a vector/int.

maffoo:
Also doesn't allow specifying initial state as a state vector, e.g. a vector of shape `(2**len(qubits),)`:
```python
>>> sim = cirq.DensityMatrixSimulator()
>>> circuit = cirq.Circuit(cirq.measure(cirq.LineQubit(0)))
>>> sim.simulate(circuit, initial_state=np.array([1, 0]))
...
ValueError: cannot reshape array of size 2 into shape (2,2)
```

<details>

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [20], in <cell line: 1>()
----> 1 sim.simulate(circuit, initial_state=np.array([1, 0]))

File ~/.virtualenvs/pyle/lib/python3.8/site-packages/cirq/sim/simulator.py:518, in SimulatesFinalState.simulate(self, program, param_resolver, qubit_order, initial_state)
    493 def simulate(
    494     self,
    495     program: 'cirq.AbstractCircuit',
   (...)
    498     initial_state: Any = None,
    499 ) -> TSimulationTrialResult:
    500     """Simulates the supplied Circuit.
    501 
    502     This method returns a result which allows access to the entire
   (...)
    516         SimulationTrialResults for the simulation. Includes the final state.
    517     """
--> 518     return self.simulate_sweep(
    519         program, study.ParamResolver(param_resolver), qubit_order, initial_state
[...]

PR #3970, opened 2021-03-25
Split the simulators' creation of ActOnArgs and iteration

### LINKED ISSUE #3872: [RFC] Split _base_iterator into two functions
**Is your design idea/issue related to a use case or problem? Please describe.**

Soon all the built-in simulators will have migrated to the _act_on_ pattern. At that point, all the simulators will have a `_base_iterator` method that creates an `ActOnArgs` and then iterates the circuit. I'm proposing that we split this into two separate abstract methods: one to create the `ActOnArgs`, and another that iterates the circuit with that ActOnArgs object. Splitting it this way enables adherence to the single responsibility pattern, and brings a number of possible use cases with it.

**Describe your design idea/issue**

Draft RFC: [tinyurl.com/split-actonargs](http://tinyurl.com/split-actonargs)

daxfohl:
Fixes #3872.

Currently all simulators implement `_base_iterator`, which does two things: it creates the ActOnArgs object, and the iterates the circuit. This PR splits ActOnArgs creation and iteration into two separate functions in each of the simulators. This allows preservation of single-responsibility principle in the simulator methods. It potentially allows moving `_core_iterator` into the base class to remove that redundancy in a future PR (all four simulators' implementations of `_core_iterator` are almost identical now).

It also addresses the problem in #3958 with a workaround: the user can now propagate the ActOnArgs to subsequent circuits. See the quantum teleportation example for an example of how to do this. Note that this process works for all simulators, and a function can be written generically to support any of them, so long as the same simulator runs each circuit.

The quantum teleportation is refactored to be implemented in terms of this. This fixes the "hack" of the teleportation in a different way than the classical control tags. Here, the circuit is split into two, and the Bob side is constructed and run after the Alice side based on Alice's measurements. Unlike the tags formulation, this cannot be serialized into a single circuit and requires the Python code to manage the classical coordination. Again, other than the final state vector printout, this example can work generically with any simulator.

Open question: should these functions be public?


daxfohl:
@tonybruguier can you consent here too? Seems like I must have merged wonkily on this one.

tonybruguier:
@googlebot I consent.

cirq/sim/simulator.py
```diff
@@ -526,7 +525,52 @@ def _base_iterator(
...
             StepResults from simulating a Moment of the Circuit.
         """
-        raise NotImplementedError()
+        acton_args = self._create_act_on_args(circuit, initial_state, qubit_order)
+        return self._core_iterator(circuit, acton_args, qubit_order)
+
+    @abc.abstractmethod
+    def _create_act_on_args(
+        self,
+        circuit: circuits.Circuit,
```
daxfohl, line 534: Seems odd to need the circuit here. Can we just use qubit_order?

cirq/sim/simulator.py
```diff
@@ -526,7 +525,52 @@ def _base_iterator(
...
+        Returns:
+            ActOnArgs for the simulator.
+        """
+
+    @abc.abstractmethod
+    def _core_iterator(
+        self,
+        circuit: circuits.Circuit,
+        initial_state: TActOnArgs,
+        qubit_order: ops.QubitOrderOrList = ops.QubitOrder.DEFAULT,
```
daxfohl, line 558: Seems odd to require the qubit_order here if we've already passed it in above. Can this be stored somewhere?

daxfohl: Plausibly if we stuff it into the ActOnArgs base class as a qubit_map, that would have the additional benefit that we'd no longer have to require simulators to manually set ActOnArgs.axes for each operation: the act_on protocol could look the axes up from the qubit_map and the operation qubits.

daxfohl: Gave this a try but it ends up being a big change. Mostly the unit tests since all the ActOnArgs creations end up needing a new parameter. Maybe a separate PR would be better.

95-martin-orion (changes requested):
If I understand this right, `ActOnArgs` is essentially a persistent state which can be used across simulations that use the same simulator. That's potentially super useful, but only if it doesn't introduce additional time costs (state-copying being a potential culprit).

On a different note, this PR opens up a whole bunch of design questions.
- Should `simulate` also use `ActOnArgs`? It seems applicable (see the teleportation example).
- What would `ActOnArgs` look like for e.g. qsim, where the state is a C++ object?
- Is there a more descriptive name for `ActOnArgs` to express how it persists state?

I think we should get solid answers on these before committing to this change.

cirq/sim/simulator.py
```diff
@@ -503,7 +505,6 @@ def _simulator_iterator(
         """
         return self.simulate_moment_steps(circuit, param_resolver, qubit_order, initial_state)
 
-    @abc.abstractmethod
     def _base_iterator(
```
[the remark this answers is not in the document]
daxfohl: Ideally we can deprecate _base_iterator at some point. @balopat is still considering how to deprecate an abstract method, but there's a chicken-and-egg aspect to it. For now, I think best to leave as-is. I can update the docstrings to highlight the differences.

examples/quantum_teleportation.py
```diff
@@ -34,26 +34,88 @@
...
+    ]
+) -> Tuple[TSimulationTrialResult, TStepResult]:
+    # Initialize our qubit state space.
     msg, alice, bob = cirq.LineQubit.range(3)
+    qubits = (msg, alice, bob)
+    args = sim._create_act_on_args(0, qubits)
+
+    # First we create a bell state circuit and simulate it on the qubits.
+    bell_circuit = cirq.Circuit(cirq.H(alice), cirq.CNOT(alice, bob))
+    list(sim._core_iterator(bell_circuit, args, qubits))
```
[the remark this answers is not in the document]
daxfohl: Sure, I can revert this file's changes. Calling `simulate_moment_steps` is incompatible with this approach because it requires two separate circuits that share the same ActOnArgs. `simulate_moment_steps` hides the ActOnArgs from you.

I'll revert this file's changes and we can revisit if we decide to create public APIs for the two new functions.

cirq/sim/density_matrix_simulator.py
```diff
@@ -214,16 +214,15 @@ def _run_sweep_repeat(
...
         measurements = {}  # type: Dict[str, List[np.ndarray]]
 
         for _ in range(repetitions):
-            all_step_results = self._base_iterator(
+            all_step_results = self._core_iterator(
                 circuit,
                 qubit_order=qubit_order,
-                initial_state=intermediate_state,
-                is_raw_state=True,
+                sim_state=acton_args.copy(),
```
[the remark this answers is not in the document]
daxfohl: Yes, in the old _base_iterator, you can see it creates a copy.

```
    if np.may_share_memory(initial_matrix, initial_state):
        initial_matrix = initial_matrix.copy()
```

With this PR, `_run_sweep_sample` no longer calls `_base_iterator` and calls `_core_iterator` instead. If anything, this approach should be faster, as it's a straight copy instead of doing `qis.to_valid_density_matrix` and then copying.

cirq/sim/sparse_simulator.py
```diff
@@ -201,59 +202,93 @@ def _run(
...
         repetitions: int,
     ) -> Dict[str, np.ndarray]:
         """Repeatedly simulate a circuit in order to produce samples."""
 
         measurements: DefaultDict[str, List[np.ndarray]] = collections.defaultdict(list)
         for _ in range(repetitions):
-            all_step_results = self._base_iterator(
-                circuit, initial_state=initial_state, qubit_order=qubit_order
+            all_step_results = self._core_iterator(
+                circuit, sim_state=acton_args.copy(), qubits=qubits
```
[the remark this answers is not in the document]
daxfohl: I can't find the specific place, but somewhere inside the old code's `_base_iterator`, `qis._to_valid_state_vector` is making a copy. This is how _brute_force_samples is designed to work: it takes the intermediate state produced by the unitary prefix, and runs repetitions on copies of it. If state was not copied for each repetition, we'd see the repetitions affect each other.

The new code calls `_core_iterator` instead of `_base_iterator`, avoiding that call to `qis._to_valid_state_vector`, so avoids that extra copy.

daxfohl:
@95-martin-orion thanks for the review. Agreed this is a big commitment and needs care. There is an RFC attached to the linked issue #3872. I'll flesh that out and we should have a full review and sign offs there before proceeding further. I wanted to make the PR available for reference, and as a sanity check that the basic idea would work.

Does qsim expose `simulate_moment_steps` via implementing `_base_iterator`? I thought it only supported final states, or returned intermediate states in some other way than `_base_iterator`, in which case this change wouldn't affect qsim. But I haven't checked.

95-martin-orion:
> Does qsim expose `simulate_moment_steps` via implementing `_base_iterator`? I thought it only supported final states {...}

You're correct - qsim does not currently implement `SimulatesIntermediateStates`. However, it's on the TODO list, so it's worth keeping in mind while going through this PR.

cirq/sim/simulator.py
```diff
@@ -526,7 +525,51 @@ def _base_iterator(
         Yields:
             StepResults from simulating a Moment of the Circuit.
         """
-        raise NotImplementedError()
+        qubits = ops.QubitOrder.as_qubit_order(qubit_order).order_for(circuit.all_qubits())
+        acton_args = self._create_act_on_args(initial_state, qubits)
+        return self._core_iterator(circuit, acton_args, qubits)
+
+    @abc.abstractmethod
+    def _create_act_on_args(
```
daxfohl, line 533: New abstract method would break any external simulators. Have to make this just a regular function returning NotImplemented.

cirq/sim/simulator.py
```diff
@@ -526,7 +525,51 @@ def _base_iterator(
...
+            qubits: Determines the canonical ordering of the qubits. This
+                is often used in specifying the initial state, i.e. the
+                ordering of the computational basis states.
+
+        Returns:
+            ActOnArgs for the simulator.
+        """
+
+    @abc.abstractmethod
+    def _core_iterator(
```
daxfohl, line 553: Same as above

examples/quantum_teleportation.py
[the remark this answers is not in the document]
daxfohl: Updated `simulate_moment_steps` to accept an `ActOnArgs`, and used that here. Also made `create_act_on_args` public.

cirq/sim/simulator.py
```diff
@@ -512,17 +515,75 @@ def _base_iterator(
...
+            The ActOnArgs for this simulator.
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def _core_iterator(
+        self,
+        circuit: circuits.Circuit,
+        initial_state: TActOnArgs,
+        qubits: Tuple['cirq.Qid', ...],
```
daxfohl, line 572: Consider putting this and qubit_map into ActOnArgs itself, to avoid having to pass this parameter around everywhere with it.

cirq/sim/simulator.py
```diff
@@ -512,17 +515,75 @@ def _base_iterator(
...
+        Returns:
+            The ActOnArgs for this simulator.
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def _core_iterator(
+        self,
+        circuit: circuits.Circuit,
+        initial_state: TActOnArgs,
```
daxfohl, line 571: Make sure this is named consistently in the simulators.

cirq/sim/simulator.py
```diff
@@ -512,17 +515,75 @@ def _base_iterator(
...
                 this state depends on the simulation implementation. See
                 documentation of the implementing class for details.
 
+        Yields:
+            StepResults from simulating a Moment of the Circuit.
+        """
+        qubits = ops.QubitOrder.as_qubit_order(qubit_order).order_for(circuit.all_qubits())
+        act_on_args = (
+            cast(TActOnArgs, initial_state)
+            if isinstance(initial_state, ActOnArgs)
```
daxfohl, line 539: Consider putting this logic directly into `create_act_on_args`. This approach will allow all the other simulation functions to accept an ActOnArgs as well.

cirq/sim/simulator.py
daxfohl: Done but let's write a unit test to make sure it works as desired.

daxfohl: Tried this but it explodes. Try again later.

cirq/sim/simulator.py
daxfohl: Test `run` with `ActOnArgs`.

daxfohl:
@95-martin-orion I think this would be easier to pull the trigger on if I made `create_act_on_args` a private function for now and reverted the teleportation example. That would give us time to ponder whether "ActOnArgs" is the right name for this etc, before going all public with it. That would make this purely a refactoring change. Thoughts?

cirq/sim/clifford/clifford_simulator.py
```diff
@@ -65,55 +68,78 @@ def is_supported_operation(op: 'cirq.Operation') -> bool:
...
         Yields:
             CliffordStepResult from simulating a Moment of the Circuit.
         """
-        qubits = ops.QubitOrder.as_qubit_order(qubit_order).order_for(circuit.all_qubits())
 
-        qubit_map = {q: i for i, q in enumerate(qubits)}
+        def create_state():
+            state = CliffordState(sim_state.qubit_map)
+            state.ch_form = sim_state.state.copy()
+            return state
```
95-martin-orion, line 121: I'd recommend handling this with an optional `ch_form` argument to the `CliffordState` constructor, rather than mutating `state_ch_form` manually.

cirq/google/calibration/engine_simulator.py
```diff
@@ -395,14 +391,20 @@ def _run(
...
 
-    def _base_iterator(
+    def _core_iterator(
         self,
         circuit: Circuit,
-        qubit_order: QubitOrderOrList,
-        initial_state: Any,
-    ) -> Iterator[StateVectorStepResult]:
+        sim_state: Any,
+    ):
```
95-martin-orion, line 398: Include the return type here (`Iterator[StateVectorStepResult]`).

examples/quantum_teleportation.py
```diff
@@ -34,26 +34,87 @@
 """
 
 import random
+from typing import Tuple
```
95-martin-orion, line 37: As discussed: postpone changing this file to a future PR.

cirq/sim/simulator.py
```diff
@@ -512,17 +515,67 @@ def _base_iterator(
...
 
+        Yields:
+            StepResults from simulating a Moment of the Circuit.
+        """
+        qubits = ops.QubitOrder.as_qubit_order(qubit_order).order_for(circuit.all_qubits())
+        act_on_args = self.create_act_on_args(initial_state, qubits)
+        return self._core_iterator(circuit, act_on_args)
+
+    @abc.abstractmethod
+    def create_act_on_args(
```
95-martin-orion, line 541: As discussed: make this method private.

95-martin-orion (changes requested):
Having reviewed the new changes, I'm inclined to agree with making _create_act_on_args private to avoid "going public" with this behavior for now. Other than that, I'm largely happy with the structure of this PR.

daxfohl:
@95-martin-orion The requested changes have been made.

95-martin-orion:
LGTM. Since the big `cirq_google` PR (#3957) is expected to land today, let's hold this until after it goes through. The fix-up process should be lightweight since this PR only touches one `cirq.google` file.

95-martin-orion (changes requested):
Balint sent out an email to the cirq-dev group with details on how to sync past the `cirq_google` change. It shouldn't be too complicated for this PR - let me know if you hit any snags.

daxfohl:
@95-martin-orion Done

95-martin-orion (approved):

PR #3971, opened 2021-03-26
Coefficient fix

FallenApart:


FallenApart:
@googlebot I signed it!

vtomole (approved):
LGTM. Thanks. While we are at it, can we also replace `>` with `⟩` here?

FallenApart:
`>`  is correct. It is according to Dirac Bra–ket notation.

FallenApart:
@vtomole Done

PR #3972, opened 2021-03-26
Fix and add docstrings to set_state_vector()

### LINKED ISSUE #3623: [docs] SimulatorStep.set_state doesn't exist
See the question on : https://quantumcomputing.stackexchange.com/questions/15187/access-and-set-state-during-quantum-computation-simulation-in-cirq

1. This part of the docs is pointing to a set_state() method wrongly, instead it should be `set_state_vector()`. 
2. Also, we should point to the API docs when referencing methods.
3. set_state_vector should get some docsstrings - it's a public method

image

unaiic:
Fixes an error in the docs, which was pointing to a `set_state()` method wrongly; instead it should be `set_state_vector()`. This also adds docstrings to the `set_state_vector()` and fixes some typos.

Fixes #3623.


cirq/sim/sparse_simulator.py
```diff
@@ -365,6 +365,18 @@ def state_vector(self, copy: bool = True):
         return vector.copy() if copy else vector
 
     def set_state_vector(self, state: 'cirq.STATE_VECTOR_LIKE'):
+        """Set the state vector.
+
+        One can pass a valid full state to this method by passing a numpy
+        array. Or, alternatively, one can pass an integer, and then the state
+        will be set lie entirely in the computation basis state for the binary
```
95-martin-orion, line 372: "will be set lie" -> "will be set to lie"

Same in the notebook.

95-martin-orion (approved):
One nit, otherwise good to go.

cirq/sim/sparse_simulator.py
unaiic: Done.

PR #3973, opened 2021-03-26
Change the remaining format string to f-string

### LINKED ISSUE #3804: Change the remaining format strings to `f-string`s
**Description of the issue** 
[Flynt](https://github.com/ikamensh/flynt) was run [here](https://github.com/quantumlib/Cirq/pull/3786) but this required covering some formated lines. This issue is to track that effort.
```
$ pip install flynt
$ flynt .
$ ./check/pytest-and-incremental-coverage 
```
Then cover the remaining lines.

**Cirq version**
The code that is on the master branch.

unaiic:
Changes the remaining format string to `f-string` using Flynt:
```
$ pip install flynt
$ flynt .
```

Fixes #3804.

### DIFF
--- a/check/format-incremental
+++ b/check/format-incremental
@@ -110,8 +110,7 @@ if (( only_print == 1 )); then
     flynt_args+=("--dry-run")
 fi
 
-# TODO: Remove '-e cirq/ops/pauli_string.py' once the line is covered (https://github.com/quantumlib/Cirq/issues/3804)
-flynt ${format_files} "${flynt_args[@]}"  -e cirq/ops/pauli_string.py
+flynt ${format_files} "${flynt_args[@]}"
 FLYNTSTATUS=$?
 
 echo "Running the black formatter..."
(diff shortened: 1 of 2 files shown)


vtomole:
Thanks for taking this @unaiic . The problem here was that this changed exposed an uncovered line and i didn't have the time to come up with a test that covered said line. This task would need to do that.

unaiic:
@vtomole Oh, okay. So a test should be added to cover that line. Does it have to be in `pauli_string_test.py`? I have no experience with tests whatsoever.

vtomole:
@unaiic 
> Does it have to be in `pauli_string_test.py`?

Yup. You need to find a case where a `cirq.Operation` that is not a `cirq.GateOperation` is passed to https://github.com/quantumlib/Cirq/blob/4bb484ae77456a92ecefdd43efb569b69d9aa946/cirq/ops/pauli_string.py#L943

Strilanc:
I would recommend changing the exception from `TypeError` to `NotImplementedError`, which is special cased by the coverage check as does-not-need-to-be-covered.

vtomole (changes requested):
Remove this line: https://github.com/quantumlib/Cirq/blob/eccc219a04cc778856cbe4c9bd6a6fb739009c0f/check/format-incremental#L113 and the following flag that excludes this file.

unaiic:
@vtomole Okay, I already updated it.

@Strilanc Got it.

vtomole (approved):
Thank you!

PR #3979, opened 2021-03-30
Density matrix simulator EVs

### LINKED ISSUE #3964: Density matrix simulator support for SimulatesExpectationValues

The density matrix simulator does not inherit from `cirq.sim.simulator.SimulatesExpectationValues`.

I would like to be able to use this interface with the density matrix simulator:
```python
q0 = cirq.GridQubit(0, 0)
test_circuit_noise = cirq.Circuit(cirq.depolarize(p=1.0)(q0))
test_op = cirq.PauliSum() + cirq.Z(q0)
cirq.DensityMatrixSimulator().simulate_expectation_values(test_circuit_noise, test_op)
```
The reason we want to use this interface is to support expectation value simulations for noisy circuits that do not support materializing the density matrix explicitly.

95-martin-orion:
Fixes #3964. CC @zaqqwerty

This is a near-exact copy of the behavior and tests for `sparse_simulator`. If we see the need in the future, this code could probably be folded into a `SimulatesEVsAndFullState` interface as the default `simulate_expectation_values_sweep` behavior, but that doesn't seem particularly urgent at the moment.

Relevant changes from the `sparse_simulator` code:
- `obs.expectation_from_density_matrix(result.final_density_matrix, qmap)` instead of state-vector equivalent
- added `test_simulate_noisy_expectation_values` test
- that's it


95-martin-orion:
Assigning to @balopat (the next triage expert) for review, since I can't review my own PRs 😞

balopat (changes requested):
I would like to reduce the duplication introduced here. 

1.) production code: I would experiment with pulling out the repeated lines and have a new "protected" method `expectation_value_from_result(obs: PauliSum, result:TSimulationResult) -> float` which then can be called from the abstract method

2.) test code: I would move factor out the expectation value calculation logic to a separate test.py file and parameterized it with the different simulator types.

95-martin-orion:
> 1.) production code: I would experiment with pulling out the repeated lines and have a new "protected" method `expectation_value_from_result(obs: PauliSum, result:TSimulationResult) -> float` which then can be called from the abstract method

Packing the duplicated code in `simulate_expectation_values_sweep` into a `Result`-targeted method would force us to repeat the construction of `pslist` for every result. Probably not a huge overhead relative to simulation costs, but if we can avoid it I think we should.

I stand by my top-of-PR comment: the correct way to de-dupe this would be something closer to #3990. (Probably the cleanest solution would involve converting `*_sweep` methods to return `Iterable`, but that's a separate discussion.)

balopat:
> I stand by my top-of-PR comment: the correct way to de-dupe this would be something closer to #3990. (Probably the cleanest solution would involve converting `*_sweep` methods to return `Iterable`, but that's a separate discussion.)

Ah, cool, I missed that, and oh, nice, yes that is exactly the direction I was thinking - having an `expectation_value_from_state` (I don't like `expectation_from_state`) implemented - though you're also right that maybe not all Simulators should be mandated to implement EV calculation, and isolating that responsibility to a separate interface like `SimulatesEVandFullState` or similar could be useful to make EV calculation implementation opt-in. 

Also, note that #3990 adds yet another test with exactly the same cases at a first blink...maybe we should think about extracting a `cirq.testing.assert_consistently_calculates_expectation_value(simulator)` function that could be reused in tests for all the different implementations!

I'm okay to get this merged but I would love to have an issue created for deduping both impl and tests for this and referenced in a comment in all these places.

balopat (approved):
LGTM - given we add an issue and a comment for deduping tests & design

balopat:
ping @95-martin-orion we could merge this, just add in a comment so we don't forget to dedupe

95-martin-orion:
Opened #4209 to track deduplication.

PR #3981, opened 2021-03-30
minor updates to cirq intro

balopat:
Fixes some typos, consistent state vector naming, portable moments/qubits diagram, fix angles in R* gate definitions.


balopat:
cc @tanujkhattar

docs/tutorials/educators/intro.ipynb
```diff
@@ -70,7 +70,7 @@
    "source": [
```
tanujkhattar, line 70: <p>Defining LineQubits was intentional in the solution so that we can introduce the learners to different types of qubits. </p><p>They would have attempted NamedQubits in the exercise above and while discussing the solution, we'd point out that you can achieve the same task more easily by using LineQubit.range(4). </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/quantumlib/Cirq/pull/3981/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

tanujkhattar (approved):
LGTM % comment

docs/tutorials/educators/intro.ipynb
balopat: Sounds good! I'll add this sentiment in the comment then - that NamedQubits are similarly okay here!

PR #3983, opened 2021-04-01
Disallow qubit_map with conflicting dimensions in subcircuits

### LINKED ISSUE #3982: CircuitOperation.qubit_map allows mapping qubits of different dimension
**Description of the issue**

If you do `CircuitOperation(..., qubit_map={LineQubit(0), LineQid(1, dimension=3)})`, it is happy but then fails during simulation with obscure message.

See above

**Expected behavior**

It should fail at circuit construction time with a friendly error message.

**Notes**

The check needs to happen in `with_qubit_mapping` as well.

daxfohl:
Fixes #3982


95-martin-orion (approved):
Thanks for catching this!

PR #3984, opened 2021-04-01
[WIP] Improve API reference

### LINKED ISSUE #3667: Improve API reference
I would like to look at the API for the Moment class. I spent a while sifting through the new [Cirq documentation](https://quantumai.google/reference/python/cirq/all_symbols) but I could not find it. I tried the search bar but it was not there. I tried to look at the code on github but a [simple search](https://github.com/quantumlib/Cirq/search?q=Moment) does not reveal where the Moment class is defined. My last resort was to use my IDE (PyCharm) to find its definition, but some users do not have this capability.

Update the documentation page to make it easy to search for class definitions.

See above.

unaiic:
Updates the documentation page to make it easy to search for class definitions. Replaces `Class` mentions with `cirq.Class`.

Fixes #3667.

### DIFF
--- a/docs/circuits.ipynb
+++ b/docs/circuits.ipynb
@@ -86,7 +86,7 @@
    "source": [
     "## Conceptual overview\n",
     "\n",
-    "The primary representation of quantum programs in Cirq is the `Circuit` class. A `Circuit` is a collection of `Moments`. A `Moment` is a collection of `Operations` that all act during the same abstract time slice. An `Operation` is a some effect that operates on a specific subset of Qubits, the most common type of `Operation` is a `GateOperation`.\n"
+    "The primary representation of quantum programs in Cirq is the `cirq.Circuit` class. A `cirq.Circuit` is a collection of `Moments`. A `cirq.Moment` is a collection of `Operations` that all act during the same abstract time slice. An `cirq.Operation` is a some effect that operates on a specific subset of Qubits, the most common type of `cirq.Operation` is a `cirq.GateOperation`.\n"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


95-martin-orion:
My first impression here (which may be based on personal preference) is that using `cirq.Class` everywhere feels verbose - almost like we're being overzealous about branding, when in reality we're just trying to get things to link properly!

A couple thoughts on how to tone it back:
1. Switch to "Wikipedia-style" linking - i.e. only link the _first_ appearance of a term on each page.
2. Somehow link `Class` to the "cirq.Class" page.

My preference would be to apply both of these changes, but I suspect (2) requires fighting with the automated tools, which I'm not very familiar with. If that's the case, I think (1) should be enough on its own.

docs/circuits.ipynb
```diff
@@ -86,7 +86,7 @@
    "source": [
     "## Conceptual overview\n",
     "\n",
-    "The primary representation of quantum programs in Cirq is the `Circuit` class. A `Circuit` is a collection of `Moments`. A `Moment` is a collection of `Operations` that all act during the same abstract time slice. An `Operation` is a some effect that operates on a specific subset of Qubits, the most common type of `Operation` is a `GateOperation`.\n"
+    "The primary representation of quantum programs in Cirq is the `cirq.Circuit` class. A `cirq.Circuit` is a collection of `Moments`. A `cirq.Moment` is a collection of `Operations` that all act during the same abstract time slice. An `cirq.Operation` is a some effect that operates on a specific subset of Qubits, the most common type of `cirq.Operation` is a `cirq.GateOperation`.\n"
```
95-martin-orion, line 89: This may need a manual pass after replacements are made to check formatting. "An `cirq.Operation`" here should be "A `cirq.Operation`" - there may be similar minor fix-ups required.

unaiic: Yeah, it seems I only changed some of them. However, I think it'll be better to wait until a clear approach is taken.

unaiic:
When I was renaming them I also felt that: it seemed so verbose to me; too repetitive, and sometimes, even harder to read. Your first idea of linking only the first appearance of such classes looks good to me. Although I guess we'll need to discuss more about this...

95-martin-orion:
> When I was renaming them I also felt that: it seemed so verbose to me; too repetitive, and sometimes, even harder to read. Your first idea of linking only the first appearance of such classes looks good to me. Although I guess we'll need to discuss more about this...

Discussion on the Cirq sync reached the following conclusions:
- We should go ahead with the "Wikipedia style" of only linking the first appearance for each class.
- There isn't a good way to link `Class` without increasing maintenance costs, so we should stick with `cirq.Class` for linking.

95-martin-orion:
@unaiic, do you have time available to complete this PR? If not, I can look into getting additional help to wrap this up.

unaiic:
@95-martin-orion Yeah, I can finish it for this weekend.

PR #3987, opened 2021-04-02
Deprecate qubits and _qubits

### LINKED ISSUE #3080: Proper deprecation of `qubit` and `_qubits` on device
There is hacky code in `cirq.Device` that falls back to these method for `qubit_set`, apparently according to the comment to preserve backward compatibility.  These other methods should be properly deprecated and removed instead.

unaiic:
Creates a new generic `def deprecation_warning(deprecation: str, deadline:str, fix: str)` in `cirq._compat`. This method uses the `_warn_or_error` method in `_compat` and it validates the `deadline` argument (_v0.12_ in this case) like the other deprecators.

It also adds a test in `device_test.py` that tests the warning emitted with `cirq.testing.assert_deprecated` from a `Device` subclass implementing `qubits` and another implementing `_qubits`.

Fixes #3080.

### DIFF
--- a/cirq/_compat.py
+++ b/cirq/_compat.py
@@ -117,6 +117,16 @@ def _validate_deadline(deadline: str):
     assert re.match(DEADLINE_REGEX, deadline), "deadline should match vX.Y"
 
 
+def deprecation_warning(deprecation: str, deadline:str, fix: str):
+    _validate_deadline(deadline)
+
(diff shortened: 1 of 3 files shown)


95-martin-orion:
Hi @unaiic,

Please fix the failing tests, then re-request review.

PR #3989, opened 2021-04-03
Use Boolean Hamiltonian Gate for QAOA

tonybruguier:


ybc1991:
@tonybruguier @balopat I just want to mention that this diagonal Hamiltonian representation is highly related with diagonal gate decomposition #3890. One main application of diagonal gate is to simulate the Hamiltonian (Ref. Sec 4. in (https://iopscience.iop.org/article/10.1088/1367-2630/16/3/033040/pdf).

tonybruguier:
> @tonybruguier @balopat I just want to mention that this diagonal Hamiltonian representation is highly related with diagonal gate decomposition #3890. One main application of diagonal gate is to simulate the Hamiltonian (Ref. Sec 4. in (https://iopscience.iop.org/article/10.1088/1367-2630/16/3/033040/pdf).

Ohhhh, nice! Thanks for the reference.

I will need more time to digest the paper and your PR. I took a little time to study both, and my temporary understanding is that the approach you just linked is that it approximates a function defined by its enumerated values while the present PR is an exact representation of a function defined by its formula.

Also, do you suggest a refactoring (now or later)? I'm open to suggestions, I just could use some more advice on how to proceed (if you have the time). I will read the paper more carefully later no matter what.

ybc1991:
@tonybruguier I don't suggest any refactoring at all. Just mention the connection. For example, here this Hamiltonian representation exploited the special format of QOQA `\sum(1-s_is_j)`, which is not considered in the diagonal gate at all. But I am interested to have a deep look and check more differences and connections here.

tonybruguier:
> @tonybruguier I don't suggest any refactoring at all. Just mention the connection. For example, here this Hamiltonian representation exploited the special format of QAOA `\sum(1-s_is_j)`, which is not considered in the diagonal gate at all. But I am interested to have a deep look and check more differences and connections here.

Thanks. I think I understand a little bit more now. I think it is indeed the case that the diagonal gates uses the values directly to construct an approximate circuit, while the present PR uses the formula to construct an exact circuit.

It seems the circuit could be a bit different though:
https://github.com/tonybruguier/sandbox/blob/master/try_out_diagonal_gate.py#L29

ybc1991:
The main difference and the biggest advantage of this work, I think, is it does not evaluate the whole truth table of boolean function directly to construct the circuit. Meanwhile, the diagonal gate has to know all values (not limited to boolean function through).
I want to correct that diagonal gates give the exact circuit instead of the approximate one.

> It seems the circuit could be a bit different though:
https://github.com/tonybruguier/sandbox/blob/master/try_out_diagonal_gate.py#L29

Yes, they will give different circuits. As for this particular circuit example you showed, they are exactly the same actually (if you ignore the identical operator`Rz(0)`). You can try `print(cirq.Circuit(cirq.decompose(circuit2)))` to see. (`cirq.decompose` will further decompose the CNOT gates)

Since the connection, I am reviewing this PR.  (so me again :) )  Will send more comments after I understand this better.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,135 @@
...
+
+def get_name_to_id(boolean_exprs):
+    # For run-to-run identicalness, we sort the symbol name lexicographically.
+    symbol_names = sorted(
+        {symbol.name for boolean_expr in boolean_exprs for symbol in boolean_expr.free_symbols}
+    )
+    return {symbol_name: i for i, symbol_name in enumerate(symbol_names)}
+
+
+def build_circuit_from_hamiltonians(hamiltonians, qubits, theta):
```
ybc1991, line 122: Do you think it is good to have another function `build_circuit_from_boolean_expressions`, wrapping `parse_expr`, `build_hamiltonian_from_boolean` and `get_name_to_id`?  I know it is NOT useful in QAOA since the same `hamiltonians` with different theta will be used multiple times. But what about potential cases?

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,135 @@
...
+    circuit = cirq.Circuit()
+    for hamiltonian in hamiltonians:
+        for h, w in hamiltonian.hamiltonians.items():
+            for i in range(1, len(h)):
+                circuit.append(cirq.CNOT(qubits[h[i]], qubits[h[0]]))
```
ybc1991, line 127: Can be written into one concise line:
`circuit.append([cirq.CNOT(qubits[c], qubits[h[0]]) for c in h[1:]])`.

I am ok with the for-loop style as well.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,135 @@
...
+        boolean_expr: A Sympy expression containing symbols and Boolean operations
+        name_to_id: A dictionary from symbol name to an integer, typically built by calling
+            get_name_to_id().
+
+    Return:
+        The HamiltonianList that represents the Boolean expression.
+    """
+
+    if isinstance(boolean_expr, (And, Not, Or, Xor)):
+        sub_hamiltonians = [
```
ybc1991, line 85: Is it true that `sub_hamiltonians` always has a length 1 or 2? If yes, I think the later construction can be simplified

ybc1991 (commented):
@tonybruguier This is an interesting work, thanks for adding an example for this! I left some comments more about the coding actually.

After understanding better the reference you gave, the following is my latest understanding of the connection between diagonal gates :
* Same: The underlying math foundation is exactly the same. One called Fourier expansions, one called Walsh-Hadamard transform (They are the same thing under Z_2^n). The construction of these basis/expansions -- `exp^(-i \gamma H_f)` through CNOT and Z gates are the same as well. (Some re-ordering for optimization is used in the diagonal gate but is not important here.)
* Diff: Diagonal gate use the full Walsh transform to get these coefficients directly, which requires `O(Nlog(N)) = O(n2^n)` even under the FFT style. Hamiltonian represents through the boolean expressions and composite rules. Under many well-structured situations, like the ISING model, it only needs to be expanded to be quadratic terms, hence it can save huge computation requirements. (I need to understand better on the cost of general cases)

examples/hamiltonian_representation.py
tonybruguier: Not with my current version of Sympy. For `test_build_hamiltonian_from_boolean` with input `x0 ^ x1 ^ x2` the XOR has three operands. I think we could try to do things recursively, of course, but I am not sure it would make the code cleaner.

examples/hamiltonian_representation.py
tonybruguier: Done.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,135 @@
...
+        return HamiltonianList(hamiltonians)
+
+    def __rmul__(self, other: float):
+        return HamiltonianList({k: other * w for k, w in self.hamiltonians.items()})
+
+    def __mul__(self, other):
+        hamiltonians = {}
+        for h1, w1 in self.hamiltonians.items():
+            for h2, w2 in other.hamiltonians.items():
+                h = tuple(sorted(set(h1).symmetric_difference(h2)))
```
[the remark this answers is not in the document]
tonybruguier: Added a comment describing what it does. The name is mine, and I am not sure it is standard. Happy to change the name if you have a favorite.

examples/hamiltonian_representation_test.py
```diff
@@ -0,0 +1,62 @@
+import math
+
+import numpy as np
+import pytest
+from sympy.parsing.sympy_parser import parse_expr
+
+import cirq
+import examples.hamiltonian_representation as hr
+
+# These are some of the entries of table 1.
```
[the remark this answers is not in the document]
tonybruguier: Done.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,135 @@
...
+            for h in sorted(self.hamiltonians)
+        )
+
+    def __add__(self, other):
+        return self._signed_add(other, 1.0)
+
+    def __sub__(self, other):
+        return self._signed_add(other, -1.0)
+
+    def _signed_add(self, other, sign: float):
```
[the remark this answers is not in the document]
tonybruguier: Done here and for others.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,135 @@
...
+# [3] https://github.com/rsln-s/IEEE_QW_2020/blob/master/Slides.pdf
+
+
+class HamiltonianList:
+    """A container class of Boolean function as equation (2) or [1]"""
+
+    def __init__(self, hamiltonians: Dict[Tuple[int, ...], float]):
+        # The representation is Tuple[int, ...] to weights. The tuple contains the integers of
+        # where Z_i is present. For example, Z_0.Z_3 would be (0, 3), and I is the empty tuple.
+        self.hamiltonians = {h: w for h, w in hamiltonians.items() if math.fabs(w) > 1e-12}
```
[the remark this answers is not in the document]
tonybruguier: Like this?

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,135 @@
...
+            hamiltonian = HamiltonianList.O()
+            for sub_hamiltonian in sub_hamiltonians:
+                hamiltonian = hamiltonian + sub_hamiltonian - 2.0 * hamiltonian * sub_hamiltonian
+        return hamiltonian
+    elif isinstance(boolean_expr, Symbol):
+        # Table 1 of [1], entry for 'x' is '1/2.I - 1/2.Z'
+        i = name_to_id[boolean_expr.name]
+        return 0.5 * HamiltonianList.I() - 0.5 * HamiltonianList.Z(i)
+    else:
+        raise ValueError(f'Unsupported type: {type(boolean_expr)}')
```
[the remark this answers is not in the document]
tonybruguier: Done.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,135 @@
...
+    symbol_names = sorted(
+        {symbol.name for boolean_expr in boolean_exprs for symbol in boolean_expr.free_symbols}
+    )
+    return {symbol_name: i for i, symbol_name in enumerate(symbol_names)}
+
+
+def build_circuit_from_hamiltonians(hamiltonians, qubits, theta):
+    circuit = cirq.Circuit()
+    for hamiltonian in hamiltonians:
+        for h, w in hamiltonian.hamiltonians.items():
```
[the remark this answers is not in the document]
tonybruguier: Renamed the variables and added some mypy comments.

examples/hamiltonian_representation.py
tonybruguier: Done.

tonybruguier:
Thanks. PTAL. I am thinking about what you said, and trying to see whether I can improve the number of gates.

examples/hamiltonian_representation.py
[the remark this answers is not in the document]
tonybruguier: Done.

tonybruguier:
I think I can now apply the speedup of the diagonal gate codes directly. What do you think?

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,135 @@
...
+            for i in range(1, len(h)):
+                circuit.append(cirq.CNOT(qubits[h[i]], qubits[h[0]]))
+
+            if len(h) >= 1:
+                circuit.append(cirq.Rz(rads=(theta * w)).on(qubits[h[0]]))
```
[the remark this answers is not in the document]
tonybruguier: Is this an issue? Presumably, the example would only be shown only in subsequent releases. Let me know.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,135 @@
...
+        self.hamiltonians = {h: w for h, w in hamiltonians.items() if math.fabs(w) > 1e-12}
+
+    def __str__(self):
+        # For run-to-run identicalness, we sort the keys lexicographically.
+        return "; ".join(
+            f"{self.hamiltonians[h]:.2f}.{'.'.join('Z_%d' % d for d in h) if h else 'I'}"
+            for h in sorted(self.hamiltonians)
+        )
+
+    def __add__(self, other):
```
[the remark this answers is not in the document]
tonybruguier: Done.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,135 @@
...
+
+class HamiltonianList:
+    """A container class of Boolean function as equation (2) or [1]"""
+
+    def __init__(self, hamiltonians: Dict[Tuple[int, ...], float]):
+        # The representation is Tuple[int, ...] to weights. The tuple contains the integers of
+        # where Z_i is present. For example, Z_0.Z_3 would be (0, 3), and I is the empty tuple.
+        self.hamiltonians = {h: w for h, w in hamiltonians.items() if math.fabs(w) > 1e-12}
+
+    def __str__(self):
```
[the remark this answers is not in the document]
tonybruguier: Done.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def __sub__(self, other):
+        return self._signed_add(other, -1.0)
+
+    def _signed_add(self, other, sign: float):
+        hamiltonians = self.hamiltonians.copy()
+        for h, w in other.hamiltonians.items():
+            if h not in hamiltonians:
+                hamiltonians[h] = 0
+            hamiltonians[h] += sign * w
```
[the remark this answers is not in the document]
tonybruguier: Done.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def __init__(self, hamiltonians: Dict[Tuple[int, ...], float]):
+        # The representation is Tuple[int, ...] to weights. The tuple contains the integers of
+        # where Z_i is present. For example, Z_0.Z_3 would be (0, 3), and I is the empty tuple.
+        self.hamiltonians = {h: w for h, w in hamiltonians.items() if math.fabs(w) > 1e-12}
+
+    def __str__(self):
+        # For run-to-run identicalness, we sort the keys lexicographically.
+        return "; ".join(
+            f"{self.hamiltonians[h]:.2f}.{'.'.join('Z_%d' % d for d in h) if h else 'I'}"
```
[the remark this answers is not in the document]
tonybruguier: Done.

examples/hamiltonian_representation.py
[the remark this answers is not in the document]
ybc1991: Then I definitely prefer to call it `HamiltonianPolynomials`. Under it, the dict`self._hamiltonians` has a clear meaning that representing each term with its coefficients. These numerical operations of the class are natural. (Like this \_\_mul\_\_  function, I was confusing what does List multiply List means until realizing it is a polynomial multiplication)

examples/hamiltonian_representation.py
ybc1991: Aha, you are right. I am not familiar with sympy_parser. Then, current style is good.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,235 @@
...
+            build_hamiltonian_from_boolean().
+        qubits: The list of qubits corresponding to the variables.
+        theta: A single float scaling the rotations.
+
+    Return:
+        A dictionary of string (the variable name) to a unique integer.
+    """
+    combined = HamiltonianList.O()
+    for hamiltonian_list in hamiltonian_lists:
+        combined += hamiltonian_list
```
ybc1991, line 176: it can be further simplified into `combined = sum(hamiltonian_lists,  HamiltonianList.O())`. Using the polynomial name, make it more meaningful in my opinion:  `combined = sum(hamiltonian_polynomial_list,  HamiltonianPolynomial.O())` without worry which is list mean what.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,235 @@
...
+            # This would be equivalent to the lines below, but it should use fewer gates.
+        else:
+            _apply_cnots(previous_h)
+            _apply_cnots(h)
+
+        if len(h) >= 1:
+            circuit.append(cirq.Rz(rads=(theta * w)).on(qubits[h[-1]]))
+
+        previous_h = h
+
```
ybc1991, line 210: I really like this change combining two ideas. But compared with the previous code, this version becomes less readable. (Mainly because here it is always from [0 to 2^n-1], the structure of gray code is not traversed.) So I suggest keep the older structure:
```
for h in sorted_hs:
      _apply_cnots(previous_h, h, forward=True)
      if len(h) >= 1:
            circuit.append(cirq.Rz(rads=(theta * w)).on(qubits[h[-1]]))
      previous_h = h
```
move the detailed cnot canceling into sub-functions

ybc1991 (commented):
Yeah, that optimizing CNOT looks good to me. Add a few more comments about the hamiltonian_representation.py

examples/hamiltonian_representation.py
tonybruguier: Done. Also did some renaming, as per your suggested convention.

examples/hamiltonian_representation.py
tonybruguier: Done. I wasn't sure what the `forward` was for though, but otherwise I did what you suggested. It's much easier to read indeed.

tonybruguier:
Thanks. PTAL.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,242 @@
...
+from typing import DefaultDict, Dict, List, Sequence, Tuple
+
+from sympy.logic.boolalg import And, Not, Or, Xor
+from sympy.core.expr import Expr
+from sympy.core.symbol import Symbol
+import sympy.parsing.sympy_parser as sympy_parser
+
+import cirq
+
+# References:
```
Strilanc, line 13: Move this into the docstring of the class.

tonybruguier: Done.

tonybruguier:
Thanks. PTAL.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,243 @@
...
+        list(combined.hamiltonians.keys()), key=functools.cmp_to_key(_gray_code_comparator)
+    )
+
+    def _apply_cnots(previous_h, h):
+        # This function applies in sequence the CNOTs from previous_h and then h. However, given
+        # that the h are sorted in Gray ordering and that some cancel each other, we can reduce the
+        # number of gates. See [4] for more details.
+
+        # We first test whether the rotation is applied on the same qubit.
+        last_qubit_is_same = previous_h and h and previous_h[-1] == h[-1]
```
ybc1991, line 193: Do we really need this `last_qubit_is_same` condition?  
Note that to implement `exp(i\theta ZZZ)`, the following two forms are equivalent:
1. CNOT(3, 1) CNOT(2, 1) Rz_1(\theta) CNOT(2, 1) CNOT(3, 1)
2. CNOT(3, 2) CNOT(2, 1) Rz_1(\theta) CNOT(2, 1) CNOT(3, 2)

As long as there are common binary representation bits in `Z` string, you can cancel the CNOT gates.

tonybruguier: Sorry, I don't quite understand what you suggest I do. I do agree with the equalities that you mentioned, I think it's figure 3 of the paper.

I also do agree that we can do the Rz on any qubit that has an Z, but choosing a different one also means changing how the CNOTs are applied. Don't they say that they "_adopt the convention in 3(b) where the CNOTs are always targeted on the highest order qubit possible_"? I mean that yes, we could change the applied qubit, but at some point, we are going to have to change it.

However, I am missing something indeed. They equations 8, 9, 10, and 11 can be applied to simplify the circuit further, but I don't understand in which order to apply them. It seems there's a general procedure, as indicated by footnote 10, but I couldn't find a new paper.

I will continue reading the paper of course, but if you have a hint, that would be helpful. In the meantime, I had realized that some of the unit tests were not very thorough, so I added some more. In particular, I'm focusing on this one:

`
(x0 ^ x1 ^ x2) | (x2 ^ x3 ^ x4)

(x0 ^ x1 ^ x2) | (x2 ^ x3 ^ x4)

                               ┌──┐
x0: ───@─────────────────────@─────@──────────────────────────@─────────────────────────────────
       │                     │     │                          │
x1: ───┼───@─────────────────┼────@┼────@─────────────────────┼───@─────────────────────────────
       │   │                 │    ││    │                     │   │
x2: ───X───X───Rz(-0.025π)───X────X┼────┼─────────────────────┼───┼───@─────────────────@───────
                                   │    │                     │   │   │                 │
x3: ───────────────────────────────┼────┼───@─────────────────┼───┼───┼─────────────────┼───@───
                                   │    │   │                 │   │   │                 │   │
x4: ───────────────────────────────X────X───X───Rz(-0.025π)───X───X───X───Rz(-0.025π)───X───X───
`

tonybruguier:
Please take a look at my comment. I'm still thinking about it, but a hit could be helpful.

examples/hamiltonian_representation.py
ybc1991: Sorry, I don't have a plan in detail on this but can elaborate more about what I want to express and why intuitively I think `last_qubit_is_same` can be removed.  Simple example:
We want to have a rotation on ZZZ and ZZI. In Hamiltonian, it is `{(0, 1, 2): w1}` and `{(0, 1): w2}`. They don't have the same last qubit for rotation (assuming we always using the largest number for rotation one). However, the CNOT between them still can be canceled. Look at the following two realization:
image
Versus
image
(BTW, i cannot read the ASCII plot you posted previously. Using SVG plot is better here.)
They are the equivalent circuit but the second case has two less CNOT gates. 

I also realize it may not be easy to have Programmable. The key should be the eq. 10,-14 as you pointed out. But I need more time to re-digest that in this case. Will add more thoughts if I have more concrete ideas.

tonybruguier: I modified the code and test. I have implemented some of the simplifications, but I still need to implement equation 11.

It seems like it's a greedy approach, though, so I am not sure it's optimal. Nonetheless, it is still a superset of what I had before.

I will continue to implement equation 11 unless you have objections.

tonybruguier: I implemented equation 11. I think the circuit you had above wasn't an example of such simplification.

The algo is greedy, but it could be optimal too. I haven't wrapped my head around it. Maybe the margin was too small for their truly remarkable proof, so they opted for a footnote :)

Either way, I think I addressed your suggestion, but happy to do more if not.

tonybruguier:
Thanks. PTAL.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,322 @@
...
+        if ladder_target:
+            cnots.extend((ph[i], ph[i + 1]) for i in reversed(range(len(ph) - 1)))
+            cnots.extend((ch[i], ch[i + 1]) for i in range(len(ch) - 1))
+        else:
+            cnots.extend((ph[i], ph[-1]) for i in range(len(ph) - 1))
+            cnots.extend((ch[i], ch[-1]) for i in range(len(ch) - 1))
+
+        while True:
+            # As per equations 9 and 10, if all the targets (resp. controls) are the same, the
+            # cnots commute. Further, if the control (resp. targets) are the same, the cnots can be
```
ybc1991, line 205: I think the code is difficult to follow is partial because the data structure of CNOTs is hard to look up.
You can transfer that into two lists by `control_qubits, target_qubits = zip(*cnots)`. Now you can use built-in like`any`, `all` instead of nested for-loop and easier to read I think.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,322 @@
...
+
+            if found_simplification:
+                continue
+
+            # Here we apply the simplification of equation 11. Note that by flipping the control
+            # and target qubits, we have an equally valid identity:
+            # CNOT(i, j) @ CNOT(j, k) == CNOT(j, k) @ CNOT(i, k) @ CNOT(i, j)
+            # CNOT(j, i) @ CNOT(k, j) == CNOT(k, j) @ CNOT(k, i) @ CNOT(j, i)
+            #
+            # Which are represented by (x, y) = (0, 1) and (x, y) = (1, 0), respectively.
```
ybc1991, line 229: I read the original paper again and it sounds very vague to me as well. They said it can help to improve but doesn't give any systematic approach. The approach you used is greedy as well. I don't know better solutions.  Your key idea is repeatedly to try the above three CNOTs to two CNOTs, right?

The complexity of code is out of my original imagination...  (I was thinking about some simple improvements..)

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,322 @@
...
+            for x, y in [(0, 1), (1, 0)]:
+                # We investigate potential pivots sequentially.
+                for j in range(1, len(cnots) - 1):
+                    # First, we look back for as long as the targets (resp. triggers) are the same.
```
ybc1991, line 233: actually, what does targets (resp. triggers) mean?

ybc1991 (commented):
I read the original paper again and it sounds very vague to me as well. They said it can help to improve but doesn't give any systematic approach. IIUC, Your key idea is repeatedly to try the three CNOTs to two CNOTs, right? This implementation is the greedy approach. But I don't know if there is a better solution. 

The complexity of code is out of my original imagination... (I was thinking about some simple improvements..)

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,322 @@
...
+
+        if ladder_target:
+            cnots.extend((ph[i], ph[i + 1]) for i in reversed(range(len(ph) - 1)))
+            cnots.extend((ch[i], ch[i + 1]) for i in range(len(ch) - 1))
+        else:
+            cnots.extend((ph[i], ph[-1]) for i in range(len(ph) - 1))
+            cnots.extend((ch[i], ch[-1]) for i in range(len(ch) - 1))
+
+        while True:
+            # As per equations 9 and 10, if all the targets (resp. controls) are the same, the
```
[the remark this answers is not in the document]
tonybruguier: Done.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,322 @@
...
+
+    circuit = cirq.Circuit()
+
+    # Here we follow improvements of [4] cancelling out the CNOTs. The first step is to order by
+    # Gray code so that as few as possible gates are changed.
+    sorted_hs = sorted(
+        list(combined.hamiltonians.keys()), key=functools.cmp_to_key(_gray_code_comparator)
+    )
+
+    def _apply_cnots(ph: Tuple[int, ...], ch: Tuple[int, ...]):
```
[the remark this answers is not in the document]
tonybruguier: Yes, changed the names to be clearer.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,322 @@
...
+
+        cnots: List[Tuple[int, int]] = []
+
+        if ladder_target:
+            cnots.extend((ph[i], ph[i + 1]) for i in reversed(range(len(ph) - 1)))
+            cnots.extend((ch[i], ch[i + 1]) for i in range(len(ch) - 1))
+        else:
+            cnots.extend((ph[i], ph[-1]) for i in range(len(ph) - 1))
+            cnots.extend((ch[i], ch[-1]) for i in range(len(ch) - 1))
+
```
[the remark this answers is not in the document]
tonybruguier: Done.

examples/hamiltonian_representation.py
tonybruguier: Ack. Maybe we leave it as is and wait to see whether they publish more?

examples/hamiltonian_representation.py
tonybruguier: (x, y) can be either (target, control) or (control, target).

examples/hamiltonian_representation.py
tonybruguier: Actually, I would need to flip control and target to get both equalities. Would renaming the loop that generate (x, y) instead clearer?

tonybruguier:
Thanks. PTAL.

examples/hamiltonian_representation.py
ybc1991: Yes, I notice you flipped the control and target. But after I read it again, maybe we don't even need to worry about the flip at least for the first several lines (208-219). I think it can be simplified into this:
```
# find consecutive same elements
for i, cnot in enumerate(cnots):
    if i > 0 and cnot == cnots[i-1]:
        cnots.pop(i)
        cnots.pop(i-1)
        break
```
Do I understand this correctly?

examples/qaoa.py
```diff
@@ -1,155 +1,71 @@
-"""Runs the Quantum Approximate Optimization Algorithm on Max-Cut.
```
daxfohl, line 1: Is there a reason to remove the max cut example?

tonybruguier: Actually, it's still max-cut, but via a Boolean functions. I restored the comment that I shouldn't have deleted.

examples/hamiltonian_representation.py
tonybruguier: Yes, you understood it correctly, I think. However, re-reading the code, I think the simplification was not complete. If the CNOTs commute, then I should look at all the commuting CNOTs for identical targets/controls. I modified the code to reflect this. Sorry about this mistake.

tonybruguier:
Thanks. PTAL.

ybc1991:
In the future PR, maybe it is a good idea to add this Hamiltonian Representation into the diagonal gate (Or some oracle gate)?
Like:
```python
Class DiagonalGate:

    @staticmethods
    def build_from_boolean_expression(bool_expr, ...) -> DiagonalGate:
           ....
```
There are many detailed things like gate versus operation, concrete protocols, etc. to consider. In general, I think it is doable and beneficial. Any thoughts?

tonybruguier:
I added a TODO

tonybruguier:
@balopat
I think last week it was decided that you would do the review. Please let me know if you have any comment on this PR.

examples/hamiltonian_representation.py
```diff
@@ -0,0 +1,338 @@
...
+import sympy.parsing.sympy_parser as sympy_parser
+
+import cirq
+
+
+# TODO(tonybruguier): Consider making a stand-alone gate given a Boolean expression.
+
+
+class HamiltonianPolynomial:
+    """A container class of Boolean function as equation (2) or [1]
```
balopat, line 18:
```suggestion
    """A container class of Boolean function as equation (2) of [1]
```

balopat (changes requested):
Thank you so much for this. Also huge thank you to @ybc1991  for the thorough review. 
I think this has the potential to become a valid operator representation in Cirq. I would move it to cirq.ops. That would also fix the issue of the fact that examples shouldn't refer to each other they should be simple examples using / demonstrating Cirq features - we shouldn't really build complex "frameworks" within the example folder.

Also, for the cases where it can beat DiagonalGate, I think DiagonalGate should delegate to it in terms of decomposition. What do you think @ybc1991, can we identify those cases?

examples/qaoa.py
```diff
@@ -1,155 +1,72 @@
...
```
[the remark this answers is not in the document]
tonybruguier: Reverted

tonybruguier:
@balopat 
Thanks for the review. I am more than happy to make it a gate; Actually, it feels like recognition, so I'll take it. I did just that and moved to `cirq.ops`. I have never created a gate before, so I inspired myself from what I saw in that directory. Happy to add missing parts, if any.

As for the example, I *think* what you meant is that it's OK to keep the changes in `examples/qaoa.py` as long we don't refer to other examples, but rather the core Cirq code. I think this is what I do now, but if you instead want me to leave the examples completely untouched, please let me know. 

I haven't done the delegation, because maybe this PR is large enough already?

balopat (changes requested):
First round of nits. I'll keep reviewing this in the next days.

cirq-core/cirq/ops/hamiltonian_gate.py
```diff
@@ -0,0 +1,354 @@
...
+        )
+
+    @property
+    def hamiltonians(self) -> DefaultDict[Tuple[int, ...], float]:
+        return self._hamiltonians
+
+    def __repr__(self):
+        # For run-to-run identicalness, we sort the keys lexicographically.
+        formatted_terms = [
+            f"{self._hamiltonians[h]:.2f}.{'.'.join('Z_%d' % d for d in h) if h else 'I'}"
```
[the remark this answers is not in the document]
tonybruguier: Done.

cirq-core/cirq/ops/hamiltonian_gate.py
```diff
@@ -0,0 +1,354 @@
...
+        elif isinstance(boolean_expr, Xor):
+            hamiltonian = HamiltonianPolynomial.O()
+            for sub_hamiltonian in sub_hamiltonians:
+                hamiltonian = hamiltonian + sub_hamiltonian - 2.0 * hamiltonian * sub_hamiltonian
+        return hamiltonian
+
+    raise ValueError(f'Unsupported type: {type(boolean_expr)}')
+
+
+def _gray_code_comparator(k1, k2, flip=False):
```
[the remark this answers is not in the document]
tonybruguier: Done.

cirq-core/cirq/ops/hamiltonian_gate.py
```diff
@@ -0,0 +1,354 @@
...
+    if max_1 != max_2:
+        return -1 if (max_1 < max_2) ^ flip else 1
+    if max_1 == -1:
+        return 0
+    return _gray_code_comparator(k1[0:-1], k2[0:-1], not flip)
+
+
+def _get_gates_from_hamiltonians(
+    hamiltonian_polynomial_list: List[HamiltonianPolynomial],
+    qubits,
```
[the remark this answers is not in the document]
tonybruguier: Done.

cirq-core/cirq/ops/hamiltonian_gate.py
```diff
@@ -0,0 +1,354 @@
...
+from sympy.core.symbol import Symbol
+import sympy.parsing.sympy_parser as sympy_parser
+
+import cirq
+from cirq import value
+from cirq.ops import raw_types
+
+
+class HamiltonianPolynomial:
+    """A container class of Boolean function as equation (2) of [1]
```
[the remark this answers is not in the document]
tonybruguier: Added a blurb.

cirq-core/cirq/ops/hamiltonian_gate.py
```diff
@@ -0,0 +1,354 @@
...
+
+        previous_h = h
+
+    # Flush the last CNOTs.
+    yield _apply_cnots(previous_h, ())
+
+
+@value.value_equality
+class HamiltonianGate(raw_types.Gate):
+    """A gate that applies an Hamiltonian from a set of Boolean functions."""
```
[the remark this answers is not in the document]
tonybruguier: Done.

cirq-core/cirq/ops/hamiltonian_gate.py
```diff
@@ -0,0 +1,354 @@
...
+
+    def __init__(self, boolean_strs: Sequence[str], theta: float, ladder_target: bool):
+        """
+        Builds an HamiltonianGate.
+
+        Args:
+            boolean_strs: The list of Sympy-parsable Boolean expressions.
+            theta: The list of thetas to scale the Hamiltonian.
+            ladder_target: Whether to use convention of figure 7a or 7b.
```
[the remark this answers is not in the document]
tonybruguier: Added a long description. I hope it helps.

cirq-core/cirq/ops/hamiltonian_gate.py
```diff
@@ -0,0 +1,354 @@
...
+            _build_hamiltonian_from_boolean(boolean_expr, name_to_id)
+            for boolean_expr in boolean_exprs
+        ]
+        self._num_qubits = len(name_to_id)
+
+    def num_qubits(self) -> int:
+        return self._num_qubits
+
+    @staticmethod
+    def get_name_to_id(boolean_exprs: Sequence[Expr]) -> Dict[str, int]:
```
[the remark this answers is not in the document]
tonybruguier: It is kind of useful to have it public, though, isn't it? If you do know in advance what your variables are, it is indeed not needed.

However, having a way to only specify the Boolean expressions and getting the variable names is useful to create the qubits, I think. As a matter of fact, I do call that function in the QAOA example.

I don't feel strongly about it though. Leaving this unresolved. What do you think?

cirq-core/cirq/ops/hamiltonian_gate.py
```diff
@@ -0,0 +1,354 @@
...
+
+    # Flush the last CNOTs.
+    yield _apply_cnots(previous_h, ())
+
+
+@value.value_equality
+class HamiltonianGate(raw_types.Gate):
+    """A gate that applies an Hamiltonian from a set of Boolean functions."""
+
+    def __init__(self, boolean_strs: Sequence[str], theta: float, ladder_target: bool):
```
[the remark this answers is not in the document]
tonybruguier: I added a TODO.

cirq-core/cirq/ops/hamiltonian_gate.py
```diff
@@ -0,0 +1,354 @@
...
+            yield cirq.Rz(rads=(theta * w)).on(qubits[h[-1]])
+
+        previous_h = h
+
+    # Flush the last CNOTs.
+    yield _apply_cnots(previous_h, ())
+
+
+@value.value_equality
+class HamiltonianGate(raw_types.Gate):
```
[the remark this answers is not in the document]
tonybruguier: I changed the name to your use your suggestion. I also took the liberty of renaming the file, since it seems to be better in line with the conventions.

tonybruguier:
@balopat 
Do you have any further comment? It's not super urgent, but I could work on them by the weekend.

balopat (changes requested):
Added a bunch of comments - I think this should be an Operation subclass similar to how PauliString is. Also instead of creating a new implementation (HamiltonianPolynomial), we should just leverage PauliSum.

Regards to `get_name_to_id`: I would get rid of it and just use cirq.parameter_names, which happens to work just okay...

`cirq.parameter_names(parse_expr("x1 ^ b | a & c"))`

And then I would recommend making the user pass in a `Dict[str,Qid]` to the `BooleanHamiltonian` similar how PauliString does it.

cirq-core/cirq/ops/boolean_hamiltonian_operation.py
```diff
@@ -0,0 +1,406 @@
...
+            yield cirq.Rz(rads=(theta * w)).on(qubits[h[-1]])
+
+        previous_h = h
+
+    # Flush the last CNOTs.
+    yield _apply_cnots(previous_h, ())
+
+
+@value.value_equality
+class BooleanHamiltonianOperation(raw_types.Gate):
```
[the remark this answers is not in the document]
tonybruguier: Done.

cirq-core/cirq/ops/boolean_hamiltonian_operation.py
```diff
@@ -0,0 +1,406 @@
...
+        an XOR.
+
+        Then, we compute exp(j * theta * polynomial), which is unitary because the polynomial is
+        Hermitian.
+
+        Args:
+            boolean_strs: The list of Sympy-parsable Boolean expressions.
+            theta: The list of thetas to scale the Hamiltonian.
+            ladder_target: Whether to use convention of figure 7a or 7b.
+            symbol_names: A list of symbol names that should be a superset of all the symbols used.
```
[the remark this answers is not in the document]
tonybruguier: Done. I think you had suggested to have `symbol_names` but it doesn't matter. The new version with the qubit map is cleaner, doesn't have it, and I like it better. Done.

cirq-core/cirq/ops/boolean_hamiltonian_operation.py
```diff
@@ -0,0 +1,406 @@
...
+from sympy.core.expr import Expr
+from sympy.core.symbol import Symbol
+import sympy.parsing.sympy_parser as sympy_parser
+
+import cirq
+from cirq import value
+from cirq.ops import raw_types
+
+
+class HamiltonianPolynomial:
```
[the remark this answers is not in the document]
tonybruguier: Much nicer indeed. This removes a lot of code. Thank you for the suggestion. Done.

tonybruguier:
@balopat 
Thank you for the thorough review. A lot has changed, but I think I have addressed all your comments so far (no more unresolved ones).

PTAL when you can?

balopat:
Thanks @tonybruguier - the design is looking now much better, it's more cirq-esque! :) 👍  The last bit to review for me is actual correctness compared to the paper, that will take a couple of days but I think I will be able to close this out this week.

examples/qaoa.py
```diff
@@ -1,155 +1,76 @@
...
-    # Make qubits
-    qubits = cirq.LineQubit.range(n)
 
-    # Print an example circuit
-    betas = np.random.uniform(-np.pi, np.pi, size=p)
-    gammas = np.random.uniform(-np.pi, np.pi, size=p)
-    circuit = qaoa_max_cut_circuit(qubits, betas, gammas, graph)
-    print('Example QAOA circuit:')
-    print(circuit.to_text_diagram(transpose=True))
+def qaoa(booleans, repetitions, maxiter, p):
```
balopat, line 21: I know it wasn't documented in the previous version either, but can we add documentation on what these parameters are?

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,337 @@
...
+            # Here we apply the simplification of equation 11. Note that by flipping the control
+            # and target qubits, we have an equally valid identity:
+            # CNOT(i, j) @ CNOT(j, k) == CNOT(j, k) @ CNOT(i, k) @ CNOT(i, j)
+            # CNOT(j, i) @ CNOT(k, j) == CNOT(k, j) @ CNOT(k, i) @ CNOT(j, i)
+            for x, y in [(_control, _target), (_target, _control)]:
+                # We investigate potential pivots sequentially.
+                for j in range(1, len(cnots) - 1):
+                    # First, we look back for as long as the targets (resp. triggers) are the same.
+                    # They all commute, so all are potential candidates for being simplified.
+                    common_A: Dict[int, int] = {}
```
balopat, line 183: don't use capital letters in variable names

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,337 @@
...
+                    for i in range(j - 1, -1, -1):
+                        if cnots[i][y] != cnots[j][y]:
+                            break
+                        # We take a note of the trigger (resp. target).
+                        common_A[cnots[i][x]] = i
+
+                    # Next, we look forward for as long as the triggers (resp. targets) are the
+                    # same. They all commute, so all are potential candidates for being simplified.
+                    common_B: Dict[int, int] = {}
```
balopat, line 192:
```suggestion
                    common_b: Dict[int, int] = {}
```

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,337 @@
...
+    ladder_target: bool = False,
+):
+    """Builds a circuit according to [1].
+
+    Args:
+        hamiltonian_polynomial_list: the list of Hamiltonians, typically built by calling
+            _build_hamiltonian_from_boolean().
+        qubit_map: map of string (boolean variable name) to qubit.
+        theta: A single float scaling the rotations.
+        ladder_target: Whether to use convention of figure 7a or 7b.
```
balopat, line 120: Can you elaborate on the difference between these two figures in words? It is hurting readability to have to read the paper back and forth.

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,337 @@
...
+    It is essentially a polynomial of Pauli Zs on different qubits. For example, this object could
+    represent the polynomial 0.5*I - 0.5*Z_1*Z_2 and it would be stored inside this object as:
+    self._hamiltonians = {(): 0.5, (1, 2): 0.5}.
+
+    While this object can represent any polynomial of Pauli Zs, in this file, it will be used to
+    represent a Boolean operation which has a unique representation as a polynomial. This object
+    only handle basic operation on the polynomials (e.g. multiplication). The construction of a
+    polynomial from a Boolean is performed by _build_hamiltonian_from_boolean().
+
+    References:
```
balopat, line 40: Let's put these references on the top of the module, that way it shows the "global scope" of the papers.

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def _simplify_cnots(cnots):
+        _control = 0
+        _target = 1
+
+        while True:
+            # As per equations 9 and 10 of [4], if all the targets (resp. controls) are the same,
+            # the cnots commute. Further, if the control (resp. targets) are the same, the cnots
+            # can be simplified away:
+            # CNOT(i, j) @ CNOT(k, j) = CNOT(k, j) @ CNOT(i, j)
```
balopat, line 149: can you add circuit diagrams? they might be more readable

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,337 @@
...
+    for pauli_string in combined:
+        w = pauli_string.coefficient.real
+        qubit_idx = tuple(sorted(qubit_indices[qubit] for qubit in pauli_string.qubits))
+        hamiltonians[qubit_idx] = w
+
+    # Here we follow improvements of [4] cancelling out the CNOTs. The first step is to order by
+    # Gray code so that as few as possible gates are changed.
+    sorted_hs = sorted(list(hamiltonians.keys()), key=functools.cmp_to_key(_gray_code_comparator))
+
+    def _simplify_cnots(cnots):
```
balopat, line 141: type hints

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,337 @@
...
+                continue
+
+            # Here we apply the simplification of equation 11. Note that by flipping the control
+            # and target qubits, we have an equally valid identity:
+            # CNOT(i, j) @ CNOT(j, k) == CNOT(j, k) @ CNOT(i, k) @ CNOT(i, j)
+            # CNOT(j, i) @ CNOT(k, j) == CNOT(k, j) @ CNOT(k, i) @ CNOT(j, i)
+            for x, y in [(_control, _target), (_target, _control)]:
+                # We investigate potential pivots sequentially.
+                for j in range(1, len(cnots) - 1):
+                    # First, we look back for as long as the targets (resp. triggers) are the same.
```
balopat, line 181: What are triggers? Also, maybe it's just me but`repr.` is hard to parse for me:)...maybe avoid these abbreviations?

cirq-core/cirq/ops/boolean_hamiltonian.py
balopat, line 141: This method should be pulled out, there is no good reason to be nested. 
Also, it should be simplified/refactored, the three layers of nested loops is hard to understand.

balopat (changes requested):
Another round of reviews, mostly addressing readability as I'm going through the code.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: Done. Also added a description.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: Done.

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,337 @@
...
+    """Builds a circuit according to [1].
+
+    Args:
+        hamiltonian_polynomial_list: the list of Hamiltonians, typically built by calling
+            _build_hamiltonian_from_boolean().
+        qubit_map: map of string (boolean variable name) to qubit.
+        theta: A single float scaling the rotations.
+        ladder_target: Whether to use convention of figure 7a or 7b.
+
+    Yield:
```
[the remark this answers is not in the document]
tonybruguier: Done.

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,337 @@
...
+def _build_hamiltonian_from_boolean(
+    boolean_expr: Expr, qubit_map: Dict[str, 'cirq.Qid']
+) -> 'cirq.PauliString':
+    """Builds the Hamiltonian representation of Boolean expression as per [1]:
+
+    It is essentially a polynomial of Pauli Zs on different qubits. For example, this object could
+    represent the polynomial 0.5*I - 0.5*Z_1*Z_2 and it would be stored inside this object as:
+    self._hamiltonians = {(): 0.5, (1, 2): 0.5}.
+
+    While this object can represent any polynomial of Pauli Zs, in this file, it will be used to
```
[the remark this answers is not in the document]
tonybruguier: Done.

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,337 @@
...
+from sympy.core.symbol import Symbol
+import sympy.parsing.sympy_parser as sympy_parser
+
+import cirq
+from cirq import value
+from cirq.ops import raw_types
+
+
+def _Hamiltonian_O():
+    return 0.0 * cirq.PauliString({})
```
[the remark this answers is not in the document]
tonybruguier: I tried to make these constants, but I hit a circular import issue (I certainly hadn't foreseen that, it wasn't the reason I had chosen this approach!). I did try to solve it for a while, but it seems not trivial. Maybe a lambda throughout is better? I did rename the function as per your convention. Finally, I did add the type hints, but it seems to make the code less readable, so let me know if you don't want them after all.

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+def _Hamiltonian_I():
```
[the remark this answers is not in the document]
tonybruguier: Same as above.

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,337 @@
...
+    return cirq.PauliString({})
+
+
+def _Hamiltonian_Z(qubit):
```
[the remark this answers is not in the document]
tonybruguier: Same as above.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: Done.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: I meant control. Sorry about this. With the refactoring, there is no longer the comment.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: Split in three functions, with comments.

tonybruguier:
@balopat 
I tried to address the comments. I suspect the test errors are not related, but let me know if you think it is (I already pinged the chat). PTAL.

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,372 @@
...
+
+        previous_h = h
+
+    # Flush the last CNOTs.
+    yield _apply_cnots(previous_h, ())
+
+
+@value.value_equality
+class BooleanHamiltonian(raw_types.Operation):
+    """A gate that applies a Hamiltonian from a set of Boolean functions."""
```
balopat, line 297:
```suggestion
    """An operation that represents a Hamiltonian from a set of Boolean functions."""
```

cirq-core/cirq/ops/boolean_hamiltonian.py
balopat: Please address this comment.

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,337 @@
...
+        return 0
+    return _gray_code_comparator(k1[0:-1], k2[0:-1], not flip)
+
+
+def _get_gates_from_hamiltonians(
+    hamiltonian_polynomial_list: List['cirq.PauliSum'],
+    qubit_map: Dict[str, 'cirq.Qid'],
+    theta: float,
+    ladder_target: bool = False,
+):
```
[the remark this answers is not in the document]
balopat: Please address this comment.

examples/qaoa.py
balopat: Please address this comment.

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,372 @@
+"""
```
balopat, line 1: Add copyright notice

cirq-core/cirq/ops/boolean_hamiltonian_test.py
```diff
@@ -0,0 +1,151 @@
+import functools
```
balopat, line 1: Add copyright notice

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,372 @@
...
+        A Boolean that tells whether a simplification has been performed.
+        The CNOT list, potentially simplified.
+    """
+    x, y = (0, 1) if flip_control_and_target else (1, 0)
+
+    # We investigate potential pivots sequentially.
+    for j in range(1, len(cnots) - 1):
+        # First, we look back for as long as the targets (resp. controls) are the same.
+        # They all commute, so all are potential candidates for being simplified.
+        common_A: Dict[int, int] = {}
```
balopat, line 169: Please don't use capital letters in variable names

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,372 @@
...
+        for i in range(j - 1, -1, -1):
+            if cnots[i][y] != cnots[j][y]:
+                break
+            # We take a note of the control (resp. target).
+            common_A[cnots[i][x]] = i
+
+        # Next, we look forward for as long as the controls (resp. targets) are the
+        # same. They all commute, so all are potential candidates for being simplified.
+        common_B: Dict[int, int] = {}
```
balopat, line 178: Please don't use capital letters in variable names

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,372 @@
...
+            # Since we removed the pivot, the length should be one fewer.
+            assert len(new_idx) == len(cnots) - 1
+            cnots = [cnots[idx] for idx in new_idx]
+            return True, cnots
+
+    return False, cnots
+
+
+def _simplify_cnots(cnots: List[Tuple[int, int]]):
+    """
```
balopat, line 204: Nit no new line - the first line should be the short description of the function.

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,372 @@
...
+
+    Args:
+        cnots: A list of CNOTS, encoded as integer tuples (control, target).
+        flip_control_and_target: Whether to flip control and target.
+
+    Returns:
+        A Boolean that tells whether a simplification has been performed.
+        The CNOT list, potentially simplified.
+    """
+    x, y = (0, 1) if flip_control_and_target else (1, 0)
```
balopat, line 163:
```suggestion
    target, control = (0, 1) if flip_control_and_target else (1, 0)
```
pick variable names that describe their responsibility.

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,372 @@
...
+            # The controls (resp. targets) are the same, so we can simplify away.
+            cnots = [cnots[n] for n in range(len(cnots)) if n != j and n != k]
+            return True, cnots
+
+        qubit_to_index[cnots[j][y]] = j
+
+    return False, cnots
+
+
+def _simplify_cnots_tripplets(
```
balopat, line 143:
```suggestion
def _simplify_cnots_triplets(
```

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,372 @@
...
+) -> Tuple[bool, List[Tuple[int, int]]]:
+    """Simplifies CNOT pairs according to equations 9 and 10 of [4].Simplifies
+
+    CNOT(i, j) @ CNOT(k, j) = CNOT(k, j) @ CNOT(i, j)
+    ───@───────       ───────@───
+       │                     │
+    ───X───X───   =   ───X───X───
+           │             │
+    ───────@───       ───@───────
+
```
balopat, line 112: This is confusing: how come this is called `simplify_cnot_pairs`? This is a commutation relation, not a simplification. Can you clarify?

balopat (changes requested):
I would love to see unit tests for the `simplify_cnot` functions. They introduce complicated logic and I'm not convinced that they are well tested. Also, eventually we might want to pull these out as separate optimizers.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: Done.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: Done.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: Done.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: Changed name to `_simplify_commuting_cnots`

cirq-core/cirq/ops/boolean_hamiltonian.py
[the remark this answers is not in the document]
tonybruguier: Sorry about having missed this. Done, I think.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: Done.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: Done.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: Expanded the comment and added figures.

examples/qaoa.py
tonybruguier: Sorry about missing this. Added mypy types and comments.

tonybruguier:
Sorry about having missed some comments. I tried to take a careful look and not miss anything more.

I did add some unit tests for the simplification functions and I did find a bug! Thank you for suggesting this. The bug wasn't catastrophic, it was just missing some potential simplifications, so the results would have been the same, just with more gates than necessary. Nonetheless, thanks again.

PTAL.

tonybruguier:
@balopat 
Hi Balint
Do you have any more comments? It's not super urgent, but I might have a little time this weekend.

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,403 @@
...
+    """Builds a circuit according to [1].
+
+    Args:
+        hamiltonian_polynomial_list: the list of Hamiltonians, typically built by calling
+            _build_hamiltonian_from_boolean().
+        qubit_map: map of string (boolean variable name) to qubit.
+        theta: A single float scaling the rotations.
+        ladder_target: Whether to use convention of figure 7a or 7b of [4]. The two formulations
+            yield the same output, but can be simplified differently.
+            For example for 3 qubits, the ladder target would be:
```
balopat, line 256:
```suggestion
            For example for 3 qubits, the `ladder_target=True` results in:
```

balopat: Is my suggestion right? Wouldn't the other one look more like a ladder? :)

balopat: Also, can we move this explanation to the public constructor of `BooleanHamiltonian`? Then, we can just refer to it from here, the private method that: `ladder_target: has the same meaning as on BooleanHamiltonian`.

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,403 @@
...
+    """An operation that represents a Hamiltonian from a set of Boolean functions."""
+
+    def __init__(
+        self,
+        qubit_map: Dict[str, 'cirq.Qid'],
+        boolean_strs: Sequence[str],
+        theta: float,
+        ladder_target: bool,
+    ):
+        """
```
balopat, line 337: no new line

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,403 @@
...
+        Args:
+            boolean_strs: The list of Sympy-parsable Boolean expressions.
+            qubit_map: map of string (boolean variable name) to qubit.
+            theta: The list of thetas to scale the Hamiltonian.
+            ladder_target: Whether to use convention of figure 7a or 7b.
+        """
+        self._qubit_map: Dict[str, 'cirq.Qid'] = qubit_map
+        self._boolean_strs: Sequence[str] = boolean_strs
+        self._theta: float = theta
+        self._ladder_target: bool = ladder_target
```
balopat, line 362: Why is this important for the user to know?

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,403 @@
...
+        Boolean expression would be true iff the vertices on that are in different cuts (i.e. it's)
+        an XOR.
+
+        Then, we compute exp(j * theta * polynomial), which is unitary because the polynomial is
+        Hermitian.
+
+        Args:
+            boolean_strs: The list of Sympy-parsable Boolean expressions.
+            qubit_map: map of string (boolean variable name) to qubit.
+            theta: The list of thetas to scale the Hamiltonian.
```
balopat, line 356: if this is a list, why is the type `float`?

balopat (changes requested):
Another set of readability comments.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: Yes, you are right, the other one looked more like a ladder. The code was correct, but the two examples in the comment were inverted, so I flipped them.

Moved to the constructor and changed the other comments.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: Done.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: I think the authors said they would provide a bound for the number of gates in a follow-up paper, but I couldn't find one. I tried to compute it myself a bit (not for too long, though) but it doesn't seem obvious. I decided to add the option as an input instead, so that it can be decided experimentally.

I expanded the comment to reflect why the option is given.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: It was a bad comment. Changed.

examples/qaoa.py
```diff
@@ -1,155 +1,85 @@
-"""Runs the Quantum Approximate Optimization Algorithm on Max-Cut.
-
-=== EXAMPLE OUTPUT ===
-
-Example QAOA circuit:
```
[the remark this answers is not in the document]
tonybruguier: Done.

tonybruguier:
Thanks. PTAL.

95-martin-orion:
Pinging @balopat for re-review.

cirq-core/cirq/ops/boolean_hamiltonian.py
balopat: Why wouldn't we try both and go with the smaller one?

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,421 @@
...
+    CNOT(i, j) @ CNOT(k, j) = CNOT(k, j) @ CNOT(i, j)
+    ───@───────       ───────@───
+       │                     │
+    ───X───X───   =   ───X───X───
+           │             │
+    ───────@───       ───@───────
+
+    Args:
+        cnots: A list of CNOTS, encoded as integer tuples (control, target).
+        flip_control_and_target: Whether to flip control and target.
```
balopat, line 128: I had to look for a while before I understood the reason for the flip. 
I think this could be explained a bit better, added some info as a suggestion above. This decides which commutation relationship to use. 

I am not convinced that this is the best way to try out the commutation relationships.
I.e. one sweep with one then another sweep with the other direction? Is there a case where this method fails to simplify? I see this one for example: `[(3,2), (1,0), (1,2),(3,2)],False` - which should simplify to `[(1,0), (1,2)]`. 

 
 ```
    ──────────X──────────   
              │              
    ──────────@──@───────   
                 │        
    ───────X─────X──X────   
           │        │    
    ───────@────────@────  
      
```

File: cirq-core/cirq/ops/boolean_hamiltonian.py [hunk 69]
```diff
@@ -0,0 +1,421 @@
...
+from cirq.ops import raw_types
+
+_hamiltonian_O: Callable[[], 'cirq.PauliString'] = lambda: 0.0 * cirq.PauliString({})
+_hamiltonian_I: Callable[[], 'cirq.PauliString'] = lambda: cirq.PauliString({})
+_hamiltonian_Z: Callable[['cirq.Qid'], 'cirq.PauliString'] = lambda qubit: cirq.PauliString(
+    {qubit: cirq.Z}
+)
+
+
+def _build_hamiltonian_from_boolean(
```
balopat, line 47: I think this should be a separate public function in itself that could be split out from this PR. That will make reviewing easier as well.

cirq-core/cirq/ops/boolean_hamiltonian.py
balopat: Another thought: this feature (simplifying CNOT pairs after trying commutation relations) might be useful as a generic optimizer as well - another candidate for a separate PR.

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,421 @@
...
+
+    Returns:
+        A Boolean saying whether a simplification has been found.
+        The simplified list of CNOTs.
+    """
+
+    found_simplification = True
+    while found_simplification:
+        for simplify_fn, flip_control_and_target in itertools.product(
+            [_simplify_commuting_cnots, _simplify_cnots_triplets], [False, True]
```
balopat, line 232: again, I'm not convinced that this is the most efficient way of simplifications.

cirq-core/cirq/ops/boolean_hamiltonian.py
balopat: That comment 

> I think the authors said they would provide a bound for the number of gates in a follow-up paper, but I couldn't find one. 

From what I see in the paper that they had a remark on the proof that the commutation operations to simplify CNOTs lead to O(N') gates where N' < N Walsh operators:


 

They had a completely different remark regarding the "ladder target" as you call it: 

It is figure 3 and not 7 (as you mentioned in the comment): 



This tells me that we should not implement - only the 3(b) that they recommend...

balopat (changes requested):
We need more changes: 

1. I'd split out and make the boolean expr to paulistring a utility, that is useful in itself
2. the cnot simplifications are not handling some edge cases and I'm not convinced about their efficacy just yet
3. we can get rid of the ladder target argument I think

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,421 @@
...
+    Args:
+        cnots: A list of CNOTS, encoded as integer tuples (control, target).
+        flip_control_and_target: Whether to flip control and target.
+
+    Returns:
+        A Boolean that tells whether a simplification has been performed.
+        The CNOT list, potentially simplified.
+    """
+
+    # If input is empty, there is no simplification.
```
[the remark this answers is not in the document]
tonybruguier: Done.

cirq-core/cirq/ops/boolean_hamiltonian.py
```diff
@@ -0,0 +1,421 @@
...
+            yield cirq.Rz(rads=(theta * w)).on(qubits[h[-1]])
+
+        previous_h = h
+
+    # Flush the last CNOTs.
+    yield _apply_cnots(previous_h, ())
+
+
+@value.value_equality
+class BooleanHamiltonian(raw_types.Operation):
```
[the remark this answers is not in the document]
tonybruguier: Done.

tonybruguier:
Had a chat with Balint. Will spilt this in smaller PRs. Let's suspend this for now and I will ping once the other PRs are merged.

cirq-core/cirq/ops/boolean_hamiltonian.py
tonybruguier: Sent #4282

tonybruguier:
I had smaller #4282 reviewed and merged (thanks!), but maybe we continue splitting this further? Thus the present PR is still suspended and instead, smaller #4309 is ready for review?

tonybruguier:
I extracted the CNOT simplification into #4386 and the present PR now only touches the example and nothing else. PTAL. Not urgent at all.

balopat (changes requested):
I find the original example to be instructive as well and hence if we'd completely replace it, we would lose a bit of value. What do you think about adding this as an additional `boolean_qaoa.py` example and refer to each other as alternative implementations?

tonybruguier:
I am also OK with abandoning this PR. I don't want to be pushing too hard for something that is not helpful to the Cirq community. OK either way, what do you think?

balopat:
I think there is value in showing it off in an example, I just think that the original construction is instructive too. Maybe not a completely separate example then, but a separate method in the same example could be interesting - "direct compilation" vs "compilation using boolean expressions".

examples/qaoa.py
```diff
@@ -1,155 +1,162 @@
...
+  │     │     │      │
+  H     H     H      H
+  │     │     │      │
+  │     │     Y^-0.5 Y^-0.5
+  │     │     │      │
+  │     @─────@      │
+  │     │     │      │
+  │     │     Y^0.5  │
+  │     │     │      │
+  │     │     Rz(0)  │
```
balopat, line 14: I'm just noticing these Rz(0), Rx(0) gates - we should filter these out I think at least in the example by passing the circuit through a `cirq.DropNegligible().optimize_circuit(circuit)`.

tonybruguier:
Thanks @balopat . I think this is what you had suggested. PTAL?

tonybruguier:
@MichaelBroughton 
Would you be interested in taking this over from Balint?

tonybruguier:
@MichaelBroughton 
Friendly ping. Not super urgent and I can withdraw it if you have objections.

PR #3990, opened 2021-04-03
Allow expectation values for density matrix, MPS, Clifford

### LINKED ISSUE #3964: Density matrix simulator support for SimulatesExpectationValues

daxfohl:
Closes #3964.

More generically, this solves it for any `SimulatesFinalState` implementation whose `TSimulationTrialResult` type implements `expectation_from_state`.

### DIFF
--- a/cirq/sim/simulator.py
+++ b/cirq/sim/simulator.py
@@ -391,7 +391,6 @@ class SimulatesFinalState(
             )
         swept_evs = []
         qubit_order = ops.QubitOrder.as_qubit_order(qubit_order)
-        qmap = {q: i for i, q in enumerate(qubit_order.order_for(program.all_qubits()))}
         if not isinstance(observables, List):
             observables = [observables]
         pslist = [ops.PauliSum.wrap(pslike) for pslike in observables]
--- a/cirq/sim/simulator.py
+++ b/cirq/sim/simulator.py
@@ -302,7 +302,9 @@ class SimulatesExpectationValues(metaclass=abc.ABCMeta):
         """
 
 
-class SimulatesFinalState(Generic[TSimulationTrialResult], metaclass=abc.ABCMeta):
+class SimulatesFinalState(
+    Generic[TSimulationTrialResult], SimulatesExpectationValues, metaclass=abc.ABCMeta
+):
     """Simulator that allows access to the simulator's final state.
 
     Implementors of this interface should implement the simulate_sweep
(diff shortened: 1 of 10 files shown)


daxfohl:
@zaqqwerty do you have a unit test I could use to check this?

daxfohl:
@tonybruguier is there a quick way to add this to MPS as well?

daxfohl:
@95-martin-orion is it possible to do the same for Clifford simulators?

daxfohl:
@95-martin-orion oops, just saw you already had a fix. Up to you whether you'd rather use this one that generifies things or stick with the yagni approach.

tonybruguier:
> @tonybruguier is there a quick way to add this to MPS as well? I see there is a state_vector field available on the trial result class. Could we add the same two lines there as I did here in state_vector_simulator and expect that to produce valid results?

I am not sure what is required. Do you want the simulator to take a density matrix as initial state and return a density matrix at the output of the circuit? That could be involved to write.

Or do you want the simulator to just output a density matrix?

daxfohl:
> I am not sure what is required. Do you want the simulator to take a density matrix as initial state and return a density matrix at the output of the circuit? That could be involved to write.

I think I figured it out. Just needed to add a way to get an expectation value from MPSTrialResult, which turned out to be a one-liner. Took the tests from density matrix simulator tests (stolen from @95-martin-orion's PR), and changed them to use MPS, and they seem to pass.

daxfohl:
Looks like Clifford also provides a function to generate a state vector from a trial result, so we can add an expectation value there too. Will add that before merging.

daxfohl:
Clifford seems to be working now too.

cirq/sim/simulator.py
```diff
@@ -372,6 +374,33 @@ def simulate_sweep(
...
+        swept_evs = []
+        qubit_order = ops.QubitOrder.as_qubit_order(qubit_order)
+        if not isinstance(observables, List):
+            observables = [observables]
+        pslist = [ops.PauliSum.wrap(pslike) for pslike in observables]
+        for param_resolver in study.to_resolvers(params):
+            result = self.simulate(
+                program, param_resolver, qubit_order=qubit_order, initial_state=initial_state
+            )
+            swept_evs.append([result.expectation_from_state(obs) for obs in pslist])
```
95-martin-orion, line 401: My concern here: the fact that `expectation_from_state` has no default implementation means that `SimulatesFullState` does not fully implement `SimulatesExpectationValues`, which means user-created subclasses will have a broken API.

We could update the class docstring to say that users must implement `expectation_from_state` in their result class, but that's kind of unwieldy for a class whose base purpose does not include calculating expectation values. This is part of why I mentioned needing a `SimulatesEVsAndFullState` type for this in #3979.

95-martin-orion (changes requested):
Regardless of how this pans out, I think we need to document the Simulator hierarchy somewhere (maybe at the top of `simulator.py`?) once it's done. Multiple nested base classes make it tricky to track which one supports what behavior.

daxfohl:
@95-martin-orion I agree. I think go ahead and merge your solution for density matrix. I'll look at how to improve this as time comes available.

95-martin-orion:
Closing as superseded by #3979. There is still interest in deduplicating this behavior (#4209), but at this point it's likely easier to do so in a fresh PR. (Feel free to reopen if that's not the case.)

PR #3993, opened 2021-04-05
Add edges to device

### LINKED ISSUE #3696: Provide a way to iterate through the edges of a device

I am working on the 2 qubit interaction visualization and wanted to generate random characterization data for each of the edges of the Sycamore chip. 

There is no way to retrieve a list of the edges on the chip - despite the `cirq.google.Sycamore` representation printing out the diagram nicely.

**Describe alternatives/workarounds you've considered**

 As currently I'm only working on an example, I will just rely on a copy-pasta from the `SerializableDevice.__str__` function: 

```
            # Find pairs that are connected by two-qubit gates.
            Pair = Tuple['cirq.GridQubit', 'cirq.GridQubit']
            pairs = {
                cast(Pair, pair)
                for gate_defs in self.gate_definitions.values()
                for gate_def in gate_defs
                if gate_def.number_of_qubits == 2
                for pair in gate_def.target_set
                if len(pair) == 2
            }
```

But this is far from ideal. It would be much better to have something like `cirq.google.Sycamore.edges()` or something similar. 

**Additional context (e.g. screenshots)**

Maybe the generic GraphDevice is the answer to this as well? 

(because I have a workaround at the moment)

daxfohl:
Closes #3696


cirq/pasqal/pasqal_device.py
```diff
@@ -361,6 +362,22 @@ def _value_equality_values_(self) -> Any:
     def _json_dict_(self) -> Dict[str, Any]:
         return cirq.protocols.obj_to_dict_helper(self, ['control_radius', 'qubits'])
 
+    @property
+    def edges(self) -> List[Tuple['cirq.Qid', 'cirq.Qid']]:
+        """Returns a list of qubit edges on the device.
+
+        Returns:
+            All qubit pairs that are separated by a distance of 1 in a single
+            dimension.
```
HGSilveri, line 371: I'd change this to something like:
```suggestion
            All qubit pairs that are less than the control radius apart.
```

HGSilveri (approved):
As far as `PasqalDevice` is concerned, this LGTM. Thanks for the enhancement!

cirq/devices/device.py
```diff
@@ -46,6 +48,36 @@ def qubit_set(self) -> Optional[AbstractSet['cirq.Qid']]:
         # Default to the qubits being unknown.
         return None
 
+    @property
+    def edges(self) -> Optional[List[Tuple['cirq.Qid', 'cirq.Qid']]]:
```
maffoo, line 52: I'd suggest typing this as `Optional[AbstractSet[Tuple['cirq.Qid', 'cirq.Qid']]]`, by analogy with `qubit_set`. In addition, we might consider changing the name to something more specific than "edges", for example `qubit_pairs` or `coupled_pairs`.

Note that in some internal code we have a property like this, but it's a persistent source of bugs that tuples are not symmetric: `(q0, q1) != (q1, q0)`. This can lead to missing coupled pairs if you forget to sort the qubits in a candidate pair before checking for it in the collection returned here. Tuples are certainly convenient, but it might be worth considering whether there is a way to avoid this issue, e.g. use our own "pair" type which is symmetric, or return a custom collection that sorts before checking membership, or just leave as-is but document that pairs here are sorted.

daxfohl: Yes, I was thinking of making the pair a frozenset instead of a tuple. So the return type would be optional(frozenset(frozenset(qid))).

Actually, maybe a custom Pair type would be better, a thin wrapper around frozen set that guarantees only two elements. I'll add that before merging.

daxfohl:
Updated the function name to `edge_set()` because there was a conflicting `edges` property in `graph_device`. There is already a `qubit_set()` function, so following that precedent with the new name.

cirq/devices/device.py
```diff
@@ -105,3 +142,27 @@ def can_add_operation_into_moment(
             Whether or not the moment will validate after adding the operation.
         """
         return not moment.operates_on(operation.qubits)
+
+
+@value.value_equality
+class QidPair:
```
mpharrigan, line 148: do we need this class? A tuple would do just fine

mpharrigan (commented):
Not all devices have (only) two qubit gates. How would this work for devices that support multi-qubit operations?

mpharrigan:
+1 using a name other than "edges"

cirq/devices/device.py
mpharrigan: I guess this is good in the case where there's no directionality to the qubit pair, which again is an assumption about how the device works :)

daxfohl:
What would be better here? The tuple implementation I had first is error prone because tuples are not symmetric. The implementation with the custom QidPair is symmetric but then the question was raised about whether that's a valid assumption for all devices.

Should we go with a third option, which is back to using tuples, but have both orderings be emitted instead of just one? That would double the output size but allow for directional edges if needed.

Per 3+ qubit gates, I don't think this is where that functionality would go. Once you have the two gate edges, you can feed those into some external function to extract n-gate sequences. I'd say that should be external to the device code. Unless there's something specific about devices and what edges are n-qubit capable?

mpharrigan:
I don't think any of my questions should be seen as blockers. Just curious if you'd thought about any of them or had any thoughts.

> Per 3+ qubit gates, I don't think this is where that functionality would go. Once you have the two gate edges, you can feed those into some external function to extract n-gate sequences. I'd say that should be external to the device code. Unless there's something specific about devices and what edges are n-qubit capable?

Yeah fair. Of course you can always dream up something crazy like if three qubits could do a toffli but you couldn't do a two-qubit interaction between any two of them.

`QidPair` is a nice addition for order-agnostic pairs. I suspect the main problem will be that there are many locations in the code where we technically *should* be using such a class but instead use a two-tuple. 

For something actionable: I do think the name should be `qubit_pairs` and not anything to do with edges.

cirq/devices/device.py
mpharrigan: This class would be a lot more usable if it supported unpacking, maybe by sorting the qubits first.

```python
qa, qb = pair
```

daxfohl:
Sure, Balint said he wanted to take a look first and think about it, since
he opened the original issue. So I'll let this hang out here until then. No
rush.

daxfohl:
@balopat ping to look at this whenever you have time, no rush. I'm by no means attached to this PR so if you want to change it completely, or table it until some other things are settled just lmk.

95-martin-orion:
Paging @balopat for review.

cirq-core/cirq/devices/device.py
```diff
@@ -105,3 +142,27 @@ def can_add_operation_into_moment(
             Whether or not the moment will validate after adding the operation.
         """
         return not moment.operates_on(operation.qubits)
+
+
+@value.value_equality
+class QidPair:
```
balopat, line 148: I'd name this `SymmetricalQidPair`

cirq-core/cirq/pasqal/pasqal_device.py
```diff
@@ -361,6 +362,22 @@ def _value_equality_values_(self) -> Any:
     def _json_dict_(self) -> Dict[str, Any]:
         return cirq.protocols.obj_to_dict_helper(self, ['control_radius', 'qubits'])
 
+    def edge_set(self) -> FrozenSet['cirq.QidPair']:
```
balopat, line 365: Let's rename this to `qid_pairs` to align with the internal terminology.

cirq-core/cirq/devices/device.py
```diff
@@ -46,6 +50,39 @@ def qubit_set(self) -> Optional[AbstractSet['cirq.Qid']]:
...
+        if all(isinstance(q, _BaseGridQid) for q in qs):
+            return frozenset(
+                [
+                    QidPair(q, q2)
+                    for q in [cast(_BaseGridQid, q) for q in qs]
+                    for q2 in [q + (0, 1), q + (1, 0)]
+                    if q2 in qs
+                ]
+            )
+        if all(isinstance(q, _BaseLineQid) for q in qs):
```
balopat, line 80: This part potentially will need refactoring when we break out different Device classes e.g. `GridDevice`, `LineDevice`, etc. But for now this should be fine.

balopat (changes requested):
Sorry for making you wait on this for so long. I'm fine with merging this, however some parts might change as the Device strategy gets sorted out.

cirq/devices/device.py
daxfohl: done

balopat (approved):

Issue #3767, opened 2021-02-09
Numpy 1.20 includes typing information, breaks mypy checks locally


kevinsung:
**Description of the issue**
mypy encounters errors.

**How to reproduce the issue**
In a fresh virtual environment, install Cirq and the dev tools (mypy version 0.782). Then execute
```
./check/mypy
```
This gives the following errors:
```
cirq/linalg/tolerance.py:31: error: Incompatible return value type (got "bool_", expected "bool")
cirq/linalg/tolerance.py:31: error: Argument 1 to "__call__" of "ufunc" has incompatible type "Union[float, complex, Iterable[float], ndarray]"; expected "Union[Union[int, float, complex, str, bytes, generic], Sequence[Union[int, float, complex, str, bytes, generic]], Sequence[Sequence[Any]], _SupportsArray]"
cirq/linalg/tolerance.py:45: error: Incompatible return value type (got "bool_", expected "bool")
cirq/linalg/combinators.py:157: error: Argument 1 to "reduce" has incompatible type "Callable[[Type[number[Any]], Type[number[Any]]], Type[number[Any]]]"; expected "Callable[[dtype[Any], dtype[Any]], dtype[Any]]"
cirq/interop/quirk/cells/parse.py:79: error: Incompatible return value type (got "float", expected "str")
cirq/google/optimizers/two_qubit_gates/math_utils.py:62: error: Item Module of "Union[Module, Any]" has no attribute "rand"
cirq/google/optimizers/two_qubit_gates/math_utils.py:63: error: Item Module of "Union[Module, Any]" has no attribute "rand"
cirq/google/optimizers/two_qubit_gates/math_utils.py:64: error: Item Module of "Union[Module, Any]" has no attribute "rand"
cirq/google/optimizers/two_qubit_gates/math_utils.py:70: error: Item Module of "Union[Module, Any]" has no attribute "rand"
cirq/experiments/purity_estimation.py:62: error: Incompatible return value type (got "number[Any]", expected "float")
cirq/linalg/states.py:53: error: Function "numpy.array" is not valid as a type
cirq/linalg/states.py:53: note: Perhaps you need "Callable[...]" or a callback protocol?
cirq/work/pauli_sum_collector.py:81: error: Unsupported operand types for % ("number[Any]" and "int")
cirq/work/pauli_sum_collector.py:81: note: Left operand is of type "Union[number[Any], ndarray]"
[...]

PR #3995, opened 2021-04-06
Fix numpy mypy and add mypy-next

mpharrigan:
Towards #3767


balopat (changes requested):
I'd prefer having the next type check always exit 0. 
I really don't love the comment based tagging, can we just add another line that greps on numpy?

mpharrigan:
kk ptal @balopat

balopat (approved):
LGTM

PR #3997, opened 2021-04-06
Fix most numpy type errors in cirq/ops

mpharrigan:
Using `check/mypy --next | grep cirq/ops` this fixes almost all of the issues raised with simple modifications.

There is a looming problem with places where we use `numbers.Complex`, which is a nice generalization for Union[int, float, complex] but does not play nice with numpy type information.

Towards #3767

### DIFF
--- a/cirq/ops/diagonal_gate.py
+++ b/cirq/ops/diagonal_gate.py
@@ -18,7 +18,7 @@ The gate is used to create a (2^n)x(2^n) matrix with the diagonal elements
 passed as a list.
 """
 
-from typing import AbstractSet, Any, Tuple, Iterator, List, Sequence, TYPE_CHECKING, Union
+from typing import AbstractSet, Any, Tuple, Iterator, List, Sequence, TYPE_CHECKING, Union, Optional
 import numpy as np
 import sympy
 
@@ -30,7 +30,7 @@ if TYPE_CHECKING:
     import cirq
 
 
-def _fast_walsh_hadamard_transform(a: Tuple[Any, ...]) -> np.array:
+def _fast_walsh_hadamard_transform(a: Tuple[Any, ...]) -> np.ndarray:
     """Fast Walsh–Hadamard Transform of an array."""
     h = 1
     a_ = np.array(a)
@@ -99,7 +99,7 @@ class DiagonalGate(raw_types.Gate):
     def _has_unitary_(self) -> bool:
         return not self._is_parameterized_()
 
-    def _unitary_(self) -> np.ndarray:
+    def _unitary_(self) -> Optional[np.ndarray]:
         if self._is_parameterized_():
             return None
         return np.diag([np.exp(1j * angle) for angle in self._diag_angles_radians])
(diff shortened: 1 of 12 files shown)


cirq/ops/diagonal_gate.py
```diff
@@ -18,7 +18,7 @@
 passed as a list.
 """
 
-from typing import AbstractSet, Any, Tuple, Iterator, List, Sequence, TYPE_CHECKING, Union
+from typing import AbstractSet, Any, Tuple, Iterator, List, Sequence, TYPE_CHECKING, Union, Optional
```
viathor, line 21: Optional: Should we keep alphabetic order?

cirq/ops/raw_types.py
```diff
@@ -552,7 +552,7 @@ def _commutes_(
 
         m12 = protocols.unitary_protocol.unitary(circuit12, default=None)
         m21 = protocols.unitary_protocol.unitary(circuit21, default=None)
-        if m12 is None:
+        if m12 is None or m21 is None:
```
viathor, line 555: Looks like a bug. Consider adding a unit test for it.

viathor (approved):

cirq/ops/diagonal_gate.py
mpharrigan: it wasn't alphabetic?

cirq/ops/raw_types.py
mpharrigan: Since the circuits are constructed from the same operations (but in different orders) they would either both admit unitaries or neither, so it's not technically a bug. In any event, the type checker caught it now.

cirq/ops/diagonal_gate.py
viathor: Nope, see "Tuple". Surprised pylint didn't catch it.

viathor (approved):

cirq/ops/diagonal_gate.py
viathor: Indeed. I should be more precise.

cirq/ops/pauli_interaction_gate.py
```diff
@@ -112,14 +112,16 @@ def _decompose_(self, qubits: Sequence['cirq.Qid']) -> 'cirq.OP_TREE':
...
         self, args: 'cirq.CircuitDiagramInfoArgs'
     ) -> 'cirq.CircuitDiagramInfo':
-        labels = cast(
-            Dict[pauli_gates.Pauli, np.ndarray],
-            {pauli_gates.X: 'X', pauli_gates.Y: 'Y', pauli_gates.Z: '@'},
-        )
+        labels: Dict['cirq.Pauli', str] = {
+            pauli_gates.X: 'X',
+            pauli_gates.Y: 'Y',
+            pauli_gates.Z: '@',
```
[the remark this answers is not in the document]
viathor: I suppose it is to make the diagram resemble the standard quantum circuit symbol for controlled-Z.

PR #4000, opened 2021-04-06
Fix most numpy type errors in cirq/linalg

mpharrigan:
Using `check/mypy --next | grep cirq/linalg` this fixes all the
problems.

#3767

### DIFF
--- a/cirq/linalg/decompositions.py
+++ b/cirq/linalg/decompositions.py
@@ -29,6 +29,7 @@ from typing import (
     TYPE_CHECKING,
     TypeVar,
     Union,
+    cast,
 )
 
 import matplotlib.pyplot as plt
@@ -121,7 +122,7 @@ def _group_similar(items: List[T], comparer: Callable[[T, T], bool]) -> List[Lis
 
 def unitary_eig(
     matrix: np.ndarray, check_preconditions: bool = True, atol: float = 1e-8
-) -> Tuple[np.array, np.ndarray]:
+) -> Tuple[np.ndarray, np.ndarray]:
     """Gives the guaranteed unitary eigendecomposition of a normal matrix.
 
     All hermitian and unitary matrices are normal matrices. This method was
@@ -249,7 +250,7 @@ def so4_to_magic_su2s(
         if mat.shape != (4, 4) or not predicates.is_special_orthogonal(mat, atol=atol, rtol=rtol):
             raise ValueError('mat must be 4x4 special orthogonal.')
 
-    ab = combinators.dot(MAGIC, mat, MAGIC_CONJ_T)
+    ab = cast(np.ndarray, combinators.dot(MAGIC, mat, MAGIC_CONJ_T))
     _, a, b = kron_factor_4x4_to_2x2s(ab)
 
     return a, b
@@ -512,12 +513,15 @@ class KakDecomposition:
         y_mat = np.array([[0, -1j], [1j, 0]])
         z_mat = np.array([[1, 0], [0, -1]])
 
-        return self.global_phase * combinators.dot(
-            after,
-            interaction_matrix(z_mat, z),
-            interaction_matrix(y_mat, y),
-            interaction_matrix(x_mat, x),
-            before,
+        return self.global_phase * cast(
+            np.ndarray,
+            combinators.dot(
+                after,
+                interaction_matrix(z_mat, z),
+                interaction_matrix(y_mat, y),
+                interaction_matrix(x_mat, x),
+                before,
+            ),
         )
 
     def _decompose_(self, qubits):
@@ -648,11 +652,13 @@ def scatter_plot_normalized_kak_interaction_coefficients(
 
     # parse input and extract KAK vector
     if not isinstance(interactions, np.ndarray):
-        interactions = [
+        interactions_np: Union[List[np.ndarray], np.ndarray] = [
             a if isinstance(a, np.ndarray) else protocols.unitary(a) for a in interactions
         ]
+    else:
+        interactions_np = interactions
 
-    points = kak_vector(interactions) * 4 / np.pi
+    points = kak_vector(interactions_np) * 4 / np.pi
 
     ax.scatter(*coord_transform(points), **kwargs)
     ax.set_xlim(0, +1)
--- a/cirq/linalg/decompositions.py
+++ b/cirq/linalg/decompositions.py
@@ -652,13 +652,13 @@ def scatter_plot_normalized_kak_interaction_coefficients(
 
     # parse input and extract KAK vector
     if not isinstance(interactions, np.ndarray):
-        interactions_np: Union[List[np.ndarray], np.ndarray] = [
+        interactions_extracted: List[np.ndarray] = [
             a if isinstance(a, np.ndarray) else protocols.unitary(a) for a in interactions
         ]
     else:
-        interactions_np = interactions
+        interactions_extracted = [interactions]
 
-    points = kak_vector(interactions_np) * 4 / np.pi
+    points = kak_vector(interactions_extracted) * 4 / np.pi
 
     ax.scatter(*coord_transform(points), **kwargs)
     ax.set_xlim(0, +1)
--- a/cirq/linalg/decompositions.py
+++ b/cirq/linalg/decompositions.py
@@ -338,7 +338,7 @@ class AxisAngleDecomposition:
 
     def __str__(self) -> str:
         axis_terms = '+'.join(
-            '{:.3g}*{}'.format(e, a) if e < 0.9999 else a
+            f'{e:.3g}*{a}' if e < 0.9999 else a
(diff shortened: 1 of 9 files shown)


cirq/linalg/decompositions.py
```diff
@@ -249,7 +250,7 @@ def so4_to_magic_su2s(
         if mat.shape != (4, 4) or not predicates.is_special_orthogonal(mat, atol=atol, rtol=rtol):
             raise ValueError('mat must be 4x4 special orthogonal.')
 
-    ab = combinators.dot(MAGIC, mat, MAGIC_CONJ_T)
+    ab = cast(np.ndarray, combinators.dot(MAGIC, mat, MAGIC_CONJ_T))
```
viathor, line 253: Can we avoid the cast if we turn `combinators.dot` into a [function overload](https://mypy.readthedocs.io/en/stable/more_types.html#function-overloading) (so that mypy can infer return type from argument types)?

cirq/linalg/decompositions.py
```diff
@@ -512,12 +513,15 @@ def interaction_matrix(m: np.ndarray, c: float) -> np.ndarray:
         y_mat = np.array([[0, -1j], [1j, 0]])
         z_mat = np.array([[1, 0], [0, -1]])
 
-        return self.global_phase * combinators.dot(
-            after,
-            interaction_matrix(z_mat, z),
-            interaction_matrix(y_mat, y),
-            interaction_matrix(x_mat, x),
-            before,
+        return self.global_phase * cast(
```
viathor, line 516: ditto

(Possibly could apply to other use of `cast` below, too.)

cirq/linalg/tolerance.py
```diff
@@ -14,14 +14,15 @@
 
 """Utility for testing approximate equality of matrices and scalars within
 tolerances."""
-from typing import Union, Iterable
+from typing import Union, Iterable, TYPE_CHECKING
```
viathor, line 17: ordering

cirq/linalg/transformations.py
```diff
@@ -14,7 +14,7 @@
 
 """Utility methods for transforming matrices or vectors."""
 
-from typing import Tuple, Optional, Sequence, List, Union, TypeVar
+from typing import Tuple, Optional, Sequence, List, Union
```
viathor, line 17: ordering

cirq/linalg/transformations.py
```diff
@@ -326,6 +324,10 @@ def partial_trace(tensor: np.ndarray, keep_indices: List[int]) -> np.ndarray:
     return np.einsum(tensor, left_indices + right_indices)
 
 
+class UnfactorizableStateVectorError(ValueError):
```
viathor, line 327: AFAICT, this is raised when the input state $|\psi_{AB}\rangle$ viewed as a state of two subsystems A and B (each possibly consisting of multiple qubits) is entangled (i.e. not a product state or IOW if $|\psi_{AB}\rangle \ne |\phi_A\rangle\otimes|\gamma_B\rangle$).

A better name for this exception is "EntangledStateError". Also, docstring would be nice.

cirq/linalg/transformations.py
```diff
@@ -471,7 +477,7 @@ def sub_state_vector(
     if default is not RaiseValueErrorIfNotProvided:
         return default
 
-    raise ValueError(
+    raise UnfactorizableStateVectorError(
```
viathor, line 480: Need to update Raises section of the function docstring above.

viathor (changes requested):

cirq/linalg/decompositions.py
mpharrigan: In the current codebase, we could get away with changing cominators.dot to be ndarray -> ndarray, but this would be a change to the function signature (although not the behavior). I've changed it to be ArrayLike -> ndarray which changes the behavior for non-array-inputs slightly by returning the numpy version of each scalar instead of the python version.

This should only cause problems in exceptional circumstances (like if someone is using dot on only scalars and checking the result with isinstance) and like I said there's no evidence in the code-base where dot is called with anything other than arrays.

cirq/linalg/decompositions.py
mpharrigan: changed this per above. what other use of cast?

cirq/linalg/transformations.py
mpharrigan: done

cirq/linalg/transformations.py
mpharrigan: good point

cirq/linalg/transformations.py
```diff
@@ -326,6 +324,14 @@ def partial_trace(tensor: np.ndarray, keep_indices: List[int]) -> np.ndarray:
     return np.einsum(tensor, left_indices + right_indices)
 
 
+class EntangledStateError(ValueError):
+    """Raised by `sub_state_error` if the result of factoring is not a pure state
```
viathor, line 328: "not a pure state" -> "not a product state"

Also, I'd remove "by `sub_state_error`" since the exception may be used in future by related functions.

Suggestion:

```python
"""Raised when a product state is expected, but an entangled state is provided."""
```

cirq/linalg/diagonalize.py
```diff
@@ -218,15 +218,15 @@ def bidiagonalize_real_matrix_pair_with_symmetric_products(
     # Merge the fixup factors into the initial diagonalization.
     left_adjust = combinators.block_diag(overlap_adjust, extra_left_adjust)
     right_adjust = combinators.block_diag(overlap_adjust.T, extra_right_adjust)
-    left = left_adjust.T.dot(base_left.T)
-    right = base_right.T.dot(right_adjust.T)
+    left = cast(np.ndarray, left_adjust.T.dot(base_left.T))
+    right = cast(np.ndarray, base_right.T.dot(right_adjust.T))
```
mpharrigan, line 222: So only the cast on `left` is necessary and I'm trying to figure out why ......

viathor (approved):

cirq/linalg/decompositions.py
viathor: Well, there are six more left at this point. It'd be nice to get rid of as many as you can. They may hide bugs and make code harder to read.

cirq/linalg/decompositions.py
viathor: SGTM

cirq/linalg/diagonalize.py
mpharrigan: it seems like in the general case

```python
x: np.ndarray
y: np.ndarray
x.dot(y) --> Union[number[Any], ndarray]
```

but 

```python
np.dot(x, y) --> ndarray
```

weird

mpharrigan: the former is defined in the top-level `__init__.pyi` type stub file. The latter may be due to the docstring of the skeleton implementation in numpy/core/multiarray.py (that dispatches to c ufuncs). Maybe an oversight on numpy's part? I still don't know why `right` doesn't need a cast in the above.

mpharrigan: I think the former is the correct signature but the latter is the easier signature. I've re-written this function to use `cirq.dot` and `np.dot` to keep everything as `ndarray`s

cirq/linalg/transformations.py
mpharrigan: Changed to use your nice wording

cirq/linalg/diagonalize.py
```diff
@@ -183,9 +183,9 @@ def bidiagonalize_real_matrix_pair_with_symmetric_products(
             raise ValueError('mat1 must be real.')
         if np.any(np.imag(mat2) != 0):
             raise ValueError('mat2 must be real.')
-        if not predicates.is_hermitian(mat1.dot(mat2.T), rtol=rtol, atol=atol):
+        if not predicates.is_hermitian(cast(np.ndarray, mat1.dot(mat2.T)), rtol=rtol, atol=atol):
             raise ValueError('mat1 @ mat2.T must be symmetric.')
-        if not predicates.is_hermitian(mat1.T.dot(mat2), rtol=rtol, atol=atol):
+        if not predicates.is_hermitian(cast(np.ndarray, mat1.T.dot(mat2)), rtol=rtol, atol=atol):
```
viathor, line 188: Optional: These are the last two casts remaining. It'd be great if we could get rid of them, too.

viathor (approved):
LGTM! Nice work finding a way to get rid of type casts!

cirq/linalg/diagonalize.py
mpharrigan: got rid of 'em!

We'll have to keep an eye out if the numpy folks change the signature for `np.dot` vs. `ndarray.dot`. It strikes me as odd that they're different.

mpharrigan:
test fails on CI and one of my python environments but not the other 🤷 

Changing back to `np.dot` rather than `@`

PR #4001, opened 2021-04-07
Fix valid PAULI_GATE_LIKE error listing 0,1,2,3 repeatedly instead of actual allowed values

Strilanc:


vtomole (commented):
Please add a test that checks for this fix.

balopat (commented):
Please add tests.

balopat:
ping @Strilanc - please add tests and a PR description!

dabacon:
Fixed in #5213

PR #4002, opened 2021-04-07
Fix most numpy errors in cirq/qis and friends

mpharrigan:
Part of #3767

### DIFF
--- a/cirq/qis/measures.py
+++ b/cirq/qis/measures.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 """Measures on and between quantum states and operations."""
 
-from typing import Optional, TYPE_CHECKING, Tuple
+from typing import Optional, TYPE_CHECKING, Tuple, cast
 
 import numpy as np
 import scipy
@@ -138,9 +138,17 @@ def fidelity(
         )
     state1 = quantum_state(state1, qid_shape=qid_shape, validate=validate, atol=atol)
     state2 = quantum_state(state2, qid_shape=qid_shape, validate=validate, atol=atol)
-    state1 = state1.density_matrix() if state1._is_density_matrix() else state1.state_vector()
-    state2 = state2.density_matrix() if state2._is_density_matrix() else state2.state_vector()
-    return _fidelity_state_vectors_or_density_matrices(state1, state2)
+    state1_arr = (
+        state1.density_matrix()
+        if state1._is_density_matrix()
+        else cast(np.ndarray, state1.state_vector())
+    )
+    state2_arr = (
+        state2.density_matrix()
+        if state2._is_density_matrix()
+        else cast(np.ndarray, state2.state_vector())
+    )
+    return _fidelity_state_vectors_or_density_matrices(state1_arr, state2_arr)
 
 
 def _numpy_arrays_to_state_vectors_or_density_matrices(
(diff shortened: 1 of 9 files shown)


cirq/qis/measures.py
```diff
@@ -138,9 +138,17 @@ def fidelity(
...
-    return _fidelity_state_vectors_or_density_matrices(state1, state2)
+    state1_arr = (
+        state1.density_matrix()
+        if state1._is_density_matrix()
+        else cast(np.ndarray, state1.state_vector())
+    )
+    state2_arr = (
+        state2.density_matrix()
+        if state2._is_density_matrix()
+        else cast(np.ndarray, state2.state_vector())
```
viathor, line 149: Optional: IIUC this is just to convert from `Optional[np.ndarray]` to `np.ndarray`. It'd be nice if you could make it more apparent to the reader. Perhaps an `assert` is understood by both mypy and humans?

cirq/qis/states.py
```diff
@@ -135,7 +136,7 @@ def density_matrix(self) -> np.ndarray:
         (a two-dimensional array).
         """
         if not self._is_density_matrix():
-            state_vector = self.state_vector()
+            state_vector = cast(np.ndarray, self.state_vector())
```
viathor, line 139: ditto (optional as above, please address neither or both for consistency)

cirq/qis/states.py
```diff
@@ -159,7 +160,10 @@ def validate(
         is_state_tensor = self.data.shape == self.qid_shape
         if is_state_vector or is_state_tensor:
             validate_normalized_state_vector(
-                self.state_vector(), qid_shape=self.qid_shape, dtype=dtype, atol=atol
+                cast(np.ndarray, self.state_vector()),
```
viathor, line 163: ditto (...)

cirq/testing/consistent_qasm.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import warnings
-from typing import Any, List, Sequence
+from typing import Any, List, Sequence, Optional
```
viathor, line 15: ordering

viathor (approved):
LGTM, thanks!

cirq/qis/measures.py
mpharrigan: assert would work but not in this ternary operation

I found this thread when going through these: https://github.com/python/typing/issues/645

honestly, I think these code patterns indicate that there's a missing method. Either "`state_vector_for_sure_because_i_already_checked`" or `array_form_irrespective_of_whether_its_density_or_state`. I'll investigate

mpharrigan: Otherwise the cast is safe since the only time `state_vector` returns none is when `_is_density_matrix` is true. Obviously mypy isn't smart enough to deduce that full logic

viathor: I agree with you that there is a missing function to aid this pattern. Perhaps something like Rust's [unwrap](https://doc.rust-lang.org/rust-by-example/error/option_unwrap.html)? This would have the benefit of being useful for all `Optional` types, not just state vectors. Ideally, we'd use it similar to how you're using `cast` here:

```python
state2_arr = (
        state2.density_matrix()
        if state2._is_density_matrix()
        else unwrap(state2.state_vector())
```

WDYT?

cirq/testing/consistent_qasm.py
mpharrigan: on the basis that none of the imports are ordered in cirq and we don't have tooling to check for or support these, I'm going to not do any alphabetizing of imports :)

cirq/qis/measures.py
viathor: I like it.

nit: maybe `state_or_density_array` -> `state_vector_or_density_matrix`. The latter is clearer (especially since `state_tensor` is also a thing so `state` alone in this context is ambiguous) and more consistent with other method names.

cirq/testing/consistent_qasm.py
viathor: Fair enough in general.

Maybe at least keep the alphabetic order where it pre-exists, e.g. here, and do whatever elsewhere?

cirq/qis/measures.py
mpharrigan: sure. I was trying to think of names that weren't *too* long but I guess this is a somewhat niche function so it can handle a long name.

cirq/testing/consistent_qasm.py
mpharrigan: discussed elsewhere, I'm going to hold out for tooling to do this for me

PR #4004, opened 2021-04-08
Attempt to Fix PauliSumCollector.collect_async warning under notebook(short-term solution)

ybc1991:
As mentioned in #3654:
> `PauliSumCollector.collect` which uses asyncio doesn't always mix well with notebooks - throws RuntimeError: This event loop is already running or a more mysterious papermill error (KeyError: 'language_info') - but collect_async seems to work fine

The current status is using collect_async. However, in the website, it shows the warning like:
image

Also, the Energy 0.0 is not desired value, which is because the collector is not finished.

The problem mainly is because notebook itself is running under the asyncio. Importing `nest-asyncio`  maybe is the easist fix.

*Note: `nest-asyncio` seems working well in most cases, but in the colab environment, `run all` can cause the cell of async function hang forever. Fortunately, it can be avoided by run cell one-by-one.*


rmlarose:
> Also, the Energy 0.0 is not desired value, which is because the collector is not finished.

Oh, I didn't realize this value was incorrect - thanks for this. I was mistakenly content with CI passing after a long battle, mea culpa. (Note the papermill error was totally unrelated.)

docs/operators_and_observables.ipynb
```diff
@@ -789,7 +803,7 @@
     "collector = cirq.PauliSumCollector(circuit, psum, samples_per_term=10_000)\n",
     "\n",
     "# Provide a sampler. See also: collector.collect(...).\n",
-    "collector.collect_async(sampler=cirq.Simulator())\n",
+    "collector.collect(sampler=cirq.Simulator())\n",
```
95-martin-orion, line 806: I tried to run this cell, but it's taking an unreasonably long time to complete (5+ minutes). To minimize the impact on test times, we should reduce `samples_per_term` to a number that allows this to complete in 30 seconds or less.

If this requires choosing an unrealistically low value for `samples_per_term`, please also add a comment explaining why we're using the new value.

95-martin-orion (changes requested):

docs/operators_and_observables.ipynb
ybc1991: That is strange. It is super quick to run that. In google colab that line takes less than 300ms. 
image
In the github action, the whole notebook execution is like one and half minutes.
image

I wonder why it takes you so long time to complete. (I guess maybe related to live lock?) Could you show your test environment?

95-martin-orion: I'm running this directly from the [Colab link](https://colab.sandbox.google.com/github/quantumlib/Cirq/blob/c6cbbfc73fc92297a5708926ffb42795d73a0b5e/docs/operators_and_observables.ipynb) (modified to point to the latest commit on this PR) and the execution consistently hangs or crashes - I haven't had a run complete successfully yet.

Logs include `IndexError: pop from an empty deque`, but it's unclear whether that's coming from the Colab restart or from asyncio.

95-martin-orion: For clarity, I've been running the linked colab with the `Runtime > Run all` command.

ybc1991: Interesting. Yes, if I click `Run all`, it will fail. But if click `Run before` that Cell and then manually `shift+Enter` run that cell can be finished quickly. I guess `Run all` added another layer of running environment caused this. I will take a look to understand this better.

95-martin-orion: > Interesting. Yes, if I click `Run all`, it will fail. But if click `Run before` that Cell and then manually `shift+Enter` run that cell can be finished quickly.

I can reproduce this result - `Run before` and then running this cell also succeeds for me. Thanks for digging into this!

ybc1991: I still do not understand why this happens. What I found is the [TensorFlow Federated colab](https://www.tensorflow.org/federated/tutorials/federated_learning_for_image_classification#before_we_start) has encountered similar asynchrony issue and they used nest_asyncio as well for solution. I tried to use `Run all` their colab, and they do not have similar issue.

95-martin-orion:
@ybc1991, have you been able to find any solution for this? Based on #4197 it appears that the issue is still occurring.

If we have no way to resolve this from our end (e.g. if it's caused by an issue with nest_asyncio that has no workaround), I'd consider moving forwards with the PR with clear warning messages on the affected cell noting that "Run all" will misbehave.

ybc1991:
> @ybc1991, have you been able to find any solution for this? Based on #4197 it appears that the issue is still occurring.
> 
> If we have no way to resolve this from our end (e.g. if it's caused by an issue with nest_asyncio that has no workaround), I'd consider moving forwards with the PR with clear warning messages on the affected cell noting that "Run all" will misbehave.

I don't find the solution to fix it. I do test on other environment such as local jupyter notebook and the note book in vscode. Both can finish correctly with run-all button. So as suggested, I added a warning cell there. PTAL.

docs/operators_and_observables.ipynb
```diff
@@ -779,6 +779,29 @@
...
+    "nest_asyncio.apply()   "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "5d8e48230d73"
+   },
+   "source": [
+    "**Warning: If you run this notebook in colab environment with \"Run all\", the process may hang forever or you may encounter crash. Please manually run this cell either press the play button to the left of the code, or use the keyboard shortcut \"Command/Ctrl+Enter\".**"
```
95-martin-orion, line 802: A few nits:
- `in colab environment` -> `in a colab environment`
- `either press...or use` -> `either by pressing...or using`
- Instead of `the process`, I recommend saying `the cell below` to be abundantly clear which cell is the source of the issue. (As it is, it's unclear if this refers to the cell above or below)

ybc1991: Done.

95-martin-orion (approved):
Thanks for this. Would you mind either commenting on #4004 or opening a new issue to track the "run all" dilemma here? I'm happy to merge this PR as-is, I just want to keep tabs on this for the future.

ybc1991:
Sure, I have updated the title and content of this PR. To point out this is the short-term solution and the problem of "run all" in the top comments.

95-martin-orion:
The error is an instance of #4293, unrelated to this PR.

balopat:
> The error is an instance of #4293, unrelated to this PR.

I reran the checks, that seems to push the situation further. This is a nasty flake from npm, as I'm looking into it, retries might be the only way to get over it until they fix it upstream.

PR #4009, opened 2021-04-08
Use duet for asynchrony, in particular for Collector

maffoo:
We are planning to switch to [duet](https://github.com/google/duet) for supporting async code, with one of the reasons being that it is "reentrant" and supports nested invocations of the event loop (`duet.run`). We don't yet have much async code in cirq, but do use it for `Collector` and `PauliSumCollector` and this can lead to problems if running in a context where the asyncio event loop is already running (#4004). The solution proposed there is to patch asyncio to make it reentrant, using the `nest-asyncio` library, but this seems like something we should not rely on. Instead, I propose that we make the switch to duet.

The main question I have is whether we need to continue to support `Collector.collect_async` and `Sampler.*_async` methods being asyncio coroutines, or whether we can just switch them to be duet coroutines. In this PR as it stands I've just switched things, but if we want to support asyncio we would need to name the duet coroutine functions differently and then wrap them into an asyncio-compatible form for the existing methods.


95-martin-orion:
Please fix the failing tests, then re-request review.

maffoo:
Hm, we don't have a version of duet for 3.6, and it'd be quite hard to do because of some asnyc/await language features that landed in 3.7. @balopat, what is the plan for dropping python 3.6 support?

balopat:
This is blocked on #3574, which is blocked on google3 having their default python runtime upgraded to 3.7. 
If we want to move forward with duet for the OSS community (I think that this is great!) then we have to think about workarounds until that upgrade is done which might take a year still. A workaround design might mean to offer the current async flow on 3.6 and duet on 3.7 somehow.

balopat:
> This is blocked on #3574, which is blocked on google3 having their default python runtime upgraded to 3.7.
> If we want to move forward with duet for the OSS community (I think that this is great!) then we have to think about workarounds until that upgrade is done which might take a year still. A workaround design might mean to offer the current async flow on 3.6 and duet on 3.7 somehow.

One idea: 

We could add a cirq-duet module that would use duet for the collector. 
People who have python 3.7+ could use that module, the ones using 3.6 can still use the old collector, on which we could add even a deprecation warning...

WDYT, @maffoo?

mpharrigan:
Do we have an update on timeline for Python 3.7+?

maffoo:
Actually, @fedimser is working on adding python 3.6 support to duet (https://github.com/google/duet/pull/23) which will let us move forward with these async changes without waiting for 3.7 support in colab etc.

maffoo:
This is ready for review. @balopat, @mpharrigan

cirq-core/cirq/conftest.py
```diff
@@ -13,6 +13,6 @@ def pytest_pyfunc_call(pyfuncitem):
     if inspect.iscoroutinefunction(pyfuncitem._obj):
         # coverage: ignore
         raise ValueError(
-            f'{pyfuncitem._obj.__name__} is async but not '
-            f'decorated with "@pytest.mark.asyncio".'
+            f'{pyfuncitem._obj.__name__} is a bare async function. '
+            f'Should be decorated with "@duet.sync".'
```
mpharrigan, line 17: nit: It should

cirq-core/cirq/testing/__init__.py
```diff
@@ -14,10 +14,6 @@
 
 """Utilities for testing code."""
 
-from cirq.testing.asynchronous import (
-    asyncio_pending,
```
mpharrigan, line 18: This is technically a breaking change. I suspect no one externally has been using our async testing helper code? but we should make a note of this for the release

cirq-core/cirq/work/collector.py
```diff
@@ -125,15 +130,12 @@ def collect(
...
-
-        See Also:
-            Python 3 documentation "Coroutines and Tasks"
-            https://docs.python.org/3/library/asyncio-task.html
         """
-        return asyncio.get_event_loop().run_until_complete(
-            self.collect_async(
-                sampler, concurrency=concurrency, max_total_samples=max_total_samples
-            )
+        return duet.run(
```
mpharrigan, line 134: lovely

cirq-core/cirq/work/collector.py
```diff
@@ -163,53 +165,56 @@ async def collect_async(
         Returns:
             The collector's result after all desired samples have been
             collected.
-
-        See Also:
-            Python 3 documentation "Coroutines and Tasks"
-            https://docs.python.org/3/library/asyncio-task.html
         """
-        pool = work_pool.CompletionOrderedAsyncWorkPool()
+        results = duet.AsyncCollector[Tuple[CircuitSampleJob, 'cirq.Result']]()
```
mpharrigan, line 169: what is going on here? Is this a form of type annotation I've never seen before or does this actually change the object that has been created?

cirq-core/cirq/work/collector.py
```diff
@@ -163,53 +165,56 @@ async def collect_async(
...
-            return job, await sampler.run_async(job.circuit, repetitions=job.repetitions)
+        async def run_job(job):
+            nonlocal job_error
+            try:
+                result = await sampler.run_async(job.circuit, repetitions=job.repetitions)
+            except Exception as error:
+                if not job_error:
+                    results.error(error)
+                    job_error = error
+            else:
```
mpharrigan, line 183: what is try .. else? why not put results.add inside the try?

cirq-core/cirq/work/work_pool.py
```diff
@@ -1,91 +0,0 @@
...
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
-import asyncio
-import collections
-from typing import Optional, Awaitable, Union
-
-
-class CompletionOrderedAsyncWorkPool:
```
mpharrigan, line 20: this seems like something someone may be relying on in cirq user land?

mpharrigan (commented):
I think the actual code changes in this PR are very straightforward. I asked some clarifying questions and posed some nits.

cirq-core/cirq/work/collector.py
maffoo: The `else` clause executes if the `try` block does not raise any exceptions. I wanted to make clear that the `except` clause is _only_ catching exceptions from the await call. If `results.add` were part of the `try` clause then the except would also handle errors from that; `results.add` shouldn't raise errors, but this is just to be safe.

cirq-core/cirq/work/work_pool.py
maffoo: It's certainly possible. I can restore this and deprecate it instead.

cirq-core/cirq/testing/__init__.py
maffoo: Brought this back since it's needed to test the work pool, which is now deprecated but not removed.

cirq-core/cirq/work/collector.py
mpharrigan: ok, glad it's just a type hint. my slight preference would be for the type annotation at the cost of some repetition but I defer to you

cirq-core/cirq/work/collector.py
mpharrigan: thanks for explaining

mpharrigan (approved):

cirq-core/cirq/work/collector.py
maffoo: Switched to type annotation. It still _just_ fits on one line :-)

dstrain115 (approved):
In general, it might be good to put more line comments in where duet is used, since it will be unfamiliar to the majority of developers, but I don't have any other objections to pulling this in.

PR #4013, opened 2021-04-09
[POC] ActOnArgs.join(other: ActOnArgs) provides random circuit speedup

daxfohl:
(This is based off the ActOnArgs branch, which hasn't been merged yet. The diff is far smaller here: https://github.com/daxfohl/Cirq/compare/spliterate...daxfohl:join.)

We add an abstract `ActOnArgs.join(self: TSelf, other: TSelf) -> TSelf` that allows joining two quantum states of the same type into a combined quantum state. This allows a simulation strategy that evaluates unentangled states independently, and join them only when a gate entangles them. This simulation strategy works generically for any simulator whose TActOnArgs implements `join`, which is usually straightforward. This POC shows it for state vector and density matrix ActOnArgs.

In the new example, I create a random circuit and simulate it with the standard `simulate_moment_steps`, and again with the new method. I see anywhere between 5 and 100x speedups, depending on how entangled the random circuit is.

Also updated the teleportation example to use `join`, as a smaller example of what this POC enables. Initially the bell state and the message state are distinct, and then they're joined only when Alice has to perform the measurements.

Eventually we could offer a `split` that could be performed after a detangling operation such as reset or measure, though that may be more complex and less frequently valuable.

Is this a direction we may want to go?


daxfohl:
Surprisingly this even shows a small (~50%) speedup on the MPS simulator. cc @tonybruguier

PR #4015, opened 2021-04-12
Improve plot_state_histogram configuration options.

tanujkhattar:
This PR introduces the following changes
- Make `tick_labels` (and other plot properties) configurable so that users can provide different labels
- Changes default tick labels from binary string ('0111') to integers ('7') to make larger plots more readable. 
- Add support for `collections.Counter` so users can directly pass the result of `result.histogram()` so that processed results can be plotted easily. 

Once this is checked-in, I'll send a PR for state_histogram.ipynb tutorial.


cirq/vis/state_histogram.py
```diff
@@ -50,32 +50,54 @@ def get_state_histogram(result: 'result.Result') -> np.ndarray:
...
 
 def plot_state_histogram(
-    values: Union['result.Result', np.ndarray], ax: Optional['plt.Axis'] = None
+    data: Union['result.Result', collections.Counter, Sequence[SupportsFloat]],
+    ax: Optional['plt.Axis'] = None,
+    *,
+    tick_label: Optional[Sequence[str]] = None,
+    xlabel: Optional[str] = 'qubit state',
+    ylabel: Optional[str] = 'result count',
+    title: Optional[str] = 'Result State Histogram',
```
95-martin-orion, line 59: Do we want to allow `None` for these fields? If not, these shouldn't have type `Optional`.

95-martin-orion (changes requested):
Mostly LGTM, just a couple details to clear up.

cirq/vis/state_histogram.py
```diff
@@ -50,32 +50,54 @@ def get_state_histogram(result: 'result.Result') -> np.ndarray:
...
-        ax:     The Axes to plot on. If not given, a new figure is created,
-                plotted on, and shown.
+        data:   The histogram values to plot. Possible options are:
+                `result.Result`: Histogram is computed using
+                `get_state_histogram` and all 2 ** num_qubits values are
+                 plotted, including 0s.
+                `collections.Counter`: Only (key, value) pairs present in
+                 collection are plotted.
+                `Sequence[SupportsFloat]`: Values in input are plotted with
+                 default labels as |0> ... |n-1>.
```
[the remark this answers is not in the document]
tanujkhattar: done.

cirq/vis/state_histogram.py
tanujkhattar: Yes, these fields can potentially be none. For eg, we might want to omit the title when drawing a figure to be included in a pdf. 
We also follow the same pattern for other visualization utilities, for eg: `integrated_histogram` 
https://github.com/quantumlib/Cirq/blob/bee1a448bc98b52ab952ccf284c0382918abaad1/cirq/vis/histogram.py#L32

95-martin-orion (approved):

PR #4016, opened 2021-04-12
Better import for sometimes failing mypy test

mpharrigan:
Would get:

dev_tools/bash_scripts_test.py:33: error: Name
'_pytest.tmpdir.TempdirFactory' is not defined

This fixes it. Don't know why it sometimes works on
some configurations.


dev_tools/bash_scripts_test.py
```diff
@@ -18,7 +18,7 @@
 from dev_tools import shell_tools
 
 if TYPE_CHECKING:
-    import _pytest
+    import _pytest.tmpdir
```
95-martin-orion, line 21: Do you have a sense for _why_ this works? I don't really understand the root problem, even - my inuition is that `import _pytest` should give access to `_pytest.tmpdir.TempdirFactory` in the original code.

mpharrigan: In python, `import package` does not give you availability to `package.subpackage`. But sometimes people like us put imports into the top-level `__init__.py`.

mpharrigan: maybe there are different versions of pytest (`_pytest`?) around.

balopat: The `_pytest` package does not have `tmpdir` attribute: https://github.com/pytest-dev/pytest/blob/main/src/_pytest/__init__.py

This should never work in mypy...but sometimes it does! :) 

```
>>> import _pytest
>>> _pytest.tmpdir
  File "<stdin>", line 1, in <module>
AttributeError: module '_pytest' has no attribute 'tmpdir'
>>> import _pytest.tmpdir
>>> _pytest.tmpdir
<module '_pytest.tmpdir' from '/Users/balintp/.virtualenvs/cirq37/lib/python3.7/site-packages/_pytest/tmpdir.py'>

```

I can confirm that this fixes the mypy issue on my 3.7 mac env where I could consistently reproduce the error report.

balopat (approved):
Thanks for the fix!!!

dev_tools/bash_scripts_test.py
mpharrigan: `_pytest.*` is listed in mypy.ini as `follow_imports=silent` and `ignore_missing_imports=true`. I don't want to actually do any sleuthing but it sounds relevant and could have been a corner case whose behavior has changed between versions.

mpharrigan:
This change got reverted somehow

95-martin-orion:
> This change got reverted somehow

Not only that, this file got reverted all the way to its version from December...

balopat:
Hmmm...my "automated rebasing" strategy was error prone, this must have been overridden then. Let me fix that.

Issue #4017, opened 2021-04-12
Make heatmaps more user friendly by supporting config params in heatmap.plot()


tanujkhattar:
Currently, the heatmaps config parameters need to be passed during construction and can be updated by calling `heatmap.update_config`. This should be changed s.t. users can directly pass the parameters to `heatmap.plot()` function. Eg:
- Current workflow
```python
heatmap = calibration.heatmap("parallel_p11_error")
heatmap.update_config(vmin=0, vmax=0.1)
heatmap.plot()
```
- Expected Workflow
```python
heatmap = calibration.heatmap("parallel_p11_error")
heatmap.plot(vmin=0, vmax=0.1)
```

**Acceptance criteria - when is the task considered done?**
- Modify the heatmap code to support the updated workflow.
- Update visualization -> heatmap tutorial and add sections on how to use various config parameters. 

Part of the larger roadmap item: #3642

rmlarose:
@tanujkhattar I think this should be closed now?

tanujkhattar:
That's right, this should have been closed by #4026

PR #4018, opened 2021-04-12
Support passing config params in heatmap.plot()

tanujkhattar:
- Part-1 for #4017
- Will send another PR to update the tutorial once this is checked in. 

Breaking Change Release Note: `heatmap.plot()` now takes all valid heatmap config arguments as `**kwargs` instead of just `**collection_options`, which is now a nested config argument (dict). Please see the tests / tutorials for more details.


cirq/vis/heatmap.py
```diff
@@ -268,21 +271,24 @@ def _plot_on_axis(self, ax: plt.Axes, **collection_options: Any) -> mpl_collecti
...
+                See __init__ for more details on the allowed arguments.
         Returns:
             A 2-tuple ``(ax, collection)``. ``ax`` is the `plt.Axes` that
             is plotted on. ``collection`` is the collection of paths drawn and filled.
         """
         show_plot = not ax
         if not ax:
             fig, ax = plt.subplots(figsize=(8, 8))
-        collection = self._plot_on_axis(ax, **collection_options)
+        self.update_config(**kwargs)
```
rmlarose, line 290: Will this make it so that future calls to `plot` have the same settings? e.g.,

```python
heatmap.plot(vmin=0, vmax=0.1)

heatmap.plot()  # Also has vmin=0, vmax=0.1?
```



If so, should there be a "local config" each time `plot` is called?

tanujkhattar: I think that makes sense. Updated the code to ensure that plot doesn't modify the internal config.

rmlarose (approved):
LGTM

cirq/vis/heatmap.py
```diff
@@ -268,23 +272,28 @@ def _plot_on_axis(self, ax: plt.Axes, **collection_options: Any) -> mpl_collecti
...
         show_plot = not ax
         if not ax:
             fig, ax = plt.subplots(figsize=(8, 8))
-        collection = self._plot_on_axis(ax, **collection_options)
+        original_config = copy.deepcopy(self._config)
+        self.update_config(**kwargs)
+        collection = self._plot_on_axis(ax)
         if show_plot:
             fig.show()
+        self._config = copy.deepcopy(original_config)
```
mpharrigan, line 296: why do you need deepcopy twice? why do you need deepcopy at all?

tanujkhattar: - I need a deepcopy once because I want to store the original state / dict (before `plot` was called) so that I can undo the changes applied by `update_config` call at the end of the function call to `plot`. 
```python
def plot(...):
  original_state = self._config  # If I don't do a deepcopy, this stores a reference to the self._config dict and hence I won't be able to undo the changes without maintaining additional information. 
  self.update_config(**kwargs) # this modifies the self._config state, which i wish to undo at the end of function call. 
  ........ 
  self._config = original_config # I think I can get rid of the deepcopy here
```

tanujkhattar: Removed the second deepcopy.

cirq/vis/heatmap.py
```diff
@@ -268,23 +272,28 @@ def _plot_on_axis(self, ax: plt.Axes, **collection_options: Any) -> mpl_collecti
...
 
     def plot(
-        self, ax: Optional[plt.Axes] = None, **collection_options: Any
+        self, ax: Optional[plt.Axes] = None, **kwargs: Any
     ) -> Tuple[plt.Axes, mpl_collections.Collection]:
         """Plots the heatmap on the given Axes.
         Args:
             ax: the Axes to plot on. If not given, a new figure is created,
                 plotted on, and shown.
-            collection_options: keyword arguments passed to mcoll.PolyCollection().
```
balopat, line 277: this is a breaking change technically, as `collection_options` was part of `Heatmap.plot()` in the previous releases as well. Given that we anticipate low usage on this though, I'm fine with it, but let's add a "breaking change" release note to the description of the PR and add the breaking change label.

balopat (approved):
LGTM, but let's take a note of the breaking change.

cirq/vis/heatmap.py
tanujkhattar: done.

Issue #4012, opened 2021-04-09
temp env creation bug in isolated notebook tests


balopat:
This came up on the cirq.google extraction PR.

```
/tmp/pytest-of-runner/quantum_volume/lib/python3.7/site-packages/networkx/utils/decorators.py in _random_state(func, *args, **kwargs)
    398             raise nx.NetworkXError("random_state_index must be an integer") from e
    399         except IndexError as e:
--> 400             raise nx.NetworkXError("random_state_index is incorrect") from e
    401 
    402         # Create a numpy.random.RandomState instance

NetworkXError: random_state_index is incorrect


=========================== short test summary info ============================
FAILED dev_tools/notebooks/isolated_notebook_test.py::test_notebooks_against_released_cirq[/home/runner/work/Cirq/Cirq/cirq-core/cirq/contrib/quimb/Cirq-to-Tensor-Networks.ipynb]
FAILED dev_tools/notebooks/isolated_notebook_test.py::test_notebooks_against_released_cirq[/home/runner/work/Cirq/Cirq/cirq-core/cirq/contrib/quimb/Contract-a-Grid-Circuit.ipynb]
FAILED dev_tools/notebooks/isolated_notebook_test.py::test_notebooks_against_released_cirq[/home/runner/work/Cirq/Cirq/examples/advanced/quantum_volume.ipynb]
=================== 3 failed, 9 passed in 1312.45s (0:21:52) ===================
Error: Process completed with exit code 1.

```

**Root cause analysis** 

I didn't figure out the exact issue yet, but it seems that the environment creation in isolated_notebook_test.py is broken somehow: while `/tmp/pytest-of-runner/proto_dir` (the template env for the environments to copy) has `decorator 5.0.6`, however somehow the virtual envs for the specific tests - that were supposed to be the clones of this template env - have `decorator 4.4.2` which is problematic, and that's why [this issue is triggered](https://stackoverflow.com/questions/66922359/unexpected-error-while-drawing-networkx-graph). e.g. `/tmp/pytest-of-runner/quantum_volume`. 

I suspect that something is wrong with `virtualenv-clone` but not sure what yet.

I'm making this issue as a reminder to fix this behavior separately from the cirq.google extraction PR.

PR #4019, opened 2021-04-13
fix networkx / decorator dependency issue

balopat:
Fixes https://github.com/quantumlib/Cirq/issues/4012 which was caused by https://github.com/networkx/networkx/issues/4718. 

Also ensures that isolated notebook test runs for all notebooks (not just changed ones) if the test itself changes.

### DIFF
--- a/dev_tools/notebooks/isolated_notebook_test.py
+++ b/dev_tools/notebooks/isolated_notebook_test.py
@@ -25,6 +25,7 @@ import os
 import subprocess
 import sys
 import warnings
+from logging import warning
 from typing import Set
 
 import pytest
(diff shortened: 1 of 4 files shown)


dev_tools/notebooks/isolated_notebook_test.py
```diff
@@ -90,10 +95,22 @@ def _find_base_revision():
     raise ValueError("Can't find a base revision to compare the files with.")
 
 
+def _list_all_notebooks() -> Set[str]:
+    try:
+        output = subprocess.check_output(['git', 'ls-files', '*.ipynb'])
+        return set(output.decode('utf-8').splitlines())
+    except subprocess.CalledProcessError as ex:
+        warning("It seems that tests are run from not a git repo, notebook tests are skipped", ex)
```
viathor, line 103: nit: misplaced "not", also s/run/running. Suggestion: "It seems that tests are not running in a git repo, skipping notebook tests"

viathor (approved):

dev_tools/notebooks/isolated_notebook_test.py
balopat: OK, as this was duplicated across the two test files, I decided to factor them out into a `utils.py` instead of modifying this sentence in both places. It grew the scope but it was the right thing to do. Can you PTAL again? Thanks!

balopat: I'm going to go ahead and merge this :) thanks for the review. Please let me know if you find any other nits!

PR #4028, opened 2021-04-16
Allow using plain qubits as keys for single-qubit heatmaps

maffoo:
It's a bit annoying that mypy seems to get confused unless I explicitly overload the `__init__` type signature. If I just try to type `value_map` as a `Union` type, typechecks fail at callsites.

Also cleaned up a few types while I was in there, in particular we are passing around polygons as lists of (x, y) coords, not as matplotlib `Polygon` objects.


balopat (approved):
LGTM, after rebase

PR #4031, opened 2021-04-19
Include py.typed files in cirq-core and cirq-google

maffoo:
These are needed to enable type-checking for users who install cirq-core and cirq-google.


balopat (approved):
Thank you!

PR #4032, opened 2021-04-19
Narrow pytest execution check in deprecation warning

balopat:
Instead of depending on `PYTEST_CURRENT_TEST` environment variable introduces `CIRQ_TESTING` to ensure that only cirq tests will see this env var. Otherwise downstream projects will fail when using deprecated cirq objects in their tests.


maffoo:
FWIW I like having deprecation warnings get turned into errors when using cirq in downstream tests, because that serves as a signal to update our usage of deprecated cirq idioms. I guess it make sense to narrow the mechanism here, but we'll probably just end up re-enabling it on our end :-)

(Other users may want to do the same thing so we should document how this mechanism works in the cirq testing docs.)

cirq-core/cirq/conftest.py
```diff
@@ -1,6 +1,12 @@
 import inspect
-
+import os
 import matplotlib.pyplot as plt
+import pytest
+
+
+@pytest.fixture(scope="session", autouse=True)
+def setup_cirq_testing_envvar():
+    os.environ['CIRQ_TESTING'] = "true"
```
maffoo, line 9: Can we put this line in `pytest_configure` instead? Also, will this get picked up if running tests on other modules such as `cirq-google`?

balopat: It does get picked up by cirq-google.
I have to check if `pytest_configure` works or not - I wanted the setting before the tests execute but _after_ the test collection, to allow for collection code to use deprecated modules - for example "module crawlers" like the json_serialization_test.py list all public objects, and during that access `cirq.google`, which triggers the error mechanism. Maybe I should just turn it off during collection for those specific cases?

balopat:
> FWIW I like having deprecation warnings get turned into errors when using cirq in downstream tests, because that serves as a signal to update our usage of deprecated cirq idioms. I guess it make sense to narrow the mechanism here, but we'll probably just end up re-enabling it on our end :-)
> 
> (Other users may want to do the same thing so we should document how this mechanism works in the cirq testing docs.)

Oh, interesting angle - I will definitely document it - I can add that to this PR. 
I'd still prefer that to be opt-in instead of opt-out, to allow smoother upgrades.

cirq-core/cirq/conftest.py
maffoo: I think using `assert_deprecated` around the module discovery would make sence, since it explicitly documents where we are using deprecated behavior. That seems preferable to relying on specific execution order of pytest_configure versus a session fixture.

balopat: Will do that!

maffoo (approved):
LGTM

PR #4033, opened 2021-04-20
fix versioning of new modules

balopat:
Fixes a bug in the version test and the actual _version.py for cirq-google.


dev_tools/version_test.py
```diff
@@ -19,14 +20,16 @@ def test_versions_are_the_same():
     packages = ["cirq-google/cirq_google"]
 
     for p in packages:
+        print(p)
```
maffoo, line 23: nit: remove print statement

maffoo (approved):
LGTM

dev_tools/version_test.py
balopat: done!

Issue #3898, opened 2021-03-10
Deprecation of abstract/protocol methods


balopat:
We need a better way to deprecate abstract methods.

Context (#3860): 

@balopat  This ends up being a breaking change though for any simulator that was overriding `_simulator_iterator`. This is why I had left the call to `_simulator_iterator` there in the original PR #3650. Otherwise I'd have just inlined and deleted `_simulator_iterator` outright: it's a private method and it's unused elsewhere.

It doesn't sound like anyone has been broken by this, so maybe it's okay. (We can just delete `_simulator_iterator` in that case; it's not called by anyone). I think there's still an unmet need for cleanly deprecate abstract methods, or methods that were intended to be overridden. The base class has to continue calling them until the deadline or it will break child classes. I like the error-by-default, but maybe there's a way to silence these errors when retaining the call to a deprecated method is required, like here.

_Originally posted by @daxfohl in https://github.com/quantumlib/Cirq/pull/3860#discussion_r587578684_

viathor:
This issue should spell out in what way the present deprecation mechanism is insufficient.

balopat:
Discussed on Cirq Cynque: 

- First of all, we should clarify underscore usage  `_method` should be considered private. 
One suggestion: 
```
public
_private
_cirq_protocol_
__python_protocol__
_protected_protected
```

- Second, deprecation seems like a good idea to have for these.

daxfohl:
I can't remember all the details but In retrospect I *think* I could have done this with some reflection in the calling methods. We could delete the method in the base class, and then the callers could do "if old method exists in the subclass then call it else call the new one" (or maybe vice-versa). Regardless, I think this isn't quite concrete enough of a problem to give a concrete answer for, and probably needs to be resolved on a case-by-case basis. I'd suggest closing the issue. @viathor @MichaelBroughton

PR #4035, opened 2021-04-20
Eliminate simulator boilerplate by pushing iteration into base class

### LINKED ISSUE #2178: Reduce required boilerplate for implementing external simulators and samplers

daxfohl:
Fixes #2178

Pushes iteration and run repetitions into simulator base class. Simulators only have to implement _create_act_on_args, and _create_step_result, and the base class handles the rest.

After the migration to ActOnArgs, all simulators had nearly identical code for iteration and repetitions. The few places where they differed were either bugs (state vector had a bug where noise was not accounted for in terminal measurement gates), or missed optimization opportunities (MPS and Clifford did not split out unitary prefixes when running iterations). This PR consolidates and dedupes that logic. It also lays the foundation for future improvements, such as the ActOnArgs.join/split/move we've discussed separately, to be implemented once and inherited by all simulators.


daxfohl:
@tonybruguier at some point I'll figure out how to rebase in a way that doesn't pull in that MPS change, but for now, can you consent again please?

tonybruguier:
@googlebot I consent.

95-martin-orion:
Self-assigning review to follow up on ActOnArgs PR chain.

cirq-core/cirq/sim/simulator.py
```diff
@@ -560,11 +589,11 @@ def _create_act_on_args(
         """
         raise NotImplementedError()
 
-    @abc.abstractmethod
     def _core_iterator(
```
95-martin-orion, line 592: [No change needed] There's some ongoing discussion around naming of "protected" methods in #3898 - nothing concrete yet, but it would be good to keep an eye on that issue.

cirq-core/cirq/sim/simulator.py
```diff
@@ -601,6 +673,67 @@ def _create_simulator_trial_result(
...
+                sim_state=act_on_args.copy(),
+            )
+            for step_result in all_step_results:
+                for k, v in step_result.measurements.items():
+                    if k not in measurements:
+                        measurements[k] = []
+                    measurements[k].append(np.array(v, dtype=np.uint8))
+        return {k: np.array(v) for k, v in measurements.items()}
+
+    def _can_be_in_run_prefix(self, val: Any):
```
95-martin-orion, line 734: Please add a docstring for this method, since it also appears outside of this class.

95-martin-orion (changes requested):
Just a few comments for the first pass here, mainly focused on clarifying the Simulator API. Specifically, we want to make sure users can easily implement the simulator interfaces; as such, it's critical that we have clear documentation on how to do so.

daxfohl:
@95-martin-orion This is ready again.

95-martin-orion (changes requested):
I think the docstrings helped me solidify my concerns about this PR - specifically, that it's blurring the line between the interface and the implementation. I agree with #2178 that there ought to be a class which captures this boilerplate code, but that class shouldn't be `SimulatesIntermediateState`; instead, it should be a new superclass to the Cirq simulators. Users could then subclass that new class if they want similar behavior, or implement one of the interfaces for different behavior.

How to translate that into this PR:
- Add a new simulator class which implements `SimulatesIntermediateState`, `SimulatesSamples`, etc.
    - This class should probably have its own file and tests, which could allow us to dedupe test code as well.
    - In a later PR, this class might also be able to implement `SimulatesExpectationValues`, similar to #3990.
- Migrate the implementation code from this PR into that new class.
- Modify existing simulators to subclass the new class instead of directly implementing the simulator interface(s).

daxfohl:
This sounds like a great plan. I will start on that and send an updated
version.

cirq-core/cirq/sim/simulation_engine.py
```diff
@@ -0,0 +1,262 @@
+# Copyright 2018 The Cirq Developers
```
daxfohl, line 1: Go through the whole file and look at docstrings etc.

daxfohl:
@95-martin-orion This looks much better, thanks for the idea. I still have to add the tests, but the change in design helps highlight that this is new functionality that can and should be tested independently. Much nicer.

daxfohl:
@95-martin-orion this is ready for review again.

Another nice thing about the approach is that now with the test module's `CountingSimulator` you can see exactly the bare minimum required to implement a simulator ... and I think it's still too much. IMO we should move toward deprecating the step results and trial results, and just yield the ActOnArgs themselves, as you'd alluded to in the RFC. That would greatly simplify and standardize things for implementors.

If we do that I also lean toward renaming ActOnArgs to QuantumState, and moving all the implementations of it to the QIS module. Maybe we should consider this in the context of your `Iterator` change so that the modifications to the simulation APIs can be done all in one go.

cirq-core/cirq/sim/simulation_engine.py
```diff
@@ -0,0 +1,238 @@
...
+    SimulatesSamples,
+    check_all_resolved,
+    split_into_matching_protocol_then_general,
+)
+
+if TYPE_CHECKING:
+    import cirq
+
+
+class SimulationEngine(
```
95-martin-orion, line 49: This could generate confusion with `cirq_google.Engine`. Maybe `SimulatorBase` instead?

95-martin-orion (changes requested):
Couple more nits, but otherwise I think this is ready. It's especially reassuring that this final version doesn't affect the existing `cirq/sim/simulator.py` API at all.

cirq-google/cirq_google/calibration/engine_simulator.py
```diff
@@ -386,20 +389,24 @@ def create_gate_with_drift(
...
-
-    def _core_iterator(
+    def run_sweep(
+        self,
+        program: Circuit,
+        params: Sweepable,
+        repetitions: int = 1,
+    ) -> List[Result]:
+        converted = _convert_to_circuit_with_drift(self, program)
+        return super().run_sweep(converted, params, repetitions)
```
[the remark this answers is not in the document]
daxfohl: Done

cirq-core/cirq/sim/simulation_engine.py
daxfohl: Done

95-martin-orion (approved):
Good to go! For tracking the remaining work, I recommend either un-linking #2178 or opening a fresh issue to track the remaining refactoring.

PR #4037, opened 2021-04-20
Fix cirq_google notebooks

balopat:
Makes notebooks using cirq_google install pre-release cirq. 

These notebooks were failing in the devsite pipeline as they were just installing cirq, not `cirq --pre`.


maffoo (approved):
LGTM

PR #4039, opened 2021-04-21
Add a `MeasurementKey` class and make it the internal representation in `MeasurementGate`

smitsanghavi:
Phase 1 of the RFC at https://tinyurl.com/structured-measurement-keys#heading=h.7xjjz7p4hj9y

Part of #4040


smitsanghavi:
@95-martin-orion (since I can't manually add reviewers)

cirq-core/cirq/ops/measurement_gate.py
```diff
@@ -69,6 +69,16 @@ def __init__(
...
 
+    @property
+    def key(self) -> str:
+        return str(self.mkey)
+
+    @key.setter
+    def key(self, key_str: str):
+        # TODO: Allow nested keys only when created via some special methods. Planned in Phase 2a
+        #   of https://tinyurl.com/structured-measurement-keys#heading=h.zafcj653k11m
+        self.mkey = value.MeasurementKey(key_str, allow_nested_key=True)
```
mpharrigan, line 80: would it make more sense to set `allow_nested_key` to `False` until the full logic is in place? Otherwise we could break stuff in between.

cirq-core/cirq/value/measurement_key.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __repr__(self):
+        return f'cirq.MeasurementKey(name={self.name})'
+
+    def __str__(self):
+        return self.name
+
+    def __hash__(self):
+        if self._hash is None:
+            object.__setattr__(self, '_hash', hash(self.name))
```
95-martin-orion, line 63: Nit: include `allow_nested_key` in the hash?

cirq-core/cirq/value/measurement_key.py
```diff
@@ -0,0 +1,82 @@
...
+class MeasurementKey:
+    """A class representing a Measurement Key.
+
+    Wraps a string key and provides validation. If you just want the string measurement key, simply
+    call `str()` on this.
+    TODO: Change allow_nested_key to is_nested_key when the nested key callers start interacting
+        with `MeasurementKey` directly. That way we can have tighter validation and won't need to
+        store allow_nested_key, is_nested_key can just be derived from name itself. This will
+        happen in Phase 2a detailed at
+        https://tinyurl.com/structured-measurement-keys#heading=h.zafcj653k11m
```
95-martin-orion, line 32: What is the use case for persisting `allow_nested_key` beyond the constructor at all? My intuition is that whether nested keys are allowed would depend on the calling method, not the key itself.

cirq-core/cirq/value/measurement_key_test.py
```diff
@@ -0,0 +1,100 @@
...
+import pytest
+
+import cirq
+
+
+def test_empty_init():
+    with pytest.raises(TypeError, match='required positional argument'):
+        _ = cirq.MeasurementKey()
+    mkey = cirq.MeasurementKey('')
+    assert mkey.name == ''
```
95-martin-orion, line 24: Now that we have a class, we might as well prevent empty keys - they can only generate confusion.

cirq-core/cirq/value/measurement_key.py
```diff
@@ -0,0 +1,82 @@
...
+    def __post_init__(self):
+        if not self.allow_nested_key and MEASUREMENT_KEY_SEPARATOR in self.name:
+            raise ValueError(
+                f'Measurement key {self.name} invalid. "{MEASUREMENT_KEY_SEPARATOR}" is not '
+                'allowed in measurement key constructor. If you want to use this character to '
+                'specify nested measurement keys, set the allow_nested_key option.'
+            )
+
+    def __eq__(self, other) -> bool:
+        return str(self) == str(other)
```
95-martin-orion, line 53: This worries me - there's a lot of classes out there, with a lot of string methods. I would instead use specific checks for `isinstance(other, str)` and `isinstance(other, MeasurementKey)`, and return NotImplemented otherwise.

95-martin-orion (changes requested):

cirq-core/cirq/ops/measurement_gate.py
```diff
@@ -69,6 +69,16 @@ def __init__(
...
             raise ValueError('len(invert_mask) > num_qubits')
 
+    @property
+    def key(self) -> str:
+        return str(self.mkey)
+
+    @key.setter
+    def key(self, key_str: str):
+        # TODO: Allow nested keys only when created via some special methods. Planned in Phase 2a
+        #   of https://tinyurl.com/structured-measurement-keys#heading=h.zafcj653k11m
```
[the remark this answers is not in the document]
smitsanghavi: Done

cirq-core/cirq/ops/measurement_gate.py
smitsanghavi: Actually, I'm setting it to `True` to ensure nothing breaks. Currently, a decomposed `CircuitOperation` will has these nested keys inside regular `MeasurementGate`s (nested keys are just keys with a `:` inside). 

Until [Phase 2a of the RFC](https://tinyurl.com/structured-measurement-keys#heading=h.zafcj653k11m), the `MeasurementKey` and `MeasurementGate` classes will not know if their keys are being remapped in a way that they need to allow nested keys. So setting a blanket allow to ensure nothing triggers the `ValueError` (except if someone creates a nested `MeasurementKey` directly).

cirq-core/cirq/value/measurement_key.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __repr__(self):
+        return f'cirq.MeasurementKey(name={self.name})'
+
+    def __str__(self):
+        return self.name
+
+    def __hash__(self):
```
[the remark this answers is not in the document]
smitsanghavi: Yeah, I defined them specifically to exclude `allow_nested_key`. I feel it should just be a InitVar but had to keep it around for serialization. More on it in the comment below.

cirq-core/cirq/value/measurement_key.py
smitsanghavi: I don't think we should. WDYT?

cirq-core/cirq/value/measurement_key.py
smitsanghavi: Yes, I wanted to make it an `InitVar` but we need it in the serialized representation to ensure that we don't raise an error when we de-serialize. We can dummy-add it by parsing the key name but that would introduce some inconsistency between construction and serialization.

cirq-core/cirq/value/measurement_key_test.py
smitsanghavi: Agreed. A bunch of tests failed since sometimes we don't bother naming the measurements in tests. I didn't want to clutter this PR with all that changes so will send that in a subsequent PR as soon as this is in.

cirq-core/cirq/value/measurement_key.py
smitsanghavi: Fair enough. Added that condition for `__eq__`

cirq-core/cirq/value/measurement_key_test.py
mpharrigan: If you don't name the measurement doesn't it give it an automatic name of `str(qubits)`?

smitsanghavi: `cirq.measure` does. But `cirq.MeasurementGate`, being a gate, does not know what qubits it's going to be operating on. So the tests which directly use the gate with `on` would fail and need explicit keys. 

As part of special treatment for qubits (phase 2b of the RFC), `MeasurementGate.on` would be overridden to pass the qubit info to `MeasurementKey` which would do remapping and maybe create the default qubit-based key, if needed.

mpharrigan: Thanks for the clarification!

cirq-core/cirq/value/measurement_key.py
95-martin-orion: Discussed offline - the deserialization case is a good argument for keeping this parameter.

cirq-core/cirq/value/measurement_key_test.py
95-martin-orion: Sounds good to me - we can handle extra validation in subsequent PRs.

95-martin-orion (changes requested):
Discussed offline - plan is to remove validation from this PR and cover it in subsequent PRs.

cirq-core/cirq/protocols/json_test_data/MeasurementKey.repr
```diff
@@ -0,0 +1 @@
+[cirq.MeasurementKey('key'), cirq.MeasurementKey('nested:key')]
```
tanujkhattar, line 1: Was this generated by actually calling `repr(g)` for some object `g` ?

The `_repr_` above returns `f'cirq.MeasurementKey(name={self.name})'`, which seems wrong. See #4440

PR #4041, opened 2021-04-21
Speed up cirq import by removing scipy.stats import

dabacon:
This speeds up import on my box by around 13%.

Before:
image
After:
image


maffoo (commented):
Looks like this is used in `von_neumann_entropy` below, so not sure that we can remove the import: https://github.com/quantumlib/Cirq/blob/master/cirq-core/cirq/qis/measures.py#L275. Tests pass, but I don't know if we have coverage of that particular case in `von_neumann_entropy`.

balopat:
> Looks like this is used in `von_neumann_entropy` below, so not sure that we can remove the import: https://github.com/quantumlib/Cirq/blob/master/cirq-core/cirq/qis/measures.py#L275. Tests pass, but I don't know if we have coverage of that particular case in `von_neumann_entropy`.

Nice find! I guess we could still move the import into the von_neumann_entropy function and add a comment that this is a local import for import performance reasons?

dabacon:
> Looks like this is used in `von_neumann_entropy` below, so not sure that we can remove the import: https://github.com/quantumlib/Cirq/blob/master/cirq-core/cirq/qis/measures.py#L275. Tests pass, but I don't know if we have coverage of that particular case in `von_neumann_entropy`.

We do have coverage, at least I checked by printing debug right before the statement then failing a test and seeing that the print showed up.

dabacon:
Oh lol, the test imports scipy.stats.

dabacon:
OK, made the import delayed in the function.  Alternatively we could just calculate the entropy using numpy, the only case you need to be careful about are the 0 * log(0) terms.

cirq-core/cirq/qis/measures_test.py
```diff
@@ -207,7 +206,10 @@ def test_von_neumann_entropy():
     probs = np.random.exponential(size=N)
     probs /= np.sum(probs)
     mat = U @ (probs * U).T.conj()
-    np.testing.assert_allclose(cirq.von_neumann_entropy(mat), scipy.stats.entropy(probs, base=2))
+
+    np.testing.assert_allclose(
+        cirq.von_neumann_entropy(mat), -np.sum(probs * np.log(probs) / np.log(2))
```
maffoo, line 211: nit: can use np.log2
```suggestion
        cirq.von_neumann_entropy(mat), -np.sum(probs * np.log2(probs))
```

maffoo (approved):
Minor comment, then LGTM

PR #4043, opened 2021-04-21
adding editable installs to development.md

balopat:


docs/dev/development.md
```diff
@@ -129,6 +129,17 @@ See the previous section for instructions.
...
+## Editable installs 
+
+If you want to pip install cirq in an editable fashion, you'll have to install it per module, e.g.: 
+
+```
+pip install -e ./cirq-core -e ./cirq-google
+```
+
+Note that `pip install -e .` will install the `cirq` metapackage only, and your code changes won't 
+get picked up!
```
mpharrigan, line 141: In my experience, it won't install anything :) 

The issue I ran into is when I didn't put one `-e` per path, i.e. I did
```
pip install -e ./cirq-core ./cirq-google
```

so cirq-core was editable but cirq-google was not

mpharrigan (approved):

docs/dev/development.md
balopat: So, `pip install .` does install the metapackage: 

`pip show cirq` 

Results in 
```
Name: cirq
Version: 0.11.0.dev0
Summary: A framework for creating, editing, and invoking Noisy Intermediate Scale Quantum (NISQ) circuits.
Home-page: http://github.com/quantumlib/cirq
Author: The Cirq Developers
Author-email: cirq-dev@googlegroups.com
License: Apache 2
Location: /Users/balintp/dev/proj/Cirq
Requires: cirq-core, cirq-google
Required-by: 

```

`pip install ./cirq-core` installs local version non-editable 
`pip install -e ./cirq-core` installs local version editable 
`pip install cirq-core` will install pypi version!! 

Thus, if you ran `pip install -e cirq-core cirq-google` that installed cirq-core locally editable, cirq-google from pypi!

mpharrigan: oh wow. What if I did `pip install -e cirq-core/ cirq-google/` with a trailing slash?

If you do `pip install .` does it install the pypi versions of `cirq-core` and `cirq-google`? IIRC when I tried it, it tried to install the specific dev version (0.11.0.dev0) of the `cirq-core` and `cirq-google` which don't exist on PyPI and it doesn't search locally, so it just gave up. Maybe it technically installed the metapackage but it complained about unsatisfiable dependencies.

Issue #3788, opened 2021-02-11
Create Sampler-like API that can support cirq.google.Engine calibrations


mrwojtek:
**Is your feature request related to a use case or problem? Please describe.**
The recently added `run_calibrations` on `cirq.google.Engine` is forcing the use the `cirq.google.Engine` itself when using the characterization and calibration features. This is especially problematic during simulations since `Engine` is not an interface but a class that handles connections to Google hardware. This causes problem for example while working with `cirq.google.calibration.PhasedFSimEngineSimulator` since it can't be easily substituted for `Engine`. There are already methods which takes both `Engine` or `PhasedFSimEngineSimulator` and choose different code paths to simulate  `run_calibrations`.

**Describe the solution you'd like**
Introduce a `cirq.Sampler`-like abstraction that allows for easy use of calibrations framework and seamless switching between real `Engine` and simulation.

**What is the urgency from your perspective for this issue? Is it blocking important work?**
P2/P3

We can wait and see how the new calibration API is used before moving forward with this issue.

balopat:
I would suggest to start with this interface living in cirq.google - as such this wouldn't be cirq.* level, like the Sampler API. 
I believe that this calibration model _might_ be useful across hardware vendors - however that's a conversation to have with our vendors, after which we can "upstream" this interface back to the Sampler level.

mpharrigan:
@dstrain115 points out that this should mock out `Engine`, not just specifically calibration api

dstrain115:
This has been completed as part of the AbstractEngine interface.  See 
https://github.com/quantumlib/Cirq/blob/master/cirq-google/cirq_google/engine/abstract_engine.py

PR #4047, opened 2021-04-23
[XEB] Local wrapper for characterization functionality

mpharrigan:
XEB calibration can be run through the calibration api, i.e. by a call to `engine.run_calibration` but since all the code is open-sourced in Cirq and only requires the ability to sample from a quantum processor, we can write a wrapper which can be dispatched to by `cg.workflow.run_calibrations` that submits circuits using a `cirq.Sampler` and performs XEB angle optimization locally.


mpharrigan:
I don't know how to test this. I mocked out the sampling and the optimization and it still takes 28 seconds. Also monkeypatching multiprocessing code can cause problems if the spawn method is not "fork"

cirq-google/cirq_google/calibration/xeb_wrapper.py
```diff
@@ -0,0 +1,84 @@
...
+import cirq.experiments.xeb_sampling as xebsamp
+from cirq_google.calibration.phased_fsim import (
+    XEBPhasedFSimCalibrationRequest,
+    PhasedFSimCalibrationResult,
+    PhasedFSimCharacterization, XEBPhasedFSimCalibrationOptions,
+)
+
+
+def run_calibration(
+        calibration: XEBPhasedFSimCalibrationRequest,
```
mrwojtek, line 28: I was hoping that for this calibration you'd create something like `LocalXEBPhasedFSimCalibrationRequest` which indicates that XEB should be done locally.

This way we could keep the single [run_calibrations](https://github.com/quantumlib/Cirq/blob/master/cirq-google/cirq_google/calibration/workflow.py#L602) method which triggers calibrations. The method could recognize the local requests and delegate them to your wrapper. The method takes a sequence of calibrations  so one issue would be if user mixes local and non-local calibrations. We could disallow for such a mix and raise an exception or filter out local and non-local calibrations.

What I would like to preserve is an ability of decoupling sampler/engine from calibration requests: user should write their code against `cirq.google.run_calibrations` and depending on passed sampler and different type of requests, the desired functionality could be realized.

There is a todo for me here as well, I see that `run_calibrations` take `engine: Union[Engine, PhasedFSimEngineSimulator]`. I think it will be better if it is just `engine: Union[Engine, Sampler]`, and if `EngineSampler` is passed, the engine object could be extracted from there. This will simplify our tutorials as well. This is related to #3788 but I'm not sure if this is the tame to make it so public, we need to test this feature more.

cirq-google/cirq_google/calibration/xeb_wrapper.py
```diff
@@ -0,0 +1,84 @@
...
+    # 4. Initial fidelities
+    initial_fids = xebf.benchmark_2q_xeb_fidelities(
+        sampled_df=sampled_df,
+        circuits=circuits,
+        cycle_depths=cycle_depths,
+    )
+
+    # 5. Characterize by fitting angles.
+    pcircuits = [xebf.parameterize_circuit(circuit, options.fsim_options) for circuit in circuits]
+    with multiprocessing.Pool() as pool:
```
mrwojtek, line 64: The default options kills the users desktops, we need to have control over number of processors used. I'd suggest an option in `LocalXEBPhasedFSimCalibrationRequest` which defaults to the number of processors minus 1.

cirq-google/cirq_google/calibration/phased_fsim.py
```diff
@@ -247,11 +247,19 @@ class PhasedFSimCalibrationResult:
...
+            Not all characterization techniques can report this.
+        final_fids: The circuit fidelities by depth after characterization of the PhasedFSim.
+            Not all characterization techniques can report this.
+
     """
 
     parameters: Dict[Tuple[Qid, Qid], PhasedFSimCharacterization]
     gate: Gate
     options: PhasedFSimCalibrationOptions
+    initial_fids: Optional[pd.DataFrame] = None
```
mrwojtek, line 261: I just noticed that this object is not JSON-serializable which is actually the most important one to serialize! We need to add serialization but because of that I don't think we should use pd.DataFrame. Shouldn't this be a plain `Dict[Tuple[Qid, Qid], Dict[int, float]]`?

As for `pd.DataFrame`, I'd suggest to have a method get_fidelities_frame() which would return frame with columns [`qubit_a`, `qubit_b`, `depth`, `initial_fidelity`, `final_fidelity`] out of the underlying data.

cirq-google/cirq_google/calibration/phased_fsim.py
```diff
@@ -247,11 +247,19 @@ class PhasedFSimCalibrationResult:
             quibts a and b either only (a, b) or only (b, a) is present.
         gate: Characterized gate for each qubit pair. This is copied from the matching
             PhasedFSimCalibrationRequest and is included to preserve execution context.
+        options: The options used to gather this result.
+        initial_fids: The circuit fidelities by depth before characterization of the PhasedFSim.
+            Not all characterization techniques can report this.
```
mrwojtek, line 252: We should list handlers which return this information.

cirq-google/cirq_google/calibration/phased_fsim.py
```diff
@@ -247,11 +247,19 @@ class PhasedFSimCalibrationResult:
...
+        final_fids: The circuit fidelities by depth after characterization of the PhasedFSim.
+            Not all characterization techniques can report this.
+
     """
 
     parameters: Dict[Tuple[Qid, Qid], PhasedFSimCharacterization]
     gate: Gate
     options: PhasedFSimCalibrationOptions
+    initial_fids: Optional[pd.DataFrame] = None
+    final_fids: Optional[pd.DataFrame] = None
```
mrwojtek, line 262: I think the names are quire misleading or inaccurate. 

I would name the `initial_fids` more like `ideal/perfect_gate_fidelities` or `ideal_fidelities` and `final_fids` more like `matched_gate_fidelities` or `matched_fidelities`. What do you think? I would rather focus on what data represents and not on how it was created.

cirq-google/cirq_google/calibration/xeb_wrapper.py
mpharrigan: Sounds good.

The dispatch behavior of `cirq.google.run_calibrations` is getting really complicated. 

1. normal requests, PhasedFSimEngineSimulator [calls special engine.get_calibrations]
2. normal requests, Engine [calls engine.run_calibration]
3. local requests, Engine [construct enginesampler, calls wrapper run_local_xeb_calibration]
4. local requests, sampler [calls wrapper run_local_xeb_calibration]

You could potentially mix-and-match (2) and (3), but that gets tricky when max_layers_per_request > 1 and they get batched together. I think for now, we should enforce only one calibration type per call to run_calibrations.

I'll put together an example with the above four conditions, but let me know if you have any thoughts.

mpharrigan:
@mrwojtek let me know if 828881dfe74a7f59a0a56399691fc0c921fa29cc is in the right direction

cirq-google/cirq_google/calibration/phased_fsim.py
mrwojtek: Sorry, this class is JSON-serializable, it was just cluttered but github diff. Could you add support for JSON serialization of the two new fields as well?

cirq-google/cirq_google/calibration/xeb_wrapper.py
mrwojtek: I think we are getting somewhere with this. I'm fine with restricting single-type calibrations per request. In fact. I'm starting to miss more and more a grouping object which we could use to group calibrations together.

The only case that I'm hesitant is number (1) and local sampler. Should it also call engine.get_calibrations? Calling sampler and doing real simulation is useful but I'm not sure if this is the right abstraction level. On the other hand, engine.get_calibrations() doesn't support fidelities that could be obtained automatically by doing simulations. Maybe boolean option to run_calibrations for that?

Please take a look at #4054, I have simplified `run_calibrations` independently; we can merge with this PR.

cirq-google/cirq_google/calibration/xeb_wrapper.py
mpharrigan: Added

cirq-google/cirq_google/calibration/phased_fsim.py
mpharrigan: Removed these fields for now. Will introduce in a follow-on PR where we can do it more thoughtfully.

cirq-google/cirq_google/calibration/phased_fsim.py
mpharrigan: (removed for future pr)

cirq-google/cirq_google/calibration/phased_fsim.py
mpharrigan: (removed for future pr)

cirq-google/cirq_google/calibration/xeb_wrapper.py
```diff
@@ -0,0 +1,102 @@
...
+)
+
+
+@contextlib.contextmanager
+def _maybe_multiprocessing_pool(
+    n_processes: Optional[int] = None,
+) -> Iterator[Union['multiprocessing.pool.Pool', None]]:
+    if n_processes == 1:
+        yield None
+        return
```
mrwojtek, line 37: This is instead of else: ? Mixing yield and return looks a little bit strange at first sight.

mrwojtek (approved):

mpharrigan:
@mrwojtek ptal!

mrwojtek:
@mpharrigan too late ;)

mpharrigan:
ninja'd!

cirq-google/cirq_google/calibration/workflow.py
```diff
@@ -649,36 +707,34 @@ def run_calibrations(
...
-    elif isinstance(sampler, PhasedFSimEngineSimulator):
-        results = sampler.get_calibrations(calibrations)
-    else:
-        raise ValueError(f'Unsupported sampler type {type(sampler)}')
+    if isinstance(sampler, PhasedFSimEngineSimulator):
+        return sampler.get_calibrations(calibration_requests)
 
-    return results
+    raise ValueError(
+        f'Unsupported engine/request combinations: {engine} cannot run {calibration_requests}'
```
balopat, line 736: When engine is None, this will read "None cannot run [XEBPhasedFSimCalibrationRequest(...), XEBPhasedFSimCalibrationRequest(...), ...]", also the user passes in `sampler`, not `engine`. I would rather use 

```suggestion
        f'Unsupported sampler/request combinations. Sampler={sampler} cannot run calibration request of type {calibration_request_type}'
```

cirq-google/cirq_google/calibration/xeb_wrapper_test.py
```diff
@@ -0,0 +1,131 @@
...
+from cirq_google.calibration.phased_fsim import (
+    LocalXEBPhasedFSimCalibrationOptions,
+    LocalXEBPhasedFSimCalibrationRequest,
+)
+from cirq_google.calibration.xeb_wrapper import (
+    run_local_xeb_calibration,
+    _maybe_multiprocessing_pool,
+)
+
+SQRT_ISWAP = cirq.ISWAP ** -0.5
```
balopat, line 35: optional: I see this in so many files, shouldn't we just introduce it in the cirq global namespace? :)

cirq-google/cirq_google/calibration/xeb_wrapper_test.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def minimize_patch(
```
balopat, line 38: make private?

cirq-google/cirq_google/calibration/xeb_wrapper_test.py
```diff
@@ -0,0 +1,131 @@
...
+        nfev=0,
+        status=0,
+        success=True,
+        message='monkeypatched',
+        x=x0.copy(),
+        final_simplex=None,
+    )
+
+
+def benchmark_patch(*args, **kwargs):
```
balopat, line 66: make private?

cirq-google/cirq_google/calibration/xeb_wrapper.py
```diff
@@ -0,0 +1,102 @@
...
+        )
+
+    return PhasedFSimCalibrationResult(
+        parameters={
+            pair: PhasedFSimCharacterization(**param_dict)
+            for pair, param_dict in char_results.final_params.items()
+        },
+        gate=calibration.gate,
+        options=options,
+        # initial_fids=initial_fids,
```
balopat, line 100: Are these commented lines necessary?

cirq-google/cirq_google/calibration/xeb_wrapper.py
```diff
@@ -0,0 +1,102 @@
...
+        return
+
+    with multiprocessing.Pool(processes=n_processes) as pool:
+        yield pool
+
+
+def run_local_xeb_calibration(
+    calibration: LocalXEBPhasedFSimCalibrationRequest,
+    sampler: 'cirq.Sampler',
+) -> PhasedFSimCalibrationResult:
```
balopat, line 46: please add docstrings to public functions

balopat (changes requested):
Found a bunch of nits, otherwise LGTM

cirq-google/cirq_google/calibration/phased_fsim.py
```diff
@@ -463,6 +464,19 @@ def _from_json_dict_(cls, **kwargs):
         return cls(**kwargs)
 
 
+@json_serializable_dataclass(frozen=True)
+class LocalXEBPhasedFSimCalibrationOptions(XEBPhasedFSimCalibrationOptions):
+
```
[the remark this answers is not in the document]
mpharrigan: added

cirq-google/cirq_google/calibration/phased_fsim.py
```diff
@@ -702,6 +716,17 @@ def _parse_characterized_angles(
     return dict(records)
 
 
+@json_serializable_dataclass(frozen=True)
+class LocalXEBPhasedFSimCalibrationRequest(PhasedFSimCalibrationRequest):
```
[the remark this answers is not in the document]
mpharrigan: added

cirq-google/cirq_google/calibration/workflow.py
```diff
@@ -600,8 +606,55 @@ def _merge_into_calibrations(
...
+        run_local_xeb_calibration(
+            cast(LocalXEBPhasedFSimCalibrationRequest, calibration_request), sampler
+        )
+        for calibration_request in calibration_requests
+    ]
+
+
 def run_calibrations(
-    calibrations: Sequence[PhasedFSimCalibrationRequest],
+    calibration_requests: Sequence[PhasedFSimCalibrationRequest],
```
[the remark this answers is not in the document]
mpharrigan: I didn't realize this was a part of v0.10; reverting. (except we recently renamed the second argument)

cirq-google/cirq_google/calibration/workflow.py
```diff
@@ -635,9 +688,14 @@ def run_calibrations(
             f'given'
         )
 
-    if not calibrations:
+    if not calibration_requests:
         return []
 
+    calibration_request_types = set(type(cr) for cr in calibration_requests)
+    if len(calibration_request_types) > 1:
+        raise ValueError("All calibrations must be of the same type.")
```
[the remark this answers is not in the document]
mpharrigan: good idea, done

cirq-google/cirq_google/calibration/workflow.py
```diff
@@ -649,36 +707,34 @@ def run_calibrations(
 
     if engine is not None:
         if processor_id is None:
-            raise ValueError('processor_id must be provided when running on the engine')
+            raise ValueError('processor_id must be provided when using an Engine')
```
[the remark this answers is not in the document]
mpharrigan: would "an Engine object" be better? the name of the class is `Engine`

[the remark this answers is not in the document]
mpharrigan: although I guess following https://github.com/quantumlib/Cirq/pull/4054 you can still reach this code path when giving a QuantumEngineSampler. I'll Just remove the conditional alltogether and the user will just have to trust the error message :)

cirq-google/cirq_google/calibration/workflow.py
mpharrigan: good catch and suggestion. The error was from a merge issue following https://github.com/quantumlib/Cirq/pull/4054

cirq-google/cirq_google/calibration/xeb_wrapper.py
```diff
@@ -0,0 +1,102 @@
...
+        cycle_depths=cycle_depths,
+        combinations_by_layer=combs_by_layer,
+    )
+
+    # 4. Initial fidelities
+    # initial_fids = xebf.benchmark_2q_xeb_fidelities(
+    #     sampled_df=sampled_df,
+    #     circuits=circuits,
+    #     cycle_depths=cycle_depths,
+    # )
```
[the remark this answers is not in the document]
mpharrigan: I'm going to re-enable in a follow-up PR. It would be helpful for me to keep this for a short period of time. Any readers of the code will be able to discern the correct, full flow although we haven't fully designed how to get this particular piece of data back to the user via the Request/Result classes.

cirq-google/cirq_google/calibration/xeb_wrapper.py
mpharrigan: removed these pending follow on pr

cirq-google/cirq_google/calibration/xeb_wrapper_test.py
mpharrigan: could be helpful for sure!

cirq-google/cirq_google/calibration/xeb_wrapper_test.py
mpharrigan: done. didn't think it mattered much in a test file

cirq-google/cirq_google/calibration/xeb_wrapper_test.py
mpharrigan: done

mpharrigan (commented):
@balopat PTAL

balopat (approved):
LGTM!

Issue #4045, opened 2021-04-22
simulations using CuPy


oumjunior:
**The simulation of circuits is somewhat slow because the underlying algebraic operations are carried using NumPy.**

**We would like to leverage the GPUs' power by using the CuPy library to speedup the simulations**

**Below is the simulation time of supremacy circuits using Cirq, Cirq with CuPy, Qsim and Qulacs**

![image](https://user-images.githubusercontent.com/32430632/115553503-c4385480-a2ad-11eb-8bdf-a86b9b4c570f.png)

vtomole:
Hey @oumjunior , from that graph it looks like `qsim` performs "better" than `CuPy`. `qsim`'s execution time is lower than `CUPy`'s up until around 20 qubits. After that, their execution times are pretty similar. Is this feature request to add the `CuPy` simulator to Cirq?

oumjunior:
Hi @vtomole, indeed the speedup starts to kick in around 20 qubits and the execution time ratio CuPy/Qsim is around 2 (for width > 20 qubits). Note also that only starting from this threshold (~20 qubits) that the time execution of the ordinary simulator (with NumPy) explodes.  

Below is another sample: simulation time of the Multiplier circuit from https://arxiv.org/pdf/1706.05113v1.pdf 

![image](https://user-images.githubusercontent.com/32430632/115847995-b7d60800-a423-11eb-919f-c27e56522a40.png)

95-martin-orion:
This is something of a tangent to the issue at hand, but it will help establish a better comparison between cupy and qsim:

-----

How many qubits appear in each gate for the Multiplier circuit, and can they be decomposed to 1- and 2-qubit gates? The fact that qsim is barely keeping pace with Cirq (when it's normally expected to beat Cirq by 10x or more for 20+ qubits) suggests to me that the gates are larger than 2 qubits, which limits the effectiveness of qsim optimizations.

This can be remedied by passing `qsim_options={f: gate_qubits}` to the QSimSimulator constructor, where `gate_qubits` is the number of qubits in each gate. This feature is documented here:
https://github.com/quantumlib/qsim/blob/795293072ef0f5f6b2bc099e94c5080c5e22473d/qsimcirq/qsim_simulator.py#L79-L86

Would you be able to re-run this experiment with the option described above? This should give a more accurate comparison with optimal qsim performance on these circuits.

oumjunior:
[...]

PR #4049, opened 2021-04-23
CuPy-based simulator

oumjunior:
CuPy-based simulator that exploits the power of GPUs and perform fast simulations especially for circuits with great width. 

The Multiplier (https://arxiv.org/pdf/1706.05113v1.pdf) execution time 
image

Supremacy circuits execution time 
image


oumjunior:
@googlebot I signed it!

Le ven. 23 avr. 2021 à 11:19, google-cla[bot] ***@***.***> a
écrit :

vtomole (changes requested):
Replacing `numpy` with `cupy` is a _huge_ change. Let's sit and think about this before we make a decision. Like i said on https://github.com/quantumlib/Cirq/issues/4045, as of now I'm not convinced that the performance improvements this change brings are worth the changes to Cirq. Maybe GPU simulation could be something that qsim handles. CC @95-martin-orion

balopat:
Hi @oumjunior, thank you for submitting this proposal. However, changing all of numpy to cupy is way too wide of a change. First of all, this change is impacting other parts of the code base like diagram drawing, qis, benchmarking, etc. which have nothing to do with simulation. Second as Victory says this might not be the place to introduce GPU support. qsim does have plans to have GPU support which will probably change the graphs. Before that happens I could _maybe_ see  creating a separate cirq module - cirq-cupy or cirq-gpu which would have a similar GPU based - however that is something to bring to the Cirq Cynque first where we can discuss whether we'd like to maintain a solution separately - given that qsim will have a GPU solution. If the verdict is none, you can still create a GPU based simulator cirq module outside of our repo and if it's well maintained we can reference it in our docs as part of our ecosystem.

95-martin-orion:
Thanks Victory for the CC.

@oumjunior, I've added a comment to #4045 that may provide a clearer comparison with qsim.

That said, I'm interested in integrating a GPU-based simulator with Cirq regardless of its performance - I think it's a worthwhile point of reference at a minimum, and there may be specific cases where a GPU-based simulator is preferable to existing options. In lieu of a full conversion to `cupy` (which, as noted above, would require significant discussion) here's what I recommend:

- Isolate the `cupy` behavior to a new implementation of the `Simulator` interface in `cirq-core/cirq/sim/simulator.py`. The result of this would be similar to `cirq-core/cirq/sim/sparse_simulator.py`
- Open a new issue for converting Cirq to `cupy`, and join our Cirq sync (Wednesdays @ 11AM PST) to drive discussion. You can get access to the Cirq sync by joining the [cirq-dev googlegroup](https://groups.google.com/g/cirq-dev).

daxfohl:
So we don't have the number of permutations explode, would it be better to add a `use_gpu: bool` field to existing simulators rather than creating a dedicated simulator?

balopat (changes requested):
As it is, this cannot go in. 
I'm completely open towards a GPU flag / maybe adding the ability to inject numpy implementations to simulators, but we should have a discussion about this. 
I think this is a great candidate for Cirq Cynque discussion.

vtomole:
Hey @oumjunior, please try signing the CLA again. It looks like the CLA bot didn't accept the first signature.

balopat:
As we discussed this on Cirq Cynque, numpy is way too central to Cirq to risk switching it out. For users who want to leverage GPU based speedups, qsim will provide GPU integrations soon. At this point we are not ready to take this risk in Cirq. Thank you for the prototype though, it was a very interesting experiment!

PR #4051, opened 2021-04-26
Fix naming of sqrt(iswap) gate constants

mrwojtek:
The constant named SQRT_ISWAP_PARAMETERS is really an inverse of it and this PR fixes that.

This is a version of #3908 without module-level parameter deprecation which I couldn't work out. The number of impacted users is marginal if not none (this is rather a private parameter).


95-martin-orion (approved):
Added the `BREAKING CHANGE` label for clarity. If there's anyone that we _know_ will be affected by this, we should ping them first; outside of that, we can move ahead with this change.

mrwojtek:
Thanks, I'll wait a few hours and proceed with a merge afterwards.

PR #4052, opened 2021-04-26
Add option to manipulate Floquet readout thresholds.

mrwojtek:


mrwojtek:
@wcourtney FYI

cirq-google/cirq_google/calibration/phased_fsim.py
```diff
@@ -476,13 +476,20 @@ class FloquetPhasedFSimCalibrationOptions(PhasedFSimCalibrationOptions):
         characterize_chi: Whether to characterize χ angle.
         characterize_gamma: Whether to characterize γ angle.
         characterize_phi: Whether to characterize φ angle.
+        readout_error_tolerance: Threshold for pairwise-correlated readout errors above which the
+            calibration will report to fail. Just before each calibration all pairwise two-qubit
+            readout errors are checked and when any of the pairs reports an error above the
+            threshold, the calibration will fail. This value is a sanity check to determine if
+            calibration is reasonable and allows for quick termination if it is not. Set to 1.0 to
+            disable readout error checks and None to used default, device-specific thresholds.
```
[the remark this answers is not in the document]
mrwojtek: Done, thanks!

cirq-google/cirq_google/json_test_data/PhasedFSimCalibrationResult.json_inward
```diff
@@ -53,6 +53,7 @@
     "characterize_zeta": true,
     "characterize_chi": false,
     "characterize_gamma": false,
-    "characterize_phi": true
+    "characterize_phi": true,
+    "readout_error_tolerance": 0.4
```
balopat, line 57: this is the `_inward` file - so maybe it would be useful to keep this without the `readout_error_tolerance` field to test the backward compatibility?

mrwojtek: I was a little bit confused with the `_inward` files. Reverted.

cirq-google/cirq_google/calibration/phased_fsim.py
```diff
@@ -641,6 +654,10 @@ def _json_dict_(self) -> Dict[str, Any]:
         }
 
 
+def _correlated_from_readout_tolerance(readout_tolerance: float) -> float:
+    return max(0.0, min(1.0, 7 / 6 * readout_tolerance - 1 / 6))
```
mpharrigan, line 658: provide a short explanation for this formula?

cirq-google/cirq_google/calibration/phased_fsim.py
```diff
@@ -577,18 +584,24 @@ def from_moment(cls, moment: Moment, options: FloquetPhasedFSimCalibrationOption
...
+            'est_zeta': self.options.characterize_zeta,
+            'est_chi': self.options.characterize_chi,
+            'est_gamma': self.options.characterize_gamma,
+            'est_phi': self.options.characterize_phi,
+            # Experimental option that should always be set to True.
+            'readout_corrections': True,
+        }
+        if self.options.readout_error_tolerance is not None:
+            args['readout_error_tolerance'] = self.options.readout_error_tolerance
+            args['correlated_readout_error_tolerance'] = _correlated_from_readout_tolerance(
```
mpharrigan, line 598: why do we need both if one is a function of the other?

mpharrigan (changes requested):
some explanation of the two different quantities would be helpful

cirq-google/cirq_google/json_test_data/PhasedFSimCalibrationResult.json_inward
mpharrigan: You can copy the old version of the json files (without the readout_error_tolerance field) and name them with the inward suffix so it checks that old versions of the serialized data can still be read in.

Not super high priority for these cirq_google files under active development where we can intuit that it should load in just fine since the new field is optional.

cirq-google/cirq_google/calibration/phased_fsim.py
mrwojtek: We probably don't need two of those. I added this because I didn't know what's the best way to go at the time of deploying this option server-side. This is hidden from the public API, it doesn't matter much; we can remove it in a future releases.

cirq-google/cirq_google/calibration/phased_fsim.py
mrwojtek: Done, I've also commented on the keys in the args dict themselves.

mpharrigan (approved):

PR #4053, opened 2021-04-26
Store job identifiers in calibration results

mrwojtek:
This change adds project_id, program_id and job_id to the result object of each PhasedFSim-like calibration. This will allow to connect the past execution to the maintenance calibration or get the time when calibration happened.

One open question is: should the triplet (project_id, program_id, jub_id) be packaged into some serializable dataclass on the Engine side? This might be a common patter and having this kind of structure would simply and promote the storage of job's executions.


mrwojtek:
@wcourtney could you advise on how to make the tests working here? It fails because `cirq_google.Engine('project_id')` tries to authenticate against real servers.

wcourtney:
It looks like you need to mock out the EngineClient so that is skips the actual request to the service. Here's an example from the Engine test: https://github.com/quantumlib/Cirq/blob/b61d1c904583fda2f62f4f6a0da4f920426d4f7c/cirq-google/cirq_google/engine/engine_test.py#L426

mrwojtek:
@dstrain115 @mpharrigan

After chatting a little bit with @wcourtney I realized that maybe the best thing which is missing is actually to make `EngineJob` JSON-serializable.

Personally I think the ability to access the job through resource id is a great future but is very obscure to use it right now; you don't really want to use it without data persistence which is missing.

The only problem is that deserialization of the `EngineJob` would require an actual permissions to the given serialized project, so deserialization might not be possible for some users.

WDYT?

cirq-google/cirq_google/calibration/phased_fsim.py
```diff
@@ -247,11 +247,38 @@ class PhasedFSimCalibrationResult:
...
     parameters: Dict[Tuple[Qid, Qid], PhasedFSimCharacterization]
     gate: Gate
     options: PhasedFSimCalibrationOptions
+    project_id: Optional[str] = None
+    program_id: Optional[str] = None
+    job_id: Optional[str] = None
+    _hash: int = 0
+
+    def __post_init__(self):
+        # Workaround on inability to specify the parameters attribute as an immutable dictionary.
```
mpharrigan, line 264: This is a potentially dangerous workaround. Since `parameters` is mutable, we shouldn't make this a `frozen=True` dataclass. People could add it to a set or dictionary and then mutate `parameters`.

cirq-google/cirq_google/calibration/phased_fsim.py
```diff
@@ -284,6 +311,34 @@ def get_parameters(self, a: Qid, b: Qid) -> Optional['PhasedFSimCharacterization
         else:
             return None
 
+    # Workaround for: https://github.com/python/mypy/issues/1362
+    @property  # type: ignore
+    @lru_cache_typesafe
```
mpharrigan, line 316: If we remove the frozen attribute and the hashability of this class, you can remove lru_cache and instead cache the `Engine` object on the object the normal way

```python
if self._engine is not None:
  return self._engine
self._engine = ...
```

cirq-google/cirq_google/calibration/phased_fsim.py
```diff
@@ -284,6 +311,34 @@ def get_parameters(self, a: Qid, b: Qid) -> Optional['PhasedFSimCharacterization
...
+        """
+        if not self.project_id or not self.program_id or not self.job_id:
+            return None
+
+        engine = Engine(project_id=self.project_id)
+        return engine.get_program(self.program_id).get_job(self.job_id)
+
+    # Workaround for: https://github.com/python/mypy/issues/1362
+    @property  # type: ignore
+    @lru_cache_typesafe
```
mpharrigan, line 330: likewise, use normal caching

cirq-google/cirq_google/json_test_data/PhasedFSimCalibrationResult.json_inward
```diff
@@ -54,5 +54,8 @@
     "characterize_chi": false,
     "characterize_gamma": false,
     "characterize_phi": true
-  }
+  },
```
mpharrigan, line 57: the `_inward` files should omit the new fields to test backwards compatibility

mpharrigan (changes requested):
hash business

cirq-google/cirq_google/calibration/phased_fsim.py
mrwojtek: I don't like mutability because this container should be by design immutable. However this and your other suggestions point out that it would be cleaner to have it mutable indeed...

Are you aware of any way to create and use immutable dictionaries?

mpharrigan: An immutable dictionary is basically a dataclass 🤷 

Or a tuple of tuples of key-value. here's stack overflow musing on the topic https://stackoverflow.com/a/2704866

cirq-google/cirq_google/json_test_data/PhasedFSimCalibrationResult.json_inward
mrwojtek: Done

cirq-google/cirq_google/calibration/phased_fsim.py
mrwojtek: Done

cirq-google/cirq_google/calibration/phased_fsim.py
mrwojtek: Done

cirq-google/cirq_google/calibration/phased_fsim.py
mrwojtek: Done, it's mutable now

mpharrigan (approved):

mrwojtek:
Since this is useful to have, I'll merge it as is. I still think would be good to have an ability to easily serialize and deserialize jobs in JSON format.

PR #4054, opened 2021-04-26
Use sampler instead of engine as a primary run_calibrations runner

mrwojtek:


mrwojtek:
@mpharrigan what do you think about this? Take a look how the notebook example is simplified now.

PR #4056, opened 2021-04-27
Fix google notebooks

balopat:
The cirq_google extraction broke these notebooks - switching to `pip install cirq --pre` fixes them.


95-martin-orion (changes requested):
Does this also need the other changes listed in the [notebooks doc](https://github.com/quantumlib/Cirq/blob/master/docs/dev/notebooks.md#pre-release-notebooks)?

balopat:
> Does this also need the other changes listed in the [notebooks doc](https://github.com/quantumlib/Cirq/blob/master/docs/dev/notebooks.md#pre-release-notebooks)?

No, because these are all exclude already from both the notebook tests, being vendor codes. This is the reason the bugs weren't found in the first place.

95-martin-orion (approved):

PR #4058, opened 2021-04-27
Allow to pass a list of circuits to Floquet preparation method

mrwojtek:


cirq-google/cirq_google/calibration/workflow.py
```diff
@@ -268,13 +270,33 @@ def prepare_characterization_for_moments(
         operations matched by gates_translator, or it mixes a single qubit and two qubit gates.
     """
     if initial is None:
-        allocations: List[Optional[int]] = []
-        calibrations: List[RequestT] = []
-        pairs_map: Dict[Tuple[Tuple[Qid, Qid], ...], int] = {}
+        initial = []
+
+    if isinstance(circuit, Circuit):
```
balopat, line 275: For clarity, I would create a list called `circuits` and just iterate over that: 

```python
circuits = [circuit] if isinstance(circuit, Circuit) else circuit 
```

cirq-google/cirq_google/calibration/workflow.py
```diff
@@ -243,7 +243,7 @@ def prepare_characterization_for_moments(
     operations and operations supported by gates_translator.
 
     Args:
-        circuit: Circuit to characterize.
+        circuit: Circuit or list of circuits to characterize.
```
balopat, line 246: optional: consider renaming to `circuits`

balopat (approved):
LGTM after some naming / restructuring nits

cirq-google/cirq_google/calibration/workflow.py
mrwojtek: I also like it better but now it's a breaking change

cirq-google/cirq_google/calibration/workflow.py
mrwojtek: There is a different return type depending on each case, I don't think that I can unify those?

cirq-google/cirq_google/calibration/workflow.py
```diff
@@ -268,13 +270,33 @@ def prepare_characterization_for_moments(
...
-        calibrations = list(initial)
-        pairs_map = {calibration.pairs: index for index, calibration in enumerate(calibrations)}
+        circuits_with_calibration = []
+        requests = list(initial)
+        for circuit in circuits:
+            circuit_with_calibration, requests = _prepare_characterization_for_circuit_moments(
+                circuit, options, gates_translator, merge_subsets, requests
+            )
+            circuits_with_calibration.append(circuit_with_calibration)
+        return circuits_with_calibration, requests
```
mpharrigan, line 287: It's really useful to have this logic/boilerplate somewhere since I've found myself writing it a lot. 

That being said: I'm really nervous about having a different return type based on the input type. This is usually a pretty bad thing. Have you considered a) always returning a list of circuits (maybe of length 1) or b) keeping the old version of the function that works one circuit at a time and introducing a wrapper that's like `prepare_characterizations_for_circuits` which always returns a list (it can optionally handle one circuit input as a special case). This is basically how you've structured the PR already but keeping the name of the old function (and keeping it public) and adding a new API

mpharrigan (changes requested):
Some concerns about the API design. Let me know what you think

cirq-google/cirq_google/calibration/workflow.py
mrwojtek: I wanted to reserve `prepare_characterizations_for_circuits` for future calibrations which are not only moment-aware but consider the entire circuit structure.... There is `prepare_characterizations_for_moments` and `prepare_characterizations_for_operations` now, we might need `prepare_characterizations_for_circuits`. Or maybe this naming is a little bit broken.

What is really missing is the function overloading which doesn't exist in Python. I would argue this code is fine exactly because of this deficiency of a language? Balint, Matthew, what do you think?

mrwojtek: Actually, what would be even better is to support arbitrary nested sequences. I'm playing right now with some code I have 2 dimensional array already; i.e., have CIRCUIT_TREE analogous to the OP_TREE. I see that typing for OP_TREE is non-trivial, maybe this should be a follow up by someone in Cirq team?

mpharrigan: Even with languages with overloading, the functions have consistent return types, which is my issue with the current proposed change

mrwojtek: You're right. I named the new method `prepare_characterizations_for_circuits_moments`, how does it sound?

mpharrigan (approved):
I think this is better. When the dust settles we should make like a table with all the request-prep functions as rows and different features/use cases as the columns like the amazon "compare similar products" tables

PR #4063, opened 2021-04-28
Speed up MPS simulator test

dabacon:
This was create 288 circuits and running 4 simulations per circuit.  The test took >9 seconds.  I've replaced this with random circuits and reduced the number to of circuits to 5 to get this to around 2 seconds.

I don't think this test was testing anything special about the two qubit circuit, but if so then this won't be a good fix so you should yell.


dabacon:
@tonybruguier who I believe wrote the MPS simulator code

tonybruguier:
LGTM.

balopat:
Yay, thank you, this lived in the back of my mind for a while now, thanks for following through with it!!!

balopat (approved):

Issue #3975, opened 2021-03-27
NumPy deprecation warnings


unaiic:
**Description of the issue**
When running Pytest checks, NumPy (version 1.20) throws some deprecation warnings, [here ](https://github.com/quantumlib/Cirq/pull/3974/checks?check_run_id=2208712674)for instance.

We could update our sources to keep them updated, assuring that the suggested changes are compatible with older versions of NumPy.

dabacon:
#4072

dabacon:
Think I have squashed them all.  Qiskit currently has the remaining ones I see.

PR #4065, opened 2021-04-28
Fix numpy warnings

dabacon:
Fixes a bunch of numpy warnings.  These were all about `np.bool` and `np.complex` and `np.int` being deprecated as they are just direct aliases of `bool`, `complex`, and `int`.

One interesting case is json test where I kept the actual numpy scalar type using `np.bool_`.


dabacon:
Fixes #3975

cirq-core/cirq/protocols/json_serialization_test.py
```diff
@@ -232,7 +232,7 @@ def test_builtins():
 def test_numpy():
     x = np.ones(1)[0]
 
-    assert_json_roundtrip_works(x.astype(np.bool))
+    assert_json_roundtrip_works(x.astype(np.bool_))
```
balopat, line 235: maybe add a comment why this is interesting? It seems like a private type.

balopat (approved):
LGTM with a request for a comment on `np.bool_`

cirq-core/cirq/protocols/json_serialization_test.py
dabacon: I read more, and I don't think we need this check, the objects we will get will al be dtype bool, so we should be same.

PR #4067, opened 2021-04-29
WIP: Speed up notebook tests

dabacon:
testing out using papermill's parameterized cell to speed up notebook tests.


dabacon:
Currently annoyed as tensorflow's nb formatting removes all metadata, and this in turn makes test barf as it has no language specified.  Unclear why this currently doesn't cause a problem.

dabacon:
It also removes the tags in cells which means that this approach is not viable.

balopat:
> It also removes the tags in cells which means that this approach is not viable.

Hmmm...I'll have a look if this is surmountable somehow - we know the people who maintain it, so maybe we can make some changes here!

dabacon:
https://github.com/tensorflow/docs/blob/9672f39b59fc7a71fc5bfbd9bcb1135ced5d05e3/tools/tensorflow_docs/tools/nbfmt/__main__.py#L207
just add "parameters" :)

dabacon:
Alternatively we could hot swap that line into the one downloaded for the test ;)

dabacon:
Closing for now will track back on main issue

PR #4069, opened 2021-04-29
Fix flakey test in engine_simulator_test

### LINKED ISSUE #4068: Engine simulator test is flaky.
If you install pytest-repeat and check for, say a couple hundred runs, you will see that
```
./check/pytest cirq-google/cirq_google/calibration/engine_simulator_test.py::test_with_random_gaussian_runs_correctly --count=1000
```
is flakey.

dabacon:
Fixes #4068 

Also removes type annotations for test methods.  These don't really add anything, since return types for vanilla tests are always `None` and never consumed by anything.


dabacon:
To verify install `pytest-repeat` and pass `--count=1000` to pytest when testing this method.

balopat:
Thank you, this is great, let's kill all the flakes!!! :) 

I confirm - `check/pytest cirq-google/cirq_google/calibration/engine_simulator_test.py --count=1000 -k test_with_random_gaussian_runs_correctly` fails 1/1000 times on my mac and passes when atol=0.1 is set... as this test is probabilistic, this only pushes the probability of failure further down the line. I'm not sure I like this test but I don't want to spend too much time investigating it now. If it gives us more grief we should consider rephrasing what we are actually testing here...

What makes me a bit uncomfortable is that one failure scenario was this:

```
(array([0.47795, 0.47095, 0.47795, 0.57315]), array([0.49825, 0.50045, 0.50615, 0.49515]), atol=0.075)
```
0.1 is %20 deviation which seems maybe too big to be meaningful, but ¯\_(ツ)_/¯ maybe this is okay for now, it does test that it runs at least :)

balopat (approved):

PR #4071, opened 2021-04-29
pylint recognizes cirq_google as known third_party

balopat:
Fixes three things: 

* check/pylint was pointing to the wrong folders (platform -> cirq-google/cirq_google)
* pylint (and the underlying isort) needs to know about cirq_google and cirq as not third_party libraries for import sorting to work properly
* there was an unused import in cirq_google that is now removed


mpharrigan (approved):
Sweet!

PR #4077, opened 2021-05-01
Speed up notebook tests by adding extra replacements file

dabacon:
Provides the ability to speed up notebook execution tests by substituting smaller values for parameters that impact the runtime of the notebook.  To do this one supplies a file for a given `.ipynb` test with the same prefix but the `.tst` suffix.  This file contains replacements of the form `pattern->replacement`.  Some notes

* Suffix `.tst` chosen instead of `.test` because the later has the potential to make some grepping for tests less reliable and more noise prone
* All the patterns must be applied during the test. This avoids the common case of a pattern being a typo and not being applied.
* Open to different syntax than `->`.  Used it because it seemed unlikely to conflict with actual patterns and was indicative of what the replacement is actually doing.  

Notebook tests are taking up to 30 minutes in Google Actions.  This PR gets that down to around 11.  Nearly as fast as the Windows test :) 

Addresses #3603

### DIFF
--- a/cirq-core/cirq/contrib/quimb/Cirq-To-Tensor-Networks.tst
+++ b/cirq-core/cirq/contrib/quimb/Cirq-To-Tensor-Networks.tst
@@ -12,8 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from dev_tools.notebooks.utils import (
-    filter_notebooks,
-    list_all_notebooks,
-    rewrite_notebook,
-)
+# Replacements to apply during testing. See devtools/notebook_test.py for syntax.
+
+max_qubits = 6->max_qubits = 4
+max_moments = 500->max_moments = 10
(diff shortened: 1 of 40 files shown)


balopat (approved):
This is fantastic, thank you for working through this! Highly impactful, will speed up feedback time on all our PRs!

mpharrigan:
Is there a reason why you chose a bespoke `->` syntax? Did you consider doing a regex style `s/pattern/replacement`? 

In that vein, it's probably pretty important to note that the left and right parts of `->` do indeed get compiled to a regular expression so if you do something like `5.1->5` it would also replace `500` or `5x0` which might be very surprising :)

dabacon:
I did consider it.  One goal was to keep this very simple, I think the goal should not be do doing anything regex in the pattern portion of the code.  So almost everything should be text->text.   In fact I considered not making this a pattern match in that it would do only a substring match. 

`s/pattern/replacement` comes from sed, right?  Is that something we think is widely known?

mpharrigan:
What do you think about my second point that it *does* behave as a regex with potentially surprising behavior for "normal" characters like `.`, `+`, `()`, and others

PR #4084, opened 2021-05-04
Fix tetris_concat docstring align arg and restore FIRST alignment option

Strilanc:
- The docstring was still talking about the arg that was removed months ago
- FIRST alignment is useful when performing bulk operations

### DIFF
--- a/cirq-core/cirq/circuits/circuit.py
+++ b/cirq-core/cirq/circuits/circuit.py
@@ -1364,7 +1364,10 @@ class AbstractCircuit(abc.ABC):
             align: When to stop when sliding the circuits together.
                 'left': Stop when the starts of the circuits align.
                 'right': Stop when the ends of the circuits align.
-                'first': Stop the first time either the starts or ends align.
+                'first': Stop the first time either the starts or the ends align. Circuits
+                    are never overlapped more than needed to align their starts (in case
+                    the left circuit is smaller) or to align their ends (in case the right
+                    circuit is smaller)
 
         Returns:
             The concatenated and overlapped circuit.
--- a/cirq-core/cirq/circuits/circuit.py
+++ b/cirq-core/cirq/circuits/circuit.py
@@ -70,8 +70,12 @@ INT_TYPE = Union[int, np.integer]
 
 
 class Alignment(enum.Enum):
+    # Stop when left ends are lined up.
     LEFT = 1
+    # Stop when right ends are lined up.
     RIGHT = 2
+    # Stop the first time left ends are lined up or right ends are lined up.
+    FIRST = 3
 
     def __repr__(self) -> str:
         return f'cirq.Alignment.{self.name}'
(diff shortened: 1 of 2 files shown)


cirq-core/cirq/circuits/circuit.py
```diff
@@ -1357,13 +1361,10 @@ def tetris_concat(
...
-                are never overlapped more than needed to align their starts (in
-                case the left circuit is smaller) or to align their ends (in
-                case the right circuit is smaller). When false, the smaller
-                circuit can be pushed deeper into the larger circuit, past the
-                first time their starts or ends align, until the second time
-                their starts or ends align.
+            align: When to stop when sliding the circuits together.
+                'left': Stop when the starts of the circuits align.
+                'right': Stop when the ends of the circuits align.
+                'first': Stop the first time either the starts or ends align.
```
balopat, line 1367: can we add the old explanation here? I think it makes it more understandable.

Strilanc: What do you mean? The old explanation is pretty coupled to the argument.

balopat (approved):
LGTM, @maffoo?

cirq-core/cirq/circuits/circuit.py
balopat: Isn't `stop_at_first_alignment = True` is equivalent to `align=FIRST`? Of course I'd rewrite it to cater for the current setup. Something like: 

```
circuits are never overlapped more than needed to align their starts (in case the left circuit is smaller) or to align their ends (in case the right circuit is smaller). 
```

maffoo:
Oh boy, sorry for messing up that tetris_concat docstring in #3821 :-)

I must admit I don't really understand what "first" alignment is trying to do. @Strilanc, can you explain the use case you have in mind? One problem I have is that the explanation for it is in terms of implementation details (the dynamic process of moving circuits past each other and stopping), rather than in terms of the end result it is trying to achieve. It seems pretty clear what it means to align circuits as far left or as far right as possible, however the alignment is implemented; if we had a similar explanation for what "first" is trying to do, I think that'd help users (it'd help me at least :-) ).

Along the same lines, I'm not sure that "first" is a good name for this alignment mode since it is very implementation-specific, referring to the "first" time a criterion is met in a process as currently implemented. If we had a succinct explanation of what result this alignment mode is trying to achieve then I think we could give it a clearer name that isn't so tied to the current implementation.

Strilanc:
@maffoo The specific case I use this for is `tetris_concat(logical_initialize, logical_bulk, logical_measure)` where I want the init to left align and the measure to right align.

maffoo:
I see. It seems to me that
```python
Circuit.tetris_concat(logical_initialize, logical_bulk, logical_measure, align='first')
```
doesn't really convey what the result will look like. What about `OUTER` for this mode instead of `FIRST`? An alternative would be to explicitly call out the alignment on both ends:
```python
Circuit.tetris_concat(logical_initialize, logical_bulk, align='left').tetris_concat(logical_measure, align='right')
```

Strilanc:
OUTER is also a reasonable name.

I'd prefer not to specify each one individually. The circuits aren't actually named in context, they're just intermediate expressions.

daxfohl:
We may want to revisit the name of the function too, lest there are any issues there. Something like shape preserving concat?

PR #4085, opened 2021-05-04
Move CliffordTableau and add unit test

### LINKED ISSUE #4075: Move Clifford Tableau Out of cirq/sim/clifford folder
**Description of the issue**

Clifford Tableau is used in the Clifford simulator. Except the simulation usage, Clifford tableau can also be used to represent a general n-qubit Clifford gates, which is not supported in cirq yet. This means we need to import tableau in the ops files to define general clifford gates. However, to avoid circular import issue, `cirq/sim` is not supposed to be imported in the top level of `cirq/ops` files. As discussed in cirq-cynq meeting, the easiest solution is move the Clifford Tableau out of cirq/sim folder. (Partially, also because Clifford Tableau itself is just a state representation not simulation.)

Question is where should we move it into? `cirq/qis` or `cirq/value` seems to be the good place.

ybc1991:
Fix #4075.

To avoid breakage, the original `cirq/sim/clifford/CliffordTableau` is untouched added with deprecated notation.
All internal import of Cliffod Tableau is pointed to the new position under `cirq/qis`. (Unless people have strong opinions of putting it into other places.)

New clifford_tableau is the same as the old one except in one place -- the `xs`, `zs`, and `rs` are protected by getter and setter pattern and do not expose the last scratch row.


cirq-core/cirq/sim/clifford/clifford_tableau.py
```diff
@@ -17,10 +17,12 @@
...
 from cirq import protocols
+from cirq._compat import deprecated_class
 from cirq.ops.dense_pauli_string import DensePauliString
 from cirq.value import big_endian_int_to_digits
 
 
+@deprecated_class(deadline='v0.12', fix='Import CliffordTableau from cirq or cirq.qis instead.')
 class CliffordTableau:
     """Tableau representation of a stabilizer state
     (based on Aaronson and Gottesman 2006).
```
balopat, line 28: You should be able to remove the implementation and just make it equal to cirq.qis.CliffordTableau.

cirq-core/cirq/sim/clifford/clifford_tableau.py
```diff
@@ -17,10 +17,12 @@
 
 import cirq
 from cirq import protocols
+from cirq._compat import deprecated_class
 from cirq.ops.dense_pauli_string import DensePauliString
 from cirq.value import big_endian_int_to_digits
 
 
+@deprecated_class(deadline='v0.12', fix='Import CliffordTableau from cirq or cirq.qis instead.')
```
balopat, line 25:
```suggestion
@deprecated_class(deadline='v0.12', fix='use cirq.CliffordTableau instead')
```

balopat (changes requested):
Thank you! 
 I added some comments + we need to add a test for the deprecated behavior - see unit test examples where `cirq.testing.assert_deprecated` is used.

cirq-core/cirq/sim/clifford/clifford_tableau.py
ybc1991: Done.

cirq-core/cirq/sim/clifford/clifford_tableau.py
ybc1991: Done.

ybc1991:
Thanks. Added the assert_deprecated. PTAL.

balopat (approved):
LGTM, beautiful! Thanks for writing the great tests for it!

Issue #4040, opened 2021-04-21
Structured Measurement Keys


smitsanghavi:
A class that wraps the string measurement key, encapsulating any measurement key related logic and provides various utils for users. Phased rollout to eventually move away from manipulating strings to a well defined interface.

**Is your design idea/issue related to a use case or problem? Please describe.**
Currently, the measurement keys are plain strings. We need more structure and logic to the key itself to have a cleaner way to:

* Validate the keys (some characters are reserved to mean special things for e.g. ':' for nesting inside a `CircuitOperation`)
* Treating measurement keys created from `Qubits` as special type. Remapping qubits for a `MeasurementGate` should also remap these keys.
* Eliminate a lot of the brittle string parsing logic inside `CircuitOperation` and enable accessing different "levels" of a nested `CircuitOperation` more cleanly.
* Unlock more expressive implementations for [Representing Classical Data](#3231), [Feed Forward](#3232) and [Flow Control](#3234)

**Describe your design idea/issue**
RFC: https://tinyurl.com/structured-measurement-keys
The initial part outlines potential approaches to go about this. The "Detailed Design" goes deeper into the preferred option (based on feedback from cirq-maintainers@) and the implementation/rollout plan for it.

dabacon:
Is this done?

daxfohl:
I think there was still a remaining task to rename and deprecate some protocols. While doing that we may want to see if we can speed up the default case for these protocols where nothing is implemented, perhaps removing some magic method options, as the protocols do get called in tight loops during circuit construction.

95-martin-orion:
@daxfohl, could you elaborate on the remaining work? IIRC, `measurement_key_(objs|names)` was the intended end state for the protocol renaming. I'm on board for optimizations, but those can be a separate post-1.0 issue.

daxfohl:
Oh, I thought you wanted to get rid of the objs suffix. If not then it
sounds like this is done.

95-martin-orion:
Opened #5465 to track optimizations; closing this as complete.

PR #4086, opened 2021-05-04
Use `MeasurementKey` in `CircuitOperation`

smitsanghavi:
Phase 2a of #4040 RFC.

* Added a concept of key `path`. Similar to file path, denotes the path to this particular key in the circuit and distinguishes from other keys with the same name in other paths.
* Added a `with_key_path` protocol and the required magic methods to make it work across Cirq.
* Used these in `CircuitOperation` to simplify the logic and remove the need for any string parsing.
* Since the `measurement_keys` protocol still speaks `str` rather than `MeasurementKey` we need to serialize to and from string at the boundaries of this protocol, leading to some string parsing. These are aims to be removed in future phases by introducing an equivalent protocol API with `MeasurementKey` representations, rather than string.
* `MeasurementKey.parse_serialized` helps in constructing the objects at the above boundaries and is the only way to parse nested key string, directly passing them to the constructor will raise `ValueError`. This is to avoid users from accidentally using the separator character in their keys, thereby confusing our logic that relies on a specific meaning of that character.
* Change the character used to separate the crosses of repetition ids (but not nesting) to `-`. For e.g. `subcircuit.repeat('a', 'b').repeat('c', 'd')` used to create a non-nested `CircuitOperation` with `repetition_ids=['a:c', 'a:d', 'b:c', 'b:d']`. But since we want to reserve the character `:` for separating different nesting levels, we will now have `repetition_ids=['a-c', 'a-d', 'b-c', 'b-d']`.
* Also had to change the JSON representations due to the addition of `path` and added new test cases there as well.


cirq-core/cirq/protocols/json_test_data/CircuitOperation.json
```diff
@@ -205,6 +205,7 @@
           "cirq_type": "ParamResolver",
           "param_dict": []
         },
+        "parent_path": [],
```
95-martin-orion, line 208: Is this JSON format backwards-compatible? i.e. can objects serialized in the old format still be deserialized in the new format? Full serialization guidelines for reference: [doc](https://github.com/quantumlib/Cirq/blob/master/docs/dev/serialization.md)

Same for the `MeasurementGate` JSON changes.

cirq-core/cirq/circuits/circuit_operation.py
```diff
@@ -40,40 +40,18 @@ def default_repetition_ids(repetitions: int) -> Optional[List[str]]:
     return None
 
 
-def cartesian_product_of_string_lists(list1: Optional[List[str]], list2: Optional[List[str]]):
+def full_join_string_lists(list1: Optional[List[str]], list2: Optional[List[str]]):
```
95-martin-orion, line 43: Should this be a private method (i.e. with leading "_")?

95-martin-orion (changes requested):
A few notes on this but otherwise it looks very clean.

cirq-core/cirq/protocols/json_test_data/CircuitOperation.json
smitsanghavi: Thanks for catching this. Moved old JSON serializations to `_inward` file and made the `parent_path` arg optional in `_from_json_dict_`.
`MeasurementGate` changes are backwards compatible. Not `_inward`-ing `MeasurementKey` JSON changes since no code or user serializes it to JSON yet.

cirq-core/cirq/ops/moment_test.py
```diff
@@ -279,16 +279,20 @@ def test_with_measurement_keys():
     m = cirq.Moment(cirq.measure(a, key='m1'), cirq.measure(b, key='m2'))
 
     new_moment = cirq.with_measurement_key_mapping(
-        m,
+        cirq.with_key_path(m, ('a', 'b')),
```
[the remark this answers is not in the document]
smitsanghavi: Done

cirq-core/cirq/ops/measurement_gate_test.py
```diff
@@ -23,6 +23,7 @@ def test_measure_init(num_qubits):
     assert cirq.MeasurementGate(num_qubits).num_qubits() == num_qubits
     assert cirq.MeasurementGate(num_qubits, key='a').key == 'a'
     assert cirq.MeasurementGate(num_qubits, key='a').mkey == cirq.MeasurementKey('a')
+    assert cirq.MeasurementGate(num_qubits, key=cirq.MeasurementKey('a')).key == 'a'
```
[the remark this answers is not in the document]
smitsanghavi: Done

cirq-core/cirq/circuits/circuit_operation.py
smitsanghavi: The test uses it. Not sure what the python convention for this is. Is it okay for unit tests to depend on private methods?

cirq-core/cirq/protocols/json_test_data/CircuitOperation.json
95-martin-orion: LGTM.

95-martin-orion:
After converting `full_join_string_lists` to private, this should be good to go. Holding off on approval to allow for requested review from @balopat - let me know if you'd like to go ahead without that review.

smitsanghavi:
> After converting `full_join_string_lists` to private, this should be good to go. Holding off on approval to allow for requested review from @balopat - let me know if you'd like to go ahead without that review.

Done. 

@balopat seems to be from last week's Cirq sync when I was trying to demonstrate the issue with adding reviewers. More reviews are definitely welcome but I'd be good with merging this as well.

PR #4090, opened 2021-05-06
[XEB] Multiprocessing spawn

mpharrigan:
There are two multiprocessing "modes": fork and spawn. Fork used to always be the default. Now sometimes on Mac it defaults to spawn (?) but in any event: spawn works way better. This could prevent some of the c++-looking errors that @mrwojtek has seen. It fixed different-but-still-c++-looking-errors for me


mrwojtek:
I'll test this one if it solves my isseus.

mrwojtek (approved):

Strilanc (approved):

PR #4092, opened 2021-05-07
[XEB] Allow default angles to be inferred from a gate

mpharrigan:
Instead of defaulting angles to 0.0 -- which is a very sharp edge that can cause optimization to fail -- default to `None` and either require the user fully specify the angle settings or provide a gate to get sensible defaults from. In `xeb_wrapper` this gate is already around as the calibration request gate.


cirq-core/cirq/experiments/xeb_fitting.py
```diff
@@ -136,6 +137,38 @@ def get_initial_simplex_and_names(
...
+    for the five PhasedFSim angles."""
+    defaults = {
+        'theta_default': 0.0,
+        'zeta_default': 0.0,
+        'chi_default': 0.0,
+        'gamma_default': 0.0,
+        'phi_default': 0.0,
+    }
+    if gate == SQRT_ISWAP:
+        defaults['theta_default'] = -np.pi / 4
```
mrwojtek, line 151: I think we should set also phi_default to np.pi / 24 here and below. This should improve the optimization speed? I'll try this setting as soon I have a chance. We can change this in a follow-up?

mrwojtek (approved):

cirq-core/cirq/experiments/xeb_fitting.py
mpharrigan: I think this would be a dangerous semantic to introduce in `cirq-core`, which is where this function currently lives. Please be aware that if you set `characterize_phi=False`, the fixed value will be set to `phi_default`. I don't know if this would help or hurt when you're only optimizing phases.

Options:
1) [my preference] have a wrapper like `SqrtISwapXEBOptions` (`GoogleSqrtISwapXEBOptions`) that sets {angle}_default values and tell users to use it instead of the automatically-determine-from-gate when you're using SqrtIswap on a google processor because it can help with phi characterization
2) Inject a different `phased_fsim_angles_from_gate` if you're running via `cirq_google.workflow`
3) [radical] Introduce a new sqrt-iswap-like gate defined to ideally have phi of np.pi/24 and have people write their algorithms with this gate. Then the angle determination would work automatically. You could still treat it as an iswap gate but you'd have to explicitly neglect the small conditional phase

mpharrigan:
I manually checked notebooks as well since I was suspicious they didn't require any changes, but they all use `SqrtISwapOptions` which is fully backwards compatible. Note: this is a breaking change if you were previously only supplying some of the angles when manually constructing `PhasedFSimCharacterizationOptions`.

cirq-core/cirq/experiments/xeb_fitting.py
mrwojtek: I like option 2 but maybe let's play with this future a little bit if that would make any difference at all?

95-martin-orion (approved):

cirq-core/cirq/experiments/xeb_fitting.py
mpharrigan: I included the gate-angle-inference function as an argument to the method to support option 2

PR #4094, opened 2021-05-07
Provide sweep-iterator API.

### LINKED ISSUE #4064: Return `Iterator` from simulator sweep methods
Simulator methods that return multiple results can be difficult to work with, since each result object may contains a massive representation of the simulator state. An example of this can be seen in #3979 and #3990, where the `simulate_expectation_value_sweep` method can't simply invoke the `simulate_sweep` method for fear of generating (and storing!) all of the output states simultaneously.

To resolve this, we should modify the simulator `*_sweep` methods to return `Iterator` objects instead of `List`s.

This would also cause sweeps to run one simulation at a time, which reduces the "black box" delay of large sweeps, and may provide opportunities for parallelizing the simulation and result-processing pipelines.

95-martin-orion:
Fixes #4064.

This PR provides `*_iter` methods for all simulator "sweep" methods. External simulators which implement the old list-based APIs will still function properly (through the magic of `@value.alternative`), but these simulators are advised to switch to the iterator-based API for better performance.

Note that the list-based API is _not_ deprecated, and we have no plans to remove it at this time.


95-martin-orion:
Paging @daxfohl for review, since this touches on the ongoing simulator refactor.

95-martin-orion:
Looks like I need to appease https://github.com/python/typing/issues/449 for python3.6 compatibility.

daxfohl:
Everything looks fine to me.

balopat (approved):
LGTM

PR #4095, opened 2021-05-09
Ensure 3 qubit gate decomposition returns unique CNOTs

madhavkrishnan:
This PR is a fix for the decompose method of three qubit gates returning multiple CNOTs which are the same object.

For example [CCZPowGate.\__decompose\__](https://github.com/quantumlib/Cirq/blob/master/cirq-core/cirq/ops/three_qubit_gates.py#L77)  returns an OP_TREE that leads to the following circuit :
```python
 0: ───p───@──────────────@───────@──────────@──────────
           │              │       │          │
 1: ───p───X───@───p^-1───X───@───X──────@───X──────@───
               │              │          │          │
 2: ───p───────X───p──────────X───p^-1───X───p^-1───X───
```
The unexpected behaviour is that there are only two unique CNOTs in the list returned. All the CNOTs between qubits 0 and 1 are the same object. This is also true for all the CNOTs between qubits 1 and 2. Looking at the [source code](https://github.com/quantumlib/Cirq/blob/master/cirq-core/cirq/ops/three_qubit_gates.py#L100), it is clear why this is the case -

```python
p = common_gates.T ** self._exponent
sweep_abc = [common_gates.CNOT(a, b), common_gates.CNOT(b, c)]

return [
      p(a),
      p(b),
      p(c),
      sweep_abc,
      p(b) ** -1,
      p(c),
      sweep_abc,
      p(c) ** -1,
      sweep_abc,
      p(c) ** -1,
      sweep_abc,
]
```
I don't think this can be desired behavior or all circuits would just have one object for an operation acting on a particular set of qubits.

This leads to problems in decomposition methods for example if you try to do a custom decomposition of the operations returned by this method there is no way to distinguish the pairs of CNOTs in the decomposition above since they are the same object.

It looks like this behaviour was introduced in an attempt to optimise three qubit gates in this issue quantumlib/Cirq#814.

### DIFF
--- a/cirq-core/cirq/ops/three_qubit_gates_test.py
+++ b/cirq-core/cirq/ops/three_qubit_gates_test.py
@@ -330,9 +330,8 @@ def unique_ops(op_list):
     """
 
     unique = []
-    while op_list:
-        op = op_list.pop(0)
-        if not sum([op is ele for ele in unique]):
+    for op in op_list:
+        if all(op is not elem for elem in unique):
             unique.append(op)
 
     return unique
--- a/cirq-core/cirq/ops/three_qubit_gates_test.py
+++ b/cirq-core/cirq/ops/three_qubit_gates_test.py
@@ -319,3 +319,55 @@ def test_resolve(resolve_fn):
     diagonal_gate = resolve_fn(diagonal_gate, {'b': 19})
     assert diagonal_gate == cirq.ThreeQubitDiagonalGate(diagonal_angles)
     assert not cirq.is_parameterized(diagonal_gate)
+
+def unique_ops(op_list):
+    """
+    Helper function for test_unique_gate_decomp.
+
+    Returns a list of unique ops from a list. Unique in the sense of op1 is op2.
+    The original list is used up by this function
+    """
+
+    unique = []
+    while op_list:
+        op = op_list.pop(0)
+        if not sum([op is ele for ele in unique]):
+            unique.append(op)
+
+    return unique
+
+def test_unique_gate_decomp():
+    """
+    Test to check that 3-qubit decompose methods return unique operations
+
+    The test is checking that no two gates in the decomposition
+    refer to the same object.
+    """
+    qubits = [cirq.GridQubit(i, 0) for i in range(3)]
+
+    ccz = cirq.CCZ.on(*qubits)
+    ccz_decomp = ccz._decompose_()
+    ccz_decomp_count = len(ccz_decomp)
+
+    ccx = cirq.CCX.on(*qubits)
+    ccx_decomp = ccx._decompose_()
+    ccx_decomp_count = len(ccx_decomp)
+
+    cs = cirq.CSWAP.on(*qubits)
+    cs_decomp = cs._decompose_()
+    cs_decomp_count = len(cs_decomp)
+
+    diagonal_angles = [2, 3, 5, 7, 11, 13, 17, 19]
+    tqd = cirq.ThreeQubitDiagonalGate(diagonal_angles)
+    tqd_decomp = tqd._decompose_(qubits)
+    tqd_decomp_count = len(tqd_decomp)
+
+    ccz_unique_count = len(unique_ops(ccz_decomp))
+    ccx_unique_count = len(unique_ops(ccx_decomp))
+    cs_unique_count = len(unique_ops(cs_decomp))
+    tqd_unique_count = len(unique_ops(tqd_decomp))
+
+    assert ccz_unique_count == ccz_decomp_count
+    assert ccx_unique_count == ccx_decomp_count
+    assert cs_unique_count == cs_decomp_count
+    assert tqd_unique_count == tqd_decomp_count
--- a/cirq-core/cirq/ops/three_qubit_gates_test.py
+++ b/cirq-core/cirq/ops/three_qubit_gates_test.py
@@ -320,6 +320,7 @@ def test_resolve(resolve_fn):
     assert diagonal_gate == cirq.ThreeQubitDiagonalGate(diagonal_angles)
     assert not cirq.is_parameterized(diagonal_gate)
 
+
 def unique_ops(op_list):
     """
     Helper function for test_unique_gate_decomp.
@@ -335,6 +336,7 @@ def unique_ops(op_list):
             unique.append(op)
 
     return unique
+
 
 def test_unique_gate_decomp():
     """
--- a/cirq-core/cirq/ops/three_qubit_gates.py
+++ b/cirq-core/cirq/ops/three_qubit_gates.py
@@ -98,20 +98,20 @@ class CCZPowGate(
                 a, b = b, a
 
         p = common_gates.T ** self._exponent
-        sweep_abc = [common_gates.CNOT(a, b), common_gates.CNOT(b, c)]
+        cnot = common_gates.CNOT
 
         return [
             p(a),
             p(b),
             p(c),
-            sweep_abc,
+            [cnot(a,b), cnot(b,c)],
             p(b) ** -1,
             p(c),
-            sweep_abc,
+            [cnot(a,b), cnot(b,c)],
             p(c) ** -1,
-            sweep_abc,
+            [cnot(a,b), cnot(b,c)],
             p(c) ** -1,
-            sweep_abc,
+            [cnot(a,b), cnot(b,c)],
         ]
 
     def _apply_unitary_(self, args: 'protocols.ApplyUnitaryArgs') -> np.ndarray:
@@ -266,7 +266,6 @@ class ThreeQubitDiagonalGate(gate_features.ThreeQubitGate):
                 b, c = c, b
             elif not b.is_adjacent(c):
                 a, b = b, a
-        sweep_abc = [common_gates.CNOT(a, b), common_gates.CNOT(b, c)]
         phase_matrix_inverse = 0.25 * np.array(
             [
                 [-1, -1, -1, 1, 1, 1, 1],
@@ -283,19 +282,19 @@ class ThreeQubitDiagonalGate(gate_features.ThreeQubitGate):
         ]
         phase_solutions = phase_matrix_inverse.dot(shifted_angles_tail)
         p_gates = [pauli_gates.Z ** (solution / np.pi) for solution in phase_solutions]
-
+        cnot = common_gates.CNOT
         return [
             p_gates[0](a),
             p_gates[1](b),
             p_gates[2](c),
-            sweep_abc,
+            [cnot(a,b), cnot(b,c)],
             p_gates[3](b),
             p_gates[4](c),
-            sweep_abc,
+            [cnot(a,b), cnot(b,c)],
             p_gates[5](c),
-            sweep_abc,
+            [cnot(a,b), cnot(b,c)],
             p_gates[6](c),
-            sweep_abc,
+            [cnot(a,b), cnot(b,c)],
         ]
 
     def _value_equality_values_(self):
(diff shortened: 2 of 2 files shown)


95-martin-orion (changes requested):
Thanks for this fix! Could you also add tests in `three_qubit_gates_test.py` that verify this works as expected? (i.e. that the first CNOT `is not` the second CNOT, etc.)

madhavkrishnan:
I have added a test to check that all 3-qubit gate decompose methods return unique objects. I thought this is a more robust test.

madhavkrishnan:
I ran ```check/format-incremental --apply HEAD^``` as you suggested and added a new commit. Should I merge the master branch since it has gone ahead?

95-martin-orion:
The new format check failure is my bad - since `three_qubit_gates.py` was not changed in the latest commit, you'll need to run `check/format-incremental --apply` (without the `HEAD^`) to catch the older changes.

> Should I merge the master branch since it has gone ahead?

We have automation tools to handle this; in general, it's only necessary to merge master manually if there are merge conflicts, or if your PR has fallen significantly out of date. Thanks for checking!

cirq-core/cirq/ops/three_qubit_gates_test.py
```diff
@@ -319,3 +319,55 @@ def test_resolve(resolve_fn):
...
+
+    Returns a list of unique ops from a list. Unique in the sense of op1 is op2.
+    The original list is used up by this function
+    """
+
+    unique = []
+    while op_list:
+        op = op_list.pop(0)
+        if not sum([op is ele for ele in unique]):
+            unique.append(op)
```
maffoo, line 336: nit:
```suggestion
    for op in op_list:
        if op not in unique:
            unique.append(op)
```

maffoo:
Drive-by comment, but could you elaborate a bit on the problem this is trying to fix? Generally speaking reusing the same gate instance in more than one part of a circuit should be fine, and I would think if there are optimizers or decompositions that rely on having distinct gate instances that would probably be a bug in the optimizer or decomposition; they should look at an operation _and_ it's position in the circuit and not rely on the identity of the gate object itself. Could you give a more concrete example where reusing gate instances fails?

cirq-core/cirq/ops/three_qubit_gates_test.py
madhavkrishnan: Done.

madhavkrishnan:
> The new format check failure is my bad - since `three_qubit_gates.py` was not changed in the latest commit, you'll need to run `check/format-incremental --apply` (without the `HEAD^`) to catch the older changes.

Done. It should clear the format check now.

95-martin-orion:
@madhavkrishnan, could you provide an answer to @maffoo's comment? This PR prevents reuse of an operation, not a gate, but his question is still valid: `GateOperation` is [treated as immutable](https://github.com/quantumlib/Cirq/blob/f7c1156c79b1ca12450b9ce8859003e95f932196/cirq-core/cirq/ops/gate_operation_test.py#L36) and should therefore be safe to reuse.

balopat:
Thanks for the drive-by comment @maffoo :) We should have caught this in the gitter chat, sorry @madhavkrishnan - as @95-martin-orion says operations are considered immutable, they are just values. I'd be curious to hear if your compiler use case that this might still break.

madhavkrishnan:
Thanks for the comments everyone. 

My initial thought process about this which I mentioned in my first comment was that if just references to the gate object was sufficient, wouldn't the entire circuit be designed like this to save memory resources? For example I could potentially describe circuits with extremely large depth with just a list of references. I didn't really think about this further.

The compiler I am using adds labels to the operations which are then used to guide the decomposition aglorithm. This is not possible if the objects are not unique.

maffoo:
The supported way to add labels to specific operations is to use the "tags" mechanism. For example, you can do `op.with_tags(label)` to add a label to a given operation. This doesn't mutate the op but instead creates a new instance, so each tagged operation object is unique (though again, if you wanted to create an operation with a tag and then reuse it, that would be ok).

I do think the current mix of mutable and immutable objects in circuits is not ideal, in particular that circuits are mutable but moments are immutable so you have to be careful about replacing ops in a moment because you can get quadratic overheads from creating moment objects over and over. Is your compiler open source? If you have example code then perhaps we can help understand your use case better and how to best use the current cirq APIs or improve them.

madhavkrishnan:
Apologies for the long comment.

@maffoo I am building on the compiler [here](https://github.com/alexandrupaler/fondq/tree/master/icm). To give you a quick summary of the source code -

A new class of qubits are defined to help add ancillas and keep track of which operation decomposition lead to that ancilla

```python
class SplitQubit(cirq.NamedQubit):
    # Static nr_ancilla
    nr_ancilla = -1

    def __init__(self, name):

        super().__init__(name)

        # A qubit/wire is split in two and these are the resulting wires
        self.children = (None, None)

        # The decision on which wire to use is based on the id of the operation
        # that generated the split
        self.threshold = opid.OperationId()

    def get_latest_ref(self, operation_id):

        # this wire has not been split
        if self.children == (None, None):
            return self

        n_ref = self

        while n_ref.children != (None, None):
            # Decide based on the threshold
            if n_ref.threshold >= operation_id:
                n_ref = n_ref.children[0]
            else:
                n_ref = n_ref.children[1]

        return n_ref

    def split_this_wire(self, operation_id):

        # It can happen that the reference is too old
        current_wire = self.get_latest_ref(operation_id)

        # The wire receives a threshold for latter updates
        current_wire.threshold = operation_id

        # It is a new wire, but keep the old name
        n_child_0 = SplitQubit(current_wire.name)

        # It is a new wire, that is introduced and gets a new name
        SplitQubit.nr_ancilla += 1
        n_child_1 = SplitQubit("_anc_{0}".format(SplitQubit.nr_ancilla))

        # Update the children tuple of this wire
        current_wire.children = (n_child_0, n_child_1)

        # Return the children as a tuple
        return current_wire.children
```

A class of flag objects called OperationId are defined with

```python


class OperationId():

    DEFAULT_NOT_INIT = -1

    def __init__(self, numbers_tuple = None):

        if numbers_tuple is None:
            numbers_tuple = (OperationId.DEFAULT_NOT_INIT, )

        if not isinstance(numbers_tuple, tuple):
            # cast to int
            x = int(numbers_tuple)
            numbers_tuple = (x, )

        self.numbers = tuple(numbers_tuple)

    def add_decomp_level(self):
        # append a zero and return a new ID
        return OperationId(self.numbers + (0,))

    def advance_decomp(self):
        # return
        last_number = self.numbers[-1]
        return OperationId(self.numbers[:-1] + (last_number + 1, ))

    def _cmp_tuple(self):
        return self.numbers

    def __eq__(self, other):
        if not isinstance(other, OperationId):
            return NotImplemented
        return self._cmp_tuple() == other._cmp_tuple()

    def __ne__(self, other):
        if not isinstance(other, OperationId):
            return NotImplemented
        return self._cmp_tuple() != other._cmp_tuple()

    def __lt__(self, other):
        if not isinstance(other, OperationId):
            return NotImplemented
        return self._cmp_tuple() < other._cmp_tuple()

    def __gt__(self, other):
        if not isinstance(other, OperationId):
            return NotImplemented
        return self._cmp_tuple() > other._cmp_tuple()

    def __le__(self, other):
        if not isinstance(other, OperationId):
            return NotImplemented
        return self._cmp_tuple() <= other._cmp_tuple()

    def __ge__(self, other):
        if not isinstance(other, OperationId):
            return NotImplemented
        return self._cmp_tuple() >= other._cmp_tuple()

``` 

Methods for manipulating these flags are also defined.

```python

from icm.icm_operation_id import OperationId

def is_op_with_op_id(op, gate_types):
    if op.gate in gate_types:
        return hasattr(op, "icm_op_id")
    return False

def reset_op_ids(circuit, gate_types):
    remove_decomposition_flags(circuit, gate_types)
    add_decomposition_flags(circuit, gate_types)

def add_op_ids(circuit, gate_types):
    nr_op = 0
        if not is_op_with_op_id(op, gate_types):
            setattr(op, "icm_op_id", OperationId(nr_op))
            # increase the id
            nr_op += 1
def remove_op_ids(circuit, gate_types):
        if is_op_with_op_id(op, gate_types):
            delattr(op, "icm_op_id")
```

The intercepting decomposer is defined as below (right now it replaces all single qubit gates with a CNOT acting on an ancilla + a measurement on the original qubit, this is a placeholder for the actual decompositions).

```python

"""
NOTE: Experimental code for testing how gate decomposition that introducesintroduces ancillas would look like in
Cirq. For this the ICM form of circuits is being implemented. As described in "Faster manipulation of large
quantum circuits using wire label reference diagrams" https://arxiv.org/abs/1811.06011
"""

import cirq

from icm.split_qubit import SplitQubitfrom icm.icm_operation_id import OperationId

def decomp_to_icm(cirq_operation):
    """
    :param cirq_operation:
    :return:
    """
    new_op_id = cirq_operation.icm_op_id.add_decomp_level()

    # Assume for the moment that these are only single qubit operations
    new_wires = cirq_operation.qubits[0].split_this_wire(new_op_id)

    """
        In this version the type of the gate, and the initialisation
         of the qubits is not considered.
    """
    # Create the cnot
    cnot = cirq.CNOT(new_wires[0], new_wires[1])

    # Assign a decomposition id, like [old].1
    cnot.icm_op_id = new_op_id.advance_decomp()

    # Create the measurement
    meas = cirq.measure(new_wires[0])

    # Because this operation follows the CNOT, has ID from the previous
    # results into something like  [oldid].2
    meas.icm_op_id = cnot.icm_op_id.advance_decomp()

    return [cnot, meas]

def keep_icm(cirq_operation):
    """
    :param cirq_operation:
    :return:
    """

    """
        Decompose if the operation is from the set of the ones to keep
    """

    if isinstance(cirq_operation.gate, (cirq.CNotPowGate, cirq.MeasurementGate)):
        return True

    """
        Keep the operation if:
    
    * this is an operation that should be decomposed
                 AND
    * is not marked for decomposition
    """

    if not flags.is_op_with_op_id(cirq_operation, [cirq_operation.gate]):
        return True

    return False
```

Here is an example of how the compiler would work

```python
import icm
import cirq

a = icm.SplitQubit("a")
b = icm.SplitQubit("b")
c = icm.SplitQubit("c")

mycircuit = cirq.Circuit(

    cirq.T.on(a), cirq.T.on(b),
    cirq.CNOT.on(a,b), cirq.S.on(a),
    cirq.CNOT.on(b,c), cirq.T.on(c),
)

icm.icm_flag_manipulations.add_op_ids(mycircuit, [cirq.T, cirq.S])

print("\nOriginal Circuit")
print("------------------\n")
print(mycircuit)

icm_circuit = cirq.Circuit(cirq.decompose(mycircuit,
                                          intercepting_decomposer=icm.decomp_to_icm,
                                          keep = icm.keep_icm))

print("\nDecomposed Circuit")
print("-------------------\n")
print(icm_circuit)

```
The output will be :

```python
Original Circuit                      
------------------                    
                                      
a: ───T───@───S───────                
          │                           
b: ───T───X───@───────                
              │                       
c: ───────────X───T───                
                                      
Decomposed Circuit                    
-------------------                   
                                      
           ┌──┐       ┌──┐            
_anc_0: ────X─────@────@─────M─────── 
            │     │    │              
_anc_1: ────┼X────X────┼@──────────── 
            ││         ││             
_anc_2: ────┼┼─────────X┼──────────── 
            ││          │             
_anc_3: ────┼┼──────────┼────X─────── 
            ││          │    │        
a: ─────────@┼────M─────┼────┼─────── 
             │          │    │        
b: ──────────@────M─────┼────┼─────── 
                        │    │        
c: ─────────────────────X────@───M─── 
           └──┘       └──┘            
```

maffoo:
Setting attributes on operations like `setattr(op, "icm_op_id", OperationId(nr_op))` or `meas.icm_op_id = cnot.icm_op_id.advance_decomp()` goes against the grain of the design of cirq, so I'd suggest changing that. The state would have to be either stored in tags (`op = op.with_tags(OperationId(nr_op))`) or else the auxiliary information would have to be stored externally, e.g. in something like a `Dict[Tuple[int, cirq.Operation], OperationId]` that stores opid's keyed on tuples of `(moment_idx, op)` where the moment index uniquely identifies which occurrence of the op in the circuit we are referring to, even if the same operation object is used.

I just worry that even if we change the 3-qubit decomposition to emit unique operation instances, there is no guarantee that other circuit manipulation routines or ways of constructing circuits manually will preserve this uniqueness, so a compiler that relies on uniqueness is likely to run into the same issue in the future.

madhavkrishnan:
@maffoo thanks I think I will go ahead and do that. Shall I go ahead and close this PR since that covers everything?

maffoo:
@madhavkrishnan, yes, please close this. Thanks for the discussion!

